PR #2570, opened 2015-05-01
Document more functions in Paths_pkgname, fixes #715

### LINKED ISSUE #715: Functions in Paths_pkgname module need documenting in the user guide
(Imported from [Trac #724](http://hackage.haskell.org/trac/hackage/ticket/724), reported by @tommd on 2010-08-13)

Some libraries and programs have a large number of files.  Instead of having explicit knowledge of each file (and instead of using a mutation on the result of getDataFilename) there should be a "getDataDirectory" or a "getDataDirectoryContents":

<blockquote>
getDataDirectoryContents = getDataDirectory &gt;&gt;= getDirectoryContents

</blockquote>

ezyang:


23Skidoo:
Merged, thanks!

PR #2574, opened 2015-05-04
Use PVP-compliant example when warning about missing base upper bound

bergmark:
Refs haskell/hackage-server#253.


23Skidoo:
Merged, thanks!

bergmark:
Yay! What an exciting first Cabal commit.

PR #2581, opened 2015-05-07
cabal-install-1.18 doesn't build on GHC 7.10

bergmark:
```
Distribution/Client/Tar.hs:679:10:
    No instance for (Applicative Partial)
      arising from the superclasses of an instance declaration
    In the instance declaration for ‘Monad Partial’
xcabal: Error: some packages failed to install:
cabal-install-1.18.1.0 failed during the building phase. The exception was:
ExitFailure 1
```


bergmark:
PS. I put upper bounds on existing versions.

23Skidoo:
Merged, thanks! Also updated the version bounds on Hackage.

Issue #1506, opened 2013-09-13
Optionally add the revison number to Paths_$pkgid.version


23Skidoo:
It'd be nice if `cabal build` could automatically add the revision number for the appropriate SCM to the `versionTags` field of `Paths_$pkgid.version`. I envision the following usage scenario:

```
$ cabal configure --revision-number # or --build-id, or whatever
$ cabal run -- --version
1.0-0a4f3c6g
$ cabal configure
$ cabal run -- --version
1.0
```

For extracting the revision number from a working copy there is the [vcs-revision](http://hackage.haskell.org/package/vcs-revision) package on Hackage (only SVN, Git and Mercurial support ATM).

23Skidoo:
Related: #3909.

Issue #1319, opened 2013-05-07
how to define a MIN_VERSION_Cabal(x,y,z) macro?


cartazio:
I'm trying to get the LLVM library bindings to build on both new and old Cabal versions... and I'm finding myself unable to do the "obvious" conditional compilation thing here.

Is this a known quirk / problem? 

Whats good workaround?

23Skidoo:
`setup-depends` [are now implemented](http://www.well-typed.com/blog/2015/07/cabal-setup-deps/).

PR #2584, opened 2015-05-08
Replacement for VersionTags

zilinc:
see :
https://ghc.haskell.org/trac/ghc/ticket/2496#comment:19
https://ghc.haskell.org/trac/ghc/ticket/2496#comment:20

### DIFF
--- a/Cabal/Distribution/PackageDescription.hs
+++ b/Cabal/Distribution/PackageDescription.hs
@@ -168,6 +168,10 @@ data PackageDescription
         sourceRepos    :: [SourceRepo],
         synopsis       :: String, -- ^A one-line summary of this package
         description    :: String, -- ^A more verbose description of this package
+        -- | Extra notes about this package. The notes can be used as
+        -- a replacement for the old VersionTags. This field will be
+        -- exported to the autogen/Paths_*.hs module. (See GHC #2496)
+        notes          :: [String],
         category       :: String,
         customFieldsPD :: [(String,String)], -- ^Custom fields starting
                                              -- with x-, stored in a
@@ -255,6 +259,7 @@ emptyPackageDescription
                       sourceRepos  = [],
                       synopsis     = "",
                       description  = "",
+                      notes        = [],
                       category     = "",
                       customFieldsPD = [],
                       library      = Nothing,
--- a/Cabal/Distribution/Simple/Build/PathsModule.hs
+++ b/Cabal/Distribution/Simple/Build/PathsModule.hs
@@ -110,7 +110,10 @@ generate pkg_descr lbi =
         "catchIO = Exception.catch\n" ++
         "\n"++
         "version :: Version"++
-        "\nversion = Version " ++ show branch ++ " " ++ show tags
+        "\nversion = Version " ++ show branch ++ " " ++ show tags++"\n"++
+        "\n"++
+        "notes :: [String]"++
+        "\nnotes = " ++ show (notes pkg_descr)++"\n"
           where Version branch tags = packageVersion pkg_descr
 
        body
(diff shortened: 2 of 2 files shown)


23Skidoo:
Related: #1506.

23Skidoo:
Why do you need this in Cabal when you can just do:

```
module Notes (notes) where

notes :: String
notes = ...
```

in your project ?

zilinc:
@23Skidoo I basically want to achieve the same thing as #1506, but more general and user-customisable. I don't see how the `Notes` modules accomplish this. Could you elaborate?

23Skidoo:
#1506 requires quite a lot more than what you've implemented. Right now your code is just a roundabout way to export a string from a module.

zilinc:
@23Skidoo Well, what I implemented alone doesn't do any of #1506, but at in my case, I used to do 
```
buildHook pkg buildInfo hooks flags = do
    let pId = package pkg
         ver = pkgVersion pId
    hash <- readProcess "git" ["rev-parse", "--short=10", "HEAD"] []
    let hash' = init hash  -- get rid of the newline
         newVer = ver { versionTags = [hash'] }
         newPId = pId { pkgVersion = newVer }
         newPkg = pkg { package = newPId }
    buildHook simpleUserHooks newPkg buildInfo hooks flags
```
in Setup.hs. With `versionTags` deprecated, I can just do it in the same way as before. It essentially just moves the versionTags field outside Version datatype.

zilinc:
Using it for revision control is just the motivation. This pull request is in fact to provide a field for any extra info associated with the package (which also covers the usage of `versionTags`), with git hash being an instance.

23Skidoo:
But you could just generate a `Notes` module in your `buildHook`, there's no need to add a field for this.

zilinc:
I do agree with you (@23Skidoo) that users can generate their own modules using the hook. But nothing would be absolutely necessary. I think the matter is how often/many people are using that field (c.f. https://ghc.haskell.org/trac/ghc/ticket/2496#comment:21). On the other hand however, cabal's pkg-descr does lack of a field for extra info.

23Skidoo:
I just don't see why you need this to be a part of the package description if you won't be editing the value of this field by hand in the `.cabal` file. I also doubt the usefulness of "notes" given its vague semantics ("a metadata string in a format that is different from package to package") - automated tools won't be able to process it in a useful way.

So, I'm -1 on this.

Note that we already have a `source-repository` stanza, which looks like a more appropriate place for revision information.

/cc @dcoutts

zilinc:
@23Skidoo I'm personally neutral on it. It doesn't stall my work. I was just asking for broader options.

zilinc:
I have a seemingly similar problem to #1319 . I tried to generate a file alongside Paths_<pkg> from Setup.hs (due to the de facto rejection of this PR) but I need duplicate some work done in `Distribution.Simple.Build`. As far as I can see, Paths_<pkg> is generated by `writeAutogenFiles` function, whose upper level functions do the preparation, e.g. create the autogen directory. I can't think of an easy way to overwrite `writeAutogenFiles` function along. What's related to #1319 is, however, a different problem. Apart from the git revision I wanted, I'd also like to have build time info. I use Data.Time and its function `defaultTimeLocale`, which is exported from different packages before and after time-1.5, with the new one from time but the old one from old-locale. So that I'd like to achieve some conditional compilation of Setup.hs. So if I understand correctly, there's no support right now for this. Am I correct?

23Skidoo:
> So if I understand correctly, there's no support right now for this. Am I correct?

Sorry for getting back to you so late. We now have [setup dependencies](http://www.well-typed.com/blog/2015/07/cabal-setup-deps/).

23Skidoo:
I think I'll close this one, but if someone is willing to implement #1506 -- patches welcome.

PR #2600, opened 2015-05-20
Provide cabal_macros in c2hs invocation.

qnikst:
This commit includes cabal_macros in c2hs call, so it will
be possible to use MIN_VERSION macros in chs files.


23Skidoo:
LGTM. Can you please also add a note to the changelog?

qnikst:
Added in the same commit.

PR #2603, opened 2015-05-22
Fix sandbox delete-source error message typo

triplepointfive:
\a is considered as a bell character, which is omitted as a result. The error message is `The 'sandbox delete-source' command expects t least one argument`


23Skidoo:
Multiline string literals work correctly for me in GHCi in this case:

```
module Main
    where

main :: IO ()
main = putStrLn $ "foo \
                  \and bar"
```

```
*Main> :main
foo and bar
```

23Skidoo:
Aha, it's due to CPP. OK, merging.

23Skidoo:
Thanks for your patch!

Issue #1360, opened 2013-06-06
Impossible to delete a non-existent source


UnkindPartition:
If a directory was add-sourced and then removed (as in `rm -rf`), `cabal sandbox delete-source` would refuse to delete it from the list of sources.

```
cabal: $DIRNAME: does not exist
```

PR #2605, opened 2015-05-22
sandbox delete-source removes only registered sources

### LINKED ISSUE #2155: cabal sandbox delete-source gives no indication of whether it did anything
`Cabal sandbox delete-source` should fail if it is given a non-existant source. At the moment it simply pretends all is well,
```
$ cabal sandbox delete-source this-is-not-a-source
Note: 'sandbox delete-source' only unregisters the source dependency, but does
not remove the package from the sandbox package DB.

Use 'sandbox hc-pkg -- unregister' to do that.
$ echo $?
0
```

triplepointfive:
This resolves #2155
To prevent #1360 from happening again `delete-source` checks only for registered sources. It still doesn't care whether input dirs exist or not.


ccressent:
Can this be merged to close off #2155?

ezyang:
Needs a test, and then I'll merge it.

martinvlk:
This has been fixed in #2931 - can now be closed.

PR #2608, opened 2015-05-24
Get 'builddir' from cabal.config or CABAL_BUILDDIR

### LINKED ISSUE #2484: Should be possible to define builddir in `cabal.config`
Currently the only way we can pass `builddir` is through a command line option. Since I use different build directories for dfiferent versions of GHC, it would be very convenient if I could specify 

```
builddir: <something>
```

in my project-local `cabal.config` file, and then specify a different `cabal.config` file using the `CABAL_CONFIG` environment variable (`cabal.config-7.8.4`, say). Currently specifying `builddir` in config file is explicit disallowed (`configFieldDescriptions` in `Config.hs`); I tried removing this and that stopped cabal from complaining about an unrecognized `builddir` option, but it still didn't take it into account either. I don't know cabal's command line/config option system well enough to fix this.

ttuegel:
Fixes #2484. The 'builddir' option may now be specified in cabal.config
as well. It will also be read from the CABAL_BUILDDIR environment
variable, if set. The order of precedence (highest to lowest) is:
1. --builddir command line option
2. CABAL_BUILDDIR environment variable
3. cabal.config setting

I've been using this myself for about a week and everything works as intended.


edsko:
Just tried this and it seems to work for me. Good stuff!

23Skidoo:
Travis says that it fails on 7.4.2:

```
Distribution/Simple/Configure.hs:151:29:

    Module `System.Environment' does not export `lookupEnv'
```

Looks like you need to move `D.Client.Compat.Environment.lookupEnv` to `D.Compat.Environment`.

ttuegel:
Oh, rats, I forgot we can't use `MIN_VERSION_foo` in Cabal.

23Skidoo:
LGTM except the loading-the-config-twice bit.

ttuegel:
This needs extra testing on Windows because I had to pull in some platform-specific stuff from `System.Environment`.

PR #2609, opened 2015-05-24
Document 'extensions' field replacements

### LINKED ISSUE #2401: Warning directs use of deprecated "extensions" field; its replacement is not documented
When I put "-XTemplateHaskell" in the "ghc-options" field a warning tells me that I should instead add "TemplateHaskell" to the "extensions" field.  I did this, and another warning told me that I should instead be using "default-extensions" and "other-extensions".  Neither "default-extensions" nor "other-extensions" is documented in "Developing Cabal Packages".

AngusL:
Resolves #2401.

### DIFF
--- a/Cabal/doc/developing-packages.markdown
+++ b/Cabal/doc/developing-packages.markdown
@@ -1334,20 +1334,21 @@ values for these fields.
     For backwards compatibility, the old variant `hs-source-dir` is also
     recognized.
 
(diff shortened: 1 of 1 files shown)


23Skidoo:
Merged, thanks! By the way, you can mark documentation-only changes with `[ci skip]` in  the body of the commit message to avoid unnecessary Travis builds.

Issue #2083, opened 2014-08-31
cabal haddock does not notice flag change


nh2:
In a package (quickcheck) we use something like:

```
{-# LANGUAGE CPP #-}

#ifdef GENERICS
-- | Random generation and shrinking of values. Can be auto derived with Generic.
#else
-- | Random generation and shrinking of values.
#endif
class Arbitrary a where ...
```

and in the cabal file

```
  if impl(ghc >= 7.2)
    cpp-options: -D_GENERICS
```

If you change the ghc version number in there, e.g. to 7.100, `cabal haddock` doesn't notice the change and passes the old flag. If you call `cabal configure` first, everything is fine.

I believe `cabal haddock` it should perform the same "did the cabal file change"-check that `cabal build` does. Like this:

```
./QuickCheck.cabal has been changed. Re-configuring with most recently used
options. If this fails, please run configure manually.
```

mkawalec:
I'll be working on this for the next two days.

dminuoso:
This was resolved in https://github.com/haskell/cabal/pull/2615

23Skidoo:
Closing as fixed by #2615

PR #2622, opened 2015-05-30
Haddock action reconfigures if need be

mkawalec:
Implements https://github.com/haskell/cabal/issues/2083. 

I'm unsure if `DontSkipAddSourceDepsCheck` should be used in a call to `reconfigure`, but `buildExFlags` is not available to `haddockAction`, so I went with a more conservative parameter (addSourceDepsCheck is always executed).


mtolly:
This appears to be *exactly* the same as #2615 :)

23Skidoo:
Closing in favour of #2615, which is already merged.

PR #2626, opened 2015-05-30
Only set rpath on dynamic executables

### LINKED ISSUE #2625: `RPATH` uselessly set on non-dynamic libraries
By default, Cabal builds binaries that statically link the used Haskell libraries. Nevertheless, the path to those libraries is encoded in  the `RPATH` section of the binary.

This is ugly, presumably increases startup time (scanning 96 directories when loading linked C libraries) and violates distribution policies (https://lintian.debian.org/tags/binary-or-shlib-defines-rpath.html).

TL;DR: When building executable binaries, do not set an rpath.

nomeata:
not static ones. They make no sense there, increase startup time and
may violate some distributions policy on rpaths. This closes #2625.


23Skidoo:
LGTM.

PR #2632, opened 2015-05-31
Allow using cabal program itself as the external setup method

tibbe:
This fixes issues when the version of Cabal that cabal-install was built
against differs from the one registered in the local package DB. Normally
we compile an external setup against the local Cabal library, which could
lead to failures or inconsistent results compared to using the internal
method.


23Skidoo:
Good idea.

tibbe:
Closing in favour of #2633.

PR #2604, opened 2015-05-22
Implement init-config 

psibi:
See #2553

Issue #2553, opened 2015-04-24
Add a command to just create a config file


snoyberg:
I've seen people ask questions in the past about how to create a ~/.cabal/config file. The answer is typically "run `cabal update`", but in some circumstances, that's not the optimal action (e.g., when you're going to change immediately to a different remote-repo). What about adding a command (strawman: `cabal init-config`) that only creates the ~/.cabal/config file, and only when missing? (Optional: I suppose it could take a flag to force it to overwrite an existing file, but I have no need of such a feature.)

yogsototh:
In my script I do a "cabal info" which only create a ~/.cabal/config file. I modify it ASAP to use Haskell LTS. On the other hand I would I found nice to have a command to pass some basic init flags, like changing the remote-repo or having profiling activated by default.

23Skidoo:
This has been fixed by #3095, by the way.

PR #2640, opened 2015-06-02
Added command 'init-config'

tvestelind:
init-config creates a default config file if it doesn't already exist.

If --config-file is set, then that file will be written.
If -f or --force is used, then the file will be overwritten if it
already exists.


mydaum:
As requested in #2553.
After implementing everything almost to completion I saw that there was already a pull request for that issue. As I had already done so much I decided to finish it (also for learning purposes) and you can decide which of the pull requests you want.

23Skidoo:
Can we make it a subcommand of `user-config`?

mydaum:
Sure, if you want to. Can you give me a couple of use cases on how you would like it work?

23Skidoo:
I only want the command name to be changed to `user-config init`, everything else should work in the same way. We already have some functionality for managing config files under `user-config`, so it feels like a logical place to put it.

23Skidoo:
Code looks good, by the way.

mydaum:
Ok, I will fix that. I have some stuff to do the coming days but I will try to squeeze it in.

Thank you! :)

23Skidoo:
Thank you for working on this!

cabal-install/tests/PackageTests/InitConfig/Check.hs
```diff
@@ -0,0 +1,46 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module PackageTests.InitConfig.Check
```
tibbe: Optional: could this be written as a unit test instead? Package tests are really slow (because we need to actually compile code) and I'm trying to avoid them. You can still do I/O in a unit test. In this case you could have the test call `createDefaultConfigFile` or `initConfigAction` directly.

You can use `Distribution.Simple.Utils.withTempDir` to create a temporary directory instead of using `packageTestsDirectory`. Using a non-temp dir risks issues e.g. if we run tests in parallel. Example test:

```haskell
readSourcePackagesInDirTest :: Assertion
readSourcePackagesInDirTest = do
  sysTmpDir <- Directory.getTemporaryDirectory
  Utils.withTempDirectory silent sysTmpDir "cabal-test" $ \ tmpDir -> do
    let pkgDir = tmpDir </> "a"
    Directory.createDirectory pkgDir
    -- Write a trivial cabal file.
    writeFile (pkgDir </> "a.cabal") $ unlines [
      "name: a",
      "version: 1.2.3.4"
      ]
    index <- SourceTree.readSourcePackagesInDir silent tmpDir
    ...
    assertBool "Package should exist in index." $
      PackageIndex.elemByPackageId index pkgid
```

mydaum: Sure, I think I should be able to do that.

mydaum: I'm looking at writing the unit tests. To be able to call initConfigAction I need to export it in Main.hs right? I could do that but it feels wrong :P
Is there another way of doing it or what do you think about exporting it?

tibbe:
How about calling `createDefaultConfigFile` and having that in a different module? I don't want to make you jump through hops, so if it's too difficult I won't insist.

mydaum:
That's no problem. I want the solution to be nice and feel right.
Right now I have one unit test but only for createDefaultConfigFile, as I had already exported that from Distribution.Client.Config it was easy to import and call.
I'm not sure that I understand what you mean. Do you think I should move some of the logic from initConfigAction to createDefaultConfigFile?

tibbe:
I guess we could have a unit test for createDefaultConfigFile and a package
test for initConfigAction. I will look into refactoring later (probably by
moving the actions out of main.)

mydaum:
Ok, I will do that and upload a new patch tomorrow or during the weekend and then we can reason about that.

bgamari:
I guess this is still waiting for tests?

mydaum:
Are you waiting for me to do something? In that case I'm not sure what I should do..

cabal-install/tests/UnitTests/Distribution/Client/UserConfig.hs
```diff
@@ -80,6 +83,19 @@ doubleUpdateConfig = bracketTest . const $ do
         listUnique (map show . fromNubList . installSummaryFile $ savedInstallFlags updated)
 
 
+newDefaultConfig :: Assertion
+newDefaultConfig = do
+    sysTmpDir <- getTemporaryDirectory
+    withTempDirectory silent sysTmpDir "cabal-test" $ \tmpDir -> do
+        let configFile  = tmpDir </> "tmp.config"
+        createDirectoryIfMissing True tmpDir
```
tibbe: This should never happen.

tibbe:
:shipit:

cabal-install/tests/UnitTests/Distribution/Client/UserConfig.hs
mydaum: Yes, that's true. I think I used to create another directory inside 'tmpDir' before writing the config file to that directory. I thought that doing that was unnecessary so I removed it and it seems like I forgot to remove createDirectoryIfMissing. I can fix it tonight if you want.
If you want to just use what's here then I could fix it in a separate commit and do a fork request on that. Is that ok?

tibbe: Either way is fine with me. I will leave it to one of the Cabal maintainers
(e.g. Ryan or Mikhail) to merge.

gbaz:
:+1:

ttuegel:
@mydaum If you could resolve the merge conflicts, I would be happy to merge this.

mydaum:
Hi,

Ok, I will try to do it tomorrow :)

On September 20, 2015 2:27:17 PM GMT+02:00, Thomas Tuegel <notifications@github.com> wrote:
>@mydaum If you could resolve the merge conflicts, I would be happy to
>merge this.
>
>---
>https://github.com/haskell/cabal/pull/2640#issuecomment-141782434

-- 

ezyang:
If you merge this, please close #2604

mydaum:
Sorry, I completely forgot about this. Is it up to me to do something or is the feature no longer needed?

23Skidoo:
Closing in favour of #3095.

PR #2649, opened 2015-06-07
Distribution.Simple.Setup: fix HaddockFlags mappend

taruti:
mappend of HaddockFlags uses the wrong function. Simple one line fix.


23Skidoo:
Merged, thanks!

PR #2652, opened 2015-06-10
Parse C compiler flags from GHC correctly

### LINKED ISSUE #2292: Cabal does not parse C compiler flags correctly
`ghc --info` reports the C compiler flags as a single string, but `configureToolChain` expects a list, which means the reported values get ignored:

https://github.com/haskell/cabal/blob/4aabb4f4b1c7fb6446ce619ddfdfa0d19516c721/Cabal/Distribution/Simple/GHC/Internal.hs#L149-L151

This should be fixed before #1910 is usable.

bgamari:
GHC hands us a `String` which we previously tried to parse as a `[String]`, causing GHC's compiler flags to be ignored.

It seems this is the result of an inconsistency between encodings of the settings exposed by `ghc --info`. In some cases (e.g. `Gcc Linker flags`) the flags are encoded as a `[String]` with `show` whereas in other cases (e.g. `C compiler flags`) the flags are just a `String`. We work around this by first attempting to parse a list, then falling back on tokenizing the string if this fails.

This fixes #2292.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -237,6 +237,7 @@ library
     Distribution.Compat.CopyFile
     Distribution.Compat.TempFile
     Distribution.GetOpt
+    Distribution.Lex
     Distribution.Simple.GHC.Internal
     Distribution.Simple.GHC.IPI641
     Distribution.Simple.GHC.IPI642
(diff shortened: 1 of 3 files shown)


bgamari:
@23Skidoo ping.

Issue #2629, opened 2015-05-31
Remove the incorrect warning "The package is missing a Setup.hs or Setup.lhs script."


nikita-volkov:
```
$ cabal sdist
Distribution quality errors:
The package is missing a Setup.hs or Setup.lhs script.
Note: the public hackage server would reject this package.
```

What's stated there is simply untrue: the public hackage server does not reject such packages. It's been there forever, but this time it got annoying.

23Skidoo:
It's not incorrect for build types other than `Simple`.

nikita-volkov:
Well, okay. But it still is incorrect in case of the `Simple` build type. Can't it be fixed with a precondition?

23Skidoo:
Fixed by #2657.

Issue #1125, opened 2012-11-20
"check" warns for missing Setup.hs, but "sdist" adds one anyway


dag:
The warning therefore seems redundant to me.

eightyeight:
I assume the fix involves removing this check? Or only performing the check if the distribution type is not `Simple`?

phadej:
https://github.com/haskell/cabal/issues/2629 is related

23Skidoo:
Fixed by #2657.

PR #2657, opened 2015-06-16
Only warn for missing Setup.hs if build type isn't Simple

eightyeight:
Because `cabal sdist` will add a default `Setup.hs` anyway. See #1125 and #2629.


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

eightyeight:
Happy to help!

PR #2664, opened 2015-06-18
Run package tests with cabal executable in build directory

grayjay:
The test "configures cabal to use the sandbox" runs the command `cabal sandbox hc-pkg list`.  That cabal may not be the development version.

When I looked into fixing the test, I noticed an inconsistency in the rest of the test suite.  The package tests locate cabal by searching both dist/build/cabal and the default program search path.  However, main also calls `canonicalizePath "dist/build/cabal"`, which throws an exception when the path doesn't exist.

Which solution is preferable?

1. Hard code the path as "dist/build/cabal/cabal" for all of the cabal-install package tests.
2. Search the default program search path if cabal is not in the build directory.  This would allow people to build cabal in other ways, such as installing it in a sandbox (See https://github.com/haskell/cabal/commit/3f99cc0c8d4b153f4341685147a329781f767).

This pull request implements the first option, but I'd be happy to change it.

### DIFF
--- a/cabal-install/tests/PackageTests/Exec/Check.hs
+++ b/cabal-install/tests/PackageTests/Exec/Check.hs
@@ -102,7 +102,12 @@ tests paths =
 
           assertMyLibIsNotAvailableOutsideofSandbox paths libNameAndVersion
 
-          result <- cabal_exec paths dir ["bash", "--", "-c", "cd subdir ; cabal sandbox hc-pkg list"]
+          result <- cabal_exec paths dir
+                    [ "bash"
+                    , "--"
+                    , "-c"
+                    , "cd subdir ; ../../../../dist/build/cabal/cabal sandbox hc-pkg list"
+                    ]
           assertExecSucceeded result
           let output = outputText result
               errMsg = "my library should have been found"
--- a/cabal-install/tests/PackageTests.hs
+++ b/cabal-install/tests/PackageTests.hs
@@ -11,7 +11,7 @@ import Distribution.Simple.Program.Builtin (ghcPkgProgram)
 import Distribution.Simple.Program.Db
         (defaultProgramDb, requireProgram, setProgramSearchPath)
 import Distribution.Simple.Program.Find
-        (ProgramSearchPathEntry(ProgramSearchPathDir), defaultProgramSearchPath)
+        (ProgramSearchPathEntry(ProgramSearchPathDir))
 import Distribution.Simple.Program.Types
         ( Program(..), simpleProgram, programPath)
 import Distribution.Simple.Utils ( findProgramVersion )
@@ -54,7 +54,7 @@ cabalProgram = (simpleProgram "cabal") {
 main :: IO ()
 main = do
     buildDir <- canonicalizePath "dist/build/cabal"
-    let programSearchPath = ProgramSearchPathDir buildDir : defaultProgramSearchPath
+    let programSearchPath = [ProgramSearchPathDir buildDir]
     (cabal, _) <- requireProgram normal cabalProgram
                       (setProgramSearchPath programSearchPath defaultProgramDb)
     (ghcPkg, _) <- requireProgram normal ghcPkgProgram defaultProgramDb


cabal-install/tests/PackageTests/Exec/Check.hs
```diff
@@ -102,7 +102,12 @@ tests paths =
 
           assertMyLibIsNotAvailableOutsideofSandbox paths libNameAndVersion
 
-          result <- cabal_exec paths dir ["bash", "--", "-c", "cd subdir ; cabal sandbox hc-pkg list"]
+          result <- cabal_exec paths dir
+                    [ "bash"
+                    , "--"
+                    , "-c"
```
grayjay: I didn't use the path in `paths` because it contains backslashes on Windows.

cabal-install/tests/PackageTests/Exec/Check.hs
```diff
@@ -102,7 +102,12 @@ tests paths =
...
+                    , "cd subdir ; ../../../../dist/build/cabal/cabal sandbox hc-pkg list"
```
grayjay: I didn't use the path in paths because it contains backslashes on Windows.

ezyang:
The pile of ../../.. surely cannot be right. Maybe #2797 will obsolete this?

BardurArantsson:
I have to dash right now, but I'm pretty sure there's definitely a better way of doing this even without #2797 making it in. (I'll also try to port this test over in #2797 just to have another non-trivial test case.)

I'll try to comment on this later.

BardurArantsson:
Well, I can at least confirm that #2797 will fix this problem automatically :).

It probably doesn't make sense to spend any time on this particular issue until that's done... which will hopefully be some time in the next few days.

grayjay:
I tried out the branch for https://github.com/haskell/cabal/issues/2797 on Windows.  I had to change the path to `sh`, but after that, the tests found the correct version of cabal.  There is still a problem with backslashes in the path that is given to bash, though: https://github.com/BardurArantsson/cabal/blob/5d513ff92a59f0100df827355ca759fb55d15149/cabal-install/tests/IntegrationTests/exec/should_run/configures_cabal_to_use_sandbox.sh#L14 .

BardurArantsson:
(I don't have any way of testing on Windows myself. so this is invaluable feedback, thank you!)

The "bash" vs. "sh" is a simple oversight -- I just copy/pasted the test and adapted it... but forgot to change "bash" to "sh". (I think I'm going to go with just "sh" for all the test scripts, seems like the safest bet that "sh" is basically always going to be installed.)

Re: backslashes: They probably come from the `programPath` invocation which happens in `main` in `IntegrationTests.hs`. (I'm guessing there's some kind of `canonicalizePath` that's ultimately responsible for them and they end up in the configuration file with backslahses in since that's the preferred way on Windows. AFAIK forward slashes should also work.). I'm sure we can do something to fix it, but I'm not sure exactly what the right fix would be at the moment. (A similar thing will happen with $GHC_PKG, btw.)

grayjay:
I meant that I changed `"/bin/sh"` to a Windows-style path (`"C:\\Program Files\\Git\\bin\\sh"`) here: https://github.com/BardurArantsson/cabal/blob/5d513ff92a59f0100df827355ca759fb55d15149/cabal-install/tests/IntegrationTests.hs#L171 .  `run` calls `canonicalizePath`, so `/bin/sh` is converted to `C:\bin\sh`.  Even without the call to `canonicalizePath`, `runProcess` can't find `/bin/sh`.  I'm not sure how to fix it.

I also had to make two smaller changes before the other tests passed:

* This test should allow `cabal.exe` or `cabal` in the error message:
https://github.com/BardurArantsson/cabal/blob/5d513ff92a59f0100df827355ca759fb55d15149/cabal-install/tests/IntegrationTests/exec/should_fail/exit_with_failure_without_args.err#L1

* This line compares program output to a text file, but the line endings differ: https://github.com/BardurArantsson/cabal/blob/5d513ff92a59f0100df827355ca759fb55d15149/cabal-install/tests/IntegrationTests.hs#L147

BardurArantsson:
OK, thanks for the follow-up. I'll see what I can do to address all of these issues in #2797.

I'm not sure what we can do about /bin/sh, but it might be feasible to just do a $PATH search first. (Or, absolute worst case, allow the shell to be specified via an env variable?)

grayjay:
@BardurArantsson No problem.  Searching PATH worked for me when I called `runProcess` with `sh`.

BardurArantsson:
@grayjay : Can you please check if everything runs OK for you in current master? If so, please just close this.

grayjay:
It works.  Thanks!

PR #2665, opened 2015-06-19
Test cabal exec error message in a more portable way

grayjay:
This change allows the tests "can run executables installed in the sandbox" and
"adds the sandbox bin directory to the PATH" to run on Windows by removing the
executable name from the required error message.


23Skidoo:
Merged, thanks!

PR #2666, opened 2015-06-19
Compare file paths with equalFilePath for portability

grayjay:
Without this change, differences in filename capitalization caused `cabal sandbox delete` to fail with this error message when used with a default sandbox on Windows:

    cabal.exe: Non-default sandbox location used:
    'C:\folder\.cabal-sandbox'.
    Assuming a shared sandbox. Please delete
    'C:\folder\.cabal-sandbox' manually.

This commit also allows the cabal-install package tests that delete sandboxes to run on Windows.


23Skidoo:
Merged, thanks!

PR #2669, opened 2015-06-20
Allow cabal freeze tests to run on Windows

grayjay:
The tests now read cabal.config strictly so that the file is closed by the time
it is deleted for the next test.


23Skidoo:
LGTM, merging.

PR #2671, opened 2015-06-20
Drop pkgna_ prefix from package keys, c.f. http://ghc.haskell.org/tra…

ezyang:
…c/ghc/ticket/10550


23Skidoo:
Looks like this makes Cabal unable to parse output of 7.10's `ghc-pkg`.

ezyang:
Yes, I have an adjusted patch coming up.

ezyang:
OK, this patch continues to parse `ghc_AAAAA` format package keys and preserves them; it just switches to generating `AAAAA` package keys by default.

PR #2675, opened 2015-06-23
Add a stack.yaml file

snoyberg:
I needed this when building cabal.exe for MinGHC (see:
https://github.com/fpco/minghc/issues/69). I thought others might find
this useful, especially for the use case of bootstrapping cabal
development.

I went the direction of no curation and explicit packages in the
extra-deps, as I thought that would make more sense for a project low on
the dependency chain, but this could certainly be tweaked to use LTS
Haskell or Stackage Nightly instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ Cabal/ghc.mk
 # TAGS files
 TAGS
 tags
+
+# stack artifacts
+/.stack-work/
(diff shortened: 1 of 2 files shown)


23Skidoo:
Merged, thanks!

snoyberg:
Thank you, that was fast!

PR #2676, opened 2015-06-24
HACKING.md: Add quotation to sed command

BardurArantsson:
The command line given to get the sandbox package database path is missing quotation which may make it fail in some shells (given certain globbing options). This adds single quotation marks to avoid such potential issues.


23Skidoo:
Merged, thanks!

PR #2680, opened 2015-06-25
Exit with an error if $CABAL_SANDBOX_CONFIG does not point to extant file

### LINKED ISSUE #2224: CABAL_SANDBOX_CONFIG error handling
I was playing around with a script to auto-update the CABAL_SANDBOX_CONFIG to the current sandbox and display the current sandbox in my PS1 variable (if any sandboxes are encountered from the current directory to the root), and encountered what I think is a bug in cabal. If the CABAL_SANDBOX_CONFIG variable is set to an empty string, a lot of commands fail in a very uninformative manner. For instance

$ export CABAL_SANDBOX_CONFIG=''
$ cabal sandbox init
Writing a default package environment file to
cabal: .tmp: does not exist
$ cabal init
cabal: .: interrupted
$ cabal sandbox delete
cabal: The package environment file '' doesn't exist

Mostly my fault since I'm mucking about things, and had assumed the empty string was synonymous with no variable in bash, but I think more informative error messages could be helpful in any case.

BardurArantsson:
--

I think this fixes #2224. I haven't done extensive manual testing, but the tests all (obviously) still pass.

Example output:

cabal: Cabal sandbox file in $CABAL_SANDBOX_CONFIG does not exist: foo

the "foo" is the content of the env variable.

I should note that the error occurs quite "aggressively", i.e. when the settings are read, and not when they are actually used -- so even things like "cabal info" will fail if the env variable is invalid. I think a more relaxed would be a lot more error-prone since it would be easy to miss cases.

### DIFF
--- a/cabal-install/Distribution/Client/Sandbox.hs
+++ b/cabal-install/Distribution/Client/Sandbox.hs
@@ -102,6 +102,7 @@ import Control.Exception                      ( assert, bracket_ )
 import Control.Monad                          ( forM, liftM2, unless, when )
 import Data.Bits                              ( shiftL, shiftR, xor )
 import Data.Char                              ( ord )
+import Data.Foldable                          ( forM_ )
 import Data.IORef                             ( newIORef, writeIORef, readIORef )
 import Data.List                              ( delete, foldl' )
 import Data.Maybe                             ( fromJust )
@@ -170,7 +171,11 @@ updateSandboxConfigFileFlag globalFlags =
   case globalSandboxConfigFile globalFlags of
     Flag _ -> return globalFlags
     NoFlag -> do
-      f' <- fmap (maybe NoFlag Flag) . lookupEnv $ "CABAL_SANDBOX_CONFIG"
+      fp <- lookupEnv "CABAL_SANDBOX_CONFIG"
+      forM_ fp $ \fp' -> do      -- Check for existence if environment variable set
+        exists <- doesFileExist fp'
+        unless exists $ die $ "Cabal sandbox file in $CABAL_SANDBOX_CONFIG does not exist: " ++ fp'
+      let f' = maybe NoFlag Flag fp
       return globalFlags { globalSandboxConfigFile = f' }
 
 -- | Return the path to the sandbox config file - either the default or the one


BardurArantsson:
Hm. I see there are some actual possibly-real failures here. I'll investigate tomorrow.

BardurArantsson:
Ah, yes, FTP.

Alright, I think is should be good to go.

edsko:
I've submitted a PR that undoes this commit, because it makes `cabal` completely unusable when `CABAL_SANDBOX_CONFIG` is set. This check happens much too early, so that for instance we get

```
# ../dist/build/cabal/cabal sandbox init
cabal: Cabal sandbox file in $CABAL_SANDBOX_CONFIG does not exist:
./cabal.sandbox.config-7.8.4
```

Moreover, even if we initialize the sandbox using an older `cabal`

```
[7.8.4 ~/wt/projects/tuf/cabal/cabal-install/t]
# ~/opt/cabal/cabal-1.22.0.0 sandbox init
Writing a default package environment file to ./cabal.sandbox.config-7.8.4
Creating a new sandbox at
/Users/e/wt/projects/tuf/cabal/cabal-install/t/.cabal-sandbox
```

trying to install any packages into the sandbox fails:

```
# ../dist/build/cabal/cabal install mtl
Resolving dependencies...
/Users/e/wt/projects/tuf/cabal/cabal-install/t/.cabal-sandbox
Configuring transformers-0.4.3.0...
Failed to install transformers-0.4.3.0
Build log ( /Users/e/wt/projects/tuf/cabal/cabal-install/t/.cabal-sandbox/logs/transformers-0.4.3.0.log ):
cabal: Cabal sandbox file in $CABAL_SANDBOX_CONFIG does not exist:
./cabal.sandbox.config-7.8.4
mtl-2.2.1 depends on transformers-0.4.3.0 which failed to install.
transformers-0.4.3.0 failed during the configure step. The exception was:
ExitFailure 1
```

BardurArantsson:
Absolutely fair enough to just revert. I guess my testing wasn't sufficient :(. I guess this is really a cue that we really need to expand the test suite and to make it easier to add tests.

(I don't think I'll get around to implementing this properly any time soon.)

PR #2683, opened 2015-06-26
Package key is not necessary for ABI calculation.

ezyang:


23Skidoo:
Breaks the Travis build.

ezyang:
Blah, it shouldn't be necessary in principle but it seems GHC uses it to do a sanity check. OK, we can just ditch this commit.

PR #2686, opened 2015-06-28
Add test dependencies to stack.yaml

DanBurton:
So that contributors can run `stack test` if they so desire.


23Skidoo:
Merged, thanks!

PR #2697, opened 2015-07-03
View support in cabal

fugyk:
This pull request adds support of view to cabal. It is to be used with my ghc-pkg repo(https://github.com/fugyk/ghc). While it is not ready for production now, as it needs some rounds of refactoring and misses some features, I just want initial review for major changes. From https://ghc.haskell.org/trac/ghc/wiki/Commentary/GSoC_Cabal_nix, a major change is that default view packages are exposed while all other packages are hidden in the DB. It is to make transition path easy as selecting the package for default view can be tricky, so with this GC can see all exposed package(also which are installed earlier) as root as well. Also there some smaller things like view is passed as config flag etc.

TODO:

* Symlink ./.ghc.environment to view file (it is required for cabal repl to work, but only better as ghc/ghci can be used with environment)
* Add view field in cabal.config and create a default view
* Register the package even if there is nothing to do (all the packages are installed)
* GC
  * Registering placeholder library for --only-dependencies(to prevent dependencies from being GC'd) and executables (It is required if it uses DLL)
  * GC core

/cc @dcoutts @ttuegel @kosmikus @ryantrinkle @ezyang


Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -268,6 +278,43 @@ list hpi verbosity packagedb = do
...
+createView hpi verbosity view =
+    when (supportsView hpi) $ runProgramInvocation verbosity invocation
+  where
+    invocation = programInvocation (hcPkgProgram hpi) args
+    args = ["view", "create", packageDbOpts hpi UserPackageDB] ++
+      case view of
+        Left name -> [name]
+        Right path -> ["--view-file", path]
+
+-- | Adds a package to the given view
```
23Skidoo: Please use proper punctuation.

Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -268,6 +278,43 @@ list hpi verbosity packagedb = do
...
+    removePackageFromView hpi verbosity view pkgid
+    when (supportsView hpi) $ runProgramInvocation verbosity invocation
+  where
+    invocation = programInvocation (hcPkgProgram hpi) args
+    args = ["view-modify", view, "add-package", "--ipid", display ipid, packageDbOpts hpi UserPackageDB]
+    -- init clashes with this module's init.
+    pkgid = List.init . reverse . snd . (span (/= '-')) . reverse . display $ ipid
+
+-- | Removes a package from the given view. As only one instance of package can be in a view
+-- we do not need ipid. TODO: Fourth argument type must be PackageId but there is no suitable InstalledPackageId
```
23Skidoo: Please respect the 80-col limit.

Cabal/Distribution/Simple/Register.hs
```diff
@@ -230,13 +244,32 @@ invokeHcPkg verbosity comp conf dbStack extraArgs =
...
-  case compilerFlavor comp of
-    GHC   -> f (GHC.hcPkgInfo conf)
-    GHCJS -> f (GHCJS.hcPkgInfo conf)
-    LHC   -> f (LHC.hcPkgInfo conf)
+  case getHcPkgInfo comp conf of
+    Just hcPkgInfo -> f hcPkgInfo
     _     -> die ("Distribution.Simple.Register." ++ name ++ ":\
                   \not implemented for this compiler")
 
+getHcPkgInfo comp conf = case compilerFlavor comp of
```
23Skidoo: This needs a type signature, otherwise the build fails on Travis due to `-Wall`:

```
    Top-level binding with no type signature:

      getHcPkgInfo :: Compiler -> ProgramConfiguration -> Maybe HcPkgInfo
```

23Skidoo: Also, please add a Haddock comment.

Cabal/Distribution/Simple/Register.hs
```diff
@@ -230,13 +244,32 @@ invokeHcPkg verbosity comp conf dbStack extraArgs =
...
+                           GHC   -> Just (GHC.hcPkgInfo conf)
+                           GHCJS -> Just (GHCJS.hcPkgInfo conf)
+                           LHC   -> Just (LHC.hcPkgInfo conf)
+                           _     -> Nothing
+
+viewSupported :: Compiler -> ProgramConfiguration -> Bool
```
23Skidoo: Please add Haddock comments to all top-level definitions.

cabal-install/Distribution/Client/Install.hs
```diff
@@ -280,9 +282,9 @@ makeInstallContext verbosity
     return (installedPkgIndex, sourcePkgDb, userTargets, pkgSpecifiers, transport)
 
 -- | Make an install plan given install context and install arguments.
-makeInstallPlan :: Verbosity -> InstallArgs -> InstallContext
-                -> IO (Progress String String InstallPlan)
-makeInstallPlan verbosity
+makeInstallPlans :: Verbosity -> InstallArgs -> InstallContext
```
23Skidoo: Haddock comment needs updating.

fugyk: Its actually worse than this. I cannot just replace the function, if compiler does not support multiple instances (ghc < 7.10). But I was too frustrated of passing HcPkgInfo, Compiler, ProgramConfiguration to every function, so that I ignored it in some of the places to make the working product. I know this pull request is poorly written, but this was just for getting reviews. I can do remaining things latter.

Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -268,6 +278,53 @@ list hpi verbosity packagedb = do
   where
     parsePackageIds = sequence . map simpleParse . words
 
+-- Create a view from either name or symlink via filepath.
```
ttuegel: This should be a Haddock comment `-- |` since it's a top-level definition.

Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -268,6 +278,53 @@ list hpi verbosity packagedb = do
...
+createView :: HcPkgInfo -> Verbosity -> Either String FilePath -> IO ()
```
ttuegel: Maybe views should be represented by an opaque data type, rather than as strings? Then, there is no problem for library users if we need to change the representation later.

cabal-install/Distribution/Client/Sandbox/PackageEnvironment.hs
```diff
@@ -227,14 +241,45 @@ sandboxPackageDBPath sandboxDir compiler platform =
...
                                                       sandboxDir
                                                       compiler
                                                       platform)]
     }
+  where viewSupported = Register.viewSupported compiler conf
+
+
+-- | A general purpose hash function. It is exported as the hash
+-- of sanbox dir needs to be calculated in two different modules.
+-- See http://en.wikipedia.org/wiki/Jenkins_hash_function
```
ttuegel: I assume this is the byte-at-a-time variant of that hash function? Can you add a comment explaining the rationale for this choice of hash function?

phadej: Also typo in the comment: sanbox :)

cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -141,28 +138,7 @@ snapshotDirectoryName = "snapshots"
 -- | Non-standard build dir that is used for building add-source deps instead of
 -- "dist". Fixes surprising behaviour in some cases (see issue #1281).
 sandboxBuildDir :: FilePath -> FilePath
-sandboxBuildDir sandboxDir = "dist/dist-sandbox-" ++ showHex sandboxDirHash ""
```
phadej: @ttuegel the has function is moved from here i.e. not introduced in this patch. Though no idea why this choice in the first place.

Cabal/Distribution/Simple/Program/HcPkg.hs
23Skidoo: Types for the type god!

cabal-install/Distribution/Client/Config.hs
```diff
@@ -430,7 +431,8 @@ initialSavedConfig = do
       globalWorldFile    = toFlag worldFile
     },
     savedConfigureFlags  = mempty {
-      configProgramPathExtra = toNubList extraPath
+      configProgramPathExtra = toNubList extraPath,
+      configView             = toFlag "default"
```
ttuegel: What will happen if the new cabal-install tries to pass this flag to an old Cabal `setup` executable?

23Skidoo:
@dcoutts @ezyang So I assume that this is obsoleted by `nix-local-build` and can be closed?

ezyang:
Yep.

PR #2706, opened 2015-07-05
Fix #2704

phadej:
Code branch with zlib <0.6 cannot recognise the uncompressed input.

---

OTOH, now as I think, maybe the property is ill-formed. QuickCheck *might* generate valid gzip or zlib stream for `[Word8]`. Opinions?

PR #2708, opened 2015-07-06
maybeDecompress: bail on all errors at the beginning of the stream with zlib < 0.6

phadej:
Related to @23Skidoo question in https://github.com/haskell/cabal/pull/2706


23Skidoo:
Merged, thanks!

PR #2734, opened 2015-07-25
Properly use test name for test libraries, with updated test.

ezyang:


23Skidoo:
LGTM. Thanks for writing a test case!

Issue #2013, opened 2014-07-24
Cabal shows a misleading message


wereHamster:
**Steps to reproduce:**

 - Start with a clean sandbox
 - Run `cabal install --dry-run --constraint 'aeson>=0.8.0.0' 'snap==0.13.2.8'`

**Actual message:** Contains the line `trying: lens-4.3 (dependency of snap-0.13.2.8)`. But lens-4.3 is not a dependency of snap-0.13.2.8. In fact, that snap version has an upper bound on lens <4.2.

**Expected result:** Something along the lines that snap-0.13.2.8 has a upper bound on aeson <0.8.

kosmikus:
I'm working on this now.

grayjay:
It looks like this issue was fixed by #2741.  Now cabal prints this message with `-v3`:

    [__9] next goal: lens (dependency of snap-0.13.2.8)
    [__9] rejecting: lens-4.13, lens-4.12.3, lens-4.12.2, lens-4.12.1, lens-4.12, lens-4.11.1, lens-4.11, lens-4.10, lens
    -4.9.1, lens-4.9, lens-4.8, lens-4.7.0.1, lens-4.7, lens-4.6.0.1, lens-4.6, lens-4.5, lens-4.4.0.2, lens-4.4.0.1, len
    s-4.4, lens-4.3.3, lens-4.3.2, lens-4.3.1, lens-4.3, lens-4.2 (conflict: snap => lens>=3.7.6 && <3.8 || >=3.7.6 && <4
    .2)
    [__9] trying: lens-4.1.2.1
    [_10] trying: lens-4.1.2.1:!bench

kosmikus:
@grayjay Thanks.

PR #2741, opened 2015-07-28
When lifting dependencies, compute the union of the ranges.

kosmikus:
For a while now, we lift out dependencies from conditionals.  If a
package is dependended on regardless of the outcome of a test, the
dependency is lifted out. This has the advantage that we can choose an
instance for the package prior to choosing the value of the flag. Often,
this is the desired behaviour for flags, i.e., to have the flag choice
implied by the possible package choices, rather than the other way
around.

Prior to this change, lifting was simply adding an unconstrained
dependency to such a package at a higher level, which would then of
course still be refined after making the flag choice. We are now making
this more precise by actually computing the union of the ranges in both
branches of the conditional. So we will not choose an instance on the
top level that will later be rejected by both branches.


kosmikus:
This should fix #2013.

23Skidoo:
Looks good.

ezyang:
Merged.

PR #2746, opened 2015-07-30
Support haddock response files

randen:
This change is for adding the ehancement to support haddock
response files.  This is what was used to build HP 7.10.2
for Windows, but it does need a corresponding change in
haddock to utilize it. Hopefully, this and its corresponding
haddock change can make it into the next ghc release and
eliminate much gnashing of teeth, error prone, and time
consuming manual steps in the HP build process.

Modified:

  * renderArgs function: I put a couple of things into locals
    since I needed another use for UTF8 support check, plus I
    added another check based on version; the temp file logic
    was just as the prologue case above but I did need to
    repeat the invocation of the 'k' function in order to
    keep the cases separate and allow proper handling of the
    temp file automatic (or not, per --keep-temp-files)
    deletion.  Important: the haddock version being check
    against for response file support, greater than 2.16.1,
    is a placeholder and may or may not be the actual value
    since that will depend on the as-yet-unreleased haddock
    (which looks like it may be >2.16.1 but at this moment is
    not released).

### DIFF
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -466,15 +466,28 @@ renderArgs :: Verbosity
               -> (([String], FilePath) -> IO a)
               -> IO a
 renderArgs verbosity tmpFileOpts version comp args k = do
+  let haddockSupportsUTF8          = version >= Version [2,14,4] []
+      haddockSupportsResponseFiles = version >  Version [2,16,1] []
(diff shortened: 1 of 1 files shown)


23Skidoo:
Does this need to be merged into the 1.22 branch?

23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 1.22.

PR #2747, opened 2015-07-31
Add frameworks when linking a dynamic library

christiaanb:
This fixes the Cabal side of [GHC trac #10568](https://ghc.haskell.org/trac/ghc/ticket/10568)


christiaanb:
I don't understand how my 1-line patch can have anything to do with the reported build failure. Perhaps somebody can explain what that unit-test is supposed to test? and how my patch, which simply adds some linker flags, would cause it to fail?

23Skidoo:
Restarted the Travis build. We've been seeing some spurious Travis failures lately.

23Skidoo:
And now it passes.

23Skidoo:
LGTM. Do you want this to also go into the 1.22 branch?

christiaanb:
I would want it to end up in the branch that will be shipped with a future GHC 7.10.3, if that's the 1.22 branch, then yes.

23Skidoo:
> I would want it to end up in the branch that will be shipped with a future GHC 7.10.3

Then you need it in 1.22. Cherry-picked.

PR #2748, opened 2015-08-01
Use ghc-pkg multiple instances option when available

fugyk:
This patch enables allowing of installations of multiple instances of same version of packages by cabal. Among other things it fixes breaking re-installations problem in cabal.

It is an attempt to divide gsoc project patches in 3-4 PRs rather than a single monolith patch:
1. Allowing multiple instances
2. Views and sharing of packages
3. GC + cabal remove and upgrade-all
4. Unique IPID calculation from source

You can expect other three PRs to come within a week.


Cabal/Distribution/Simple/Register.hs
```diff
@@ -230,13 +232,23 @@ invokeHcPkg verbosity comp conf dbStack extraArgs =
...
                   \not implemented for this compiler")
 
+getHcPkgInfo :: Compiler -> ProgramConfiguration -> Maybe HcPkgInfo
+getHcPkgInfo comp conf = case compilerFlavor comp of
+                           GHC   -> Just (GHC.hcPkgInfo conf)
+                           GHCJS -> Just (GHCJS.hcPkgInfo conf)
+                           LHC   -> Just (LHC.hcPkgInfo conf)
+                           _     -> Nothing
+
+multInstEnabled :: Compiler -> ProgramConfiguration -> Bool
```
23Skidoo: `D.S.Compiler` is probably a better place for this, since it already includes similar functionality (`parmakeSupported`).

23Skidoo:
LGTM modulo comments. 

/cc @kosmikus

Cabal/Distribution/Simple/Register.hs
fugyk: This module already has `withHcPkg ` which does the same thing, but it returns in unrequired IO monad. So I thought that this module is good to go for getHcPkgInfo. This function is also not exported, but if you want to change module for this function, I do not have a problem.

23Skidoo: The difference is that `withHcPkg` is an unexported helper function.

fugyk: This function is also unexported. I have mentioned that in last comment.

23Skidoo: `multInstEnabled` is exported.

fugyk: Sorry, I overlooked the line you had commented and thought that you were referring to `getHcPkgInfo`. But I looked into `D.S.Compiler` and it seems to be focused only on compiler features with `ghc --info` and `multInstEnabled` is a ghc-pkg feature. Also `D.S.Compiler` is imported by `D.S.Programs.HcPkg` now and so `HcPkgInfo` cannot be available to `D.S.Compiler` unless we move things from one module to another.

23Skidoo: Then maybe you can move it to `D.S.Programs.HcPkg`?

fugyk: I have tried that now, but found that `D.S.GHC`(and variants) imports `D.S.Programs.HcPkg` and so `getHcPkgInfo` cannot be moved into `D.S.Programs.HcPkg` also.

ezyang:
OK.

ezyang:
I did some feature testing on this with `cabal-install` in GHC 7.10, and it does seem to work well. My test went like this:

1. I installed `yesod-1.4.2`
2. I installed `yesod-1.4.1` (this one actually failed because it has a bogus version bounds on network WAI handler :)

Reinstalls of libraries proceeded fine. For example, `xml-conduit-1.3.1` got installed twice, with different versions of `conduit`, and there were no problems.

fugyk:
This PR has been closed due to github's bug. I was trying to update my email address in git for submission of code samples in GSoC website, but I have overwritten all of the history due to carelessness. But after reverting back, this PR is not updating and is in permanently closed status. I have contacted github support and got the reply that it can't be opened back and I have to create a new PR. I am not opening a new PR as both of the closed PR is superseded by @ezyang's PR.

PR #2750, opened 2015-08-03
Fix windows build

anton-dessiatov:
This one is trivial. Commit 8ea6f33c34e10dcc9faa046040266e91676e8062 has broken Windows build, this looks to me like a fix. The problem is that ```InstallSymlink.hs``` file is conditionally compiled and Windows part wasn't updated.


23Skidoo:
Merged, thanks!

PR #2516, opened 2015-04-01
Eliminate duplicate hsc2hs command line arguments

anton-dessiatov:
I've recently stumbled upon an issue that didn't let me to build my Haskell project on Windows. It manifested itself as an error:

`cabal: w:\haskell\bin\hsc2hs.exe: does not exist`

The file that seemed to be missing (hsc2hs.exe binary) was actually present in necessary directory, so I started investigating the issue.

It turned out that the core reason was that hsc2hs command line exceeded the maximum allowed length of 32767 characters. The reason why it exceeded the limit was the huge amount of duplicate `--cflag` arguments that appeared from the dependent packages. As a quick fix I just added a `nub` function call to eliminate duplicate arguments and the problem is fixed at least for my project.

While this fix doesn't completely eliminate the problem (at some point there eventually will be user with huge number of distinct arguments to hsc2hs), it still does offer some help to me, so I decided to make a pull request.

PR #2751, opened 2015-08-03
Fix issues with OSX builds after nubbing hsc2hs arguments

anton-dessiatov:
This is a fix I came up for @christiaanb issues (see #2516 for additional details). I've just splitted hsc2hs arguments in two parts. One part is nub'ed, another isn't. ```-framework``` flags are in the latter part.

@christiaanb could you please check if this PR fixes build failures on OS X?

### DIFF
--- a/Cabal/Distribution/Simple/PreProcess.hs
+++ b/Cabal/Distribution/Simple/PreProcess.hs
@@ -395,59 +395,59 @@ ppHsc2hs bi lbi =
       (gccProg, _) <- requireProgram verbosity gccProgram (withPrograms lbi)
       -- ordNubRight function helps a bit in case dependent packages bring too many
       -- command line arguments (see pull request #2516).
-      rawSystemProgramConf verbosity hsc2hsProgram (withPrograms lbi) . ordNubRight $
-          [ "--cc=" ++ programPath gccProg
-          , "--ld=" ++ programPath gccProg ]
+      let allowedToNub = 
+            [ "--cc=" ++ programPath gccProg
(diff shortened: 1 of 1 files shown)


23Skidoo:
@anton-dessiatov Are you sure that this bug exists only in OSX code and won't bite us in other contexts?

anton-dessiatov:
Not really. The issue is that if there are cflags that work in groups then these groups have to be treated  as a whole (like ```-framework CoreVideo```, which are essentially two flags, but they cannot be splitted). If some other code decides to bring agrument groups like that, the thing will break again.

23Skidoo:
Then we probably should revert that change altogether.

anton-dessiatov:
Probably yes, because it was just a workaround for limited length of Windows command line argument string.

The ultimate solution would be to patch ```hsc2hs``` to accept command line arguments from a file (like Visual Studio tools do with ```.rsp``` files) and pass them in a temporary file instead of a command line. Maybe I'll come with that later.

23Skidoo:
We already do it for Haddock.

23Skidoo:
Reverted #2516 in fe7b8784ac0a5848974066bdab76ce376ba67277, so closing.

23Skidoo:
@anton-dessiatov BTW, if you decide to write a patch for `hsc2hs`, one thing to think about is how to handle arguments with spaces in them. With Haddock we went for a simple solution: each argument on its own line. However, the standard solution (used by `gcc` for example) is to use quotation marks (both single and double quotation marks are supported).

Issue #1919, opened 2014-06-04
cabal haddock --executables fails with "internal error when calculating transitive package dependencies"


massysett:
When running

```
cabal haddock --executables
```

the command fails with

```
cabal: internal error when calculating transitive package dependencies.
Debug info: []
```

I have tagged my source code tree where this problem exists:

https://github.com/massysett/penny/tree/cabal_error

This is cabal-install version 1.20.0.2.

taladar:
Has any progress been made on this issue? We have some projects with significant code in the modules which only appear in the executable and it would be great if documentation could be generated for those as well. Of course removing the dependency on the local library is not really an option.

edsko:
This problem still occurs with 1.24.0.0.

OlivierSohn:
I'm running into this error, and like @taladar says, removing the dependency on the local library (or convenience library) fixes haddock generation, but I need the dependency nonetheless for projects that depend on the main library (else I have link-time errors).

Has anyone found a workaround for this issue?

Thanks!

c-lewis:
This problem is also present in 2.2.0.1.

alexbiehl:
I think this is fixed with latest Cabal.

Christopher Lewis <notifications@github.com> schrieb am Sa., 27. Okt. 2018,
06:54:

int-e:
FWIW, this bug still exists for the `v1-haddock` command in `cabal-install 2.4.1.0`. It occurs when an executable depends on the library defined in the current package.

mietek:
My `cabal-fail-one` repository is now named [`cabal-issue-1919`](https://github.com/mietek/cabal-issue-1919).

int-e:
Note that `new-haddock` is affected by a *different* issue that also causes executable documentation to not be built in some cases (including @mietek's repository, I think): #5890.

madgen:
I get the same error when I use `cabal haddock` on a package with internal libraries with Cabal 2.4.1.0. Since Stack currently supports up to Cabal 2.4, I imagine this affects some number of people.

Is the only workaround to just not use internal libraries? Does anyone have an idea where to start poking to work towards a patch?

brandon-leapyear:
Seems like this is still an issue? I'm surprised this issue is still open after 7 years. Is there any progress yet? Is this work blocked by another issue, or can it be worked on today?

Mikolaj:
I don't see any label suggesting this is blocked (e.g., on a haddock bug), so I guess any work on this would be very welcome.

[...]

PR #2753, opened 2015-08-07
[RFC] Handle executables depending on package library for Haddock

rimmington:
... sort of. Altering `haddock` to use the same fix as `build` did back in 1.8.0.2 / 2009 / 7a69bf67, ie. to use an inplace package DB, allows the generation of Haddocks with intra-package dependencies (issue #1919). Unfortunately this means you cannot generate Haddocks without building first, because `ghc-pkg` gets understandably upset if you try to register a package with nothing built.

Considering this downside and my unfamiliarity with the Cabal codebase, this is more of an area for discussion and **not ready for merge**.

### DIFF
--- a/Cabal/Distribution/Simple/Build.hs
+++ b/Cabal/Distribution/Simple/Build.hs
@@ -20,6 +20,8 @@ module Distribution.Simple.Build (
     startInterpreter,
 
     initialBuildSteps,
+    inplaceRegisterLibrary,
+    createInternalPackageDB,
     writeAutogenFiles,
   ) where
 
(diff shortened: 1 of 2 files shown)


23Skidoo:
Do you know what causes the Travis failure?

rimmington:
After a brief look at the testing code, I suspect it's the aforementioned issue that `cabal haddock` no longer works before you run `cabal build` because it always tries to register the library. I think this could be fixed by registering the library only when required, but I haven't figured out a nice way to do that.

rimmington:
I'm not working on this.

PR #2765, opened 2015-08-15
Update documentation for test-only packages

grayjay:
After #2699, the documentation no longer needs to explain that the example .cabal files with only tests or benchmarks are invalid.


ttuegel:
Thanks!

PR #2773, opened 2015-08-17
Warn if 'license-file' is absolute or outside of source tree (fixes #2742)

### LINKED ISSUE #2742: license-file: ../LICENSE is accepted, but the file does not show up in the sdist
The `.cabal` file line `license-file: ../LICENSE` is accepted, but then the file does not show up in the sdist.  If there's no way of meaningfully including a file from higher up the directory tree then the configuration line should be rejected.

grayjay:


23Skidoo:
Merged, thanks!

Issue #2772, opened 2015-08-17
Remove global DB restriction when converting PackageDbStack into ghc/ghc-pkg flags


23Skidoo:
In #2593, we removed the global DB restriction from `getInstalledPackages` for compilers that don't require this restriction. We should also change the function that converts the db stack into ghc/ghc-pkg flags (`D.S.Program.GHC.packageDBArgs`) accordingly. See this comment by @dcoutts: https://github.com/haskell/cabal/pull/2617#issuecomment-107149392.

phadej:
I can check this during this week (no hard promise though).

snoyberg:
Hope you don't mind @phadej, I took a crack at this in #2776.

PR #2776, opened 2015-08-18
Generate package DB flags for skipped global #2772

snoyberg:
I tried to implement this in as small a way as possible, though overall the PackageDBStack approach feels slightly wrong: if we're hardcoding an order of optional global, optional user, other DBs, a data type like the following would be better:

```haskell
data PackageDBStack = PackageDBStack
    { includeGlobal :: Bool
    , includeUser :: Bool
    , extraDBs :: [FilePath]
    }
```

However, ship's probably sailed on that one. Assuming the configure step guarantees that databases come in in the correct order, this patch should be fine.

### DIFF
--- a/Cabal/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/Distribution/Simple/Program/GHC.hs
@@ -457,6 +457,11 @@ packageDbArgs implInfo dbstack = case dbstack of
   (GlobalPackageDB:UserPackageDB:dbs) -> concatMap specific dbs
   (GlobalPackageDB:dbs)               -> ("-no-user-" ++ packageDbFlag)
                                        : concatMap specific dbs
+  (UserPackageDB:dbs)                 -> ("-no-global-" ++ packageDbFlag)
+                                       : concatMap specific dbs
+  dbs                                 -> ("-no-global-" ++ packageDbFlag)
+                                       : ("-no-user-" ++ packageDbFlag)
+                                       : concatMap specific dbs
   _ -> ierror
   where
     specific (SpecificPackageDB db) = [ '-':packageDbFlag , db ]


phadej:
Is global and user databases somehow special (otherwise than GHC knows their location), or could one have? `ghc -clear-package-db -package-db my-own-global-db -user-package-db` ?

At least `% ghc --make -clear-package-db -global-package-db -package-db ddd -user-package-db Main.hs` seems to work (with empty `ddd`), that said: I have no idea why anyone would do that.

snoyberg:
I think it may be handled specially by `cabal configure`, but GHC does not seem to have any special treatment of the order of the stack. And indeed, there are good reasons to want to rearrange the stacks in interesting ways, but that can always be achieved by passing them all in directly to cabal via `--package-db=(path-to-global)` etc.

Cabal/Distribution/Simple/Program/GHC.hs
```diff
@@ -457,6 +457,11 @@ packageDbArgs implInfo dbstack = case dbstack of
   (GlobalPackageDB:UserPackageDB:dbs) -> concatMap specific dbs
   (GlobalPackageDB:dbs)               -> ("-no-user-" ++ packageDbFlag)
                                        : concatMap specific dbs
+  (UserPackageDB:dbs)                 -> ("-no-global-" ++ packageDbFlag)
+                                       : concatMap specific dbs
+  dbs                                 -> ("-no-global-" ++ packageDbFlag)
+                                       : ("-no-user-" ++ packageDbFlag)
+                                       : concatMap specific dbs
   _ -> ierror
```
phadej: the `_` case is not needed.

phadej:
I just don't like the partiality of `specific`, we could do for recent enough GHC:
```hs
packageDbArgs :: GhcImplInfo -> PackageDBStack -> [String]
packageDbArgs implInfo dbstack = case dbstack of
  -- special case to have nicer arguments in common situations
   (GlobalPackageDB:UserPackageDB:dbs) -> concatMap specific dbs
   (GlobalPackageDB:dbs)               -> ("-no-user-" ++ packageDbFlag)
                                        : concatMap specific dbs
   dbs                                 -> ("-clear-package-db)
                                        : concatMap specific dbs
   where
     specific (SpecificPackageDB db) = [ '-':packageDbFlag , db ]
     specific GlobalPackageDB        = [ "-global-package-db" ]
     specific UserPackageDB          = [ "-user-package-db" ]
```

snoyberg:
I agree with your analysis and think it's a better approach. Do you want to send that add a separate PR, or should I update my branch with your suggestions?

I'll be afk for the next few hours, so if the latter we'll need to wait a bit.

phadej:
I can do another PR

23Skidoo:
Closing in favour of #2778 (which is already merged).

ezyang:
Actually close.

PR #2778, opened 2015-08-18
Resolve #2772: generate database flags (may omit global db)

### LINKED ISSUE #2772: Remove global DB restriction when converting PackageDbStack into ghc/ghc-pkg flags

phadej:
Alternative solution to https://github.com/haskell/cabal/pull/2776

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -332,11 +332,10 @@ checkPackageDbEnvVar =
     Internal.checkPackageDbEnvVar "GHC" "GHC_PACKAGE_PATH"
 
 checkPackageDbStack :: Compiler -> PackageDBStack -> IO ()
-checkPackageDbStack comp =
-  let ghcVersion = compilerVersion comp
-  in if ghcVersion < Version [7,6] []
-        then checkPackageDbStackPre76
-        else checkPackageDbStackPost76
+checkPackageDbStack comp = if flagPackageConf implInfo
+                              then checkPackageDbStackPre76
+                              else checkPackageDbStackPost76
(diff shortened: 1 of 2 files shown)


phadej:
ping @dcoutts, @23Skidoo

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -332,11 +332,10 @@ checkPackageDbEnvVar =
...
 checkPackageDbStack :: Compiler -> PackageDBStack -> IO ()
-checkPackageDbStack comp =
-  let ghcVersion = compilerVersion comp
-  in if ghcVersion < Version [7,6] []
-        then checkPackageDbStackPre76
-        else checkPackageDbStackPost76
+checkPackageDbStack comp = if flagPackageConf implInfo
+                              then checkPackageDbStackPre76
+                              else checkPackageDbStackPost76
+  where implInfo = ghcVersionImplInfo (compilerVersion comp)
```
23Skidoo: We probably should do something to avoid recomputing `GhcImplInfo` each time it's needed.

23Skidoo: Not in this patch, of course.

23Skidoo:
OK, I think I'll just merge it and a comment myself.

23Skidoo:
Merged, thanks!

phadej:
Np, thanks for commening!

Issue #1622, opened 2013-12-18
--enable-library-stripping


accursoft:
Installed libraries can be shrunk by about 10% with `strip --strip-unneeded`. It would be nice if cabal could do this automatically, the same way it strips executables.

PR #2782, opened 2015-08-19
Don't strip libraries in build directory

thomie:
Two changes:

  * strip static libraries (.a files) at install/copy time, instead of
    at build time. Note that Cabal strips shared libraries (.so files)
    and executables at install/copy time also.

    This is needed for GHC, because it uses Cabal (via ghc-cabal copy)
    to install the boot libraries, and relies on Cabal to take of
    library stripping. Currently .so files indeed get properly stripped,
    but .a files do not.

    (Stripping static libraries at build time was introduced in
    60409cb9493de99f6549956e09604596e3f28938, with the explanation:
    "Call 'stripLib' from createLibArchive so that it's done only
    once.")

    This bug (if you want to call it that) partially explains the
    difference in size of a GHC installation before and after manual
    stripping, reported in (edit: https://github.com/haskell/cabal/issues/1622#issuecomment-62076817) .
    (The other parts are that the
    GHC build system currently doesn't strip executables on
    installation: https://ghc.haskell.org/trac/ghc/ticket/9087, and
    neither are the RTS library stripped, since they don't go through
    Cabal).

  * never strip the copy of a library in the build directory. Only
    strip the copy that is installed. Note that Cabal never strips
    executables in the build directory either.

    This might speed up compilation of a package under development,
    since stripping won't be performed for every 'cabal build'. It is
    also consistent with the GNU coding standards for 'install-strip':

      "install-strip should not strip the executables in the build
      directory which are being copied for installation. It should only
      strip the copies that are installed."

    Source:
    http://www.gnu.org/prep/standards/html_node/Standard-Targets.html

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -1075,11 +1075,13 @@ installLib verbosity lbi targetDir dynlibTargetDir builtDir _pkg lib clbi = do
       let src = srcDir </> name
           dst = dstDir </> name
       createDirectoryIfMissingVerbose verbosity True dstDir
+
       if isShared
-        then do when (stripLibs lbi) $ Strip.stripLib verbosity
-                                       (hostPlatform lbi) (withPrograms lbi) src
-                installExecutableFile verbosity src dst
+        then installExecutableFile verbosity src dst
         else installOrdinaryFile   verbosity src dst
+
+      when (stripLibs lbi) $ Strip.stripLib verbosity
+                             (hostPlatform lbi) (withPrograms lbi) dst
 
     installOrdinary = install False
     installShared   = install True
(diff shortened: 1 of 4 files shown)


23Skidoo:
Fails on Travis due to `-Werror`.

thomie:
Hm, this patch isn't quite ready yet.
```
DeterministicAr:                     FAIL (0.73s)
    PackageTests.DeterministicAr.checkMetadata: Metadata has changed at offset 8 in /home/travis/.cabal/lib/x86_64-linux-ghc-7.4.2/DeterministicAr-0/libHSDeterministicAr-0.a
```

thomie:
Ready for review.

23Skidoo:
Merged, thanks!

PR #2785, opened 2015-08-20
Remove dead code

BardurArantsson:
Only the NoComments case of IncludeComments was ever used and we can
simplify the generation of the default "package environment file"
accordingly.

--
This is simplifies some of the work I'm doing around searching for sandboxes (so it's a prerequisite for that), but should be fine for merge as-is since it just eliminates dead code.


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #2786, opened 2015-08-21
use a liberal import of Control.Exception to grab the new ErrorCall pattern in base

bgamari:
This is a prerequisite for GHC Phabricator [D861](https://phabricator.haskell.org/D861).

### DIFF
--- a/Cabal/Distribution/Compat/Binary.hs
+++ b/Cabal/Distribution/Compat/Binary.hs
@@ -1,4 +1,7 @@
 {-# LANGUAGE CPP #-}
+#if __GLASGOW_HASKELL__ >= 711
+{-# LANGUAGE PatternSynonyms #-}
+#endif
 
 #ifndef MIN_VERSION_binary
 #define MIN_VERSION_binary(x, y, z) 0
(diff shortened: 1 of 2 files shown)


23Skidoo:
Can't we do something like

```
#if __GLASGOW_HASKELL__ >= 711
import Control.Exception (pattern ErrorCall)
#endif
```

instead? Explicit imports give us forwards compat and help navigating code.

bgamari:
@gridaphobe, what say you?

gridaphobe:
Sure, that ought to work too (we'd also need to enable `PatternSynonyms` of course). I try to avoid using CPP personally, which is why I used the full import, but I don't have a strong opinion there.

(I'm not at my computer ATM so I haven't tried to compile anything)

bgamari:
@gridaphobe, do you think you could give this a shot?

gridaphobe:
Sure, I'll do it later today.

gridaphobe:
@23Skidoo is this better?

@bgamari I validated the new patch against the rest of [D861](https://phabricator.haskell.org/D861), so we should be good to go once it's merged.

23Skidoo:
@gridaphobe Looks OK. You sure that the `ErrorCall (..)` import doesn't trigger a warning on GHC 7.11? We strive to be warning-free.

gridaphobe:
Ah, it did indeed, fixed!

23Skidoo:
Great, thanks. Merging.

PR #2787, opened 2015-08-21
Warn when constraints or preferences refer to unknown packages (fixes #2719)

### LINKED ISSUE #2719: Missing warning when constraint is not used
I lost some time using the command

```bash
$ cabal install foo --constraint=trasformers==0.4.3.0
```
after I realised that there was a typo in `trasformers`. I warning like

```bash
the constraint trasformers==0.4.3.0 was not used
```
would have been very useful.

grayjay:
`cabal install` and `cabal configure` warn when any constraint or preference names a package that is not in the source package index or installed package index.

I tried to keep it simple and unobtrusive by only printing the first bad constraint or preference.  Would it be better to summarize all of them?

### DIFF
--- a/cabal-install/Distribution/Client/Configure.hs
+++ b/cabal-install/Distribution/Client/Configure.hs
@@ -15,23 +15,25 @@ module Distribution.Client.Configure (
     configure,
     configureSetupScript,
     chooseCabalVersion,
+    checkConfigExFlags
   ) where
 
 import Distribution.Client.Dependency
 import Distribution.Client.Dependency.Types
          ( AllowNewer(..), isAllowNewer, ConstraintSource(..)
-         , LabeledPackageConstraint(..) )
+         , LabeledPackageConstraint(..), showConstraintSource )
 import qualified Distribution.Client.InstallPlan as InstallPlan
 import Distribution.Client.InstallPlan (InstallPlan)
 import Distribution.Client.IndexUtils as IndexUtils
          ( getSourcePackages, getInstalledPackages )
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Configure.hs
```diff
@@ -223,6 +230,31 @@ configureSetupScript packageDBs
...
+          ++ showConstraint (head unknownConstraints)
+  unless (null unknownPreferences) $ warn verbosity $
+             "Preference refers to an unknown package: "
+          ++ display (head unknownPreferences)
+  where
+    unknownConstraints = filter (unknown . userConstraintPackageName . fst) $
+                         configExConstraints flags
+    unknownPreferences = filter (unknown . \(Dependency name _) -> name) $
+                         configPreferences flags
+    unknown pkg = null (lookupPackageName installedPkgIndex pkg)
```
23Skidoo: Why do we need to look up the name twice?

23Skidoo:
LGTM. @kosmikus, what do you think?

cabal-install/Distribution/Client/Configure.hs
grayjay: I looked up the package in the installed index as well as the source index because an installed package may no longer be available as source.  I also found that `cabal freeze` can create a constraint on rts, even though rts isn't in the source index.

23Skidoo: OK.

>  also found that `cabal freeze` can create a constraint on rts

It probably shouldn't do that, though.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #2790, opened 2015-08-24
Add architecture to hash sandbox dir

microtorus:
This patch allows for 32-bit and 64-bit concurrent builds.

This ensures that dependencies built for different architectures will not get mixed into the same dist-sandbox directory thus allowing for x86 and x86_64 builds in the same working space and even concurrently.

### DIFF
--- a/cabal-install/Distribution/Client/Sandbox.hs
+++ b/cabal-install/Distribution/Client/Sandbox.hs
@@ -93,7 +93,7 @@ import Distribution.Simple.Utils              ( die, debug, notice, info, warn
                                               , intercalate, topHandlerWith
                                               , createDirectoryIfMissingVerbose )
 import Distribution.Package                   ( Package(..) )
-import Distribution.System                    ( Platform )
+import Distribution.System                    ( Platform(..) )
 import Distribution.Text                      ( display )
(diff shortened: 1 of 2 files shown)


23Skidoo:
Instead of just `$arch`, please use `$arch-$os-$compiler`, like we do elsewhere. Instead of just hashing it, please change the directory to `dist/dist-$HASH/$arch-$os-$compiler` (aside: we should probably be using a platform-specific directory separator here instead of a slash).

BardurArantsson:
@ezyang Is this relevant after your changes to "pkgid", etc?

If yes, then why aren't we merging this?

ezyang:
This looks orthogonal, because it is related to sandbox directory selection which is not based off of a computed component ID. Which mean it should be relevant.

23Skidoo:
> why aren't we merging this?

I guess because GitHub doesn't send notifications when the submitter updates the PR by pushing additional commits. I'll postpone this until after 1.24 is branched because I think there's some potential for breakage.

BardurArantsson:
Ugh. Another "process" thing forced upon us by GitHub :(

(Good work, btw, getting a release together!)

phadej:
Sandbox is deprecated.

PR #2791, opened 2015-08-24
Cabal IPID flags

ezyang:
This is https://github.com/haskell/cabal/pull/2752 but with an extra commit for adding support for an `--ipid` flag for explicitly programming the IPID on configuration.


ezyang:
Subsumed by PR #2817

PR #2792, opened 2015-08-24
Eliminate library names and package keys, unifying them with IPIDs

ezyang:
PR'd to get Travis run.


fugyk:
Correct me if I am wrong. The use of package key is to completely rebuild the package if any of its direct or indirect dependency is changed by passing `--this-package-key` to ghc in the build step. After that there is no use of package key except to calculate package key of the package that will depend on it.

If that's the case than you are using inplace package key in the build step, meaning even if any dependency is changed, there will be no rebuilds. This seems to nullify the entire use of package key.

ezyang:
No, what we called the package key had one other important effect, which was that it was what was used for internal type equalities and symbol names we write out to object and interface files. It is NOT, in any sense, intended to be used to induce recompilation (recompilation is just an unfortunate side effect of what happens when you change the package key.)

Also, you WILL recompile if the dependency changes, because GHC always checks the IPIDs of depended upon packages to figure out if it needs to recompile.

ezyang:
OK, I think the latest revision should be green on all versions of GHC and be ready for merge.

ezyang:
Subsumed by PR #2817

PR #2799, opened 2015-08-27
Use the sandbox config for "cabal get"

simonmar:
If I try to use a local cabal-install repo cache by adding a
"remote-repo-cache: <dir>" setting to the local cabal.config, this is
respected by "cabal update" but then ignored by "cabal get", which seems
wrong.  This respects the local cabal.config in "cabal get" too.  Maybe
all commands should do this?


23Skidoo:
Merged, thanks!

23Skidoo:
> Maybe all commands should do this?

Yes, I think we should just use `loadConfigOrSandboxConfig` everywhere. This is not the first time an issue like this one comes up.

Issue #2810, opened 2015-09-04
Search upwards for package directory


BardurArantsson:
The idea here is that if the user's current working directory is somewhere inside a cabal package, then cabal-install should automatically find the .cabal file in one of the parent directories, instead of forcing the user to change to that directory.

BardurArantsson:
Implementation issues/notes:

1. Currently Cabal *probably* doesn't change the current directory to the .cabal file's directory (why would it?)
2. The implementation (see pull request) is quite messy currently (quite a few FIXMEs) that need to be addressed. Most of them are simple (and I'll address them over the new few days), but a few are rather more complicated.

hvr:
Fwiw, things like `cabal test` may benefit from temporarily changing the CWD as test-frameworks which read external files assume to operate from the folder where the `.cabal` file is located

BardurArantsson:
Yup, and the problem is... properly handling CWD has some potentially thorny issues:

* We'd need to find all the places that require a change of CWD. It might be possible to find a single spot in cabal-install/Main.hs where this could be done -- I'm not sure if we have all the necessary location/configuration metadata in some common spot.
* CWD handling on Windows is a royal pain because it's a *per-drive* and *per-process* thing. However, I guess we could pretty much assume we're always in a CWD on the same drive since we're doing a recursive upwards search, not crossing drives.
* We'd have to reliably switch back the CWD to whatever it was before. The only *really* reliable way to do that is by spawning a new subprocess and then have that subprocess change the CWD and do all the work. *This is the biggie* -- it complicates things rather a lot and I'm not even sure we can do this sanely on Windows since it doesn't(?) support fork(). Either way this means that we'd have to make cabal-install a two-stage thing.

A possible workaround would be to export an environment variable (like `CABAL_SANDBOX_PACKAGE_PATH` e.g. `CABAL_PACKAGE_PATH`) so that any code currently relying on CWD could at least be adjusted to look at that?

BardurArantsson:
If we're going to do a forking type thing, an interesting idea might be to follow git's model of handling subcommands, namely just compiling each subcommand into a separate executable and then calling that from the main command. That would automatically "protect" the user's CWD from any interference by subcommands. What do you think about that idea?

(I realize that such a change definitely isn't trivial, but is it totally crazy?)

hvr:
As I mentioned in #2349, why do we need to restore the CWD at all during `cabal` process' lifetime?

BardurArantsson:
[...]

Issue #2797, opened 2015-08-27
Migrate integration tests to some Makefile/shell based format


ezyang:
Currently, the process for package tests in Cabal is something like:

1. Create a new Haskell file and write a lot of boilerplate to call a bunch of Cabal/ghc-pkg commands
2. Wire it up to the main test caller
3. Call 'cabal test' and wait for the test suite to build
4. Actually run the test suite

This really sucks, for a few reasons:

1. It takes a lot of boilerplate to write the actual package test, way more than an equivalent shell script plus expected output to diff against would
2. Under the most normal workflow, we have to rebuild the test suite every time we make a change (e.g. tweak a command). This makes for a pretty annoying compile-run cycles
3. It's opaque: it's much harder to reverse engineer what the source level commands are being run

The way GHC does these tests is that it just has a test runner which runs some shell scripts (really a Makefile) which runs a test, and the diffs it against some expected output. There's a bit of fuzz to make it all work out, but it is quite nice. Cabal should grow something similar; maybe that would encourage more tests! (Maybe there is a good out-of-the-box thing we can use here.)

BardurArantsson:
+1, if you (or someone else) could make this happen it would be great.

An additional advantage: We wouldn't have to worry about pollution of the process state, e.g. CWD or similar. Each test case would be a completely independent OS process, as it should be.

I guess a feasible way to go here would to put this new system in place and not bother converting the existing test cases? (I may be sufficiently motivated to do so because I may need to add some PackageTest cases soonish.)

Just for clarification: Why bit does the diffing, the each shell script or the test runner? (I'm assuming the latter, but just want to clarify.) There also might be some issues here if Cabal ever uses absolute paths, time stamps or similar things in output. (Perhaps the diffing should actually use some sort of pattern/regex-based matching to allow for such cases?)

BardurArantsson:
Given that one of my "contributions" had to be reverted a few days ago (basically due to an error caused by a lack of automated tests), I think I'll take a stab at an initial implementation this weekend. @ezyang : Just to make sure we're not duplicating effort: are you actively working on this, or...?

ezyang:
Nope, have at it!

BardurArantsson:
So, I think I have a basic test runner set up working now:

```
Using cabal: /home/bardur/wd/cabal/cabal-install/dist/build/cabal/cabal
Using ghc-pkg: /usr/bin/ghc-pkg
Current directory is: /home/bardur/wd/cabal/cabal-install
[...]

PR #2811, opened 2015-09-04
WIP/Search upwards for package directory

BardurArantsson:
FOR REVIEW ONLY

### DIFF
--- a/cabal-install/Distribution/Client/Exec.hs
+++ b/cabal-install/Distribution/Client/Exec.hs
@@ -19,9 +19,8 @@ import Data.Foldable (forM_)
 import qualified Distribution.Simple.GHC   as GHC
 import qualified Distribution.Simple.GHCJS as GHCJS
 
-import Distribution.Client.Sandbox (getSandboxConfigFilePath)
 import Distribution.Client.Sandbox.PackageEnvironment (sandboxPackageDBPath)
-import Distribution.Client.Sandbox.Types              (UseSandbox (..))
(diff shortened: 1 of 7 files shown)


BardurArantsson:
High-level overview of what I've done here:

  - Added a new SandboxMetadata that contains both the sandbox
  directory and the config file path.

  - I've changed the UseSandbox construct to take such a
  SandboxMetadata value -- this simplifies the lookup logic
  drastically since it can be unified in a single function which
  handles both the lookup of the .cabal file and the
  configuration file. There are some cases error cases that would
  be very hard to detect if we didn't combine the logic this.
  (See getSandboxMetadata for the search/lookup logic.)

Observation: The config flag handling seems rather messy -- I
find it very hard to judge which configuration is actually being
used.

BardurArantsson:
See also comments on the #2810 issue regarding implementation

BardurArantsson:
Oh, yeah, I haven't spent any time actually testing this yet -- I just want to get feedback on the approach before I spend any time on adding/updating tests. I expect a feature this complex will require some PackageTests at the very least.

hvr:
Thanks for doing this.

Could you maybe outline the algorithm you use (or plan to use) for locating the various config-files?

BardurArantsson:
I'm closing this for now. I think it'll need a bit of reworking and most of all *TESTS*. Hopefully, #2797 will be enough :).

BardurArantsson:
Meh, meant #2868. (Hard to follow all the indirections!)

PR #2814, opened 2015-09-08
Change "XXX" markers to "TODO"

BardurArantsson:
I don't like TODO markers, but at least this is more consistent.


23Skidoo:
Merged, thanks!

PR #2823, opened 2015-09-15
Call addInfoForKnownRepos in initialSavedConfig

edsko:
@dcoutts introduced `addInfoForKnownRepos` in 8617d39d3c0be275e4373c1c6d5ddc7368083960, but while we called `addInfoForKnownRepos` when _parsing_ a file, we didn't call it when creating the initial file. This meant that when we created an initial file, we would end up with a different internal configuration (which would not have `addInfoForKnownRepos` applied) than if we wrote out and then read back that same initial file (which would).


23Skidoo:
Merged, thanks!

PR #2826, opened 2015-09-17
Lower memory usage during index file parsing #2396

### LINKED ISSUE #2396: cabal update uses a ridiculous amount of memory
`cabal update` uses 500-600Mb on my system, causing builds to fail completely on small virtual servers.

Memory usage sky-rockets after `Updating the index cache file...`, so I'm guessing that something in `Distribution.Client.IndexUtils.updatePackageIndexCacheFile` is constructing the entire package list in memory.

sznurek:
Fixes #2396

As stated in one of the commit messages:

The parsePackageIndex wasn't streaming because resulting list was
hidden in two constructors: Either and tuple. By removing that, error
semantics changed in updatePackageIndexCacheFile.

The updatePackageIndexCacheFile problem was sequence call that evaluated
whole list produced by parsePackageIndex. By using lazySequence, any error
during index file parsing will be raised during writing to cache file.

Also removed readPackageIndexFile, as it wasn't used anywhere as far as
I can tell. If this is a mistake I'll include it again.

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -288,33 +288,6 @@ packageDesc :: PackageEntry -> GenericPackageDescription
 packageDesc (NormalPackage  _ descr _ _) = descr
 packageDesc (BuildTreeRef _ _ descr _ _) = descr
 
--- | Read a compressed \"00-index.tar.gz\" file into a 'PackageIndex'.
---
--- This is supposed to be an \"all in one\" way to easily get at the info in
--- the Hackage package index.
---
--- It takes a function to map a 'GenericPackageDescription' into any more
--- specific instance of 'Package' that you might want to use. In the simple
--- case you can just use @\_ p -> p@ here.
---
-
-{-
-readPackageIndexFile :: Package pkg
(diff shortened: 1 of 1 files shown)


ttuegel:
I can confirm this does improve memory usage significantly!

Before:
```
ttuegel ~$ rm -fr .cabal/packages/hackage.haskell.org                            
ttuegel ~$ command time cabal update                                             
5.59user 0.93system 0:08.73elapsed 74%CPU (0avgtext+0avgdata 514544maxresident)k
376inputs+428352outputs (3major+121981minor)pagefaults 0swaps
```

After:
```
ttuegel ~$ rm -fr .cabal/packages/hackage.haskell.org                            
ttuegel ~$ command time /tmp/cabal-dist-cabal-install/build/cabal/cabal update   
4.70user 0.56system 0:07.60elapsed 69%CPU (0avgtext+0avgdata 14984maxresident)k
16inputs+428352outputs (0major+1277minor)pagefaults 0swaps
```

Great work!

23Skidoo:
I feel a bit uneasy about solving problems caused by lazy I/O with more lazy I/O. Would be nice to move to a more principled scheme at some point.

PR #2828, opened 2015-09-19
Add hasUpperBound and hasLowerBound

snoyberg:
This is useful for performing PVP bounds checking, especially for
tooling that wants to automate such dependency bounds. A demonstration
of this being used in practice is the --pvp-bounds flag for stack:

https://github.com/commercialhaskell/stack/commit/c774f9905b3b112f58fc3789dfb9c26389d33b24

### DIFF
--- a/Cabal/Distribution/Version.hs
+++ b/Cabal/Distribution/Version.hs
@@ -44,6 +44,8 @@ module Distribution.Version (
   simplifyVersionRange,
   foldVersionRange,
   foldVersionRange',
+  hasUpperBound,
+  hasLowerBound,
 
   -- ** Modification
   removeUpperBound,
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Version.hs
```diff
@@ -805,3 +807,29 @@ instance Text VersionRange where
...
+--
+-- @since 1.24.0.0
+hasUpperBound :: VersionRange -> Bool
+hasUpperBound AnyVersion = False
+hasUpperBound (ThisVersion _) = True
+hasUpperBound (LaterVersion _) = False
+hasUpperBound (EarlierVersion _) = True
+hasUpperBound (WildcardVersion _) = True
+hasUpperBound (UnionVersionRanges x y) = hasUpperBound x && hasUpperBound y
+hasUpperBound (IntersectVersionRanges x y) = hasUpperBound x && hasUpperBound y
```
23Skidoo: I don't think that this is correct. Consider the intersection of `>=1.0` with `<=1.1`.

23Skidoo:
This may be easier to implement by converting `VersionRange` to `VersionIntervals` first. Also, it looks like other code in that file assumes that if the lower bound is not specified, then it is `>=0` by default, so in a sense a lower bound always exists.

Cabal/Distribution/Version.hs
snoyberg: You're right, good catch, thanks.

snoyberg:
Thanks for the pointer on VersionIntervals. It looks like it will work for upper bounds. However, with lower bounds, it's not going to work (at least for the way I'd want to implement this, which has to do with your second point). For PVP bounds purposes, the presence of an *implicit* lower bound (i.e., that `0` is the minimum version possible) is not my goal: it's to detect if the user explicitly stated a lower bound that should be respected (as opposed to omitting a lower bound, where tooling can add it in for the user).

23Skidoo:
LGTM now, so merging.

hvr:
c.f. #2774 which also contains similiar predicates

snoyberg:
Interesting, that looks like a very similar feature to what I was adding to stack

PR #2833, opened 2015-09-23
Prevent detailed-0.9 test suites from freezing on Windows (fixes #2762)

### LINKED ISSUE #2762: Detailed-0.9 test suite freezes on Windows
The test suite in `Cabal/tests/PackageTests/TestSuiteTests/LibV09/LibV09.cabal` freezes, whether I run the Cabal package tests or run `Setup.exe test` directly.  I also tried running the simpler test, BuildTestSuiteDetailedV09, which was replaced in #2508, and it also freezes on Windows.  Both tests pass on Linux.

I played around with the code that runs the tests as a sub-process and found something that works.  I'm not proposing it as a fix, since I still don't know the cause of the problem.  Here is the commit: grayjay@e69fe268a92592f122b67f4232249736e211ba5e

Summary:
* Allow `System.Process.createProcess` to create the handles to the sub-process, instead of creating the handles with `D.Compat.CreatePipe.createPipe` and passing them to `createProcess`.
* Read and write to the handles after the process is created.
* `rawSystemIOWithEnv'` is a modified copy of `rawSystemIOWithEnv`, since `rawSystemIOWithEnv` is called elsewhere.

grayjay:
This commit changes the way that Setup creates the `stdin` handle for the detailed test subprocess.  It allows `System.Process.createProcess` to create the handle, instead of passing in a handle from `D.Compat.CreatePipe.createPipe`.

I also added the function `D.S.Utils.createProcessWithEnv`, which is more general than the existing `rawSystemIOWithEnv`.

This change also allows Setup to run detailed tests when it is not compiled with `-threaded`.

/cc @ttuegel

### DIFF
--- a/Cabal/Distribution/Simple/Test/LibV09.hs
+++ b/Cabal/Distribution/Simple/Test/LibV09.hs
@@ -23,15 +23,14 @@ import Distribution.Simple.Setup
     ( TestFlags(..), TestShowDetails(..), fromFlag, configCoverage )
 import Distribution.Simple.Test.Log
 import Distribution.Simple.Utils
-    ( die, notice, rawSystemIOWithEnv, addLibraryPath )
+    ( die, debug, notice, createProcessWithEnv, addLibraryPath )
 import Distribution.System ( Platform (..) )
 import Distribution.TestSuite
 import Distribution.Text
 import Distribution.Verbosity ( normal )
 
-import Control.Concurrent (forkIO)
 import Control.Exception ( bracket )
-import Control.Monad ( when, unless, void )
+import Control.Monad ( when, unless )
 import Data.Maybe ( mapMaybe )
 import System.Directory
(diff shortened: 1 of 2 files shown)


ttuegel:
LGTM. Thanks!

Cabal/Distribution/Simple/Test/LibV09.hs
```diff
@@ -108,9 +97,22 @@ runTest pkg_descr lbi flags suite = do
...
+                                     CreatePipe (UseHandle wOut) (UseHandle wOut)
+
+        hPutStr wIn $ show (tempLog, PD.testName suite)
+        hClose wIn
+
+        -- Append contents of temporary log file to the final human-
+        -- readable log file
+        logText <- hGetContents rOut
+        -- Force the IO manager to drain the test output pipe
+        length logText `seq` return ()
```
23Skidoo: Could have used `evaluate`.

Issue #2205, opened 2014-11-06
Investigate link failures on Windows caused by using Cygwin instead of MSYS


23Skidoo:
In #2202, a user was trying to build `cabal-install` from the Cygwin shell (using native GHC/`cabal-install`). This caused link failures, which was solved by switching to MSYS2 shell.

Since we now get the `gcc` path from `ghc --info`, this was probably caused by Cygwin somehow tricking the network package's `configure` script into using a wrong version of `gcc`. We should look into whether it's possible to fix this by e.g. invoking `configure` with `CC=/path/to/gcc` or something like it.

jasagredo:
Nowadays, Haskell on Windows relies on the Mingw toolchain and not on Cygwin. I'm closing this ticket as it is about Cygwin, please re-open if you disagree.

Issue #1785, opened 2014-04-17
Can't build 1.20 on Windows


YoEight:
I get that error on `cabal build`

```
Preprocessing executable 'cabal' for cabal-install-1.20.0.0...
Linking dist/dist-sandbox-83973e47\build\cabal\cabal.exe ...
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x40de): undefined reference to `shutdownWinSock'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x4176): undefined reference to `shutdownWinSock'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x455c): undefined reference to `initWinSock'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x45d0): undefined reference to `getWSErrorDescr'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x472c): undefined reference to `getWSErrorDescr'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x4873): undefined reference to `shutdownWinSock'
C:\Users\sqlserver\cabal-1.20\cabal-install\.cabal-sandbox\x86_64-windows-ghc-7.8.2\network-2.4.2.2/libHSnetwork-2.4.2.2.a(Socket.o):fake:(.text+0x116d5): undefined reference to `newAcceptParams'
[...]

PR #2840, opened 2015-09-26
Use extra-prog-path in running build-type configure

gbaz:
Before this patch, configure scripts were always run just using whatever "sh" was lying around instead of using the searchpath setup -- hence e.g. on windows even with the searchpath set to point to a msys sh, the cygwin would get chosen, causing a mess. Also, the configure scripts weren't being _run_ with an augmented path, so they wouldn't manage to necessarily find the proper compiler toolchain for testing with autoconf. That is also corrected with this patch.

Also note that by putting the proper msys path in place, this means we get the proper uname for msys, which means that even in a generally cygwin system, a "config.guess" script should now properly detect msys for the build process.

I've verified that on a raw windows shell, without msys in my path, but with the proper pointers provided using extra-prog-path in my cabal/config, I can now configure and build the network library.

This _should_ work even in a cygwin shell, but I haven't tried yet :-)

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -66,8 +66,12 @@ import Distribution.PackageDescription.Parse
 import Distribution.PackageDescription.Configuration
          ( flattenPackageDescription )
 import Distribution.Simple.Program
-         ( defaultProgramConfiguration, addKnownPrograms, builtinPrograms
-         , restoreProgramConfiguration, reconfigurePrograms )
+         ( defaultProgramConfiguration, builtinPrograms
+         , restoreProgramConfiguration)
+import Distribution.Simple.Program.Db
+import Distribution.Simple.Program.Find
+import Distribution.Simple.Program.Run
+import Distribution.Simple.Program.Types
 import Distribution.Simple.PreProcess (knownSuffixHandlers, PPSuffixHandler)
 import Distribution.Simple.Setup
 import Distribution.Simple.Command
@@ -91,9 +95,8 @@ import Distribution.Simple.Haddock (haddock, hscolour)
 import Distribution.Simple.Utils
          (die, notice, info, warn, setupMessage, chattyTry,
           defaultPackageDesc, defaultHookedPackageDesc,
-          rawSystemExitWithEnv, cabalVersion, topHandler )
-import Distribution.System
-         ( OS(..), buildOS )
+          cabalVersion, topHandler )
+import Distribution.Utils.NubList
 import Distribution.Verbosity
 import Language.Haskell.Extension
 import Distribution.Version
@@ -106,10 +109,7 @@ import System.Environment(getArgs, getProgName)
 import System.Directory(removeFile, doesFileExist,
                         doesDirectoryExist, removeDirectoryRecursive)
 import System.Exit       (exitWith,ExitCode(..))
-import System.IO.Error   (isDoesNotExistError)
-import Control.Exception (throwIO)
 import Distribution.Compat.Environment (getEnvironment)
-import Distribution.Compat.Exception (catchIO)
 
 import Control.Monad   (when)
 import Data.Foldable   (traverse_)
@@ -626,7 +626,6 @@ autoconfUserHooks
 runConfigureScript :: Verbosity -> Bool -> ConfigFlags -> LocalBuildInfo
                    -> IO ()
 runConfigureScript verbosity backwardsCompatHack flags lbi = do
-
   env <- getEnvironment
   let programConfig = withPrograms lbi
   (ccProg, ccFlags) <- configureCCompiler verbosity programConfig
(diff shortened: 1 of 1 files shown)


gbaz:
While #1785 and is closed, this should actually also better address some of the issues that came up there. I wonder if it may have some impact on #2205 as well.

ezyang:
LGTM, but you need to pass the tests ;)

BardurArantsson:
Please squash into a single commit before merge! Having individual commits that don't build is somewhere between annoying and painful if one ever needs to do "git bitsect".

gbaz:
@ezyang is that standard cabal practice here? I think i've seen branches that had unsquished bits.

gbaz:
(also, I guess you can remove the action-required tag -- I don't have the rights to do so)

BardurArantsson:
Even if it isn't standard practice I'll argue that it should be done in this particular case, regardless. Basically: Having a commit like "whoops, ..." separate from the "main" commit adds  nothing of any value to anyone who's going to be looking back through the history or using "git annotate". The commit history should be meaningful to *humans*.

Of course, if there are logically separate commits which each add value on their own, then they *shouldn't* just be blindly squashed.

What I'm basically saying is that the commits should be "curated" such that they form a logical progression for *human* reviewer, each one of them builds individually and they don't needlessly clutter up history with "wip" or "whoops" or "blah" commits. Believe me, you wouldn't ever want to see the un-curated commit history of any of my feature branches -- they usually go into the tens or hundres of "wip" commits :). This also means that you can commit willy-nilly to your topic branch while you're working on it and clean up one you're ready for final review+merge.

This has been standard practice on every single git-based project I've ever worked on and it works *really well*. (Especially since GitHub actually supports rebase+force-push to a pull request branch quite well.)

23Skidoo:
@gbaz 
> is that standard cabal practice here? I think i've seen branches that had unsquished bits.

Not in general case, but sometimes I ask people to squish/rebase to make code easier to review. And @BardurArantsson is right about `git bisect`.

@BardurArantsson 
> Having individual commits that don't build is somewhere between annoying and painful if one ever needs to do "git bitsect".

I agree, but unfortunately it's the reality that one has to live with. Use `bisect skip` or return 125 from your `bisect run` script.

ezyang:
This one should be squashed. There's ~no reason not to, it's really small.

With larger sets I'm more sympathetic to not squashing.

BardurArantsson:
Sorry, didn't want to derail. I'll be quiet after this post.

@ezyang : Agreed about larger sets. My criterion is spelled out here

> Of course, if there are logically separate commits which each add value on their own, then they shouldn't just be blindly squashed.

... and it's one that I think makes pretty good sense even though it usually requires a judgement call about "adds value" and "logically separate".

gbaz:
Ok, let's sort out this last issue and I'm happy to squash -- just wanted to know what the standard practice is here (I know some people run relatively squash-free repos on principle, etc).

So on this last issue, I'm trying to figure out what is being recommended. You're suggesting I don't step on the whole env, but only the PATH and CFLAGS variables I guess? Note that we still will have certain clunkiness here. We want to _change_ the PATH and CFLAGS variables, not just step on them willy-nilly. With PATH there's a helper function used elsewhere, but CFLAGS we still have to fish it out of the env and alter it -- it is just the case that we won't clobber the parts of the env that stay the same -- so the net effect is maybe slightly more "precise" code but the cost will imho be further verbosity and less of a "wholemeal" character.

That said, I'll go ahead and refactor the commit as so. We can review it on this branch, and then I guess I'll need to make a new PR when I squash?

23Skidoo:
> We can review it on this branch, and then I guess I'll need to make a new PR when I squash?

No, you can just do `push -f`.

gbaz:
Ok, refactored and squashed. The refactor ended up reducing the linecount, so actually does feel cleaner overall :-)

Cabal/Distribution/Simple.hs
```diff
@@ -636,32 +635,24 @@ runConfigureScript verbosity backwardsCompatHack flags lbi = do
   -- to ccFlags
   -- We don't try and tell configure which ld to use, as we don't have
   -- a way to pass its flags too
-  let env' = appendToEnvironment ("CFLAGS",  unwords ccFlags)
-             env
+  let extraPath = fromNubList $ configProgramPathExtra flags
+  let cflagsEnv = maybe (unwords ccFlags) (++ (" " ++ unwords ccFlags)) $ lookup "CFLAGS" env
+      pathEnv = maybe (intercalate ";" extraPath) ((intercalate ";" extraPath ++ ";")++) $ lookup "PATH" env
```
23Skidoo: I think you need to use [`searchPathSeparator`](http://hackage.haskell.org/package/filepath-1.4.0.0/docs/System-FilePath-Windows.html#v:searchPathSeparator) here.

Cabal/Distribution/Simple.hs
```diff
@@ -636,32 +635,24 @@ runConfigureScript verbosity backwardsCompatHack flags lbi = do
...
-      = action
-          `catchIO` \ioe -> if isDoesNotExistError ioe
-                              then die notFoundMsg
-                              else throwIO ioe
-
-    notFoundMsg = "The package has a './configure' script. This requires a "
-               ++ "Unix compatibility toolchain such as MinGW+MSYS or Cygwin."
+    notFoundMsg = "The package has a './configure' script. If you are on Windows, This requires a "
+               ++ "Unix compatibility toolchain such as MinGW+MSYS or Cygwin. "
+               ++ "If you are not on Windows, ensure that an 'sh' command is discoverable in your path."
```
23Skidoo: > an 'sh' command

s/an/a/?

gbaz: "an" since "sh" is pronounced "ess-eich" :-) https://www.englishclub.com/pronunciation/a-an.htm

23Skidoo:
LGTM.

Probably could be simplified still further by extending `getEffectiveEnvironment` with support for appending/prepending instead of `getEnvironment`/`lookup`/`maybe`, but that's not a requirement for merging.

23Skidoo:
Merged, thanks!

PR #2841, opened 2015-09-26
Carry package dbs into testsuite

enolan:
Prior to this patch, the testsuite used your global package db, and
sometimes, the inplace package db, which led to errors if you didn't
have old-time installed globally. They looked like this:

BuildDeps/SameDepsAllRound:          Cabal result was Result {successful = False, success = Failure, outputText = "\"/home/enolan/cabal/Cabal/tests/Setup configure --user -w  /home/enolan/.nix-profile/bin/ghc\" in PackageTests/BuildDeps/SameDepsAllRound\nConfiguring SameDepsAllRound-0.1...\nSetup: At least the following dependencies are missing:\nold-time -any\n"}

### DIFF
--- a/Cabal/tests/PackageTests.hs
+++ b/Cabal/tests/PackageTests.hs
@@ -39,6 +39,7 @@ import PackageTests.ReexportedModules.Check
 
 import Distribution.Simple.Configure
     ( ConfigStateFileError(..), findDistPrefOrDefault, getConfigStateFile )
+import Distribution.Simple.Compiler (PackageDB(..))
 import Distribution.Simple.LocalBuildInfo (LocalBuildInfo(..))
 import Distribution.Simple.Program.Types (programPath)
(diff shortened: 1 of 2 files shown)


enolan:
Fixed to compile on base<4.8.

enolan:
Oops, deleted branch by accident

23Skidoo:
Hmm, is there a guarantee that old-time is installed in the in-place package DB?

enolan:
The in-place package DB only includes the new Cabal library. The idea of this patch is to work in the case that you made a sandbox and ran `./Setup --package-db=.cabal-sandbox/foo`. If you did `cabal install --only-dependencies --enable-tests` everything necessary will be either in the global package db or the sandbox one. `old-time` is a dependency of `package-tests`.

23Skidoo:
OK, this looks good, merging. Thanks!

PR #2844, opened 2015-09-27
Fail more gracefully when tests can't be run

enolan:
If cabal-install isn't built yet, we can't run it's tests yet. Prior to
this commit, if you tried running "./Setup test" on cabal-install without
having built cabal-install yet you got:

package-tests: dist/build/cabal: canonicalizePath: does not exist (No such file or directory)

Now you get:

package-tests: user error (The program 'cabal' is required but it could not be found.)

I moved the canonicalization logic into findProgramOnSearchPath. This is
a change to the semantics of the library, but shouldn't break anything.

### DIFF
--- a/Cabal/Distribution/Simple/Program/Find.hs
+++ b/Cabal/Distribution/Simple/Program/Find.hs
@@ -35,7 +35,7 @@ import Distribution.Simple.Utils
 import Distribution.System
          ( OS(..), buildOS )
 import System.Directory
-         ( findExecutable )
+         ( canonicalizePath, findExecutable )
 import Distribution.Compat.Environment
          ( getEnvironment )
 import System.FilePath
@@ -109,7 +109,7 @@ findProgramOnSearchPath verbosity searchpath prog = do
     findFirstExe (f:fs) = do
       isExe <- doesExecutableExist f
       if isExe
-        then return (Just f)
+        then Just `fmap` canonicalizePath f
         else findFirstExe fs
 
 -- | Interpret a 'ProgramSearchPath' to construct a new @$PATH@ env var.
--- a/cabal-install/tests/IntegrationTests.hs
+++ b/cabal-install/tests/IntegrationTests.hs
@@ -15,7 +15,7 @@ import Distribution.Simple.Program.Builtin (ghcPkgProgram)
 import Distribution.Simple.Program.Db
         (defaultProgramDb, requireProgram, setProgramSearchPath)
 import Distribution.Simple.Program.Find
-        (ProgramSearchPathEntry(ProgramSearchPathDir), defaultProgramSearchPath)
+        (ProgramSearchPathEntry(ProgramSearchPathDir))
 import Distribution.Simple.Program.Types
         ( Program(..), simpleProgram, programPath)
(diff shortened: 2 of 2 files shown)


23Skidoo:
Any ideas about what caused the build bot failure?

enolan:
It's the AMP, I think. `<$>` got moved into Prelude. Needs some CPP to compile on <7.10. I'll fix it tonight.

enolan:
Fixed now :)

Cabal/Distribution/Simple/Program/Find.hs
```diff
@@ -109,7 +109,7 @@ findProgramOnSearchPath verbosity searchpath prog = do
     findFirstExe (f:fs) = do
       isExe <- doesExecutableExist f
       if isExe
-        then return (Just f)
+        then Just `fmap` canonicalizePath f
```
23Skidoo: I don't think this is needed, this function already returns an absolute path.

23Skidoo:
So the problem is that `canonicalizePath` exits with an error if the `dist/build/cabal` directory doesn't exist. I suggest adding a `doesDirectoryExist` check before the `canonicalizePath` path line if the goal is to improve the error message.

Cabal/Distribution/Simple/Program/Find.hs
enolan: It only returns absolute paths if all of the `ProgramSearchPathDir`s are absolute. That's the whole point. You'll notice my change is the only call to `canonicalizePath` in the function.

cabal-install/tests/PackageTests.hs
```diff
@@ -60,8 +60,9 @@ main = do
     distPref <- findDistPrefOrDefault NoFlag
     -- Use the default builddir for all of the subsequent package tests
     setEnv "CABAL_BUILDDIR" "dist"
-    buildDir <- canonicalizePath (distPref </> "build/cabal")
-    let programSearchPath = ProgramSearchPathDir buildDir : defaultProgramSearchPath
```
enolan: This can get the wrong cabal if dist/build/cabal exists but is empty.

enolan:
The counterargument is that we already have a system for finding executables and throwing errors if they aren't there. It makes sense to me to use it. The diffstat for these changes is negative if you ignore comments.

phadej:
AFAICS the original issue is already fixed in master. Can @enolan confirm?

enolan:
Bug is still there. On 39231018:
```
Test suite integration-tests: RUNNING...
integration-tests: dist/build/cabal: canonicalizePath: does not exist (No such file or directory)
```

enolan:
Updated to work with the newer `IntegrationTests` suite. Don't pull yet, still need to handle #2910

quasicomputational:
Is this still relevant? The test-suite's gone through some significant changes in the past few years and `cabal-install/tests/IntegrationTests.hs` doesn't even exist any more. 40846f0698567422cd63b2f38bea9c5d06efdf83 suggests that this PR isn't relevant since the integration tests don't use the `cabal` binary any more, and `cabal-testsuite` has stuff for test with and without a `cabal` binary.

PR #2846, opened 2015-09-29
Implement ComponentInstanceId, replacing PackageKey and InstalledPack…

ezyang:
…ageId.

Today in Cabal, when you build and install a package, it is
uniquely identified using an InstalledPackageId which is computed
using the ABI hash of the library that was installed.  There
are few problems with doing it this way:

    - In a Nix-like world, we should instead uniquely identify
      build products by some sort of hash on the inputs to the
      compilation (source files, dependencies, flags).  The ABI
      hash doesn't capture any of this!

    - An InstalledPackageId suggests that we can uniquely identify
      build products by hashing the source and dependencies of
      a package as a whole.  But Cabal packages contain many components:
      a library, test suite, executables, etc.  Currently, when
      we say InstalledPackageId, we are really just talking about
      the dependencies of the library; however, this is unacceptable
      if a Cabal package can install multiple libraries; we need
      different identifiers for each.

    - We've also needed to compute another ID, which we've called
      the "package key", which is to be used for linker symbols
      and type equality GHC-side.  It is confusing what the distinction
      between this ID and InstalledPackageIds are; the main reason
      we needed another ID was because the package key was needed
      prior to compilation, whereas the ABI hash was only available
      afterwards.

This patch replaces InstalledPackageId and PackageKey with a
new identifier called ComponentInstanceId, which has the following
properties:

    - It is computed per-component, and consists of a package
      name, package version, hash of the ComponentInstanceIds
      of the dependencies it is built against, and the name
      of the component.  For example, "foo-0.1-abcdef" continues
      to identify the library of package foo-0.1, but
      "foo-0.1-123455-foo.exe" would identify the executable,
      and "foo-0.1-abcdef-bar" would identify a private sub-library
      named bar.

    - It is passed to GHC to be used for linker symbols and
      type equality.  So as far as GHC is concerned, this is
      the end-all be-all identifier.

    - Cabal the library has a simple, default routine for computing
      a ComponentInstanceId which DOES NOT hash source code;
      in a later patch Duncan is working on, cabal-install can
      specify a more detailed ComponentInstanceId for a package
      to be built with.

Here are some knock-on effects:

    - 'id' is a ComponentInstanceId

    - 'depends' is now a list of ComponentInstanceIds

    - New 'abi' field to record what the ABI of a unit is (as it is no longer
      computed by looking at the output of ghc --abi-hash).

    - The 'HasInstalledPackageId' typeclass is renamed to
      'HasComponentInstanceId'.

    - GHC 7.10 has explicit compatibility handling with
      a 'compatPackageKey' (an 'ComponentInstanceId') which is
      in a compatible format.  The value of this is read out
      from the 'key' field.

cc @dcoutts


23Skidoo:
:+1: Nice to see package keys and installed package IDs unified, that distinction was confusing.

Cabal/Distribution/InstalledPackageInfo.hs
```diff
@@ -66,17 +65,19 @@ import qualified Distribution.Compat.ReadP as Parse
 import Distribution.Compat.Binary  (Binary)
 import Data.Maybe   (fromMaybe)
 import GHC.Generics (Generic)
+import qualified Data.Char as Char
 
 -- -----------------------------------------------------------------------------
 -- The InstalledPackageInfo type
 
-
+-- For BC reasons, we continue to name this record an InstalledPackageInfo;
```
23Skidoo: What would break if we renamed this type?

Cabal/Distribution/Simple/Build/Macros.hs
```diff
@@ -99,10 +99,10 @@ generateMacros prefix name version =
 -- | Generate the @CURRENT_PACKAGE_KEY@ definition for the package key
 --   of the current package, if supported by the compiler.
 --   NB: this only makes sense for definite packages.
-generatePackageKeyMacro :: LocalBuildInfo -> String
-generatePackageKeyMacro lbi
+generateComponentInstanceIdMacro :: LocalBuildInfo -> String
+generateComponentInstanceIdMacro lbi
   | packageKeySupported (compiler lbi) =
-      "#define CURRENT_PACKAGE_KEY \"" ++ display (localPackageKey lbi) ++ "\"\n\n"
+      "#define CURRENT_PACKAGE_KEY \"" ++ display (localComponentInstanceId lbi) ++ "\"\n\n"
```
23Skidoo: We may want to deprecate `CURRENT_PACKAGE_KEY` and introduce a `CURRENT_COMPONENT_INSTANCE_ID` macro.

23Skidoo:
LGTM. Feel free to merge.

Cabal/Distribution/InstalledPackageInfo.hs
ezyang: Not much, I just didn't git grep rename it yet.

ezyang:
merged

Issue #2845, opened 2015-09-29
Get rid of Cabal bootstrapping


ezyang:
Suggested patch:

```
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -19,7 +19,7 @@ description:
   organizing, and cataloging Haskell libraries and tools.
 category: Distribution
 cabal-version: >=1.10
-build-type: Custom
+build-type: Simple
 -- Even though we do use the default Setup.lhs it's vital to bootstrapping
 -- that we build Setup.lhs using our own local Cabal source code.
```

(well, also remove the comments). I feel like Cabal's Cabal file is stable enough and supported by enough previous versions that this is not really a necessary step anymore.

ttuegel:
I suggest we follow the same rule as with GHC versions: if we haven't made any breaking changes to the package description in 3 years, then we can keep it at `build-type: Simple`. If we need to make breaking changes in the future, we can always go back to `build-type: Custom`. Travis will keep us honest here.

23Skidoo:
Fixed by #2845.

PR #2847, opened 2015-09-30
Don't bootstrap Cabal, fixing #2845.

ezyang:


23Skidoo:
OK, let's merge. @dcoutts can revert this patch later if he disagrees.

Issue #1839, opened 2014-05-03
`defaultRemoteRepo` points to obsolete folder


hvr:
I've noticed in the access logs, that even recent `cabal-install` versions still use `/packages/archive/`-based download paths for the tarballs, see [code here](https://github.com/haskell/cabal/blob/d64c539ebb81da410b74d4d2dacda60f7559497b/cabal-install/Distribution/Client/Config.hs#L268-L272)

However, this results in an additional HTTP request due to the 301-redirect. Here's what typical requests for e.g. `cabal update` followed by `cabal install streaming-commons` look like:

```
… "GET /packages/archive/00-index.tar.gz HTTP/1.1" 301 5 "-" "cabal-install/1.20.0.0" "-"
… "GET /packages/index.tar.gz HTTP/1.1" 200 6962668 "-" "cabal-install/1.20.0.0" "-"
… "GET /packages/archive/streaming-commons/0.1.2/streaming-commons-0.1.2.tar.gz HTTP/1.1" 301 5 "-" "cabal-install/1.20.0.0" "-"
… "GET /package/streaming-commons-0.1.2/streaming-commons-0.1.2.tar.gz HTTP/1.1" 200 75785 "-" "cabal-install/1.20.0.0" "-"
```

Is there a reason not to update to the new URL scheme (NB: it's not just the prefix that changes)?

edsko:
`defaultRemoteRepo` no longers refers to the legacy URI (https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Config.hs#L483), but of course lots of people will have cabal `config` files that have this URI baked in. This was also a problem in the Hackage Security work, where the `hackage-security` library would try to download files such as http://hackage.haskell.org/packages/archive/root.json instead of (http://hackage.haskell.org/root.json). In my `using-hackage-security` I've changed [`addInfoForKnownRepos`](https://github.com/edsko/cabal/blob/using-hackage-security/cabal-install/Distribution/Client/Config.hs#L509) to automatically change these legacy URIs to new URIs:

``` haskell
addInfoForKnownRepos :: RemoteRepo -> RemoteRepo
addInfoForKnownRepos repo@RemoteRepo{ remoteRepoName = "hackage.haskell.org" } =
      tryHttps
    . addCredentials
    $ if isOldHackageURI (remoteRepoURI repo)
        then defaultRemoteRepo
        else repo
  where
    tryHttps       r = r { remoteRepoShouldTryHttps = True }
[...]

PR #2849, opened 2015-10-01
Always use new hackage URI

bennofs:
This addresses the issue #1839.
As noted in the commit message, it also fixes a small bug in the upload URI generation when
the repo URI does not end with a slash.


bennofs:
Oops, please don't merge this right now, I just found out that `relativeTo` also doesn't have the right behavior (it just discards the old uriPath).

23Skidoo:
Please also fix the Travis errors while you're at it.

bennofs:
Ok, mistake and travis build fixed.

23Skidoo:
LGTM, merging. Thanks!

PR #2850, opened 2015-10-01
Use a common sandbox for building Cabal and cabal-install.

23Skidoo:
Should speed up the Travis build a little.


23Skidoo:
OK, seems to work now.

Issue #2080, opened 2014-08-31
Implement 'cabal upload --haddock'


23Skidoo:
See [this comment](https://github.com/haskell/cabal/issues/2079#issuecomment-53965427). Uploading locally-built Haddock documentation to Hackage is a common use case, and we should support it directly.

bennofs:
I would like to work on this. However, I had the following design in mind:

1. add a new option to `cabal sdist`: use `--doc` to generate a documentation tarball.
2. add a new option to `cabal upload`: use `--doc` to upload a documentation tarball.
3. Perhaps allow `cabal upload` to be run without arguments (and `cabal upload --doc`), which will automatically generate sdists/documentation tarballs and upload them.

That way, `upload`'s functionality stays constrained to uploading stuff to hackage and it's consistent with how uploading source tarballs works today. Also, I opted for `cabal sdist --doc` instead of `cabal haddock --for-hackage` since the latter would clobber existing documentation in `dist/doc`, which users still might want to use for development (and that documentation then of course shouldn't refer to the hackage links).

Would that design be ok?

bennofs:
Now, after looking at the code some more, I'm still not exactly sure where all the parts for this fit. I think there are three parts:

1. generate the documentation with links etc suitable for hackage upload. This now looks like it would fit best into the `haddock` command, in Cabal-the-library.
2. from that, generate a documentation tarball. This must be done in `cabal-install`, since only that may depend on the library required for generating tarballs.
3. upload a documentation tarball. This should go in `cabal upload`, provided by `cabal-install`.

I'm reasonably sure about 1 and 3, but I don't know what to do about 2. Should it be part of `cabal sdist`? (since that already deals with creating temporary directories and creating archives from them). Or rather a special, cabal-install only, flag for `cabal haddock`?

23Skidoo:
@bennofs I'd say that 2 should also be a part of `cabal haddock`, and I think you should reuse the `--for-hackage` option for this. The workflow would then be:

```
$ cabal haddock --for-hackage
$ cabal upload --doc
```

Having to remember to call `cabal sdist` in between is too complicated, IMO. Also, `cabal upload --doc` should invoke `cabal haddock --for-hackage` if there's no existing documentation tarball (ideally, it'd also look at source timestamps and regenerate it if needed, but that can be a future improvement).

bitemyapp:
I agree with the UX improvements @23Skidoo has laid out here.

bennofs:
[...]

PR #2852, opened 2015-10-04
cabal haddock: add --for-hackage flag

bennofs:
This is the first step to implement #2080. It adds a new flag
to `cabal haddock`, called `--for-hackage`, which will generate
documentation suitable for upload to hackage. It's only a collection
of flags, and matches the flags used by the hackage doc builder.


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

Issue #2179, opened 2014-10-24
Attempting to install a non-existent dependency does not suggest the package doesn't exist


owst:
If we try to directly install a non-existent package we get a warning that the package doesn't exist:

    > cabal install blergh
    cabal: There is no package named 'blergh'.
    You may need to run 'cabal update' to get the latest list of available
    packages.

However, if we add the same non-existent package as a dependency, the warning we get is obscured:

    > cabal install --only-dep
    Resolving dependencies...
    cabal: Could not resolve dependencies:
    next goal: foo (user goal)
    rejecting: foo-1.0 (global constraint requires ==0.1.0.0)
    trying: foo-0.1.0.0
    next goal: blergh (dependency of foo-0.1.0.0)
    Dependency tree exhaustively searched.

    Note: when using a sandbox, all packages are required to have consistent
    dependencies. Try reinstalling/unregistering the offending packages or
    recreating the sandbox.

My proposal is to change the wording (in the case where *no* rejections were made) of the failure thus:

    > cabal install --only-dep
    Resolving dependencies...
    cabal: Could not resolve dependencies:
    next goal: foo (user goal)
    rejecting: foo-1.0 (global constraint requires ==0.1.0.0)
    trying: foo-0.1.0.0
    next goal: blergh (dependency of foo-0.1.0.0)
    There is no package named 'blergh'.
    You may need to run 'cabal update' to get the latest list of available
    packages.

    Note: when using a sandbox, all packages are required to have consistent
    dependencies. Try reinstalling/unregistering the offending packages or
    recreating the sandbox.

grayjay:
The warning is a little better after #2853:

    $ cabal install --only-dep
    Resolving dependencies...
    cabal: Could not resolve dependencies:
    trying: foo-1.0 (user goal)
    next goal: blergh (dependency of foo-1.0)
    fail (unknown package: blergh)
    Dependency tree exhaustively searched.

23Skidoo:
I guess this can be closed then.

Issue #2816, opened 2015-09-10
Cabal should offer a suggestion for how to resolve dep when a package doesn't exist at all in cached index


ezyang:
Recently I attempted to build Stack and I got this error:

```
[ezyang@hs01 stack]$ cabal install
Resolving dependencies...
cabal: Could not resolve dependencies:
trying: stack-0.1.4.1 (user goal)
next goal: binary-tagged (dependency of stack-0.1.4.1)
```

The reason was that I hadn't run `cabal update`. It would be helpful if Cabal gave a little tip about this.

grayjay:
This is a duplicate of #2179.

PR #2853, opened 2015-10-04
Always warn when 'cabal install' cannot find a dependency

grayjay:
This commit prevents cabal from filtering out messages about missing
packages when summarizing the solver trace.

In this example, it adds the line `fail (unknown package: either)`.

    $ ../dist/build/cabal/cabal install 
    Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
    update' to download it.
    Resolving dependencies...
    trying: stack-0.1.4.1 (user goal)
    next goal: either (dependency of stack-0.1.4.1)
    fail (unknown package: either)

This commit partially fixes issues #2179 and #2816 but doesn't remind the user to run `cabal update`.  If we mention `update`, I think it should be a separate message, because this line can appear many times in the full trace.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency/Modular/Message.hs
+++ b/cabal-install/Distribution/Client/Dependency/Modular/Message.hs
@@ -47,6 +47,9 @@ showMessages p sl = go [] 0
     go v l (TryF qfn b : Enter : Failure c fr : Leave : ms) = (atLevel (add (F qfn) v) l $ "rejecting: " ++ showQFNBool qfn b ++ showFR c fr) (go v l ms)
     go v l (TryS qsn b : Enter : Failure c fr : Leave : ms) = (atLevel (add (S qsn) v) l $ "rejecting: " ++ showQSNBool qsn b ++ showFR c fr) (go v l ms)
     go v l (Next (Goal (P qpn) gr) : TryP qpn' i : ms@(Enter : Next _ : _)) = (atLevel (add (P qpn) v) l $ "trying: " ++ showQPNPOpt qpn' i ++ showGRs gr) (go (add (P qpn) v) l ms)
+    go v l (Next (Goal (P qpn) gr) : Failure c fr : ms) =
+        let v' = add (P qpn) v
+        in (atLevel v' l $ showPackageGoal qpn gr) $ (atLevel v' l $ showFailure c fr) (go v l ms)
     go v l (Failure c Backjump : ms@(Leave : Failure c' Backjump : _)) | c == c' = go v l ms
     -- standard display
     go v l (Enter                  : ms) = go v          (l+1) ms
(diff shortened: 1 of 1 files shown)


kosmikus:
Looking at this now ...

cabal-install/Distribution/Client/Dependency/Modular/Message.hs
```diff
@@ -54,10 +57,14 @@ showMessages p sl = go [] 0
...
     go v l (TryF qfn b             : ms) = (atLevel (add (F qfn) v) l $ "trying: " ++ showQFNBool qfn b) (go (add (F qfn) v) l ms)
     go v l (TryS qsn b             : ms) = (atLevel (add (S qsn) v) l $ "trying: " ++ showQSNBool qsn b) (go (add (S qsn) v) l ms)
-    go v l (Next (Goal (P qpn) gr) : ms) = (atLevel (add (P qpn) v) l $ "next goal: " ++ showQPN qpn ++ showGRs gr) (go v l ms)
+    go v l (Next (Goal (P qpn) gr) : ms) = (atLevel (add (P qpn) v) l $ showPackageGoal qpn gr) (go v l ms)
     go v l (Next _                 : ms) = go v l ms -- ignore flag goals in the log
     go v l (Success                : ms) = (atLevel v l $ "done") (go v l ms)
-    go v l (Failure c fr           : ms) = (atLevel v l $ "fail" ++ showFR c fr) (go v l ms)
+    go v l (Failure c fr           : ms) = (atLevel v l $ showFailure c fr) (go v l ms)
+
+    showPackageGoal qpn gr = "next goal: " ++ showQPN qpn ++ showGRs gr
```
kosmikus: I somewhat prefer if local functions have type signatures, too.

cabal-install/Distribution/Client/Dependency/Modular/Message.hs
```diff
@@ -47,6 +47,9 @@ showMessages p sl = go [] 0
     go v l (TryF qfn b : Enter : Failure c fr : Leave : ms) = (atLevel (add (F qfn) v) l $ "rejecting: " ++ showQFNBool qfn b ++ showFR c fr) (go v l ms)
     go v l (TryS qsn b : Enter : Failure c fr : Leave : ms) = (atLevel (add (S qsn) v) l $ "rejecting: " ++ showQSNBool qsn b ++ showFR c fr) (go v l ms)
     go v l (Next (Goal (P qpn) gr) : TryP qpn' i : ms@(Enter : Next _ : _)) = (atLevel (add (P qpn) v) l $ "trying: " ++ showQPNPOpt qpn' i ++ showGRs gr) (go (add (P qpn) v) l ms)
+    go v l (Next (Goal (P qpn) gr) : Failure c fr : ms) =
+        let v' = add (P qpn) v
+        in (atLevel v' l $ showPackageGoal qpn gr) $ (atLevel v' l $ showFailure c fr) (go v l ms)
```
kosmikus: Could you explain the idea here? If a package fails immediately, without any tries, then it already indicates that the package does not exist? And in such a case, you add the package to the filter, so that the message will always be shown? Wouldn't it be easier to just directly catch `BuildFailureNotInIndex` and always show that? Are there other failures that will now also be shown, and if so, is that desirable, or unexpected?

kosmikus:
Thanks a lot for this. I have some minor questions, but I'm in principle in favour of this to go in as soon as possible.

cabal-install/Distribution/Client/Dependency/Modular/Message.hs
grayjay: @kosmikus Thanks for reviewing this.

I assumed that when `go` is called with `Next (Goal (P qpn) _) : _`, the following messages also relate to `qpn`, and it would make sense for `P qpn` to be added to the stack of variables.  However, there isn't an easy way to determine when to pop the variable.  I instead tried to replicate that behavior by adding a special case for a failure after a package goal choice.  Since `Next` is only followed by `Failure` or `TryP`, and the `TryP` case already adds the right variable to the stack, I thought that my change would handle the remaining cases.

It seems like `BuildFailureNotInIndex` is the only `FailReason` that should come directly after a package goal choice, so I think that there is no difference in behavior between my PR and your suggestion.  I would still need to match on the message before `BuildFailureNotInIndex`, because `BuildFailureNotInIndex` contains a `PN`, and the stack needs a `QPN`.  Would you like me to make that change?

kosmikus: @grayjay Thanks for the reply. I don't necessarily think that a change is required, and I currently believe that (y)our reasoning is correct and both versions do the same. But adding a comment explaining what this does might be good.

grayjay:
@kosmikus I added the type signatures and a comment.  I can combine the commits if you want.

kosmikus:
Thanks again.

grayjay:
Thanks!

PR #2859, opened 2015-10-11
Use configured compiler for cabal sandbox hc-pkg

### LINKED ISSUE #1935: sandbox-config's `package-db` path contains hardcoded GHC version
When creating a new sandbox via `cabal sandbox init`, the auto-generated `cabal.sandbox.config` contains the following settings:

```
local-repo: /tmp/sb/.cabal-sandbox/packages
logs-dir: /tmp/sb/.cabal-sandbox/logs
world-file: /tmp/sb/.cabal-sandbox/world
user-install: False
package-db: /tmp/sb/.cabal-sandbox/x86_64-linux-ghc-7.8.2-packages.conf.d
build-summary: /tmp/sb/.cabal-sandbox/logs/build.log

install-dirs
  prefix: /tmp/sb/.cabal-sandbox
  bindir: $prefix/bin
  libdir: $prefix/lib
  libsubdir: $arch-$os-$compiler/$pkgid
  libexecdir: $prefix/libexec
  datadir: $prefix/share
  datasubdir: $arch-$os-$compiler/$pkgid
  docdir: $datadir/doc/$arch-$os-$compiler/$pkgid
  htmldir: $docdir/html
  haddockdir: $htmldir
  sysconfdir: $prefix/etc
```

The problematic line is

```
package-db: /tmp/sb/.cabal-sandbox/x86_64-linux-ghc-7.8.2-packages.conf.d
```

which makes it harder to share the same `cabal.sandbox.config` for multiple GHC versions. Ideally it would read like

```
package-db: /tmp/sb/.cabal-sandbox/$arch-$os-$compiler-packages.conf.d
```

Which, however, doesn't seem to be supported as `cabal-install` doesn't seem to support string-interpolation in the `package-db` field.

adamgundry:
For most commands (e.g. `cabal install`), the `package-db` field in `cabal.sandbox.config` is ignored, and the path is reconstructed from the `prefix` and current compiler instead.  This is arguably the right behaviour, because the right package DB depends on the arch/compiler, so it doesn't make sense to specify only one.  This commit makes `cabal sandbox hc-pkg` behave similarly, to avoid invoking ghc-pkg on an incompatible package DB (fixes #1935).

Moreover, the compiler version to use is now picked up from the most recently configured compiler, if any.  Otherwise, the global default compiler is used, as before.

I think the `package-db` field should probably be removed from `cabal.sandbox.config` completely, but I haven't yet investigated this.

@luite hopefully this will make `cabal configure -w ghcjs && cabal sandbox hc-pkg list` just do the right thing; please yell if not!

### DIFF
--- a/cabal-install/Distribution/Client/Sandbox.hs
+++ b/cabal-install/Distribution/Client/Sandbox.hs
@@ -39,7 +39,7 @@ module Distribution.Client.Sandbox (
     updateInstallDirs,
 
     -- FIXME: move somewhere else
-    configPackageDB', configCompilerAux'
+    configPackageDB', configCompilerAux', getPersistOrConfigCompiler
   ) where
 
 import Distribution.Client.Setup
--- a/cabal-install/Distribution/Client/Sandbox.hs
+++ b/cabal-install/Distribution/Client/Sandbox.hs
@@ -65,7 +65,8 @@ import Distribution.Client.Sandbox.PackageEnvironment
   , createPackageEnvironmentFile, classifyPackageEnvironment
   , tryLoadSandboxPackageEnvironmentFile, loadUserConfig
   , commentPackageEnvironment, showPackageEnvironmentWithComments
-  , sandboxPackageEnvironmentFile, userPackageEnvironmentFile )
+  , sandboxPackageEnvironmentFile, userPackageEnvironmentFile
+  , sandboxPackageDBPath )
 import Distribution.Client.Sandbox.Types      ( SandboxPackageInfo(..)
                                               , UseSandbox(..) )
 import Distribution.Client.SetupWrapper
@@ -82,7 +83,10 @@ import Distribution.Simple.Compiler           ( Compiler(..), PackageDB(..)
 import Distribution.Simple.Configure          ( configCompilerAuxEx
                                               , interpretPackageDbFlags
                                               , getPackageDBContents
+                                              , maybeGetPersistBuildConfig
+                                              , findDistPrefOrDefault
                                               , findDistPref )
+import qualified Distribution.Simple.LocalBuildInfo as LocalBuildInfo
 import Distribution.Simple.PreProcess         ( knownSuffixHandlers )
 import Distribution.Simple.Program            ( ProgramConfiguration )
 import Distribution.Simple.Setup              ( Flag(..), HaddockFlags(..)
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -791,3 +797,18 @@ configCompilerAux' configFlags =
   configCompilerAuxEx configFlags
     --FIXME: make configCompilerAux use a sensible verbosity
     { configVerbosity = fmap lessVerbose (configVerbosity configFlags) }
+
+-- | Try to read the most recently configured compiler from the
```
23Skidoo: I wonder if this function can be also used in other places in this file.

23Skidoo:
Looks OK.

cabal-install/Distribution/Client/Sandbox.hs
adamgundry: I wondered this too, which is why I factored it out as a separate function. But I haven't really investigated the other uses of `configCompilerAux'`, `configCompilerAuxEx`, and so on...

luite:
Hmm, it doesn't appear to quite work correctly for GHCJS. Packages end up in the correct package db, but `cabal sandbox hc-pkg list` does not seem to read the configured compiler. Setting the default compiler to ghcjs with a line `compiler: ghcjs` in a local `cabal.config` file makes the command appear to use the correct dir, but the package list is empty.

I'll investigate a bit more.

luite:
Oh I was looking at the wrong thing, it is using `ghcjs-pkg` correctly, but passes the wrong path.

luite:
Erm, next time I should build the correct branch first

luite:
`cabal sandbox hc-pkg` appears to work now. `cabal sandbox exec` does not seem to use the currently configured compiler. can this be fixed in the same way?

adamgundry:
I'm on it.

adamgundry:
Now `cabal exec` should be fixed, and a bogus warning that it was generating is now suppressed.

luite:
Works ok here with GHCJS / ghcjs-pkg. I'd be happy to merge this if no one objects.

23Skidoo:
LGTM, merging.

23Skidoo:
Merged, thanks!

PR #2862, opened 2015-10-11
Fix #2804: Text Platform roundtrip (with tests)

### LINKED ISSUE #2804: Text instance for Text.System.Platform can't do a round trip
Is that intentional?

~~~~ haskell
λ> simpleParse (display (Platform X86_64 Linux)) :: Maybe Platform
Nothing
~~~~

phadej:


23Skidoo:
LGTM.

phadej:
Have to fix the ambiguous parse though.
- Most obvious change would be to not support `-` dash in ident.

otherwise:
- `foo-bar-quux` is indeed ambiguous `Platform`

phadej:
ping @23Skidoo, now this is green. AFAICS parse part of `Text Platform` isn't used in `Cabal` itself, so there shouldn't be any regressions.

Cabal/Distribution/System.hs
```diff
@@ -193,8 +209,14 @@ buildPlatform = Platform buildArch buildOS
 -- Utils:
 
 ident :: Parse.ReadP r String
-ident = Parse.munch1 (\c -> Char.isAlphaNum c || c == '_' || c == '-')
-  --TODO: probably should disallow starting with a number
+ident = liftM2 (:) first rest
```
23Skidoo: Looks like `ident` is now identical to `dashlessIdent`. Is this intentional?

phadej: My bad, `ident` should have dash. Will fix shortly.

phadej:
@23Skidoo does it look ok now?

23Skidoo:
It does! Merging, thanks!

PR #2864, opened 2015-10-11
Migrate integration tests to shell based format

BardurArantsson:
This is an implementation of #2797.

Let me know if further description/explanation is necessary and I'll add it to the commit message(s).


BardurArantsson:
@grayjay Could you please try this out on Windows and let me know what is still missing relative to fully working on Windows?

(I think I fixed all the test cases + line ending things, but I could easily have missed something.)

BardurArantsson:
Hm... that's an interesting failure. What I'm doing is copying a two-level directory hierarchy from the source directory into a temporary folder and running a shell script in the 2nd level directory. Apparently "source ../foo.sh" isn't picking up the "foo.sh" that *should* have been copied into the 1st level directory.

Anyone have any ideas what's going on with Travis here?

(I would have guessed a CWD issue could have caused this, but I explicitly set the CWD when running the shell script that does the "source" thing.)

BardurArantsson:
Oh, wait, maybe I'm setting CWD using a *relative* path. I'll check that tomorrow.

cabal-install/tests/IntegrationTests/exec/should_run/configures_cabal_to_use_sandbox.sh
```diff
@@ -0,0 +1,14 @@
...
+
+cabal sandbox init > /dev/null
+cabal install > /dev/null
+
+# The library should not be available outside the sandbox
+$GHC_PKG list | grep -v "my-0.1"
+
+# When run inside 'cabal-exec' the 'sandbox hc-pkg list' sub-command
+# should find the library.
+cabal exec sh -- -c "cd subdir && $CABAL sandbox hc-pkg list" | grep "my-0.1"
```
grayjay: `CABAL` contains backslashes on Windows, so I had to quote it here.  The quoted path still doesn't work with Cygwin, though.

grayjay:
@BardurArantsson This branch works well on Windows, except for three things:

* The path to sh is hard-coded as `/bin/sh`.
* The variables `CABAL` and `GHC_PKG` refer to Windows-style paths.  I tested with `sh` from MinGW and Cygwin this time and found that only MinGW can handle the paths.
* (See the comment in the code.)

BardurArantsson:
Thanks.

I'm not sure what we can do about "/bin/sh": That's actually close to being the *only* guaranteed way to get a shell on a POSIX(ish) system. The only other one I know of is to use "env". Does "/usr/bin/env sh" work for you?

(I suppose we could do a lookup for /bin/sh first and try plain "sh" if "/bin/sh" isn't found.)

I don't think supporting Cygwin is necessarily a goal since (I believe) msys is the "primary" Windows-compatibility system... @ezyang: Do you concur?

Will fix the quoting issue.

grayjay:
@BardurArantsson  "/usr/bin/env sh" doesn't work because the path is different on Windows  ("C:/Program Files/Git/usr/bin/env" on my computer).  Trying "/bin/sh" and then "sh" sounds like a good solution.

23Skidoo:
Yes, lack of Cygwin support is not a blocker. Only supporting msys2 is fine. But we must get this to work on Travis before merging.

23Skidoo:
Closing in favour of #2868, which subsumes this PR.

Issue #2711, opened 2015-07-07
package-tests do not work with stack


ezyang:
Steps to reproduce: run `stack test` in Cabal.

Actual result:

```
Cabal-1.23.0.0: test (suite: package-tests)
package-tests: dist: canonicalizePath: does not exist (No such file or directory)
```

OK, so let's set `CABAL_BUILDDIR` explicitly, like ` env CABAL_BUILDDIR=.stack-work/dist/x86_64-linux/Cabal-1.22.2.0/ stack test`. Then we get:

```
    expected: 'setup install' should succeed
      output: "/home/hs01/ezyang/cabal-edge/Cabal/tests/Setup configure --user -w /bin/ghc" in 
PackageTests/BuildDeps/InternalLibrary2/to-install
    Configuring InternalLibrary2-0.1...
    Setup: Use of GHC's environment variable GHC_PACKAGE_PATH is incompatible with
    Cabal. Use the flag --package-db to specify a package database (it can be used
    multiple times).
```

23Skidoo:
Patches welcome.

ezyang:
An adequate workaround is to call the test executable directly with `CABAL_BUILDDIR`, e.g. `env CABAL_BUILDDIR=.stack-work/dist/x86_64-linux/Cabal-1.22.2.0/ .stack-work/dist/x86_64-linux/Cabal-1.22.2.0/build/package-tests/package-tests`

ezyang:
I have a better idea, which is to use the executable name of the package tests to infer the build directory.

ttuegel:
@ezyang You can work around this by running `stack exec --no-ghc-package-path stack test`.

PR #2868, opened 2015-10-13
Migrate integration tests to shell based format, fixed travis

phadej:
Improves on https://github.com/haskell/cabal/pull/2864


phadej:
ping @BardurArantsson @23Skidoo

23Skidoo:
@phadej Does it also work on Windows (with msys2)?

/cc @grayjay

BardurArantsson:
I'm 99% percent sure it won't run on Windows, I believe a previous iteration did this exact thing.

@phadej Thanks for the fixes. I'll rework the "/bin/bash" one (see comment there) and see if I can improve on the "extra-source-files" bit sometime during the next week or so (again, see comment there).

phadej:
ie10 - win8

The only test failing on windows.

phadej:
I'm not sure where to look for the bug. Is it `Cabal` or test setup. @BardurArantsson do you have any ideas?

phadej:
For extra-source-files, IIRC recent enough `Cabal` supports globs? can we use them in `Cabal` itself?

phadej:
And if someone wants to test on windows (now or later): https://dev.modern.ie/tools/vms is handy (minghc works nice too).

phadej:
And last comments:
- for OSX you might need to `cabal update`with just builded `cabal` once to get `00-index.tar` into the right place (not the place Haskell Platform puts it) - alternatively tell where HP placed it to configure script. This is very minor, won't do anything to fix this.
- `stack test` doesn't work still (https://github.com/haskell/cabal/issues/2711):

```
 cabal: Use of GHC's environment variable GHC_PACKAGE_PATH is incompatible with
Cabal.
```

this LGTM otherwise. The `extra-source-files` cleanup could be done later.

I need this PR in to get going with parsec parser (there is something weird happening in my branch, and this version of integration tests is much much more nicer to debug :+1:).

BardurArantsson:
Yeah, I agree that the glob vs. extra-sources thing *isn't* a show-stopper, it can be fixed in a diferent PR. (I'll create the issue after this PR lands. If I forget, please somebody remind me.)

I'm not sure what to make of #2711, but given that it's a previously existing issue and I don't *think* this necessarily changes anything much about a possible resolution, I think we should just punt on fixing #2711 with this PR -- and get this merged ASAP so everybody else who wants to use EasyTests(TM) becomes unblocked.

I'll go squash into a series of meaningful commits which build individually a little later tonight (or tomorrow). At least, I think it would be best to have that, I don't know if there's a project policy for this in Cabal/cabal-install land?

@phadej : I'm not sure if I can preserve authorship of all your patches if we want each individual commit to build -- do you mind if I just pull the "source" -> "." and "extra-sources" ones into the main commit here? (I think the Windows-compat bits can be separate, but if you don't mind I'd also be happy to just pull them in.)

23Skidoo:
Globs of form `foo/*.bar` are supported, but recursive globs are not (yet).

BardurArantsson:
Re: the glob thing. I think the only downside here that it's yet-another-surprise for when building on travis suddenly fails rather inexplicably, but perhaps we should just have a FAQ-type thing? (Or maybe a "checklist-before-creating-a-PR"?). Obviously, it would be ideal to just have an "extra-source-dir" or full recursive glob support, but until then this really only affects the PR phase of development. During normal running-of-tests, it's completely irrelevant.

cabal-install/tests/IntegrationTests.hs
```diff
@@ -0,0 +1,276 @@
+-- | Groups black-box tests of cabal-install and configures them to test
```
23Skidoo: I wonder if the `tasty-golden` package can be used to simplify code in this file.

phadej:
@BardurArantsson, I'm not particularly attached to those commits, there are quite mechanical anyway. Actually I'll just squash everything myself.

cabal-install/tests/IntegrationTests.hs
BardurArantsson: I did check out `tasty-golden`, but it doesn't really seem to get us much since all the traversal stuff it pretty specialized anyway. Did you have something particular in mind?

BardurArantsson:
@phadej Great, thanks. AFAICT this is good to go from my perspective.

cabal-install/tests/IntegrationTests.hs
23Skidoo: Nope, it just looked like `tasty-golden` does something similar. I haven't read the code closely.

cabal-install/tests/IntegrationTests/exec/should_run/exit_with_failure_without_args.sh
```diff
@@ -0,0 +1,6 @@
+. ../common.sh
+
+# We should probably be using a .err file and should_fail,
```
23Skidoo: One possible future solution for these kinds of problems is to add support for platform-specific `.err` files.

23Skidoo: Or we could add support for matching against a set of regexes.

cabal-install/cabal-install.cabal
```diff
@@ -27,7 +27,35 @@ Extra-Source-Files:
   README.md bash-completion/cabal bootstrap.sh changelog
 
   -- Generated with '../Cabal/misc/gen-extra-source-files.sh | sort'
-  tests/PackageTests/Freeze/my.cabal
+  tests/IntegrationTests/exec/common.sh
+  tests/IntegrationTests/exec/should_run/Foo.hs
+  tests/IntegrationTests/exec/should_run/My.hs
+  tests/IntegrationTests/exec/should_run/adds_sandbox_bin_directory_to_path.sh
```
23Skidoo: The "prints error message if given no argument" test for `exec` is missing, is this due to the Windows `.exe` problem?

cabal-install/tests/IntegrationTests/exec/should_run/configures_cabal_to_use_sandbox.sh
```diff
@@ -0,0 +1,14 @@
+. ../common.sh
+
+cabal sandbox delete > /dev/null
```
23Skidoo: These `> /dev/null` redirections are kinda ugly, can we get rid of them at some point?

cabal-install/tests/IntegrationTests/exec/should_run/exit_with_failure_without_args.sh
23Skidoo: Also see how LLVM solves this: http://llvm.org/docs/CommandGuide/FileCheck.html

23Skidoo:
OK, I think this is good to go. 

@phadej @BardurArantsson Thanks for your contributions!

cabal-install/cabal-install.cabal
BardurArantsson: Yes. This is also part of the reason for some of the > /dev/null bits. (I may have forgotten to remove some that weren't strictly necessary.).

The problem is the you get the "configuring blahblah..." message which contains the executable name and if you're trying to compare the contents of stdout with a file...

(There was a little back and forth on this on the original branch. I decided to keep the *option* of having .out and .err files, but in some cases I had to replace them with a "cabal ... | grep ..." to avoid the .exe problem.)

cabal-install/tests/IntegrationTests/exec/should_run/exit_with_failure_without_args.sh
BardurArantsson: Yeah, I considered line-by-line regex comparison instead, but it's a bit noisy if you have to escape all the lines containing special characters (and where you *don't* want special treatment). FileCheck looks somewhat similar to that, but I think it's probably overkill for what we want.

Maybe we could add something simple here like line-based comparison, but with

```
   RE:this is a regex
   this is not
```

i.e. just using the a RE: prefix to designate regexes. (It's ulikely enough to appear in actual output that it wouldn't need to be "escapable".)

BardurArantsson:
... and thanks to @ezyang and @grayjay for early review, testing and feedback!

cabal-install/tests/IntegrationTests/exec/should_run/exit_with_failure_without_args.sh
23Skidoo: The `RE:` prefix idea sounds good.

cabal-install/cabal-install.cabal
BardurArantsson: Oh, I should say: The test *is* still in here in one of the *.sh files, but just with a "cabal ... | grep ..." instead of direct-compare with an output file.

cabal-install/tests/IntegrationTests/exec/should_run/exit_with_failure_without_args.sh
BardurArantsson: I'll create an issue to add it and assign it to myself. (No guarantees on when I'll get 'round to it, though.)

PR #2869, opened 2015-10-15
Backwards compatibility type synonym for InstalledPackageId.

ezyang:


23Skidoo:
Merged, thanks!

PR #2877, opened 2015-10-17
Manual page generation

### LINKED ISSUE #848: cabal-install needs a manpage
(Imported from [Trac #858](http://hackage.haskell.org/trac/hackage/ticket/858), reported by @nomeata on 2011-06-19)

The cabal binary shipped by cabal-install ought to be documented with a manage. Ideally, cabal could generate one by itself from the Command declarations, similar to how darcs does it, this way there is no duplication of documentation.

mmakowski:
Manual page generation from command list defined in `Main`. Based on the approach and implementation in darcs.

Fixes #848.

### DIFF
--- a/cabal-install/Distribution/Client/Manpage.hs
+++ b/cabal-install/Distribution/Client/Manpage.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, ExistentialQuantification #-}
+{-# LANGUAGE CPP #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Distribution.Client.Manpage
@@ -7,17 +7,13 @@
 --
 -- Maintainer  :  cabal-devel@haskell.org
 -- Stability   :  provisional
--- Portability :  non-portable (ExistentialQuantification)
+-- Portability :  portable
 --
--- Functions and types for building the manual page.
+-- Functions for building the manual page.
 
 module Distribution.Client.Manpage
-  ( -- * Command list
-    CommandVisibility (..)
-  , CommandSpec (..)
-  , commandFromSpec
-    -- * Manual page generation
-  , manpage
+  ( -- * Manual page generation
+    manpage
   ) where
 
 import Distribution.Simple.Command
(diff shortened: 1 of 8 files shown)


23Skidoo:
Thanks! Looks good on a first glance.

cabal-install/Distribution/Client/Manpage.hs
```diff
@@ -0,0 +1,187 @@
...
+  , manpage
+  ) where
+
+import Distribution.Simple.Command
+import Distribution.Client.Setup (globalCommand)
+
+import Data.Char (toUpper)
+import Data.List (intercalate)
+
+data CommandVisibility = Visible | Hidden
```
23Skidoo: Looks like this duplicates the functionality of `CommandType` in `D.S.Command`.

cabal-install/Distribution/Client/Manpage.hs
```diff
@@ -0,0 +1,187 @@
...
+
+-- | wraps a @CommandUI@ together with a function that turns it into a @Command@.
+-- By hiding the type of flags for the UI allows construction of a list of all UIs at the
+-- top level of the program. That list can then be used for generation of manual page
+-- as well as for executing the selected command.
+data CommandSpec action
```
23Skidoo: This looks like something that belongs better in `D.S.Command`. Consider refactoring that module.

cabal-install/Distribution/Client/Manpage.hs
```diff
@@ -0,0 +1,187 @@
...
+  = forall flags. CommandSpec (CommandUI flags) (CommandUI flags -> Command action) CommandVisibility
+
+commandFromSpec :: CommandSpec a -> Command a
```
23Skidoo: Ditto.

23Skidoo:
LGTM but I'd like to see some parts moved to `D.S.Command`.

23Skidoo:
A potential future improvement may be to add support for generating per-command manpages (e.g. `cabal manpage install`). Then we could make `cabal help` display a manpage when run from the terminal, a-la `git help`.

23Skidoo:
FYI, there are also some formatting problems (but I don't see them as impediment for merging):

formatting_problem

mmakowski:
@23Skidoo, thanks for the review. I will try to address the comments over the next couple of days.

23Skidoo:
> Regarding `cabal manpage install`, I think that woud require making `manpage` a library-level special command similar to `help`.

I am not so sure. If you can generate a manpage with information about all commands from `cabal-install`, it should also be possible to do the same for a single command (maybe after some refactoring of `D.S.Command`).

Code for help formatting and option handling lives in the Cabal lib because we want to support things like `./Setup foo --help`.

> Another thing to consider with help displaying manual pages is that man is not availabe on Windows.

In these cases we can just revert to the old mode of output.

mmakowski:
I have now moved command-related bits to `D.S.Command` and fixed formating of `sandbox` sub-commands.

23Skidoo:
LGTM now.

23Skidoo:
Merged, thanks!

PR #2881, opened 2015-10-22
Refactoring and two Doc/Wording changes

lspitzner:
regarding the changes of wording:

1. a "no-op" would not call `cont` in the Nothing case.
2. I think saying

> cabal-install version 1.22.6.0
> __compiled__ using version 1.22.4.0 of the Cabal library 

is better in order to avoid the whole "what, why is ghc-pkg saying something different" confusion.


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #2890, opened 2015-10-26
Implement `cabal upload --doc`

bennofs:
This is the second part that was still remaining for the implementation of #2080. I still need to test this on powershell though, I'll report back when I've done that.

### DIFF
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -49,7 +49,7 @@ import Distribution.Simple.PreProcess
 import Distribution.Simple.Setup
          ( defaultHscolourFlags
          , Flag(..), toFlag, flagToMaybe, flagToList, fromFlag
-         , HaddockFlags(..), HscolourFlags(..) )
+         , fromFlagOrDefault, HaddockFlags(..), HscolourFlags(..) )
 import Distribution.Simple.Build (initialBuildSteps)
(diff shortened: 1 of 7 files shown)


23Skidoo:
Thanks for working on this! Can you look into what's causing the Travis failure?

bennofs:
Ok, the upload works on windows, but cabal still reports an error (this is related to the powershell code, which doesn't deal with non-empty response bodies well). There is also some misbehaviour (reporting failure while the upload actually worked) when using curl. I'll look into those two issues, after that the PR should be ready to merge.

About the travis build, it seems that there is no longer any failure? Was it restarted?

bennofs:
Fixed the remaining errors on Windows and the curl issue. Should be good to merge now.

bennofs:
ping @23Skidoo (hope pinging is OK)

23Skidoo:
Sorry for the delay. Will take a look now.

23Skidoo:
LGTM except that small security issue. Also would be nice to see a changelog entry. Thanks again for working on this!

bennofs:
@23Skidoo changelog entry added, security issues fixed. I'm not quite sure here about the powershell situation regarding password security: the password is in cleartext in the generated script, so anyone who can read the script also has access to the password. Are the temporary files generated by cabal only readable by the current user?

23Skidoo:
Merged, thanks!

PR #2892, opened 2015-10-27
CopyFile.hs: use withBinaryFile instead of bracket

Heather:


23Skidoo:
Merged, thanks!

Heather:
sorry, does this break build?

23Skidoo:
Yes, but I already fixed that.

PR #2896, opened 2015-10-27
why DeriveGeneric is there?

Heather:
sorry for commit message - I'm not sure whether it's used here but for now I can't find where


23Skidoo:
Merged, thanks!

PR #2898, opened 2015-10-27
Add --ipid-extra flag.

ezyang:
This flag uses the default IPID generation algorithm, but
permits users to add some extra information which can be used
to ensure that the IPID is unique.


ezyang:
Actually, I don't think I need this.

PR #2904, opened 2015-11-01
CPP-reducing refactoring

hvr:
This refactoring allows to drop `{-# LANGUAGE CPP #-}` from a couple of
modules again. While at it, this also canonicalises the `Applicative`/`Monad` instances, 
thereby future proofing those.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -263,7 +263,9 @@ library
       Distribution.Compat.Binary.Generic
 
   default-language: Haskell98
-  default-extensions: CPP
+  -- starting with GHC 7.0, rely on {-# LANGUAGE CPP #-} instead
+  if !impl(ghc >= 7.0)
+    default-extensions: CPP
 
 -- Small, fast running tests.
 test-suite unit-tests
--- a/Cabal/Distribution/Compat/Binary/Class.hs
+++ b/Cabal/Distribution/Compat/Binary/Class.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, FlexibleContexts #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE Trustworthy #-}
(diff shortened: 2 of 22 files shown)


Cabal/Cabal.cabal
```diff
@@ -263,7 +263,9 @@ library
       Distribution.Compat.Binary.Generic
 
   default-language: Haskell98
-  default-extensions: CPP
+  -- starting with GHC 7.0, rely on {-# LANGUAGE CPP #-} instead
+  if !impl(ghc >= 7.0)
```
23Skidoo: Doesn't `impl(ghc < 7.0)` work?

hvr: The main reason I write `!impl(ghc >= 7.0)` is that I'm not sure what the semantics are if the `impl(ghc)` is false. If in fact those two variants are equivalent, there's no reason to use the `!`-version :-)

23Skidoo: [Looks like you're right](https://github.com/haskell/cabal/blob/master/Cabal/Distribution/PackageDescription/Configuration.hs#L112), they're different. `impl(ghc < 7)` is basically a shorthand for `impl(ghc) && impl(ghc < 7)`.

Hugs is no longer supported by Cabal, though, and I don't think we expect anything except GHC/GHCJS to be able to compile Cabal.

23Skidoo:
Merged, thanks. Some bits (e.g. `Mon.mempty`) are slightly ugly, but I guess getting rid of CPP is worth it.

hvr:
Yeah, it's not perfect. I noticed CPP breaks automatic refactoring tools quite a lot, as CPP conditionals are not represented in the AST. I mostly bothered about the cases where CPP was trivially used for warning suppression.

There's also the variant of defining a kind of internal common module that gets imported by all modules, and provides a common set of verbs/identifiers used by all modules. It's not uncommon for larger code-bases, as this allows to cut down `import`-lines quite a bit.

PR #2905, opened 2015-11-01
	Bump lower bounds of dependencies 

phadej:


23Skidoo:
7.2.2 was released more than three years ago; per [our policy](https://github.com/haskell/cabal/blob/master/HACKING.md#dependencies-policy) Cabal no longer has to be buildable by it.

phadej:
@23Skidoo this was inspired by IRC conversation:
```
17:45  hvr dcoutts: I was wondering what the oldest GHC was you expect 'Cabal' to compile against
17:47  hvr right now the lower bound is base>=4.4
17:47  dcoutts is it
17:47  hvr i.e. GHC >=7.2
...
18:37  hvr tbh, I'd drop support for GHC 7.2
18:37  hvr GHC 7.2 has some annoying broken corners in the install-plan space
18:37  hvr due to an incompelte SafeHaskell
```

If travis says *red*, then I'll bump base lower bound

hvr:
To quote `#hackage` a bit more, 

```
16:45 < hvr> dcoutts: I was wondering what the oldest GHC was you expect 'Cabal' to compile against
16:46 < dcoutts> hvr: I'm not sure we have a hard and fast rule
16:46 < dcoutts> i prefer a bit more than some people do for Cabal, since there are bootstrapping issues
16:47 < dcoutts> it's handy for people with old ghc to install a new cabal-install
16:47 < dcoutts> hvr: but I think we can certainly require bsae 4
16:47 < dcoutts> base
16:47 < dcoutts> so can use new exceptions etc
16:47 < hvr> right now the lower bound is base>=4.4
```

and I guess we can easily support GHC>=7.4 for quite some time to come, as iirc there's no major feature that was added after GHC 7.4. IMHO GHC 7.10 is the next big changing releases, where it makes sense to start dropping support for pre GHC 7.10 versions.

phadej:
Build is green. ping @hvr @23Skidoo

cabal-install/cabal-install.cabal
```diff
@@ -160,22 +160,22 @@ executable cabal
     -- NOTE: when updating build-depends, don't forget to update version regexps
     -- in bootstrap.sh.
     build-depends:
-        array      >= 0.1      && < 0.6,
-        base       >= 4.3      && < 5,
-        bytestring >= 0.9      && < 1,
         Cabal      >= 1.23     && < 1.24,
```
23Skidoo: Can we please keep the deps sorted alphabetically like before?

23Skidoo:
Yes, I think we can continue supporting 7.4 for a while.

phadej:
@23Skidoo done.

23Skidoo:
Merged, thanks!

PR #2916, opened 2015-11-08
Fix space leak in solver backjumping

grayjay:
This commit refactors backjumping so that it uses the `Progress` type instead of
separate references to a node's children and the conflict set calculated from
those children.  The behavior should be unchanged.

I added to #2914 so that I could test memory usage more easily.  These heap profiles were generated with `cabal --ignore-sandbox install --dry-run --max-backjumps -1 hackage-server-0.4 aeson yesod --with-compiler C:/ghc-7.6.3_64/bin/ghc +RTS -p -s -h -L50`.

after #2914 (https://github.com/haskell/cabal/commit/90c777299251deda571083ec0b95b68ca20e0873):
cabal-refactored-logging

after both commits (https://github.com/haskell/cabal/commit/c1189c3d61585207e99818531e3cd990d4babf3a):
cabal

longer run with both commits, using ` -i10`.  I don't know why it increases later.
cabal-longer-run

I have a few concerns about this PR.  I combined backjumping and exploring the tree into one traversal, which is less modular.  Is there a way that we could use two passes without storing conflict sets directly on the tree's nodes?  The `combine` function is also less recognizable after the refactoring.  I'd like to make it clearer, if possible.

~~I deleted several unused functions that didn't compile with my change, such as `explore`.  Is that alright, or should I refactor them?~~

Issue #2860, opened 2015-10-11
Attempt to find optimal build plans


jdnavarro:
This includes a weight penalty system to measuer the quality of different build plans.

/cc @kosmikus

kosmikus:
I also talked to @grayjay about this before.

kosmikus:
Plan of doing this incrementally:

1. Modify `PSQ` to take an extra parameter for weights / penalties, which generally should be an instance of `Ord`. Don't use it yet, instantiate it to `()` everywhere.

2. Replace the current reorderings (mainly in `Preference`) to modify weights instead. Try to reimplement the current behaviour precisely. This will require to use lexicographic ordering, i.e., a weight type that is in essence a tuple.

3. Either prior to or during exploration, calculate the total penalty of each node (that is visited). When finding a success node, return the accumulated penalty, so that it can e.g. be printed together with the found solution.

4. Add some functionality for pruning, again either immediately before or during exploration. This can  appear in multiple flavours. For example, we could just allow a command line flag to specify a "maximum" score and prune everything that exceeds that maximum score. But we could also continue after finding the first solution and dynamically adjust the limit, trying to find incrementally better and ultimately optimal solutions.

5. Tweak the actual penalties used, adjusting the preference preference phase so that we get what feels really like optimal build plans. This is mostly heuristics and experience.

grayjay:
Should we compare solutions found with different goal orderings?  The solutions might vary more than those found at one end of a large tree.

kosmikus:
@grayjay Goal reordering is somewhat independent. It can significantly speed up the overall search. It can also change the score of the "first few" solutions, although it will obviously not change the score of the optimal solution. Implementing the machinery to do the scoring is necessary in any case.

grayjay:
@kosmikus I've started working on this.  Here is my design, so far:

* I created a type `newtype WeightedPSQ w k v = WeightedPSQ [(w, k, v))]` with the invariant that the list is always sorted by weight.  All choice nodes use `WeightedPSQ`, except for goal choices.  `GoalChoice` still uses `PSQ` because goal order shouldn't affect the score of the install plan.

[...]

PR #2917, opened 2015-11-09
[RFC] Add option to find best install plan before backjump limit

grayjay:
I've started implementing install plan scoring, as described in #2860.  `cabal install` now prints the score of each install plan and allows the user to find better plans with the option `--max-score`, where zero is perfect.  `cabal` should find the same install plans as before when no maximum is specified.  This pull request isn't ready to be merged yet.  It needs documentation and tests, and I have a few questions in TODO comments.  Also, this is the most basic way to expose install plan scoring, and it might be better to automate determining when to continue searching before this is merged.

I rebased onto #2916 because this change made the existing space leak worse.  Then I compared performance between #2916 and the latest commit.  Memory usage is similar, but the latest commit is noticeably slower when the solver runs for longer.  I compiled cabal with GHC 7.10.2 and ran `cabal --ignore-sandbox install --dry-run --max-backjumps -1` with the following packages and compilers:

| Package | Compiler | Measurement | https://github.com/haskell/cabal/commit/c1189c3d61585207e99818531e3cd990d4babf3a (#2916) | https://github.com/haskell/cabal/commit/868909732813b584aa15c3ccd62e8469063f9122 (#2917) |
| --- | --- | --- | --- | --- |
| aeson | GHC 7.10.2 | time | 0.884 s | 0.876 s |
| hackage-server-0.4 | GHC 7.10.2 | time | 6.01 s | 6.66 s |
| hackage-server-0.4 | GHC 7.6.3 | time | 122.6 s | 140.5 s |
| hackage-server-0.4 | GHC 7.6.3 | total memory (`+RTS -s`) | 256MB | 258MB |

These comments that I made on the issue are still relevant:
https://github.com/haskell/cabal/issues/2860#issuecomment-149114193
https://github.com/haskell/cabal/issues/2860#issuecomment-151069126


grayjay:
I rebased this to deal with conflicts and made a few minor changes that shouldn't affect the behavior.

23Skidoo:
@kosmikus, would be nice if you could provide @grayjay with some feedback on this.

BardurArantsson:
@grayjay Note that there are now new conflicts, probably because my dead code removal patch was merged.

grayjay:
I resolved the conflicts.  The dead code removal actually shortened the diff by about 80 lines.

grayjay:
I tried exposing install plan scoring in a more user-friendly way.  The latest commit adds a new option, `--find-best-solution`.  The solver searches until the backjump limit and then uses the best-scoring install plan found.  If there is no backjump limit, the search is exhaustive.  The only benefit of `--max-score` now is that it can allow the solver to run faster by pruning low-scoring nodes during the whole run.  `--find-best-solution` works by decreasing the maximum allowed score as the solver finds better and better solutions.

There are a few things I'd like to improve if this feature is added to cabal:
* The solver is much slower than on master.  This branch ran for 43% longer in one long-running test.
* On master, `explore` creates a log from the search tree, and then `logToProgress` processes the log and applies the backjump limit.  I think it would be easier to implement the interaction between `--max-backjumps` and `--find-best-solution` if I first moved application of the backjump limit into `explore`.  That would take significant refactoring, though.  This branch currently just stores the best install plans in the log entries, which is awkward.
* I think `--max-score` is unnecessary if we have `--find-best-solution`, but I left it in for testing.

kosmikus:
@grayjay Thanks for rebasing this. I'm hoping that I can finally take a look at this soon.

grayjay:
@kosmikus Thanks.  I just rebased on top of the module reorganization and added some comments and quickcheck tests.  I also added a commit that does dynamic goal reordering based on conflict sets.  It sometimes finds the best solution faster, especially when combined with `--reorder-goals`.

kosmikus:
@grayjay Thanks! I had just been experimenting with the same/a similar thing. My "dynamic-goal-reordering" branch is at https://github.com/kosmikus/cabal/tree/count-conflicts

My initial experiments with this have been very promising. There are only very few cases where it seems to be slower, and many where it seems to perform vastly better. So I'd actually be optimistic that something like this can become the new default.

I will have a look at how our approaches differ, if at all.

grayjay:
@kosmikus Cool!  I'm interested in looking at your branch when I have time.

I added `--dynamic-goal-reordering` a while ago without pushing it, and I just realized that I never finished it.  I wanted the solver to try choosing goals that were in the last conflict set, because they are likely to also fail in the next subtree.  My code is using the wrong conflict set, though.  Sorry about the mix-up.  The first two commits should be ready for review.

kosmikus:
@grayjay Right, this corresponds to what I'm doing, only that I'm actually not just considering the "last" conflict set, but counting how often each goal was involved in conflicts and use that for sorting.

grayjay:
@kosmikus I fixed the bug in `--dynamic-goal-reordering`.  Now it actually seems like an improvement!  I haven't tested it thoroughly, though.

I wanted to do something more complex, but I haven't implemented it.  The solver would prefer goals whose variables are contained in conflict sets at any level, but prefer variables at lower levels over variables at higher levels.  For example, when the solver is at node 4, it would first look for goals in the conflict set at node 3, then node 2, and finally node 1. (I left out the goal nodes and other children.)  This PR only uses the conflict set at node 3.

tree-2

The reason that I used conflict sets from non-`Fail` nodes is that I thought it might deal with the case where the solver spends a long time solving for a few goals, only to hit an unrelated, unresolvable conflict farther down in the tree. If that pattern repeats in siblings' subtrees, the solver could try the goals involved in the unresolvable conflict first and give up faster on those subtrees.

I tried your branch on a few packages that my PR couldn't handle, at least without `--reorder-goals`, and it was fast!  It seems like a good default, so far.  I'm curious to see how it interacts with `--find-best-solution`.

kosmikus:
@grayjay FYI: I have pushed https://github.com/kosmikus/cabal/tree/count-conflicts-2 that is a version of `count-conflicts` on top of current master.

grayjay:
@kosmikus I'm getting a compiler error on the new branch:

```
Distribution/Solver/Modular/Explore.hs:122:35:
    Couldn't match type ‘Goal QPN’ with ‘OpenGoal ()’
    Expected type: P.PSQ
                     (OpenGoal ())
                     (Assignment
                      -> ConflictMap
                      -> (ConflictSetLog (Assignment, RevDepMap), ConflictMap))
      Actual type: P.PSQ
                     (Goal QPN)
                     (Assignment
                      -> ConflictMap
                      -> (ConflictSetLog (Assignment, RevDepMap), ConflictMap))
    In the second argument of ‘getBestGoal’, namely ‘ts’
    In the expression: getBestGoal cm ts
```

kosmikus:
@grayjay Sorry, I made a mistake when pushing. I force-updated the branch.

grayjay:
@kosmikus Thanks.  I tried it on a few more packages, and it seems to work really well.  I'll look at the code when I have more time.

grayjay:
@kosmikus I started rebasing this onto master, because I think it will be easier to split it apart when it is up to date.

grayjay:
I'm done rebasing.

grayjay:
I rebased onto #3594.

grayjay:
This PR needs more work.

PR #2918, opened 2015-11-10
Make `cabal init` reject extra arguments.

ben0x539:
This is better than ignoring them because one is likely in some state of
confusion when passing extraneous arguments to `cabal init` and would
probably appreciate being taken out of that state sooner rather than
later.


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #2923, opened 2015-11-13
Disable Ar.wipeMetadata for AIX

hvr:
AIX uses its own `ar` format variant (with a different header),
which causes Cabal to abort with

    Distribution.Simple.Program.Ar.wipeMetadata: Bad global header

This patch simply disables this non-essential operation for AIX.


23Skidoo:
Merged, thanks!

PR #2927, opened 2015-11-16
Use `-fno-warn-implicit-prelude' in Paths_* module

trskop:
When GHC is invoked with `-fwarn-implicit-prelude', then autogenerated
Paths_* modules are source of unnecessary warnings.

This commit adds OPTIONS_GHC pragma with `-fno-warn-implicit-prelude' in
to Paths_* module, so that such warnings are suppressed.


23Skidoo:
Merged, thanks!

PR #2930, opened 2015-11-17
Conditionally omit `TemplateHaskell` from set of supported extensions

hvr:
This emulates what GHC>=8 will do on its own for older GHC versions.

This is particularly useful in combination with #2644 allowing the
solver to toggle flags depending on the availability of
`-XTemplateHaskell`.

For instance, cross-compilers and/or unregisterised GHC builds often
don't have TemplateHaskell support. Having support for toggling flags
based on availability of `-XTemplateHaskell` allows `cabal` to
support such environments with less manual intervention.

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE PatternGuards #-}
+
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Distribution.Simple.GHC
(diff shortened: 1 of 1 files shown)


hvr:
See also https://phabricator.haskell.org/D1484

/cc @bgamari

phadej:
Would be nice to have a package test for that...

23Skidoo:
LGTM. 

@phadej 
Coming up with a test for this may be hard because you need a non-standard GHC.

phadej:
@23Skidoo Ah indeed. Pity. Have to live without test then.

PR #2931, opened 2015-11-20
Fix for #2155 including tests

martinvlk:
Here is my take on fixing https://github.com/haskell/cabal/issues/2155 (cabal sandbox delete-source gives no indication of whether it did anything)

I included integration tests that capture the problem and the solution.

I also did a bit of refactoring along the way and added corresponding unit tests.

Could somebody review my code as I am a Haskell beginner and would appreciate feedback and making sure it's sound before it's potentially accepted into the project.

Cheers
Martin


martinvlk:
BTW, I am handling two different use cases:
1) when the source being deleted is not an existing directory
2) when the source being deleted exists on disk, but was not registered with the sandbox prior to removing

In both cases cabal gives nonzero exit status, and it outputs corresponding error message.

martinvlk:
Ah, missing import for older GHC versions. I'll fix that.

Martin

23Skidoo:
Thanks, will review soon.

martinvlk:
I just found an old [pull request 2605](https://github.com/haskell/cabal/pull/2605) in the queue, that also attempts to fix #2155, but hasn't been merged since May. So we have two conflicting pull requests for that issue.

The fix in the pull request 2605 does work as far as I can see, but is done quite differently than my fix. It has no tests going with it.

One other thing I noticed is that the pull request 2605 mentions issue #1360, and I can see that one of the use cases I handle here goes against that fix. I am going to remove it so and tolerate nonexistent source directory as per #1360.

23Skidoo:
#2155 also mentions that the exit code is wrong, which still seems to be the case with your patch.

cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -453,12 +453,19 @@ sandboxDeleteSource verbosity buildTreeRefs _sandboxFlags globalFlags = do
   indexFile            <- tryGetIndexFilePath (pkgEnvSavedConfig pkgEnv)
 
   withRemoveTimestamps sandboxDir $ do
-    Index.removeBuildTreeRefs verbosity indexFile buildTreeRefs
+    removedRefs <- Index.removeBuildTreeRefs verbosity indexFile buildTreeRefs
+    when (length buildTreeRefs > length removedRefs) $
+      die $ "Skipping nonregistered sources: " ++ (showL $ prune buildTreeRefs removedRefs)
```
23Skidoo: I think this will cause corruption of the timestamps file when some refs were removed and some were not.

cabal-install/Distribution/Client/Tar.hs
```diff
@@ -453,12 +461,28 @@ mapEntries :: (Entry -> Entry) -> Entries -> Entries
 mapEntries f = foldrEntries (Next . f) Done Fail
 
 filterEntries :: (Entry -> Bool) -> Entries -> Entries
-filterEntries p =
-  foldrEntries
-    (\entry rest -> if p entry
-                      then Next entry rest
-                      else rest)
-    Done Fail
+filterEntries p = isoR . filterEntriesW (return . p)
```
23Skidoo: Do `filterEntriesW` and `foldrEntriesW` have to use `WriterT` or can we generalise them to an arbitrary `Monad`? I don't see anything `WriterT`-specific in the implementation.

23Skidoo:
LGTM for the most part. Will merge once the comments are addressed.

martinvlk:
By the way, I wrote more tests and found that #1360 has regressed since it was fixed in 2013.

I think though that now it is caused by a different function then before. It now fails when updating the cache (Distribution.Client.Sandbox/Index.removeBuildTreeRefs, when calling the updatePackageIndexCacheFile function) rather than when canonicalizing the paths.

Not sure whether to fix this in this PR or do a new one..

23Skidoo:
Looks almost ready to go in now, thanks. See my comments above.

martinvlk:
> Can't you just call (p $ map snd srcRefs) here instead of doing fmap snd in p?

No, the point of p is that is observes the filtering of entries using a Writer (see the "tell" call?) and at the end I know what elements were filtered out. So I can't get rid of p.

23Skidoo:
I'm not asking to get rid of `p`, I'm just saying that you didn't need to change the type of `p` in that patch. Instead of passing `srcRefs` to `p` you can pass `map snd srcRefs` to it.

martinvlk:
Ok, I see what you mean now.. your suggestion is better because p in principle does not need to know anything about the other value from the tuple... fair enough!

martinvlk:
> removedPaths <- withRemoveTimestamps sandboxDir $ removeBuildTreeRefs

This was the way it was originally written. Unfortunately if we want to die on removal of non-registered refs then it can't work that way without excessively large refactoring.  

> Or maybe just add a removeTimestamps function and call removeTimestamps removedPaths.

I am going to just change the existing withRemoveTimestamps to do the right thing for me. This is the only place it's used so won't hurt anywhere. The only thing is it then won't follow the same pattern as the other related functions in Timestamp.hs, but that's not too bad I guess.

martinvlk:
Hmm, looks like one of the build targets failed for some reason unrelated to my changes:
https://travis-ci.org/haskell/cabal/jobs/93416013

How to investigate and/or request a new build without me pushing a change?

martinvlk:
Ah well, I'll push the rename of that function and it'll do another integration build then.

23Skidoo:
Merged, thanks!

martinvlk:
Awesome, thanks for your guidance!

PR #2939, opened 2015-11-30
Fix the regression of #1360: Impossible to delete a non-existent source

martinvlk:
Right now this pull request contains a partial fix, the test I added is still not passing.
I am going public with it now so that I can discuss some specifics with the others - details to be added later.

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -58,7 +58,7 @@ import Distribution.Simple.Utils
          ( die, warn, info, fromUTF8, ignoreBOM )
 
 import Data.Char   (isAlphaNum)
-import Data.Maybe  (mapMaybe)
+import Data.Maybe  (mapMaybe, catMaybes)
 import Data.List   (isPrefixOf)
 #if !MIN_VERSION_base(4,8,0)
 import Data.Monoid (Monoid(..))
@@ -75,7 +75,7 @@ import Distribution.Client.Utils ( byteStringToFilePath
                                  , tryFindAddSourcePackageDesc )
 import Distribution.Compat.Exception (catchIO)
 import Distribution.Client.Compat.Time (getFileAge, getModTime)
-import System.Directory (doesFileExist)
+import System.Directory (doesFileExist, doesDirectoryExist)
 import System.FilePath ((</>), takeExtension, splitDirectories, normalise)
 import System.FilePath.Posix as FilePath.Posix
          ( takeFileName )
@@ -245,7 +245,7 @@ typeCodeFromRefType :: BuildTreeRefType -> Tar.TypeCode
 typeCodeFromRefType LinkRef     = Tar.buildTreeRefTypeCode
 typeCodeFromRefType SnapshotRef = Tar.buildTreeSnapshotTypeCode
 
-type MkPackageEntry = IO PackageEntry
+type MkPackageEntry = IO (Maybe PackageEntry)
 
 instance Package PackageEntry where
   packageId (NormalPackage  pkgid _ _ _) = pkgid
@@ -262,7 +262,7 @@ packageDesc (BuildTreeRef _ _ descr _ _) = descr
 data PackageOrDep = Pkg PackageEntry | Dep Dependency
 
 parsePackageIndex :: ByteString
-                  -> [IO PackageOrDep]
+                  -> [IO (Maybe PackageOrDep)]
 parsePackageIndex = accum 0 . Tar.read
   where
     accum blockNo es = case es of
@@ -280,11 +280,11 @@ parsePackageIndex = accum 0 . Tar.read
 
         tryExtractPkg = do
           mkPkgEntry <- maybeToList $ extractPkg entry blockNo
-          return (fmap Pkg mkPkgEntry)
+          return $ fmap (fmap Pkg) mkPkgEntry
 
         tryExtractPrefs = do
           (_,prefs') <- maybeToList $ extractPrefs entry
-          map (return . Dep) $ prefs'
+          fmap (return . Just . Dep) prefs'
 
 extractPkg :: Tar.Entry -> BlockNo -> Maybe MkPackageEntry
 extractPkg entry blockNo = case Tar.entryContent entry of
@@ -292,7 +292,7 @@ extractPkg entry blockNo = case Tar.entryContent entry of
      | takeExtension fileName == ".cabal"
     -> case splitDirectories (normalise fileName) of
         [pkgname,vers,_] -> case simpleParse vers of
-          Just ver -> Just $ return (NormalPackage pkgid descr content blockNo)
+          Just ver -> Just . return $ Just (NormalPackage pkgid descr content blockNo)
             where
               pkgid  = PackageIdentifier (PackageName pkgname) ver
               parsed = parsePackageDescription . ignoreBOM . fromUTF8 . BS.Char8.unpack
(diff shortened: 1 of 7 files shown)


23Skidoo:
Can you please look into the Travis failure?

23Skidoo:
Ah, OK, I see that you're aware of the failure.

cabal-install/Distribution/Client/IndexUtils.hs
```diff
@@ -307,12 +307,15 @@ extractPkg entry blockNo = case Tar.entryContent entry of
   Tar.OtherEntryType typeCode content _
     | Tar.isBuildTreeRefTypeCode typeCode ->
       Just $ do
-        let path   = byteStringToFilePath content
-            err = "Error reading package index."
-        cabalFile <- tryFindAddSourcePackageDesc path err
-        descr     <- PackageDesc.Parse.readPackageDescription normal cabalFile
-        return $ BuildTreeRef (refTypeFromTypeCode typeCode) (packageId descr)
-                              descr path blockNo
+        let path = byteStringToFilePath content
```
23Skidoo: So the problem is that `extractPkg` is called from `parsePackageIndex`, which is called from `updatePackageIndexCacheFile`, which is called from `removeBuildTreeRefs`. But isn't `updatePackageIndexCacheFile` called only after `doRemove` - that is, after the invalid refs were already removed? So why should we fix `extractPkg` then - maybe it just suffices to improve the error message?

martinvlk:
I did more investigation yesterday and I found out that yes, it may be a documentation problem, at least in part, but for a different reason than you suggest.

The thing is that when the directory doesn't exist, then the reference is NOT removed from the index, unless you specify its full path, as recorded in the index. This is because canonicalization of a nonexistent directory in `removeBuildTreeRefs` fails and therefore you never get to reconstruct the full path from a relative one specified on the command line. Also in this scenario the `updatePackageIndexCacheFile` bails out with a "does not exist" for the file and wipes the whole cache file, instead of just removing the bad reference.

The former of the two issues is hard to fix, and the only thing we can possibly do is document this in the user guide. The user has to specify the full path as given by `cabal sandbox list-sources` and then it gets removed successfully.

The problem of wiping the index cache file then is fixed by my commit and I think we should merge it, along with a documentation update. If you agree, I'll put together the doc update and also add tests to capture the behaviour.

23Skidoo:
We can update the manual, but it's more important to make `delete-source` print a helpful warning when it's can't canonicalise the provided path. Something like:

```
$ cabal sandbox delete-source foo
Can't figure out the absolute path 'foo' refers to. If you are trying to delete a reference to a removed directory, please provide the full absolute path (as given by `sandbox list-sources`).
```

Also, `remove-source` should fail even before it gets to updating the cache in this case.

> The latter problem of wiping the index cache file then is fixed by my commit and I think we should merge it

Yes, since `updatePackageIndexCacheFile` is also called from other places.

martinvlk:
Ok, this is now ready for review.

I have streamlined the error handling as we are now dealing with different types of errors within a possibly list of input values. Each type of error will be reported only once for all affected paths and the same goes for success.

23Skidoo:
Can you please rebase this against master to get rid of the spurious merge commits? Also would be nice to see some of the commits squashed together to provide a more logical view of the changes.

cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -452,26 +455,47 @@ sandboxDeleteSource verbosity buildTreeRefs _sandboxFlags globalFlags = do
...
   notice verbosity $ "Note: 'sandbox delete-source' only unregisters the " ++
     "source dependency, but does not remove the package " ++
     "from the sandbox package DB.\n\n" ++
     "Use 'sandbox hc-pkg -- unregister' to do that."
   where
-    convertWith dict pth = fromMaybe pth $ fmap fst $ find ((==pth) . snd) dict
-    showL = concat . intersperse " " . fmap show
+    showPaths f = concat . intersperse " " . fmap (show . f)
+
+    putNewlineStdErr = hPutStr stderr "\n"
```
23Skidoo: Why is this needed? Can't you just use `warn`?

cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -452,26 +455,47 @@ sandboxDeleteSource verbosity buildTreeRefs _sandboxFlags globalFlags = do
...
+    showE [] = return ' '
+    showE errs@(Index.ErrNonregisteredSource{}:_) = showPaths Index.nrPath errs
+    showE errs@(Index.ErrNonexistentSource{}:_) = showPaths Index.nePath errs
+
+    errorType Index.ErrNonregisteredSource{} Index.ErrNonregisteredSource{} = True
+    errorType Index.ErrNonexistentSource{} Index.ErrNonexistentSource{} = True
+    errorType _ _ = False
+
+    handleErrors [] = return ()
+    handleErrors errs@(Index.ErrNonregisteredSource{}:_) = putNewlineStdErr >> warn verbosity ("Sources not registered: " ++ showE errs)
```
23Skidoo: Please try to stay within 80-col limit.

cabal-install/Distribution/Client/Sandbox/Index.hs
```diff
@@ -156,39 +158,56 @@ addBuildTreeRefs verbosity path l' refType = do
...
+                       | ErrNonexistentSource   { nePath :: FilePath } deriving Show
+
 -- | Remove given local build tree references from the index.
 --
 -- Returns a tuple consisting of removed build tree refs and mappings from
 -- provided build tree refs to corresponding full directory paths).
 removeBuildTreeRefs :: Verbosity -> FilePath -> [FilePath]
-                       -> IO ([FilePath], [(FilePath, FilePath)])
+                       -> IO ([Either DeleteSourceError FilePath],
+                              (FilePath -> FilePath)) -- ^ A tuple consisting of:
```
23Skidoo: I changed the Haddock string for this function, so this comment is no longer needed.

23Skidoo:
Looks OK, but please rebase your branch and squash related commits together.

cabal-install/Distribution/Client/Sandbox.hs
martinvlk: It formats lists of paths for inclusion in messages which then get reported by functions like `warn` or `notice`. It is needed, unless there is a more appropriate way to do it I am not aware of.

23Skidoo: I was talking about `putNewlineStderr`.

martinvlk:
I'm struggling a little with git here, squashing commits, etc. Hope the result will be what we expect. Also after squashing, you seem to be recorded as an author of my changes on the commit... oh well.

cabal-install/Distribution/Client/Sandbox.hs
martinvlk: `putNewlineStderr` just adds a newline to stderr, whereas `warn` will also prepend the string "Warning: ", which is undesirable for my purposes.

23Skidoo:
That's probably because you squashed some of your commits with mine. Try doing a rebase first, and then squash. You can go back to a previous state with `git reflog` and `git reset`.

cabal-install/Distribution/Client/Sandbox.hs
23Skidoo: But why do you need a newline before `Warning: ...`? We don't do that in other places in the code. Can you show an example of output with and without the newline?

martinvlk:
Closing this pull request, will create a new one from a cleaned-up branch.

PR #2943, opened 2015-12-08
Implement better error handling for "sandbox delete-source".

martinvlk:
This is a continuation of [pull/2939](https://github.com/haskell/cabal/pull/2939) with cleaned-up commits.

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -58,7 +58,7 @@ import Distribution.Simple.Utils
          ( die, warn, info, fromUTF8, ignoreBOM )
 
 import Data.Char   (isAlphaNum)
-import Data.Maybe  (mapMaybe)
+import Data.Maybe  (mapMaybe, catMaybes)
 import Data.List   (isPrefixOf)
 #if !MIN_VERSION_base(4,8,0)
 import Data.Monoid (Monoid(..))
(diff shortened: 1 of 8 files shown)


martinvlk:
In response to the question about the newline in the output -

with the newline it might look like this:
```
Warning: Source directory not found for paths: "r" "s"
If you are trying to delete a reference to a removed directory, please provide
the full absolute path (as given by `sandbox list-sources`).

Warning: Sources not registered: "q"

cabal: The sources with the above errors were skipped.
```

While without newline it would be like this:
```
Warning: Source directory not found for paths: "r" "s"
If you are trying to delete a reference to a removed directory, please provide
the full absolute path (as given by `sandbox list-sources`).
Warning: Sources not registered: "q"
cabal: The sources with the above errors were skipped.
```

martinvlk:
I guess I could add `\n\n` into the message string itself and not do it with a separate function, although that's just a matter of style, not principle.

grayjay:
Listing the skipped sources again might make it easier to see where the errors start:

    Warning: Source directory not found for paths: "r" "s"
    If you are trying to delete a reference to a removed directory, please provide
    the full absolute path (as given by `sandbox list-sources`).
    Warning: Sources not registered: "q"
    cabal: The sources with the above errors were skipped ("q", "r", "s").

martinvlk:
Good idea - I'll do that, thanks!

martinvlk:
Ready for merge.

BardurArantsson:
LGTM.

(If there are still quibbles about the exact error message formatting, that can be addressed in a follow-up PR.)

Btw, you actually don't have to do a new PR when rebasing -- you can just force-push to the branch in your own repository -- GitHub will automatically pick those commits up as the "new set of commits".

23Skidoo:
@martinvlk Thanks for your contribution!

PR #2949, opened 2015-12-16
Type refactoring

edsko:
This PR introduces four small changes; none of these change any functionality: they just change types around a bit to facilitate further changes in later PRs.

* `Index`

The first commit introduces

``` haskell
-- | Which index do we mean?
data Index =
    -- | The global index for the specified repository
    GlobalIndex Repo
    -- | A (sandbox) local repository
  | LocalIndex { localIndexFile :: FilePath, localCacheFile :: FilePath }
```

and then uses this in lieu of passing in two `FilePath`s to lots of functions. This means the definitions

``` haskell
indexFile (GlobalIndex repo) = repoLocalDir repo </> "00-index.tar"
cacheFile (GlobalIndex repo) = repoLocalDir repo </> "00-index.cache"
```

now only live in once place, and moreover we can, if we wanted, do something special in certain cases for certain kinds of `Repo`s (we will need that functionality later).

* `Cache`

The second commit introduces a type for the index cache; right now it's just a wrapper around the cache entries

```
data Cache = Cache {
    cacheEntries :: [IndexCacheEntry]
  }
```

This should make it a little easier to extend or change the cache format if we wanted to do that.

* `Repo`

The old `Repo` type was defined as

``` haskell
data Repo = Repo {
    repoKind     :: Either RemoteRepo LocalRepo,
    repoLocalDir :: FilePath
  }
```

where `LocalRepo` was isomorphic to unit:

``` haskell
data LocalRepo = LocalRepo
```

This commit changes `Repo` to

``` haskell
data Repo =
    -- | Local repositories
    RepoLocal {
        repoLocalDir :: FilePath
      }

    -- | Standard (unsecured) remote repositores
  | RepoRemote {
        repoRemote   :: RemoteRepo
      , repoLocalDir :: FilePath
      }
```

instead, which is a little more idiomatic and will make adding more repository types easier.

* `withGlobalRepos`

Finally, the last commit changes

``` haskell 
globalRepos :: GlobalFlags -> [Repo]
```

to

``` haskell
withGlobalRepos :: Verbosity -> GlobalFlags -> ([Repo] -> IO a) -> IO a
```

This will be necessary for repositories that need to do some repository-specific initialization (even if we don't currently have any, we will soon). The `Verbosity` flag is not used yet, but will be.

### DIFF
--- a/cabal-install/Distribution/Client/BuildReports/Storage.hs
+++ b/cabal-install/Distribution/Client/BuildReports/Storage.hs
@@ -79,7 +79,9 @@ storeAnonymous reports = sequence_
                -> [(BuildReport, Repo, RemoteRepo)]
     onlyRemote rs =
       [ (report, repo, remoteRepo)
-      | (report, Just repo@Repo { repoKind = Left remoteRepo }) <- rs ]
+      | (report, Just repo) <- rs
+      , Just remoteRepo     <- [repoRemote' repo]
+      ]
 
 storeLocal :: CompilerInfo -> [PathTemplate] -> [(BuildReport, Maybe Repo)]
            -> Platform -> IO ()
--- a/cabal-install/Distribution/Client/BuildReports/Storage.hs
+++ b/cabal-install/Distribution/Client/BuildReports/Storage.hs
@@ -80,7 +80,7 @@ storeAnonymous reports = sequence_
     onlyRemote rs =
       [ (report, repo, remoteRepo)
       | (report, Just repo) <- rs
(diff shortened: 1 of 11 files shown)


23Skidoo:
LGTM with some minor comments.

edsko:
@23Skidoo I think that commit addresses all your comments.

edsko:
@23Skidoo Rebased the PR on the latest `master`.

23Skidoo:
Merged, thanks!

PR #2950, opened 2015-12-16
Remove magic value from `globalCommand`

edsko:
In the definition of `globalCommand` we had

``` haskell
    (case showOrParseArgs of ShowArgs -> take 8; ParseArgs -> id) [..]
```

the intention here is that the first 8 options will be shown when the user asks for `--help`, and the rest are not. However, this is rather error prone. If we forget to update that value when adding a new command line argument, we might either be wondering why the option isn't showing in the help text, or--worse--push another flag out of the help text without noticing.

This commit changes this to

``` haskell
    args ShowArgs  = argsShown
    args ParseArgs = argsShown ++ argsNotShown

    argsShown    = [..]
    argsNotShown = [..]
```

which is a lot more robust.


23Skidoo:
Merged, thanks!

PR #2951, opened 2015-12-16
Extend repo config with security settings

edsko:
This extends `RemoteRepo` with new fields

``` haskell
-- | Enable secure access to Hackage?
remoteRepoSecure :: Bool,

-- | Root key IDs (for bootstrapping)
remoteRepoRootKeys :: [String],

-- | Threshold for verification during bootstrapping
remoteRepoKeyThreshold :: Int,
```

and modifies the parser accordingly. This does not yet require a dependency on the security library (nor are these settings yet used anywhere).

NOTE: We suggest to make the security functionality opt-in, rather than opt-out, for the first official release. Thus, the 'secure' field will be set to `False` when initialising a `~/.cabal/config` file, and existing `~/.cabal/config` files will be interpreted as if they had `secure: False`.

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -72,7 +72,7 @@ import Distribution.ParseUtils
          , ParseResult(..), PError(..), PWarning(..)
          , locatedErrorMsg, showPWarning
          , readFields, warning, lineNo
-         , simpleField, listField, spaceListField
+         , simpleField, listField, boolField, spaceListField
          , parseFilePathQ, parseTokenQ )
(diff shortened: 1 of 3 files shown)


23Skidoo:
LGTM.

23Skidoo:
How will this work with old config files? Would be nice if we printed a sensible error message at least.

edsko:
This commit doesn't change that, that was already implemented. Old config files will be still be parsed just fine and will get sensible defaults. (What "sensible defaults" means depends on whether we go for opt-in to or opt-out of the security functionality; right now this is opt-in.)

23Skidoo:
OK, let's merge then!

PR #2952, opened 2015-12-16
WIP Nix local build

ezyang:
Creating a PR for the nix-local-build branch so I can make comments on the diff. Not ready for merge yet, I think.


Cabal/Distribution/Simple/PackageIndex.hs
```diff
@@ -304,6 +305,30 @@ deleteDependency (Dependency name verstionRange) =
   delete' name (\pkg -> packageVersion pkg `withinRange` verstionRange)
 -}
 
+-- | Map over all the packages in the index, but each package's id is not
```
ezyang: This isn't used anywhere, do we still need it?

cabal-install/Main.hs
```diff
@@ -268,6 +270,9 @@ mainWorker args = topHandler $
       , hiddenCmd  win32SelfUpgradeCommand win32SelfUpgradeAction
       , hiddenCmd  actAsSetupCommand actAsSetupAction
       , hiddenCmd  manpageCommand (manpageAction commandSpecs)
+
+      , hiddenCmd  installCommand { commandName = "new-configure" } newConfigureAction
```
ezyang: Can I use cabal install with the new configure / new build?

cabal-install/Distribution/Client/SetupWrapper.hs
```diff
@@ -122,6 +122,7 @@ import qualified System.Win32 as Win32
 
 data SetupScriptOptions = SetupScriptOptions {
     useCabalVersion          :: VersionRange,
+    useCabalSpecVersion      :: Maybe Version,
```
ezyang: What is this? Needs a comment, or SetupScriptOptions needs a reference to where these are defined. Ditto with useVersionMacros.

cabal-install/Distribution/Client/SetupWrapper.hs
```diff
@@ -256,7 +262,8 @@ type SetupMethod = Verbosity
 internalSetupMethod :: SetupMethod
 internalSetupMethod verbosity options _ bt mkargs = do
   let args = mkargs cabalVersion
-  debug verbosity $ "Using internal setup method with build-type " ++ show bt
+  --TODO: just temporarily higher verbosity
+  info verbosity $ "Using internal setup method with build-type " ++ show bt
```
ezyang: Don't forget to revert this

cabal-install/Main.hs
dcoutts: No, there's no corresponding install command yet.

Cabal/Distribution/Simple/PackageIndex.hs
dcoutts: Right, I stopped using it. Removed.

cabal-install/Distribution/Client/SetupWrapper.hs
```diff
@@ -338,18 +345,24 @@ externalSetupMethod verbosity options pkg bt mkargs = do
...
-              -- Does the previously compiled setup executable still exist and
-              -- is it up-to date?
-              useExisting <- canUseExistingSetup version
-              if useExisting
-                then return (version, Nothing, options)
-                else installedVersion
-      _ -> installedVersion
+  cabalLibVersionToUse = 
+    case useCabalSpecVersion options of
+      Just version -> do
```
ezyang: The case on useCabalSpecVersion is new; the rest is just indent fuzz.

cabal-install/Distribution/Client/Search.hs
```diff
@@ -0,0 +1,589 @@
+-----------------------------------------------------------------------------
```
ezyang: This isn't imported by anyone; let's put in another branch/CR?

cabal-install/Distribution/Client/RebuildMonad.hs
```diff
@@ -0,0 +1,90 @@
...
+import Data.Maybe      (fromMaybe)
+import Data.Binary     (Binary)
+import System.FilePath (takeFileName)
+
+
+-------------------------------
+-- Simple rebuild abstraction
+--
+
+newtype Rebuild a = Rebuild (StateT [MonitorFilePath] IO a)
```
ezyang: Shouldn't this be WriterT?

cabal-install/Distribution/Client/RebuildMonad.hs
```diff
@@ -0,0 +1,90 @@
...
+import Distribution.Verbosity    (Verbosity)
+
+import Control.Applicative
+import Control.Monad.State as State
+import Data.Maybe      (fromMaybe)
+import Data.Binary     (Binary)
+import System.FilePath (takeFileName)
+
+
+-------------------------------
```
ezyang: More docs please.

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,619 @@
...
+     -- and content hash. The monitored file is considered to have changed if
+     -- it no longer exists or if its modification time and content hash have
+     -- changed.
+     --
+   | MonitorFileHashed !FilePath
+
+     -- | Monitor a single non-existant file for changes. The monitored file
+     -- is considered to have changed if it exists.
+     --
+   | MonitorNonExistantFile !FilePath
```
ezyang: s/Existant/Existent/

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,619 @@
...
+   | MonitorFileGlob !FilePathGlob
+
+  deriving (Show, Generic)
+
+instance Binary MonitorFilePath
+
+-- | A file path specified by globbing
+--
+data FilePathGlob
+   = GlobDir  !Glob !FilePathGlob
```
ezyang: What's a GlobDir. Docs!

cabal-install/Main.hs
ezyang: OK, I see now, cabal new-configure actually goes and downloads deps; and cabal new-build will build them all. Strange names.

cabal-install/Distribution/Client/SetupWrapper.hs
dcoutts: Added a comment for that and another one.

cabal-install/Distribution/Client/SetupWrapper.hs
dcoutts: Ta. I'll eventually prepare patches by a manual diff process, so I'll notice things like this.

cabal-install/Distribution/Client/Search.hs
dcoutts: Oh, oops, yes it's not supposed to be in this branch.

cabal-install/Distribution/Client/RebuildMonad.hs
dcoutts: No, WriterT is always wrong. :-)

This monad could be an instance of the writer class, but the WriterT impl is not a good implementation of that abstraction. It's necessarily leaky.

cabal-install/Distribution/Client/SetupWrapper.hs
dcoutts: Is this a complaint or a request for a change? I'm not sure what you mean here.

cabal-install/Distribution/Client/RebuildMonad.hs
ezyang: Oh, very interesting! Maybe we should remark that it is just used as Writer, but it's not Writer to avoid space leak.

cabal-install/Main.hs
dcoutts: Not really. Currently configure just does the first phases of build, and currently the planning phase will download packages. Eventually configure will persist config flags by writing them info a local file that extends the cabal.project file.

Note that once we have the hackage security stuff merged then the planning phase will not need to download tarballs, since it'll have the tarball hashes up front.

ezyang: Ah, OK.  But I think it is true that "new-build" is effectively
"install"?

Excerpts from Duncan Coutts's message of 2015-12-16 16:42:44 -0800:

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,619 @@
...
+monitorFileHashedSearchPath :: [FilePath] -> FilePath -> [MonitorFilePath]
+monitorFileHashedSearchPath notFoundAtPaths foundAtPath =
+    MonitorFileHashed foundAtPath
+  : map MonitorNonExistentFile notFoundAtPaths
+
+-- | A file used to store the state of a file monitor.
+--
+-- It is typed to ensure it's used at consistent types.
+--
+newtype FileMonitorCacheFile a b
```
ezyang: What are the phantom types a and b supposed to represent?

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,619 @@
...
+   | MonitorStateFileHashChanged
+  deriving (Show, Generic)
+
+instance Binary MonitorStateFile
+
+-- | The state necessary to determine whether the files matched by a globbing
+-- match have changed.
+--
+data FileGlobMonitorState
+   = MonitorStateGlobDirs  !Glob !FilePathGlob
```
ezyang: Would appreciate an explanation of what's going on here.

cabal-install/Distribution/Client/FileStatusCache.hs
dcoutts: Will document. It's a key and value. If the key has not changed, you get back the cached value. If the key has changed, you have to run an action to produce a new value.

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,619 @@
...
+    getGlobPath (MonitorStateGlobDirs  glob globs _ _) =
+      MonitorFileGlob (GlobDir  glob globs)
+    getGlobPath (MonitorStateGlobFiles glob       _ _) =
+      MonitorFileGlob (GlobFile glob)
+
+------------------------------------------------------------------------------
+-- Checking the status of monitored files
+--
+
+data Changed b = Unchanged b | Changed (Maybe FilePath)
```
ezyang: I don't understand this type. When is Changed Nothing return? When the file went away? What does the file represent? The usage immediately below doesn't illuminate much either.

cabal-install/Distribution/Client/FileStatusCache.hs
dcoutts: Yeah, this stuff is subtle. The big picture is we're efficiently determining if a set of files has changed or not. But rather than just a set of files, we can also monitor files defined by a file glob. Doing that efficiently is a little more tricky.

cabal-install/Distribution/Client/FileStatusCache.hs
dcoutts: The result in the Changed case is not very important. It's for reporting which file if any changed, that caused us to report that something changed. It can be nothing because it might not be a file that changed (it could be the key value, or it could be the first time or the binary cache could be in an old format).

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,627 @@
...
+    MonitorStateFileSet
+      <$> Map.traverseWithKey (probeFileStatus root)     singlePaths
+      <*> traverse            (probeGlobStatus root ".") globPaths
+
+-----------------------------------------------
+-- Monad for checking for file system changes
+--
+-- We need to be able to bail out if we detect a change (using ExceptT),
+-- but if there's no change we need to be able to rebuild the monitor
+-- state. And we want to optimise that rebuilding by keeping track if
```
ezyang: I would love an example where there is no change, but we need to rebuild the monitor state (because something changed.)

cabal-install/Distribution/Client/FileStatusCache.hs
```diff
@@ -0,0 +1,627 @@
...
+
+    return (MonitorStateGlobFiles glob mtime' children)
+    -- Again, we don't force a cache rewite with 'cacheChanged', but we do use
+    -- the new mtime' if any.
+  where
+    isInBoth :: MergeResult a b -> Bool
+    isInBoth (InBoth _ _) = True
+    isInBoth _            = False
+
+------------------------------------------------------------------------------
```
ezyang: I don't know what any of this does and I'd really appreciate short pointers on what these are supposed to do.

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -178,6 +181,10 @@ data PackagePreference =
      -- | If we prefer versions of packages that are already installed.
    | PackageInstalledPreference PackageName InstalledPreference
 
+   | PackageStanzasPreference   PackageName [OptionalStanza]
```
ezyang: Obviously, this is some preference related to stanzas. But what kind of preference is it? Are we trying to respect the dependencies added by the optional stanzas? Needs a comment.

cabal-install/Distribution/Client/FileStatusCacheTest.hs
```diff
@@ -0,0 +1,157 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
ezyang: Dumb question, but why isn't this in the tests directory?

cabal-install/Distribution/Client/InstallPlan.hs
```diff
@@ -177,16 +189,27 @@ instance (HasComponentId ipkg, HasComponentId srcpkg) =>
 
 
 data GenericInstallPlan ipkg srcpkg iresult ifailure = GenericInstallPlan {
-    planIndex      :: (PlanIndex ipkg srcpkg iresult ifailure),
-    planFakeMap    :: FakeMap,
+    planIndex      :: !(PlanIndex ipkg srcpkg iresult ifailure),
+    planFakeMap    :: !FakeMap,
```
ezyang: I wonder if we can get rid of this now.

cabal-install/Distribution/Client/PackageCollection.hs
```diff
@@ -0,0 +1,580 @@
+module Distribution.Client.PackageCollection (
```
ezyang: This isn't used anywhere.

cabal-install/Main.hs
hvr: > Ah, OK.  But I think it is true that "new-build" is effectively "install"?

not currently; afaiui, it's more like a hypothetical `cabal build --auto-install-dependencies` with an implicit project-sandbox and a nix-style build-dep cache

cabal-install/Distribution/Client/FileStatusCache.hs
dcoutts: Consider an example like this:

 * matching glob: ./*/*.cabal
 * initial files: ./foo/foo.cabal  ./tmp/

So the cache has info that the top level dir has foo and tmp, and has subtrees that tell us that foo contains a match but tmp does not.

Now I do mkdir bar; touch bar/blah and re-check. There's a new dir at the top level that matches the '*' glob. So we want to check if anything underneath matches (if it does then that's an actual change). But if nothing matches then we still want to update the cache about what is there (we might not discover there's no matches 'til we've done a great deal of work). So this is a circumstance in which we will rebuild and rewrite the cache. There's potentially a lot of work that would otherwise have to be repeated. Think about unpacking a new package (but where for some reason it didn't end up matching our glob), if we didn't update the cache we might have to rescan the whole thing every time (depending on the glob pattern).

There are other cases where we could update the cache, but we don't because the what we save in those cases is more limited, and there is a cost to rewriting the cache. But the new subtree case is potentially a large amount of work that'd have to be repeated each time.

cabal-install/Distribution/Client/FileStatusCacheTest.hs
dcoutts: It should be moved.

cabal-install/Distribution/Client/InstallPlan.hs
dcoutts: I'd really like to get rid of the fake id map.

cabal-install/Distribution/Client/PackageCollection.hs
dcoutts: Oh, also a file that should be in a different feature branch.

cabal-install/Main.hs
dcoutts: Right, it's also not install because install changes the user's environment and build does not. By change the user's environment I mean either: adding binaries to user visible path, or adding packages to a user visible environment, like the per-user env.

ezyang:
Closing this.

PR #2960, opened 2015-12-17
Define Semigroup instances

hvr:
This makes the code `-Wcompat`-clean with GHC 8.0

Due to the amount of `Monoid` instances, a compat-layer is used
rather than flooding the code-base with CPP conditionals.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -253,6 +253,7 @@ library
   other-modules:
     Distribution.Compat.Binary
     Distribution.Compat.CopyFile
+    Distribution.Compat.Semigroup
     Distribution.GetOpt
     Distribution.Lex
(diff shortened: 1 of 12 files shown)


23Skidoo:
Can you please fix the Travis failure?

23Skidoo:
Otherwise looks good.

hvr:
The code should be reasonably future-proof, requiring only minor tweaks when the next phases of the Semigroup proposal become active.

hvr:
`Distribution.PackageDescription` is a bit more finnicky as there's a nameclash with `Text.PrettyPrint ((<>))` which I didn't want to resolve yet in this patch.

hvr:
@23Skidoo *bump*

23Skidoo:
Merged, thanks!

PR #2962, opened 2015-12-18
Separate out reading index from building cache

edsko:
The cache construction was using lazy I/O to avoid building the cache in memory all at once, but I found the algorithm a little hard to follow. This commit makes a few changes; they don't change anything fundamental, just make it a bit easier to follow:

* Separated out

    ``` haskell
    tarEntriesList :: Tar.Entries -> [(BlockNo, Tar.Entry)]
    ```

  into its own function. This now makes the top-level of `parsePackageIndex` a straight-forward composition

    ``` haskell
    concatMap (uncurry extract) . tarEntriesList . Tar.read
    ```
 
  so that it's easier to see that this indeed as lazy as we want. (Previously especially the name "accum" suggests an accumulator and hence a foldl rather than a foldr style, which would not be good; that bit of the algorithm is now the separate `tarEntriesList` function.)

* Changed the definition of `lazySequence` to avoid the unnecessary nested use of `unsafeInterleaveIO`; a single one suffices:

    ``` haskell
    lazySequence :: [IO a] -> IO [a]
    lazySequence = unsafeInterleaveIO . go
      where
        go []     = return []
        go (x:xs) = do x'  <- x
                       xs' <- lazySequence xs
                       return (x' : xs')
    ```

* Introduced `withIndexEntries`, which makes it very clear in which scope we can, and must, consume the lazy list we get from all this.

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -60,7 +60,7 @@ import Distribution.Simple.Utils
          ( die, warn, info, fromUTF8, ignoreBOM )
 
 import Data.Char   (isAlphaNum)
-import Data.Maybe  (mapMaybe, catMaybes)
+import Data.Maybe  (mapMaybe, catMaybes, maybeToList)
 import Data.List   (isPrefixOf)
 #if !MIN_VERSION_base(4,8,0)
 import Data.Monoid (Monoid(..))
@@ -242,8 +242,6 @@ typeCodeFromRefType :: BuildTreeRefType -> Tar.TypeCode
 typeCodeFromRefType LinkRef     = Tar.buildTreeRefTypeCode
 typeCodeFromRefType SnapshotRef = Tar.buildTreeSnapshotTypeCode
 
-type MkPackageEntry = IO (Maybe PackageEntry)
-
 instance Package PackageEntry where
   packageId (NormalPackage  pkgid _ _ _) = pkgid
   packageId (BuildTreeRef _ pkgid _ _ _) = pkgid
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/IndexUtils.hs
```diff
@@ -258,32 +256,41 @@ packageDesc (BuildTreeRef _ _ descr _ _) = descr
...
+-- and pair each entry with its block number.
+--
+-- NOTE: This preserves the lazy nature of 'Entries': the tar file is only read
+-- as far as the list is evaluated.
+tarEntriesList :: Tar.Entries -> [(BlockNo, Tar.Entry)]
+tarEntriesList = go 0
+  where
+    go _ Tar.Done         = []
+    go _ (Tar.Fail e)     = error ("tarEntriesList: " ++ e)
+    go n (Tar.Next e es') = (n, e) : go (n + Tar.entrySizeInBlocks e) es'
```
23Skidoo: Probably should be `go !n` to avoid creating unnecessary thunk.

cabal-install/Distribution/Client/IndexUtils.hs
```diff
@@ -319,31 +326,38 @@ extractPkg entry blockNo = case Tar.entryContent entry of
...
-      prefs     = parsePreferredVersions (BS.Char8.unpack content)
+      prefs     = parsePreferredVersions content
   _ -> Nothing
 
-parsePreferredVersions :: String -> [Dependency]
+parsePreferredVersions :: ByteString -> [Dependency]
 parsePreferredVersions = mapMaybe simpleParse
                        . filter (not . isPrefixOf "--")
                        . lines
+                       . BS.Char8.unpack -- TODO: Are we sure no unicode?
```
23Skidoo: I think we are, Hackage does not allow libraries with Unicode names.

23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #2968, opened 2015-12-21
Introduce ignore-expiry cmdline arg

edsko:
If used, this will instruct the hackage-security library (once integrated_ to ignore expiry dates on TUF metadata (to be used only if, say, the main server is down and we need to rely on mirrors that have not been updated for longer than the expiry period on the timestamp). Right now this command line argument is unused.

I think this will be the last PR before the PR that adds support for secure repositories. I've set the series of PRs up in such a way that that last PR will be very small and will also make it very obvious that this does not change any of the old code paths (i.e., if you don't opt-in for security you will not get a change in behaviour). However, it will rely on as-yet-unreleased versions of both hackage-security and tar. 

Anyway, this commit is still independent.

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -129,6 +129,7 @@ data GlobalFlags = GlobalFlags {
     globalWorldFile         :: Flag FilePath,
     globalRequireSandbox    :: Flag Bool,
     globalIgnoreSandbox     :: Flag Bool,
+    globalIgnoreExpiry      :: Flag Bool,    -- ^ Ignore security expiry dates
     globalHttpTransport     :: Flag String
   }
 
@@ -146,6 +147,7 @@ defaultGlobalFlags  = GlobalFlags {
     globalWorldFile         = mempty,
     globalRequireSandbox    = Flag False,
     globalIgnoreSandbox     = Flag False,
+    globalIgnoreExpiry      = Flag False,
     globalHttpTransport     = mempty
   }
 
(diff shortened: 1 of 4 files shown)


cabal-install/Distribution/Client/Setup.hs
```diff
@@ -309,6 +311,11 @@ globalCommand commands = CommandUI {
          globalIgnoreSandbox (\v flags -> flags { globalIgnoreSandbox = v })
          trueArg
 
+      ,option [] ["ignore-expiry"]
+         "Ignore expiry dates on signed metadata (use only in exception circumstances)"
```
phadej: *exceptional* ?

edsko: Haha, fair point :) Will fix.

23Skidoo:
LGTM, merged!

23Skidoo:
One minor nit: in general, I prefer using custom datatypes (e.g. `data IgnoreExpiry = IgnoreExpiry | DoNotIgnoreExpiry`) instead of [boolean flags](https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/). This rule is not followed across the whole codebase, though.

> I've set the series of PRs up in such a way that that last PR will be very small and will also make it very obvious that this does not change any of the old code paths (i.e., if you don't opt-in for security you will not get a change in behaviour)

Thanks, that has definitely made you PRs easier to review!

edsko:
Yeah, fair enough. Actually, this `Bool` is not passed to the hackage-security library as is but is instead translated to `Maybe UTCTime` (the time to check expiry against, if doing such a check). Since it only appears in a handful of places in the rest of `cabal` I didn't introduce a new datatype. But I take your point nonetheless.

23Skidoo:
> Since it only appears in a handful of places in the rest of `cabal` I didn't introduce a new datatype.

Fair enough, OTOH creating new datatypes is quite cheap.

PR #2975, opened 2015-12-22
Also probe in canonicalized path, if binary was a symlink.

ezyang:


ezyang:
This makes things like, you have a solitary ghc-7.10 symlinked to the real ghc, and that's the directory that has ghc-pkg. (Yes, adding a ghc-pkg-7.10 symlink would also work)

23Skidoo:
Merged, thanks! Would be also nice to see a comment.

PR #2984, opened 2015-12-22
Fix #2645 - make cabal merge package preferences if more than one is specified

### LINKED ISSUE #2645: Properly merge preferences (dependency resolution)
Preferences, as in "soft constraints" on package versions are not properly merged. The primary symptom is that for any package with any "preferred versions" defined via hackage, any preferences given on the commandline are ignored.

In the output of

~~~~.bash
> cabal install --dry-run gl --preference="gl<0.7" -v3
~~~~

contains the following two (!) lines for the gl package:

1) `gl <0.7` (from commandline flag)
2) `gl <0.7.2.1 || >0.7.2.1` (hackage preferred version)

And seemingly, the second one blindly overwrites the first. (You can observe the changed behaviour by editing the index.cache in ~/.cabal/packages/.. and temporarily removing the gl preference.) I believe changing the order as a solution for the issue is not sufficient, because then hackage preferences are ignored as soon as the user provides some preference. Instead, the two preferences should be merged. (This merging functionality must already exist for hard constraints, so this should be no problem, right?)

I would suggest giving info to the user (even without verbose) when such a merging happens.

martinvlk:
Creating PR for this fix, need to discuss one thing before it is ready for merge. I will comment below.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -587,13 +587,12 @@ interpretPackagesPreference :: Set PackageName
                             -> (PackageName -> PackagePreferences)
 interpretPackagesPreference selected defaultPref prefs =
   \pkgname -> PackagePreferences (versionPref pkgname) (installPref pkgname)
-
   where
     versionPref pkgname =
-      fromMaybe anyVersion (Map.lookup pkgname versionPrefs)
-    versionPrefs = Map.fromList
-      [ (pkgname, pref)
-      | PackageVersionPreference pkgname pref <- prefs ]
+      fromMaybe [anyVersion] (Map.lookup pkgname versionPrefs)
+    versionPrefs = Map.fromListWith (++)
+                   [(pkgname, [pref])
+                   | PackageVersionPreference pkgname pref <- prefs]
 
     installPref pkgname =
       fromMaybe (installPrefDefault pkgname) (Map.lookup pkgname installPrefs)
@@ -818,7 +817,8 @@ resolveWithoutDependencies (DepResolverParams targets constraints
                            . InstalledPackageIndex.lookupSourcePackageId
                                                      installedPkgIndex
                            . packageId
-        versionPref   pkg = packageVersion pkg `withinRange` preferredVersions
+        versionPref pkg = length . filter (packageVersion pkg `withinRange`) $
+                          preferredVersions
 
     packageConstraints :: PackageName -> VersionRange
     packageConstraints pkgname =
--- a/cabal-install/Distribution/Client/Dependency/TopDown.hs
+++ b/cabal-install/Distribution/Client/Dependency/TopDown.hs
@@ -127,7 +127,9 @@ explore pref (ChoiceNode _ choices)  =
       where
         isInstalled (SourceOnly _) = False
         isInstalled _              = True
-        isPreferred p = packageVersion p `withinRange` preferredVersions
+        isPreferred p = length . filter (packageVersion p `withinRange`) $
+                        preferredVersions
+
         (PackagePreferences preferredVersions packageInstalledPreference)
           = pref pkgname
 
@@ -669,9 +671,11 @@ finaliseSelectedPackages pref selected constraints =
         -- version constraints. TODO: distinguish hacks from prefs
         bounded = boundedAbove versionRange
         isPreferred p
-          | bounded   = True -- any constant will do
-          | otherwise = packageVersion p `withinRange` preferredVersions
+          | bounded   = boundedRank -- this is a dummy constant
+          | otherwise = length . filter (packageVersion p `withinRange`) $
+                        preferredVersions
           where (PackagePreferences preferredVersions _) = pref (packageName p)
+        boundedRank = 0 -- any value will do
 
         boundedAbove :: VersionRange -> Bool
         boundedAbove vr = case asVersionIntervals vr of
--- a/cabal-install/Distribution/Client/Dependency/Types.hs
+++ b/cabal-install/Distribution/Client/Dependency/Types.hs
@@ -32,6 +32,7 @@ module Distribution.Client.Dependency.Types (
     ConstraintSource(..),
     unlabelPackageConstraint,
     showConstraintSource
+
   ) where
 
 #if !MIN_VERSION_base(4,8,0)
@@ -156,17 +157,18 @@ showPackageConstraint (PackageConstraintStanzas pn ss) =
     showStanza TestStanzas  = "test"
     showStanza BenchStanzas = "bench"
 
--- | A per-package preference on the version. It is a soft constraint that the
+-- | Per-package preferences on the version. It is a soft constraint that the
 -- 'DependencyResolver' should try to respect where possible. It consists of
--- a 'InstalledPreference' which says if we prefer versions of packages
--- that are already installed. It also has a 'PackageVersionPreference' which
--- is a suggested constraint on the version number. The resolver should try to
--- use package versions that satisfy the suggested version constraint.
+-- an 'InstalledPreference' which says if we prefer versions of packages
+-- that are already installed. It also has (possibly multiple)
+-- 'PackageVersionPreference's which are suggested constraints on the version
+-- number. The resolver should try to use package versions that satisfy
+-- the maximum number of the suggested version constraints.
 --
 -- It is not specified if preferences on some packages are more important than
 -- others.
 --
-data PackagePreferences = PackagePreferences VersionRange InstalledPreference
+data PackagePreferences = PackagePreferences [VersionRange] InstalledPreference
 
 -- | Whether we prefer an installed version of a package or simply the latest
 -- version.
--- a/cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/DSL.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/DSL.hs
@@ -2,7 +2,10 @@
 -- | DSL for testing the modular solver
 module UnitTests.Distribution.Client.Dependency.Modular.DSL (
     ExampleDependency(..)
+  , ExPreference(..)
   , ExampleDb
+  , ExampleVersionRange
+  , ExamplePkgVersion
   , exAv
   , exInst
   , exResolve
@@ -84,6 +87,7 @@ type ExamplePkgVersion = Int
 type ExamplePkgHash    = String  -- for example "installed" packages
 type ExampleFlagName   = String
 type ExampleTestName   = String
+type ExampleVersionRange = C.VersionRange
 
 data ExampleDependency =
     -- | Simple dependency on any version
@@ -104,6 +108,7 @@ data ExampleDependency =
     -- | Dependency on a language version
   | ExLang Language
 
+data ExPreference = ExPref String ExampleVersionRange
 
 data ExampleAvailable = ExAv {
     exAvName    :: ExamplePkgName
@@ -304,8 +309,9 @@ exResolve :: ExampleDb
           -> [Language]
           -> [ExamplePkgName]
           -> Bool
+          -> [ExPreference]
           -> ([String], Either String CI.InstallPlan.InstallPlan)
-exResolve db exts langs targets indepGoals = runProgress $
+exResolve db exts langs targets indepGoals prefs = runProgress $
     resolveDependencies C.buildPlatform
                         compiler
                         Modular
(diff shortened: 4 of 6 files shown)


cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -587,13 +587,19 @@ interpretPackagesPreference :: Set PackageName
...
     versionPref pkgname =
-      fromMaybe anyVersion (Map.lookup pkgname versionPrefs)
-    versionPrefs = Map.fromList
-      [ (pkgname, pref)
-      | PackageVersionPreference pkgname pref <- prefs ]
+      fromMaybe [anyVersion] (Map.lookup pkgname versionPrefs)
+    versionPrefs = foldr mergeIfMultiplePrefs Map.empty prefs
+    mergeIfMultiplePrefs (PackageVersionPreference pkgname pref) m =
+      Map.insertWith (++) pkgname [pref] m
+    -- Previously the two cases below were simply ignored, and I think for a good
```
martinvlk: See the comment here - how do we handle "installed" or "latest" preference in the input pref list? Previously they were simply ignored by a partial list comprehension.

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -587,13 +587,19 @@ interpretPackagesPreference :: Set PackageName
...
+    -- reason, because it is not possible to specify installed or latest preference
+    -- for a package. What to do?
+    mergeIfMultiplePrefs
```
martinvlk: See the comment here - how do we handle "installed" or "latest" preference in the input pref list? Previously they were simply ignored by a partial list comprehension.

23Skidoo:
Travis failures are due to `-Werror`.

grayjay:
Do you mind if I review this pull request, too?

23Skidoo:
Sure, go ahead.

grayjay:
I'll add some advice and questions to the diff.  Most of the changes look fine.

cabal-install/Distribution/Client/Dependency.hs
grayjay: `InstalledPreference` is handled below in `installPrefs`, so you can continue filtering it out here.  `interpretPackagesPreference` needs to handle both constructors because sandboxes add preferences on installed instances.

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -818,7 +824,8 @@ resolveWithoutDependencies (DepResolverParams targets constraints
                            . InstalledPackageIndex.lookupSourcePackageId
                                                      installedPkgIndex
                            . packageId
-        versionPref   pkg = packageVersion pkg `withinRange` preferredVersions
+        versionPref   pkg = packageVersion pkg `withinRange`
+                            combinePrefRanges preferredVersions
```
grayjay: Why does this function handle preferences differently than in the modular solver?  This seems more consistent:
```versionPref   pkg = length $ filter (packageVersion pkg `withinRange`) preferredVersions```

File: cabal-install/Distribution/Client/Dependency/TopDown.hs [hunk 4]
```diff
@@ -127,7 +127,8 @@ explore pref (ChoiceNode _ choices)  =
       where
         isInstalled (SourceOnly _) = False
         isInstalled _              = True
-        isPreferred p = packageVersion p `withinRange` preferredVersions
+        isPreferred p = packageVersion p `withinRange`
+                        combinePrefRanges preferredVersions
         (PackagePreferences preferredVersions packageInstalledPreference)
```
grayjay: The comment in `resolveWithoutDependencies` also applies here, and on the next change in this file.  I think that it would be better to either keep the old behavior or use the same comparison as in the modular solver, as long as the change isn't much larger.

cabal-install/Distribution/Client/Dependency/Types.hs
```diff
@@ -166,7 +168,11 @@ showPackageConstraint (PackageConstraintStanzas pn ss) =
 -- It is not specified if preferences on some packages are more important than
```
grayjay: It would be good to update this comment and explain the meaning of multiple ranges.

cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/DSL.hs
```diff
@@ -104,6 +107,7 @@ data ExampleDependency =
     -- | Dependency on a language version
   | ExLang Language
 
+type ExPreference = (String, ExampleVersionRange)
```
grayjay: I think that converting `ExPreference` into a data type would make the test cases easier to read.

cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/DSL.hs
```diff
@@ -303,9 +307,9 @@ exResolve :: ExampleDb
           -- A compiler can support multiple languages.
           -> [Language]
           -> [ExamplePkgName]
-          -> Bool
+          -> (Bool, [ExPreference])
           -> ([String], Either String CI.InstallPlan.InstallPlan)
-exResolve db exts langs targets indepGoals = runProgress $
+exResolve db exts langs targets (indepGoals, prefs) = runProgress $
```
grayjay: Why are independent goals and preferences combined in a pair?

cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/Solver.hs
```diff
@@ -85,9 +88,25 @@ tests = [
...
         ]
+
+     , testGroup "Soft Constraints" [
+          runTest $ soft [("A", mkvrThis [1,0,0])]      $ mkTest db13 "selectPreferredVersionSimple" ["A"] (Just [("A", 1)])
+        , runTest $ soft [("A", mkvrOrEarlier [2,0,0])] $ mkTest db13 "selectPreferredVersionSimple2" ["A"] (Just [("A", 2)])
+        , runTest $ soft [("A", mkvrOrEarlier [2,0,0])
+                         ,("A", mkvrOrEarlier [1,0,0])] $ mkTest db13 "selectPreferredVersionMultiple" ["A"] (Just [("A", 1)])
+        , runTest $ soft [("A", mkvrOrEarlier [1,0,0])
+                         ,("A", mkvrOrEarlier [2,0,0])] $ mkTest db13 "selectPreferredVersionMultiple2" ["A"] (Just [("A", 1)])
+        ]
```
grayjay: I like the choice of test cases.

cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/Solver.hs
```diff
@@ -85,9 +88,25 @@ tests = [
...
     ]
   where
     indep test = test { testIndepGoals = True }
+    soft prefs test = test { testSoftConstraints = prefs }
+
+mkvrThis :: [Int] -> V.VersionRange
+mkvrThis = V.thisVersion . V.makeVersion 
+
+mkvrOrEarlier :: [Int] -> V.VersionRange
+mkvrOrEarlier = V.orEarlierVersion . V.makeVersion
```
grayjay: You could simplify the versions in the test cases by making `mkvrThis` and `mkvrOrEarlier` both have type `ExamplePkgVersion -> V.VersionRange`.

martinvlk:
Hi, addressing review comments, all fair points - thanks!

cabal-install/Distribution/Client/Dependency/TopDown.hs
```diff
@@ -669,8 +671,9 @@ finaliseSelectedPackages pref selected constraints =
         -- version constraints. TODO: distinguish hacks from prefs
         bounded = boundedAbove versionRange
         isPreferred p
-          | bounded   = True -- any constant will do
-          | otherwise = packageVersion p `withinRange` preferredVersions
+          | bounded   = 999 -- any constant will do
```
martinvlk: I am not sure here about this `bounded` case, what weight should we assign to it relative to the other preferences? Not sure I understand this bit.

23Skidoo:
@grayjay Thanks for helping out with code review!

grayjay:
@23Skidoo Glad I can help.  I'm reviewing the latest set of changes now.

grayjay:
@martinvlk Thanks for making the changes.  I have a few more suggestions.

It looks like your first commit was combined with an existing commit, maybe because you rebased against the wrong branch.  Since you only modified code in the second commit, you can fix it by cherry-picking your new second commit on top of your original first commit, if you still have it.

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -587,13 +587,12 @@ interpretPackagesPreference :: Set PackageName
...
   where
     versionPref pkgname =
-      fromMaybe anyVersion (Map.lookup pkgname versionPrefs)
-    versionPrefs = Map.fromList
-      [ (pkgname, pref)
-      | PackageVersionPreference pkgname pref <- prefs ]
+      fromMaybe [anyVersion] (Map.lookup pkgname versionPrefs)
+    versionPrefs = foldr (\(PackageVersionPreference pkgname pref) m ->
+                            Map.insertWith (++) pkgname [pref] m)
+                   Map.empty prefs
```
grayjay: This is a partial function now, and it could fail in a sandbox.  You could use a list comprehension and `Map.fromListWith`, since a pattern in a list comprehension acts as a filter.

cabal-install/Distribution/Client/Dependency/TopDown.hs
grayjay: I'm not familiar with the Topdown solver, but it looks like this function ignores preferred ranges when the dependency is bounded above.  Since `bounded` doesn't depend on the package instances being compared, the value here will only be compared with itself and should have no effect on behavior (which is consistent with the comment).  I would use something like `0` or `1`, so that it doesn't look like the constant has special meaning.  I guess moving the guard into `bestByPref` would remove the need for an arbitrary value, but I don't recommend refactoring it now that the solver is deprecated.

cabal-install/tests/UnitTests/Distribution/Client/Dependency/Modular/Solver.hs
```diff
@@ -85,9 +88,22 @@ tests = [
...
+
+     , testGroup "Soft Constraints" [
+          runTest $ soft [ ExPref "A" $ mkvrThis 1]      $ mkTest db13 "selectPreferredVersionSimple" ["A"] (Just [("A", 1)])
+        , runTest $ soft [ ExPref "A" $ mkvrOrEarlier 2] $ mkTest db13 "selectPreferredVersionSimple2" ["A"] (Just [("A", 2)])
+        , runTest $ soft [ ExPref "A" $ mkvrOrEarlier 2
+                         , ExPref "A" $ mkvrOrEarlier 1] $ mkTest db13 "selectPreferredVersionMultiple" ["A"] (Just [("A", 1)])
+        , runTest $ soft [ ExPref "A" $ mkvrOrEarlier 1
+                         , ExPref "A" $ mkvrOrEarlier 2] $ mkTest db13 "selectPreferredVersionMultiple2" ["A"] (Just [("A", 1)])
+        ]
     ]
```
grayjay: All of the test changes look good.  I just thought of another test case, if you want to add more.  Testing a preference on A==1 and a preference on A==2 would distinguish between the new behavior and merging preferences by taking the intersection of version ranges.

martinvlk:
Right so, I fixed the situation with wrongly squashed commits and also addressed the rest of the comments. BTW - thanks for them, this is helping me to bring my largely theoretical Haskell knowledge to real life - your patience is much appreciated!

This should now be ready for merging.

grayjay:
@martinvlk  No problem.  Thanks for fixing this issue!

The two new test cases have duplicate names.  Other than that, I think this pull request is ready to be merged.

martinvlk:
Ready for merge so!

grayjay:
Thanks!  It looks good.

BardurArantsson:
@grayjay Do you also want @kosmikus to review, or do think your sign-off is enough, i.e. do you think you have enough of a handle on the changes[1]? If you feel comfortable with it, I'm happy to press the "Merge" button.

[1] I'd assume you do, but you know what they say about assumptions...

grayjay:
@BardurArantsson I've reviewed the whole pull request, but I want to check with @23Skidoo before it is merged.

BardurArantsson:
A-okay :)

martinvlk:
Ping, any movement on this one?

BardurArantsson:
OK, I've reviewed this and even though I don't understand the solver in much detail, it all seems reasonable to me. (See comments.)

@23Skidoo Any objection to me merging this?

BardurArantsson:
Merged. Thanks for the contribution @martinvlk !

PR #2987, opened 2015-12-23
InstalledPackageId has constructor

Heather:


Heather:
@23Skidoo is it my mistake even?

ezyang:
InstalledPackageId got turned into a type synonym for backwards compatibility. I suppose for sufficiently recent GHC we could also add a pattern synonym so it really does look BC OK.

23Skidoo:
Going to close this for now, please send another PR when/if you incorporate @ezyang's feedback.

ezyang:
@23Skidoo , this is not quite related to this PR, but maybe should we adopt a policy where we close PRs if it's not necessary for maintainers to look at it (so, open PR means, "Maintainer, look at me!" And closed PR means, "Don't look at me.")

23Skidoo:
I do want to get to Inbox 0 on Cabal PRs, but right now we're not there yet. Mostly because I don't want to close PRs that are (50/75/90 %)-complete because it makes easy to forget about them.

PR #2993, opened 2015-12-24
Reexport exeExtension from System.Directory.

ezyang:


23Skidoo:
`exeExtension` was added in 1.2.4.0, so you need a version check.

ezyang:
Ugh, nevermind then. (Note to self: submit doc update.)

Issue #2994, opened 2015-12-25
In principle, Setup --builddir=/path/to/dist should work even if CWD isn't in /path/to


ezyang:
Right now it doesn't. Would be as easy as saving the current working directory in LBI and changeDirectory'ing to it first before doing anything.

23Skidoo:
I guess we can also use `getExecutablePath`? Not sure if it works with `runghc`, though.

ezyang:
It doesn't, and if you're using the default setup executable it won't even give you the right directory.

bgamari:
Note that f907f1a6b4241ab4b75eaafa2ffb9031de8f008a caused a regression in GHC, [#11326](https://ghc.haskell.org/trac/ghc/ticket/11326).

ezyang:
OK, let's revert this and add a new field to the LBI that actually contains the absolute path.

ezyang:
Recording the absolute directory isn't a great idea because then you can't move the build directory somewhere else and expect it to work.

23Skidoo:
> Recording the absolute directory isn't a great idea because then you can't move the build directory somewhere else and expect it to work.

This is the reason sandboxes are not relocatable  (#2302).

PR #2995, opened 2015-12-25
Canonicalize path to Cabal file, partially towards #2994.

ezyang:
The actual reason I want this is because it gives me a reliable way to find the test files for the Cabal test suite. An alternate thing that would work is to store the CWD in the LBI when we make it.

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -107,7 +107,8 @@ import Distribution.Text
 -- Base
 import System.Environment(getArgs, getProgName)
 import System.Directory(removeFile, doesFileExist,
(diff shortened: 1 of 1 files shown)


BardurArantsson:
Certainly looks reasonable to me. One potential worry here might changing behavior in cases of non-existent directores and symlinks (where ISTR canonicalizePath being rather inconcsistent?). Another one would be Custom-type builds that rely on this, but that seems pretty far-fetched.

I'm not sure I'm qualified to think through all the potential consequences and to just merge this, but +1 review from me.

ezyang:
If we are worried about BC, then we can simply put the current directory in a new field in the LBI.

PR #2998, opened 2015-12-27
Rewrite the package test suite

ezyang:
I've rewritten the test suite to be more concise and "correct by
construction".  The primary method that this is achieved by
is introducing the 'TestM' monad, which carries around the
important state for the tests so that (1) we don't have to
pass it as an argument all around, and (2) we can automatically
make the correct decisions about how to do things.  This new
method emphasises "configuration by convention": we assume
that a test-case named "Foo" has its test packages in the
directory "tests/PackageTests/Foo".

A secondary change is that all command functions automatically fail if
they have a non-zero exit code (unless you use the 'shouldFail'
combinator which inverts the sense.)  This saves a lot of typing
on test-cases.  (In fact, I've reorganized all of the commands
related here.)

In the process, I've tightened up the logic for how to find the
LocalBuildInfo of the Cabal we've testing, so we should now
reliably be testing the inplace Cabal library, and not the
system library (as was often the case.)

Because things are a lot shorter, there is no good reason to
make Check modules except for the biggest test cases.  Most
test-cases have been folded into PackageTests.Tests; if you
have a small test-case you should just put it there.


Cabal/Cabal.cabal
```diff
@@ -341,16 +314,16 @@ test-suite package-tests
     tasty-quickcheck,
     tasty-hunit,
     QuickCheck >= 2.1.0.1 && < 2.9,
+    transformers,
```
BardurArantsson: Wouldn't this require changes in bootstrap scripts...? Or perhaps it doesn't when it's just "test" deps?

Cabal/tests/PackageTests/DeterministicAr/Check.hs
```diff
@@ -3,51 +3,29 @@
 
 module PackageTests.DeterministicAr.Check where
 
+
```
BardurArantsson: Spurious whitespace

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
-             deriving (Eq, Show)
-
-data Result = Result
-    { successful :: Bool
-    , success    :: Success
-    , outputText :: String
-    } deriving Show
+-- TODO: remove me
+inplaceDB :: SuiteConfig -> FilePath
+inplaceDB suite = cabalDistPref suite </> "package.conf.inplace"
```
BardurArantsson: *Can* this be removed now?

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+-- this tests; usually @Check.hs@ and any test packages.
+testDir :: TestM FilePath
+testDir = do
+    (suite, test) <- ask
+    return $ absoluteCWD suite </> "tests/PackageTests" </> testMainName test
+
+-- | The absolute path to the root of the package directory; it's
+-- where the Cabal file lives.  This is what you want the CWD of cabal
+-- calls to be.
+-- TODO: arguably this should not take a 'PackageSpec' argument
```
BardurArantsson: It *doesn't* take a PackageSpec argument any more :)

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+    prefix_dir <- prefixDir
+    when ((cmd == "register" || cmd == "copy") && not (testPackageDb test)) $
+        error "Cannot register/copy without using 'withPackageDb'"
+    let extraArgs1 = case cmd of
+            "configure" ->
+                -- If the package database is empty, setting --global
+                -- here will make us error loudly if we try to install
+                -- into a bad place.
+                [ "--global"
+                -- TODO: Should be configurable, but it's tricky:
```
BardurArantsson: What does this TODO apply to? The "--with-ghc" bit seems to actually be configurable, so...

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+rawCompileSetup :: Verbosity -> SuiteConfig -> [(String, Maybe String)] -> FilePath -> IO ()
+rawCompileSetup verbosity suite e path = do
+    r <- rawRun verbosity (Just path) (ghcPath suite) e $
+        [ "--make"] ++
+        ghcPackageDBParams (packageDBStack suite) ++
+        [ "-hide-all-packages"
+        , "-package base"
+        -- This is better, but we don't necessarily have it
+        -- if we're bootstrapping with old Cabal.
+        -- , "-package-id " ++ CURRENT_PACKAGE_KEY
```
BardurArantsson: Is bootstrapping with old Cabal a problem when looking at the *tests*? (I mean, if you want to do development, you could always go "old Cabal -> new Cabal -> new Cabal w/tests", right?)

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+        -- GHC doesn't seem to necessarily reliably determine
+        -- when to rebuild Setup, so this is safest.
+        , "-fforce-recomp"
+        , "-O0"
+        , "Setup.hs" ]
+    unless (resultExitCode r == ExitSuccess) $
+        error $
+        "could not build shared Setup executable\n" ++
+        "  ran: " ++ resultCommand r ++ "\n" ++
+        "  output:\n" ++ resultOutput r
```
BardurArantsson: This is not a new thing, but I think readability could be improved by adding a few more newlines (in the output!) around the resultOutput bit

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+        error "Must initialize package database using withPackageDb"
+    let db_stack = packageDBStack config ++ [SpecificPackageDB db_path]
+        extraArgs = ghcPkgPackageDBParams db_stack
+    run Nothing (ghcPkgPath config) (cmd : extraArgs ++ args)
+
+ghcPkgPackageDBParams :: PackageDBStack -> [String]
+ghcPkgPackageDBParams dbs = concatMap convert dbs where
+    convert :: PackageDB -> [String]
+    -- TODO: this is a bit dodgy but there's no way good
+    -- way to handle this case in ghc-pkg
```
BardurArantsson: Just "There's no good way to..." would be sufficient, IMO.

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
-            compileSetup config (directory spec)
-            path <- canonicalizePath $ directory spec </> "Setup"
-            run (Just $ directory spec) path envOverrides cabalArgs
-        else do
-            -- Use shared Setup executable (only for Simple build types).
-            path <- canonicalizePath "Setup"
-            run (Just $ directory spec) path envOverrides cabalArgs
+-- | Running an executable that was produced by cabal.
+runExe :: String -> [String] -> TestM Result
+runExe exe_name args = do
```
BardurArantsson: The naming convention seems to be a bit... inconsistent: runExe here, cabal_build above, etc. Perhaps it would be worth trying to bring it into line while you're rewriting everything anyway?

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
 
-    -- fork off a thread to start consuming the output
-    out <- suckH [] readh
+    out <- hGetContents readh
+    evaluate (length out) -- force the output
     hClose readh
 
     -- wait for the program to terminate
     exitcode <- waitForProcess pid
+    -- TODO: suboptimal, not copypastable
```
BardurArantsson: Meh, remove TODO.

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
-assertRegisterFailed :: Result -> Assertion
-assertRegisterFailed result = when (successful result) $
-    assertFailure $
-    "expected: \'setup register\' should fail\n" ++
-    "  output: " ++ outputText result
+record :: Result -> TestM ()
+record res = do
+    build_dir <- distDir
+    liftIO $ createDirectoryIfMissing True build_dir
+    -- TODO: make a runnable shell script
```
BardurArantsson: Seems big enought and complex enough to be worthy of a GH-issue linked to this one? (Though, if not sure how feasible it *actually* is -- it might need to reset a lot of ambient state if it's going to be robust.)

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+                        , resultOutput = output } = do
+    (_, test) <- ask
+    let lastN n = unlines . reverse . take n . reverse . lines
+    when (exitCode /= ExitSuccess && not (testShouldFail test)) $
+        assertFailure $ "Command " ++ cmd ++ " failed.\n" ++
+        "last 5 lines of output (see test.log for full):\n" ++ lastN 5 output
+    when (exitCode == ExitSuccess && testShouldFail test) $
+        assertFailure $ "Command " ++ cmd ++ " succeeded.\n" ++
+        "last 5 lines of output (see test.log for full):\n" ++ lastN 5 output
+    -- TODO: breadcrumb for test.log for full output
```
BardurArantsson: Just show all the output. The last N lines are *almost always* the *most* uninteresting. (We've removed the "tail N" effect in regular builds... unless it's been re-enabled since I removed it ~6-12 months or so ago.)

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -1,367 +1,540 @@
...
+
+whenGhcVersion :: (Version -> Bool) -> TestM () -> TestM ()
+whenGhcVersion p m = do
+    (suite, _) <- ask
+    when (p (ghcVersion suite)) m
+
+withPackage :: FilePath -> TestM a -> TestM a
+withPackage f = withReaderT (\(suite, test) -> (suite, test { testCurrentPackage = f }))
+
+-- TODO: Really should accumulate... but I think to do this
```
BardurArantsson: I think this should be added before merge, or the comment should be removed. Doing this post-merge would change the semantics of withEnv rather subtly -- now it might not matter greatly, but then again it might.

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -0,0 +1,157 @@
...
+         . tcs "Deadlock" $ do
+            cabal_build ["--enable-tests"]
+            shouldFail $ cabal "test" []
+         ]
+    ]
+
+  -- Inline tests
+  , tc "BenchmarkExeV10" $ cabal_build ["--enable-benchmarks"]
+
+  , tc "BenchmarkOptions" $ do
```
BardurArantsson: One minor complaint: It's a bit unclear what the tests are actually trying to test. E.g. is this test just trying to see if cabal accepts the options without failing badly? Is it trying to test whether benchmarks executables are actually produced? (etc.)

Clarifying comments would suffice, I think, but personally, I'm a big fan of doing a little more formal "setup/exercise/verify" stanzas. See for example the first bit of example code on http://xunitpatterns.com/Four%20Phase%20Test.html for the gist of the idea. Doing this slightly more formally with "setup", "exercise", "verify" combinators would be trivial. *Rationale:* I find that doing this formally (or at least semi-formally using always-identical comments) tends to make it a lot easier to figure out unfamiliar failing tests.

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -0,0 +1,157 @@
...
+
+  , tc "Haddock" $ do
+      dist_dir <- distDir
+      let haddocksDir = dist_dir </> "doc" </> "html" </> "Haddock"
+      cabal "configure" []
+      cabal "haddock" []
+      let docFiles
+              = map (haddocksDir </>)
+                    ["CPP.html", "Literate.html", "NoCPP.html", "Simple.html"]
+      -- TODO: refactor me
```
BardurArantsson: This TODO is just weird. Remove.

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -0,0 +1,157 @@
...
+
+  , tc "UniqueIPID" . withPackageDb $ do
+      withPackage "P1" $ cabal "configure" []
+      withPackage "P2" $ cabal "configure" []
+      withPackage "P1" $ cabal "build" []
+      withPackage "P1" $ cabal "build" [] -- rebuild should work
+      r1 <- withPackage "P1" $ cabal "register" ["--print-ipid", "--inplace"]
+      withPackage "P2" $ cabal "build" []
+      r2 <- withPackage "P2" $ cabal "register" ["--print-ipid", "--inplace"]
+      -- TODO: hella janky
```
BardurArantsson: You may want to golf it, but it's *test code* and it's reasonably *understandable*, so remove the TODO. If someone wants to golf it or extract some common logic at some point, they will...

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -0,0 +1,157 @@
...
+        assertEqual
+            ("executable should have linked with the " ++ expect ++ " library")
+            ("myLibFunc " ++ expect)
+            (concat $ lines (resultOutput r))
+
+    tc :: FilePath -> TestM a -> TestTree
+    tc name m
+        = testCase name (runTestM config name Nothing m)
+
+-- TODO: get rid of me
```
BardurArantsson: Remove TODO

BardurArantsson:
This looks like a huge win for readability to me[1], so definitely a qualified +1 (see review comments) from me, FWIW.

[1] In fact, I kind of wish for something similar for cabal-install now so one can avoid shell script! :) Of course being able to re-run changed tests without compiling is nice for cabal-install, so...

Cabal/Cabal.cabal
ezyang: It's fine because it's a test dep. Otherwise the deps on tasty/co would be unacceptable.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: ghcPath and ghcPkgPath always come from Cabal's LBI, so in that sense it's not configurable. I've moved this comment somewhere more reasonable.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: You can do that, and that's what we used to do. As before, the problem is that if you do this, you need to configure Cabal twice (once with the old Cabal to actually build it with old Cabal, once with the new Cabal for the test suite to open it with LBI. There's a long comment in PackageTests.hs explaining the situation.)

Cabal/tests/PackageTests/PackageTester.hs
ezyang: OK I added one.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: A better name for runExe would be OK. I actually like the distinction between cabal "build" and cabal_build; cabal "build" actually means, "run the build command with Cabal", while cabal_build is a mini-recipe which says, "compile, then build". runExe is not a great name but I couldn't think of a better one.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: Actually pretty easy to fix, I just fixed it.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: In that case, we need some workaround to deal with the ExeV10 tests, which print a thousand lines of "The quick brown fox..." to test whether or not buffer is flushed correctly.

BardurArantsson: We do? Doesn't the output only fire on failure? Sure, it'll be a bit annoying, but it's a *lot* better to have annoying output for *one* test rather than annoying output for *all-but-one* test :), i.e. the ones that are more likely to fail.

(I think most terminals these days should have several thousands of lines scroll-back by default anyway?)

Cabal/tests/PackageTests/PackageTester.hs
ezyang: OK I just implemented it.

Cabal/tests/PackageTests/PackageTester.hs
ezyang: It does only fire on failure. But the hpc tests are touchy and I accidentally broke them for a while :) And there's like, 16 of them because there are 16 different configurations for it. But OK, the test passes now :)

Cabal/tests/PackageTests/PackageTester.hs
ezyang: It's OK, notice that it errors if testEnvironment is non-null. I'm happy making the first person who needs this implement it :)

Cabal/tests/PackageTests/Tests.hs
ezyang: I think we should leave this for another PR; when I originally decided to port these tests, my plan was to shorten the tests but otherwise not change them at all. Most of the tests that call cabal_build and then some commands originally (and still) were *solely* testing if ./Setup ran without exiting non-zero.

Actually, originally, the test suite was a bit more like you suggested: you had to explicitly assert that a given command had succeeded or failed. I hated this, because there were lots of tests which outright FORGOT to test the exit code, and it was a pain for "setup" operations, where you implicitly did not want it to fail.

I'd like to avoid boilerplate. If by and large most tests are `successfulExitOnly` and `noSetup`, can we just keep that implicit?

BardurArantsson: Sure, and I definitely agree that it should be hard to "do the wrong thing" (e.g. forget to test exit status). There are certain things that should always be "implicitly-check-this-unless-told-otherwise" and running ./Setup would definitely be one of those, I think.

It can definitely wait since it requires quite a lot of thought, but it'd be nice to have some sort of comment for each test, just briefly explaining what it's supposed to be testing. (I guess this may not actually be knowable in all cases, though...)

ezyang:
OK, I addressed all comments besides the splitting tests into setup/run/verify phases. Still need to fix build on old version of GHC.

[1] Shake was the project that convinced me that shell invocations from GHC could be reasonably readable. They actually have even more type-class magic to make "cmd" invocations even more convenient, but I didn't go that far. Rebuilds suck, but the automatically generated `test.sh` files should make the rebuild/test cycle more pleasant.

BardurArantsson:
Your changes look great. Enthusiatic :+1: from me!

Cabal/tests/PackageTests/Tests.hs
ezyang: Oh! This makes sense, because the bulk of the test is in the input data. I can go ahead and add some comments summarizing what's going on in the Cabal files.

ezyang:
Successfully greened. I rebased and after CI approves I'll push.

23Skidoo:
Good stuff.

> 1,008 additions and 1,307 deletions. 

:+1

PR #3002, opened 2015-12-28
Two minor package test fixes

erikd:


ezyang:
Thank you!

We should merge the first patch ASAP, but as for the second, my intent was to disable the unused return value warning when I added `-fno-warn-unused-do-bind` and `-fno-warn-incomplete-patterns` to the Cabal file. If we insist on not squelching these warnings, then my preference is to change the `cabal` commands to return unit, and add `cabal_` variants which return (or maybe we can name them something different)

erikd:
I build cabal in my Jenkins instance using `cabal build --ghc-option=-Wall --ghc-option=-Werror` which seems to override the warning suppressions you wanted.

However, I agree having two versions `cabal` and `cabal_` one of which returns unit is probably a better idea.

BardurArantsson:
The "-fno-warn-incomplete-patterns" one is definitely something I'd want removed just as a safety precaution when expanding data type definitions. I also don't quite see the necessity of it.

BardurArantsson:
Cherry-picked the first commit into master. Checking build status now...

23Skidoo:
> If we insist on not squelching these warnings, then my preference is to change the `cabal` commands to return unit, and add `cabal_` variants which return (or maybe we can name them something different)

I like this better than using `-fno-warn-incomplete-patterns`.

ezyang:
Do people have preferences about which of `cabal` and `cabal_` return unit? I'd like `cabal` to return unit since that's the most commonly used one, but `mapM_` and co set a precedent it should go the other way.

BardurArantsson:
Even if it's a bit annoying to have the common one be the harder-to-type one, I'd definitely prefer to follow the forM_ (et al.) convention here.

ezyang:
How about cabal and cabal' (prime meaning returns something)?

23Skidoo:
> How about cabal and cabal' (prime meaning returns something)?

Fine with me.

ezyang:
Subsumed in favor of #3008.

Issue #3004, opened 2015-12-28
Incorrect special character handling in Haddock response files


snoyberg:
This has been causing us quite some trouble with our Stackage builds (see: https://github.com/fpco/stackage/issues/1018). I recommend considering this a major issue worthy of an emergency patch, since the following combination seems to always cause failed Haddock builds:

* Newest released Cabal (1.22.6.0)
* Upcoming Haddock release (> 2.16.1)
* Any cabal file with a newline in the synopsis

When generating a response file for the Haddock arguments, the current code in Cabal simply calls `hPutStrLn` on the argument without any escaping. When an argument has a newline (like the synopsis), anything after the newline is treated as its own argument.

Solution: use response file escaping to pass in all arguments. As an example, that's how I implemented response file support in GHC, see: https://github.com/ghc/ghc/commit/296bc70b5ff6c853f2782e9ec5aa47a52110345e#diff-782eb6a039b579ec019e5a4009f45058R1281

ezyang:
Reproduced.

ezyang:
If I understand correctly, we can't fix this until https://github.com/haskell/haddock/issues/379 is fixed, because Haddock does not yet support the standard GCC response file format.

ezyang:
If you like, we can disable response files for now until Haddock fixes this.

23Skidoo:
Yes, needs to be fixed in Haddock first.

ezyang:
The relevant code to fix is Distribution.Simple.Haddock, in renderArgs. Here's a hokey patch that doesn't work:

{{{
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -497,7 +497,7 @@ renderArgs verbosity tmpFileOpts version comp args k = do
                  withTempFileEx tmpFileOpts outputDir "haddock-response.txt" $
                     \responseFileName hf -> do
                          when haddockSupportsUTF8 (hSetEncoding hf utf8)
-                         mapM_ (hPutStrLn hf) renderedArgs
+                         mapM_ (hPutStrLn hf . escape) renderedArgs
                          hClose hf
                          let respFile = "@" ++ responseFileName
                          k ([respFile], result)
@@ -515,6 +515,19 @@ renderArgs verbosity tmpFileOpts version comp args k = do
               pkgstr = display $ packageName pkgid
               pkgid = arg argPackageName
       arg f = fromFlag $ f args
[...]

PR #3005, opened 2015-12-28
Failing test for #3004.

ezyang:
Putting it here so we don't lose it. #3004


ezyang:
(Haha, I like how Travis passes this, because there's no recent enough GHC.)

23Skidoo:
There is, but it is marked with `allow_failures`.

PR #3006, opened 2015-12-28
Make misc/gen-extra-source-files.sh edit the Cabal.cabal file directly.

ezyang:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ dist-*
 /Cabal/dist/
 /Cabal/tests/Setup
 /Cabal/Setup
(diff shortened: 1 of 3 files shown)


23Skidoo:
For bonus points this you can also add something like this to `.travis.yml`:

```
./Cabal/misc/gen-extra-source-files.sh
[[ `git diff | wc -l` == 0 ]]
```

ezyang:
Haha, even better would be to do the tests out of the sdist (right now we only build, not test.)

ezyang:
I'm puzzled why GitHub doesn't think this is merged.

PR #3007, opened 2015-12-28
Travis no bootstrap

ezyang:
Want to see if Travis will accept this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ dist-*
 /Cabal/dist/
 /Cabal/tests/Setup
 /Cabal/Setup
+/Cabal/misc/source-file-list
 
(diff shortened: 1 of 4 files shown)


ezyang:
OK, right, this doesn't work because package-tests reads the LBI, and it's just not recent enough. (This should affect people who do dev work on Cabal... I always run a very recent Cabal so it's not affected me.) I'll have to think about what to do about this.

23Skidoo:
One nice thing about our current Travis setup is that it tests that bootstrapping still works (even though it's not enabled by default).

ezyang:
Yeah, I decided it's not worth trying to work around this, and the bootstrapping test is useful.

PR #3009, opened 2015-12-29
Run gen-extra-source-files.sh in Travis.

ezyang:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ script:
  - cd Cabal
 # Test if gen-extra-source-files.sh was run recently enough
  - ./misc/gen-extra-source-files.sh
- - ! git diff-files -p --exit-code
+ - ./misc/travis-diff-files.sh
  - mkdir -p ./dist/setup
  - cp Setup.hs ./dist/setup/setup.hs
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -33,6 +33,9 @@ script:
 
 # Cabal
  - cd Cabal
+# Test if gen-extra-source-files.sh was run recently enough
+ - ./misc/gen-extra-source-files.sh
+ - git diff-files
```
23Skidoo: You need to run `diff-files` with `--exit-code`, otherwise it always returns 0:

```
$ rm LICENSE    
$ git diff-files
:100644 000000 489b8a152a9aba3b4700d6374533590c01c259b3 0000000000000000000000000000000000000000 D	LICENSE
$ echo $?       
0
$ git diff-files --exit-code
:100644 000000 489b8a152a9aba3b4700d6374533590c01c259b3 0000000000000000000000000000000000000000 D	LICENSE
$ echo $?                   
1
```

ezyang:
Oh OK, so it looks like this test is NOT working; there is something out-of-sync. Maybe a diff would be more useful.

PR #3012, opened 2015-12-30
The Cabal part for fully gcc-like response files

randen:
  * VERY IMPORTANT! this feature depends on a particular
    version of haddock, that includes the corresponding
    support for these style of response files.  However,
    we do not really know that version number!  So, this
    change here must be co-ordinated with the haddock project.
This is intended to fix haskell/cabal#3004. The other half is haskell/haddock#379.

### DIFF
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -87,7 +87,7 @@ import Language.Haskell.Extension
 import Control.Monad    ( when, forM_ )
 import Data.Char        ( isSpace )
 import Data.Either      ( rights )
-import Data.Foldable    ( traverse_ )
+import Data.Foldable    ( traverse_, foldl' )
 import Data.Maybe       ( fromMaybe, listToMaybe )
 
 import System.Directory (doesFileExist)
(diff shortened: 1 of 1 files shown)


23Skidoo:
Looks like you've forgotten an import:

```
Distribution/Simple/Haddock.hs:531:29:
    Not in scope: ‘foldl'’
    Perhaps you meant one of these:
      ‘foldl’ (imported from Prelude), ‘foldl1’ (imported from Prelude),
      ‘foldr’ (imported from Prelude)
```

>  However, we do not really know that version number! 

OK, I'll wait until the Haddock patch is merged. And I guess this should also go into 1.22?

bgamari:
I believe this should appear in Haddock 2.17.

randen:
@bgamari , should I update this pull request with >= 2.17 for the supporting Haddock version?

bgamari:
@23Skidoo I have merged the Haddock portion locally. Could you merge this, bumping the haddock version bound appropriately?

bgamari:
I have merged the Haddock portion to both Haddock's `v2.16` and `master` branches. While I don't expect to release haddock-2.16.2 myself, this means that the bound currently in the Cabal pull request is nevertheless accurate.

ezyang:
Don't forget to merge this with #3005.

PR #3039, opened 2016-01-13
Tests for #2731 (Ignore dependencies that are not Buildable)

grayjay:
I added a package test for the Cabal component and several unit tests for the solver component.  This PR also resolves a conflict between #2731 and #2732.  I had to make a few changes to the solver DSL:

* Allow a flag to be used multiple times in a package.  Reusing a flag previously caused this error:

        Exception: internal error: could not construct a valid install plan.
        The proposed (invalid) plan contained the following problems:
        Package B-1.0.0 has an invalid configuration, in particular:
          missing an assignment for the flag: my-flag

* Represent the Buildable field.  Now both sides of a conditional can be either not buildable, or buildable with a list of dependencies.
* Allow a compiler to support zero languages or extensions.  I changed the signature of `DSL.exResolve` to distinguish between not checking languages/extensions and not allowing any languages/extensions.

I thought of two other improvements, but I left them for separate issues or PRs.  I didn't see an easy way to check flag choices directly in the DSL.  I instead had to check the dependencies that the flags introduced.  I also noticed that the default flag value is false in the DSL and true in cabal-install/Cabal.  Some of these tests rely on the default, so it should probably be consistent.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -77,6 +77,8 @@ extra-source-files:
   tests/PackageTests/BuildDeps/TargetSpecificDeps3/lemon.hs
   tests/PackageTests/BuildDeps/TargetSpecificDeps3/my.cabal
   tests/PackageTests/BuildTestSuiteDetailedV09/Dummy2.hs
+  tests/PackageTests/BuildableField/BuildableField.cabal
+  tests/PackageTests/BuildableField/Main.hs
   tests/PackageTests/CMain/Bar.hs
   tests/PackageTests/CMain/foo.c
   tests/PackageTests/CMain/my.cabal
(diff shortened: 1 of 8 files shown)


BardurArantsson:
LGTM. Have to run tests, etc. for each individual commit?

23Skidoo:
/cc @kosmikus 

I think this can be merged now.

grayjay:
@BardurArantsson I tested each commit with GHC 7.10.2, but I can push the commits one at a time to build them with Travis.  I'd like to make sure that the two commits from #2731 build.

BardurArantsson:
Fair enough, just checking :). I wonder if there's something that could be done to get Travis to always build each commit in a series...

ezyang:
https://docs.travis-ci.com/user/triggering-builds/ might be of help.

BardurArantsson:
I was hoping for something which could be set up once and for all for the whole repository.

(Oh, how I long for a proper review + CI system...)

ezyang:
Well, we do have a Phabricator instance :) (Though I don't think builds are setup.)

BardurArantsson:
No direct experience with Phab, but doesn't it require some sort of weird custom command line tool instead of just using standard git commands? (When reviewing and pushing patches to review, obviously.)

Anyway, this is probably enough off-topic spam for this thread...

BardurArantsson:
Looks like they all succeeded. Merged

grayjay:
Thanks!

PR #3047, opened 2016-01-15
Distinguish between component ID and unit ID.

ezyang:
GHC 8.0 is switching the state sponsored way to specify
linker names from -this-package-key to -this-unit-id, so
it behooves us to use the right one.  But it didn't make
much sense to pass ComponentIds to a flag named UnitId,
so I went ahead and finished a (planned) refactoring
to distinguish ComponentIds from UnitIds.

At the moment, there is NO difference between a ComponentId
and a UnitId; they are identical.  But semantically, a
component ID records what sources/flags we chose (giving us enough
information to typecheck a package), whereas a unit ID records
the component ID as well as how holes were instantiated
(giving us enough information to build it.)  MOST code
in the Cabal library wants unit IDs, but there are a few
places (macros and configuration) where we really do
want a component ID.

Some other refactorings that got caught up in here:

    - I axed IPI641, fixing #3034.

    - Removed some defunct Bacpack fields from ExposedModule
      (GHC doesn't use them.)

    - Removed instantiatedWith (also unused) and all of the
      relevant code from configure (it will need to be rewritten
      in any case.)

    - Changed the type of componentCompatPackageKey to String, reflecting the
      fact that it's not truly a UnitId or ComponentId.

    - Changed the behavior of CURRENT_PACKAGE_KEY to unconditionally
      give the compatibility package key, which is actually what you
      want if you're using it for the template Haskell trick.  I also
      added a CURRENT_COMPONENT_ID macro for the actual component ID,
      which is something that the Cabal test-suite will find useful.

    - Added the correct feature test for GHC 8.0 ("Uses unit IDs").


23Skidoo:
Do you want this to go into 1.24?

ezyang:
Yes, we'll release GHC 8.0 with it.

ezyang:
Note: Cabal library changes should be good, but I need to work out some details for cabal-install. Will update shortly.

ezyang:
Should be ready.

BardurArantsson:
I started to review this, but realized just how big it is[1]... any chance it could be split up a *little* bit? (The removal of IPI641 would seem like a reasonable candidate at least.)

I realize that you've probably already tried to, but it can't hurt to ask...

[1] I probably couldn't contribute much in-depth review commentary anyway, but...

ezyang:
I separated out some bits but I'm not sure it helps very much. Not sure if the intermediate commits build.

BardurArantsson:
I'm not sure if this is general policy for this project, but would you mind triggering a build for each of them? Alternatively I suppose it might be possible to just create a PR for each? (I realize that this is annoying, but I guess we'll have to like with it if we want to make sure commits build individually.)

(It can get really annoying when bisecting that a really big commit next to the prime suspect doesn't build.)

BardurArantsson:
... and actually, thinking about it a little more: It might be fine to actually *NOT* bother trying to build but to just leave this as four commits until review is done and then to *re-squash* before merge. That way there aren't any weird compatibility effects either, etc.

BardurArantsson:
Well, the three "Remove ..." commits certainly LGTM. I don't have time to review the really big one right now, but will try to do it a little later today.

BardurArantsson:
(And I may not be the best to review this, so I'd encourage others to review too...)

BardurArantsson:
(Mark. End of review)

23Skidoo:
> I'm not sure if this is general policy for this project, but would you mind triggering a build for each of them?

We don't have a formal policy, but we've never been strict about this. Bisect allows ignoring broken commits, so personally I'm okay with imposing this cost on the person doing bisect instead of each and every contributor. Don't see any other disadvantages besides slightly worse bisect experience to having broken commits in history.

> It might be fine to actually NOT bother trying to build but to just leave this as four commits until review is done and then to re-squash before merge.

Personally I prefer a history composed of smaller self-contained commits instead of large monolithic ones.

> Alternatively I suppose it might be possible to just create a PR for each?

Please don't do that, related changes should be grouped together.

BardurArantsson:
OK. I don't necessarily agree, but it's probably not worth fighting over :).

BardurArantsson:
Btw, if you use the setup-dev.sh script it should be pretty easy to check if things at least build. (There's a caveat in that tests are currently not *run* if you do this, but they should be built.)

23Skidoo:
LGTM, let's merge.

PR #3052, opened 2016-01-15
allow building with recent (master) HTTP

Heather:
I need it for GHC 8 support


23Skidoo:
Merged, thanks!

Heather:
@23Skidoo but so far changes for GHC 8 is not accepted in HTTP library

23Skidoo:
Well, we can't force Ganesh to accept that patch.

Heather:
@23Skidoo I don't want to just provided tracking information

PR #3057, opened 2016-01-16
Fix typos in solver test case comment.

grayjay:


23Skidoo:
Tip: when doing comment-only changes you can use `[ci skip]` in the commit message body to avoid triggering builds on Travis.

grayjay:
Thanks, I'll keep that in mind.

PR #3058, opened 2016-01-16
Change default flag value to True in solver DSL.

grayjay:
This default is consistent with Cabal.

I updated the test cases from #3039 that rely on the default.  As far as I can tell, the other four tests that use flags are unaffected.  We could also add the ability to set defaults for individual flags.  That would be a bigger change, though, because it would require flags to be declared before they are used.


23Skidoo:
Merged, thanks!

PR #3060, opened 2016-01-17
Enable tests in setup-dev.sh

BardurArantsson:
It seems there are no longer any spurious failures, so we
can enable tests in setup-dev.sh


BardurArantsson:
This is low-impact enough that I'm just going to go ahead and merge it myself.

PR #3063, opened 2016-01-19
Make configIPID a bit more type-safe, fix a bug.

ezyang:
Previously, configIPID was a String, which made it a bit confusing
whether or not it was a ComponentId or a UnitId.  I've decided (for now)
it should be a UnitId, for consistency with the existing backwards
compatibility InstalledPackageId, but this might have to change in the
future (I'm not sure, I have to work out which ends up being better.)
Since right now the string representations of UnitId and ComponentId are
interchangeable, it should be harmless to conflate the two.


ezyang:
Actually, scratch this, I think it should be a `ComponentId`, not a `UnitId`.

PR #3066, opened 2016-01-19
Per component copy

### LINKED ISSUE #2780: copy command fails when building only some components
Actions:
1. Create a project with both a library and executable
2. `cabal clean`
3. `cabal configure`
4. `cabal build lib:packagename`
5. `cabal copy`

Expected: command succeeds, copying over the library only, and not the test suite

Actual: errors out since the executable does not exist, e.g.:

```
Installing library in
/home/vagrant/.cabal/lib/x86_64-linux-ghc-7.10.2/tmp-0.1.0.0-JdaAODhQa7p5WhxMngbYYt
Installing executable(s) in /home/vagrant/.cabal/bin
Setup.hs: dist/build/tmp-exe/tmp-exe: does not exist
```

This problem could be mitigated by allowing `copy` to take a list of components like `build` does.

ezyang:
Accept components to copy in ./Setup copy, fixes #2780. 

This is based off of #3022 (which has a very useful refactoring.) Only the last commit is relevant.


ezyang:
CC @snoyberg, @mgsloan, @LukeHoersten

LukeHoersten:
@ezyang thank you so much!

ezyang:
(If someone wants this in 1.24 they'll have to backport it past #2780. It won't be trivial, but maybe it won't be too hard either.)

ezyang:
This got folded into #3022.

PR #3070, opened 2016-01-23
Add some more backwards-compatibility functions, fixes #3068.

### LINKED ISSUE #3068: installedPackageId vanished
`Distribution.Package.installedPackageId` and `Distribution.InstalledPackageInfo.installedPackageId` no longer exist. It turns out that these are used by a number of packages. I seem to recall there were more, but all I can find at the moment is `packunused` (see https://github.com/hvr/packunused/blob/master/packunused.hs#L12)

ezyang:


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

Issue #3003, opened 2015-12-28
Cabal-1.22.6.0 incompatible with binary-0.8


hvr:
Here's what happens if you have `binary-0.8` installed (this was with GHC 7.10.3):

```
$ cabal install Cabal
Resolving dependencies...
[ 1 of 79] Compiling Distribution.GetOpt ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/GetOpt.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/GetOpt.o )
[ 2 of 79] Compiling Distribution.Compat.Binary ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/Compat/Binary.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/Compat/Binary.o )
[ 3 of 79] Compiling Distribution.TestSuite ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/TestSuite.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/TestSuite.o )
[ 4 of 79] Compiling Distribution.Simple.PreProcess.Unlit ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/Simple/PreProcess/Unlit.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/Simple/PreProcess/Unlit.o )
[ 5 of 79] Compiling Distribution.Simple.CCompiler ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/Simple/CCompiler.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/Simple/CCompiler.o )
[ 6 of 79] Compiling Distribution.PackageDescription.Utils ( /tmp/cabal-tmp-7012/Cabal-1.22.6.0/Distribution/PackageDescription/Utils.hs, /tmp/cabal-tmp-7012/Cabal-1.22.6.0/dist/setup/Distribution/PackageDescription/Utils.o )
[...]

PR #3076, opened 2016-01-26
Fix space leak in ./Setup configure

grayjay:
This commit fixes the space leak in #2777 without modifying the logic for choosing flags.  It is a lot simpler than #2925, though I think an improvement to the algorithm like in #2925 is still needed.  I tested the practicality of solving for more flags by running `./Setup configure` on git-annex with 3 of the 21 flags removed.  Most of the dependencies were missing, which meant that Setup had to try all 2^18 flag combinations.  It took 13 minutes.

/cc @joeyh

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Configuration.hs
+++ b/Cabal/src/Distribution/PackageDescription/Configuration.hs
@@ -176,8 +176,6 @@ instance Semigroup d => Semigroup (DepTestRslt d) where
     x     <> DepOk = x
     (MissingDeps d) <> (MissingDeps d') = MissingDeps (d <> d')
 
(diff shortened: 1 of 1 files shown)


23Skidoo:
Travis failure is unrelated to your change (see #3003).

joeyh:
Incidentally, git-annex HEAD is down to 16 flags;
under 10 minutes worst case for Setup configure is fine by me.

-- 
see shy jo

23Skidoo:
LGTM, merging. Sorry for taking so long!

grayjay:
@23Skidoo No problem. Thanks for merging it.

PR #3082, opened 2016-01-27
Improve algorithm for choosing flags with './Setup configure'

grayjay:
Cabal previously tried all flag combinations, which was very slow.  The new
algorithm assigns one flag at a time, and backtracks when a flag introduces a
dependency that is unavailable.

The new algorithm handles the Buildable field by adding an extra conditional at
the top level of each component that represents the condition for which the
component is buildable.  Since all dependencies go under the "then" branch, they
are only required when a flag choice makes the component buildable.  The buildable
logic is taken from the cabal-install dependency solver.

This commit also changes the error message that is shown when dependencies are
missing.  Previously, Cabal printed the shortest list of missing dependencies
from a single flag assignment.  Now it takes the union of all dependencies that
caused it to backtrack when trying different combinations of flags, which
requires less searching.

I updated #2925 to allow backtracking to work with the Buildable field.  I moved the `extractCondition` function (added in #2731) from the modular solver to `Distribution.PackageDescription.Configuration` so that it can be used by Cabal and cabal-install.  Since `extractCondition` is not a very general utility function, I was wondering if there is a better place to put it.

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Configuration.hs
+++ b/Cabal/src/Distribution/PackageDescription/Configuration.hs
@@ -22,6 +22,7 @@ module Distribution.PackageDescription.Configuration (
     -- Utils
     parseCondition,
     freeVars,
+    extractCondition,
     mapCondTree,
     mapTreeData,
     mapTreeConds,
@@ -45,6 +46,7 @@ import Data.Char ( isAlphaNum )
 import Data.Maybe ( mapMaybe, maybeToList )
 import Data.Map ( Map, fromListWith, toList )
 import qualified Data.Map as Map
+import Data.Tree ( Tree(Node) )
 
 ------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


23Skidoo:
Just checked that the `containers` test suite still works with this change.

grayjay:
@23Skidoo What is different in the containers test suite?

23Skidoo:
The `containers` test suite depends on `containers`, so the `cabal-install` solver fails to find a solution when you run `configure --enable-tests` because of circular dependencies. The `./Setup configure` solver works in this case, so it's important to make sure that it continues to work. See e.g. #1806, #1575, #935, #960.

23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

grayjay:
@23Skidoo Thanks.  I just realized that if I export `addBuildableCondition` instead of `extractCondition` from `Distribution.PackageDescription.Configuration`, I can reuse more code between cabal-install and Cabal.  I'll make another PR.  Sorry I didn't think of it sooner!

PR #3091, opened 2016-01-29
Refactor #3082 to reduce code duplication and improve Haddock comment.

grayjay:


Cabal/src/Distribution/PackageDescription/Configuration.hs
```diff
@@ -22,7 +22,7 @@ module Distribution.PackageDescription.Configuration (
     -- Utils
     parseCondition,
     freeVars,
-    extractCondition,
```
23Skidoo: Let's continue exporting `extractCondition`, just in case.

23Skidoo:
LGTM.

grayjay:
@23Skidoo I exported `extractCondition` again and added to its Haddock comment.

PR #3095, opened 2016-01-30
Add new 'user-config' subcommand 'init'.

23Skidoo:
An updated version of #2640.


grayjay:
I noticed that `init` doesn't use the path specified by CABAL_CONFIG.  I can submit a pull request to make it consistent with the other `user-config` commands.

PR #3097, opened 2016-01-31
Allow 'user-config init' to use CABAL_CONFIG.

grayjay:


ezyang:
Tests?

grayjay:
Sure.  I'll add an integration test.

grayjay:
@ezyang I saw that the integration tests already set the config file using `--config-file`.  `IntegrationTests.hs` sets CABAL_ARGS to `--config-file=config-file`, and then `common.sh` gives that argument to cabal.  The argument interferes with this test because it takes precedence over the environment variable.  I dealt with it by renaming CABAL_ARGS to CABAL_CONFIG_FILE_ARG and then leaving it out of `common.sh` for the user-config tests.  Does that sound reasonable?  I wasn't sure what other arguments might need to be shared between tests.

ezyang:
Hmm, in principle one could add extra args to the set (much like TEST_HC_OPTS). Interestingly, in GHC, there is an extra set of args for when you need something tweaked, TEST_HC_OPTS_NORECOMP. Maybe we should follow that naming convention? I definitely that this sort of thing is tricky to test, let's not get too hung up on it unless it's hurting someone.

grayjay:
How is TEST_HC_OPTS_NORECOMP used?

ezyang:
Essentially, when a test wants to exercise the recompilation facilities (most tests don't, and always want to do a full run) it just specifies TEST_HC_OPTS_NORECOMP so that the `-fforce-recomp` flag isn't passed.

grayjay:
That makes sense.  So we could specify different sets of arguments in `IntegrationTests.hs`, if we wanted to add more.

grayjay:
I'll undo the rename and use CABAL_ARGS_NO_CONFIG_FILE for the user-config tests.

cabal-install/cabal-install.cabal
```diff
@@ -58,6 +58,16 @@ Extra-Source-Files:
   tests/IntegrationTests/multiple-source/should_run/p/Setup.hs
   tests/IntegrationTests/multiple-source/should_run/q/q.cabal
   tests/IntegrationTests/multiple-source/should_run/q/Setup.hs
+  tests/IntegrationTests/user-config/common.sh
```
23Skidoo: Good catch, I completely forgot about this.

23Skidoo:
Merged, thanks!

grayjay:
Thanks @23Skidoo and @ezyang

PR #3100, opened 2016-02-01
Allow man page integration test to run on Windows.

grayjay:
The man page generated on Windows lists commands such as `cabal.exe install`, but the test expects `cabal install`.  I'm not completely sure how the failure should be fixed.  This commit changes the test so that it accepts either `cabal` or `cabal.exe`.

### DIFF
--- a/cabal-install/tests/IntegrationTests/manpage/should_run/outputs_manpage.sh
+++ b/cabal-install/tests/IntegrationTests/manpage/should_run/outputs_manpage.sh
@@ -3,9 +3,9 @@
 OUTPUT=`cabal manpage`
 
 # contains visible command descriptions
-echo $OUTPUT | grep -q '\.B cabal install' || die "visible command description line not found in:\n----$OUTPUT\n----"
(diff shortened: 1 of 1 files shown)


23Skidoo:
LGTM.

BTW, you might want to look into enabling Windows CI for Cabal with AppVeyor or [GHC Builder](https://ghc.haskell.org/trac/ghc/wiki/Builder) (@gbaz said that haskell.org could provide us with resources).

23Skidoo:
Another approach is to change manpage generation code - the `.exe` suffix is redundant in this case. Don't know if anyone uses the manpage on Windows, however.

grayjay:
Good idea.  I'll look into CI when I have time.  There have been enough Windows failures that it seems worthwhile.

I think that removing the `.exe` would make it easier to skim the list of sub-commands.  How about I use `"cabal"` as the program name and leave the test unchanged?

23Skidoo:
> I think that removing the `.exe` would make it easier to skim the list of sub-commands.  How about I use `"cabal"` as the program name and leave the test unchanged?

Sure, that'd also be accepted.

grayjay:
I modified the manpage command in #3103 instead.

PR #3103, opened 2016-02-01
Always use "cabal" as the program name when generating the man page.

grayjay:
The man page previously listed commands like "cabal.exe install" on Windows, which was unnecessary.

### DIFF
--- a/cabal-install/Main.hs
+++ b/cabal-install/Main.hs
@@ -160,7 +160,8 @@ import qualified Paths_cabal_install (version)
 
 import System.Environment       (getArgs, getProgName)
 import System.Exit              (exitFailure)
-import System.FilePath          (splitExtension, takeExtension, (</>), (<.>))
+import System.FilePath          ( splitExtension, takeBaseName
+                                , takeExtension, (</>), (<.>))
 import System.IO                ( BufferMode(LineBuffering), hSetBuffering
 #ifdef mingw32_HOST_OS
                                 , stderr
@@ -1266,4 +1267,4 @@ manpageAction commands _ extraArgs _ = do
   unless (null extraArgs) $
     die $ "'manpage' doesn't take any extra arguments: " ++ unwords extraArgs
   pname <- getProgName
-  putStrLn (manpage pname commands)
+  putStrLn $ manpage (takeBaseName pname) commands


ezyang:
Hmm. So if the Cabal executable is installed at a different name, you will still put cabal in the manpage? But I suppose that is OK.

cabal-install/Main.hs
```diff
@@ -1265,5 +1265,4 @@ manpageAction :: [CommandSpec action] -> Flag Verbosity -> [String] -> Action
 manpageAction commands _ extraArgs _ = do
   unless (null extraArgs) $
     die $ "'manpage' doesn't take any extra arguments: " ++ unwords extraArgs
-  pname <- getProgName
-  putStrLn (manpage pname commands)
+  putStrLn (manpage "cabal" commands)
```
23Skidoo: Maybe use `takeBaseName` instead? Use case: a user has both `cabal-1.18` and `cabal-1.20` installed.

grayjay:
I used `takeBaseName` instead.  Thanks for the suggestion!

23Skidoo:
Merged, thanks!

grayjay:
@23Skidoo I just realized that `takeBaseName` doesn't work.  It converts `cabal-1.20` to `cabal-1`.  Should I revert this and reopen my previous pull request that updated the man page test to allow `cabal.exe` in the output?

23Skidoo:
@grayjay Just tested, it seems to work for me on Windows:

```
Prelude System.FilePath> takeBaseName $ "C:\\foo\\cabal-1.22.exe"
"cabal-1.22"
```

This is with GHC 7.8/HP 2014.2.0.0. Maybe there was a bug in a previous version of the `filepath` library?

grayjay:
I meant on Linux.  This PR calls `takeBaseName` unconditionally, and it treats `.20` as an extension.

23Skidoo:
I see. Then we should only call `takeBaseName` on Windows.

grayjay:
@23Skidoo I don't know how to avoid incorrectly removing the extension.  Apparently the `.exe` isn't required, so I get the same result when I rename the executable to `cabal-1.22` and run `manpage` on Windows.  I think it would be safer to just revert this PR and change the test.  Or I could remove `.exe` specifically.

23Skidoo:
> Or I could remove `.exe` specifically.

Let's do this. And since this should be harmless on non-Windows, let's not hide this code behind CPP.

PR #3106, opened 2016-02-02
cbits: WIN32 guard for wider range of compilers

Heather:


23Skidoo:
Merged, thanks!

hvr:
Related: http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowsCygwinnonPOSIXandMinGW

23Skidoo:
> Related: http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowsCygwinnonPOSIXandMinGW

From this table it looks like `_WIN32` (what we had originally) works for all compilers.

@Heather Which compilers were you having problems with?

Heather:
@hvr @23Skidoo honestly didn't know that, keeping just `_WIN32` should be fine, but before there was `WIN32` and `_WIN32` in different places and I really doubt that it was made for some reason, also I used to exclude cygwin from list

23Skidoo:
GHC doesn't support Cygwin nowadays. Might be a good idea to be future-proof, though. So I think we can simplify the conditional to `defined(_WIN32) && !defined(__CYGWIN__)` and apply this change consistently everywhere.

Heather:
@23Skidoo agreed, should I open correcting PR for this?

23Skidoo:
Sure, go ahead.

Issue #3108, opened 2016-02-03
Properly say what our GHC version support window is


ezyang:
Not our "we build with this version of GHC" window, but the "Cabal knows how to drive this version of GHC" window. Based on #3034, we claim to support GHCs all the way back to 6.4.2. When, if ever, do we drop support? Is this claim even meaningful, given we don't actually test these GHCs?

Once we define a support window, we should start testing these versions on Travis. This is a bit delicate, we need to install both the old GHC and the new GHC. Fortunately, hvr's PPA installs each version of GHC to a separate folder, which is great; even better, builds for precise go all the way back to 6.12 (perhaps that should be our actual support window?) Another thing that needs to be dealt with is being able to run the tests using a `--with-ghc` which is not the same as the version of Cabal that was built.

hvr:
Btw, what's the oldest GHC not-yet-EOL'ed Linux distribution still bundle?

ezyang:
Ubuntu Precise (released 2012-04-26, EOL 2017-04-26) : 7.4.1
Debian wheezy (released 2013-05-02, became oldstable in 2015-04-25; probably EOL some time in 2017): 7.4.1
CentOS 5 (released 2007-04-12, EOL 2017-04-31): there appears to be no GHC in the EPEL repository, and very difficult to get it to work. This guy couldn't get cabal-install to work https://mail.haskell.org/pipermail/cabal-devel/2010-April/006212.html ; this guy successfully got 7.0.2 by bootstrapping from some 6.12.1 blobs https://posterousklevstul.wordpress.com/2011/03/23/haskell-ghc-7-0-2-on-centos-5-5/
CentOS 6 (released 2011-07-10, EOL 2020-11-30): EPEL has 7.0.4 (https://dl.fedoraproject.org/pub/epel/6/i386/repoview/ghc.html). But EPEL does not count as "supported", it is just backports from Fedora which has a much smaller release cycle. Usually people use 7.0.4 to bootstrap to a newer compiler.
Fedora 22: GHC 7.8 (lol!)

ezyang:
Here are some things we can refactor in the codebase once we drop support:

* `-package` compatibility (`-package-id` as added in GHC 6.12)
* `IPI642` (the correct format was added in GHC 6.10)
* Generally, grep the codebase for `Version [6`, many of these things would be out of the support window.

PR #3111, opened 2016-02-05
Change Cabal's GHC support window to GHC 6.12 or later

ezyang:
JUST THE LATEST PATCH, it was based off of #3066, but I think it should not be difficult to commute earlier. See #3108 for some rationale.


23Skidoo:
> JUST THE LATEST PATCH

Can you please cherry-pick it into a new branch based off `master` and run `push -f`?

23Skidoo:
Though I think this should wait until after 1.24, if it's fine with you.

ezyang:
Yeah, I didn't bother cherry-picking it because I'm pretty sure this is post 1.24.

23Skidoo:
Merged, thanks!

PR #3113, opened 2016-02-05
Fix unit test that didn't compile

kolmodin:
Broke with 335ac900e3f571ea9222ba4934ab9ab930153ad2.


23Skidoo:
@dcoutts I'm merging this, because it looks quite harmless, even though the build still fails.

PR #3114, opened 2016-02-06
Require QuickCheck >= 2.7

grayjay:
The tests use `===` and `counterexample`, which were added in QuickCheck-2.7.


23Skidoo:
Merged, thanks!

PR #3116, opened 2016-02-07
Fix bootstrap.sh: add hashable

tuncer:
Since 1e1d82b8e, hashable is needed.  Adapt bootstrap.sh accordingly.


23Skidoo:
Merged, thanks!

PR #3117, opened 2016-02-07
Cabal with ghc package tests

ezyang:
In preparation for testing Cabal against old versions of GHC, I've made the tests more robust to being run against the boot package database, and added a new environment variable for PackageTests to make it use a different version of GHC.


23Skidoo:
The patch itself looks good, can you fix the Travis failure?

ezyang:
Closed.

PR #3118, opened 2016-02-07
Cabal: define MonadFail instances

hvr:
This silences `-Wnoncanonical-monadfail-instances` warnings in the Cabal
library

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -179,6 +179,10 @@ library
   if impl(ghc < 7.6)
     build-depends: ghc-prim >= 0.2 && < 0.3
 
+  if !impl(ghc >= 8.0)
+    -- provides compat Control.Monad.Fail module
+    build-depends: fail == 4.9.*
+
   if !os(windows)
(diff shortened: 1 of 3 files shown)


hvr:
NB: This patch avoids enabling CPP for the two modules by adding a conditional dependency on the tiny [`fail`](http://hackage.haskell.org/package/fail) package...

23Skidoo:
> NB: This patch avoids enabling CPP for the two modules by adding a conditional dependency on the tiny [`fail`](http://hackage.haskell.org/package/fail) package...

The problem with this is that Cabal will be no longer buildable out-of-the box on GHC < 8.

23Skidoo:
Given that the `fail` package is so tiny, can we solve this by adding a `D.Compat.MonadFail` module instead?

hvr:
sure... that would have been plan b... :-)
-- 

hvr:
@23Skidoo see #3124 for the compat-module variant

23Skidoo:
Closing in favour of #3124.

PR #3119, opened 2016-02-08
bootstrap.sh: fix error introduced in 46584a2fdbd

tuncer:
In 46584a2fdbd I've accidentally put the hashable version string on a
separate line, causing the shell to try to execute the command
"1.2.4.0".  Fix it by removing the line.


23Skidoo:
Merged, thanks!

PR #3131, opened 2016-02-10
Give `cabal-install` the Wcompat treatment

hvr:


hvr:
Btw, if don't need to support GHC < 7.4 we can leverage `Generic`-based deriving for many of the tedious boilerplate-y `Semigroup`/`Monoid` instances in `Cabal` and `cabal-install`... assuming we can derive `Generic` for the respective types.

23Skidoo:
Yes, we should do that, but after the 1.24 release.

23Skidoo:
Merged, thanks!

hvr:
btw, we're currently using cabal install 1.22 in the Travis job for GHC HEAD... but GHC 8.0/HEAD requires at least 1.23 ; see also https://phabricator.haskell.org/D1892

23Skidoo:
Yes, I know it's broken.

PR #3134, opened 2016-02-10
Correctly remove .exe from the cabal command listed in the man page.

grayjay:
This commit only removes the extension when it is equal to ".exe".  It fixes the change made in #3103, which didn't handle executable names like `cabal-1.23`.


23Skidoo:
Merged, thanks!

PR #3135, opened 2016-02-11
Quote cabal and ghc-pkg paths in cabal-install integration tests

grayjay:
Spaces in the path to `ghc-pkg` caused two tests to fail when run with the Haskell Platform on Windows.


ezyang:
LGTM.

23Skidoo:
Merged, thanks!

PR #3137, opened 2016-02-12
 {test,bench}: Ignore "bad version" error when loading LBI.

23Skidoo:
See https://github.com/haskell/cabal/pull/1821#commitcomment-16052702.


ezyang:
This fix seems questionable but I can't think of a better way of working around it. (Maybe having cabal-install just parse the package description itself.) It's a bit counterintuitive that returning an empty list means "build everything".

23Skidoo:
> It's a bit counterintuitive that returning an empty list means "build everything".

Good point, I'll add a special-purpose ADT to make the code more readable.

23Skidoo:
> Good point, I'll add a special-purpose ADT to make the code more readable.

Done.

ttuegel:
> I'll add a special-purpose ADT to make the code more readable.
> 
> Done.

Is that commit in there? I don't see it.

23Skidoo:
> Is that commit in there? I don't see it.
I rebased and force-pushed.

PR #3138, opened 2016-02-12
Reinstall add-source deps even when LBI can't be loaded.

### LINKED ISSUE #3136: add-source deps are not reinstalled when LBI can't be loaded
Reinstallation of add-source deps is done in `reconfigure`, and that function silently ignores bad version errors and returns when loading LBI fails. This can happen when a package is using `build-type: Custom` and the setup script is built against a different version of Cabal than `cabal-install`. So changes in add-source deps are ignored in this case.

23Skidoo:
But only if it happens due to version mismatch. Fixes #3136.


23Skidoo:
This bites me quite often when working on `cabal-install`, BTW, now that it uses `build-type: Custom`.

23Skidoo:
Replaced calls to `die` with `notice`.

23Skidoo:
This is somewhat fragile when you're using multiple compilers inside the same sandbox because we can't tell which compiler the project was configured with (since we couldn't load the LBI). Fixing this properly will be a bit more involved, so I'm postponing this for after 1.24.

23Skidoo:
> This is somewhat fragile when you're using multiple compilers inside the same sandbox because we can't tell which compiler the project was configured with (since we couldn't load the LBI). Fixing this properly will be a bit more involved, so I'm postponing this for after 1.24.

Having thought a bit more about this, I now believe we should make the case of unreadable LBI an error and just always compile setup scripts with the same Cabal version that `cabal-install` was built with, installing it automatically if necessary. Now that we have setup dependencies (thanks, @edsko!) this should be easier to implement. The only potentially problematic case is when an old `cabal-install` tries to compile a setup script that doesn't support old versions of Cabal, but then we should just advise the user to upgrade (and the error shouldn't preclude upgrading `cabal-install` itself).

Also, we should get better at bumping development versions when the LBI format is changed (#3181).

ttuegel:
> Having thought a bit more about this, I now believe we should make the case of unreadable LBI an error and just always compile setup scripts with the same Cabal version that cabal-install was built with, installing it automatically if necessary. Now that we have setup dependencies (thanks, @edsko!) this should be easier to implement. The only potentially problematic case is when an old cabal-install tries to compile a setup script that doesn't support old versions of Cabal, but then we should just advise the user to upgrade (and the error shouldn't preclude upgrading cabal-install itself).

I strongly agree. I would never expect an old cabal-install to work with a newer Cabal, or if it did, I would consider it a happy accident. Not only does this simplify things for us, it will be easier for users to understand what's gone wrong and how to correct it ("Your cabal-install is too old, please upgrade.")

23Skidoo:
> I now believe we should make the case of unreadable LBI an error and just always compile setup scripts with the same Cabal version that `cabal-install` was built with

This will also make life a bit harder for developers, but with #3181 fixed, I think we'll cope.

23Skidoo:
This has bitrotted and was obsoleted by new-build, closing.

PR #3141, opened 2016-02-13
configureToolchain: Use the same program names as 'ghc --info'.

### LINKED ISSUE #3140: configureToolchain tests the wrong ld for -x support
Steps to reproduce: Set `ld command` to `false` in the ghc settings file. Then try to build any package.

Output with `-v` includes
```
/home/rwbarton/bin/ghc -c /tmp/16816927771714636915.c -o /tmp/1957747793424238335.o
/usr/bin/ld -x -r /tmp/1957747793424238335.o -o /tmp/7198853861649760492.o
...
```
Expected output was `false -x -r /tmp/1957747793424238335.o -o /tmp/7198853861649760492.o`.

This is bad for two reasons:

* We're presumably going to use the `ld` from `ghc --info` later, rather than the system linker, so that's the one for which we need to know whether it supports `-x`. (Though I don't know when Cabal actually uses `ld`, if ever.)

* Using the system linker here can fail in some scenarios, such as when running a 32-bit ghc on a 64-bit host system.

As a workaround I have to use the `--with-ld` flag manually. Then this works as expected.

23Skidoo:
For example, we used to search for `ld` when the program name specified by `ghc --info` was `i386-unknown-linux-ld`.

Fixes #3140.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -189,7 +189,8 @@ library
 
   ghc-options: -Wall -fno-ignore-asserts -fwarn-tabs
   if impl(ghc >= 8.0)
-    ghc-options: -Wcompat -Wnoncanonical-monad-instances -Wnoncanonical-monadfail-instances
+    ghc-options: -Wcompat -Wnoncanonical-monad-instances
+                 -Wnoncanonical-monadfail-instances
 
   exposed-modules:
     Distribution.Compat.CreatePipe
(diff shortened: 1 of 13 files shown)


23Skidoo:
@dcoutts This moves `handleDoesNotExist` to `D.S.Utils`, hope you don't mind.

BardurArantsson:
Everything LGTM.

rwbarton:
Yes, it works for me now, thanks! One typo though: in the commit message you mean `ghc --info`, not `gcc --info`.

23Skidoo:
@BardurArantsson Thanks for the code review, by the way.

Issue #3112, opened 2016-02-05
Continuous integration on Windows


grayjay:
I experimented with building cabal on [AppVeyor] (https://www.appveyor.com/) by enabling it on a branch in my repository.  I wrote a simple script that can run both packages' tests.  Here's my proposed plan for setting up Windows CI:

- Wait until the cabal 1.24 release so that we can use the newer version of cabal to install dependencies.  Then `cabal update` will be able to download the index using curl.  cabal-install-1.22.0.0 failed: https://ci.appveyor.com/project/grayjay/cabal/build/1.0.6#L10764
- Address the issues relating to cabal-install test failures on Windows: #2872, #3059
- Handle the Cabal tests that fail due to missing shared libraries with GHC >= 7.8 (https://ghc.haskell.org/trac/ghc/ticket/8228).  I disabled those tests individually by using CPP on my branch.  I think it would be better to separate out the tests that use shared libraries in order to highlight what tests are being skipped.  Maybe they could go in a separate test suite, or at least a separate top-level test tree.
- Test with GHC 7.10.3 or 8.0 first, and possibly add more versions later.

Branch: https://github.com/grayjay/cabal/tree/appveyor-ghc-7.10.3
Log: https://ci.appveyor.com/project/grayjay/cabal/branch/appveyor

I also tried some other versions of GHC and the Haskell Platform but didn't have much luck.

A related issue: https://github.com/elm-lang/elm-platform/issues/105.

Remaining Windows CI issues:

- [ ] #3123 - `cabal exec` integration tests
- [ ] #3126 - file monitor unit tests

23Skidoo:
Thanks for looking into this! The `cabal update` failure looks weird, maybe AppVeyor is doing something  funny with the network (like forcing an HTTPS connection). We may also want to patch the HP installer script to better support silent installation - this [shouldn't be very hard](http://stackoverflow.com/questions/14445395/nsis-installer-silent-mode), and the new release will be out soon (simultaneously with GHC 8.0).

grayjay:
@23Skidoo No problem.  I'll plan to make a pull request once cabal-install-1.24 is available.  I was able to install HP 2014.2.0.0 from Chocolatey, but then I had trouble installing cabal-install's dependencies.

grayjay:
[...]

PR #3146, opened 2016-02-14
Add flag to disable Cabal package tests that use shared libraries

grayjay:
The flag `--skip-shared-library-tests` allows the tests to run when shared libraries are unavailable, such as during the AppVeyor build.

This is a step towards #3112.  I wanted to minimize the changes to existing tests but also make it easy to see what tests are being skipped.  I'm not sure if this is the best solution.

### DIFF
--- a/Cabal/tests/PackageTests.hs
+++ b/Cabal/tests/PackageTests.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+
 -- The intention is that this will be the new unit test framework.
 -- Please add any working tests here.  This file should do nothing
 -- but import tests from other modules.
@@ -25,9 +27,12 @@ import Distribution.Verbosity (normal, flagToVerbosity)
 import Distribution.ReadE (readEOrFail)
 
 import Control.Exception
+import Data.Typeable (Proxy(..), Typeable)
 import Distribution.Compat.Environment ( lookupEnv )
 import System.Directory
 import Test.Tasty
+import Test.Tasty.Options
+import Test.Tasty.Ingredients
 import Data.Maybe
 
 #if MIN_VERSION_base(4,6,0)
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -14,6 +14,7 @@ build_script:
   - Setup configure --user --ghc-option=-Werror --enable-tests
   - Setup build
   - Setup test unit-tests --show-details=streaming
+  # - Setup test package-tests --show-details=streaming --test-options=--skip-shared-library-tests
```
23Skidoo: Is this intentional?

grayjay: I added it because I thought we could run the package tests soon.  The only failure is #3147.

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -244,5 +242,28 @@ tests config =
...
 
     tc :: FilePath -> TestM a -> TestTree
-    tc name m
-        = testCase name (runTestM config name Nothing m)
+    tc = testCase config
+
+-- | Tests that require shared libraries.
+sharedLibTests :: SuiteConfig -> [TestTree]
+sharedLibTests config =
+  tail [ undefined
```
23Skidoo: What's the purpose of `undefined` here?

23Skidoo: OK, I see, it's to make rearranging test cases easier (you don't need to delete commas).

23Skidoo:
Merged, thanks!

Cabal/tests/PackageTests/Tests.hs
grayjay: I copied the pattern from the top of the file. I meant to ask if I understood it correctly but forgot.

grayjay:
Thanks!  I'll update HACKING.md and/or the test readme.

Cabal/tests/PackageTests.hs
```diff
@@ -25,9 +27,12 @@ import Distribution.Verbosity (normal, flagToVerbosity)
 import Distribution.ReadE (readEOrFail)
 
 import Control.Exception
+import Data.Typeable (Proxy(..), Typeable)
```
23Skidoo: BTW, this broke the build on GHC <= 7.6. I changed this line to import `Data.Proxy` instead and added a dependency on `tagged`.

grayjay: Thanks.  I was just looking into that.

ezyang:
This is late to the party, but the prevailing style for this test sutie has been to use environment variables to pass everything in. I guess now that we've added support for extra command line arguments we should make all of the other parameters support it too.

grayjay:
I chose a flag because I wanted the parameter to be more explicit.  An environment variable would also work.  Should I change it, for consistency?

23Skidoo:
I don't think we need to change it, but as @ezyang says, it'd be nice if the other parameters could be also passed via the command line.

PR #3144, opened 2016-02-13
Start documenting cabal-install integration tests

grayjay:
/cc @BardurArantsson

PR #3148, opened 2016-02-14
Format cabal-install/tests/README.md

grayjay:
This is a follow-up to #3144 that splits the README into 80-character lines.


grayjay:
Thanks!

23Skidoo:
If you use Emacs, you can just hit `M-q` to do this.

PR #3164, opened 2016-02-18
Run cabal-install unit tests on AppVeyor

grayjay:
I excluded the FileMonitor tests because of exceptions from `removeDirectoryRecursive` (https://github.com/haskell/cabal/issues/3126#issuecomment-184080769).

I also wanted to enable the integration tests, but I couldn't figure out how to exclude more than one group of tests.  `tasty` doesn't support multiple filters: https://github.com/feuerbach/tasty/issues/38.


23Skidoo:
Merged, thanks!

PR #3166, opened 2016-02-19
cabal-install.cabal: Update 'other-modules'

erikd:
Building from sdist files fails without this fix.


23Skidoo:
Merged, thanks!

PR #3171, opened 2016-02-19
Allow limiting the scope of '--allow-newer'. 

### LINKED ISSUE #2756: Extend `--allow-newer` to limit scope to single packages

23Skidoo:
Fixes #2756.

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -315,7 +315,9 @@ configure (pkg_descr0', pbi) cfg = do
           -- Ignore '--allow-newer' when we're given '--exact-configuration'.
           if fromFlagOrDefault False (configExactConfiguration cfg)
           then pkg_descr0'
-          else relaxPackageDeps (configAllowNewer cfg) pkg_descr0'
+          else relaxPackageDeps
+               (fromMaybe AllowNewerNone $ configAllowNewer cfg)
+               pkg_descr0'
 
     setupMessage verbosity "Configuring" (packageId pkg_descr0)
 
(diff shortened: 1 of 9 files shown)


hvr:
@23Skidoo :+1:

one question though, does the allow-newer field only allow `True`/`False`? @dcoutts: as the new `cabal.project` features allows per-package flags; it'd be neat if we could then write something like

```
-- applies to all
with-compiler: /opt/ghc/8.0.1/bin/ghc
-- corresponds to `--allow-newer: ghc-prim`
allow-newer: ghc-prim
tests: False

-- corresponds to `--allow-never aeson:base,aeson:process`
package aeson
   allow-newer: base,process

23Skidoo:
> one question though, does the `allow-newer` field only allow `True`/`False`?

Yes, but I can change it if you want.

23Skidoo:
> Yes, but I can change it to also support the full syntax if you want.
Done.

PR #3172, opened 2016-02-19
Make secure remote repos a bit more discoverable.

23Skidoo:
Before this change, the relevant section of the default config file looked like this:
```
```
With this change, it looks like this:
```
  -- secure: False
  -- root-keys:
  -- key-threshold:
```

/cc @edsko @dcoutts


23Skidoo:
The idea is that people will google for `cabal key-threshold` and discover http://www.well-typed.com/blog/2015/07/hackage-security-alpha/.

edsko:
Nice!

PR #3175, opened 2016-02-20
Warn user when deleting package database

### LINKED ISSUE #2381: cabal sandbox delete should reset $GHC_PACKAGE_PATH
Currently if you delete a sandbox whilst inside `cabal exec bash` then `$GHC_PACKAGE_PATH` is still set to point to a non-existent file.

ttuegel:
Fixes #2381. If CABAL_SANDBOX_PACKAGE_PATH, GHC_PACKAGE_PATH, or
GHCJS_PACKAGE_PATH refer to a package database inside the sandbox, warn
the user that the package database has been deleted.

### DIFF
--- a/cabal-install/Distribution/Client/Sandbox.hs
+++ b/cabal-install/Distribution/Client/Sandbox.hs
@@ -109,18 +109,23 @@ import qualified Data.Map                          as M
 import qualified Data.Set                          as S
 import Data.Either                            (partitionEithers)
 import Control.Exception                      ( assert, bracket_ )
-import Control.Monad                          ( forM, liftM2, unless, when )
+import Control.Monad                          ( forM, liftM, liftM2, unless, when )
 import Data.Bits                              ( shiftL, shiftR, xor )
 import Data.Char                              ( ord )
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/Sandbox.hs
```diff
@@ -368,6 +374,47 @@ sandboxDelete verbosity _sandboxFlags globalFlags = do
...
 
+      absSandboxDir <- makeAbsolute sandboxDir
+
+      let
+        pathInsideSandbox = isPrefixOf absSandboxDir
+
+      let
+        -- Warn the user if CABAL_SANDBOX_PACKAGE_PATH points inside the
+        -- deleted sandbox.
+        checkSandboxPackagePath path = do
```
23Skidoo: `checkSandboxPackagePath`, `checkGHCPackagePath` and `checkGHCJSPackagePath` look quite similar. Can we merge them into a single function? Also, the `liftM (maybe [] splitSearchPath) ...` bit looks like it could be part of that function as well.

23Skidoo:
LGTM module copy-paste.

PR #3176, opened 2016-02-20
Add field for setting framework search path (fixes #182).

### LINKED ISSUE #182: Handle framework paths (-F) in Cabal

23Skidoo:
An updated version of #3158 with the following changes:

* Renamed `framework-dirs` to `extra-framework-dirs`.
* Added `extra-framework-dirs` to `ConfigFlags`.
* Added some checks to `D.PD.Check`.
* Updated the changelog.


komadori:
Ah, you beat me to it! One difference between that and my own patch is that I made the check PackageDistWarning rather than PackageDistInexcusable, so that you wouldn't be forced to raise the minimum Cabal version for users not using MacOS or shared libraries, but MacOS users would still pick it up if they had a recent enough Cabal. What do you think?

Also, the new entry in "Cabal/doc/installing-packages.markdown" has the wrong option name.

23Skidoo:
@komadori Thanks! Fixed the typo and changed the check to use `PackageDistSuspiciousWarn`.

PR #3177, opened 2016-02-20
Don't pass OSX-specific options to GHC on non-OSX.

23Skidoo:
Basically a two-line change, the rest is about passing a `Platform` value to `renderGhcArgs`.


23Skidoo:
Probably not a very big deal, since most packages hide those options behind Cabal flags, but makes the code a bit more robust.

PR #3183, opened 2016-02-23
Fix exitcode on `upload`ing failure

lspitzner:
When upload failed (e.g. wrong auth), the previous exitcode was not nonzero as expected.


23Skidoo:
Merged, thanks!

Issue #1634, opened 2013-12-31
cabal-install does not quote the path in the --with-gcc parameter


tibbe:
According to haskell/network#118 cabal-install does properly quote the path passed to `--with-gcc`, causing failures on Windows when the path contains a space.

drbo:
The only solution I know of is to either fix or ditch Autoconf, which has a bad track record with handling spaces in variables.

gbaz:
For the record, I've found the short names solution to work. What a pain :-/

23Skidoo:
IIUC, @hvr fixed this in #2946 (we now use `./configure CC=/path/to/gcc` instead of `./configure --with-gcc`). If it still doesn't work, then it's a bug in Autoconf.

23Skidoo:
> If it still doesn't work, then it's a bug in Autoconf.

It doesn't, see #3185.

PR #3184, opened 2016-02-24
Unbreak the build on AppVeyor.

23Skidoo:
Take 2.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,9 @@
 install:
-  - choco install HaskellPlatform -version 2014.2.0.0
-  - SET PATH=C:\Program Files\Haskell Platform\2014.2.0.0\bin;C:\Program Files\Haskell Platform\2014.2.0.0\mingw\bin;C:\Program Files\Haskell Platform\2014.2.0.0\lib\extralibs\bin;%PATH%
+  - choco install ghc -version 7.10.3
+  - SET PATH=%PATH%;C:\tools\ghc\ghc-7.10.3\bin
+  - curl -o cabal.zip https://www.haskell.org/cabal/release/cabal-install-1.24.0.0-rc1/cabal-install-1.24.0.0-rc1-x86_64-unknown-mingw32.zip
+  - 7z x cabal.zip
+  - cabal --version
   - cabal update
 
 build_script:
@@ -9,7 +12,7 @@ build_script:
 
   # 'echo "" |' works around an AppVeyor issue:
   # https://github.com/commercialhaskell/stack/issues/1097#issuecomment-145747849
-  - echo "" | cabal install --only-dependencies --enable-tests
+  - echo "" | ..\cabal install --only-dependencies --enable-tests
 
   - Setup configure --user --ghc-option=-Werror --enable-tests
   - Setup build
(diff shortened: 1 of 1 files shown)


grayjay:
@23Skidoo Thanks for uploading the binary. I'm trying to update my previous branch that downloaded GHC and cabal separately: https://github.com/grayjay/cabal/tree/appveyor-newer-ghc-and-cabal

23Skidoo:
Apparently when you run `7z x cabal.tar.gz` it unpacks only the `cabal.tar` file.

grayjay:
Now it's failing to download the packages: https://ci.appveyor.com/project/grayjay/cabal/build/1.0.73

23Skidoo:
@grayjay Maybe the problem is that AppVeyor needs a 64-bit binary? The one I uploaded is 32-bit, but the one that comes with Haskell Platform is 64-bit (since it's from `HaskellPlatform-2014.2.0.0-x86_64-setup.exe`).

23Skidoo:
Something funky is going on:

```
ansi-terminal-0.6.2.3 failed while downloading the package. The exception was:
DeleteFile
"C:\\Users\\appveyor\\AppData\\Roaming\\cabal\\packages\\hackage.haskell.org\\ansi-terminal\\0.6.2.3\\curl-headers2812.txt":
permission denied (The process cannot access the file because it is being used
by another process.)
```

grayjay:
I don't know.  I looked back at the other builds that I tried, and I think that all of the cabal binaries that failed to download the index were 32-bit.

23Skidoo:
OK, I'll build a 64-bit binary then.

23Skidoo:
The 64-bit binary seems to be able to download packages. Yay!

grayjay:
Great!

23Skidoo:
So now it fails when trying to configure `network`:

```
cabal: Entering directory 'C:\Users\appveyor\AppData\Local\Temp\1\cabal-tmp-2404\network-2.6.2.1'
Configuring network-2.6.2.1...
configure: WARNING: unrecognized options: --with-compiler
checking build system type... x86_64-pc-msys
checking host system type... x86_64-pc-msys
checking for gcc... C:\Program Files\Haskell Platform\2014.2.0.0\mingw\bin\gcc.exe
checking whether the C compiler works... no
configure: error: in `/tmp/cabal-tmp-2404/network-2.6.2.1':
configure: error: C compiler cannot create executables
See `config.log' for more details
cabal: Leaving directory 'C:\Users\appveyor\AppData\Local\Temp\1\cabal-tmp-2404\network-2.6.2.1'
Failed to install network-2.6.2.1
```

grayjay:
`network` works with GHC 7.10.3: https://ci.appveyor.com/project/grayjay/cabal

23Skidoo:
Thanks, changed the script to use GHC 7.10.3.

23Skidoo:
And it finally passed!

grayjay:
I can reproduce the network configure error locally with `cabal-install-1.24.0.0`, but not `cabal-install-1.22.6.0`.  I think it can't handle the spaces in the path to gcc.

23Skidoo:
That's #1634. So it's still broken? Maybe we need to add a workaround.

23Skidoo:
Filed an issue: #3185.

grayjay:
Thanks.  It looks like 1.22 was able to configure network because it used `--with-gcc` instead of `CC`.  The configure script ignored `--with-gcc` and found `gcc` on the path.

Issue #3169, opened 2016-02-19
Replace boilerplate Monoid/Semigroup instances with generics


23Skidoo:
We have a large number of boilerplate `Monoid`/`Semigroup` instances for large records in `D.S.Setup` and `D.C.Setup` (as well as in #3156). Those should be replaced with generics.

Relevant Hackage package: https://hackage.haskell.org/package/generic-deriving-1.10.1/docs/Generics-Deriving-Monoid.html.

hvr:
@23Skidoo the `generic-deriving` package may be overkill here... we need as little as (stolen from `semigroups`):

```hs
import Data.Semigroup
import GHC.Generics

-- | Generically generate a 'Semigroup' ('<>') operation for any type
-- implementing 'Generic'. This operation will append two values
-- by point-wise appending their component fields. It is only defined
-- for product types.
--
-- @
-- 'gmappend' a ('gmappend' b c) = 'gmappend' ('gmappend' a b) c
-- @
gmappend :: (Generic a, GSemigroup (Rep a)) => a -> a -> a
gmappend x y = to (gmappend' (from x) (from y))

class GSemigroup f where
  gmappend' :: f p -> f p -> f p

instance GSemigroup U1 where
  gmappend' _ _ = U1

instance GSemigroup V1 where
  gmappend' x y = x `seq` y `seq` error "GSemigroup.V1: gmappend'"

instance Semigroup a => GSemigroup (K1 i a) where
  gmappend' (K1 x) (K1 y) = K1 (x <> y)

instance GSemigroup f => GSemigroup (M1 i c f) where
  gmappend' (M1 x) (M1 y) = M1 (gmappend' x y)

instance (GSemigroup f, GSemigroup g) => GSemigroup (f :*: g) where
  gmappend' (x1 :*: x2) (y1 :*: y2) = gmappend' x1 y1 :*: gmappend' x2 y2

-- | Generically generate a 'Monoid' 'mempty' for any product-like type
-- implementing 'Generic'.
--
-- It is only defined for product types.
--
-- @
-- 'gmappend' 'gmempty' a = a = 'gmappend' a 'gmempty'
-- @

gmempty :: (Generic a, GMonoid (Rep a)) => a
gmempty = to gmempty'

class GSemigroup f => GMonoid f where
  gmempty' :: f p

instance GMonoid U1 where
  gmempty' = U1

instance (Semigroup a, Monoid a) => GMonoid (K1 i a) where
  gmempty' = K1 mempty

[...]

PR #3193, opened 2016-02-26
Add `gmappend`/`gmempty` Generics-helpers (re #3169)

hvr:
This adds `gmappend`/`gmempty` to `Distribution.Compat.Semigroup` so it becomes available for `Cabal` and `cabal-install`.

### DIFF
--- a/Cabal/Distribution/Compat/Semigroup.hs
+++ b/Cabal/Distribution/Compat/Semigroup.hs
@@ -21,6 +21,12 @@ import qualified Data.Monoid as Mon
 #else
 -- provide internal simplified non-exposed class for older GHCs
 import Data.Monoid as Mon (Monoid(..), All(..), Any(..), Dual(..))
+-- containers
+import Data.Set (Set)
+import Data.IntSet (IntSet)
+import Data.Map (Map)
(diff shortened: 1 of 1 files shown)


hvr:
I'd suggest to make these helpers available in `Cabal-1.24` as well, even if we choose not to make use of them in `Cabal-1.24` yet, as this would give us the option to make use of them in `cabal-install-1.24.x`.

23Skidoo:
Merged, thanks! Also cherry-picked into the 1.24 branch.

hvr:
@23Skidoo quite possible, I wonder if there was any reason for @RyanGlScott (assuming he implemented that code originally) to add instances for `V1`

RyanGlScott:
This is originally Eric Mertens's code ([PR here](https://github.com/ekmett/semigroups/pull/40)).

The `U1` case needed for a datatype is one constructor with no arguments (e.g., `data Foo = Foo`), and the `V1` case is needed for datatypes with no constructors (e.g., `data Bar`). Since this is a `Cabal`-specific backport, it's up to you whether you think those are cases you'd ever need to support. But it's really not _that_ much extra code.

PR #3195, opened 2016-02-27
Turn 'configPrograms' field into a 'Last'-monoid

hvr:
This implements the suggestions mentioned at
https://github.com/haskell/cabal/issues/3169#issuecomment-189281916

The main benefit of this change is turning 'ConfigFlags' into a uniform
product-type suitable for generic derivation of pointwise
`Semigroup`/`Monoid` instances.

NB: This changes the `Binary` serialisation of `ConfigFlags` since there's
now an additional `Maybe` inserted in `configPrograms`'s type

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -327,8 +327,8 @@ data ConfigFlags = ConfigFlags {
     -- because the type of configure is constrained by the UserHooks.
     -- when we change UserHooks next we should pass the initial
     -- ProgramConfiguration directly and not via ConfigFlags
-    configPrograms      :: ProgramConfiguration, -- ^All programs that cabal may
-                                                 -- run
+    configPrograms      :: Last' ProgramConfiguration, -- ^All programs that
+                                                       -- @cabal@ may run
 
     configProgramPaths  :: [(String, FilePath)], -- ^user specified programs paths
     configProgramArgs   :: [(String, [String])], -- ^user specified programs args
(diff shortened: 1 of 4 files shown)


23Skidoo:
Still fails (on < 7.6):

```
Distribution/Simple/Setup.hs:397:10:
    No instance for (Binary (Last' ProgramConfiguration))
      arising from a use of `Distribution.Compat.Binary.Class.$gdmput'
    Possible fix:
      add an instance declaration for
      (Binary (Last' ProgramConfiguration))
    In the expression: (Distribution.Compat.Binary.Class.$gdmput)
    In an equation for `put':
        put = (Distribution.Compat.Binary.Class.$gdmput)
    In the instance declaration for `Binary ConfigFlags'
```

hvr:
@23Skidoo that one's weird... I'm investigating...

hvr:
@23Skidoo I'm quite optimistic the builds will work this time...

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -404,7 +404,7 @@ configAbsolutePaths f =
 
 defaultConfigFlags :: ProgramConfiguration -> ConfigFlags
 defaultConfigFlags progConf = emptyConfigFlags {
-    configPrograms     = progConf,
+    configPrograms     = pure progConf,
```
23Skidoo: `Last' progConf` would be more readable IMO.

23Skidoo:
Merged, thanks! I'll need to check whether this patch breaks any setup scripts before it can be cherry-picked into 1.24 as well (`ConfigFlags` are used by hooks, but I don't think hooks can make use of `configPrograms`).

hvr:
@23Skidoo IIRC @dcoutts  was convinced that changing the internal representation of `ConfigFlags` shouldn't break user-hooks

23Skidoo:
Yes, I don't think it does, but it never hurts to double-check.

PR #3196, opened 2016-02-28
Replace Boilerplate Monoid/Semigroup instances with generics

hvr:
addresses #3169


hvr:
@dcoutts @23Skidoo There's a few instances that we can't rewrite yet without more refactoring. Specifically those which have `Bool` flags without `Data.Monoid.All` or `Data.Monoid.Any` `newtype`-wrappers to guide the `Monoid` derivation.

hvr:
Oh, and the `Semigroup` instances for `TestSuite`/`Executable`/`Benchmark` are partial via `error`!

```haskell
instance Semigroup Executable where
  a <> b = Executable{
    exeName    = combine' exeName,
    modulePath = combine modulePath,
    buildInfo  = combine buildInfo
  }
    where combine field = field a `mappend` field b
          combine' field = case (field a, field b) of
                      ("","") -> ""
                      ("", x) -> x
                      (x, "") -> x
                      (x, y) -> error $ "Ambiguous values for executable field: '"
                                  ++ x ++ "' and '" ++ y ++ "'"

```

hvr:
Interesting... the `cabal-install` changes seem to cause integration-test failures

23Skidoo:
Nice! Boolean flags should use `Any`, I think.

hvr:
> Nice! Boolean flags should use `Any`, I think.

there are a few cases which have it inverted:

```haskell
instance Semigroup Library where
 -- ...
       libExposed     = libExposed a && libExposed b, -- so False propagates
```

```haskell
instance Semigroup BuildInfo where
  a <> b = BuildInfo {
    buildable           = buildable a && buildable b,
   -- ...
 }
```

23Skidoo:
Merged, thanks!

hvr:
@23Skidoo what are your plans re 1.24 branch? I mostly want to know to decide what to do about the nix-local branch

23Skidoo:
@hvr Well, I didn't want to backport this into 1.24 initially, but it's probably not a very big deal after all. And merges will be easier. I just need to find a way to download all custom setup scripts from Hackage to check that #3195 didn't break anything.

hvr:
That shouldn't be too hard:

```bash
cabal list --simple | awk '{ print $1 }' | uniq
```

gives you a list of all package names,

and

```bash
curl -s --fail http://hackage.haskell.org/package/$PKGNAME/src/Setup.lhs || curl -s --fail http://hackage.haskell.org/package/$PKGNAME/src/Setup.hs
```

tries to retrieve you the latest version of `$PKGNAME`'s Setup.lhs or Setup.hs

you could also try to be more clever and look at the .cabal files first (e.g. from your 01-index.tar ball) before downloading the Setup.(l)hs files...

23Skidoo:
> you could also try to be more clever and look at the .cabal files first (e.g. from your 00-index.tar ball) and check for the build-type before downloading the Setup.(l)hs files...

Yep, that's what I want to do.

hvr:
In that case:

```bash
cd $TMP
tar xf ~/.cabal/packages/hackage.haskell.org/00-index.tar.gz 
find -name '*.cabal' | xargs grep -li '^build-type:.*custom' | cut -f2 -d/ | uniq
```

23Skidoo:
@hvr Thanks! That was easier than I expected. Uploaded the results to https://github.com/23Skidoo/all-custom-setups. And it turns out `llvm-general` actually uses `configPrograms` in its setup script.

hvr:
@23Skidoo I think that specific case in llvm-general can be made more compatible by providing a function `configPrograms_` with the old type-sig (as it's only used as a function, rather than a field-record-label). and name the actual field differently, e.g. `configPrograms_`

23Skidoo:
Yes, it should work given that the script doesn't use explicit import lists.

hvr:
@23Skidoo btw, @dcoutts tells me he "used to have a script to try compiling all custom Setup.hs scripts with old and new Cabal lib version to look for breakages"

did you try to compile those Setup.hs files or did you just `grep` for configPrograms?

23Skidoo:
@hvr Just grepped.

hvr:
@23Skidoo fwiw, if you're still worried about merging to 1.24, you could try merging only the cabal-install parts, or we could simply back out the `ConfigFlags` instance from the part merged to 1.24

23Skidoo:
@hvr Cherry-picked all your patches into 1.24.

Issue #879, opened 2012-05-24
Add option to list recursive dependencies of a package along with their licenses


bos:
(Imported from [Trac #889](http://hackage.haskell.org/trac/hackage/ticket/889), reported by joeyadams on 2011-09-27)

It should be possible to list all of the dependencies of a package, including their licenses.  This would be useful for:

<ul><li>Developers writing proprietary applications.
</li></ul><ul><li>Developers who don't want their liberally-licensed libraries to be "upgraded" to the GPL or similar by subtle dependencies.
</li></ul>Attached is a simple patch that makes cabal-install list this information when the verbosity level is 2 or higher.  Sample usage:

<pre class="wiki">$ cabal install --verbose=2 --dry-run gnutls
Reading available packages...
Resolving dependencies...
base         BSD3
bytestring   BSD3
ffi          BSD3
ghc-prim     BSD3
gnutls       GPL (Just (Version {versionBranch = [3], versionTags = []}))
integer-gmp  BSD3
monads-tf    BSD3
rts          BSD3
transformers BSD3
In order, the following would be installed:
monads-tf-0.1.0.0 (new package)
gnutls-0.1 (new package)
</pre>This is a quick hack, as I am not familiar with the Cabal codebase.  This should probably be wrapped as a <tt>--list-licenses</tt> command line switch.  Pretty-printing licenses would be a plus.

I am already finding this feature useful, and hope others will too.

libeako:
I was about to request this feature.

Many common licenses require a reproduction of the license when the software is distributed in binary form. For example : BSD3 : 
"Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution."

It is too tedious and error-prone to manually traverse the dependency graph and the individual licenses.

I imagine a command line tool that collects the license info into a folder dedicated for this purpose, which would be distributed together with the executable.

hvr:
fwiw, @ekmett mentioned on `#hackage` he hacked up a script which generates
output like  https://gist.github.com/ekmett/34024e26479763d77bc3e1ecaa24b819

Profpatsch:
A command line feature like this would be awesome, to see which package introduced which dependencies. For reference, here’s how `nix-store 1.11` implements it (though I’m not a fan of the naming at all).

```
OPERATION --QUERY
   Synopsis
       nix-store {--query | -q} {--outputs | --requisites | -R | --references |
[...]

PR #3202, opened 2016-03-01
Extend cabal list cmd

mtullsen:
Implements a solution to #879 by extending the cabal 'list' command with a --dependencies option.

I generalize the idea of listing the license field to listing multiple, arbitrary fields (a --field option) of the packages that are dependencies.

This doesn't address the extraction of License files (as discussed in #879).

### DIFF
--- a/cabal-install/Distribution/Client/List.hs
+++ b/cabal-install/Distribution/Client/List.hs
@@ -13,6 +13,8 @@ module Distribution.Client.List (
   list, info
   ) where
 
+import Distribution.Client.SetupWrapper
+         ( SetupScriptOptions(..), defaultSetupScriptOptions )
 import Distribution.Package
          ( PackageName(..), Package(..), packageName, packageVersion
          , Dependency(..), simplifyDependency
@@ -26,8 +28,11 @@ import Distribution.PackageDescription
 import Distribution.PackageDescription.Configuration
          ( flattenPackageDescription )
 
+import Distribution.ParseUtils ( fieldName, ppFields )
 import Distribution.Simple.Compiler
         ( Compiler, PackageDBStack )
+import Distribution.Simple.Configure (getPersistBuildConfig)
+import qualified Distribution.Simple.LocalBuildInfo
 import Distribution.Simple.Program (ProgramConfiguration)
 import Distribution.Simple.Utils
         ( equating, comparing, die, notice )
@@ -36,7 +41,7 @@ import Distribution.Simple.PackageIndex (InstalledPackageIndex)
 import qualified Distribution.Simple.PackageIndex as InstalledPackageIndex
 import qualified Distribution.Client.PackageIndex as PackageIndex
 import Distribution.Version
-         ( Version(..), VersionRange, withinRange, anyVersion
+         ( Version(..), VersionRange, withinRange, anyVersion, thisVersion
          , intersectVersionRanges, simplifyVersionRange )
 import Distribution.Verbosity (Verbosity)
 import Distribution.Text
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/List.hs
```diff
@@ -138,27 +235,39 @@ list :: Verbosity
...
-                                           ++ sourceVersions    pkg ]
-             -- Note: this only works because for 'list', one cannot currently
-             -- specify any version constraints, so listing all installed
-             -- and source ones works.
-      else
-        if null matches
-            then notice verbosity "No matches found."
-            else putStr $ unlines (map showPackageSummaryInfo matches)
+list verbosity packageDBs repoCtxt comp conf listFlags pats = do
+  unless (simpleOutput || null fields) $
```
23Skidoo: So just `--field=foo --simple-output` is allowed, even though it has no effect? From the error message it looks like the test should be `when (not.null fields && (not simpleOutput || not onlyDependencies)) $ die "..."`.

cabal-install/Distribution/Client/List.hs
```diff
@@ -129,6 +138,94 @@ getPkgList verbosity packageDBs repoCtxt comp conf listFlags pats = do
...
+          let
+              showPkgSimple pkg =
+                display (packageName pkg)
+                ++ " "
+                ++ display (packageVersion pkg)
+                ++ showPackageFields pkg
+
+              showPkgStd = showPackageSummaryInfo . mkDisplayInfo srcPkgIndex
+
+              pkgs = InstalledPackageIndex.topologicalOrder installedPkgIndex
```
23Skidoo: I don't think that this is correct, it will print all packages in the package index. I think you should use `dependencyClosure`.

cabal-install/Distribution/Client/List.hs
```diff
@@ -129,6 +138,94 @@ getPkgList verbosity packageDBs repoCtxt comp conf listFlags pats = do
...
+            else putStr $ unlines $ map showPkgStd    pkgs
+
+  where
+
+  -- flags:
+  simpleOutput = fromFlag (listSimpleOutput listFlags)
+  fields       = fromNubList $ listFields listFlags
+
+  -- Left if any of the fields are bad, otherwise Right with the show function:
+  possiblyShowPackageFields :: Either String
```
23Skidoo: This feels a bit overcomplicated. Why not just build a `Map String (FieldDescr InstalledPackageInfo)` from `fieldsInstalledPackageInfo`? Then the code would look roughly like this:

```
fieldDescrMap <- buildFieldDescrMap
validateFieldNames fieldDescrMap fields
showPackageFields <- mkShowPackageFields fieldDescrMap
```

cabal-install/Distribution/Client/List.hs
```diff
@@ -129,6 +138,94 @@ getPkgList verbosity packageDBs repoCtxt comp conf listFlags pats = do
...
+    where
+    getFieldDescr nm =
+      case [ fd | fd <- Installed.fieldsInstalledPackageInfo
+                , nm == fieldName fd ] of
+         []   -> Left nm
+         fd:_ -> Right fd
+
+    showFields fds = render . nest 2 . ($+$ text "") . ppFields fds
+
+  mkDisplayInfo :: PackageIndex.PackageIndex SourcePackage
```
23Skidoo: This function looks like something that could be reused in other parts of the file.

IvanMalison:
Whats the status of this PR?

23Skidoo:
Looks like the author never addressed the review points. I'm inclined to close, but anyone is willing to take over and finish the patches.

gbaz:
We should probably close. There's a different path to some of this now through the plan.json stuff, and arguably that can be extended to the rest. I'm closing now, but if anyone objects speak up.

PR #3208, opened 2016-03-03
Improve goal reorder heuristics.

kosmikus:
This change primarily does two things:

1. For `--reorder-goals`, we use a dedicated datatype `Degree`
   rather than an `Int` to compute the approximate branching
   degree. We map 0 and 1 to the same value. We then use a
   lazy ordering and a shortcutting minimum function to look
   for the "best" goal.

   The motivation here is that we do not want to spend
   unnecessary work. Following any goal that has 0 or 1 as degree
   cannot really be "wrong", so we should not look at any others
   and waste time.

   This will still not always make the use of `--reorder-goals`
   better than not using it, but it will reduce the overhead
   introduced by it.

2. We use partitioning rather than sorting for most of the other
   goal reordering heuristics that are active in all situations.
   I think this is slightly more straightforward and also slightly
   more efficient, whether `--reorder-goals` is used or not.

I have run some preliminary performance comparisons and they
seem to confirm that in both cases separately (with or without
`--reorder-goals`), these changes are a relative improvement
over the status quo. I will run additional tests before
merging this into master.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency/Modular/PSQ.hs
+++ b/cabal-install/Distribution/Client/Dependency/Modular/PSQ.hs
@@ -1,21 +1,29 @@
 {-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}
 module Distribution.Client.Dependency.Modular.PSQ
     ( PSQ(..)  -- Unit test needs constructor access
+    , Degree(..)
     , casePSQ
     , cons
+    , degree
     , delete
+    , dminimumBy
     , length
-    , llength
     , lookup
     , filter
     , filterKeys
+    , firstOnly
     , fromList
+    , isZeroOrOne
     , keys
     , map
     , mapKeys
     , mapWithKey
     , mapWithKeyState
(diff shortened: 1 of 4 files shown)


hvr:
I've ran a few random tests for cases which stress the solver a bit (solving a full stackage-set, trying to find (when there isn't any) install plans for packages for GHC8, etc..), and they all show a significant improvement of `--reorder-goals`. This looks very promising!

kosmikus:
@23Skidoo After running some tests, it still looks like this is a clear improvement, and I haven't found any result deviations either. Are you ok if I merge this? I have no idea what your policies for 1.24 are, but I wouldn't mind having it go in that release as well.

23Skidoo:
@kosmikus Sure, please do. It'd be also nice to include #2916 and #2914 in 1.24.

kosmikus:
@23Skidoo Thanks. And agreed about #2916 and #2914.

23Skidoo:
Cherry-picked into the 1.24 branch.

PR #3210, opened 2016-03-05
Parameterize SourcePackage and ConfiguredPackage on package location

BardurArantsson:
This is a step towards breaking the dependency from the modular
solver on cabal-install.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -73,6 +73,7 @@ import Distribution.Client.PkgConfigDb (PkgConfigDb)
 import Distribution.Client.Types
          ( SourcePackageDb(SourcePackageDb), SourcePackage(..)
          , ConfiguredPackage(..), ConfiguredId(..)
+         , UnresolvedPkgLoc
          , OptionalStanza(..), enableStanzas )
 import Distribution.Client.Dependency.Types
          ( PreSolver(..), Solver(..), DependencyResolver, ResolverPackage(..)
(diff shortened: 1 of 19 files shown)


BardurArantsson:
I should say: The reason we want to break the "PackageLocation" dependency is that that brings in a lot of unnecessary baggage, namely URI (i.e. the `network` package), etc. I can't recall all the details, but ISTR it also brings in other parts of cabal-install which aren't needed for the solver.

BardurArantsson:
Most of the remainder of splitting out the solver into its own library "between" Cabal and cabal-install is just moving files around and splitting up `Distribution.Client.Types` appropriately.

I think we can hold off on that for a bit if we still want to keep master close to 1.24.x for a while. This bit is the most code-edit heavy bit.

23Skidoo:
Looks harmless enough, the only complaint is bikeshedding: the `PackageLocation'` type name is pretty uninformative. Maybe we could have something like `UnresolvedPackageLocation = PackageLocation (Maybe FilePath)` and `ResolvedPackageLocation = PackageLocation FilePath` instead? Better naming suggestions welcome, of course.

BardurArantsson:
That's a good point. I'll have a look at it tomorrow.

BardurArantsson:
Looking over the other code, we have three variants:

- PackageLocation (Maybe FilePath)
- PackageLocation FilePath
- PackageLocation ()

I'm not sure what that last one really represents.

My only concern about doing what you suggest is really that the names are a bit long... though obviously not as long as spelling out the full types. Would `UnresolvedPkgLoc` and `ResolvedPkgLoc` be acceptable? (Or maybe spell out Location in full?)

(BTW, if we're going to be systematic about it, this does involve changes in a few further places that this patch doesn't currently touch. It's not too bad, but worth mentioning.)

BardurArantsson:
OK, so I've changed this to two commits:

- The first commit was amended this: Renamed `PackageLocation'` to `UnresolvedPkgLoc` and replaced stray instances of `PackageLocation (Maybe FilePath)` with the `UnresolvedPkgLoc` alias.
- The second commit adds `ResolvedPkgLoc` and replaces instances of `PackageLocation FilePath` with a reference to the alias.

Hopefully that should be it.

Let me know if you'd like the longer names -- it's pretty trivial to do a global search/replace.

ezyang:
Sorry for ALL THE BIKESHEDS, but I'd be keen for an alias on `SourcePackage UnresolvedPkgLoc` too.

BardurArantsson:
Can we please do that in a follow-up? I'd really like to get this merged ASAP since it's pretty prone to conflicts with other stuff.

BardurArantsson:
@23Skidoo OK to merge?

23Skidoo:
@BardurArantsson 
> `PackageLocation ()`

I think we can leave this one as it is.

23Skidoo:
Merged, thanks!

PR #3212, opened 2016-03-06
Introduce UnresolvedSourcePackage alias

BardurArantsson:
The alias is a shorthand for 'SourcePackage UnresolvedPkgLoc' which is
used all over the place.

---
Per @ezyang's request in #3210.

I'm a bit torn on whether this should be shortened to `UnresolvedSrcPkg`, but decided against it based on the fact that we still have `SourcePackage` around. Opinions?


BardurArantsson:
Ah, thanks for the title correction. I've pushed a new revision with updated commit message.

23Skidoo:
Merged, thanks!

PR #3213, opened 2016-03-08
update -prof Check message

gbaz:
The flag enable-executable-profiling warns to change it to just enable-profiling. So we'll fix that here too so we don't send people down an old path.


23Skidoo:
Merged, thanks!

PR #3218, opened 2016-03-10

23Skidoo:
See https://github.com/haskell/cabal/pull/3197#issuecomment-194567139.

Please don't merge until Travis is green.


23Skidoo:
Great, now it passes.

PR #3219, opened 2016-03-11

23Skidoo:
Same fix as #3218, but now for `master` instead of `1.24`.


23Skidoo:
Looks like `cabal test` needs an analogue of `filterConfigureFlags`.

23Skidoo:
The question is why this doesn't fail on the `1.24` branch.

PR #3235, opened 2016-03-19
Add current pkg info to generated `cabal_macros.h` file

erikd:
Add the version string macro for the package currently being built
to the generated `dist/build/autogen/cabal_macros.h` file. Perviously
this file only contained the version strings of all the dependent
packages, but not the version string of the package itself.


23Skidoo:
This will also generate the `MIN_VERSION_foo` macro for the current package, but that's probably OK.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 1.24.

PR #3245, opened 2016-03-29
Dependency solver quickcheck tests

grayjay:
I started adding quickcheck tests for the solver.  There is currently only one test, which checks that parameters like `--reorder-goals` do not affect whether the solver finds an install plan.
 
Here is an example failure: https://gist.github.com/grayjay/86eea907cd08878855ae

I created a new test suite because the test is currently too slow to run as a unit test.  The time to find one of the known bugs on `master` ranges from under a second to over a minute on my computer, with a median of about 6 seconds (10,000 tests).  We may want to run it even longer after those bugs are fixed.  Another problem is that the test occasionally runs out of memory.  I haven't tried to debug it yet.

There are two other issues.  I'm not sure how we should print failing test cases in a human-readable format.  The PR currently uses the `pretty-show` package to parse and reformat the results from the derived `Show` instance, but it requires `happy`.  I'm also not sure how to organize the test suite.  I put the new test in the `UnitTests` directory since it uses the existing `DSL` module.  Another option is to have three directories under `tests`, one each for unit tests, solver quickcheck tests, and shared modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,11 +73,13 @@ script:
  - cp Setup.hs ./dist/setup/setup.hs
  - ghc --make -odir ./dist/setup -hidir ./dist/setup -i -i. ./dist/setup/setup.hs -o ./dist/setup/setup -Wall -Werror -threaded  # the command cabal-install would use to build setup
 
+ - cabal install happy
  - cabal install --only-dependencies --enable-tests --enable-benchmarks
  - ./dist/setup/setup configure --user --ghc-option=-Werror --enable-tests --enable-benchmarks -v2 # -v2 provides useful information for debugging
  - ./dist/setup/setup build
  - ./dist/setup/setup haddock # see https://github.com/haskell/cabal/issues/2198
- - ./dist/setup/setup test --show-details=streaming --test-option=--hide-successes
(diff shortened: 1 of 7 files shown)


BardurArantsson:
Re: Showing output/data: [present](https://hackage.haskell.org/package/present) looks pretty nice, but the GHC version requirement may be too steep :(. (It seems to require GHC 7.10.x.)

kosmikus:
This looks quite interesting. Regarding the actual failure: can I assume that you only get failures for test cases involving setup dependencies?

23Skidoo:
LGTM, I think this can be merged.

Re: dependency on `happy`, I think it's fine in the case of the test suite.

grayjay:
@BardurArantsson I'll continue using `pretty-show` since the dependency on `happy` isn't a problem, but thanks for the suggestion.

@kosmikus I just ran the test for over an hour without `--independent-goals` or setup dependencies, and there were no failures.  I also didn't notice any other types of failure while working on the test.

@23Skidoo Thanks! Would you like me to merge it?

23Skidoo:
@grayjay Do the honours :-)

PR #3249, opened 2016-03-30

ezyang:
Commandeered from #3219


ezyang:
Yeah, I guessed this was the case. If cabal-install does not have a custom-setup, on master this means we'll use an old version of the Cabal library to build the test library, which means that it didn't work. On master (since I recently merged a patch to add custom-setup to cabal-install) this works fine.

Issue #220, opened 2012-05-24
cabal-install doesn't consider build-tools as dependencies


bos:
**Summary by @ezyang.** When you write `build-tools: alex`, Cabal only checks that an executable of this name is installed; `cabal-install` does not know that if there is no such executable, `alex` needs to be installed. One complication is that `build-tools` does not necessarily correspond to a package, so it is not a priori obvious to `cabal-install` what should be installed to get an executable.

Here is **Proposal A: Build-tools only**

* We introduce a new type of qualified goal, a `build-tools` goal, similar to how `setup` qualified goals work.
* We modify the solver to interpret each field of `build-tools` as a package name, and we solve for each build-tool as a separate `build-tools` dependency (so we don't do something goofy like try to solve for `happy` and `alex`'s dependency sets consistently when they really couldn't care less.)
* Some `build-tools` fields don't correspond to actual packages. We give two escape hooks in this situation. First, Cabal has an internal mapping of build tools to `Maybe PackageName`.  If `Nothing`, this indicates that the build-tool is not managed by Cabal and we shouldn't try to solve for it. If `Just pn`, then it means that installing `pn` will also install the requested executable. This mapping will be configurable using the `build-tool-packages` field in a `cabal-project` and also modifiable by command line argument. (Names can be bikeshedded of course.) Our second escape hook is, if a `build-tool` refers to an unknown package name, we print a warning but continue to solve, simply ignoring that package. A user can suppress the warning by disabling this using the aforementioned flag.

Consider these examples:

* `build-tools: mytoolpkg`: cabal-install will install all executables defined by a package named `mytoolpkg`, and ensure they are available prior to building this package.
* `build-tools: gtk2hsc2hs`: cabal-install is hard-coded to know that the `gtk2hsc2hs` executable is provided by the `gtk2hs-buildtools`, so it ensures that it builds and installs all executables from the package before building this package.
* `build-tools: someprog`, plus a flag `--provides-build-tool someprog=somepkg`, cabal-install sees `someprog` (which has no package), but using the flag mapping it knows that `somepkg` is the package that provides this build tool.

Here is **Proposal B: tool-depends**

* We introduce a new type of qualified goal, a `build-tools` goal, similar to how `setup` qualified goals work. (Like before)
[...]

PR #3251, opened 2016-03-30
Testing with GHC 7.0 and 7.2 AND some improvements to package-tests

ezyang:
In the process, I refactored the test suite a bit to be more robust
and handle the --with-compiler more correctly in most cases.

One note: I turned off the HPC tests with the GHC and WITH_GHC
parameters don't match.  In principle this should work but there
is some sort of bug.  I don't plan on fixing the bug.

UPDATE: This patch set now also has some refactorings to try to eliminate the LBI parsing nonsense, by using `build-tools` as per #220.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -346,6 +346,14 @@ library
   if !impl(ghc >= 7.0)
     default-extensions: CPP
 
+executable cabal-setup
+  build-depends: base, Cabal
+  main-is: Setup.hs
+  -- Actually, we don't really want anything from tests; just
+  -- want to make sure we don't pick up source files in .
+  hs-source-dirs: tests
+  default-language: Haskell98
+
 -- Small, fast running tests.
 test-suite unit-tests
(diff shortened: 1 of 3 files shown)


ezyang:
One downside with this patch is it increases the runtime of all tests as we repeatedly run the test suites from 7.0 and 7.2 for every built version. Perhaps we should have a dedicated 7.0/7.2 test ways, which build using some picked version of GHC (latest version?) and then test against 7.0/7.2. It wouldn't be necessary to even bother building test cabal-install, in this case.

23Skidoo:
That'd be nicer, I think. Hopefully that doesn't make `.travis.yml` overcomplicated.

ezyang:
It looks like, to do this, I'm going to have to take the `script` contents and put them in their own shell script; otherwise I can't write a big enough conditional. If people are OK with this I can do it.

23Skidoo:
Yes, I think it's fine. Some small parts of the build script already live under `Cabal/misc`.

PR #3252, opened 2016-03-30
Miscellaneous refactorings

ezyang:


Cabal/Distribution/Package.hs
```diff
@@ -115,6 +119,25 @@ instance NFData PackageIdentifier where
 -- * Component Source Hash
 -- ------------------------------------------------------------
 
+data Module =
```
23Skidoo: Can you please add a Haddock comment here?

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -489,8 +535,10 @@ configure (pkg_descr0', pbi) cfg = do
          ++ "supported by " ++ display (compilerId comp) ++ ": "
          ++ intercalate ", " (map display exts)
 
-    -- configured known/required programs & external build tools
-    -- exclude build-tool deps on "internal" exes in the same package
+    -- Configure known/required programs & external build tools.
+    -- Exclude build-tool deps on "internal" exes in the same package
+    --
+    -- ToDo: Factor this into a helper package.
```
23Skidoo: `TODO`s in existing code are written in all caps: let's be consistent, it makes them easier to grep for.

Cabal/Distribution/Simple/LocalBuildInfo.hs
```diff
@@ -325,25 +325,25 @@ data ComponentLocalBuildInfo
     componentCompatPackageName :: PackageName,
     componentExposedModules :: [Installed.ExposedModule],
     componentIsPublic :: Bool,
-    componentPackageRenaming :: Map PackageName ModuleRenaming
+    componentIncludes :: [(UnitId, ModuleRenaming)]
```
23Skidoo: Would be nice to have a comment here saying what this field is for.

23Skidoo:
LGTM modulo minor comments.

PR #3254, opened 2016-03-30
Axe getnumcores, as we don't support pre-6.12 anymore.

ezyang:


23Skidoo:
What about GHC < 7.6 on ARM? I think it still needs this workaround. Otherwise LGTM.

ezyang:
OK, I looked at it more carefully and I guess we should just not specify `-threaded` in this situation.

ezyang:
OK, updated. Once it goes green I'll push.

Issue #3257, opened 2016-03-31
Testing plan for cabal-install and convenience libraries


ezyang:
Convenience libraries need some integration tests with cabal-install to make sure, e.g., cabal-install's dependency solver and builder can handle them correctly.

In the solver:

- [ ] When public library depends on internal library, dependencies of internal library should be considered.
- [ ] Conversely, when public library does NOT depend on library, dependencies of internal library should only be considered if the buildable stanza associated with it (e.g., a test suite or executable) is being built.

(TODO: more)

Issue #1541, opened 2013-10-13
Haskell source pre-processing in Haskell no longer works


ygale:
The standard technique for pre-processing Haskell source code using Haskell has stopped working in recent versions of Cabal. The technique is as follows:

The cabal file is structured like this:

    Name: foo
    ...
    Executable foo-meta
      ...
    Library
      ...
    Executable foo
      ...

Then you pass "`-F -pgmF dist/build/foo/foo-meta`" to GHC for the source file you want to pre-process, e.g., using an `OPTIONS_GHC` pragma.

The effect is that first cabal builds `foo-meta`, then uses it to build the library, then builds the final executable.

But now, Cabal refuses to build the artifacts in the order I specified in the Cabal file. It always tries to build the library first, before the `foo-meta` executable, and the build fails.

Note: We have only tried this so far inside a sandbox. However, we have verified that the problem is definitely that cabal is not even trying to build `foo-meta`, not a problem with differences in the path to the `foo-meta` executable in a sandbox.

ezyang:
This is a little known feature of Cabal, not mentioned in the manual, and rarely known by many who have this problem in the wild.

Additionally, it would be quite nice if we could support the preprocessor use-case without needing people to add a custom Setup.hs. Perhaps we can support `-pgmF` style: what if Cabal adds every mentioned build-tool to the PATH?

PR #3259, opened 2016-03-31
Fix build-tools ordering regression (#3257, #1541)

ezyang:
When converting the component graph to operate in terms of UnitIds
instead of CNames I accidentally introduced a regression where we
stopped respecting build-tools when determining an ordering to
build things.  This commit fixes the regression (though perhaps
not in the most clean/performant way you could manage it.)  It
also fixes a latent bug if internal libraries aren't processed
in the correct order.

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -127,6 +127,8 @@ import Text.PrettyPrint
 import Distribution.Compat.Environment ( lookupEnv )
 import Distribution.Compat.Exception ( catchExit, catchIO )
 
+import Data.Graph (graphFromEdges, topSort)
+
 -- | The errors that can be thrown when reading the @setup-config@ file.
 data ConfigStateFileError
     = ConfigStateFileNoHeader -- ^ No header found.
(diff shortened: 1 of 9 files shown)


Cabal/Distribution/Simple/Configure.hs
```diff
@@ -1592,11 +1600,16 @@ mkComponentsLocalBuildInfo cfg comp installedPackages pkg_descr
...
-        let dep_uids = map fst (filter (\(_,e) -> e `elem` internalPkgDeps)
-                                       (componentPackageDeps clbi))
+        -- NB: We want to preserve cdeps because it contains extra
+        -- information like build-tools ordering
+        let dep_uids = [ componentUnitId dep_clbi
+                       | cname <- dep_cnames
+                       -- Being in z relies on topsort!
+                       , (dep_clbi, _) <- z
+                       , componentLocalName dep_clbi == cname ]
+        print (componentLocalName clbi, dep_cnames, dep_uids)
```
23Skidoo: Forgotten debug output?

23Skidoo:
Looks OK.

23Skidoo:
And thanks for adding that test.

Cabal/Distribution/Simple/Configure.hs
ezyang: Yes, I'll kill it.

PR #3260, opened 2016-03-31
Have Cabal build a Setup executable and test with it.

ezyang:
I recently discovered that build-tools can be used to specify
dependencies on internal executables (#220).  This means that
we can make a nice improvement to the package-tests test suite:
instead of manually building a Setup.hs script ourselves,
we can add one to the Cabal file for Cabal, and have the
test suite have a build-tools dependency on it.

Perhaps the only objection to this is that the Cabal *library*
now has an executable cabal-setup.  Maybe with some buildable
shenanigans we can make it so that we never build this executable
unless the user asks for it, or the test suite is enabled.
Alternately, cabal-setup is a generally handy executable to have
around.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -342,6 +342,14 @@ library
   if !impl(ghc >= 7.0)
     default-extensions: CPP
 
+executable cabal-setup
+  build-depends: base, Cabal
+  main-is: Setup.hs
+  -- Actually, we don't really want anything from tests; just
+  -- want to make sure we don't pick up source files in .
+  hs-source-dirs: tests
+  default-language: Haskell98
+
 -- Small, fast running tests.
 test-suite unit-tests
   type: exitcode-stdio-1.0
@@ -380,6 +388,7 @@ test-suite package-tests
     PackageTests.TestSuiteTests.ExeV10.Check
     PackageTests.PackageTester
   hs-source-dirs: tests
+  build-tools: cabal-setup
   build-depends:
     base,
     containers,
--- a/Cabal/tests/PackageTests.hs
+++ b/Cabal/tests/PackageTests.hs
@@ -187,10 +187,6 @@ main = do
                     (showPackageDbList (uninterpretPackageDBFlags
                                         packageDBStack0)))
 
-    -- Create a shared Setup executable to speed up Simple tests
-    putStrLn $ "Building shared ./Setup executable"
-    rawCompileSetup verbosity suite [] "tests"
-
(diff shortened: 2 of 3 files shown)


Cabal/Cabal.cabal
```diff
@@ -342,6 +342,14 @@ library
   if !impl(ghc >= 7.0)
     default-extensions: CPP
 
+executable cabal-setup
```
23Skidoo: This means that `cabal install Cabal` will now also install `cabal-setup`. Would be nice to avoid that, perhaps by building `cabal-setup` only when a flag is enabled.

ezyang: Still testing, but how does something like this look:

```
diff --git a/Cabal/Cabal.cabal b/Cabal/Cabal.cabal
index 494eb97..b182bc6 100644
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -212,6 +212,9 @@ source-repository head
 flag bundled-binary-generic
   default: False
 
+flag cabal-setup-buildable
+  default: False
+
 library
   build-depends:
     array      >= 0.1 && < 0.6,
@@ -347,6 +350,10 @@ library
     default-extensions: CPP
 
 executable cabal-setup
+  if flag(cabal-setup-buildable)
+    buildable: True
+  else
+    buildable: False
   build-depends: base, Cabal
   main-is: Setup.hs
   -- Actually, we don't really want anything from tests; just
@@ -392,7 +399,11 @@ test-suite package-tests
     PackageTests.TestSuiteTests.ExeV10.Check
     PackageTests.PackageTester
   hs-source-dirs: tests
-  build-tools: cabal-setup
+  if flag(cabal-setup-buildable)
+    buildable: True
+    build-tools: cabal-setup
+  else
+    buildable: False
   build-depends:
     base,
     containers,
```

23Skidoo: Yep, that's what I had I mind. Is Cabal smart enough to see that `--enable-tests` must imply `cabal-setup-buildable: True`? Then maybe the second conditional could be elided.

ezyang: It's not, because `build-tools` are not considered during dependency resolution; this is bug #220

ezyang: Ugh, this interacts poorly with `new-build`, now it thinks that the package has no component named `package-tests`.

ezyang:
OK, well, this patch doesn't simplify that much, so I guess if the added executable is a deal breaker lets not put it in.

ezyang:
Canned.

PR #3266, opened 2016-04-01
Revert separate shell script but preserve old GHC testing.

ezyang:
I stared at the new Travis output, and decided I didn't like it: we lose coloration for individual exit codes. So away goes the shell script.


ezyang:
Appveyor seems to be flaking right now.

BardurArantsson:
Wouldn't it be sufficient to add "-x" to the sh command line? (Yes, coloration would be gone, but you could still see pretty easily which command failed.)

ezyang:
@BardurArantsson So, actually, the shell script I had used -v, so you could see the commands being run as they were done, and it used -e so it stopped executing on the first failed command. I can see merits of going either way: if we stop executing immediately on failure, we learn about failed builds more quickly but you don't get to see what happens afterwards (so you might end up having to submit to Travis multiple times to keep fixing errors... that's true for -Werror anyway though). I guess if we fail on the first failed command, you can just scroll to the end of the log and that's the error, no need for coloring.

BardurArantsson:
Ah, I missed the "-v" bit. (I think the canonical thing here is "-x" and "-v" just so happens to work with bash, but I may be mistaken.)

IME getting the first error (without subsequent errors) is much more valuable than wasting CPU time on trying subsequent commands that, most likely, won't work anyway. (To be clear: there *are* exceptions for example in our own `setup-dev.sh` script, the "test" commands will often fail spuriously, but the rest of the build will work. My preferred solution that that is to explicitly allow failure by appending a "|| true" to commands that may fail spuriously.)

I guess in the end it's just preference.

ezyang:
@BardurArantsson There's a slight difference between -x and -v, which is that -v outputs the shell script as it is read in (so what you source level), while -x actually prints out the command that is being run, with things like variables expanded. But yes, they do similar things.

In that case, I think I'm going to kill this PR; I think I am convinced to like the new output.

BTW, for inline Travis you can get fail fast: https://github.com/travis-ci/travis-ci/issues/1066 by just setting "set -e". But you can't easily NOT fail fast with a shell script, since the exit code won't get propagated by default.

Issue #3269, opened 2016-04-02
Planned improvements to cabal-install integration-tests


ezyang:
I was using `integration-tests` and was annoyed by a few things, so I'm planning to fix them. Here is my work plan:

- [x] I'm going to follow @phadej's suggestion to just use `!` to invert the exit code of command which are expected to fail. Then I'll eliminate `should_run`/`should_fail` directories, so we just have a single level of hierarchy. This makes tests which want to work with the same packages but want to test both successful and failing variants more convenient.
- [x] It's annoying to copy paste the same `common.sh` for each category. So I'm going to have a single global common script in `IntegrationTests` which gets copied into all scripts. It will just get pasted into the category directory.
- [ ] Something a little annoying is that I can't run the test script in the temporary directory directly to rerun the test, since the envvars aren't set.  So, instead of passing through the environment variables implicitly, I'll write them to a file. So the expectation is you also write `. ./env.sh` at the top of your scripts.
- [ ] I don't really like `RE:`. It's annoying to write, and it means we can't implement `accept` later. Furthermore, we're only using it to munge cabal.exe to cabal. So I'm going to replace it with a direct munge.
- [ ] The idiom `>/dev/null` is really irritating for debugging, since it means we suppress the output. Cabal's package tests does a lot better here: we turn on verbose by default (for good test logs) and we don't dev null any output. I think perhaps the right thing to do here is to `tee` the output we want to compare against to a log file, and then employ a naming convention for checking these log files (similar to how in package tests we only test the stdout/stderr of a few, important command)
- [ ] The de-interleaving of stderr and stdout is a pain. For display, put them together, and also say which command was run before each.

For now I'm just planning to do the first two, since they seem like obviously good improvements.

CC @BardurArantsson

BardurArantsson:
Not sure what you mean about the "RE:"/munging, but FWIW everything else sounds sensible.

ezyang:
@BardurArantsson I looked at all use sites of `RE:` and it seems to be purely for ensuring we are indifferent to `cabal` versus `cabal.exe` in output. If that's the only case it's used, we can just add a special case to the comparator to substitute occurrences of `cabal.exe` with `cabal` and then drop the regex.

I added one more work item, which is the fact that `>/dev/null` is bad for good test log output.

23Skidoo:
@ezyang 
>  it means we can't implement `accept` later.

What is `accept`?

ezyang:
[...]

PR #3270, opened 2016-04-02
Improvements to cabal-install integration-tests #3269

ezyang:


BardurArantsson:
LGTM, but should probably be squashed before merge.

PR #3272, opened 2016-04-02
Respect "--" in bash-completion, fixes #3271

lspitzner:
(I looked at `git`'s bash-completion, and simplified slightly.)

Issue #3271, opened 2016-04-02
Bash completion triggers execution


Tuplanolla:
There is a bug in the file `cabal-install/bash-completion/cabal` that
causes incomplete and potentially dangerous commands to be executed when
the shell is requested to complete them.
Typing in `cabal run --`, adding an argument to the end and
trying to complete it by hitting Tab twice triggers the bug.
The behavior is as follows.

    $ cat > Main.hs
    module Main where

    import System.Environment

    main :: IO ()
    main = getArgs >>= putStrLn . ("(" ++) . (++ ")") . unwords
    $ cabal init --is-executable --non-interactive

    Guessing dependencies...

    Generating LICENSE...
    Warning: unknown license type, you must put a copy in LICENSE yourself.
    Generating Setup.hs...
    Generating bug.cabal...

    Warning: no synopsis given. You should edit the .cabal file and add one.
    You may want to edit the .cabal file and add a Description field.
    $ touch LICENSE
    $ cabal run -- bug with arguments
    (with            Running          bug-0.1.0.0...   for
    --list-options)  arguments        bug...
    Preprocessing    bug              executable
    $ cabal run -v0 -- bug with arguments
    (with            --list-options)  arguments

PR #3273, opened 2016-04-02
Fix bash-completion after "--"

lspitzner:
#3272 accidentally disabled completions at the start of flags,
e.g. on "cabal install --".
[CI SKIP]

(I shouldn't have considered more why git did not use a for-loop - because it was not a for loop.)

(Tested all of "cabal run", "cabal run -", "cabal run --", "cabal run -- ". This change does not seem to break #3271 again.)

Sorry for the noise.

### DIFF
--- a/cabal-install/bash-completion/cabal
+++ b/cabal-install/bash-completion/cabal
@@ -55,11 +55,16 @@ _cabal_subcommands()
 
 __cabal_has_doubledash ()
 {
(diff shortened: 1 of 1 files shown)


23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked both patches into 1.24.

PR #3276, opened 2016-04-03
Use unique component names in solver quickcheck test

grayjay:


ezyang:
Is it an error if the arbitrary instance produces the same name twice? If that's the case, maybe some sort of deterministic name generation would be better.

ezyang:
Is it an error if the arbitrary instance produces the same name twice? If that's the case, maybe some sort of deterministic name generation would be better.

grayjay:
Duplicate tests within a package started causing an error in a recent commit.  I'm not sure how to generate names deterministically without adding a lot of complexity, but I'll look into removing the duplicates.

ezyang:
This is not necessarily a thing you should do, but one way to do this is to pass in a parameter to the arbitrary functions specifying the set of names "in scope", plus some sort of candidate name generator e.g. ``candidateNames = [1..] >>= (`replicateM` ['a'..'z'])`` (c.f.  http://stackoverflow.com/q/9542313/23845). To then select a fresh name you simply filter out candidates which are in scope ` head (filter (`Set.notMember` in_scope) candidateNames)`. Of course you need to plumb the scoping parameter around. I've done this before when writing an arbitrary instance for lambda calculus terms, where I wanted some variables fresh and others bound.

grayjay:
@ezyang Thanks for the explanation.  I looked at the failure more carefully and realized that `ComponentDeps` should handle components with the same name by merging them.  The reason that duplicate tests caused a failure is that test suites were handled differently from library and setup components.  I refactored it in #3277.

ezyang:
Great, always good when a test case is right and leads to a bugfix :+1:

grayjay:
I meant they were handled differently in the solver DSL.

PR #3278, opened 2016-04-03
Delete ReadyPackage

ezyang:
Delete ReadyPackage.

Previously, ReadyPackage was a ConfiguredPackage elaborated with
a dependencies data structure which had InstalledPackageInfo
rather than ConfiguredId.  Well, it turned out that we only
used the data from ConfiguredId!  So ReadyPackage is completely
useless.

I retained some type synonyms to avoid changing type signatures,
and as a reminder that not all ConfiguredPackages can be built,
only the ready ones.

There is also some extra documentation in the rest of the patchset.

### DIFF
--- a/cabal-install/Distribution/Client/Types.hs
+++ b/cabal-install/Distribution/Client/Types.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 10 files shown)


23Skidoo:
Seems to fail on Travis/AppVeyor.

23Skidoo:
> Well, it turned out that we only used the data from ConfiguredId! So ReadyPackage is completely useless.

Can't we fix this by making `ReadyPackage` a `newtype` then? I still like the `ReadyPackage`/`ConfiguredPackage` distinction, it makes the intent clearer.

ezyang:
I don't think the newtype would add much (besides line noise), but if you insist, I can reintroduce a newtype.

cabal-install/Distribution/Client/Types.hs
```diff
@@ -163,29 +163,7 @@ instance PackageFixedDeps (ConfiguredPackage loc) where
...
-         PackageFixedDeps (GenericReadyPackage srcpkg ipkg) where
-  depends (ReadyPackage _ deps) = fmap (map installedUnitId) deps
-
-instance HasUnitId srcpkg =>
-         HasUnitId (GenericReadyPackage srcpkg ipkg) where
-  installedUnitId (ReadyPackage pkg _) = installedUnitId pkg
-
-instance (Binary srcpkg, Binary ipkg) => Binary (GenericReadyPackage srcpkg ipkg)
-
+type ReadyPackage = ConfiguredPackage UnresolvedPkgLoc
```
23Skidoo: Please keep the Haddock comment.

23Skidoo:
Please do. Since we're not removing the `ReadyPackage` type completely, I think it's better to get some additional type safety.

ezyang:
OK, newtyped.

23Skidoo:
Thanks! Looks good to go now.

ezyang:
I'll wait for it to green.

ezyang:
I'm really confused: the error message in the Travis build appears to be for a stale version of the commit.

PR #3282, opened 2016-04-05
Delete FakeMap

ezyang:
This is only the last patch; the rest is #3281.

Compute 'UnitId' when we compute a 'ConfiguredPackage';
consequently, we can eliminate 'FakeMap' (finally!)
There is one hack remaining, which is that 'SolverInstallPlan'
gins up fake unit IDs so that it can be keyed on UnitIds.
But this data structure exists only very briefly before
being converted into an 'InstallPlan' or 'ElaboratedInstallPlan'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ progress.txt
 
 # test files
 dist-test
+register.sh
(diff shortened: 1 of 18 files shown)


ezyang:
Goingt or eopen

PR #3238, opened 2016-03-21
Require QuickCheck >= 2.8.2 for cabal-install unit tests

grayjay:
The `ProjectConfig` tests use the Arbitrary instance for Map added in 2.8.2.

PR #3285, opened 2016-04-05
Require QuickCheck >= 2.8.2 for cabal-install unit tests

grayjay:
The 'ProjectConfig' tests use the Arbitrary instance for Map added in 2.8.2.

(cherry picked from commit 68ef50116132cbcfae3857d46540a313d717df4a)

This commit is from #3238, but it also applies to 1.24.


ezyang:
OK.

23Skidoo:
Merged, thanks!

Issue #3203, opened 2016-03-02
Sandbox targets can prevent reinstallation of packages with setup dependencies


grayjay:
While experimenting with the `custom-setup` section, I was unable to reinstall a package that depended on two versions of another package.  The list of targets in the `-v3` log contained all of the installed packages, as well as a second occurrence of the main package.  The extra target seemed to prevent the solver from choosing setup dependencies.

I wrote some tests that demonstrate the behavior.  The unit tests show the effect of duplicate targets, and the integration test reproduces the problem in a sandbox.  I'm not sure if this is a bug in sandboxes or the solver.

EDIT: The integration test installs the two dependencies in the sandbox, but I got a similar error when they were installed globally.

Tests: https://github.com/grayjay/cabal/commit/42a8bd1833a8d887e7d24675f162ae83058b4d62
Log: https://ci.appveyor.com/project/grayjay/cabal/build/1.0.78

Unit test database:
```Haskell
  [
       Right $ exAv "A" 1 [ExAny "B"] `withSetupDeps` [ExFix "C" 1]
     , Right $ exAv "B" 1 []          `withSetupDeps` [ExFix "C" 2]
     , Right $ exAv "C" 1 []
     , Right $ exAv "C" 2 []
  ]
```

Installing A and B:
```
targets: A, B
[...]
[__0] trying: A-1.0.0 (user goal)
[__1] trying: B-1.0.0 (user goal)
[__2] next goal: A-setup.C (dependency of A-1.0.0)
[__2] rejecting: A-setup.C-2.0.0 (conflict: A => A-setup.C==1.0.0)
[__2] trying: A-setup.C-1.0.0
[__3] next goal: B-setup.C (dependency of B-1.0.0)
[__3] rejecting: B-setup.C~>A-setup.C-1.0.0 (conflict: B => B-setup.C==2.0.0)
[__3] trying: B-setup.C-2.0.0
[__4] done
```

Specifying each target twice:
```
targets: A, A, B, B
[...]
[__0] trying: A-1.0.0 (user goal)
[__1] trying: A~>A-1.0.0 (user goal)
[__2] trying: B-1.0.0 (user goal)
[__3] trying: B~>B-1.0.0 (user goal)
[__4] next goal: A-setup.C (dependency of A-1.0.0)
[__4] rejecting: A-setup.C-2.0.0 (conflict: A => A-setup.C==1.0.0)
[__4] trying: A-setup.C-1.0.0
[...]

PR #3290, opened 2016-04-06
Allow "cycles" through test suite dependencies (VERSION TWO)

ezyang:
This is a version of #3232 updated for HEAD.

CC @edsko


23Skidoo:
If the tests pass, can you please squash fix-up commits with the original ones? Would be nice to avoid having broken commits in history.

ezyang:
There's still a bug in this PR though, `cabal install` in the sandbox is not idempotent. In the integration test, if I run it the second time, here's what I get:

```
      A-2 *test (via: T-1) (reinstall) (changes: T-1 added)
      T-1 (via: A-2) (reinstall) (changes: A-1 -> 2)
```
@23Skidoo Of course.

grayjay:
The log from the third run looks like #3203.

ezyang:
(I guess, at some point, we should move this to a branch on `origin` so @edsko can push to it.)

23Skidoo:
So I think I won't delay the first 1.24 release for this. If @kosmikus thinks that this PR should go into 1.24, I think it can be made a part of a point release. Which means that 1.24 is now good to go.

kosmikus:
Independently of the first 1.24 release, I'd like to understand better what is going on here. There's quite a bit of history, and perhaps someone who has been following that more closely than I have can provide a good summary of what the status here is? @ezyang ?

ezyang:
I don't really know how the patch internally works, but I know that the obvious test for the feature (cabal installing a package with a cyclic dependency in a sandbox) doesn't work properly. So someone needs to dig in and figure out why it's failing and then fix it.

23Skidoo:
> the obvious test for the feature (cabal installing a package with a cyclic dependency in a sandbox) doesn't work properly. So someone needs to dig in and figure out why it's failing and then fix it.

@edsko, can you look into it?

edsko:
@23Skidoo the solver part of this is working, as far as I can tell. I'm not sure what's going on with the sandbox, some additional check somewhere I think. I don't know when I'll get a chance to look at that; perhaps somebody who is more familiar with the sandboxing stuff would be able to identify the problem more quickly?

23Skidoo:
> I don't know when I'll get a chance to look at that; perhaps somebody who is more familiar with the sandboxing stuff would be able to identify the problem more quickly?

I'll try to take a look at this next month.

hvr:
@ezyang I wanted to to try this PR to see whether it fixes the issues I have with a couple of packages (`time`, `bytestring`, `binary`, `cassava`, ...) but sadly this PR has bitrotted  and needs to be rebased with the merge-conflicts resolved...

23Skidoo:
Closing in favour of #3402.

23Skidoo:
@hvr See #3402 for an updated version.

hvr:
@23Skidoo thanks!

PR #3298, opened 2016-04-07
Try and enable Travis notifications

ezyang:


.travis.yml
```diff
@@ -37,3 +37,9 @@ matrix:
   allow_failures:
    - env: GHCVER=head
   fast_finish: true
+
+notifications:
+  irc:
+    channels:
+      - "chat.freenode.net##haskell-cabal"
```
23Skidoo: I thought we were using the `#hackage` channel?

ezyang: We can send to both. In any case, the notifications have turned out to be a bit less useful than I hoped, because (1) the IRC notifications don't get notifications for builds of PRs, and (2) you don't get the notification until the build is done, as opposed to when it is doomed (because one job failed.)

BTW, the Slack signup is https://haskell-cabal.slack.com/x-32997886514-32984962528/signup not sure what the best place to publicize is.

23Skidoo: Maybe the wiki front page?

ezyang: Ugh, it looks like I'll have to set up some autoinviter for Slack, since "public Slack" is not a by default supported operation. Let's leave it for now, I guess.

PR #3299, opened 2016-04-08
Test suite: create symlinks as needed.

### LINKED ISSUE #3190: Symlinks for Cabal package tests are not included in the sdist
These symlinks aren't listed in `Cabal.cabal`.  When I list them, `./Setup sdist` copies their targets, but we need the symlinks in this case.

```
tests/PackageTests/GhcPkgGuess/Symlink/ghc
tests/PackageTests/GhcPkgGuess/SymlinkGhcVersion/ghc
tests/PackageTests/GhcPkgGuess/SymlinkVersion/ghc
```

23Skidoo:
Fixes #3190.


Cabal/tests/PackageTests.hs
```diff
@@ -231,9 +235,26 @@ main = do
...
+        "tests/PackageTests/GhcPkgGuess/SymlinkGhcVersion/ghc"
+    maybeCreateSymlink "tests/PackageTests/GhcPkgGuess/SymlinkVersion/bin/ghc-7.10"
+        "tests/PackageTests/GhcPkgGuess/SymlinkVersion/ghc"
+
     defaultMainWithIngredients options $
         runTestTree "Package Tests" (tests suite)
 
+maybeCreateSymlink :: FilePath -> FilePath -> IO ()
+#ifdef mingw32_HOST_OS
+maybeCreateSymlink _oldpath _newpath = return ()
```
ezyang: Can we error here instead? Help us notice if we forgot to Windows gate a test.

ezyang:
Thanks LGTM.

Cabal/tests/PackageTests.hs
23Skidoo: Sure, but then I'll have to move symlink creation from `main` to `Tests.hs`.

23Skidoo:
Updated.

23Skidoo:
The Travis failure was due to the fact that when the symlink target is a relative path, it should be specified relative to the directory the symlink is in. That is,  

    ln - s bin/ghc-7.10 tests/PackageTests/GhcPkgGuess/SymlinkGhcVersion/ghc

instead of 

    ln - s tests/PackageTests/GhcPkgGuess/SymlinkGhcVersion/bin/ghc-7.10 tests/PackageTests/GhcPkgGuess/SymlinkGhcVersion/ghc

23Skidoo:
Looks like an existence check is needed after all.

Cabal/tests/PackageTests/PackageTester.hs
```diff
@@ -741,6 +746,18 @@ ghcFileModDelay = do
...
     liftIO $ threadDelay delay
 
+-- | Create a symlink if necessary. Does not work on Windows.
+createSymlink :: FilePath -> FilePath -> TestM ()
+#ifdef mingw32_HOST_OS
+createSymlink _oldpath _newpath =
+  error "PackageTests.PackageTester.createSymlink: does not work on Windows!"
+#else
+createSymlink oldpath newpath = do
+  symlinkExists <- liftIO $ doesFileExist newpath
```
ezyang: I'm pretty sure you want to delete the old symlink if it exists, otherwise a stale symlink will stick around and cause problems.

23Skidoo:
OK, changed it to create a symlink only for the duration of the test and to delete any existing ones.

Issue #1746, opened 2014-04-01
Better failure on missing module in `exposed-modules` or `other-modules`


ekmett:
If I forget a module in `exposed-modules` or `other-modules`, but a module I _do_ remember to add references I get a very hard to track down error.

`cabal` will let me build a package like that locally.

Hackage lets me upload the package, but then users get problems, because the resulting `.hs` files don't get included as they aren't referenced by the `cabal` file.

It would be really nice if there was a way that `cabal` could complain during `cabal build` if a module gets brought in transitively that isn't in either of those lists.

I'm sure I'm missing some corner cases it'll break, like `Paths_foo` or some such, but I felt the need to ask as this probably hits me about every 3 months.

hdgarrood:
Here's another potential approach to this, which I expect would be significantly simpler: before building, search through `hs-source-dirs` recursively for any valid Haskell modules which are not mentioned in any of `exposed-modules`, `other-modules`, `main-is`, etc., and emit warnings if any are found.

This approach assumes that there is no reason for someone to want valid Haskell modules in their Haskell source directories, but for those modules to never be referenced in the cabal file. It seems fairly probable that this is a safe assumption, but I'm not really sure.

MedeaMelana:
You would do this on cabal build? Or only on cabal dist?

hdgarrood:
Maybe both? It would be nice to be made aware of this earlier, and I can't imagine that it would be a noticeable performance hit.

hdgarrood:
Also it seems that if a module is referenced in other-modules and it doesn't exist in any of the source dirs, there's no warning either. See https://github.com/purescript/purescript/pull/980 for example; the actual module was called `Ctags` but the cabal file contained `CTags`.

MedeaMelana:
Could this be a Mac OS filesystem case sensitivity issue where it did work for the author?

hdgarrood:
Ohh, true. I'm on Linux, and I think he was on Mac. Perhaps cabal should check that the case of module names in the cabal file and files on the file system match up too.

hvr:
Related: commercialhaskell/stack#32

ttuegel:
@23Skidoo I just realized, I think we can do this without modifying GHC! All we must do is scan the build directory for interface files; that will indicate which modules `ghc --make` loaded. To avoid false positives, we should clean out the build directory when reconfiguring, but I think that is required for the new build infrastructure anyway.

23Skidoo:
[...]

PR #3302, opened 2016-04-08
Cabal.cabal: Add missing files

erikd:


23Skidoo:
```
tests/PackageTests/CabalMacros/my.cabal
tests/PackageTests/Regression/T3294/Main.hs
```

These files are not in the repo, they are probably auto-generated by the test suite.

erikd:
Oops, fixed.

23Skidoo:
Merged, thanks!

ezyang:
Thanks for the catch. I wonder if we can get Travis to check for these files too.

23Skidoo:
See #1746. This check is on the wishlist, but it shouldn't be specific to our Travis setup.

PR #3328, opened 2016-04-12
Only catch and unwrap user errors (simplified fix for #3310).

ezyang:


23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #3331, opened 2016-04-13
Solver quickcheck improvements

grayjay:
I added another test and applied the backjump limit.  The backjump limit prevents the test from running out of memory when the solver's log becomes too large.

I tried merging this PR,  #3221, #3234, #3237, @edsko's branch off of #3268, and #3327 into master.  I haven't seen any failures, and the memory usage has been constant.  Maybe we can start running the tests on Travis.


23Skidoo:
Merged, thanks!

grayjay:
Thanks!

Issue #3330, opened 2016-04-13
Cabal 1.24 breaks ghcjs


hvr:
There's currently a serious bug in Cabal/cabal-install 1.24 which makes it unusable with `ghcjs`, specifically `/usr/bin/strip` seems to be applied to .js files, which fails terribly...

@luite told me yesterday:

```
hvr> luite: did you take care of that Cabal 1.24 <-> ghcjs bug?
luite> hvr: not yet, i have a patch now that works here, but unfortunately i missed the rc3
```

So I'm creating this issue mostly so this issue is not forgotten, and possibly makes it into 8.0 final's Cabal (tbh, I don't know if the mentioned patch touches `Cabal` or only `cabal-install`)

/cc @23Skidoo

ezyang:
So this is something like, when using GHCJS, `stripExes` and `stripLibs` should always be false?

23Skidoo:
@luite, can we take a look at the patch?

Also, it would be nice if we had a Travis configuration for testing with GHCJS set up.

23Skidoo:
@luite 

This is basically the last bug blocking the 1.24 release. If I won't hear from you for the next few days, I'll postpone this until 1.24.0.1.

PR #3339, opened 2016-04-14
An attempt to unbreak GHCJS with Cabal 1.24.

23Skidoo:
See #3330.

Apparently we can't just disable stripping for GHCJS because it has a `--native-too` mode.

/cc @hvr


hvr:
@23Skidoo I'm testing this right now with ghcjs-7.10.3, and `cabal install --ghcjs ...` now, and your patch seems to work fine

23Skidoo:
@hvr Great, thanks.

PR #3341, opened 2016-04-15
Don't include requests for non-existent stanzas in pkgStanzasRequested, fixes #3340.

### LINKED ISSUE #3340: [nix-local-build] Packages without tests should not considered --enable-tests as part of configuration
Steps to reproduce:
1. Create a Cabal package with no test suite
2. `cabal new-build`
3. `cabal new-build --enable-tests`

Expected behavior: Cabal reports nothing to do

Actual behavior: Cabal decides to rebuild the package, because configuration has changed

CC @dcoutts

ezyang:
First three commits are just some extra commenting from when I was analyzing the code.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning.hs
@@ -177,15 +177,20 @@ import           System.FilePath
 -- data BuildStyle =
 
 
+-- | Check that an 'ElaboratedConfiguredPackage' actually makes
+-- sense under some 'ElaboratedSharedConfig'.
 sanityCheckElaboratedConfiguredPackage :: ElaboratedSharedConfig
                                        -> ElaboratedConfiguredPackage
                                        -> Bool
 sanityCheckElaboratedConfiguredPackage sharedConfig
                                        pkg@ElaboratedConfiguredPackage{..} =
 
+    -- we should only have enabled stanzas that actually can be built
+    -- (according to the solver)
     pkgStanzasEnabled `Set.isSubsetOf` pkgStanzasAvailable
 
-    -- the stanzas explicitly enabled should be available and enabled
+    -- the stanzas that the user explicitlyr requested should be
+    -- enabled (by the previous test, they are also available)
  && Map.keysSet (Map.filter id pkgStanzasRequested)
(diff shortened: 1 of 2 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -177,26 +177,38 @@ import           System.FilePath
...
                                        -> Bool
 sanityCheckElaboratedConfiguredPackage sharedConfig
                                        pkg@ElaboratedConfiguredPackage{..} =
 
+    -- we should only have enabled stanzas that actually can be built
+    -- (according to the solver)
     pkgStanzasEnabled `Set.isSubsetOf` pkgStanzasAvailable
 
-    -- the stanzas explicitly enabled should be available and enabled
+    -- the stanzas that the user explicitlyr requested should be
```
23Skidoo: Typo: explicitlyr.

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -181,37 +181,41 @@ import           System.FilePath
...
+                                       -> a
 sanityCheckElaboratedConfiguredPackage sharedConfig
-                                       pkg@ElaboratedConfiguredPackage{..} =
+                                       pkg@ElaboratedConfiguredPackage{..}
+                                       ret =
 
     -- we should only have enabled stanzas that actually can be built
     -- (according to the solver)
-    pkgStanzasEnabled `Set.isSubsetOf` pkgStanzasAvailable
+    assert (pkgStanzasEnabled `Set.isSubsetOf` pkgStanzasAvailable)
```
23Skidoo: Aren't asserts removed when optimisation is enabled?

23Skidoo:
LGTM.

cabal-install/Distribution/Client/ProjectPlanning.hs
ezyang: Yes. I just converted it into multiple asserts so that we know WHICH one failed when one fails. But arguably we should have a proper assert mechanism that is decoupled from optimization.

23Skidoo: Or just compile with `-fno-ignore-asserts`.

PR #3344, opened 2016-04-15
Fix boostrap.sh, to adapt to changes

tuncer:
Without this, bootstrap.sh fails to build.

### DIFF
--- a/cabal-install/bootstrap.sh
+++ b/cabal-install/bootstrap.sh
@@ -213,8 +213,16 @@ OLD_LOCALE_VER="1.0.0.7"; OLD_LOCALE_VER_REGEXP="1\.0\.?"
                        # >=1.0.0.0 && <1.1
 BYTEABLE_VER="0.1.1";  BYTEABLE_VER_REGEXP="0\.?"
                        # 0.1.1
-CRYPTOHASH_VER="0.11.7"; CRYPTOHASH_VER_REGEXP="0\.11\.?"
+MEMORY_VER="0.12";     MEMORY_VER_REGEXP="0\."
+                       # 0.*
+CRYPTONITE_VER="0.15"; CRYPTONITE_VER_REGEXP="0\."
+                       # 0.*
+CRYPTOHASH_VER="0.11.9"; CRYPTOHASH_VER_REGEXP="0\.11\.?"
                        # 0.11.*
+CRYPTOHASH_SHA256_VER="0.11.7.1"; CRYPTOHASH_SHA256_VER_REGEXP="0\.11\.?"
+                       # 0.11.*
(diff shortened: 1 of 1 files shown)


tuncer:
bootstrap.sh with a clean ~/.cabal and ~/.ghc should probably be part of the CI tests.
Maybe something like this:
```
rm -fr ~/.cabal ~/.ghc
cd cabal/Cabal
ghc --make Setup
./Setup configure --user
./Setup build
./Setup install
cd ../cabal-install
env EXTRA_CONFIGURE_OPTS="" ./bootstrap.sh --no-doc
```

ezyang:
LGTM, and I would love it if you could add that to our Travis!

tuncer:
I'll try to extend the travis-ci script later.

tuncer:
While this was needed to fix bootstrap.sh, I'm actually confused by the changes in https://github.com/haskell/cabal/commit/cf789ea314463c01f6fd56c10c1060a88b9ead1a#diff-16b83139b05c1138b7674bbd23a8a3a6. It appears that cryptohash was replaced with cryptohash-sha256, but hackage-security still needs cryptohash.

23Skidoo:
See https://github.com/well-typed/hackage-security/issues/152.
@dcoutts promised to make a new release of hackage-security soon that won't pull in the whole of `cryptonite`.

tuncer:
@23Skidoo, any idea when the new hackage-security release will land? As it is now, the build is broken. If it will take more than a day or two, I'd suggest to either temporarily revert the other change or merge this as an intermittent fix.

23Skidoo:
All right, let's merge this for now.

cabal-install/bootstrap.sh
```diff
@@ -213,8 +213,16 @@ OLD_LOCALE_VER="1.0.0.7"; OLD_LOCALE_VER_REGEXP="1\.0\.?"
                        # >=1.0.0.0 && <1.1
 BYTEABLE_VER="0.1.1";  BYTEABLE_VER_REGEXP="0\.?"
                        # 0.1.1
-CRYPTOHASH_VER="0.11.7"; CRYPTOHASH_VER_REGEXP="0\.11\.?"
+MEMORY_VER="0.12";     MEMORY_VER_REGEXP="0\."
+                       # 0.*
+CRYPTONITE_VER="0.15"; CRYPTONITE_VER_REGEXP="0\."
+                       # 0.*
+CRYPTOHASH_VER="0.11.9"; CRYPTOHASH_VER_REGEXP="0\.11\.?"
```
hvr: why was `cryptohash` updated? this way you've pulled two unneeded additional dependencies for no benefit

23Skidoo: We'll revert this once the new version of `hackage-security` is out anyway.

Issue #3345, opened 2016-04-15
cabal install assumes Custom packages will be installed with configured IPID, not true for old Cabal


tuncer:
With 1223a39, this is how far installing yi gets:

```
$ cabal -j4 install yi yi-fuzzy-open
...
Installed dynamic-state-0.2.2.0
Configuring executable-path-0.0.3...
Building executable-path-0.0.3...
Configuring ghc-paths-0.1.0.9...
Installed blaze-builder-0.4.0.1
Configuring oo-prototypes-0.1.0.0...
Installed executable-path-0.0.3
Building oo-prototypes-0.1.0.0...
Configuring io-storage-0.3...
Installed oo-prototypes-0.1.0.0
Building io-storage-0.3...
Configuring parallel-3.2.1.0...
Installed io-storage-0.3
Building ghc-paths-0.1.0.9...
Configuring parsec-3.1.9...
Installed charsetdetect-ae-1.1.0.1
Building parallel-3.2.1.0...
Configuring pointedlist-0.6.1...
Installed ghc-paths-0.1.0.9
InstallPlan: internal error: configured package depends on a non-library
package
```

ezyang:
Hmm, does this occur on master?

tuncer:
Yes, 1223a39 is master on github.

tuncer:
The only non-default setting I have in ~/.cabal/config is `split-objs: True`.

tuncer:
I haven't bisected when this started to happen, but I've been using cabal from master all the time.

tuncer:
The previous cabal executable was built on April 4th.

23Skidoo:
@tuncer Does this happen with 1.24? My guess is that this is related to internal libraries.

ezyang:
Well, I've seen this error before but it was in the case of `new-build` and fixed by https://github.com/haskell/cabal/commit/61fb6fadbefa2bd341fa8165bf77dea56624ff39 I feel that the install logic in this case should be correct. In any case I'll try to repro.

ezyang:
OK, repro'd.

ezyang:
OK, I've diagnosed the problem: `ghc-paths` is a Custom script built against an old version of Cabal. Consequently, we don't actually pass `--ipid` when configuring, and then the resulting IPID of the package is not what we expect. We can fix this by adopting some code from `new-build`. I have a small working test case too.

ezyang:
PR'd.

tuncer:
[...]

PR #3349, opened 2016-04-16
travis-script.sh: test bootstrapping

tuncer:


tuncer:
FYI, in case there are errors, I probably won't have time to fix them before Sunday.

tuncer:
Works as expected, meaning fails right now (see #3344).

23Skidoo:
I'm a bit worried about how this will affect compile times on Travis.

23Skidoo:
@tuncer Can you rebase this against master and do `push -f`? I want to evaluate the impact of this PR on compile times before merging.

ezyang:
@23Skidoo If we replace the code which uses cabal-install to build cabal-install, with code that uses the bootstrap script, I think that the time to build should be comparable.

tuncer:
Rebased, which also incorporates the fixed bootstrap.sh, so it should work now.

tuncer:
If we bootstrap, do we still need the initial `cabal install parsec`?

tuncer:
@ezyang, I've thought about replacing bits in there as well, but I wasn't sure which parts are okay to remove. Is it a better idea for you to make those changes in a follow-up commit?

tuncer:
Seems like I can remove `cabal install parsec`. Will update.

ezyang:
Yes, assuming @23Skidoo is OK with the Travis time we can defer it.

tuncer:
We might have to run `cabal update` after bootstrap, since ~/.ghc and ~/.cabal are nuked, but it's probably best to streamline it when trimming the redundant bits by, for example, running `cabal update` in travis-script.sh` instead (after bootstrap.sh).

tuncer:
After cabal-install has been bootstrapped, we probably don't have to install parsec anymore.

tuncer:
Extra commit that moves `cabal update` and optional `cabal install parsec` to a later stage, so that we don't run update twice and also use the freshly bootstrapped cabal-install.

23Skidoo:
This is now blocked by #3345.

tuncer:
I'm afraid I don't see what's causing this, given that bootstrap.sh is the one from #3345 since the rebase. Any idea?

23Skidoo:
It looks like #3356 needs to be merged to make the build bot error go away.

tuncer:
It's interesting that ghc-8.0.0.20160411 succeeds.

23Skidoo:
So the bootstrap script also needs to check that a recent version of `bytestring` is installed (and install it if it's not).

tuncer:
Added bytestring, and while at it, fixed a tiny whitespace issue separately.

23Skidoo:
Still fails, `bytestring` has to be installed before `binary`.

tuncer:
Fixed order.

tuncer:
Adjusted regex to require >=0.10.2.

tuncer:
Changed bytestring version that gets fetched to 0.10.6.0, in order to relax its base vsn bound. The previous error was that bytestring >=0.10.2.0 is needed, and 0.10.6.0 is the last bytestring release before the lower bound is bumped.

tuncer:
First, binary-0.8.3.0 complained that it needs bytestring >=0.10.2:

```
Configuring binary-0.8.3.0...
Setup: Encountered missing dependencies:
bytestring >=0.10.2
```

To fix that, I've tried giving it 0.10.8.0, but that failed because GHC 7.6.3's base is too old. Therefore, I've changed it to 0.10.6.0, since that's the last version before the base lower bound was bumped, and we cannot go older than 0.10.2.0 anyway.

Now we have a new error:

```
Configuring cabal-install-1.25.0.0...
Warning: This package indirectly depends on multiple versions of the same
package. This is highly likely to cause a compile failure.
package Cabal-1.25.0.0 requires binary-0.5.1.1
package text-1.2.2.1 requires binary-0.8.3.0
package cabal-install-1.25.0.0 requires binary-0.8.3.0
package unix-2.6.0.1 requires bytestring-0.10.0.2
package binary-0.5.1.1 requires bytestring-0.10.0.2
package Cabal-1.25.0.0 requires bytestring-0.10.0.2
package zlib-0.6.1.1 requires bytestring-0.10.6.0
package text-1.2.2.1 requires bytestring-0.10.6.0
package tar-0.5.0.3 requires bytestring-0.10.6.0
package parsec-3.1.9 requires bytestring-0.10.6.0
package network-2.6.2.1 requires bytestring-0.10.6.0
package hashable-1.2.4.0 requires bytestring-0.10.6.0
package hackage-security-0.5.1.0 requires bytestring-0.10.6.0
package ed25519-0.0.5.0 requires bytestring-0.10.6.0
package cryptohash-sha256-0.11.7.1 requires bytestring-0.10.6.0
package cabal-install-1.25.0.0 requires bytestring-0.10.6.0
package binary-0.8.3.0 requires bytestring-0.10.6.0
package base64-bytestring-1.0.0.1 requires bytestring-0.10.6.0
package base16-bytestring-0.1.1.6 requires bytestring-0.10.6.0
package HTTP-4000.3.3 requires bytestring-0.10.6.0
Building cabal-install-1.25.0.0...
[...]
Distribution/Client/World.hs:103:33:
    Couldn't match type `B.ByteString'
                  with `bytestring-0.10.0.2:Data.ByteString.Lazy.Internal.ByteString'
    Expected type: [Char]
                   -> bytestring-0.10.0.2:Data.ByteString.Lazy.Internal.ByteString
      Actual type: [Char] -> B.ByteString
    In the second argument of `(.)', namely `B.pack'
    In the expression: writeFileAtomic world . B.pack
    In a stmt of a 'do' block:
      writeFileAtomic world . B.pack
      $ unlines [(display pkg) | pkg <- pkgsNewWorld]
```

What should we do?

hvr:
@tuncer fwiw, `bytestring-0.10.8.1` will support GHC 7.6 again

tuncer:
@hvr, thanks.

@23Skidoo, what happened to the new hackage-security release? Is it ready, and would it help us avoid adding bytestring and also allow us to remove the other temporary packages?

23Skidoo:
I think what we can try doing to fix the error is check the version of bytestring installed and then install an older version of binary if bytestring is < 0.10.2 (binary-0.8.2.1 should work).

tuncer:
@hvr, now fetching bytestring 0.10.8.1.

tuncer:
@23Skidoo, can you file a ticket to remove the extra packages once the new hackage-security release can be integrated in bootstrap.sh?

23Skidoo:
I'm probably missing something. The failure on 7.6 and 7.4 is due to us having installed two versions of bytestring, so I'm not sure how using a newer `hackage-security` will help here. I believe that the way to fix this is to use the `bytestring` that comes with GHC instead (which means using `binary-0.8.2.1` on 7.4 and 7.6).

tuncer:
Some of the deps were meant to be intermittent until the new hackage-security package is out, so assuming we have a new hackage-security, I'd prefer to first trim deps and then fix any issues left.

23Skidoo:
Which ones? If you're referring to the dependency on `cryptohash`, it has already been removed in favour of `cryptohash-sha256`. That happened in `hackage-security` 0.5.1.0

tuncer:
That explains why I haven't been able to find it.

tuncer:
I'll modify bootstrap.sh accordingly.

tuncer:
Waiting for CI.

cabal-install/bootstrap.sh
```diff
@@ -181,8 +181,18 @@ PARSEC_VER="3.1.9";    PARSEC_VER_REGEXP="[3]\.[01]\."
                        # >= 3.0 && < 3.2
 DEEPSEQ_VER="1.4.2.0"; DEEPSEQ_VER_REGEXP="1\.[1-9]\."
                        # >= 1.1 && < 2
-BYTESTRING_VER="0.10.8.1"; BYTESTRING_VER_REGEXP="0\.10\.[2-9]"
-                       # >=0.10.2
+
+case "$GHC_VER" in
```
23Skidoo: I think we can remove all mentions of `bytestring` from the bootstrap script if we add a similar conditional for `binary` instead. So GHC <= 7.6 will use `binary-0.8.2.1` (which depends on `bytestring >= 0.9` and therefore should work with the [versions of `bytestring` that come with GHC 7.6 and 7.4](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory)) and GHC > 7.6 will use `binary-0.8.3.0`.

tuncer: Hmm, won't we have to repeat the conditional or at least conditionalise the do_pkg and info_pkg calls as well?

tuncer: Added a commit, please let me know if that's what you had in mind.

23Skidoo: If you remove this conditional and all mentions of `bytestring` from the script, I think that just conditionally defining `BINARY_VER` and `BINARY_VER_REGEXP` will work.

tuncer: Ah, right, now I got what you're saying.

cabal-install/bootstrap.sh
```diff
@@ -184,17 +184,22 @@ DEEPSEQ_VER="1.4.2.0"; DEEPSEQ_VER_REGEXP="1\.[1-9]\."
 
 case "$GHC_VER" in
     7.4*|7.6*)
+        # GHC 7.4 or 7.6
         BYTESTRING_VER="0.8.2.1"
```
23Skidoo: As I said, I think that we no longer need to install `bytestring` or check its version once we have a conditional for `binary`. Otherwise looks good.

tuncer: Yep, that's what the last commit does. Should I collapse the commits?

23Skidoo: Yep, looks good now. Let's wait for the CI tests to pass, then you can squash the commits.

23Skidoo:
And now it fails because `tar` requires `bytestring-builder` when compiled against `bytesting < 0.10`. So we'll also need to conditionally install `bytestring-builder` depending on the version `bytestring`.

tuncer:
Okay, added another commit.

23Skidoo:
Merged: 2645970ae942f13a942b369ad938db20ff641f75.

23Skidoo:
@tuncer Thanks for your patience!

23Skidoo:
One other thing we can do is to patch `binary` to support `bytestring-0.9` via `bytestring-builder` (like `tar` does, see https://github.com/haskell/tar/pull/14). That'd allow us to always install the latest version of `binary` on all supported versions of GHC.

PR #3357, opened 2016-04-17
Fix construction of conflict sets once and for all (WIP) 

edsko:
At @kosmikus ' request I have started a new `solver` branch in the main repository so that we can push solver fixes to this branch and then once we are happy with them, push them to master.

I realized that my understanding of conflict sets and goal reason chains was flawed. I think it's not just me though; they badly need explaining and documenting. To this end I started writing a blog post about the relation between conflict sets, pre assignments, constrained instances, and goal reason chains; a first draft can be found at http://www.well-typed.com/blog/preview/backjumping/ . 

On this branch so far we first have @grayjay 's test case for backjumping after the SIR check. We previously fixed this in #3221, but I now realize the fix there was incorrect. (In my defence, I did mention I was not very convinced.) I hope it is correct now (and if so, this PR will obsolete #3221). In addition we have a fix to the cycle check detection (#3170). That check too was wrong, and for similar reasons. The correct solution is also similar in both cases. Instead of collecting which _packages_ are involved in the conflict, we instead collect `Goals` (packages + goal reason chains). Then to construct the conflict set we construct conflict sets for each of these individual goals and then merge them. This is just a generalization of the process that happens in during validation (described in the aforelinked blog post). For the cycle check this is a tad awkward since we start with the final reverse dependency map which does not include goal reason chains, but we can just record these separately.

Sadly I am running out of time now. I was hoping to get more done today but I cannot. However, I think what needs to happen in the linker is similar (#3327) : instead of trying to construct a separate `lgBlame` set, we instead should collect `Goal`s as members of the link group, and use those to construct the conflict set.  separate `lgBlame` is not necessary (I hope); though perhaps @grayjay will correct me on that. Note that #3268 is (mostly) independent from this, as that one is less about constructing conflict set as it is about constructing the right link groups in the first place. That said, that PR should probably also be rebased to this branch (or at least, my version of it at https://github.com/edsko/cabal/tree/pr/FixLinkDeps ; still on my TODO list). 

It is somewhat frustrating that this is so difficult, but this is a subtle topic and I think fixing these issues and documenting and writing this up is rather important. 

@grayjay can I ask you for a favour? Can you rebase all your evil solver unit tests that are still missing from `master` and add them to this `solver` branch, even if they are currently failing? That would be very helpful.


grayjay:
Will this branch be merged into 1.24, if there's enough time?

@edsko I'll cherry-pick my tests and your comments on the tests.  I can also rebase `pr/FixLinkDeps`.

> Instead of collecting which packages are involved in the conflict, we instead collect Goals (packages + goal reason chains).

Does this change apply to all conflict sets?  Will the new `ConflictSet` type only allow insertion of `Goal`s?

> separate lgBlame is not necessary (I hope); though perhaps @grayjay will correct me on that.

Here's the reason that I thought we needed `lgBlame`.  I think it is possible for packages to become linked for reasons other than the actual linking choice or the introduction of the packages' goals.  `lgBlame` stores those other variables so that they can be added to the conflict set if the link group ever leads to a failure.

Say A depends on B and C, and introduces both goals. B also depends on C, but only when flagB is true.  For some dependency problem, the solver has to choose 0.B and 1.B.  If it links 0.B and 1.B and tries +flagB, it will have to link 0.C and 1.C.  `lgBlame` for 0.C and 1.C's link group will contain 0.flagB.  If linking 1.C leads to a failure, the solver should try -flagB.  Without `lgBlame`, the conflict set would only include 0.C, 1.C, 0.A, and 1.A (as well as any packages that introduced 0.A or 1.A).

edsko:
> Will this branch be merged into 1.24, if there's enough time?

I really hope so.

> @edsko I'll cherry-pick my tests and your comments on the tests.  I can also rebase `pr/FixLinkDeps`.

If you can rebase pr/FixLinkDeps and push it to this branch, that would be awesome. I am incredibly strapped for time at the moment. 

> > Instead of collecting which packages are involved in the conflict, we instead collect Goals (packages + goal reason chains).
> 
> Does this change apply to all conflict sets?  Will the new `ConflictSet` type only allow insertion of `Goal`s?

I _think_ it would make sense to say that `ConflictSet`s can only be constructed from two operations: from a `Goal`, or by merging other `ConflictSet`s. If that is not the case, then the exceptions would be quite illustrative. But perhaps @kosmikus can comment here as well.

> > separate lgBlame is not necessary (I hope); though perhaps @grayjay will correct me on that.
> 
> Here's the reason that I thought we needed `lgBlame`.  I think it is possible for packages to become linked for reasons other than the actual linking choice or the introduction of the packages' goals.  `lgBlame` stores those other variables so that they can be added to the conflict set if the link group ever leads to a failure.
> 
> Say A depends on B and C, and introduces both goals. B also depends on C, but only when flagB is true.  For some dependency problem, the solver has to choose 0.B and 1.B.  If it links 0.B and 1.B and tries +flagB, it will have to link 0.C and 1.C.  `lgBlame` for 0.C and 1.C's link group will contain 0.flagB.  If linking 1.C leads to a failure, the solver should try -flagB.  Without `lgBlame`, the conflict set would only include 0.C, 1.C, 0.A, and 1.A (as well as any packages that introduced 0.A or 1.A).

Yes, it is important that `flagB` will be included in the conflict set here. However, in your example, the dependency on `C` will have a goal reason chain that includes `flagB`; so, we if base this on `Goal`s (packages + goal reason chains), I think this will be okay? I could be wrong though.

kosmikus:
> > > Instead of collecting which packages are involved in the conflict, we instead collect Goals (packages + goal reason chains).
> > 
> > Does this change apply to all conflict sets?  Will the new `ConflictSet` type only allow insertion of `Goal`s?
>
> I _think_ it would make sense to say that `ConflictSet`s can only be constructed from two operations: > from a `Goal`, or by merging other `ConflictSet`s. If that is not the case, then the exceptions would be quite illustrative. But perhaps @kosmikus can comment here as well.

Yes, I think that makes sense. If we need other ways to create `ConflictSet`s for whatever reason, it still makes sense to explicitly add them to the `ConflictSet` module now that the type is abstract.

> > > separate lgBlame is not necessary (I hope); though perhaps @grayjay will correct me on that.
> > 
> > Here's the reason that I thought we needed `lgBlame`.  I think it is possible for packages to become linked for reasons other than the actual linking choice or the introduction of the packages' goals.  `lgBlame` stores those other variables so that they can be added to the conflict set if the link group ever leads to a failure.
> > 
> > Say A depends on B and C, and introduces both goals. B also depends on C, but only when flagB is true.  For some dependency problem, the solver has to choose 0.B and 1.B.  If it links 0.B and 1.B and tries +flagB, it will have to link 0.C and 1.C.  `lgBlame` for 0.C and 1.C's link group will contain 0.flagB.  If linking 1.C leads to a failure, the solver should try -flagB.  Without `lgBlame`, the conflict set would only include 0.C, 1.C, 0.A, and 1.A (as well as any packages that introduced 0.A or 1.A).
>
> Yes, it is important that `flagB` will be included in the conflict set here. However, in your example, the dependency on `C` will have a goal reason chain that includes `flagB`; so, we if base this on `Goal`s (packages + goal reason chains), I think this will be okay? I could be wrong though. 

Hmm, I'm not sure here, because I'm still not completely confident about some aspects of linking. Normally, if `A` already depends on `C`, then it's perfectly possible for `C` to get GRC `A` only. If it's later discovered that `B` also depends on `C` under certain circumstances, then that GRC would not be updated. In the unlinked scenario, this is ok. If we figure out that there's a conflict involving `C`, then we do not have a chance to avoid the `C` goal if we not _at least_ change the choice for `A`.

With linking, the situation is a bit more complicated. If we link `0.B` and `1.B` (but not `0.A` and `1.A` if they both exist), then as @grayjay says, it depends on the choice of `flagB` if we a forced to link `0.C` and `1.C` or not. Once again, without looking at the code, the intuition here is the constraint that `0.C` and `1.C` have to be linked is introduced by the choice of `flagB`, so written as a constraint, it'd be something like
```
1.B:flagB => 0.C ~ 1.C
```
So indeed, it's clear that if a conflict is arising in the context of (not) linking `1.C` to `0.C`, then `1.B:flagB` should definitely be a goal in the conflict set. The question is how we get there. For normal goals, we just have a set of active constraints and if we find a conflict between two constraints, we create a conflict set from the two goals that introduced these constraints. I'm not quite sure how this translates to link groups.

kosmikus:
@edsko Everything that is in this PR up until now looks ok to me.

23Skidoo:
@edsko 

 > > Will this branch be merged into 1.24, if there's enough time?

> I really hope so.

Since these patches only touch `cabal-install` code, they can always go into a patch-level release of `cabal-install`.

kosmikus:
@23Skidoo That's true and good to know, but we are in the somewhat uncomfortable situation that we have merged certain new features (such as in particular the setup/component dependencies) that are currently quite fragile due to subtle bugs. Conflict-set related bugs mean that the solver is incomplete, i.e., it does not find solutions where there are any, and can be very difficult to work around.

On the one hand, it's fortunate that this only affects new features that aren't in wide use yet. On the other hand, it'd be nice not to spoil the experience that users have with these new features by shipping them in a state that is known to be broken.

I'm sorry that I have started looking at these so late. But unfortunately, the issues involved are quite subtle, and solving them properly requires a lot of thinking time. I think that after last week's discussions, @edsko and I are on a good track now, and I'm optimistic that with the help of @grayjay, the three of us can come up with a robust solution soon.

23Skidoo:
@kosmikus @hvr told me that there will be a fourth RC before the GHC 8 release, so we probably can (yet again) postpone the 1.24 release for a short while.

kosmikus:
@23Skidoo :) Ok, I'll try to keep the momentum up this time.

hvr:
@23Skidoo Fwiw, RC4 is officially mentioned in the RC3 announcement: https://mail.haskell.org/pipermail/ghc-devs/2016-April/011862.html :-)

edsko:
@kosmikus , @grayjay : I've been thinking more about this problem this morning. In particular I've been thinking about Andres' remarks in the comment above (https://github.com/haskell/cabal/pull/3357#issuecomment-211518643) that the GCR might not include all the information we need if goals were already introduced previously, and his remark on my draft blog post (http://www.well-typed.com/blog/preview/backjumping/) that we examples with flags could be replaced by examples without flags. I think there is still a gap in my understanding of how this stuff works, exactly. 

Suppose we have the following database

``` haskell
dbGRC6a :: ExampleDb
dbGRC6a = [ Right $ exAv "A"  1 [ExAny "AF"]
          , Right $ exAv "AF" 1 [ExFix "B" 1]
          , Right $ exAv "AF" 2 [ExFix "B" 2]
          , Right $ exAv "B"  1 [ExFix "C" 1]
          , Right $ exAv "B"  2 [ExFix "C" 2]
          , Right $ exAv "C"  1 []
          ]
```

and we are solving for `A`. (Package `AF` here I think of as simulating `flagA`, though that doesn't necessarily matter.) The tree after validation looks like

simulateflags1

The node of interest is the right-most failure node: we pick version 2.0 for `AF`, then pick version 2.0 for `B`, and then get into trouble because we cannot pick a suitable version for `C`. In order to get out of trouble we need to change our mind about `AF`, and we can do that because thankfully the conflict set includes `AF`. It includes `AF` because although the conflict here is between `B` and `C`, the GRCs for `B` and `C` both include `AF` (it would have been sufficient of course if only one of them included it).

Ok, so far so good; this all makes sense. But then I wondered, ok, what if `A` has dependencies on both `B` and `C`:

``` haskell
dbGRC6b :: ExampleDb
dbGRC6b = [ Right $ exAv "A"  1 [ExAny "AF", ExAny "B", ExAny "C"]
          -- .. as before
          ]
```

In this case, the GRC might _not_ include `AF`. And indeed it does not:

simulateflags2

This tree is a lot larger because we have more choice in the order in which we solve the various constraints, but I am again interested in the right-most failure:

simulateflags2-path

On this path we solve the goals in precisely the same order as in the first example, **BUT** since both `B` and `C` are now also direct dependencies of `A`, their GRC does not include `AF`; and consequently, the conflict set recorded on the failure node does not in fact include `AF`! I'm sure this is a bug in my understanding of how this is supposed to work as opposed to a bug in the solver, but I cannot fix bugs in my head :sob: [mumbles something about Godel] 

So my question is, why is this tree correct? Until I understand that, I don't think I can finally suggest what we should do for the linking problem...

kosmikus:
I've discussed this with @edsko right now.

Some updates.

In the example posted above, the conflict sets are actually ok. In fact, even `{B,C}` would be sufficient as the conflict set of the conflict in question in this case. The conflict is between the constraints
```
C == 1.0.0
B => C == 2.0.0
```
so it is really between `B` and `C`. Every valid solution must make either a different choice for `B` or a different choice for `C`.

The connection to `AF` is going to be discovered once we backjump to `B` and try to make a different choice for that. We then discover (or have already discovered) that the other choice for `B` is impossible because of a conflict between
```
B == 1.0.0
AF => B == 2.0.0
```
so here we have `{AF, B}` ...

Why/when do we need goal reason chains at all? In fact, in the example @edsko gives above, we don't. The only reason GRCs are used by the solver currently is because the set of goals we have is in general not statically determined, but the presence or absence of goals depends on other goals.

Let's consider this example:

A-2 depends on C-any
A-1 depends on B-any
C-1 is unusable (e.g. due to an unsupported language extension)
B-1 is available

Now without GRCs, the solver would mark the choice C-1 as a failure, and since it is a failure that is not due to another constraint, the conflict set of this failure would be `{C}`.

Normally, the solver would start trying `A-2` and then go to `C` and arrive at the fail node. But this means that as there are no other choices for `C` available, we'd immediately backjump past `A` and declare the problem to be unsolvable, which is clearly wrong!

Storing that `C` is there because of `A` and using the conflict set `{A, C}` instead makes it possible for us to avoid `C` completely by making another choice for `A`.

In general, the only reason GRCs are included is because they may make it possible to avoid certain goals completely. If the goal set in a particular scenario is statically determined (as it is in @edsko's example above), then GRCs are superfluous.


While discussing this, @edsko and I discovered that there may be a much simpler way to look at all this that avoids the concept of GRCs. We can look at the goal set as being statically determined after all, by taking the union of all goals that occur anywhere in the search tree. The solver will, in general, only make explicit choices for a subset of these goals and leave others unassigned. We can make this explicit by saying that unassigned goals correspond to a choice of _|_ for that goal.

Now, every goal in the tree can get an additional choice that is called _|_ which corresponds to leaving that goal unassigned. By construction, this choice will *always* be a failure. The builder phase of the solver would not have created that choice node if a non-_|_ choice was not needed. But we can now store the justification of the goal's presence as the conflict set of the _|_ choice. In other words, the solver knows that it cannot just leave the goal unassigned because of the goal reason for that goal, which is stored in the _|_ option. Then we don't need the concept of GRCs explicitly anymore, and can construct conflict sets just by collecting the immediately conflicting goals and use the normal principles of backjumping and taking unions of conflict sets everywhere.

In the implementation, we may or may not make the _|_ options explicit. The GRCs are currently already stored in the tree for each node, so it would probably not be difficult to just change the backjumping code to include them at the right place.


The current plan is to fix the conflict set construction while linking without making this more fundamental change first and to get everything merged, and then try to do the general simplification after that and see if it works. For linking, we @edsko and I believe that `lgBlame` is needed, or perhaps to be replaced by just `lgConflictSet`, storing the conflict set associated with a link group directly. And for the time being, we'll need to collect goals (with GRCs) in there, rather than plain variables.

@grayjay, and feedback is of course appreciated.

kosmikus:
Actually, I think the diff for removing goal reason chains may be smaller than I thought. Here is an experimental ad-hoc implementation of the idea described above on top of current master (not including the changes currently in the solver branch): https://github.com/kosmikus/cabal/tree/no-grcs

This passes the test suite (on master) and seems to do ok on a few comparison scripts as well. As it seems to simple, I'm willing to consider just doing it unless we detect any problems that I overlooked. I'll try to merge the current solver branch into no-grcs tomorrow.

/cc @grayjay @edsko

grayjay:
I haven't looked at the code very thoroughly, but it looks good so far.  I think it's a very clean way to handle goal reasons.

I also diffed the logs from the unit tests to see the change in conflict sets.  I noticed one minor problem.  When the solver encounters a missing package, the package's variable seems to disappear from the conflict set higher in the tree.  I don't think it affects backjumping, only the summarized log.

If we merge `no-grcs` into `solver`, then #3237 and #3327 are the only issues remaining.  Do you think we can use the original commit from #3327, now that goal reasons are handled properly?

I can try to write a quickcheck test that compares the results of running the solver with and without backjumping, if I have time.

kosmikus:
@grayjay Thanks a lot for looking. Yes, missing packages were one of the things requiring an extra change, because in that scenario, there is no package choice with 0 options created, but rather a `Fail` node immediately. I then decided that it's rather useless to put the _missing_ package itself in the conflict set (it will be nowhere else), but only the immediate goal reason for introducing the missing package as a goal.

There will be lots of other differences in the logs, right? So unless we think there's an actual problem, I would not try to restore old behaviour here.

I still have to look at the other PRs ...

Having more quickcheck tests would be great, and indeed, the one you describe sounds like a good one.

Thanks again.

grayjay:
@kosmikus Glad I can help.

I thought the missing variable was a problem because the top-level conflict set is also used as a filter to summarize the log.  The message about the unknown package is not included in the summary, even though it is relevant.

I did notice other changes in the logs.  The conflict sets on failure nodes are smaller, as I expected, and there are more lines with "fail (backjumping, conflict set: ...".  I think the extra backjumping messages appear because messages with identical conflict sets are filtered out, and the solver now adds more variables as it backtracks, creating more distinct messages.  I assume these changes are fine.

kosmikus:
@grayjay Oh, the error message filter is a very good point. I'll add the unknown package back in, and will also add a comment that the error messages are the reason that we have it.

Right, the other changes are expected. I guess we could mostly reproduce the old output by adding more log-processing magic so that subsequent fail/backjump messages where one conflict set is a subset of the other are merged into just one, showing the largest. But I'm not sure if that's worth it. In the case that anything turns out to be wrong with conflict sets again for whatever reason, we probably appreciate the more detailed and more correct messages.

edsko:
@kosmikus @grayjay I've cherry-picked Andres's no-grc branch into the `solver` branch. There appear to be no regressions, but sadly the two expected failures did not magically turn into unexpected passes; that's what we did expect to happen, no? @grayjay am I wrong here? We though the `lgBlame` was correct but just missing GRCs (which are now no longer necessary)?

As for the unknown package issue discussed above (https://github.com/haskell/cabal/pull/3357#issuecomment-212776259, https://github.com/haskell/cabal/pull/3357#issuecomment-212780124) -- Andres and I just discussed this and we concluded that it would be better to fix this in the construction of the log messages (we can easily detect an empty `POption` node) rather than in the construction of the tree. If we try to include the name of the unknown package in the conflict set explicitly here it would make this the single exception of how we construct conflict sets, which is rather unfortunate. Moreover, since the only reason for including it in the first place is to have better error messages, it feels like the better place for this to live is in the generation of those error messages.

Some TODOs:

- [ ] Change the tree to have an empty `POption` for unknown packages (`PChoiceF qpn gr []` ... `BuildFailureNotInIndex pn`, as discussed just now)
- [ ] Add a `varToConflictSet` function alongside `toConflictSet` (which should probably be called `goalToConflictSet`). We should then audit where exactly we use `toConflictSet` and if those calls can be replaced by `varToConflictSet` instead. For example, in the cycle detection set we explicitly record some state mapping `QPN`s to `QGoalReason`s; that is no longer necessary. 
- [ ] There might also be some places where we can use `Var` instead of `Goal`; Anders suggests we should be able to replace `Goal` by `Var` in `VROrigin` for example. I think this would be a good simplification -- the construction of the conflict set is now becoming far more comprehensible: it just involves the packages that are in conflict with each other.
- [ ] It might be possible to remove `Unknown` from `GoalReason` (low priority)

edsko:
@kosmikus just reminded me that https://github.com/haskell/cabal/pull/3237 has not yet been merged into this branch. That might fix some problems.

kosmikus:
The `POption` fix @edsko describes above does not in itself fix the conflict set either. An empty package choice will right now only add its goal reason to the conflict set, not itself. But we can change it to add both itself and its reason.

edsko:
@kosmikus @grayjay Merged #3237; `indepGoals3` is now passing but `indepGoals4` is still failing. @grayjay were we expecting that to happen?

edsko:
Re the missing package: see detailed commit log in https://github.com/haskell/cabal/pull/3357/commits/5a4149f4ee2f0ce3aad8bcb1aeb8b34b21949954 .

edsko:
@grayjay ah, sorry, it's because your PR https://github.com/haskell/cabal/pull/3327 was still necessary. I've now brought that into this branch and now all tests are passing. @kosmikus / @grayjay after fixing the unknown package issue, we are now happy to recommend this to be merged with master?

cabal-install/Distribution/Client/Dependency/Modular/Cycles.hs
```diff
@@ -48,26 +47,16 @@ detectCyclesPhase = (`runReader` Set.empty) .  cata go
 -- as the full conflict set containing all decisions that led to that 'Done'
 -- node, check if the solution is cyclic. If it is, return the conflict set
 -- containing all decisions that could potentially break the cycle.
-findCycles :: ConflictSet QPN -> RevDepMap -> Maybe (ConflictSet QPN)
-findCycles fullSet revDeps = do
+findCycles :: Map QPN QGoalReason -> RevDepMap -> Maybe (ConflictSet QPN)
+findCycles grs revDeps = do
     guard $ not (null cycles)
```
kosmikus: I'd certainly prefer a simple
```
case cycles of
  [] -> Nothing
  (c : _) -> Just $ ...
```
over this combintion of `guard`, `null` and `head`.

cabal-install/Distribution/Client/Dependency/Modular/Cycles.hs
```diff
@@ -7,33 +7,32 @@ import Prelude hiding (cycle)
...
 #endif
 
 import Distribution.Client.Dependency.Modular.Dependency
-import Distribution.Client.Dependency.Modular.Flag
 import Distribution.Client.Dependency.Modular.Package
 import Distribution.Client.Dependency.Modular.Tree
+import qualified Distribution.Client.Dependency.Modular.ConflictSet as CS
 
-type DetectCycles = Reader (ConflictSet QPN)
+type DetectCycles = Reader (Map QPN QGoalReason)
```
kosmikus: Ok, I know we talked about this already and I said to do it later, but looking at it now, I agree with you. Collecting all these `QGoalReason`s now for no purpose is not worth it. We really don't need any `Reader` here now, right? Just the `RevDepMap` in the end?

cabal-install/Distribution/Client/Dependency/Modular/Dependency.hs
```diff
@@ -285,6 +242,26 @@ qualifyDeps QO{..} (Q pp@(PP ns q) pn) = go
...
+
+    goD :: Dep QPN -> Dep PN
+    goD (Dep qpn ci) = Dep (unq qpn) (fmap unq ci)
+    goD (Ext  ext)   = Ext ext
+    goD (Lang lang)  = Lang lang
+    goD (Pkg pn vr)  = Pkg pn vr
+
+    unq :: QPN -> PN
+    unq (Q _ pn) = pn
+
```
kosmikus: Could you please add a comment for `unqualifyDeps` stating what it does and (more importantly) where/why we need it? I'm sure I'll forget this again.

cabal-install/Distribution/Client/Dependency/Modular/Dependency.hs
```diff
@@ -361,31 +335,29 @@ instance ResetGoal Dep where
 instance ResetGoal Goal where
   resetGoal = const
 
--- | Compute a conflict set from a goal. The conflict set contains the closure
--- of goal reasons as well as the variable of the goal itself.
+-- | Compute a singleton conflict set from a goal, containing just
+-- the goal variable.
 toConflictSet :: Ord qpn => Goal qpn -> ConflictSet qpn
-toConflictSet (Goal g grs) = S.insert (simplifyVar g) (goalReasonChainToVars grs)
+toConflictSet (Goal g _gr) = CS.singleton (simplifyVar g)
```
kosmikus: The call to `simplifyVar` is redundant.

Also, I propose to rename this to `goalVarToConflictSet`, hopefully making it more obvious from the name that it only makes use of the variable.

cabal-install/Distribution/Client/Dependency/Modular/Dependency.hs
```diff
@@ -361,31 +335,29 @@ instance ResetGoal Dep where
...
+extendConflictSet = CS.insert . simplifyVar
 
-goalReasonChainToVars :: Ord qpn => GoalReasonChain qpn -> ConflictSet qpn
-goalReasonChainToVars = S.unions . L.map goalReasonToVars
+goalReasonToVars :: Ord qpn => GoalReason qpn -> ConflictSet qpn
+goalReasonToVars UserGoal                 = CS.empty
+goalReasonToVars (PDependency (PI qpn _)) = CS.singleton (P qpn)
+goalReasonToVars (FDependency qfn _)      = CS.singleton (simplifyVar (F qfn))
+goalReasonToVars (SDependency qsn)        = CS.singleton (S qsn)
+goalReasonToVars Unknown                  = CS.empty
```
kosmikus: The call to `simplifyVar` in L350 is redundant.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
```diff
@@ -216,37 +224,46 @@ pickConcrete qpn@(Q pp _) i = do
...
+
+    -- Find the link group for the package we are linking to
     --
     -- Since the builder never links to a package without having first picked a
     -- concrete instance for that package, and since we create singleton link
     -- groups for concrete instances, this link group must exist (and must
     -- in fact already have a canonical member).
-    let lg = vsLinks vs ! Q pp' pn
+    let target   = Q pp' pn
+        lgTarget = vsLinks vs ! Q pp' pn
```
kosmikus: Looks like the intent was to have L244 read
```
lgTarget = vsLinks vs ! target
```

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
```diff
@@ -216,37 +224,46 @@ pickConcrete qpn@(Q pp _) i = do
...
+    lgTarget' <- lift' $ lgMerge [] lgSource lgTarget
+    updateLinkGroup lgTarget'
 
-    -- Since we already have a canonical member, we just need to add the new
-    -- member into the group
-    let lg' = lg { lgMembers = S.insert pp (lgMembers lg) }
-    updateLinkGroup lg'
-    linkDeps [P qpn] pp' deps
+    -- Make sure all dependencies are linked as well
+    linkDeps target [P qpn] deps
```
kosmikus: I have to think about this one a bit more to make sure that I actually believe it :)

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
```diff
@@ -263,35 +280,45 @@ makeCanonical lg qpn@(Q pp _) i =
...
-    go :: FlaggedDep comp QPN -> UpdateState ()
-    go (Simple (Dep qpn@(Q _ pn) _) _) = do
-      vs <- get
-      let qpn' = Q pp' pn
-          lg   = M.findWithDefault (lgSingleton qpn  Nothing) qpn  $ vsLinks vs
-          lg'  = M.findWithDefault (lgSingleton qpn' Nothing) qpn' $ vsLinks vs
-      lg'' <- lift' $ lgMerge parents lg lg'
-      updateLinkGroup lg''
+    go :: [Var QPN] -> FlaggedDeps Component QPN -> FlaggedDeps Component QPN -> UpdateState ()
+    go parents deps rdeps = mapM_ (uncurry (go1 parents)) $ zip deps rdeps
```
kosmikus: Why not `zipWithM_`?

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
```diff
@@ -263,35 +280,45 @@ makeCanonical lg qpn@(Q pp _) i =
 -- because having the direct dependencies in a link group means that we must
 -- have already made or will make sooner or later a link choice for one of these
 -- as well, and cover their dependencies at that point.
-linkDeps :: [Var QPN] -> PP -> FlaggedDeps comp QPN -> UpdateState ()
-linkDeps parents pp' = mapM_ go
+linkDeps :: QPN -> [Var QPN] -> FlaggedDeps Component QPN -> UpdateState ()
+linkDeps parent = \parents deps -> do
```
kosmikus: Why the RHS lambda?

Do we really need to call the two args `parent` and `parents`? Seems very easy to confuse.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
```diff
@@ -401,7 +428,7 @@ verifyFlag' (FN (PI pn i) fn) lg = do
         vals  = map (`M.lookup` vsFlags vs) flags
     if allEqual (catMaybes vals) -- We ignore not-yet assigned flags
       then return ()
-      else conflict ( S.fromList (map F flags) `S.union` lgConflictSet lg
+      else conflict ( CS.fromList (map F flags) `CS.union` lgConflictSet lg
                     , "flag " ++ show fn ++ " incompatible"
                     )
```
kosmikus: I don't think I fully understand why this whole flag / stanza verification business is needed.

cabal-install/Distribution/Client/Dependency/Modular/Preference.hs
```diff
@@ -358,37 +358,41 @@ preferReallyEasyGoalChoices = trav go
...
+enforceSingleInstanceRestriction :: Tree QGoalReason -> Tree QGoalReason
 enforceSingleInstanceRestriction = (`runReader` M.empty) . cata go
   where
-    go :: TreeF QGoalReasonChain (EnforceSIR (Tree QGoalReasonChain)) -> EnforceSIR (Tree QGoalReasonChain)
+    go :: TreeF QGoalReason (EnforceSIR (Tree QGoalReason)) -> EnforceSIR (Tree QGoalReason)
 
     -- We just verify package choices.
     go (PChoiceF qpn gr cs) =
-      PChoice qpn gr <$> sequence (P.mapWithKey (goP qpn) cs)
+      PChoice qpn gr <$> sequence (P.mapWithKey (goP qpn gr) cs)
```
kosmikus: This is another point where we no longer need the goal reason now, right?

cabal-install/Distribution/Client/Dependency/Modular/Preference.hs
```diff
@@ -358,37 +358,41 @@ preferReallyEasyGoalChoices = trav go
     go x                = x
 
 -- | Monad used internally in enforceSingleInstanceRestriction
-type EnforceSIR = Reader (Map (PI PN) QPN)
+--
+-- For each package instance we record the goal for which we picked a concrete
+-- instance. The SIR means that for any package instance there can only be one.
+type EnforceSIR = Reader (Map (PI PN) (Goal QPN))
```
kosmikus: This can be changed back to `QPN`, right?

travis-script.sh
```diff
@@ -75,7 +75,7 @@ install_from_tarball
 cd ../cabal-install
 
 cabal install happy
-cabal install --only-dependencies --enable-tests --enable-benchmarks
+cabal install --only-dependencies --enable-tests --enable-benchmarks --allow-newer=tasty-expected-failure:base
```
kosmikus: Do we still need `tasty-expected-failures`?

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
```diff
@@ -125,8 +125,13 @@ build = ana go
...
       error "Distribution.Client.Dependency.Modular.Builder: build.go called with Pkg goal"
     go bs@(BS { index = idx, next = OneGoal (OpenGoal (Simple (Dep qpn@(Q _ pn) _) _) gr) }) =
+      -- If the package does not exist in the index, we construct an emty PChoiceF node for it
+      -- After all, we have no choices here. Alternatively, we could immediately construct
+      -- a Fail node here, but that would complicate the construction of conflict sets.
+      -- We will probably want to give this case special treatment when generating error
+      -- messages though.
       case M.lookup pn idx of
-        Nothing  -> FailF (toConflictSet (Goal (P qpn) gr)) (BuildFailureNotInIndex pn)
+        Nothing  -> PChoiceF qpn gr (P.fromList [])
```
kosmikus: I'm leaning towards reverting this, and simply adapting the conflict set for the `FailF`, which is easy enough. The log manipulation seems fragile to me, and having to adapt the initial conflict set in backjumping instead is also not really nicer.

kosmikus:
@edsko Ok, looked through everything once:

- There are two points I still want to think about a bit more. (I made comments in the code involving "I do not understand" or "I want to think about this again/more".)
- The unknown package issue I'd prefer to handle like I originally did, just with the modified package set. I think the changes are not really an improvement after looking at them now, and considering they're incomplete.
- For everything else, I'm confident it's correct, but I've made a few minor comments here and there ( @grayjay, some are probably about changes introduced by you, so you might want to have a look too).

But it looks like we're close to be able to merge this. Thanks a lot so far.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
grayjay: FWIW, I thought that including the current variable in the initial conflict set was more correct.  If we were using an explicit `_|_` node, and it wasn't last in the list of children, its conflict set would need to contain the current variable in order for the solver to try the other siblings.

cabal-install/Distribution/Client/Dependency/Modular/Explore.hs
```diff
@@ -5,18 +5,19 @@ module Distribution.Client.Dependency.Modular.Explore
...
 import Distribution.Client.Dependency.Modular.Message
 import Distribution.Client.Dependency.Modular.Package
 import qualified Distribution.Client.Dependency.Modular.PSQ as P
+import qualified Distribution.Client.Dependency.Modular.ConflictSet as CS
 import Distribution.Client.Dependency.Modular.Tree
 import qualified Distribution.Client.Dependency.Types as T
 
--- | This function takes the variable we're currently considering and a
+-- | This function takes the variable we're currently considering, an
+-- initial conflict set (the goal reason for the current goal) and a
```
grayjay: I think it would be nice to include an explanation of how this corresponds to a `_|_` choice, like in https://github.com/haskell/cabal/pull/3357#issuecomment-212298258.  That definitely made it clearer for me.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
grayjay: How about `target` instead of `parent`? It matches the variable in `pickLink`.

grayjay:
@edsko The two failures were expected, since they weren't related to `GoalReasonChain`s.

> I guess we could mostly reproduce the old output by adding more log-processing magic so that subsequent fail/backjump messages where one conflict set is a subset of the other are merged into just one, showing the largest. But I'm not sure if that's worth it. In the case that anything turns out to be wrong with conflict sets again for whatever reason, we probably appreciate the more detailed and more correct messages.

I tried solving for a few real packages, and I didn't notice many backjumping messages in the logs.  I don't think the extra messages are a problem, especially since the messages are still filtered out when the conflict set doesn't contain the current level.

> @kosmikus / @grayjay after fixing the unknown package issue, we are now happy to recommend this to be merged with master?

Another thing that isn't finished is documentation of the invariant from https://github.com/haskell/cabal/pull/3268#issuecomment-209217491.

travis-script.sh
grayjay: I removed it.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
kosmikus: @grayjay I'm certainly still undecided on this, but I don't think I understand your argument: the original code creates an explicit Failure node here in place of a package choice node, so it'll be a child of a goal choice node. I don't understand what you mean about it being last in a list of children in this context ...

Or are you talking about something different?

cabal-install/Distribution/Client/Dependency/Modular/Explore.hs
kosmikus: @grayjay You are right. Will do.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
grayjay: @kosmikus Maybe I misunderstood, but I thought part of the change to use an empty `PChoice` involved adding the current variable to the initial conflict set that is given to the backjump function.  Then `initial` would contain the current variable and the goal reason variable.  I meant to say that I thought using both variables was more correct than using just the goal reason, regardless of how we deal with unknown packages.

I think it makes sense to revert the rest of the empty `POption` change, because it complicates the log processing.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
edsko: Because if I link Q.A to Q'.A, then it doesn't make sense to allow Q.A-flag to be true but Q'.A-flag to be false.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
kosmikus: @grayjay But why do you think it's "more correct"? Except in the case where we have no available choices at all (the one we're talking about here), the only way we could be asked to `Fail` here is because we discovered conflicts lower in the tree. But then we're backjumping already, and we wouldn't even consider the current node if their conflict didn't include our package name.

edsko:
@grayjay @kosmikus implemented almost all the requests above:

- [x] Simplify coding style in `findCycles`
- [x] Remove the Reader from `findCycles`
- [x] Replace `toConflictSet` with `goalVarToConflictSet` and `varToConflictSet`
- [x] Add comment to `unqualifyDeps`
- [x] Remove redundant calls to `simplifyVar`
- [x] Use `target` in definition of `lgTarget`
- [x] Use `zipWithM` in `linkDeps`
- [x] Rename arguments of `linkDeps` to `target` and `blame` (left the RHS lambda; I do this on purpose to avoid variables being in scope unecesssarily and avoid referring to the wrong variable in the `where`-bound definitions)
- [x] Remove use of the GRC in `enforceSingleInstanceRestriction` (collecting `QPN`s rather than `Goal`s)
- [x] Document the package qualification invariant proposed in https://github.com/haskell/cabal/pull/3268#issuecomment-209217491
- [ ] Change the unknown package treatment. I do not fully understand the issue, and also I'm not sure what @grayjay means in https://github.com/haskell/cabal/pull/3357#discussion_r60681294 -- which solution she prefers
- [ ] Related: I have not implemented @grayjay's request at https://github.com/haskell/cabal/pull/3357#discussion_r60681435 ; I think that's something @kosmikus should do, perhaps.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
kosmikus: Ok, this still makes me feel uneasy. @grayjay, @edsko, am I right that the correctness of this relies on the SIR being enforced? I would normally expect that we have to include both `P qpn` and `P target` in the blame set.

Consider the following database:
```
db22 :: ExampleDb
db22 = [
    Right $ exAv "X" 1 [ExFix "C" 2, ExAny "A"]
  , Right $ exAv "Y" 1 [ExFix "C" 1, ExFix "A" 2]
  , Right $ exAv "A" 1 []
  , Right $ exAv "A" 2 [ExAny "B"]
  , Right $ exAv "B" 1 [ExAny "C"]
  , Right $ exAv "C" 1 []
  , Right $ exAv "C" 2 []
  ]
```
and let's assume we solve for `0.X` and `1.Y`. We get the following trace:
```
[__0] trying: 0.X-1.0.0 (user goal)
[__1] trying: 0.A-2.0.0 (dependency of 0.X-1.0.0)
[__2] trying: 0.B-1.0.0 (dependency of 0.A-2.0.0)
[__3] trying: 0.C-2.0.0 (dependency of 0.X-1.0.0)
[__4] trying: 1.Y-1.0.0 (user goal)
[__5] trying: 1.A~>0.A-2.0.0 (dependency of 1.Y-1.0.0)
[__6] trying: 1.B~>0.B-1.0.0 (dependency of 1.A-2.0.0)
[__7] next goal: 1.C (dependency of 1.Y-1.0.0)
[__7] rejecting: 1.C~>0.C-2.0.0 (conflict: 1.Y => 1.C==1.0.0)
[__7] rejecting: 1.C-2.0.0 (multiple instances)
[__7] rejecting: 1.C-1.0.0 (dependencies not linked: cannot make 1.C canonical member of {*0.C-2.0.0,1.C-2.0.0}; conflict set: 0.C, 1.B, 1.C)
[__7] fail (backjumping, conflict set: 0.C, 1.B, 1.C, 1.Y)
[__6] rejecting: 1.B-1.0.0 (multiple instances)
[__6] fail (backjumping, conflict set: 0.B, 0.C, 1.A, 1.B, 1.C, 1.Y)
[__5] rejecting: 1.A-2.0.0 (multiple instances)
[__5] rejecting: 1.A-1.0.0 (conflict: 1.Y => 1.A==2.0.0)
[__4] fail (backjumping, conflict set: 0.A, 0.B, 0.C, 1.A, 1.B, 1.C, 1.Y)
[__3] rejecting: 0.C-1.0.0 (conflict: 0.X => 0.C==2.0.0)
[__2] fail (backjumping, conflict set: 0.A, 0.B, 0.C, 0.X, 1.A, 1.B, 1.C, 1.Y)
[__1] trying: 0.A-1.0.0
[__2] trying: 0.C-2.0.0 (dependency of 0.X-1.0.0)
[__3] trying: 1.Y-1.0.0 (user goal)
[__4] next goal: 1.A (dependency of 1.Y-1.0.0)
[__4] rejecting: 1.A~>0.A-1.0.0 (conflict: 1.Y => 1.A==2.0.0)
[__4] trying: 1.A-2.0.0
[__5] trying: 1.B-1.0.0 (dependency of 1.A-2.0.0)
[__6] next goal: 1.C (dependency of 1.Y-1.0.0)
[__6] rejecting: 1.C~>0.C-2.0.0 (conflict: 1.Y => 1.C==1.0.0)
[__6] rejecting: 1.C-2.0.0 (multiple instances)
[__6] trying: 1.C-1.0.0
[__7] done
```

The interesting place is on level 7, where we reject all options for `C` and then backjump with the conflict set `{0.C, 1.B, 1.C, 1.Y}` which does *not* contain `0.B`.

But `0.B` in the conflict set is ultimately necessary in order to make us reconsider `0.A`, which was the choice that really needs to be fixed.

We do pick up `0.B` here, but only on level 6, because of
```
[__6] rejecting: 1.B-1.0.0 (multiple instances)
[__6] fail (backjumping, conflict set: 0.B, 0.C, 1.A, 1.B, 1.C, 1.Y)
```
This is because for the computation of the SIR conflict set, we indeed include both the link source and the link target.

I am, however, somewhat reluctant of relying on the presence of the SIR for the correctness of the solver. After all, it's possible we will remove or relax it at some point in the future.


The reason is that it's quite possible that we discover a failure underneath the current node that cannot be fixed by changing

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
edsko: > I think it makes sense to revert the rest of the empty POption change, because it complicates the log processing.

This I disagree with. Log processing is just that -- log processing. Having the conflict sets created in the same way everywhere seems a worthwhile goal to have to me. Having _one_ place in the tree construction where the conflict set is created in a different way to everywhere else seems very unfortunate.

grayjay: I thought that `initial` represents the `_|_` node, and `_|_` should have a conflict set containing the current variable.  Say we used an explicit `_|_` node, and it happened to be first in the list of children.  If `_|_` didn't contain the current variable, the solver would backjump after reaching the failure and would skip the siblings, even though those siblings might lead to solutions.

Since we're not using an explicit failure node, adding the current variable to `initial` wouldn't actually affect the behavior.  I just thought it might be more consistent with representing the `_|_` node.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
edsko: Certainly any reliance on the SIR is purely accidental. Having both `qpn` (which would probably be renamed to `source`) and `target` in the blame set makes sense to me.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
kosmikus: @grayjay Oh ok, I understand your argument now, and I agree.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
kosmikus: This being said, I've disabled the SIR and run a few more tests, and I think we're fine, but nevertheless for relatively subtle reasons. I'll see whether I can come up with an explanation that convinces me. I still think it's harmless to add the target at this point, and it would generally make me feel better :)

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
kosmikus: Yes, because `Q.A-flag` and `Q'.A-flag` are "direct dependencies" of `A` and therefore have to be linked. But you seem to be doing something different and assert that it's equivalent?

edsko: We don't have a direct representation of linking flags. We just check if their values are the same.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
kosmikus: One more note here, mainly for myself: without the SIR, linking is never necessary, just perhaps desirable. This changes the kind of counterexamples we could possibly hope to produce.

kosmikus:
@edsko, @grayjay: I think I'm happy to propose the current state of this branch for merging into master.

Done:

> - [x] Change the unknown package treatment. I do not fully understand the issue, and also I'm not sure what @grayjay means in #3357 (comment) -- which solution she prefers
> - [x] Related: I have not implemented @grayjay's request at #3357 (comment) ; I think that's something @kosmikus should do, perhaps.

Regarding the "unknown package" treatment: added current package to the initial conflict set, as @grayjay suggested. The log formatting seems to be ok now, even though @edsko apparently had observed in the past that it doesn't always work?

Also done:

- [x] Further simplifications enabled by removing goal reasons, among other things, avoiding the need for `Unknown` as a goal reason.
- [x] Added ability to the solver tests to check error messages. Added a test case to verify that "unknown package" occurs in the errors.
- [x] Fix a few warnings for various GHC versions.

Regarding the one issue that I had marked, whether to add "target" to the blame set in implied link constraints: Let's keep things as they are fow now. I'm confident this is a non-issue as long as the SIR is actively enforced. Even without the SIR, it's never going to turn a problem where solutions exist into one where no solutions can found. However, I think that without the SIR, the current solver will not have the potential to find *all* possible solutions, so it's not truly complete. I also think, though, that the correct fix for this would not be to add the link target to the blame set for the implied link constraints, but to add the link target to the goal reason for the goal being linked itself (and thereby to the "initial" conflict set for that goal). I think we should discuss this again, but as I'm confident the current code is ok because of the SIR, I'm eager to finally get this merged and not hold this up any longer.

What do you think?

@23Skidoo Is there anything we can do about the ghc-8 CI test failing due to exceeding the time limit?

23Skidoo:
@kosmikus Is the test expected to take such a long time? I'd recommend making it optional then.

kosmikus:
@23Skidoo I really have no idea. I haven't invented the CI tests for cabal. But if you look at https://travis-ci.org/haskell/cabal/builds/125010104 and https://travis-ci.org/haskell/cabal/builds/125010110, you can see that the GHC-8 builds seem to take far longer than any of the others ...

23Skidoo:
Sorry, I thought that this was something specific to the `solver` branch. Try changing `TEST_OLDER=YES` to `NO` on line 11 in `.travis.yml`.

/cc @ezyang

ezyang:
I guess we'll have to move `TEST_OLDER` to its own matrix entry, though I surprised the extra tests are doubling the size. Here is a comparison: https://travis-ci.org/haskell/cabal/builds/124624027
Whatever the reason, Travis run time has gotten substantially longer.

kosmikus:
@ezyang OOC, what does `TEST_OLDER` do?

ezyang:
Runs the integration tests (Cabal and cabal-install) but passing `-w older-ghc`.

kosmikus:
@ezyabg Hmm, it is a bit strange that the times seem to be significantly higher than for recent PRs to master. I cannot imagine anything in this PR making things so much worse (and then only for GHC-8 ...).

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
grayjay: I think `[P qpn]` is correct.  `target` (0.B) did not cause any packages to become linked, because it was not linked when its version was picked.  I tried running the test without enforcing the SIR, and

    [__6] rejecting: 1.B-1.0.0 (multiple instances)

became

    [__6] rejecting: 1.B-1.0.0 (dependencies not linked: cannot make 1.B canonical member of {*0.B-1.0.0,1.B-1.0.0}; conflict set: 0.B, 1.A, 1.B)

The SIR adds 0.B to the conflict set only because the solver happens to enforce the SIR before enforcing linked dependencies.  Only one broken constraint is needed to make a valid conflict set.

cabal-install/Distribution/Client/Dependency/Modular/Builder.hs
grayjay: > This I disagree with. Log processing is just that -- log processing. Having the conflict sets created in the same way everywhere seems a worthwhile goal to have to me. Having one place in the tree construction where the conflict set is created in a different way to everywhere else seems very unfortunate.

@edsko I just meant that I was okay with reverting the change; I don't have a strong opinion.  Now that we show the unknown package in the summarized log, I'm happy with the empty `PChoice` approach.  I like how it simplifies conflict sets.

cabal-install/Distribution/Client/Dependency/Modular/Linking.hs
kosmikus: @grayjay I actually believe that too, after doing similar tests. But I think that to be completely correct, `0.B` would have to be added to the goal reasons / initial conflict set of `1.B`, simply for the fact that it makes the "link to whatever version we have chosen for `0.B`" choice available, but no other linking option. I think that if we want that even without the SIR, the solver is able to find *all* solutions (including all solutions based on linking), this extra entry is necessary. With the SIR, it clearly makes no difference. So as I suggest in the general comment I've made, I suggest to merge this now without change and have this discussion in more detail independently.

grayjay: Okay.

grayjay:
> What do you think?

As far as I can tell, this is ready to be merged.  All of the issues have been addressed, and I think that the new documentation is clear.  I'm not familiar with all of the code, though.

kosmikus:
@grayjay Ok, thanks. @edsko, in case you agree, just hit the button :)

23Skidoo:
@kosmikus Will there be any other solver fixes that to merge into 1.24?

kosmikus:
@23Skidoo If I'm keeping track correctly, then once this PR is merged, the following PRs can be closed because they have been integrated into / made obsolete by this one:

- #3221 
- #3234 
- #3237 
- #3268 
- #3327 

@edsko, @grayjay: Can you verify this list? Am I missing any PRs? Am I listing PRs that you think aren't fixed by this?

The only other possible candidate for 1.24 that I'm currently aware of and that isn't covered by this PR is #3290. If @edsko agrees that this one (#3357) is ready to merge, I'm planning to talk to him about what he thinks about #3290. I don't fully understand it yet (there are a number of changes outside of the solver, in code I'm less familiar with), and @ezyang seems to claim that it's still buggy ...

ezyang:
I restarted the build to see if it's a transient failure. But looking at the logs, it just seems like Travis got slower recently.

grayjay:
@kosmikus The list of PRs looks correct.  I can close mine.

I looked at http://scribu.github.io/travis-stats/#haskell/cabal, and one thing that has increased build times on master is the solver quickcheck tests, since the solver needs to be compiled again.  I guess that's another reason to split the solver into a library.  Another option is to add the tests to `unit-tests`.  The only problem is that the solver tests can take minutes to find bugs.  The default number of tests doesn't take long to run, it just isn't very thorough.

jennycraig:
ehh ufa

edsko:
Merged with master :grin: Let me just for the record extend a huge kudos once more to @grayjay -- you're a star!

@23Skidoo I have closed all the outstanding PRs that were subsumed by this branch. As @kosmikus says, the only one remaining is #3232/#3290, and I'm inclined to propose that we do not try to merge that before release. However, Andres mentioned that @hvr or @dcoutts said that that PR is actually needed for the `nix-local-build` stuff to be useable. I don't fully understand why that would the case, but if true, then it's a higher priority than I thought.

grayjay:
@edsko Thanks!

kosmikus:
Thanks to both @grayjay and @edsko for making this happen. Sorry again for letting the original PRs linger for so long, and thanks for nevertheless investing so much time now to make this happen pre-1.24 after all.

Regarding #3290: I think I had misunderstood which bug was responsible. Had a discussion with @dcoutts yesterday, and we verified that the main problem he was thinking about was a conflict set problem that is fixed by the solver branch. I would nevertheless still at least try to understand the current status of #3290 and the trade-offs involved before I make a final decision on whether to try to merge it pre-1.24 or not.

23Skidoo:
@kosmikus @edsko @grayjay Thanks, everyone!

I'll wait for a verdict on #3290 from @kosmikus before releasing 1.24 final.

23Skidoo:
@kosmikus 

I've now backported the solver fixes from `master` to 1.24 (see #3373).

grayjay:
@kosmikus @23Skidoo Thanks!

kosmikus:
@23Skidoo I unfortunately haven't managed to talk to @edsko over the weekend or today. But having looked a bit at #3290 myself, I think I agree with @edsko that we should do that post-1.24. It all still looks somewhat confusing to me, and while it's without doubt nice to have, it doesn't seem to be a big blocker either.

23Skidoo:
@kosmikus OK. Thanks for looking into this.

edsko:
@grayjay I've emailed you with a link to a draft blog post about backjumping and conflict sets; I used the email address listed on your github profile. [Let me know](http://www.well-typed.com/who_we_are/) if this isn't the right address.

PR #3359, opened 2016-04-18
Solver test cases

grayjay:
I cherry-picked the solver test cases from open PRs onto the solver branch (See #3357) and added issue numbers. I used `tasty-expected-failure` to ignore the tests that are currently failing.  It doesn't seem to ignore exceptions with GHC 7.8.4, though.  I can look into it later, but I didn't want to delay the PR.

/cc @edsko

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -335,6 +335,7 @@ Test-Suite unit-tests
         random,
         hackage-security,
         tasty,
+        tasty-expected-failure,
         tasty-hunit,
         tasty-quickcheck,
         tagged,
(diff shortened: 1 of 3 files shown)


kosmikus:
@grayjay Thanks. This looks good to me, but I'll leave merging into `solver` to @edsko in order not to interfere with his work.

I wouldn't worry about "expected failures" too much for the time being. In fact, in the `solver` branch, it can even be helpful to have these as explicit failures for now, so that we can be absolutely sure we fix them all.

edsko:
Merged, many thanks! I agree with @kosmikus , I'd say let the failing test cases just fail. After all, on this branch, if there's something failing, we should really fix it. Bu

edsko:
Merged, many thanks! I agree with @kosmikus , I'd say let the failing test cases just fail. After all, on this branch, if there's something failing, we should really fix it. But it's okay for now.

grayjay:
No problem.

> After all, on this branch, if there's something failing, we should really fix it. But it's okay for now.

I figured we would remove the dependency on `tasty-expected-failure` before merging into master, so we wouldn't miss anything.

PR #3360, opened 2016-04-18
Rebase fixLinkDeps onto solver branch

grayjay:
All but the last two commits are from  #3359.

See #3357.  I rebased https://github.com/edsko/cabal/tree/pr/FixLinkDeps, except for the last commit, which handles a different issue.  I also removed my commits, because they weren't needed.

/cc @edsko

### DIFF
--- a/cabal-install/Distribution/Client/Dependency/Modular/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency/Modular/Dependency.hs
@@ -19,6 +19,7 @@ module Distribution.Client.Dependency.Modular.Dependency (
   , flattenFlaggedDeps
   , QualifyOptions(..)
   , qualifyDeps
+  , unqualifyDeps
     -- ** Setting/forgetting components
   , forgetCompOpenGoal
   , setCompFlaggedDeps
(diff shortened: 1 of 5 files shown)


edsko:
@kristen I'm somewhat confused. What is the status on #3327 now? Do we have a solution? Does the `solver` branch now have my fix for the incorrect link groups (the unqualifation-requalification thing) and your original solution to #3327? And the discussion at the end of https://github.com/haskell/cabal/pull/3357#issuecomment-211518643 we still don't have a good answer for? Is that correct?

grayjay:
@edsko This PR only contains your solution to #3268, which uses `unqualifyDeps`.  I wasn't sure what to do about #3327.  I think that my PR is incomplete because it doesn't contain `GoalReasonChain`s, and we haven't decided what to do with `lgBlame` and `lgConflictSet`.

edsko:
Yes, I think the answer to that isn't clear yet, but it would be good if we have have the test cases you have for #3327 available on this branch (where we expected them to fail right now).

grayjay:
I cherry-picked that test case, but I think it is currently passing because of a different fix.

edsko:
Aha, I see. That explains my confusion :) That's a little unfortunate.

grayjay:
Yeah, there is a lot of overlap between conflict sets. I'll try to come up with a more specific test.

grayjay:
I added a test in #3363.

PR #3365, opened 2016-04-21
Remove the 'run tests' step from setup-dev.sh

BardurArantsson:
As it turns out we often get spurious test failures, and it's generally
more useful to be able to actually build *everything* rather than having
the tests pass while building. (We already have pre-merge Travis checks,
so it's not that likely that something will slip through anyway.)


BardurArantsson:
I'm going to take the liberty of just merging this as it should be pretty harmless.

PR #3369, opened 2016-04-23
Add quickcheck test to compare dependency solving with and without backjumping

grayjay:
Here is the test that I mentioned in #3357.  It added a lot more code than I expected, and it doesn't seem to find bugs in conflict sets any faster than the existing tests, so I'm not sure it's worth merging.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -52,6 +52,7 @@ module Distribution.Client.Dependency (
     setShadowPkgs,
     setStrongFlags,
     setMaxBackjumps,
+    setEnableBackjumping,
     addSourcePackages,
     hideInstalledPackagesSpecificByUnitId,
     hideInstalledPackagesSpecificBySourcePackageId,
@@ -77,6 +78,7 @@ import Distribution.Client.Types
          , OptionalStanza(..), enableStanzas )
 import Distribution.Client.Dependency.Types
          ( PreSolver(..), Solver(..), DependencyResolver, ResolverPackage(..)
+         , EnableBackjumping(..)
          , PackageConstraint(..), showPackageConstraint
          , LabeledPackageConstraint(..), unlabelPackageConstraint
          , ConstraintSource(..), showConstraintSource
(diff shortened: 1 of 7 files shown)


cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -152,7 +154,8 @@ data DepResolverParams = DepResolverParams {
        depResolverAvoidReinstalls   :: Bool,
        depResolverShadowPkgs        :: Bool,
        depResolverStrongFlags       :: Bool,
-       depResolverMaxBackjumps      :: Maybe Int
+       depResolverMaxBackjumps      :: Maybe Int,
+       depResolverEnableBackjumping :: EnableBackjumping
```
kosmikus: Just trying to understand: by keeping `EnableBackjumping` separate from `MaxBackjumps`, then even with disabled "backjumping", in fact every step up in the tree will still be counted as a backjump, is that right?

cabal-install/Distribution/Client/Dependency/Types.hs
```diff
@@ -105,6 +106,9 @@ instance Text PreSolver where
       "choose"  -> return Choose
       _         -> Parse.pfail
 
+newtype EnableBackjumping = EnableBackjumping Bool
+  deriving Show
```
kosmikus: I think it's good that you introduce a `newtype` here. We should probably do the same for all the other `Bool` arguments to the solver. Would you perhaps be willing to do this (in a separate PR)?

kosmikus:
It looks good to me, and it makes sense as a test. Also, I like to have the option to switch off backjumping, also for testing manually. So I think this is useful. I'm just not sure if it wouldn't be better to combine the maximum backjumps and the enable/disable functionality in a single configuration option ... But that would probably be more invasive, so I think the current proposal is fine.

@23Skidoo Should we hold off merging things that aren't essential for 1.24 until after the release to make your life easier?

23Skidoo:
@kosmikus Thanks, but I think I can manage. Go ahead.

kosmikus:
@23Skidoo Ok, done. @grayjay Thanks!

cabal-install/Distribution/Client/Dependency.hs
grayjay: That's right.  Maybe I should rename the option, because it doesn't disable the backjump limit.  I used the backjump limit to limit the run time and memory usage of the individual test cases.  The backjump messages in the solver's log are also necessary because they are used for truncating the summarized log.  They prevent the solver from keeping the whole log in memory.

cabal-install/Distribution/Client/Dependency/Types.hs
grayjay: I'll look into it if I have time.

grayjay:
@kosmikus Thanks! Do you still want the two options to be combined?  One reason that I kept them separate is that the new option is only used for testing.

cabal-install/Distribution/Client/Dependency/Types.hs
grayjay: I added newtypes for the other options in #3377.

PR #3377, opened 2016-04-24
Use newtypes for Bool solver options

grayjay:
This commit shouldn't change any behavior, except possibly the serialization of GenericInstallPlan, InstallFlags, etc.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency/Types.hs
+++ b/cabal-install/Distribution/Client/Dependency/Types.hs
@@ -16,7 +16,14 @@
 module Distribution.Client.Dependency.Types (
     PreSolver(..),
     Solver(..),
+
+    ReorderGoals(..),
+    IndependentGoals(..),
+    AvoidReinstalls(..),
+    ShadowPkgs(..),
+    StrongFlags(..),
     EnableBackjumping(..),
+
     DependencyResolver,
     ResolverPackage(..),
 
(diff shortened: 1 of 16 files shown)


23Skidoo:
LGTM.

We may want to add a type class for such types, something like

```
class BooleanFlag a where
    asBool :: a -> Bool
```

and use `asBool` instead of `unReorderGoals`/`unAvoidReinstalls`/[...].

cabal-install/Distribution/Client/Dependency/Types.hs
```diff
@@ -106,9 +113,30 @@ instance Text PreSolver where
...
+
+newtype AvoidReinstalls = AvoidReinstalls { unAvoidReinstalls :: Bool }
+  deriving (Eq, Generic, Show)
+
+newtype ShadowPkgs = ShadowPkgs { unShadowPkgs :: Bool }
+  deriving (Eq, Generic, Show)
+
+newtype StrongFlags = StrongFlags { unStrongFlags :: Bool }
+  deriving (Eq, Generic, Show)
+
```
kosmikus: If all these have an extractor, and `Eq`, `Generic`, and `Show`, shouldn't then `EnableBackjumps` get the same treatment? (Yes, it's an option not exposed to the user, but still ...)

kosmikus:
Looks good to me as well.

grayjay:
I added a `BooleanFlag` type class in the last commit.  I think it would be more useful if we moved it to the library and used it to generalize the functions that handle boolean options, but that would be a bigger change.  I created a Bool instance and used GeneralizedNewtypeDeriving to reduce boilerplate.  Does that make sense, or would it be better to write out the instances?

I also derived Eq, Show, Generic, and BooleanFlag for EnableBackjumps.

23Skidoo:
Great, I'll merge this PR once the build bot is green.

23Skidoo:
>  I created a Bool instance and used GeneralizedNewtypeDeriving to reduce boilerplate. Does that make sense, or would it be better to write out the instances?

GHC extensions (except Template Haskell) are allowed, provided that they are available on all GHC versions we need to support. Using `GeneralizedNewtypeDeriving` is certainly fine.

>  I think it would be more useful if we moved it to the library and used it to generalize the functions that handle boolean options

The right place for `BooleanFlag` is probably in `Distribution.Simple.Setup`, near the `Flag` type:
https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Setup.hs#L122

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

> The right place for BooleanFlag is probably in Distribution.Simple.Setup, near the Flag type:
https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Setup.hs#L122

Would you like me to move it?

23Skidoo:
> Would you like me to move it?

If it's not too much trouble. It's probably better to do that right now than later.

grayjay:
I opened  #3379.

Issue #3348, opened 2016-04-16
bash completion at "cabal configure -w "


lspitzner:
"--list-options" is interpreted as the -w argument. which breaks:

    cabal configure -w <tab>cabal: Cannot find the program 'ghc'. User-specified path '--list-options' does not refer to an executable and the program is not on the system path.

How _should_ `cabal configure -w --list-options` behave, even? The general problem here is obvious, but there are two approaches to fixing.

23Skidoo:
> The general problem here is obvious, but there are two approaches to fixing.

And those are?

lspitzner:
1. make the bash completion script more clever. i.e. not only replace the last "word"/"param" (in bash sense) but the last "flag group"; grouped as in `cabal (install) (-w foo) (--whatever)`.
2. change cabal's (the argument parser's) behaviour, "fixing" the behaviour of `cabal configure -w --list-options`: to not parse `--list-options` as an argument to `-w`. But then, maybe the user is using the very popular `--list-options` compiler. This would be implemented by testing that any/the last parameter is `--list-options`. I don't know how much special-casing in the parser this requires.

As 1. requires knowledge about any (future?) flags that have (mandatory) parameters, I favor solution 2.

23Skidoo:
I don't think the completion script should be invoking `cabal --list-options` with any arguments other than the command name. The output for `cabal install --list-options` and `cabal install -w foo --list-options` will be identical anyway. I'm also reluctant to add special-casing to Cabal's command line parser to work around what is essentially a bug in the completion script.

lspitzner:
but with that approach, you get no completion for `cabal configure -w /usr/b` (…`in/ghc`) (which i don't mind too much, as long as it does not break.) But consider that `cabal --no-require-sandbox sandbox --verbose init` is a perfectly valid command; trying to remove flags with the intention of passing `--list-options` would require knowledge about which flags have parameters.

atm i do not see any quick fix. other tools must have solved this already; analysing other's solutions would be my first step if i find time to look at this more.

@dcoutts you are listed as author, any suggestions?

23Skidoo:
I think that a simple solution would be to assume that there are no flags with parameters (unless you want to add a way of extracting that info from `cabal`). So you'd start looking at the input string from the left, and once you'd find a known command name, you would invoke `cabal $cmd --list-options` to get the completion context. Otherwise the completion context would be given by `cabal --list-options`.

[...]

PR #3384, opened 2016-04-29
Fix bugs with bash completion

kolmodin:
The bash completion suffers from a number of issues,
some fix with this patch;

With this patch we expand the path to cabal as well as its arguments.
For example;
~/code/cabal configure --with-ghc=$GHC <tab><tab>
will now work. Both ~ and the $GHC variable will be expanded.

If there are no .cabal files, or multiple files, the script will no longer
crash.

We still invoke cabal to get completions. If cabal detects an error and writes
errors to STDERR, these will no longer be printed to the console. Instead we
don't offer any completions.

### DIFF
--- a/cabal-install/bash-completion/cabal
+++ b/cabal-install/bash-completion/cabal
@@ -10,9 +10,10 @@
 #   - executable|test-suite|benchmark for the three
 _cabal_list()
 {
-	cat *.cabal |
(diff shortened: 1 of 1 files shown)


23Skidoo:
Looks OK, from what I can tell.

lspitzner:
LGTM.

fixes #3348

23Skidoo:
Merged, thanks!

PR #3385, opened 2016-05-02
Don't ignore 206 response in plain-http transport

edsko:
The plainHttpTransport implementation saved the resulting file only when the
server responded with 200 OK, thus ignoring the payload from the server when
the server responded with a 206 Partial Content. This was causing incremental
updates through hackage-security to fail (it wouldn't fail fatally;
hackage-security would notice the invalid timestamp and download the whole
thing instead, but of course this means we're not actually getting any
incremental behaviour).

### DIFF
--- a/cabal-install/Distribution/Client/HttpUtils.hs
+++ b/cabal-install/Distribution/Client/HttpUtils.hs
@@ -635,7 +635,7 @@ plainHttpTransport =
       (_, resp) <- cabalBrowse verbosity Nothing (request req)
       let code  = convertRspCode (rspCode resp)
           etag' = lookupHeader HdrETag (rspHeaders resp)
-      when (code==200) $
+      when (code==200 || code==206) $
         writeFileAtomic destPath $ rspBody resp
       return (code, etag')
 


23Skidoo:
Are other transports not affected?

hvr:
@23Skidoo `curl` is not; Haven't tried `wget` nor `powershell` (latter one requires windows for testing...)

23Skidoo:
Does the code that handles etags also need to be updated? It only saves the etag [when the response code is 200](https://github.com/edsko/cabal/blob/pr/plainHttpTransport206/cabal-install/Distribution/Client/HttpUtils.hs#L121).

23Skidoo:
In general, this looks OK, but needs a comment explaining when we can expect the 206 response code.

hvr:
@23Skidoo at least as far as the incremental 01-index updating is concerned, the .etag file is not used   (it's subsumed by the `snapshot.json` meta-data) afaik. But I haven't looked yet too closely at the HTTP implementation inside `cabal-install` to where the etag-handling occurs

hvr:
PS: The 206 code is only expected when the GET request had a `Range:` header (however, if the server doesn't understand/support range-request, it would most likely respond as if it was an ordinary request, i.e. with a 200 code and the complete non-partial file content).

23Skidoo:
Our powershell code always returns 200 in case of success, which may be incorrect when doing a range request. Also, it doesn't seem to support etags.

23Skidoo:
OK, I'll merge this right now and add a comment later.

23Skidoo:
Also cherry-picked into 1.24.

23Skidoo:
Related: #3388.

Issue #3389, opened 2016-05-02
bootstraping cabal-install-1.24.0.0 fails because of hackage-security haddocks


aisamanra:
In particular: hackage-security has Haddock comments on GADT constructors, and Haddock chokes on those. This is an [old bug](http://trac.haskell.org/haddock/ticket/43) (eight years ago!) and [will be fixed in GHC 8.](https://phabricator.haskell.org/D1086), but that doesn't help with bootstrapping for current GHC versions, where `bootstrap.sh` will fail with

```
Error during cabal-install bootstrap:
Documenting the hackage-security package failed.
```

The specific error that arises when generating the haddocks for `hackage-security` looks like:

```
src/Hackage/Security/TUF/Layout/Index.hs:48:5:
    parse error on input ‘-- | Package-specific metadata (@targets.json@)’
```

I did get around this on my own by modifying [the variable `NO_DOCS_PACKAGES_VER_REGEXP`](https://github.com/haskell/cabal/blob/c397b68b50129c1fe009cbaba76b37d75a94c824/cabal-install/bootstrap.sh#L217) so that bootstrap.sh wouldn't build haddocks for `hackage-security`, and if that's an acceptable workaround, I'd be happy to open a PR with that (admittedly trivial) change, but I don't know if that'd be the preferred long-term workaround.

23Skidoo:
Duplicate of https://github.com/well-typed/hackage-security/issues/149.

> I did get around this on my own by modifying [the variable `NO_DOCS_PACKAGES_VER_REGEXP`](https://github.com/haskell/cabal/blob/c397b68b50129c1fe009cbaba76b37d75a94c824/cabal-install/bootstrap.sh#L217) so that bootstrap.sh wouldn't build haddocks for `hackage-server`, and if that's an acceptable workaround, I'd be happy to open a PR with that (admittedly trivial) change, but I don't know if that'd be the preferred long-term workaround.

Please do, I think that this is an acceptable workaround for the time being.

23Skidoo:
Closing this ticket as a duplicate, but please send us a PR with that change.

PR #3393, opened 2016-05-02
Prevent bootstrap.sh from building haddocks for hackage-security

aisamanra:
An outstanding bug in GHC pre-8 causes haddock comments on GADT constructors to fail, which in turn prevents the bootstrapping process from finishing. This works around it by skipping haddocks for hackage-security.

This addresses the problem described in #3389 and well-typed/hackage-security#149, and this fix won't be necessary at all for GHC 8+.


23Skidoo:
Merged, thanks!

23Skidoo:

PR #3396, opened 2016-05-04
Make `Validate` a proper monad

edsko:
This makes it a bit easier to modify it (for example, during debugging).


23Skidoo:
Merged, thanks!

PR #3397, opened 2016-05-04
Add `conflictSetOrigin` debugging field

edsko:
This field is only added if the `debug-conflict-sets` flag is specified to
`cabal config`; it adds a tree to `CallStack`s to a `ConflictSet`. This is very
useful when trying to understand how a certain conflict set was constructed.

/cc @kosmikus


edsko:
Note that this requires GHC 7.10; trying to configure with `-f debug-conflict-sets` with older GHC will result in a solver failure. (The docs for `CallStack` claim we need ghc 8.0; I think this is a documentation bug: http://ghc.haskell.org/trac/ghc/ticket/12013 .)

23Skidoo:
Merged, thanks!

PR #3399, opened 2016-05-04
Simplify `GoalChoice`

### LINKED ISSUE #3398: Solver tree contains redundant information
As I was getting ready to write the documentation for the new backjumping behaviour in the solver, I realized that the tree now contains redundant information (maybe it always did?). In particular, in a `GoalChoice` node we record, for each goal, why we are introducing that goal; to be precise, `OpenGoal` contains the `GoalReason`. But then we record that same `GoalReason` again on the `PChoice`/`FChoice`/`SChoice` nodes in the tree. I've written a function to check that these are always the same:

``` haskell
checkInvariant :: Tree QGoalReason -> Tree QGoalReason
checkInvariant tree =
    if go Nothing tree then tree
                       else error "GoalReason invariant violated"
  where
    go :: Maybe QGoalReason -> Tree QGoalReason -> Bool
    go _   (GoalChoice       psq) =                   all (uncurry goG) (PSQ.toList psq)
    go mgr (PChoice _ gr     psq) = mgr == Just gr && all (uncurry goP) (PSQ.toList psq)
    go mgr (FChoice _ gr _ _ psq) = mgr == Just gr && all (uncurry goF) (PSQ.toList psq)
    go mgr (SChoice _ gr _   psq) = mgr == Just gr && all (uncurry goS) (PSQ.toList psq)
    go _   (Done _)               = True
    go _   (Fail _ _)             = True

    goG :: OpenGoal () -> Tree QGoalReason -> Bool
    goG (OpenGoal _dep gr) subtree = go (Just gr) subtree

    goP :: POption -> Tree QGoalReason -> Bool
    goP _ = go Nothing

    goF, goS :: Bool -> Tree QGoalReason -> Bool
    goF _ = go Nothing
    goS _ = go Nothing
```

and on the examples where this finished the invariant was true (this invariant verifies the entire tree so clearly we can only check this for small examples).

If @kosmikus and @grayjay agree that this information is redundant, I suggest we remove the redundancy from the tree, and just work with `Tree ()` instead of `Tree QGoalReason` (or, barring that, at the minimum document the invariant that these much always equal).

edsko:
This PR makes two changes:

* Replace `OpenGoal` with `Goal` as keys for the PSQ in `GoalChoice`
* Document the invariant that the `QGoalReason` cached in `PChoice` and co must equal the reason stored on `GoalChoice` node directly above them.

Fixes #3398.

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -111,8 +111,9 @@ build = ana go
     -- it from the queue of open goals.
     go bs@(BS { rdeps = rds, open = gs, next = Goals })
       | P.null gs = DoneF rds
-      | otherwise = GoalChoiceF (P.mapWithKey (\ g (_sc, gs') -> bs { next = OneGoal g, open = gs' })
-                                              (P.splits gs))
+      | otherwise = GoalChoiceF $ P.mapKeys close
+                                $ P.mapWithKey (\ g (_sc, gs') -> bs { next = OneGoal g, open = gs' })
+                                $ P.splits gs
 
     -- If we have already picked a goal, then the choice depends on the kind
     -- of goal.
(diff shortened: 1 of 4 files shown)


23Skidoo:
I presume that this is part of fixing #3398?

edsko:
Sorry, I should have mentioned that. Yes, this PR would close #3398.

kosmikus:
LGTM.

23Skidoo:
Merged, thanks!

PR #3405, opened 2016-05-06
Restore Verbose-level logging of the haddock command line

randen:
(This is against tag 1.24 but should be considered for where ever convenient.)

Prior to the response file code, all of the details of the
command line for haddock would have been logged (with level
== Verbose), so this corrects an oversight and brings the
information in the logging back to what it used to be for
cabal's haddock invocations, by logging the contents of the
response file.

  * Restore the logging of the entire command line used when
    invoking haddock to what it used to be prior to adding
    the response-file creation.


23Skidoo:
LGTM.

23Skidoo:

PR #3383, opened 2016-04-27
Move ConstraintSource and ResolverPackage to Distribution.Solver.*

BardurArantsson:
As a follow-up to #3381 these commits move ConstraintSource and ResolverPackage too.

Incidentally, ResolverPackage is a bit anemic as a module, but I've decided to stick with the general pattern of `Distribution.Solver.Types.*` for now. I think it'll be easier to see if there's anything that should perhaps be re-combined later on.

PR #3406, opened 2016-05-06
Move solver types to Distribution.Solver.* namespace

BardurArantsson:
This completes the move of all the solver types which the modular solver requires to Distribution.Solver.*.

Like last time there are few *tiny* modules, but again I've decided to go for consistency; see #3383.

This means that we now have:

```sh
$ grep -r Client cabal-install/Distribution/Solver|wc -l
0
```

which is to say that there are no longer any dependencies from the Solver namespace to the Client namespace. We should probably add a Travis check or something that ensures that none are added back accidentally.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency/Types.hs
+++ b/cabal-install/Distribution/Client/Dependency/Types.hs
@@ -17,7 +17,6 @@ module Distribution.Client.Dependency.Types (
 
     DependencyResolver,
 
-    PackagePreferences(..),
     PackagesPreferenceDefault(..),
 
   ) where
@@ -25,10 +24,9 @@ module Distribution.Client.Dependency.Types (
 import Data.Char
          ( isAlpha, toLower )
 
-import Distribution.Solver.Types.InstalledPreference
 import Distribution.Solver.Types.LabeledPackageConstraint
-import Distribution.Solver.Types.OptionalStanza
 import Distribution.Solver.Types.PkgConfigDb ( PkgConfigDb )
+import Distribution.Solver.Types.PackagePreferences
 import Distribution.Solver.Types.PackageIndex ( PackageIndex )
 import Distribution.Solver.Types.Progress
 import Distribution.Solver.Types.ResolverPackage
@@ -39,8 +37,6 @@ import qualified Distribution.Compat.ReadP as Parse
 import Distribution.Simple.PackageIndex ( InstalledPackageIndex )
 import Distribution.Package
          ( PackageName )
-import Distribution.Version
-         ( VersionRange )
 import Distribution.Compiler
          ( CompilerInfo )
 import Distribution.System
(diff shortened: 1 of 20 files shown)


cabal-install/Distribution/Client/Dependency/Types.hs
```diff
@@ -1,16 +1,4 @@
...
--- |
--- Module      :  Distribution.Client.Dependency.Types
--- Copyright   :  (c) Duncan Coutts 2008
--- License     :  BSD-like
---
--- Maintainer  :  cabal-devel@haskell.org
--- Stability   :  provisional
--- Portability :  portable
---
--- Common types for dependency resolution.
```
23Skidoo: I'm okay with removing the header, but it'd be nice to preserve the top comment about the purpose of the module.

BardurArantsson:
Actually, perhaps we can drop the Travis check and just use an "internal library"?

@ezyang Do you think "internal library" support is robust enough at this point to move the Dependency.Solver.* bits into such a library? (I'm *assuming* tests will have access to such a library implicitly. Right?)

cabal-install/Distribution/Client/Dependency/Types.hs
BardurArantsson: I actually removed it because the "purpose" isn't very clear any longer. The module is now a sort of weird mish-mash.

(I think this module should perhaps also be split, but I'm kind of waiting to see where migrating the Tests takes me... The module may become entirely redundant.)

23Skidoo:
LGTM, but would be nice to see some of the tiny modules recombined later on.

BardurArantsson:
Yup, I fully intend to at some point, but I think it'll have to wait to see what happens when I try migrating the test cases, etc. I'm not sure we actually have the "full set" yet.

BardurArantsson:
Hm, has Travis been disabled or something? I'm not seeing any "CI status" on this PR.

23Skidoo:
Try tweaking something in the code (e.g. a comment), amend the last commit, and do a force-push.

23Skidoo:
Actually don't do that, I found a way to fix this.

BardurArantsson:
Check that. Thanks!

23Skidoo:
OK, for some reason I can't re-trigger the Travis hook, but closing and reopening the PR should work.

23Skidoo:
Merged, thanks!

ezyang:
Re adding an internal library, I think we just need to merge https://github.com/haskell/cabal/pull/3356

BardurArantsson:
OK, sounds good. I'll postpone the attempt until that's merged then :).

23Skidoo:
@ezyang 
> I think we just need to merge #3356

I think that all review comments in #3355 and #3356 have been addressed, so feel free to merge.

PR #3415, opened 2016-05-09
Add '--{ghc,haddock}-{option[s],path}' flag to 'upload'.

### LINKED ISSUE #3414: Add support for `--hadock-options` to `cabal upload --doc`
The `cabal upload --doc` feature implemented by @bennofs  is super useful. However, currently I can't seem to be able to pass `--haddock-options` to tweak the generated documentation (e.g. to set `--use-unicode` or `--hyperlinked-source`).

So right now I have to fallback to `cabal haddock --for-hackage` + uploading the generated doc-tarball. However, if `cabal upload` supported the `--haddock-options` flag it'd be more convenient.

/cc @thoughtpolice

23Skidoo:
Fixes #3414.


hvr:
wow... that was quick :-)

23Skidoo:
@dcoutts, do you think this is acceptable or is it better to just refer the users to `haddock --for-hackage`?

23Skidoo:
So we discussed this with @dcoutts and decided that it's better to make `haddock --for-hackage` more visible instead of cluttering the interface for `upload`. Note that you can also set haddock options in `cabal.config` or `cabal.project`.

PR #3421, opened 2016-05-11
Remove 'upload --check'.

### LINKED ISSUE #1823: 'cabal upload --check' broken
I noticed that `cabal upload --check` is broken, as it uses the [hardcoded](https://github.com/haskell/cabal/blob/3f061c77fa3/cabal-install/Distribution/Client/Upload.hs#L45) URL:

```haskell
checkURI :: URI
Just checkURI = parseURI "http://hackage.haskell.org/cgi-bin/hackage-scripts/check-pkg"
```

Which doesn't seem to work anymore (it responds with a 404). Is it planned to have this functionality restored?

23Skidoo:
Plus some other small tweaks.

Fixes #1823.

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -351,7 +351,6 @@ instance Semigroup SavedConfig where
 
       combinedSavedUploadFlags = UploadFlags {
         uploadCandidate   = combine uploadCandidate,
-        uploadCheck       = combine uploadCheck,
         uploadDoc         = combine uploadDoc,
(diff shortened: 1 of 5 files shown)


gbaz:
Now if someone uploads with -c, what will happen? Will it error? I hope so, as if someone had a -c based workflow before I'd hate for that to turn into an accidental upload, even if just of a candidate...

23Skidoo:
@gbaz 
```
$ ./cabal-install/dist/build/cabal/cabal upload -c foo.tar.gz
cabal: unrecognized 'upload' option `-c'
```

PR #3429, opened 2016-05-13
Fix various behavior differences between the transports

bennofs:
This PR has a few commits that each fix a difference in what some transport
does in comparision to all the others. See the commits for more details,
they should be pretty self-explanatory.


23Skidoo:
LGTM.

gbaz:
Thanks especially for fixing the wget stuff. I had a nightmare of a time trying to get it working even that well, and as the comments indicate, at a certain point just got stuck :-)

23Skidoo:
Merged, thanks! Also cherry-picked into 1.24.

PR #3439, opened 2016-05-17
travis-script.sh: test bootstrapping

23Skidoo:
Same as #3349, but now also installs `bytestring-builder` when `bytestring` is < 0.10.2.

### DIFF
--- a/travis-script.sh
+++ b/travis-script.sh
@@ -1,7 +1,38 @@
 #!/usr/bin/env bash
 set -ev
 
+# ---------------------------------------------------------------------
+# Bootstrap cabal, to verify bootstrap.sh script works.
+# ---------------------------------------------------------------------
+
+OLD_CWD=$PWD
+
+# Bootstrap
+cd Cabal
+ghc --make Setup
+./Setup configure --user
+./Setup build
+./Setup install
+cd ../cabal-install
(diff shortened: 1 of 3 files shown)


travis-script.sh
```diff
@@ -1,8 +1,39 @@
...
+OLD_CWD=$PWD
+
+# Bootstrap
+cd Cabal
+ghc --make Setup
+./Setup configure --user
+./Setup build
+./Setup install
+cd ../cabal-install
+env EXTRA_CONFIGURE_OPTS="" ./bootstrap.sh --no-doc
```
23Skidoo: So the problem now is that when we install Cabal before bootstrapping, Cabal will be compiled against the version of binary that came with GHC. I think we should modify `bootstrap.sh` to support installing the currently checked-out version of Cabal, then we can install it after `binary`.

23Skidoo:
Travis build times seem to be a bit slower, though sometimes comparable:

https://travis-ci.org/haskell/cabal/builds/130951419
https://travis-ci.org/haskell/cabal/builds/130658375
https://travis-ci.org/haskell/cabal/builds/130429059

I think we should look into enabling caching on Travis next.

23Skidoo:
Passes on Travis.

PR #3441, opened 2016-05-18
Added empty exposed modules check for library targets

headprogrammingczar:
What it says on the tin.

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -218,6 +218,10 @@ checkLibrary pkg lib =
             "Duplicate modules in library: "
          ++ commaSep (map display moduleDuplicates)
 
+  , check (not (null (libModules lib))) $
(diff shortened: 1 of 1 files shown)


23Skidoo:
Apparently this breaks the test suite. Can you please look into that?

headprogrammingczar:
The fails appear to be that `(libModules lib)` is erroneously empty, but I can't immediately see why. Will keep looking.

headprogrammingczar:
It looks like I had the test backwards. I'm still not sure why tests were passing on my machine, but when I ran the sandbox's result manually I finally managed to reproduce it and catch my mistake. We'll see what CI makes of this.

23Skidoo:
Merged in 0252a0aed79e881d4f6210d9296216aef3c301d0, thanks!

hvr:
I hope this is only an advisory warning not causing `cabal check` to return non-zero? I have a few packages which have empty exposed-modules in their library stanze on purpose

PR #3447, opened 2016-05-20
Implement -j in travis and bootstrap script

tuncer:


tuncer:
Need to modify so that -j is not used with older GHC.

23Skidoo:
I think we should also support passing `-j` without argument to make use of `cabal-install`'s core number autodetection feature (IIRC, GHC and `Setup build` also support this). E.g. what if Travis switches to 8-core instances tomorrow?

tuncer:
Sounds reasonable. I've only had to adjust bootstrap.sh for that.

1. travis-script.sh uses getopts, and if passed `-j`, then `$OPTARG` should be empty and therefore `$jobs` = `-j`. If no `-j` is supplied, the default is `-j 1`.

2. bootstrap.sh also defaults to `-j 1`, and if `-j` or `--jobs` is used, then we either set the explicit number if provided, or `-j`.

tuncer:
Once this is all working, we can omit the `4` in `.travis.yml`.

tuncer:
There's a problem with the bootstrap.sh option parsing loop. On 7.4 it exits with `Unknown argument or option, quitting: ` and when given `-j 4`, it's `Unknown argument or option, quitting: -j 4`.

tuncer:
I cannot reproduce the while loop error locally. I can successfully call `bootstrap.sh `, `bootstrap.sh -j`, and `bootstrap.sh -j 4`.

tuncer:
TODO: remove ` 4` in `travis-script.sh` call.

23Skidoo:
Merged, thanks!

ezyang:
@tuncer, out of curiosity, have you ever called `./travis-script.sh` on a local machine (as opposed to having Travis run it?)

tuncer:
@ezyang, I recall at least one full run, yes. However, I relied on travis for most of it, due to the tested GHC versions. Why do you ask?

ezyang:
Oh, because I forgot to reimplement the cores/command line parsing when I rewrote the Travis script to use new-build. I'll add it back.

tuncer:
@ezyang, is the branch visible somewhere?

ezyang:
I have not added back the flags yet, so no branch yet -_-.

juhp:
Late, but I think `setup build -j1` fails with the Cabal in ghc-7.8 also...

It would be better not to do anything for the default "-j1" case.

tuncer:
@ezyang, can we restore `-jN` in bootstrap.sh and travis-script.sh now, or is there a blocker?

PR #3448, opened 2016-05-20
No need to install cabal-install package

tuncer:
Now that we bootstrap Cabal and cabal-install, we don't have to install
the package anymore.


23Skidoo:
Merged, thanks!

PR #3459, opened 2016-05-25
bootstrap.sh bugfixes and sh compat

tuncer:
Fix errors (like vsn regex var name typo) and also POSIX compliance issues in bootstrap.sh.


tuncer:
I've only double quoted what shellcheck complained about. Therefore, STM_VER_REGEXP, ASYNC_VER_REGEXP and HASHABLE_VER_REGEXP are left unquoted. Maybe we should for consistency of the code.

tuncer:
I've only double quoted what shellcheck complained about. Therefore, STM_VER_REGEXP, ASYNC_VER_REGEXP and HASHABLE_VER_REGEXP are left unquoted. Maybe we should for consistency of the code.

tuncer:
@23Skidoo, because we don't have a test suite for the shell script and it's not a small diff, I'd suggest a thorough review to avoid regressions.

23Skidoo:
This breaks Travis.

tuncer:
I'll bisect it.

tuncer:
I wish Github would allow triggering CI for each commit in the branch. That would help with bisecting.

tuncer:
I'll have to git-bisect locally.

tuncer:
Only 7.10.3 fails. @23Skidoo, any idea which of the changes is at fault?

23Skidoo:
Failure looks quite weird:

```
/usr/bin/ld:/tmp/cabal- 418b5a2c/ghc10394_0/ghc_198.ldscript: file format not recognized; treating as linker script
/usr/bin/ld:/tmp/cabal- 418b5a2c/ghc10394_0/ghc_198.ldscript:1: syntax error
collect2: ld returned 1 exit status
<no location info>: ghc: phase `Linker' failed (exitcode = 1)
```

Maybe related to ld detection that the bootstrap script does?

tuncer:
I'll back out one diff that might cause this, but it's odd that this only breaks 7.10.3 out of all the GHC versions.

tuncer:
While we're at it, shellcheck complains about constructs like

``
echo "-j   Concurrent workers"
``

as found in travis-script.sh, warning that IIUC -j would be passed to echo. I left that as is because I'm unconvinced. Thoughts?

Other than that, the double quoting warnings of shellcheck seem overzealous and were the cause of the the breakages in this pr.

tuncer:
I need to check the while loop changes again, to make sure it's correct.

cabal-install/bootstrap.sh
```diff
@@ -52,24 +55,32 @@ SCOPE_OF_INSTALLATION="${SCOPE_OF_INSTALLATION:---user}"
 DEFAULT_PREFIX="${HOME}/.cabal"
 
 # Try to respect $TMPDIR.
-[ -"$TMPDIR"- = -""- ] &&
-  export TMPDIR=/tmp/cabal-$(echo $(od -XN4 -An /dev/random)) && mkdir $TMPDIR
+if [ -"$TMPDIR"- = -""- ]; then
+    TMPDIR=/tmp/cabal-$(od -XN4 -An /dev/random)
+    mkdir "$TMPDIR"
+    export TMPDIR
+fi
```
tuncer: @23Skidoo, a couple of questions regarding TMPDIR:

1. Where is this used? Is the env var reused by cabal-install?
2. It goes against convention to name it TMPDIR because you set TMPDIR to deviate from /tmp, so referencing a directory inside /tmp via TMPDIR isn't right.

How about this instead: `mktemp -d -p /tmp -t cabal-XXXXXXXX`. This is portable to OpenBSD, FreeBSD, Linux and Solaris. The parent dir falls back to /tmp (if given via -p or assumed to be default if not) if TMPDIR is not set. The result will look like `/tmp/cabal-XYXYXYXY`.

Because I didn't figure out where TMPDIR is used in cabal and to see if this is the cause of the 7.10.3 build failure, I'll add a test commit.

tuncer: @23Skidoo ping?

cabal-install/bootstrap.sh
```diff
@@ -63,8 +63,9 @@ for c in $CC gcc clang cc icc; do
 done
 
 # None found.
-[ -"$CC"- = -""- ] && die 'C compiler not found (or could not be run).
-  If a C compiler is installed make sure it is on your PATH, or set $CC.'
+[ -"$CC"- = -""- ] && die "\
+C compiler not found (or could not be run).\n\
+If a C compiler is installed make sure it is on your PATH, or set $CC."
```
23Skidoo: I don't think we want to expand `$CC` here.

cabal-install/bootstrap.sh
23Skidoo: @tuncer I'm actually not sure myself where it's used, but your change looks fine to me.

23Skidoo:
LGTM except the `$CC` expansion in error message (first commit).

cabal-install/bootstrap.sh
tuncer: Then maybe it should be `CC` instead?

tuncer: I guess we can leave it as $CC to make the message more obvious.

tuncer: Makes sense.

tuncer:
Removed the $CC commit.

23Skidoo:
OK, I'll wait until Travis is green and then merge.

cabal-install/bootstrap.sh
tuncer: We should figure out how and where it's used so that we can ideally use a separate name here.

23Skidoo:
Merged, thanks!

PR #3463, opened 2016-05-28
travis.sh cleanup

tuncer:


tuncer:
@23Skidoo, as discussed.

tuncer:
Let's see if keeping the dirs will improve CI times.

tuncer:
@23Skidoo, keeping the dirs got us the following error:
```
In order, the following would be installed:
Cabal-1.25.0.0 (reinstall)
cabal: The following packages are likely to be broken by the reinstalls:
hackage-security-0.5.1.0
Use --force-reinstalls if you want to install anyway.
```

tuncer:
@23Skidoo, should I omit 7872bcc?

tuncer:
I'll omit 7872bcc for now, and we can tackle it separately from the other commits.

23Skidoo:
Merged, thanks!

PR #3468, opened 2016-05-30
Documentation for known os and arch aliases

### LINKED ISSUE #2138: documentation needs an os() and arch() list
The documentation says that os() uses the value of System.Info.os, and arch() System.Info.arch. But when doing something OS-specific, it's pretty common to not have easy access to every OS installed in order to go look up what the System.Info.os value is. Probably even more common when dealing with something architecture specific.

For example, I know that i need to add ! os(hurd) to my cabal file to fix it on the hurd, without needing to make any other hurd-specific changes. But I don't know what the actual os value is, and will need to get a hurd VM to find out. Update: Apparently it's os(gnu).

So, it would be great to include a list of common (and not so common) os names in the documentation. Those I know of include: linux, windows, darwin (OSX), solaris, gnu (hurd).

Or, perhaps, provide these lists not in the documentation, but in the code.

accraze:
This PR contains documentation for the known `os()` and `arch()` aliases.

Resolves #2138

### DIFF
--- a/Cabal/Distribution/System.hs
+++ b/Cabal/Distribution/System.hs
@@ -73,6 +73,18 @@ data ClassificationStrictness = Permissive | Compat | Strict
 -- * Operating System
 -- ------------------------------------------------------------
 
+-- | These are the known OS names: Linux, Windows, OSX
+--  ,FreeBSD, OpenBSD, NetBSD, DragonFly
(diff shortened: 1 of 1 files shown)


23Skidoo:
Merged, thanks! May be also worth it to do the same thing for the manual.

PR #3478, opened 2016-06-04
travis-script.sh: use shell arith expansion

tuncer:
Instead of spawning expr we can use sh arithmetic expansion.


23Skidoo:
LGTM, the segfault is probably just Travis being wonky. Restarted the build.

23Skidoo:
Merged, thanks!

Issue #3453, opened 2016-05-22
Resolving dependencies for wai repo on GHC 7.6.3 times out with Travis


snoyberg:
We have regular build failures on the [wai repo](https://github.com/yesodweb/wai) due to the build timing out, e.g. https://travis-ci.org/yesodweb/wai/jobs/132019472. I don't know if there is legitimately no build plan, or if the solver is simply having difficulty finding one. I've kicked off some dependency solving on one of my servers to try to determine if there's a workaround. Nonetheless, taking over 10 minutes to solve dependencies is surprising.

Reproduction:

* Tested with GHC 7.6.3 on Linux
* Tested with cabal-install 1.16, 1.22, and 1.24
* Check out the wai repo at commit c9427c887a40dbc35b6f4cca58f5096cd0a1b6e4 to the directory /home/travis/build/yesodweb/wai
* Run the following command:

```
cabal install --enable-tests --enable-benchmarks --force-reinstalls --ghc-options=-O0 --reorder-goals --max-backjumps=-1 /home/travis/build/yesodweb/wai/warp/ /home/travis/build/yesodweb/wai/wai-conduit/ /home/travis/build/yesodweb/wai/wai-websockets/ /home/travis/build/yesodweb/wai/wai/ /home/travis/build/yesodweb/wai/wai-handler-launch/ /home/travis/build/yesodweb/wai/auto-update/ /home/travis/build/yesodweb/wai/warp-tls/ /home/travis/build/yesodweb/wai/wai-handler-fastcgi/ /home/travis/build/yesodweb/wai/mime-types/ /home/travis/build/yesodweb/wai/wai-extra/ /home/travis/build/yesodweb/wai/wai-app-static/
```

Expected:

* Either dependency solving succeeds or fails, in a reasonable amount of time
* Very long period of inactivity

grayjay:
I started looking into this issue.  There is no solution, because there is a conflict between `warp`'s `doctest` test suite and its benchmarks with GHC 7.6.3.

The test requires `doctest`, which requires the package `ghc`.  `ghc` cannot be reinstalled, and the installed version depends on the installed `bytestring` (`0.10.0.2`).  The benchmark requires `criterion >= 1`.  `criterion-1.0.0.0` through `1.1.1.0` all require `aeson >= 0.8`, and `aeson-0.8.0.0` through `0.11.2.0` require `bytestring >= 0.10.4.0`.

[...]

PR #3480, opened 2016-06-05
Fix bug in dependency solver stanza preferences

grayjay:
The first commit creates a newtype for the weak/trivial attribute of package flags and stanzas.

The second commit fixes a bug in dependency solver stanza preferences. Previously, stanza choices were set to "weak" even when the stanza preference did not apply. See https://github.com/haskell/cabal/issues/3453#issuecomment-222580111.

@kosmikus Does it make sense to call the newtype `WeakOrTrivial`?  I'm still not completely sure I understand the meaning of the weak/trivial fields in the solver's `FChoice` and `SChoice` nodes.


kosmikus:
LGTM.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #3500, opened 2016-06-16
cabal-install: make user config optional if possible

### LINKED ISSUE #3494: cabal-install-1.24 seems to require the creation of ~/.cabal/config for commands that did not require it in the past
It seems in cabal-install-1.24 that commands like `register` and `sdist` now require the create of a `~/.cabal/config` file which fails during nix builds.

See this bug for details:

https://github.com/NixOS/nixpkgs/issues/16144

At the core, it seems that `loadConifg` requires the creation of a `~/.cabal/config` if such a file does not exist. While that has been the case for a while, changes made by this patch,

https://github.com/haskell/cabal/commit/3bb8c7a9a2499dc2a06325d290628a8c7d8ab6bf#diff-53e6b83dd13dd8fb7781bc0ce3e6b36bL282

caused more commands to be affected by it. It's not clear that is a 'bug' -- but it does seems to cause trouble anyway.

ttuegel:
Fixes #3494.

### DIFF
--- a/cabal-install/Main.hs
+++ b/cabal-install/Main.hs
@@ -177,6 +177,7 @@ import Data.Maybe               (listToMaybe)
 import Data.Monoid              (Monoid(..))
 import Control.Applicative      (pure, (<$>))
 #endif
+import Control.Exception        (SomeException(..), try)
 import Control.Monad            (when, unless)
 
 -- | Entry point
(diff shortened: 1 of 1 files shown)


23Skidoo:
LGTM.

PR #3506, opened 2016-06-21
Exit with non-0 code when build fails

23Skidoo:
Fixes issue #3495

This also adds a stub where we can add more detailed failure reporting,
but this starts with nothing more than the exit code.

Unlike the existing "cabal install" failure reporting, we don't need or
want "build" failure reporting to be quite so noisy in common cases like
the one local package failing to build, since ghc reports its own errors
and there's enough context to see what's going on.

(cherry picked from commit ba2065f41b90a99045500bbdacac00a150c03601)


23Skidoo:
This is a backport of #3504 for the 1.24 branch.

PR #3507, opened 2016-06-22
init: Only add exposed-modules which are in an hs-source-dir

### LINKED ISSUE #3484: cabal init picks up source files without adding directories to hs-source-dirs
Steps to reproduce:

1. Create a subfolder `foo` and put `A.hs` in it
2. Run `cabal init` (in the parent folder)

Expected behavior:

EITHER:
1. No modules are added to `exposed-modules` and `hs-source-dirs` is commented out
2. `A` is added to `exposed-modules` and `hs-source-dirs` contains `foo`

Actual behavior: `A` is added to `exposed-modules` but `hs-source-dirs` is not updated.

A fairly minor bug but a bug nonetheless.

byorgey:
Only those modules found in a directory in hs-source-dirs will be added
to exposed-modules.  If a directory 'src' is found it will be used as
hs-source-dirs (note in this case any .hs files in the root directory
will NOT be added to exposed-modules).  If any directories are specified
on the command line with --source-dir, they will be used.  If no
directories are specified on the command line and there is no directory
named src, the root directory will be used.

Fixes #3484.

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -29,7 +29,7 @@ import System.Directory
   ( getCurrentDirectory, doesDirectoryExist, doesFileExist, copyFile
   , getDirectoryContents, createDirectoryIfMissing )
 import System.FilePath
-  ( (</>), (<.>), takeBaseName )
+  ( (</>), (<.>), takeBaseName, equalFilePath )
 import Data.Time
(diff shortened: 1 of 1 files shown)


ezyang:
Good! If it were me, I'd s/sourceFiles/sourceFiles0/ and s/sourceFiles'/sourceFiles/ (because you never want sourceFiles0) but o/w looks fine.

23Skidoo:
Merged as fb51463008ba0d98b5268f8d15987f1097076eee. Thanks!

Issue #3489, opened 2016-06-09
Add a goal order parameter to the dependency solver


grayjay:
This parameter would allow unit tests to specify the exact order of goals in the solver's tree.  Without it, changes in goal order heuristics could break the backjumping tests that rely on a specific order.  It would also allow us to simplify those tests by specifying the goal order more directly.  This issue came up in https://github.com/haskell/cabal/pull/3422#issuecomment-219205230.

kosmikus:
Yes, I very much support this.

grayjay:
Implemented in #3510.

PR #3517, opened 2016-07-04
Freeze solver goal order in two more unit tests

grayjay:
See #3489 and #3510.

@kosmikus Do you think we should specify the goal order for these two tests?  I wasn't completely sure.


kosmikus:
@grayjay Perhaps both, actually? I think for these situations, we want to make sure we find the right install plan both if the goal order is the fixed one, but also in practice (with whatever goal order the solver actually picks).

grayjay:
I updated it to test both goal orders.

kosmikus:
LGTM.

grayjay:
Thanks!  I'll merge it.

PR #3523, opened 2016-07-06
Add Distribution.Compat.Graph, fixes #3521.

### LINKED ISSUE #3521: Proposed new graph interface (refactor of PackageIndex)
I propose we introduce a general, user-friendly graph type to Cabal (Data.Graph is NOT user-friendly), which supports arbitrary key types (as opposed to PackageIndex, which bakes in UnitId as the key.) This will give us more flexibility to pick different keys for solver plans, install plans, etc.

Here is the proposed module:

```
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE FlexibleContexts #-}
-----------------------------------------------------------------------------
-- |
-- Module      :  Data.Graph.Node
-- Copyright   :  (c) Edward Z. Yang
-- License     :  BSD3
--
-- Maintainer  :  libraries@haskell.org
-- Stability   :  experimental
-- Portability :  portable
--
-- This alternative interface to 'Data.Graph' stores nodes (identified
-- by the 'IsNode' type class) which are associated with a key and
-- record the keys of their neighbors.
--
-- Special attention is paid to graphs which are "broken"; i.e., they
-- have neighbor keys which are not in the map.
--
-----------------------------------------------------------------------------

module Data.Graph.Node (SCC(..), module Data.Graph.Node) where

import Prelude hiding (lookup)
import Data.Graph (SCC(..))
import qualified Data.Graph as G
import Data.Map (Map)
import qualified Data.Map as Map
import Data.Maybe
import qualified Data.Array as Array
import Data.Array ((!))
import Data.List hiding (lookup)
import Data.Ord
import qualified Data.Tree as Tree
import Data.Either

-- | A graph of nodes @a@.  The nodes are expected to have instance
-- of class 'IsNode'.
data Graph a
    = Graph {
        graphMap          :: Map (Key a) a,
        -- Lazily cached graph representation
        graphForward      :: G.Graph,
        graphAdjoint      :: G.Graph,
        graphVertexToNode :: G.Vertex -> a,
        graphKeyToVertex  :: Key a -> Maybe G.Vertex,
        graphBroken       :: [(a, [Key a])]
    }

class Ord (Key a) => IsNode a where
    type Key a :: *
    nodeKey :: a -> Key a
    nodeNeighbors :: a -> [Key a]

fromMap :: IsNode a => Map (Key a) a -> Graph a
fromMap m
    = Graph { graphMap = m
            , graphForward = g
            , graphAdjoint = G.transposeG g
            , graphVertexToNode = vertex_to_node
            , graphKeyToVertex = key_to_vertex
            , graphBroken = broken
            }
  where
    try_key_to_vertex k = maybe (Left k) Right (key_to_vertex k)

    (brokenEdges, edges)
        = unzip
        $ [ partitionEithers (map try_key_to_vertex (nodeNeighbors n))
          | n <- ns ]
    broken = filter (not . null . snd) (zip ns brokenEdges)

    g = Array.listArray bounds edges

    ns              = sortBy (comparing nodeKey) (Map.elems m)
    vertices        = zip (map nodeKey ns) [0..]
    vertex_map      = Map.fromList vertices
    key_to_vertex k = Map.lookup k vertex_map

    vertex_to_node vertex = nodeTable ! vertex

    nodeTable   = Array.listArray bounds ns
    bounds = (0, Map.size m - 1)

fromList :: IsNode a => [a] -> Graph a
fromList ns = fromMap
            . Map.fromList
            . map (\n -> (nodeKey n, n))
            $ ns

-- Map-like operations

-- TODO: More efficient graph rebuilding strategy

insert :: IsNode a => a -> Graph a -> Graph a
insert n g = fromMap (Map.insert (nodeKey n) n (toMap g))

deleteKey :: IsNode a => Key a -> Graph a -> Graph a
deleteKey k g = fromMap (Map.delete k (toMap g))

lookup :: IsNode a => Key a -> Graph a -> Maybe a
lookup k g = Map.lookup k (toMap g)

toList :: Graph a -> [a]
toList g = Map.elems (toMap g)

toMap :: Graph a -> Map (Key a) a
toMap = graphMap

-- Graph-like operations

-- From stronglyConnCompR in Data.Graph
stronglyConnComp :: Graph a -> [SCC a]
stronglyConnComp g = map decode forest
  where
    forest = G.scc (graphForward g)
    decode (Tree.Node v [])
        | mentions_itself v = CyclicSCC  [graphVertexToNode g v]
        | otherwise         = AcyclicSCC (graphVertexToNode g v)
    decode other = CyclicSCC (dec other [])
        where dec (Tree.Node v ts) vs
                = graphVertexToNode g v : foldr dec vs ts
    mentions_itself v = v `elem` (graphForward g ! v)

cycles :: Graph a -> [[a]]
cycles g = [ vs | CyclicSCC vs <- stronglyConnComp g ]

broken :: IsNode a => Graph a -> [(a, [Key a])]
broken g = graphBroken g

closure :: IsNode a => Graph a -> [Key a] -> Maybe (Graph a)
closure g ks = do
    vs <- mapM (graphKeyToVertex g) ks
    return (fromList (decodeVertexForest g (G.dff (graphForward g))))

revClosure :: IsNode a => Graph a -> [Key a] -> Maybe [a]
revClosure g ks = do
    vs <- mapM (graphKeyToVertex g) ks
    return (decodeVertexForest g (G.dff (graphAdjoint g)))

flattenForest :: Tree.Forest a -> [a]
flattenForest = concatMap Tree.flatten

decodeVertexForest :: IsNode a => Graph a -> Tree.Forest G.Vertex -> [a]
decodeVertexForest g = map (graphVertexToNode g) . flattenForest

topSort :: Graph a -> [a]
topSort g = map (graphVertexToNode g) $ G.topSort (graphForward g)

revTopSort :: Graph a -> [a]
revTopSort g = map (graphVertexToNode g) $ G.topSort (graphAdjoint g)
```

CC @dcoutts

ezyang:
The intent is for this to be used to replace most of
the current PackageIndex and InstallPlan implementations.

CC @dcoutts @23Skidoo

### DIFF
--- a/Cabal/Distribution/Compat/Graph.hs
+++ b/Cabal/Distribution/Compat/Graph.hs
@@ -1,0 +1,370 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE BangPatterns #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Compat.Graph
+-- Copyright   :  (c) Edward Z. Yang 2016
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-dev@haskell.org
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- A data type representing directed graphs, backed by "Data.Graph".
+-- It is strict in the node type.
+--
+-- This is an alternative interface to "Data.Graph".  In this interface,
+-- nodes (identified by the 'IsNode' type class) are associated with a
+-- key and record the keys of their neighbors.  This interface is more
+-- convenient than 'Data.Graph.Graph', which requires vertices to be
+-- explicitly handled by integer indexes.
+--
+-- The current implementation has somewhat peculiar performance
+-- characteristics.  The asymptotics of all map-like operations mirror
+-- their counterparts in "Data.Map".  However, to perform a graph
+-- operation, we first must build the "Data.Graph" representation, an
+-- operation that takes /O(V + E log V)/.  However, this operation can
+-- be amortized across all queries on that particular graph.
+--
+-- Some nodes may be broken, i.e., refer to neighbors which are not
+-- stored in the graph.  In our graph algorithms, we transparently
+-- ignore such edges; however, you can easily query for the broken
+-- vertices of a graph using 'broken' (and should, e.g., to ensure that
+-- a closure of a graph is well-formed.)  It's possible to take a closed
+-- subset of a broken graph and get a well-formed graph.
+--
+-----------------------------------------------------------------------------
+
+module Distribution.Compat.Graph (
+    -- * Graph type
+    Graph,
+    IsNode(..),
+    -- * Query
+    null,
+    size,
+    lookup,
+    -- * Construction
+    empty,
+    insert,
+    deleteKey,
+    updateLookup,
+    -- * Combine
+    unionLeft,
+    unionRight,
+    -- * Graph algorithms
+    stronglyConnComp,
+    SCC(..),
+    cycles,
+    broken,
+    closure,
+    revClosure,
+    topSort,
+    revTopSort,
+    -- * Conversions
+    -- ** Maps
+    toMap,
+    -- ** Lists
+    fromList,
+    toList,
+    keys,
+    -- ** Graphs
+    toGraph,
+    -- * Node type
+    Node(..),
+    nodeValue,
+) where
+
+import qualified Prelude as Prelude
+import Prelude hiding (lookup, null)
+import Data.Graph (SCC(..))
+import qualified Data.Graph as G
+import Data.Map (Map)
+import qualified Data.Map as Map
+import qualified Data.Array as Array
+import Data.Array ((!))
+import qualified Data.Tree as Tree
+import Data.Either
+import Data.Typeable
+import qualified Data.Foldable as Foldable
+import Control.DeepSeq
+import Distribution.Compat.Binary (Binary(..))
+
+-- | A graph of nodes @a@.  The nodes are expected to have instance
+-- of class 'IsNode'.
+data Graph a
+    = Graph {
+        graphMap          :: !(Map (Key a) a),
+        -- Lazily cached graph representation
+        graphForward      :: G.Graph,
+        graphAdjoint      :: G.Graph,
+        graphVertexToNode :: G.Vertex -> a,
+        graphKeyToVertex  :: Key a -> Maybe G.Vertex,
+        graphBroken       :: [(a, [Key a])]
+    }
+    deriving (Typeable)
+
+-- NB: Not a Functor! (or Traversable), because you need
+-- to restrict Key a ~ Key b.  We provide our own mapping
+-- functions.
+
+-- General strategy is most operations are deferred to the
+-- Map representation.
+
+instance Show a => Show (Graph a) where
+    show = show . toList
+
+instance (IsNode a, Read a) => Read (Graph a) where
+    readsPrec d s = map (\(a,r) -> (fromList a, r)) (readsPrec d s)
+
+instance (IsNode a, Binary a) => Binary (Graph a) where
+    put x = put (toList x)
+    get = fmap fromList get
+
+instance (Eq (Key a), Eq a) => Eq (Graph a) where
+    g1 == g2 = graphMap g1 == graphMap g2
+
+instance Foldable.Foldable Graph where
+    fold = Foldable.fold . graphMap
+    foldr f z = Foldable.foldr f z . graphMap
+    foldl f z = Foldable.foldl f z . graphMap
+    foldMap f = Foldable.foldMap f . graphMap
+#ifdef MIN_VERSION_base
+#if MIN_VERSION_base(4,6,0)
+    foldl' f z = Foldable.foldl' f z . graphMap
+    foldr' f z = Foldable.foldr' f z . graphMap
+#endif
+#if MIN_VERSION_base(4,8,0)
+    length = Foldable.length . graphMap
+    null   = Foldable.null   . graphMap
+    toList = Foldable.toList . graphMap
+    elem x = Foldable.elem x . graphMap
+    maximum = Foldable.maximum . graphMap
+    minimum = Foldable.minimum . graphMap
+    sum     = Foldable.sum     . graphMap
+    product = Foldable.product . graphMap
+#endif
+#endif
+
+instance (NFData a, NFData (Key a)) => NFData (Graph a) where
+    rnf Graph {
+        graphMap = m,
+        graphForward = gf,
+        graphAdjoint = ga,
+        graphVertexToNode = vtn,
+        graphKeyToVertex = ktv,
+        graphBroken = b
+    } = gf `seq` ga `seq` vtn `seq` ktv `seq` b `seq` rnf m
+
+-- TODO: Data instance?
+
+-- | The 'IsNode' class is used for datatypes which represent directed
+-- graph nodes.  A node of type @a@ is associated with some unique key of
+-- type @'Key' a@; given a node we can determine its key ('nodeKey')
+-- and the keys of its neighbors ('nodeNeighbors').
+class Ord (Key a) => IsNode a where
+    type Key a :: *
+    nodeKey :: a -> Key a
+    nodeNeighbors :: a -> [Key a]
+
+-- | A simple, trivial data type which admits an 'IsNode' instance.
+data Node k a = N a k [k]
+    deriving (Show, Eq)
+
+-- | Get the value from a 'Node'.
+nodeValue :: Node k a -> a
+nodeValue (N a _ _) = a
+
+instance Functor (Node k) where
+    fmap f (N a k ks) = N (f a) k ks
+
+instance Ord k => IsNode (Node k a) where
+    type Key (Node k a) = k
+    nodeKey (N _ k _) = k
+    nodeNeighbors (N _ _ ks) = ks
+
+-- TODO: Maybe introduce a typeclass for items which just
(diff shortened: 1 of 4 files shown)


23Skidoo:
Fails on AppVeyor due to `-Werror`.

ezyang:
Huh, I wonder why that didn't light up Travis.

Cabal/tests/UnitTests/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,88 @@
...
+arbitraryGraph len = do
+    -- Careful! Assume k is much larger than size.
+    ks <- vectorOf len arbitrary `suchThat` hasNoDups
+    ns <- forM ks $ \k -> do
+        a <- arbitrary
+        neighbors <- listOf (elements ks)
+        return (N a k (ordNub neighbors))
+    return (fromList ns)
+
+ordNub :: Ord a => [a] -> [a]
```
23Skidoo: You can use the one in `D.S.Utils`.

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,327 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE UndecidableInstances #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Compat.Graph
+-- Copyright   :  (c) Edward Z. Yang 2016
+-- License     :  BSD3
+--
+-- Maintainer  :  libraries@haskell.org
```
23Skidoo: I think that this should be `cabal-dev@`.

ezyang:
Addressed comments and force-pushed. Also a few more functions for interface.

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,358 @@
...
+
+instance Functor (Node k) where
+    fmap f (N a k ks) = N (f a) k ks
+
+instance Ord k => IsNode (Node k a) where
+    type Key (Node k a) = k
+    nodeKey (N _ k _) = k
+    nodeNeighbors (N _ _ ks) = ks
+
+-- TODO: Maybe split this type class into two
```
phadej: which type, into what?

ezyang: The thought is that IsNode can be decomposed into two typeclasses: one with Key and nodeKey (representing data types that have "keys") and the one that adds nodeNeighbors. The former typeclass is independently useful; e.g., you can provide an alternate Map with insert :: HasKey a => a -> Map a -> Map a (no need to specify key; it's embedded in the item.) But it's not useful for what I needed this for...

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,358 @@
...
+    -- * Node type
+    Node(..),
+    nodeValue,
+) where
+
+import qualified Prelude as Prelude
+import Prelude hiding (lookup, null)
+import Data.Graph (SCC(..))
+import qualified Data.Graph as G
+import Data.Map.Strict (Map)
```
phadej: `containers >= 0.5` ?

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,358 @@
...
+-- value if updated, and the original value if it was deleted.
+updateLookup :: IsNode a => (a -> Maybe a) -> Key a -> Graph a -> (Maybe a, Graph a)
+updateLookup f k g =
+    let (r, m') = Map.updateLookupWithKey (const f) k (toMap g)
+    in (r, fromMap m')
+
+-- Combining
+
+-- | /O(V + V')/. Union with a combining function.
+-- The combining function @f@ must be key-preserving; i.e.,
```
phadej: This feels fragile, should there be `safe` and `unsafe` versions. Or at least `valid` function to validate that `Map` invariants are correct?

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,358 @@
...
+
+-- | Reverse topologically sort the nodes of a graph.
+-- Requires amortized construction of graph.
+revTopSort :: Graph a -> [a]
+revTopSort g = map (graphVertexToNode g) $ G.topSort (graphAdjoint g)
+
+-- Conversions
+
+-- | /O(1)/. Convert a map from keys to nodes into a graph.
+fromMap :: IsNode a => Map (Key a) a -> Graph a
```
phadej: there should be a note about `Map` invariant.

phadej: Or even having `unsafeFromMap :: IsNode a => Map (Key a) a -> Graph a` and `fromMap :: Map (Key a) a -> Maybe (Graph a)`

Cabal/tests/UnitTests/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,88 @@
...
+prop_nodes_consistent :: (Eq a, IsNode a) => Graph a -> Bool
+prop_nodes_consistent g = all p (toList g)
+  where
+    (_, vtn, ktv) = toGraph g
+    p n = case ktv (nodeKey n) of
+            Just v  -> vtn v == n
+            Nothing -> False
+
+-- A non-broken graph has the 'nodeNeighbors' of each node
+-- equal the recorded adjacent edges in the node graph.
```
phadej: what if `nodeNeighbors ` return duplicates?

Cabal/Distribution/Compat/Graph.hs
ezyang: Yeah, we can't add that constraint; I fixed it but lost the change. Planning to switch it to Data.Map.

Cabal/Distribution/Compat/Graph.hs
ezyang: I think what I'll probably do is axe this function and just add a right-biased union, which is what we actually seem to need.

Cabal/Distribution/Compat/Graph.hs
ezyang: Yeah, you are right.

Cabal/tests/UnitTests/Distribution/Compat/Graph.hs
ezyang: Data.Graph stores adjacency list, so duplicates should be preserved. Although I guess the current code doesn't test it,

Cabal/Distribution/Compat/Graph.hs
ezyang: I think the "safe" version is unproductive; they should just call Graph.fromList (Map.elems m). I'll add a note about the invariant.

Cabal/Distribution/Compat/Graph.hs
phadej: I'd rather use `CPP`. It's unfortunate if we run into space leaks with lazy `Map`, but still better not to run into them with newer `transformers`.

Cabal/Distribution/Compat/Graph.hs
phadej: I'm not sure whether `fromMap` has to exported at all? Do you see where you need it?

ezyang: You are right, it is not used. I will remove its export.

Cabal/Distribution/Compat/Graph.hs
ezyang: Space leaks here are fairly unlikely, because we have to force @nodeKey n@ to, e.g., insert. I think space leaking with old transformers would be extra-bad; better to just add enough strictness annotations here to avoid problems.

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,370 @@
...
+insert !n g = fromMap (Map.insert (nodeKey n) n (toMap g))
+
+-- | /O(log V)/. Delete the node at a key from the graph.
+deleteKey :: IsNode a => Key a -> Graph a -> Graph a
+deleteKey k g = fromMap (Map.delete k (toMap g))
+
+-- | /O(log V)/. Lookup and update.  This function returns the changed
+-- value if updated, and the original value if it was deleted.
+updateLookup :: IsNode a => (a -> Maybe a) -> Key a -> Graph a -> (Maybe a, Graph a)
+updateLookup f k g =
```
phadej: should `f` preserve the `Key` of the argument, it it returns `Just` ?

Cabal/Distribution/Compat/Graph.hs
phadej: functions like `unionWith` could introduce the leaks. For example: https://neilmitchell.blogspot.fi/2016_05_01_archive.html

But I see you removed `unionWith`. `updateLookupWIthKey` could introduce space leak too

Cabal/Distribution/Compat/Graph.hs
ezyang: Actually I don't need the full generality; I'm going to replace it with deleteLookup.

Cabal/Distribution/Compat/Graph.hs
ezyang: Fascinatingly, you can avoid space leak with updateLookupWithKey with an appropriate seq (since the inner implementation needs to see if it's a Nothing or Just). But I'm replacing it with deleteLookup so the point's moot. Still not going to add a Distribution.Compat.StrictMap ;) (yes I'm lazy.)

Cabal/Distribution/Compat/Graph.hs
```diff
@@ -0,0 +1,370 @@
...
+
+instance Functor (Node k) where
+    fmap f (N a k ks) = N (f a) k ks
+
+instance Ord k => IsNode (Node k a) where
+    type Key (Node k a) = k
+    nodeKey (N _ k _) = k
+    nodeNeighbors (N _ _ ks) = ks
+
+-- TODO: Maybe introduce a typeclass for items which just
```
23Skidoo: s/which/with/

ezyang:
Typofixed, and merged.

PR #3525, opened 2016-07-06
Refactor of PackageIndex and InstallPlan 

ezyang:
The point of this PR is to (1) split InstallPlan into two separate, decoupled data types, InstallPlan (for installing things) and SolverInstallPlan (the output of the solver, which we run consistency checks on) and (2) switch PackageIndex, InstallPlan, and SolverInstallPlan to use the new Graph data type introduced in #3523. Yes there are more lines of code, but they are less coupled and can support future evolution.

CC @dcoutts, @23Skidoo


23Skidoo:
Haven't reviewed in detail, but looks promising to me.

ezyang:
I realized this changeset has one BC-breaking change: the `PackageIndex` type constructor got removed. I suppose there isn't any good reason why we should remove it, so I suppose I should bring it back.

ezyang:
Ugh, Mac OS X GHC 8.0 build is timing out.

ezyang:
Merged.

phadej:
It might be only me, but I'd realy like a bit more tidied up history.

```
- Swizzle
- travis fixes
- Minor fixes.
- I think this should fix all the bugs
- Try this
- Fix up
- Hack
- Hmmmm
```

I'm very sorry, but that doesn't look very professional. Consider using `--fixup`: http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

ezyang:
Ugh, this is the "GitHub is bad at noticing what you actually pushed to master" bug (aka I forgot to push to the topic branch after rebasing.) If you look at the integrated history in master these commits are all gone. This branch was somewhat of a strange outlier because I was debugging Travis issues and needed to keep pushing to CI.

PR #3527, opened 2016-07-08
Add test build directory to LD_LIBRARY_PATH when -executable-dynamic

### LINKED ISSUE #2039: "ghc: panic! (the 'impossible' happened)" using "--enable-executable-dynamic"
From: https://github.com/coreyoconnor/vty/issues/54

Using `--enable-executable-dynamic` leads to a ghc panic. Filing here as Cabal handles `--enable-executable-dynamic`. If this is confirmed to not be caused by Cabal (likely so IMO) then I'll move to ghc's bug tracker.

To reproduce:

1. clone https://github.com/coreyoconnor/executable-dynamic-issue
2. `cabal configure --enable-tests --enable-executable-dynamic`
3. `cabal test`

Reproduced on GHC 7.8.2, GHC 7.8.3, Cabal 1.18, Cabal 1.20. Ubuntu. The vty report is on Nix, which I do not have a working environment to reproduce with.

thomie:
Since commit 18fcd9c1aaeddd9d10a25e44c0e986c9889f06a7, Cabal builds the stub executable and the test library for a LibV09 (detailed-0.9) test in different directories.

This is good, but when running the test with --enable-executable-dynamic, we now have to add the test build directory to (DY)LD_LIBRARY_PATH, so that the linker can find the test library when running the stub executable (see #2289 for details).

This fixes #2039 (read: no parse). The GHC panic mentioned in that ticket must have been resolved earlier already, perhaps also by commit 18fcd9c1aaeddd9d10a25e44c0e986c9889f06a7.

CC @ezyang, @christiaanb

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -54,6 +54,10 @@ module Distribution.Simple (
   ) where
 
 -- local
+import qualified Distribution.InstalledPackageInfo as Installed
+import qualified Distribution.PackageDescription as PD
+import qualified Distribution.Simple.PackageIndex as PackageIndex
+
 import Distribution.Simple.Compiler hiding (Flag)
 import Distribution.Simple.UserHooks
 import Distribution.Package
@@ -88,16 +92,20 @@ import Distribution.Text
 
 -- Base
 import System.Environment (getArgs, getProgName)
-import System.Directory   (removeFile, doesFileExist
+import System.Directory   (removeFile, doesFileExist, getTemporaryDirectory
                           ,doesDirectoryExist, removeDirectoryRecursive)
 import System.Exit                          (exitWith,ExitCode(..))
-import System.FilePath                      (searchPathSeparator)
+import System.FilePath                      (searchPathSeparator, (</>))
+import System.IO                            (hPutStrLn, hClose)
 import Distribution.Compat.Environment      (getEnvironment)
+import Distribution.Compat.Exception        (catchExit, catchIO)
 import Distribution.Compat.GetShortPathName (getShortPathName)
 
-import Control.Monad   (when)
+import Control.Monad   (when, filterM)
 import Data.Foldable   (traverse_)
-import Data.List       (unionBy)
+import Data.List       (unionBy, inits)
+import Data.Maybe      (isNothing)
+
 
 -- | A simple implementation of @main@ for a Cabal setup script.
 -- It reads the package description file using IO, and performs the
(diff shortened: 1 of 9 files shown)


Cabal/Distribution/Simple/Test/LibV09.hs
```diff
@@ -99,8 +99,15 @@ runTest pkg_descr lbi flags suite = do
...
+                                      (_, _, _, _, _, _, exeClbi) =
+                                          testSuiteLibV09AsLibAndExe
+                                              pkg_descr suite clbi lbi
+                                              distPref pwd
+                                  -- Get the dependencies of the stub
+                                  -- executable. They should include the test
+                                  -- library itself (#2039).
                                   paths <- LBI.depLibraryPaths
-                                             True False lbi clbi
+                                             True False lbi exeClbi
```
ezyang: This needs a comment explaining why it's `exeClbi` and not `clbi`.

thomie: I added the following comment already:
```
-- Get the dependencies of the stub
-- executable. They should include the test
-- library itself (#2039).
```

If you still think it isn't clear, could you suggest what the comment should be?

ezyang: I investigated and here is what I discovered:

1. Here is something confusing: the function is called `depLibraryPaths` (library?!), yet we are passing it the CLBI for the executable! I looked at the definition and the point is that depLibraryPaths doesn't have anything to do about libraries; it has to do with the DEPENDENT libraries, and whose dependencies do we care about? The executables. So I guess I'd personally be a fan of a ``-- NB: the library path is for the dependencies of the *executable*, not library`` (or if you want to shave yaks, renaming `depLibraryPaths`)
2. I don't understand why this code has to go through all these gyrations to even get its hands on the executable CLBI. This seems wrong-headed. The real problem is that the CLBIs for this type of test is a massive hack, as explained in Build:

```
               clbi -- This ComponentLocalBuildInfo corresponds to a detailed
                    -- test suite and not a real component. It should not
                    -- be used, except to construct the CLBIs for the
                    -- library and stub executable that will actually be
                    -- built.
```

   A variant of this comment seems apt here. If this were GHC I'd tell you to refer to the Note.

Does that help?

Cabal/Distribution/Simple.hs
```diff
@@ -522,6 +530,165 @@ simpleUserHooks =
       where
         verbosity = fromFlag (configVerbosity flags)
 
+-- Try to build a test C program which includes every header and links every
+-- lib. If that fails, try to narrow it down by preprocessing (only) and linking
+-- with individual headers and libs.  If none is the obvious culprit then give a
+-- generic error message.
+-- TODO: produce a log file from the compiler errors, if any.
+checkForeignDeps :: PackageDescription -> LocalBuildInfo -> Verbosity -> IO ()
```
23Skidoo: Why do you need to move `checkForeignDeps`?

23Skidoo:
LGTM, I think this can go in once @ezyang's comments are addressed.

Mikolaj:
@thomie: hi! it's been a while and the code must have bitrotted, but AFAICT it's still an important issue to solve, so would you like to work on it and merge it?

thomie:
@Mikolaj: sorry, I will not work on this anymore.

Mikolaj:
@thomie: that's fine, thank you for your contributions and all the best!

Issue #2899, opened 2015-10-28
cabal uses too much memory when solving for setup dependencies


grayjay:
Some combinations of setup and library dependencies cause the solver to use more and more memory.  I ran `cabal install --dry-run --max-backjumps -1` with this cabal file for five minutes with about 16GB of memory.  I used GHC 7.10.2.

```
name:                example
version:             0.1.0.0
cabal-version:       >=1.8
build-type:          Custom

custom-setup
  setup-depends:       aeson, hackage-server, yi

library
  build-depends:       aeson, mtl, leksah
```

![cabal-memory-usage](https://cloud.githubusercontent.com/assets/4276753/10780987/483a8bba-7d00-11e5-8734-47b18ef62a04.png)

My best guess is that it is related to the `addLinking` function creating linked nodes from existing unlinked nodes.  When a node is referenced by a later sibling, it probably isn't garbage collected immediately after it is traversed. https://github.com/haskell/cabal/blob/c5b90c4d157f64156b6c8efde2aafb1ac86a851b/cabal-install/Distribution/Client/Dependency/Modular/Linking.hs#L67

grayjay:
I encountered this issue again while testing with a long-running command: `cabal install leksah yesod hackage-server`

This uses GHC 8.0.1.
![issue-2899](https://cloud.githubusercontent.com/assets/4276753/16710313/c72c463c-45de-11e6-88f8-a76b40c555f5.png)

The -v3 log shows that the solver chose to link a setup dependency near the top of the tree (level 8).  It didn't return to that level until around the time that the memory usage dropped.  When it descended again, it used a different goal order and chose the setup dependencies much farther down in the tree.  I think that the spike in memory usage can be explained by the solver retaining the whole subtree starting where it chose the first setup dependency.

grayjay:
Fixed in #4110.

PR #3530, opened 2016-07-10
Solver: Fix space leak in 'addlinking' (issue #2899)

grayjay:
I took a pass at fixing #2899.  I created a tree type with subtrees that have dummy arguments, so that `addLinking` can copy the subtrees without causing them to share data.  I'm not sure if this is the best approach, though.  Another option would be to add the linked nodes during the "build" phase.

This isn't ready to be merged yet.  I at least need to document and rename the tree type.

/cc @kosmikus @edsko

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -102,17 +102,18 @@ data BuildType =
   | Instance QPN I PInfo QGoalReason  -- ^ build a tree for a concrete instance
   deriving Show
 
-build :: BuildState -> Tree QGoalReason
-build = ana go
+build :: BuildState -> Tree' QGoalReason
+build = ana' go
   where
-    go :: BuildState -> TreeF QGoalReason BuildState
+    go :: BuildState -> TreeF QGoalReason (() -> BuildState)
 
     -- If we have a choice between many goals, we just record the choice in
     -- the tree. We select each open goal in turn, and before we descend, remove
     -- it from the queue of open goals.
     go bs@(BS { rdeps = rds, open = gs, next = Goals })
       | P.null gs = DoneF rds
-      | otherwise = GoalChoiceF $ P.mapKeys close
+      | otherwise = GoalChoiceF $ addDummyArgs
+                                $ P.mapKeys close
                                 $ P.mapWithKey (\ g (_sc, gs') -> bs { next = OneGoal g, open = gs' })
                                 $ P.splits gs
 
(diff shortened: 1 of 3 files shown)


ezyang:
Unfortunately, the dummy argument method is not guaranteed to thwart sharing, because inlining and floating can reintroduce the sharing. For example, try compiling:

```
module F where
g =
    let f () = [1..]
    in (f (), f ())
```

You might be OK in this case because your functions are big enough that not much inlining is present, but in cases like this it's very tricky.

As far as I can tell, the only way to guarantee sharing (or lack thereof) is to work in the IO monad. But that is probably a bigger refactoring than you are willing to bear.

In any case, this sort of trick deserves a big honking comment. Maybe do a GHC style Note.

phadej:
One way to prevent sharing is to use church encoding, i.e. not carry a value, but a function to consume that value (e.g. `\c n -> foldr cn xs`).

```hs
data ListF a b
    = Nil
    | Cons a b

foldr :: (a -> b -> b) -> b -> [a] -> b
cata :: (ListF a b -> b) -> [a] -> b
```

i.e. `Tree a ~ forall b. TreeF a b -> b`, but even then we have to sure to produce RHS variant in the way, so the big closure (instead of big value) isn't persisted,

I'm exploring this way of sharing-prevention for `universe` package. I could try to prioritise that experimentation, if this is urgent problem. https://github.com/dmwit/universe/issues/23

edsko:
Hmmm, @kosmikus has expressed an preference for adding the link options as part of the build phase ; if that fixes this problem, perhaps that's a nicer solution rather than trying clever tricks to prevent sharing?

grayjay:
Thank you for the feedback, and thank you @phadej for offering to look into preventing sharing.  I didn't realize that the dummy argument trick was unreliable.  Combining the build and linking phases should work, because the linked and unlinked subtrees will differ from the start.  I'll try that solution when I have more time.

grayjay:
I just realized that the linked and unlinked subtrees are still identical after the linking traversal, because linking at one level doesn't affect how nodes are linked farther down in the tree.  So combining the build and linking phases may not be enough to guarantee that the subtrees are not shared.  I don't think it would even reduce the likelihood of a space leak over this PR.

Currently, `buildTree` creates a search tree with only unlinked choices for each package.  Then `addLinking` traverses the tree and adds linking choices under some of those packages by transforming the existing unlinked choices.  `addLinking` adds a choice to link each package to each other instance of that package that has already been chosen, whether or not that choice is consistent with the choices that were made higher up in the tree.  Since linking has no effect on the linking choices lower in the tree, the subtrees under linking choices are equal to the subtrees under the unlinked choices that they were created from.  E.g., the tree under the choice to link package C to A-setup.C-1 is equal to the tree under the choice to link to B-setup.C-1, or the choice to not link at all.

AFAICT, the subtrees only differ after the `validateLinking` step, when the solver finally prunes nodes whose linking is inconsistent with the choices that were made above.  If we combined all steps up to `validateLinking`, then the subtrees of the initial tree would all be different.  I hope that isn't necessary, though, because it would be significantly less modular.

That said, I would be surprised if the subtrees were shared if we only combined `buildTree` and `addLinking`.  The logic that leads to identical subtrees isn't straightforward.

I'm not really sure what to do.  We could make the subtrees differ in some way, such as only creating one new linking choice for each distinct group of linked packages.  That also seems like a trick, though, even if it has its own benefits.

grayjay:
I created a test case with the solver DSL: https://github.com/grayjay/cabal/commit/1f1b95d7b4d836ac79a60f943bb36e0ac6ef41e8

PR #3532, opened 2016-07-11
Correctly pass setup options to sdist invocation.

ezyang:
This fixes a bug I noticed when converting cabal-install to use convenience libraries.


23Skidoo:
Shouldn't the result of `sdist --list-sources` always be the same and independent of flags?

Otherwise LGTM.

ezyang:
No, the difference was that if it's a Custom setup, then you need to pass the correct flags to build the setup script.

ezyang:
Merged.

23Skidoo:
So this is required e.g. when the setup script depends on a library that is only installed in a sandbox?

ezyang:
Hmm, I guess so, in which case, there should be a test.

ezyang:
After shooting myself in the foot I have a test case. I'll push it tomorrow.

ezyang:
#3536

PR #3533, opened 2016-07-11
cabal-install uses internal libraries.

ezyang:
There are lots of benefits:

    * We only build solver code once, instead of repeatedly
      rebuilding it for each test suite.

    * Duplication in the cabal-install file is greatly
      reduced.

There is one downside, which is that we don't rebuild
cabal-install with asserts turned on.  I think we should
look into making this a runtime toggleable chunk of code.


23Skidoo:
LGTM, but build bot failures will need to be fixed.

>  +36 −122

Nice!

BardurArantsson:
So I'm guessing that this means (assuming it's merged) that we could also do so for the solver-only bits? (That is, the `Distribution.Solver.*` modules.)

(I'd be happy to do a PR for that, time permitting.)

23Skidoo:
> So I'm guessing that this means (assuming it's merged) that we could also do so for the solver-only bits? (That is, the Distribution.Solver.* modules.)

Yes.

> (I'd be happy to do a PR for that, time permitting.)

Please do!

ezyang:
I guess even when I get this green, we can't do this for 1.26, because the need for internal library support means that you can't "cabal install cabal-install" from 1.24 (which doesn't understand internal libraries.) So we need one intervening release before we can put this in.

BardurArantsson:
Ah, true. Guess it'll have to wait for post-1.26.

23Skidoo:
Well, in principle we can release cabal-install 1.26 faster than originally planned (e.g. in September).

phadej:
Won't this need a `Cabal-1.26` too?

23Skidoo:
Yes, but I don't see this as a problem.

ezyang:
Well, I'm going to close this for now, no way to merge.

23Skidoo:
We should revive this after the next release.

PR #3536, opened 2016-07-11
Add test for #3532.

ezyang:


ezyang:
Just my luck. I add a test and uncover another bug -_-

ezyang:
Closing because I don't want to deal with this bug at the moment.

PR #3537, opened 2016-07-11
Import hygiene and miscellaneous fixes.

ezyang:

### DIFF
--- a/Cabal/Distribution/Compat/Graph.hs
+++ b/Cabal/Distribution/Compat/Graph.hs
@@ -87,10 +87,10 @@ import qualified Data.Map as Map
 import qualified Data.Array as Array
 import Data.Array ((!))
 import qualified Data.Tree as Tree
-import Data.Either
-import Data.Typeable
+import Data.Either (partitionEithers)
+import Data.Typeable (Typeable)
 import qualified Data.Foldable as Foldable
-import Control.DeepSeq
+import Control.DeepSeq (NFData(..))
 import Distribution.Compat.Binary (Binary(..))
 
 -- | A graph of nodes @a@.  The nodes are expected to have instance
@@ -186,7 +186,7 @@ instance Ord k => IsNode (Node k a) where
     nodeKey (N _ k _) = k
     nodeNeighbors (N _ _ ks) = ks
 
--- TODO: Maybe introduce a typeclass for items with just
+-- TODO: Maybe introduce a typeclass for items which just
 -- keys (so, Key associated type, and nodeKey method).  But
 -- I didn't need it here, so I didn't introduce it.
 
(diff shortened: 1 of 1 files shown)


ezyang:
@phadej 

* The revert was because I botched the push (I accidentally also grabbed some local changes for a non-ready patch which broke the build.)
* I can start using GitHub merge (don't like it but for each project its own, I suppose.)
* Re #3476 that was an oversight, I've pushed the updated comment.
* Yes, I don't know if we expect @23Skidoo to merge all pulls to master (bus factor?), or if developers judge that their PR is stabilized they can merge themselves. Every bit helps—we're not very good at shepherding PRs to completion.

phadej:
This is precisely why we use Github merge: Travis the CI verifies that we don't have something which breaks the build in the commit / branch you push / merge. It's not perfect, but still better than (even rare) accidental breakages.

I'm also quite sure that people are lazy to verify with every GHC we tend to support, that boring job is left for CI on purpose.

ezyang:
(FYI: I edited my above comment, in case you didn't see.)

23Skidoo:
I agree with @ezyang's points and appreciate his efforts. In general, I think we should move in the direction of less bureaucracy, which among other things means trusting developers with a commit bit to merge open PRs (after all, that's what the commit bit is for).

I also think that non-GitHub merges are fine, since our build bots will catch any eventual errors anyway, in which case the offending commits can be reverted.

23Skidoo:
Merged, thanks!

phadej:
> And so the question I would ask you is this: should Cabal value stability or forward progress more?

This is a hard question. I do value forward progress, do value your efforts, and I have bitten myself by long-lead time and numerous rebasing due conflicts (not in `Cabal` though). But I do value stability too.

To return to #3525. I do agree that it's welcomed change. But I'm a bit confused. There is an issue to introduce `Cabal.Distribution.Compat.Graph` #3521, referencing #3525. Yet, I cannot find any discussion about either change. I think that I follow all media where discussion could have happen, but apparently I don't. Hopefully there was some discussion somewhere, but I totally missed it. 

About large refactorings. I have no silver-bullet for this. Try to split them into smaller PRs (as you did with `Graph`), squash / fixup history so there aren't numerous *fix small thing* -commits: you'll help both the reviewer and yourself, if you need to rebase. For example I'd try to split #3525, into a separate PR for rewrite of `PackageIndex`, even that would require redundant fixing of `cabal-install` in between.

I do dislike that we step over CI, and push to `master` directly. I *greatly* appreciate, that if I have a bit a time, and I want use it to make `Cabal` better, I can pull newest `master`, and I have the repository in good shape to start with. Travis builds don't take *that long*, you cannot wait for them.

If I can propose a rule: let someone else than the author merge the PR.

PR #3542, opened 2016-07-12
Introduce new ComponentEnabledSpec, removing testEnabled/benchmarkEna…

ezyang:
…bled.

As per an existing TODO in the code, the use of
testEnabled/benchmarkEnabled to indicate if a component
was enabled/disabled by the user command line was an
egregious violation of abstraction.  This commit removes
these two fields, instead passing along the necessary
enabling information with ComponentEnabledSpec instead.

As there were not many uses of testEnabled/benchmarkEnabled,
this was not too difficult to do.


Cabal/Distribution/PackageDescription.hs
```diff
@@ -631,15 +619,6 @@ emptyTestSuite = mempty
 hasTests :: PackageDescription -> Bool
 hasTests = any (buildable . testBuildInfo) . testSuites
 
--- | Get all the enabled test suites from a package.
-enabledTests :: PackageDescription -> [TestSuite]
-enabledTests = filter testEnabled . testSuites
-
--- | Perform an action on each buildable 'TestSuite' in a package.
-withTest :: PackageDescription -> (TestSuite -> IO ()) -> IO ()
```
phadej: Are this removed because we favour `withTestLBI`?

Cabal/Distribution/Simple/LocalBuildInfo.hs
```diff
@@ -152,6 +161,37 @@ data LocalBuildInfo = LocalBuildInfo {
 
 instance Binary LocalBuildInfo
 
+-- | Describes what components are enabled by user-interaction.
+data ComponentEnabledSpec
```
phadej: should this have `@since 1.26.0.0` annotation?

Cabal/Distribution/Simple/LocalBuildInfo.hs
```diff
@@ -152,6 +161,37 @@ data LocalBuildInfo = LocalBuildInfo {
...
+    = ComponentEnabledSpec {
+        testsEnabled :: Bool,
+        benchmarksEnabled :: Bool
```
phadej: will this have e.g. `executablesEnabled` flag. Can it be uses to more precise selection of buildable components (i.e. only single test-suite)? If yes, could we directly have the target `Spec`, even we don't use all features of it?

Cabal/Distribution/PackageDescription/Configuration.hs
```diff
@@ -416,15 +418,15 @@ newtype TargetSet a = TargetSet [(DependencyMap, a)]
...
+overallDependencies enabled (TargetSet targets) = mconcat depss
   where
     (depss, _) = unzip $ filter (removeDisabledSections . snd) targets
     removeDisabledSections :: PDTagged -> Bool
-    removeDisabledSections (Lib _ l)     = buildable (libBuildInfo l)
-    removeDisabledSections (Exe _ e)   = buildable (buildInfo e)
-    removeDisabledSections (Test _ t)  = testEnabled t && buildable (testBuildInfo t)
-    removeDisabledSections (Bench _ b) = benchmarkEnabled b && buildable (benchmarkBuildInfo b)
+    removeDisabledSections (Lib _ l)   = componentEnabled enabled (CLib l)
+    removeDisabledSections (Exe _ e)   = componentEnabled enabled (CExe e)
```
phadej: what did (does) `buildable` check do? I see there is still `componentBuildable`?

Cabal/Distribution/PackageDescription.hs
ezyang: Correct. Unfortunately a backwards compatible interface is not possible because you really do need the CLBI to determine what is enabled.

Cabal/Distribution/Simple/LocalBuildInfo.hs
ezyang: Sure.

Cabal/Distribution/Simple/LocalBuildInfo.hs
ezyang: Yep, that's the idea, and the original reason why I started this refactor!

Cabal/Distribution/PackageDescription/Configuration.hs
ezyang: Buildable merely says whether or not `buildable: true/false` syntactically in the Cabal file. It's a "static" notion of enabling which gets resolved when we pick flags, in contrast to the enable spec which is passed in from command line. I don't know a good place to doc this.

ezyang:
I guess I need to push the change through cabal-install too.

Cabal/Distribution/PackageDescription/Configuration.hs
phadej: To recap:

- `componentEnabled` is whether I asked for component to be built
- `componentBuildable` is whether component is buildable in my environment (i.e. package author)

AFAICS there aren't currently a possiblity to *enable* component which isn't *buildable* (i.e. we say *enable whatever you can *build(), but in the future we might run into that situation?

If this were GHC, I'd add a `[Note]` and refer to it from both `componentEnabled` and `componentBuildable` docs.

*EDIT:* `componentBuildable` could depend only on the environmen (`os`, `impl` etc)?

ezyang:
Hmm, I just realized that if I remove 'withTest' and 'withBenchmark' then I really ought to remove 'withLib'  and 'withExe'. Maybe it's better to keep them and note that there's been a change in behavior.

phadej:
@ezyang you could `{-# DEPRECATE #-}` them too, say for removal in 1.28?

ezyang:
I decided to put them back in with copious hyperlinks.

ezyang:
OK so I came up with a way to do Notes that get rendered by Haddock (since the info is relevant to clients). Along the way I fixed a lot of Haddock documentation. Take a look.

ezyang:
I guess this PR is blocked on #3545.

ezyang:
Unblocked!

ezyang:
Calling attention to the `finalizePackageDescription` BC change: it takes a new `EnabledComponentSpec` parameter. http://hs01.scs.stanford.edu:8910/search/?q=finalizePackageDescription There seem to be a decent number of packages using it, so I think what I will do is undo this BC-breaking change. The default behavior will be to assume tests and benchmarks are NOT enabled.

And the rest:

* `enabledTests` and `enabledBenchmarks` are gone. Fortunately noone is using them http://hs01.scs.stanford.edu:8910/search/?q=enabledTestsq=finalizePackageDescription In contract `overallDependencies` and `resolveWithFlags` are unchanged.
* `checkBuildTargets` grows a `LocalBuildInfo` but no one is using it. http://hs01.scs.stanford.edu:8910/search/?q=checkBuildTargets
* `componentEnabled` and `componentDisabledReason` have different types. No one is using it http://hs01.scs.stanford.edu:8910/search/?q=componentEnabled 
* `pkgEnabledComponents` is removed, but no one is using it http://hs01.scs.stanford.edu:8910/search/?q=pkgEnabledComponents

After fixing BC, my plan is to push this and the hack for #3545; the yak for that ticket can be shaved later.

Cabal/Distribution/PackageDescription.hs
```diff
@@ -765,14 +758,14 @@ emptyBenchmark = mempty
 hasBenchmarks :: PackageDescription -> Bool
 hasBenchmarks = any (buildable . benchmarkBuildInfo) . benchmarks
 
--- | Get all the enabled benchmarks from a package.
-enabledBenchmarks :: PackageDescription -> [Benchmark]
-enabledBenchmarks = filter benchmarkEnabled . benchmarks
-
 -- | Perform an action on each buildable 'Benchmark' in a package.
+-- You probably want 'withBenchLBI' if you have a 'LocalBuildInfo', see the note in
+-- "Distribution.Simple.LocalBuildInfo#buildable_vs_enabled_components"
```
23Skidoo: This note lives in `Cabal/Distribution/Types/ComponentEnabledSpec.hs` now.

23Skidoo: Fixed in 463c9a873104a9e8fc4bcdfae9060219fb9185cd.

Issue #3552, opened 2016-07-14
readDesc hook is still not plumbed sufficiently


ezyang:
Hello @thumphries, you might be interested in this one.

While working on an unrelated refactoring, I realized that the `readDesc` hook is not being plumbed everywhere it needs to be. To make matters worse, I'm not sure what we can do.

The issue is this code in `Cabal/Distribution/PackageDescription/Check.hs`, which checks that the `.cabal` file does not have a byte order mark:

```
checkCabalFileBOM :: Monad m => CheckPackageContentOps m
                  -> m (Maybe PackageCheck)
checkCabalFileBOM ops = do
  epdfile <- findPackageDesc ops
  case epdfile of
    Left pc      -> return $ Just pc
    Right pdfile -> (flip check pc . startsWithBOM . fromUTF8) `liftM` (getFileContents ops pdfile)
                       where pc = PackageDistInexcusable $
                                    pdfile ++ " starts with an Unicode byte order mark (BOM). This may cause problems with older cabal versions."
```

No hooks to be found, so they can't possibly be applied here. To add insult to injury, this function gets invoked when you configure. So whatever you were using `readDesc` for, Cabal is STILL banging on the actual file. (If it even exists!)

@phadej, this is your code. How do you think it should interact with the `readDesc` hook? Should we drop this code? Or maybe add a hook that passes in the unparsed string of the Cabal file?

PR #3553, opened 2016-07-14
Implement --cabal-file, allows multiple Cabal files in directory

ezyang:
This is primarily intended for use with the Cabal test suite (allowing
us to easily specify multiple Cabal packages for the same Haskell source
files), but maybe some end-users will find it useful as well.  If there
are multiple Cabal files in the current working directory, --cabal-file
(for configure) allows you to disambiguate which one to build with.

There's a big hack to handle the BOM check, as it is inconvenient to
plumb the flag value all the way to the check code.  Some bigger
refactoring needed, see #3552.


phadej:
So `cabal check` doesn't support `--cabal-file` atm?

ezyang:
Yes, and supporting it would require quite a bit of faffing about.

23Skidoo:
Bikeshedding: we might want to call it `--package-description` in case we later add support for other package description formats (e.g. hpack).

/cc @dcoutts

ezyang:
OK, sure (although it's hard to imagine the library getting support for hpack; it's more of a cabal-install thing.)

phadej:
I'd keep it as `--cabal-file`, and if we later would like to add `--package-description` we can, yet `--cabal-file` would be always parsed as cabal file (e.g. we could add `--hpack-file` too)

 `--package-description` needs to guess, and sometimes you really want the tool you use not to guess at all, just do what it's told. 

*note to myself:* verify what automatic reconfiguration can be disabled with some flag

ezyang:
Well, I have the patch to put it either way.

hvr:
Please take into account we'll also need a way to specify the `cabal.project` file location when designing a name for this

ezyang:
Well... `--cabal-project` for that one?

hvr:
Alright, we'd have 

 - `--config-file` to specify an alternative `~/.cabal/config` 

 - (suggested) `--cabal-file` to speficy an alternative `*.cabal` file

but then, for consistency, I'd suggest

 - `--project-file` to specify  a non-default `cabal.project` file

23Skidoo:
@ezyang BTW, don't we need a note in the manual about `--cabal-file`?

ezyang:
Well, I wasn't intending for it to be a user-visible feature. If we put it in the manual, we need to teach cabal-install how to handle it too, because cabal-install also reads in cabal files itself.

23Skidoo:
There should be a place for documenting non-user-visible features like `--exact-configuration` and this one, but I guess this will have to wait until someone rewrites the user guide.

23Skidoo:
Added a line to the changelog (858981e7d2eb57148872ad581ed5e70ecd9fe671).

Issue #3550, opened 2016-07-14
Mac OS X continuous integration


ezyang:
According to https://docs.travis-ci.com/user/multi-os/ Travis supports OS X. We should add it to the build matrix.

23Skidoo:
IIRC, Stack's Travis script supports OS X, so we can start by looking there.

phadej:
@23Skidoo they only use `stack` on osx

ezyang:
Once I finish #3525 we will have on-time Mac OS X builds.

ezyang:
Fixed!

Issue #3464, opened 2016-05-28
bootstrap: TMPDIR redefinition


tuncer:
When creating https://github.com/haskell/cabal/commit/ce58148ff6849c5adb56bcf5cfcb92c2263e3092, I was surprised to see that `TMPDIR` is reused for something else than the convention of pointing to a custom `/tmp`.
1. Where is the re-exported `TMPDIR` used? Maybe by cabal-install?
2. It goes against convention to name it TMPDIR because you set TMPDIR to deviate from /tmp, so referencing a directory inside /tmp by the name `$TMPDIR` isn't right.

/cc @23Skidoo.

ezyang:
Hi @tuncer.

This is a textbook case of (1) someone before you not documenting something tricky they did, and (2) if you see something tricky, you better investigate.

I blame'd the Git code, and this line was introduced in #1710 to fix some sort of awful problem on Mavericks. But the fix here wasn't complete and so @23Skidoo wrote another patch to solve @creswick's report at #2542.

As the grep now doesn't pick up ld's, it is possible the patch from #1710 can be deleted. But I would not feel confident about it without checking to see if #2542 is still fixed.

tuncer:
@ezyang, the referenced tickets seem resolved. Does that mean the OS X issues are fixed and, if so, can we get rid of this now? One side effect is that this seems to produce a leftover cabal-XXXXXXX in /tmp.

ezyang:
Erm, do you mean, get rid of the `TMPDIR` setting? I think it's possible but I certainly want to test. Maybe here is a reasonable workplan:

1. Adjust the Mac OS X bootstrap script to take place in a folder named `arnold` (or anything with `ld`) so we keep testing #2542
2. The existing bootstrap script should test #1710. So now just drop the TMPDIR setting and make sure Travis is green, then push.

Does that sound good?

tuncer:
> Erm, do you mean, get rid of the TMPDIR setting? 

Yes, overriding TMPDIR is one thing, but what's going on here is confusing.

> Does that sound good?

Sounds like a workable plan.

tuncer:
@ezyang, should we close this?

ezyang:
Sure.

PR #3554, opened 2016-07-14
Try to install ghc manually on osx

phadej:
Tries to implement https://github.com/haskell/cabal/issues/3550

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,30 @@
 # The following enables several GHC versions to be tested; often it's enough to
 # test only against the last release in a major GHC version. Feel free to omit
 # lines listings versions you don't need/want testing for.
-env:
- - GHCVER=7.4.2
- - GHCVER=7.6.3
- - GHCVER=7.8.4
- - GHCVER=7.10.3
- - GHCVER=8.0.1 TEST_OLDER=NO
+matrix:
+  include:
+   - env: GHCVER=7.4.2
+     os: linux
+   - env: GHCVER=7.6.3
+     os: linux
+   - env: GHCVER=7.8.4
+     os: linux
+   - env: GHCVER=7.10.3
+     os: linux
+   - env: GHCVER=8.0.1 TEST_OLDER=NO
+     os: linux
+   - env: GHCVER=7.8.4
+     os: osx
+   - env: GHCVER=7.10.3
+     os: osx
+   - env: GHCVER=8.0.1 TEST_OLDER=NO
+     os: osx
+  allow_failures:
+   - env: GHCVER=head
+
  # TODO add PARSEC_BUNDLED=YES when it's so
  # It seems pointless to run head if we're going to ignore the results.
  #- GHCVER=head
-
-os:
- - linux
- - osx
 
 # Note: the distinction between `before_install` and `install` is not important.
 before_install:
@@ -40,10 +51,6 @@ script:
 # Deploy Haddocks to the haskell/cabal-website repo.
 after_success:
  - ./travis-deploy.sh deploy
-
-matrix:
-  allow_failures:
-   - env: GHCVER=head
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 # NB: don't set `language: haskell` here
+language: c
+
+sudo: required
 
 # The following enables several GHC versions to be tested; often it's enough to
 # test only against the last release in a major GHC version. Feel free to omit
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 # NB: don't set `language: haskell` here
+# We specify language: c, so it doesn't default to e.g. ruby
 language: c
 
 sudo: required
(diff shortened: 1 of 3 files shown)


phadej:
Well at least it caught, that `mktemp -p` doesn't work on mac, which is in `bootstrap.sh`.

phadej:
In the last commit build failure is interesting:
```
Package Tests
  InternalLibraries/gen-script:                 FAIL
    Exception: ./register.sh: runProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

What might be the cause?

ezyang:
Re mktemp, that is a regression from https://github.com/haskell/cabal/commit/ce58148ff6849c5adb56bcf5cfcb92c2263e3092 see also discussion in #3464. Let's revert that commit.

ezyang:
I wrote that test. `cabal register` generates a `register.sh` and then we try to run it. Unfortunately the test isn't clear about why it failed, but I guess there's some non-portability in the process execution code. Too bad I don't have an OS X laptop handy :(

phadej:
> The job exceeded the maximum time limit for jobs, and has been terminated.

ezyang:
Wait, where does it say that?

phadej:
Ok, it was something with travis. Maybe bad time of day.

There are cpp problems `GHC-7.6.3`, and `GHC-7.4.2`, dunno what to do. The support window allows us to drop those, OTOH if installing `cpphs` and using it would fix the issue, then why not...

ezyang:
So, it's this bug, right? http://justtesting.org/post/64947952690/the-glasgow-haskell-compiler-ghc-on-os-x-109

I'm OK with declaring our official Mac OS X support starts with 7.8.4. Then let's just remove it from the build matrix in that case.

23Skidoo:
> I'm OK with declaring our official Mac OS X support starts with 7.8.4. Then let's just remove it from the build matrix in that case.

+1.

Blaisorblade:
Caveat: 7.8.4 doesn't work on 10.11 without hacks. Thankfully Travis supports OS X 10.9 and 10.10, but might be worth taking note of that.
https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version

According to
https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Adding this to .travis.yml (after line https://github.com/haskell/cabal/pull/3554/files#diff-354f30a63fb0907d4ad57269548329e3R39) should be enough 
```
matrix:
#[...]
  exclude:
  - os: osx
    env: GHCVER=7.4.2
  - os: osx
    env: GHCVER=7.6.3
```

phadej:
@23Skidoo I made improvements to docs uploading
- runs only in single matrix entry
- everything related is done in `after_success`
    - probably still doesn't work, but at least doesn't fail PR CI jobs

phadej:
Green. Please review.

.travis.yml
```diff
@@ -1,26 +1,41 @@
 # NB: don't set `language: haskell` here
+language: c
```
ezyang: Put a comment here why we want language: c

.travis.yml
```diff
@@ -1,26 +1,41 @@
...
+   - env: GHCVER=7.4.2
+     os: linux
+   - env: GHCVER=7.6.3
+     os: linux
+   - env: GHCVER=7.8.4
+     os: linux
+   - env: GHCVER=7.10.3
+     os: linux
+   - env: GHCVER=8.0.1 TEST_OLDER=NO
+     os: linux
```
ezyang: Comment here (e.g. some text plus ref to PR) why we are not testing 7.6.3

.travis.yml
```diff
@@ -1,26 +1,41 @@
...
 before_install:
- - travis_retry sudo add-apt-repository -y ppa:hvr/ghc
- - travis_retry sudo apt-get update
- - travis_retry sudo apt-get install --force-yes ghc-$GHCVER-prof ghc-$GHCVER-dyn happy
- - if [ "$TEST_OLDER" == "YES" ]; then travis_retry sudo apt-get install --force-yes ghc-7.0.4-prof ghc-7.0.4-dyn ghc-7.2.2-prof ghc-7.2.2-dyn; fi
- - export PATH=$HOME/.cabal/bin:/opt/ghc/$GHCVER/bin:/opt/cabal/1.24/bin:$PATH
- - git version
+ - ./travis-install.sh
+ - export PATH=/opt/ghc/$GHCVER/bin:$PATH
+ - export PATH=$HOME/.ghc-install//bin:$PATH;
```
ezyang: doubleslash?

.travis.yml
```diff
@@ -1,26 +1,41 @@
...
+     os: linux
+   - env: GHCVER=7.10.3
+     os: linux
+   - env: GHCVER=8.0.1 TEST_OLDER=NO
+     os: linux
+   - env: GHCVER=7.8.4
+     os: osx
+   - env: GHCVER=7.10.3
+     os: osx
+   - env: GHCVER=8.0.1 TEST_OLDER=NO
```
ezyang: NB: Although TEST_OLDER is currently disabled because (1dd14ccb91b7719c15c516cea76761bc8a01531e), you didn't implement TEST_OLDER for the Mac OS X installer so you shouldn't imply here that we could turn it on once timeouts are fixed.

ezyang:
I put some comments but if you don't want to do them just go ahead and push.

Issue #3545, opened 2016-07-13
Moving location where library build files are generated


ezyang:
Hey guys. I want to propose a BC-breaking change to the `dist` directory layout. I'll first explain what the change is, and then why I need it. I'd like to know if people are OK with it. It WILL break code which needs to know where Cabal puts build products. CC @mgsloan, @23Skidoo , @dcoutts 

**Cabal-1.24 behavior.** Suppose we are building a Haskell module M.hs. If this module is in the public library, the interface file is placed in `dist/build/M.hi`. If the module is in any other component (e.g., a test suite) named `comp`, the interface file is placed in `dist/build/comp/M.hi`)

**Proposed new behavior.** I propose that libraries also get their own subdirectory: `dist/build/pkgname`; thus if we build `M.hs` in a library named `foobar` it will be placed in `dist/build/foobar/M.hi`. Obviously, this will break any scripts which rely on the interface files/object files having been placed in the old location.

**Motivation.** Cabal supports pre-processors which may generate extra object files that need to be linked into the final libraries/executables we build. This is tested by the `PreProcessExtraSources` `cabal:package-test`. For example, if we build `Foo.hsc`, this results in a `Foo_hsc.o` in our build directory.

If a preprocessor generates this extra object file, how does Cabal know to link it in? The "obvious" design is that when you run the preprocessor, it tells you what extra files need to be compiled and you go ahead and compile them. But this is not what Cabal does. Instead, what it does is simply look in the build directory for any files which look like ones that the preprocessor generated, and make sure to link them in. (https://github.com/ian-ross/cabal/commit/e38cb0c076fed041a8ea986e1dd2b44115dcda34)

Now here is the problem with Cabal-1.24's behavior: the build directories for the test suites/benchmarks/etc are all INSIDE the library's build directory. So clearly the search code will pick up the extra build products: disaster! And so the most obvious fix (without completely changing the preprocessor interface) is to move the directory out of the way.

The bug got uncovered by #3542 although I am not really sure why it didn't manifest earlier.

**Alternatives.** I think the most plausible alternative which doesn't break BC is to fix the search behavior to filter out subdirectories which correspond to other components in the Cabal file. But I am morally opposed to it. The current directory layout really is just WRONG. Consider:

```
name:                xyz
version:             0.1.0.0
license:             BSD3
author:              Edward Z. Yang
maintainer:          ezyang@cs.stanford.edu
extra-source-files:  ChangeLog.md
cabal-version:       >=1.10

library
[...]

PR #3557, opened 2016-07-15
A terrible hack which solves #3545.

ezyang:
See the comment in the code.  I don't like it but it's BC!


ezyang:
@23Skidoo @phadej What do you think?

phadej:
I don't think that re-arranging `dist` is breaking change as it's not clearly specified either. IIRC @dcoutts have commented about that somewhere.

Also e.g. https://github.com/ekmett/bound/blob/master/Setup.lhs doesn't rely on where stuff goes, as it asks `Cabal` where it puts stuff.

Also https://github.com/haskell/cabal/pull/3022 broke stuff already (for `cabal_macros.h`)

---

But if it helps for now, then LGTM

---

I'd propose to separate `dist/build` clearly per component, as we ultimately want to treat them fully independently apart that they come in single source package.

23Skidoo:
I'd prefer the solution originally proposed in #3545, i.e. rearranging the directories. Using `dist/build` for library files always felt wrong to me.

ezyang:
This change is vetoed!

PR #3571, opened 2016-07-17
Avoid`fromFlag`

ttuegel:
I was debugging a failing test in my `reconfigure` branch; in the end, I cleaned up many uses of the partial `fromFlag` function. Given that there is an unambiguous default for these cases, I didn't think there was much sense in even admitting the possibility of raising an exception.

### DIFF
--- a/Cabal/Distribution/Make.hs
+++ b/Cabal/Distribution/Make.hs
@@ -128,7 +128,7 @@ defaultMainHelper args =
 configureAction :: ConfigFlags -> [String] -> IO ()
 configureAction flags args = do
   noExtraFlags args
-  let verbosity = fromFlag (configVerbosity flags)
+  let verbosity = getVerbosity configVerbosity flags
   rawSystemExit verbosity "sh" $
     "configure"
     : configureArgs backwardsCompatHack flags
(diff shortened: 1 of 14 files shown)


ezyang:
Oh yes please! Does this nail all uses of `fromFlag` or did you just do two that you were looking at? (We really should deprecate that function.)

ttuegel:
I agree we should deprecate it. I *was* going to eliminate every use, but I found some uses had been added recently, so I wondered if I was alone in thinking it should be eliminated. There are few enough remaining calls that I'll go ahead and remove them all, then mark the function deprecated.

ezyang:
I think some of those uses are my fault, I bet I didn't realize `fromFlag` was partial and was just mimicking how it was used in nearby code.

To go even further down the rabbit hole, why do we even use the `Flag` constructor in the first place? Under what circumstances should a configuration flag use `Flag`? Very mysterious to me!

ttuegel:
> To go even further down the rabbit hole, why do we even use the Flag constructor in the first place? Under what circumstances should a configuration flag use Flag? Very mysterious to me!

This situation is tricky. Officially, `field = Flag foo` indicates a command-line argument was given to Cabal which indicates `foo` is the proper value for `field`; `NoFlag` indicates no command-line argument given.

However, for each flags structure, e.g. `BuildFlags`, there is a default value, e.g. `defaultBuildFlags`, intended to capture the default settings. So, a setting of `field = Flag foo` could have been specified on the command-line, or it could be a default value; generally there is no way to know. The most notorious offenders are the verbosity fields; most action wrappers set a default value for the verbosity in `Distribution/Simple.hs` before calling the real implementation.

There is also a `Monoid` instance with all the fields populated by `NoFlag`. Usually you can use this in place of `defaultBuildFlags`, but sometimes it will cause problems if the recipient of your `BuildFlags` does not pick a default value. (That was the true cause of my problem here.)

I would prefer if all the command-line argument handling were done in one place, i.e. `buildAction` should take a `BuildFlags` and call `build` with some structure that has all the default values specified (no more `Flag`s at all). I can't think of any reason `Distribution/Simple/GHC.hs` should be aware of the actual flags passed to Cabal on the command-line, but it is. Instead of `defaultBuildFlags`, perhaps a function `finalizeBuildConfig :: BuildFlags -> BuildConfig` where the latter structure has only fully specified fields?

While we're at it, maybe all the flags for all the different phases don't have to live in a single, 2300-line module that takes several minutes to recompile the `*Flags`-type for each action when you just wanted to add a single option to one action! But, I dream.

ttuegel:
I am closing the pull request because I am going to implement the refactoring I described above.

I decided we *need* this refactoring because I found WTFs like [this](https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Configure.hs#L663):
```.haskell
    reloc <-
       if not (fromFlag $ configRelocatable cfg)
            then return False
            else return True
```
I'm not even trying to shame anybody! There's a good chance I wrote that code; I've had my hands all over `configure`! But I can't let this go on.

ezyang:
@ttuegel The flags data structure are part of the very public Cabal API, specifically the `UserHooks` interface. So are you planning on finalizing into a `BuildConfig` after hooks are run?

ttuegel:
> The flags data structure are part of the very public Cabal API, specifically the UserHooks interface. So are you planning on finalizing into a BuildConfig after hooks are run?

I think I'll have `configToFlags :: BuildConfig -> BuildFlags` that returns a structure with no `NoFlag` entries. That should be even safer than what we do now.

ezyang:
OK. I guess this is going to need a disturbing amount of boilerplate but what else are you going to do.

Note that `BuildFlags` still makes sense as an interface to INVOKE Setup, because `NoFlag` lets you defer the choice of a default until later.

ttuegel:
> I guess this is going to need a disturbing amount of boilerplate but what else are you going to do.

I may decide to back off on the `finalizeBuildConfig` thing, but `fromFlag` has to go!

> Note that BuildFlags still makes sense as an interface to INVOKE Setup, because NoFlag lets you defer the choice of a default until later.

Absolutely!

ezyang:
> I may decide to back off on the finalizeBuildConfig thing, but fromFlag has to go!

Hear hear!

PR #3572, opened 2016-07-17
[RFC] Refactor Flag type

ttuegel:
Before I get too deeply invested in this, I wanted to request feedback.

This begins to realize my vision for refactoring the `Flag` type in Cabal. I am trying to solve two problems:

1. The use of the partial function `fromFlag`.
2. The problem of "configuration creep". I mean "configuration" in the sense of "The Configuration Problem," not in the sense of "cabal configure". The primary symptom of this is that default values for flags are strewn throughout the codebase.

To get feedback, I have refactored `TestFlags` as a model for what I would like to do to the rest of the codebase. Except for deprecations and new exports, I have maintained API compatibility. I have not removed any exports or (AFAIK) changed the types of any exports.

The `TestFlags` type is replaced with `TestConfig a` where `a` has kind `* -> *`. The easiest way to understand what this type parameter does is to consider this synonym:
```.haskell
type TestFlags = TestConfig Flag
```
which is equivalent to the old definition of `TestFlags`. `TestConfig a`'s sole constructor is named `TestFlags` to keep compatibility, I have just generalized the type.

Besides `Flag`, I have introduced
```.haskell
newtype Final a = Final { fromFinal :: a }
```
and I provide
```.haskell
finalizeTestFlags :: TestFlags -> TestConfig Final
```
which calls `fromFlagOrDefault` on each field with the appropriate default values. The result of `finalizeTestFlags` is a `TestConfig` with *all* fields populated. For compatibility, there is a trivial
```.haskell
testConfigToFlags :: TestConfig Final -> TestFlags
```
which does the obvious thing. There is marginally more boilerplate this way, but I think having concentrated boilerplate is better than having `fromFlag` throughout the codebase. To organize the boilerplate better and reduce recompilation times, I am splitting up `Distribution.Simple.Setup`, starting with `Distribution.Simple.Command.Test`.

Before I continue this refactoring for the other commands, do I hear any comments or objections?

/cc interested parties: @ezyang @23Skidoo @dcoutts


Cabal/Distribution/Simple/Command/Test.hs
```diff
@@ -0,0 +1,188 @@
+{-# LANGUAGE DeriveGeneric #-}
```
ttuegel: The interesting changes are mainly in this module.

ezyang:
I like this plan. Given the constraint that we want both flaggy and non-flaggy data types, I think this is the only possible thing you can do. I do have some inline comments so I'll put them in.

Cabal/Distribution/Flag.hs
```diff
@@ -0,0 +1,99 @@
...
+-- Copyright   :  Isaac Jones 2003-2004
+--                Duncan Coutts 2007
+--                Thomas Tuegel 2016
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- A monoid for command-line options such that successive options override
+-- their precedents.
```
ezyang: Obviously, eventually we should have docs for everything here.

Cabal/Distribution/Simple/Command.hs
```diff
@@ -184,16 +191,87 @@ optArg' ad mkflag showflag =
 noArg :: (Eq b) => b -> MkOptDescr (a -> b) (b -> a -> a) a
 noArg flag sf lf d = choiceOpt [(flag, (sf,lf), d)] sf lf d
 
-boolOpt :: (b -> Maybe Bool) -> (Bool -> b) -> SFlags -> SFlags
```
ezyang: OMG, you don't have to do this, but I would really love docs for these functions.

Cabal/Distribution/Simple/Command/Test.hs
```diff
@@ -0,0 +1,188 @@
...
+import Distribution.Simple.Command
+import Distribution.Simple.InstallDirs
+import Distribution.Simple.Utils
+import Distribution.Text
+import Distribution.Verbosity
+
+import Data.Char ( isAlpha )
+
+
+data TestConfig a = TestFlags { testDistPref    :: a FilePath
```
ezyang: Conventionally type variables of kind `* -> *` use `f` not `a`.

Cabal/Distribution/Simple/Command/Test.hs
```diff
@@ -0,0 +1,188 @@
...
+                              , testVerbosity   :: a Verbosity
+                              , testHumanLog    :: a PathTemplate
+                              , testMachineLog  :: a PathTemplate
+                              , testShowDetails :: a TestShowDetails
+                              , testKeepTix     :: a Bool
+                              , testOptions     :: [PathTemplate]
```
ezyang: In the putative docs, there needs to be an explanation when you say `f MyType` and when you say `[MyType]`. I think the reason is that the default for a list argument is always `[]`, and then if you specify multiple times you want it to cons on, e.g. `--dependency foo --dependency bar`?

23Skidoo:
I like this change.

> ```.haskell 
newtype Final a = Final { fromFinal :: a }
```

Starting with base-4.8 you can also use `Data.Functor.Identity`, but I guess that using it doesn't really buy us anything. `Final` is easier to type OTOH.

**Comment by ttuegel** [comment]
> Starting with base-4.8 you can also use Data.Functor.Identity, but I guess that using it doesn't really buy us anything. Final is easier to type OTOH.

Also, I don't think we want to pick up the stray instances. For example, `Monoid a => Monoid (Identity a)` would be particularly confusing, I think.

**Comment by ttuegel** [comment]
There is one incompatible change here that I didn't think of before. Import lists like
```.haskell
import Distribution.Simple.Setup ( TestFlags(..) )
```
must now become
```.haskell
import Distribution.Simple.Setup ( TestFlags, TestConfig(..) )
```

On the one hand, this is a pretty breaking change. On the other hand, we have setup dependencies now, so in principle this is not so bad. Thoughts?

Cabal/Distribution/Simple/Command/Test.hs
23Skidoo: Though sometimes you want Flag-like mappend behaviour for list flags as well, see https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Config.hs#L175.

23Skidoo:
That will break a lot of setup scripts. We'll probably have to change the default custom-setup upper bound to 1.26.

/cc @hvr

ttuegel:
If that's much trouble, I could also increase the boilerplate level a notch by introducing separate data types for `TestFlags` and `TestConfig`. That would actually eliminate the need for `Final`, so the amount of boilerplate probably stays about the same.

23Skidoo:
> That would actually eliminate the need for `Final`, so the amount of boilerplate probably stays about the same.

If that's indeed the case, the BC solution is of course preferable (if it's not too much trouble).

ttuegel:
I can maintain backwards compatibility with minimal boilerplate using `RecordWildCards`. I recognize that's not a popular extension. If we want to go without, record field names become verbose very quickly.

23Skidoo:
Using `RecordWildCards` in Cabal code is perfectly fine, there is already a bunch of modules with this extension enabled.

lspitzner:
I have been using something rather similar in a recent project, and used the following abstraction:

~~~~
class CZip k where
  cZip :: (forall a . f a -> g a -> h a) -> k f -> k g -> k h

instance CZip SomeConf where ...
~~~~

One has to write that (boilerplaty) instance, but automatically gets:

~~~~
fromMaybeSomeConf :: SomeConf Final -> SomeConf Maybe -> SomeConf Final
fromMaybeSomeConf = cZip (\def -> maybe def Final)

toMaybeSomeConf :: SomeConf Final -> SomeConf Maybe
toMaybeSomeConf c = cZip (\_ x -> Just $ copoint x) c c
~~~~

For that to work, SomeConf needs to have the form

~~~~
data SomeConf f = SomeConf { field1 :: f m1, field2 :: f m2, .. }
~~~~

where `m1`, `m2`, .. are Semigroups. E.g. most likely stuff like `Semigroup.Last Bool`, `Semigroup.Last Int` or `[FilePath]`.

* The current `Flag` combines the semantics of `Maybe` and `Semigroup.Last`. In this approach, these are separated.
* This means that the merging behaviour is encoded in the type of each field. It is implicit documentation (and annoying for usage). But you need to `copoint` the `Final`s anyway with the above approach.
* The default is a trivial `SomeConf Final` value, not mixed with any other logic.
* It should be trivial to automate the `CZip` instance via TH.

Not sure if this would be applicable for cabal, or if this would cause (more) breakage.

ezyang:
@lspitzner Very interesting. Unfortunately we need to keep the old version of `*Flags` around for BC. I wonder if we shouldn't convert into one of these functor-parametrized types, in preparation for one day making the BC-breaking change. I'll defer to @ttuegel on this one!

ttuegel:
> I wonder if we shouldn't convert into one of these functor-parametrized types, in preparation for one day making the BC-breaking change.

That's what I originally did, but we discovered that it breaks downstream import lists. 

However, I'm happy enough with the fully-compatible, separate `TestFlags` and `TestConfig` + `RecordWildCards` implementation. It has a little (but not much) more boilerplate than I originally wanted, but *all* the boilerplate is concentrated in `Distribution.Simple.Command.*` modules. We can even remove common boilerplate patterns like
```.haskell
let distPref = fooDistPref flags
    verbosity = fooVerbosity flags
```
So, I think it's a big win.

hvr:
I mostly agree with @23Skidoo here. But I'd also point out we can empirically find out how many packages would be affected (i.e. break) on Hackage to help decide whether it'd be more economical to retrofit setup-depends or tigthen cabal's implicit upper bound.

lspitzner:
Sure. I was in a "if we break things, break them correctly" mode, but if we can maintain BC that is better than avoiding small amounts of boilerplate.

ttuegel:
> I was in a "if we break things, break them correctly" mode

Agree 100%! :+1: 

Thanks for the feedback, everyone. I will close this for now and reopen when I have the complete, backwards-compatible refactoring.

Issue #3574, opened 2016-07-19
Convenience libraries BC regret


ezyang:
Pursuant to #3573 I went back and looked at the convenience libraries changeset to find out where I broke BC. Some of these are avoidable and I'm going to BC them.

- [x] `ComponentName` had `CLibName` turn to `CLibName String`.  This made the code nice and symmetric but plenty of client code needs the old version. So I'm going to unwind this back into `CSubLibName String` and `CLibName`. This also means a cluster of functions showBuildTarget, disambiguateBuildTargets, renderBuildTarget and componentStringName go back to their old type (since they need PackageId now)
- [x] Similarly, we'll bring back `library :: Maybe Library` in `PackageDescription` (nominating it to be the public library), and add a new field `subLibraries :: [Library]`
- [x] componentsConfig is going to come back as a helper function to massage the field into the old form. It won't be very correct with Backpack but that's ok for BC.
- [ ] I already restored prefixRelativeInstallDirs to old behavior but maybe we should deprecate it

Here are the ones I am not going to fix unless people tell me to:

* Fixing autogeneration to be per-component meant I plumbed `ComponentLocalBuildInfo` through a number of places:     initialBuildSteps,    writeAutogenFiles, generate,  generateMacros / generateComponentIdMacro,  autogenModulesDir, getLibSourceFiles,  getExeSourceFiles, installIncludeFiles, PPSuffixHandler, preprocessComponent, ppGreenCard, etc. I am NOT going to rename these and bring back BC versions of them
* HookedBuildInfo changed types. The change is for the better and I am unrepentant.
* OriginalModule got renamed to Module. The new name is much better, and not many people were using OriginalModule anyway.
* componentPackageRenaming was a "new" feature, it got renamed again to componentIncludes and I am unrepentant.

Some functions whose types changed but virtually no one on Hackage uses them: maybeHasName, mkComponentsLocalBuildInfo

ezyang:
OK closing.

PR #3580, opened 2016-07-21
BC fix: libraries -> library/subLibraries (#3574)

ezyang:
The resulting code is more verbose, but it is more backwards-compatible
and actually is simpler to understand in some cases (because
CLibName uniquely identifies the "public library"; no faffing
about with package names to figure it out.)

### DIFF
--- a/Cabal/Distribution/PackageDescription.hs
+++ b/Cabal/Distribution/PackageDescription.hs
@@ -32,6 +32,7 @@ module Distribution.PackageDescription (
         descCabalVersion,
         BuildType(..),
         knownBuildTypes,
+        allLibraries,
 
         -- ** Renaming
         ModuleRenaming(..),
@@ -125,7 +126,7 @@ import Distribution.Text
 import Language.Haskell.Extension
 
 import Data.Data                  (Data)
-import Data.List                  (nub, intercalate)
+import Data.List                  (nub, intercalate, find)
 import Data.Maybe                 (fromMaybe, maybeToList)
 import Data.Foldable as Fold      (Foldable(foldMap))
 import Data.Traversable as Trav   (Traversable(traverse))
@@ -189,7 +190,8 @@ data PackageDescription
         buildType      :: Maybe BuildType,
         setupBuildInfo :: Maybe SetupBuildInfo,
         -- components
-        libraries      :: [Library],
+        library        :: Maybe Library,
+        subLibraries   :: [Library],
         executables    :: [Executable],
         testSuites     :: [TestSuite],
         benchmarks     :: [Benchmark],
@@ -256,7 +258,8 @@ emptyPackageDescription
                       category     = "",
                       customFieldsPD = [],
                       setupBuildInfo = Nothing,
-                      libraries    = [],
+                      library      = Nothing,
+                      subLibraries = [],
                       executables  = [],
                       testSuites   = [],
                       benchmarks   = [],
@@ -393,7 +396,7 @@ instance Text ModuleRenaming where
 -- The Library type
 
 data Library = Library {
-        libName :: String,
+        libName :: Maybe String,
         exposedModules    :: [ModuleName],
         reexportedModules :: [ModuleReexport],
         requiredSignatures:: [ModuleName], -- ^ What sigs need implementations?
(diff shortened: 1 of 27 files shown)


Cabal/Distribution/PackageDescription.hs
```diff
@@ -419,39 +422,37 @@ instance Monoid Library where
...
+
+allLibraries :: PackageDescription -> [Library]
+allLibraries p = maybeToList (library p) ++ subLibraries p
 
 -- |If the package description has a library section, call the given
 --  function with the library build info as argument.
 withLib :: PackageDescription -> (Library -> IO ()) -> IO ()
-withLib pkg_descr f =
-   sequence_ [f lib | lib <- libraries pkg_descr, buildable (libBuildInfo lib)]
+withLib pkg_descr f = do
```
23Skidoo: `do` here looks unnecessary.

Cabal/Distribution/PackageDescription/Parse.hs
```diff
@@ -1213,14 +1223,14 @@ parseHookedBuildInfo inp = do
...
-            = liftM (\bis -> [(CLibName "", bis)]) (parseBI bi)
+            = liftM (\bis -> [(CLibName, bis)]) (parseBI bi)
     parseLib _ = return []
 
     parseStanza :: HookedBuildInfo -> [Field] -> ParseResult HookedBuildInfo
     parseStanza bis (F line inFieldName mName:bi)
         | Just k <- case lowercase inFieldName of
                         "executable" -> Just CExeName
-                        "library"    -> Just CLibName
+                        "library"    -> Just CSubLibName
```
23Skidoo: Don't you need to check whether this is a named library here?

ezyang: Nope, because it was handled in the code above. Historically HookedBuildInfo format, the public library was un-stanza'd.

Cabal/Distribution/Simple/Build.hs
```diff
@@ -406,7 +408,7 @@ testSuiteLibV09AsLibAndExe pkg_descr
   where
     bi  = testBuildInfo test
     lib = Library {
-            libName = testName test,
```
23Skidoo: This looks suspicious.

ezyang: Well, all of the code here is suspicious, really. The choice of libName here doesn't make a difference, it's never used later.

Cabal/Distribution/PackageDescription/Parse.hs
23Skidoo: `HookedBuildInfo` really needs to be better documented...
Also looks like `isStanzaHeader` should be made aware of test suites and benchmarks, right now it assumes that "executable" is the only non-library stanza type.

Cabal/Distribution/Simple/Build.hs
ezyang: Actually, I was wrong, it does matter. I'll add a fix.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -200,15 +207,14 @@ checkSanity pkg =
...
   ]
   where
-    -- The public library gets special dispensation, because it
+    -- The public 'library' gets special dispensation, because it
     -- is common practice to export a library and name the executable
-    -- the same as the package.  We always put the public library
-    -- in the top-level directory in dist, so no conflicts either.
-    libNames = filter (/= unPackageName (packageName pkg)) . map libName $ libraries pkg
+    -- the same as the package.
+    subLibNames = catMaybes . map libName $ subLibraries pkg
```
23Skidoo: Shouldn't this be `map (fromMaybe (packageName pkg) . libName) $ subLibraries`?

ezyang: Mmm. I guess the invariant is that everything in `subLibraries` has a non-Nothing package name. I guess we can check that!

23Skidoo: Yeah, I was looking for such a check, but it seem we don't have one.

PR #3593, opened 2016-07-22
Consistently update 'localPkgDescr' with 'HookedBuildInfo' + docs

### LINKED ISSUE #1757: Document HookedBuildInfo
Many of the hooks in a Cabal build are supposed to return that type, but the type itself has no documentation attached to it, and the type alias resolves to a rather uninformative `(Maybe BuildInfo, [(String, BuildInfo)])`. It looks like the `fst` is applied to the library and the `snd` has entries for executables.

ezyang:
I got sucked into this rathole while puzzling over why
we pass both 'PackageDescription' and 'LocalBuildInfo' (which
has a 'PackageDescription' in 'localPkgDescr').  The conclusion
is that they were different, but not for a good reason. Now they're
the same.

Added docs fixes #1757.

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -187,16 +187,13 @@ configureAction :: UserHooks -> ConfigFlags -> Args -> IO LocalBuildInfo
 configureAction hooks flags args = do
     distPref <- findDistPrefOrDefault (configDistPref flags)
     let flags' = flags { configDistPref = toFlag distPref }
+
+    -- See docs for 'HookedBuildInfo'
     pbi <- preConf hooks args flags'
 
     (mb_pd_file, pkg_descr0) <- confPkgDescr hooks verbosity
 
-    --get_pkg_descr (configVerbosity flags')
-    --let pkg_descr = updatePackageDescription pbi pkg_descr0
     let epkg_descr = (pkg_descr0, pbi)
-
-    --(warns, ers) <- sanityCheckPackage pkg_descr
-    --errorOut (configVerbosity flags') warns ers
 
     localbuildinfo0 <- confHook hooks epkg_descr flags'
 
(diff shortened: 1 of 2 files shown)


Cabal/Distribution/Simple.hs
```diff
@@ -292,6 +294,12 @@ cleanAction hooks flags args = do
     pbi <- preClean hooks args flags'
 
     (_, ppd) <- confPkgDescr hooks verbosity
+    -- It might seem like we are doing something clever here
+    -- but we're really not: if you look at the implementation
+    -- of 'clean' in the end all the package description is
+    -- used for is to clear out @extra-tmp-files@.  IMO,
+    -- the configure script goo should go into @dist@ too!
+    --          -- ezyang
```
tibbe: Blame will get this. No need to sign messages like is done in GHC. :)

ezyang: You know, you'd be shocked how often people ask me questions that could be solved by blame ;)

Cabal/Distribution/Types/HookedBuildInfo.hs
```diff
@@ -9,6 +9,55 @@ module Distribution.Types.HookedBuildInfo (
 import Distribution.Types.ComponentName
 import Distribution.Types.BuildInfo
 
+-- | 'HookedBuildInfo' is mechanism that hooks can use to
```
23Skidoo: Nice!

Cabal/Distribution/Types/HookedBuildInfo.hs
```diff
@@ -9,6 +9,55 @@ module Distribution.Types.HookedBuildInfo (
...
+--      1. We run the pre-hook, which produces a 'HookedBuildInfo'
+--         (e.g., in the Autoconf case, it reads it out from a file).
+--      2. We sanity-check the hooked build info with
+--         'sanityCheckHookedBuildInfo'.
+--      3. We update our 'PackageDescription' (either freshly read
+--         or cached from 'LocalBuildInfo') with 'updatePackageDescription'.
+--         In doing so, we must be careful to also update it in
+--         'localPkgDescr' in 'LocalBuildInfo', where a user can
+--         also get to it.  In practice the code also passes around
+--         the update 'PackageDescription' around explicitly (redundantly)
```
23Skidoo: s/update/updated/

Cabal/Distribution/Types/HookedBuildInfo.hs
```diff
@@ -9,6 +9,55 @@ module Distribution.Types.HookedBuildInfo (
...
+--         which is what everyone used to get the actual 'Component'.
+--         And that's a good thing because up until recently
+--         'localPkgDescr' was NOT updated.  Although, it doesn't
+--         look like anyone saw a bug because of this, thanks to
+--         1c20a6328579af9e37677d507e2e9836ef70ab9d.
+--
+-- It is not well-specified whether or not a 'HookedBuildInfo' applied
```
23Skidoo: @dcoutts, you might want to comment on this.

23Skidoo:
LGTM.

Issue #3488, opened 2016-06-09
Solver: Combine goal-ordering heuristics more effectively by assigning scores to goal choices


grayjay:
@kosmikus mentioned this in https://github.com/haskell/cabal/issues/3453#issuecomment-222078560.

The scoring would be similar to the scoring described in #2860, except that it would apply to goals, instead of package, flag, and stanza choices.  Scoring all nodes might actually be a simplification after #2860 is implemented.

Currently, the solver applies goal-ordering heuristics by either stably sorting the available goals or pruning non-preferred goals (when other goals are available).  This approach gives too much weight to some of the heuristics, particularly the earlier pruning steps.  Pruning is also problematic because it leaves fewer goals for features like `--reorder-goals` and "dynamic goal reordering" to work with. See https://github.com/haskell/cabal/issues/3453#issuecomment-222064219.

If heuristics were instead expressed as scores on goals, then the heuristics could be combined in a way that gives more control over their relative impact on the goal order that is used to explore the tree.

PR #3594, opened 2016-07-22
Add 'WeightedPSQ' and use it to sort package, flag, and stanza choices.

grayjay:
I split this out from #2917 because the data structure is also needed for goal-scoring (#3488).  `WeightedPSQ` associates weights with choices in the search tree.  This PR uses `WeightedPSQ` for package, flag, and stanza choices and `PSQ` for goal choices, but eventually all choices should use `WeightedPSQ`.  There are a few TODOs in the code.

/cc @kosmikus 

EDIT: This PR uses `[Double]` as the weight type to avoid changing the behavior of the solver.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/WeightedPSQ.hs
+++ b/cabal-install/Distribution/Solver/Modular/WeightedPSQ.hs
@@ -23,7 +23,7 @@ import Data.Ord (comparing)
 import qualified Data.Traversable as T
 import Prelude hiding (filter, length, lookup)
 
-import Distribution.Solver.Modular.PSQ (Degree(..))
+import Distribution.Solver.Modular.Degree
 
 newtype WeightedPSQ w k v = WeightedPSQ [(w, k, v)]
   deriving (Eq, Show, Functor, F.Foldable, T.Traversable)
--- a/cabal-install/Distribution/Solver/Modular/WeightedPSQ.hs
+++ b/cabal-install/Distribution/Solver/Modular/WeightedPSQ.hs
@@ -1,21 +1,19 @@
 {-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}
 module Distribution.Solver.Modular.WeightedPSQ (
     WeightedPSQ
-  , filter
   , fromList
+  , toList
   , keys
+  , weights
   , length
   , degree
   , isZeroOrOne
+  , filter
   , lookup
-  , mapWeightsWithKey
   , mapWithKey
-  , toList
+  , mapWeightsWithKey
   , union
-  , weights
   ) where
-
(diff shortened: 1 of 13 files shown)


ezyang:
Let me be the grumpy code reviewer:

1. Needs docs!
2. Needs asymptotics!
3. Degree is in an odd place

But I'm all for defining data structures we need.

kosmikus:
@grayjay Looks promising so far. It doesn't seem to have any negative performance impact afaics.

grayjay:
Thanks.  I moved `Degree` into its own module, and I'll add documentation and unit tests next.

@kosmikus How do you want to proceed?  Should I open a new PR with just `WeightedPSQ` so that it is easier to merge, or should I continue cleaning up all of these changes?  I realized that I can remove some of the TODOs because they only relate to install plan scoring.

grayjay:
I cleaned it up, added tests, and added documentation with asymptotics, similar to D.Compat.Graph.  It's ready for review.  I'd like to clean up the history before it is merged, though.

cabal-install/Distribution/Solver/Modular/WeightedPSQ.hs
```diff
@@ -0,0 +1,102 @@
...
+-- Each element has a key ('k'), value ('v'), and weight ('w'). All operations
+-- that add elements or modify weights stably sort the elements by weight.
+newtype WeightedPSQ w k v = WeightedPSQ [(w, k, v)]
+  deriving (Eq, Show, Functor, F.Foldable, T.Traversable)
+
+-- | /O(N)/.
+filter :: (v -> Bool) -> WeightedPSQ k w v -> WeightedPSQ k w v
+filter p (WeightedPSQ xs) = WeightedPSQ (L.filter (p . triple_3) xs)
+
+-- | /O(N)/.
```
ezyang: Ouch! Should there be a TODO here to track the length separately?

grayjay: `length` is actually no longer called.  I think it was used by --reorder-goals, which now uses `degree` to avoid traversing the whole list.  Calling `length` there had a significant impact on performance because it was called on the children of every available goal choice after the children were filtered with an expensive predicate.  I'll remove it.

ezyang: Cool, thanks.

23Skidoo:
@grayjay @kosmikus Is this ready to go in?

grayjay:
I was done working on it, but I wasn't sure if @kosmikus wanted to take another look.

23Skidoo:
Let's merge then, if @kosmikus dislikes something he can always revert.

PR #3596, opened 2016-07-22
Pretty utils

phadej:
Move pretty-printing related stuff out of `Parse*` modules.

The move of `writePackageDescription` is breaking change, but on the way for `Parsec` the whole 
`Distribution.PackageDescription.Parse` will be redone (types will change).


ezyang:
OK sure.

23Skidoo:
LGTM. I don't think any custom setup scripts are using any of this stuff.

phadej:
appveyor seems to have some transient error

PR #3602, opened 2016-07-23
Parsec parser

phadej:
types are in places, values are `undefined`.

TODO (not exhaustive):
- [x] test for tab warning (wont do in this PR)
- [x] test for bom warning
- [x] parse but warn about nbsp (`\194`)
- [x] test about nbsp
- [x] lots of tests for invalid fields, we want this parser to be as strict as it could be. Relaxing it is easier.
- [x] warn if flag declration doesn't specify `default` and `manual` fields (wont do in this PR)
- [x] add classier to `PWarning`
- [x] equip `PWarning` / `ParseResult` with severity, some warnings are recoverable (e.g. tags in versions), but we don't want them into Hackage. Do we?
- [x] amend `parser-test` to count warnings for both `readp` and `parsec`

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -327,6 +327,11 @@ flag parsec
   default:      False
   manual:       True
 
+flag parsec-struct-diff
+  description:  Use StructDiff in parsec tests. Affects only parsec tests.
+  default:      False
+  manual:       True
+
 library
   build-depends:
     array      >= 0.1 && < 0.6,
@@ -648,9 +653,7 @@ test-suite parser-tests
 
   type: exitcode-stdio-1.0
   main-is: ParserTests.hs
-  other-modules:
-    DiffInstances
-    StructDiff
+
   hs-source-dirs: tests
   build-depends:
     base,
@@ -659,11 +662,19 @@ test-suite parser-tests
     bytestring,
     directory,
     filepath,
-    generics-sop ==0.2.*,
-    these >=0.7.1 && <0.8,
-    singleton-bool >=0.1.1.0 && <0.2,
-    keys,
     Cabal
+
+  if flag(parsec-struct-diff)
+    build-depends:
+      generics-sop ==0.2.*,
+      these >=0.7.1 && <0.8,
+      singleton-bool >=0.1.1.0 && <0.2,
+      keys
+    other-modules:
+      DiffInstances
+      StructDiff
+    cpp-options: -DHAS_STRUCT_DIFF
+
   ghc-options: -Wall -rtsopts
   default-extensions: CPP
   default-language: Haskell2010
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -322,6 +322,11 @@ flag old-directory
   description:  Use directory < 1.2 and old-time
   default:      False
 
+flag parsec
+  description:  Use parsec parser
+  default:      False
+  manual:       True
+
 library
   build-depends:
     array      >= 0.1 && < 0.6,
@@ -482,6 +487,30 @@ library
     Language.Haskell.Extension
     Distribution.Compat.Binary
 
+  if flag(parsec)
+    cpp-options: -DCABAL_PARSEC
+    build-depends:
+      transformers,
+      parsec >= 3.1.9 && <3.2
+    build-tools:
+      alex >= 3.1.4
+    exposed-modules:
+      Distribution.Compat.Parsec
+      Distribution.PackageDescription.Parsec
+      Distribution.PackageDescription.Parsec.FieldDescr
+      Distribution.Parsec.Class
+      Distribution.Parsec.ConfVar
+      Distribution.Parsec.Lexer
+      Distribution.Parsec.LexerMonad
+      Distribution.Parsec.Parser
+      Distribution.Parsec.Types.Common
+      Distribution.Parsec.Types.Field
+      Distribution.Parsec.Types.FieldDescr
+      Distribution.Parsec.Types.ParseResult
+      Distribution.Compat.DList
+
+  -- Move DList to other-module if/when D.C.Lens is done
+
   other-modules:
     Distribution.Backpack.PreExistingComponent
     Distribution.Backpack.ReadyComponent
@@ -609,6 +638,32 @@ test-suite package-tests
     old-time
   if !os(windows)
     build-depends: unix, exceptions
+  ghc-options: -Wall -rtsopts
+  default-extensions: CPP
+  default-language: Haskell2010
+
+test-suite parser-tests
+  if !flag(parsec)
+    buildable: False
+
+  type: exitcode-stdio-1.0
+  main-is: ParserTests.hs
+  other-modules:
+    DiffInstances
+    StructDiff
+  hs-source-dirs: tests
+  build-depends:
+    base,
+    containers,
+    tar >=0.5 && <0.6,
+    bytestring,
+    directory,
+    filepath,
+    generics-sop ==0.2.*,
+    these >=0.7.1 && <0.8,
+    singleton-bool >=0.1.1.0 && <0.2,
+    keys,
+    Cabal
   ghc-options: -Wall -rtsopts
   default-extensions: CPP
   default-language: Haskell2010
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -543,6 +543,7 @@ library
     Distribution.Compat.GetShortPathName
     Distribution.Compat.MonadFail
     Distribution.Compat.Prelude
+    Distribution.Compat.SnocList
     Distribution.GetOpt
     Distribution.Lex
     Distribution.Utils.String
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -498,7 +498,7 @@ library
       transformers,
       parsec >= 3.1.9 && <3.2
     build-tools:
-      alex >= 3.1.4
+      alex >=3.1.4 && <3.3
     exposed-modules:
       Distribution.Compat.Parsec
       Distribution.PackageDescription.Parsec
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -303,6 +303,7 @@ extra-source-files:
   tests/PackageTests/UniqueIPID/P2/M.hs
   tests/PackageTests/UniqueIPID/P2/my.cabal
   tests/PackageTests/multInst/my.cabal
+  tests/ParserTests/warnings/bom.cabal
   tests/Setup.hs
   tests/hackage/check.sh
   tests/hackage/download.sh
@@ -651,7 +652,25 @@ test-suite parser-tests
     buildable: False
 
   type: exitcode-stdio-1.0
+  hs-source-dirs: tests
   main-is: ParserTests.hs
+  build-depends:
+    base,
+    bytestring,
+    filepath,
+    tasty,
+    tasty-hunit,
+    tasty-quickcheck,
+    Cabal
+  ghc-options: -Wall
+  default-language: Haskell2010
+
+test-suite parser-hackage-tests
+  if !flag(parsec)
+    buildable: False
+
+  type: exitcode-stdio-1.0
+  main-is: ParserHackageTests.hs
 
   hs-source-dirs: tests
   build-depends:
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -304,6 +304,7 @@ extra-source-files:
   tests/PackageTests/UniqueIPID/P2/my.cabal
   tests/PackageTests/multInst/my.cabal
   tests/ParserTests/warnings/bom.cabal
+  tests/ParserTests/warnings/nbsp.cabal
   tests/Setup.hs
   tests/hackage/check.sh
   tests/hackage/download.sh
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -528,7 +528,6 @@ library
     Distribution.Compat.CopyFile
     Distribution.Compat.GetShortPathName
     Distribution.Compat.MonadFail
-    Distribution.Compat.DList
     Distribution.Compat.Prelude
     Distribution.GetOpt
     Distribution.Lex
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -304,7 +304,19 @@ extra-source-files:
   tests/PackageTests/UniqueIPID/P2/my.cabal
   tests/PackageTests/multInst/my.cabal
   tests/ParserTests/warnings/bom.cabal
+  tests/ParserTests/warnings/bool.cabal
+  tests/ParserTests/warnings/deprecatedfield.cabal
+  tests/ParserTests/warnings/extratestmodule.cabal
+  tests/ParserTests/warnings/gluedop.cabal
   tests/ParserTests/warnings/nbsp.cabal
+  tests/ParserTests/warnings/newsyntax.cabal
+  tests/ParserTests/warnings/oldsyntax.cabal
+  tests/ParserTests/warnings/subsection.cabal
+  tests/ParserTests/warnings/trailingfield.cabal
+  tests/ParserTests/warnings/unknownfield.cabal
+  tests/ParserTests/warnings/unknownsection.cabal
+  tests/ParserTests/warnings/utf8.cabal
+  tests/ParserTests/warnings/versiontag.cabal
   tests/Setup.hs
   tests/hackage/check.sh
   tests/hackage/download.sh
--- a/Cabal/Distribution/Compat/Parsec.hs
+++ b/Cabal/Distribution/Compat/Parsec.hs
@@ -56,7 +56,8 @@ integral = toNumber <$> some d P.<?> "integral"
     f '9' = Just 9
     f _   = Nothing
 
--- | Greedely munch characters while predicate holds.
+-- | Greedily munch characters while predicate holds.
+-- Require at least one character.
 munch1
     :: P.Stream s m Char
     => (Char -> Bool)
@@ -64,6 +65,7 @@ munch1
 munch1 = some . P.satisfy
 
 -- | Greedely munch characters while predicate holds.
+-- Always succeeds.
 munch
     :: P.Stream s m Char
     => (Char -> Bool)
--- a/Cabal/Distribution/Compat/Parsec.hs
+++ b/Cabal/Distribution/Compat/Parsec.hs
@@ -1,0 +1,71 @@
+{-# LANGUAGE FlexibleContexts #-}
+module Distribution.Compat.Parsec (
+    P.Parsec,
+    P.ParsecT,
+    P.Stream,
+    (P.<?>),
+
+    P.runParser,
+
+    -- * Combinators
+    P.between,
+    P.option,
+    P.optional,
+    P.optionMaybe,
+    P.try,
+    P.sepBy,
+    P.sepBy1,
+    P.choice,
+
+    -- * Char
+    integral,
+    P.char,
+    P.anyChar,
+    P.satisfy,
+    P.space,
+    P.spaces,
+    P.string,
+    munch,
+    munch1,
+    P.oneOf,
+    ) where
+
+import           Distribution.Compat.Prelude
+import           Prelude ()
+
+import qualified Text.Parsec                 as P
+import qualified Text.Parsec.Pos             as P
+
+integral :: (P.Stream s m Char, Integral a) => P.ParsecT s u m a
+integral = toNumber <$> some d P.<?> "integral"
+  where
+    toNumber = foldl' (\a b -> a * 10 + b) 0
+    d = P.tokenPrim
+          (\c -> show [c])
+          (\pos c _cs -> P.updatePosChar pos c)
+          f
+    f '0' = Just 0
+    f '1' = Just 1
+    f '2' = Just 2
+    f '3' = Just 3
+    f '4' = Just 4
+    f '5' = Just 5
+    f '6' = Just 6
+    f '7' = Just 7
+    f '8' = Just 8
+    f '9' = Just 9
+    f _   = Nothing
+
+-- | Greedely munch characters while predicate holds.
+munch1
+    :: P.Stream s m Char
+    => (Char -> Bool)
+    -> P.ParsecT s u m String
+munch1 = some . P.satisfy
+
+-- | Greedely munch characters while predicate holds.
+munch
+    :: P.Stream s m Char
+    => (Char -> Bool)
+    -> P.ParsecT s u m String
+munch = many . P.satisfy
--- a/Cabal/Distribution/PackageDescription/Parsec.hs
+++ b/Cabal/Distribution/PackageDescription/Parsec.hs
@@ -1,0 +1,541 @@
+{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE Rank2Types #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.PackageDescription.Parsec
+-- Copyright   :  Isaac Jones 2003-2005
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- This defined parsers and partial pretty printers for the @.cabal@ format.
+
+module Distribution.PackageDescription.Parsec (
+    -- * Package descriptions
+    readGenericPackageDescription,
+    parseGenericPackageDescription,
+
+    -- ** Parsing
+    ParseResult,
+
+    -- ** Supplementary build information
+    -- readHookedBuildInfo,
+    -- parseHookedBuildInfo,
+    ) where
+
+import           Distribution.Compat.Prelude
+import           Prelude ()
+
+import qualified Data.ByteString                                   as BS
+import           Data.List                                         (partition)
+import qualified Data.Map                                          as Map
+import           System.Directory
+                 (doesFileExist)
+import qualified Text.Parsec                                       as P
+import qualified Text.Parsec.Error                                 as P
+
+import           Distribution.PackageDescription
+import           Distribution.Simple.Utils
+                 (die, fromUTF8BS, warn)
+import           Distribution.Verbosity                            (Verbosity)
+
+import           Distribution.PackageDescription.Parsec.FieldDescr
+import           Distribution.Parsec.Class                         (parsec)
+import           Distribution.Parsec.ConfVar
+                 (parseConditionConfVar)
+import           Distribution.Parsec.Parser
+import           Distribution.Parsec.Types.Common
+import           Distribution.Parsec.Types.Field                   (getName)
+import           Distribution.Parsec.Types.FieldDescr
+import           Distribution.Parsec.Types.ParseResult
+
+import Distribution.Text (display)
+import Distribution.Version (mkVersion, Version, asVersionIntervals, orLaterVersion, LowerBound (..))
+
+--import Debug.Trace
+
+-- ---------------------------------------------------------------
+-- Parsing
+
+-- | Helper combinator to do parsing plumbing for files.
+--
+-- Given a parser and a filename, return the parse of the file,
+-- after checking if the file exists.
+--
+-- Argument order is chosen to encourage partial application.
+readAndParseFile
+    :: (BS.ByteString -> ParseResult a)
+    -> Verbosity
+    -> FilePath
+    -> IO a
+readAndParseFile parser verbosity fpath = do
+    exists <- doesFileExist fpath
+    unless exists
+        (die $ "Error Parsing: file \"" ++ fpath ++ "\" doesn't exist. Cannot continue.")
+    bs <- BS.readFile fpath
+    let (warnings, errors, result) = runParseResult (parser bs)
+    traverse_ (warn verbosity . showPWarning fpath) warnings
+    traverse_ (warn verbosity . showPError fpath) errors
+    case result of
+        Nothing -> die $ "Failing parsing \"" ++ fpath ++ "\"."
+        Just x  -> return x
+
+-- | Parse the given package file.
+readGenericPackageDescription :: Verbosity -> FilePath -> IO GenericPackageDescription
+readGenericPackageDescription = readAndParseFile parseGenericPackageDescription
+
+------------------------------------------------------------------------------
+-- | Parses the given file into a 'GenericPackageDescription'.
+--
+-- In Cabal 1.2 the syntax for package descriptions was changed to a format
+-- with sections and possibly indented property descriptions.
+--
+-- TODO: add lex warnings
+parseGenericPackageDescription :: BS.ByteString -> ParseResult GenericPackageDescription
+parseGenericPackageDescription bs = case readFields bs of
+    Right fs  -> parseGenericPackageDescription' fs
+    -- | TODO: better marshalling of errors
+    Left perr -> parseFatalFailure (Position 0 0) (show perr)
+
+runFieldParser :: FieldParser a -> [FieldLine Position] -> ParseResult a
+runFieldParser p ls = runFieldParser' pos p =<< fieldlinesToString pos ls
+  where
+    -- TODO: make per line lookup
+    pos = case ls of
+        []                     -> Position 0 0
+        (FieldLine pos' _ : _) -> pos'
+
+fieldlinesToBS :: [FieldLine ann] -> BS.ByteString
+fieldlinesToBS = BS.intercalate "\n" . map (\(FieldLine _ bs) -> bs)
+
+-- TODO: Take position  from FieldLine
+-- TODO: Take field name
+fieldlinesToString :: Position -> [FieldLine ann] -> ParseResult String
+fieldlinesToString pos fls =
+    let str = intercalate "\n" . map (\(FieldLine _ bs') -> fromUTF8BS bs') $ fls
+    in if '\xfffd' `elem` str
+        then str <$ parseWarning pos PWTUTF "Invalid UTF8 encoding"
+        else pure str
+
+runFieldParser' :: Position -> FieldParser a -> String -> ParseResult a
+runFieldParser' (Position row col) p str = case P.runParser p' [] "<field>" str of
+    Right (pok, ws) -> do
+        -- | TODO: map pos
+        traverse_ (\(PWarning t pos w) -> parseWarning pos t w) ws 
+        pure pok
+    Left err        -> do
+        let ppos = P.errorPos err
+        -- Positions start from 1:1, not 0:0
+        let epos = Position (row - 1 + P.sourceLine ppos) (col - 1 + P.sourceColumn ppos)
+        let msg = P.showErrorMessages
+                "or" "unknown parse error" "expecting" "unexpected" "end of input"
+                (P.errorMessages err)
+
+        parseFatalFailure epos $ msg ++ ": " ++ show str
+  where
+    p' = (,) <$ P.spaces <*> p <* P.spaces <* P.eof <*> P.getState
+
+data GPDS = Fields | Sections
+
+-- Note [Accumulating parser]
+parseGenericPackageDescription'
+    :: [Field Position]
+    -> ParseResult GenericPackageDescription
+parseGenericPackageDescription' fs = do
+    let (newSyntax, fs') = sectionizeFields fs
+    (_, gpd) <- foldM go (Fields, emptyGpd) fs'
+    -- Various post checks
+    maybeWarnCabalVersion newSyntax (packageDescription gpd)
+    checkForUndefinedFlags gpd
+    -- TODO: do other validations
+    return gpd
+  where
+    go :: (GPDS, GenericPackageDescription)
+       -> Field Position
+       -> ParseResult (GPDS, GenericPackageDescription)
+    go s                (IfElseBlock pos _ _ _)  = do
+        parseFailure pos "if else block on the top level"
+        return s
+    go (Fields, gpd)   (Field (Name pos name) fieldLines) =
+        case Map.lookup name pdFieldParsers of
+            -- | TODO: can be more accurate
+            Nothing -> fieldlinesToString pos fieldLines >>= \value -> case storeXFieldsPD name value (packageDescription gpd) of
+                Nothing -> do
+                    parseWarning pos PWTUnknownField $ "Unknown field: " ++ show name
+                    return (Fields, gpd)
+                Just pd -> return (Fields, gpd { packageDescription = pd })
+            Just parser -> (\pd -> (Fields, gpd { packageDescription = pd }))
+                <$> runFieldParser (parser $ packageDescription gpd) fieldLines
+    go (Sections, gpd) (Field (Name pos _) _) = do
+        parseWarning pos PWTTrailingFields "Ignoring trailing fields after sections"
+        return (Sections, gpd)
+    go (s, gpd)        (Section name args fields) =
+        (,) s <$> parseSection gpd name args fields
+
+    pdFieldParsers :: Map FieldName (PackageDescription -> FieldParser PackageDescription)
+    pdFieldParsers = Map.fromList $
+        map (\x -> (fieldName x, fieldParser x)) pkgDescrFieldDescrs
+
+    emptyGpd :: GenericPackageDescription
+    emptyGpd = GenericPackageDescription emptyPackageDescription [] Nothing [] [] [] []
+
+    parseSection
+        :: GenericPackageDescription
+        -> Name Position
+        -> [SectionArg Position]
+        -> [Field Position]
+        -> ParseResult GenericPackageDescription
+    parseSection gpd (Name pos name) args fields
+        | name == "library" && null args = do
+            -- TODO: check that library is defined once
+            l <- parseCondTree libFieldDescrs storeXFieldsLib (targetBuildDepends . libBuildInfo) emptyLibrary fields
+            let gpd' = gpd { condLibrary = Just l }
+            pure gpd'
+
+        | name == "executable" = do
+            -- TODO: make a combinator for this, repeated in tests and bench
+            name' <- parseName pos args
+            -- Note: we don't parse the "executable" field here, hence the tail hack. Duncan 2010
+            exe <- parseCondTree (tail executableFieldDescrs) storeXFieldsExe (targetBuildDepends . buildInfo) emptyExecutable fields
+            -- TODO check duplicate name here?
+            let gpd' = gpd { condExecutables = condExecutables gpd ++ [(name', exe)] }
+            pure gpd'
+
+        | name == "test-suite" = do
+            name' <- parseName pos args
+            testStanza <- parseCondTree testSuiteFieldDescrs storeXFieldsTest (targetBuildDepends . testStanzaBuildInfo) emptyTestStanza fields
+            testSuite <- traverse (validateTestSuite pos) testStanza
+            -- TODO check duplicate name here?
+            let gpd' = gpd { condTestSuites = condTestSuites gpd ++ [(name', testSuite)] }
+            pure gpd'
+
+        | name == "benchmark" = do
+            name' <- parseName pos args
+            benchStanza <- parseCondTree benchmarkFieldDescrs storeXFieldsBenchmark (targetBuildDepends . benchmarkStanzaBuildInfo) emptyBenchmarkStanza fields
+            bench <- traverse (validateBenchmark pos) benchStanza
+            -- TODO check duplicate name here?
+            let gpd' = gpd { condBenchmarks = condBenchmarks gpd ++ [(name', bench)] }
+            pure gpd'
+
+        | name == "flag" = do
+            name' <- parseName pos args
+            name'' <- runFieldParser' pos parsec name' `recoverWith` FlagName ""
+            flag <- parseFields flagFieldDescrs warnUnrec (emptyFlag name'') fields
+            -- Check default flag
+            let gpd' = gpd { genPackageFlags = genPackageFlags gpd ++ [flag] }
+            pure gpd'
+
+        | name == "custom-setup" && null args = do
+            sbi <- parseFields setupBInfoFieldDescrs warnUnrec mempty fields
+            let pd = packageDescription gpd
+            -- TODO: what if already defined?
+            let gpd' = gpd { packageDescription = pd { setupBuildInfo = Just sbi } }
+            pure gpd'
+
+        | name == "source-repository" = do
+            kind <- case args of
+                [SecArgName spos secName] ->
+                    runFieldParser' spos parsec (fromUTF8BS secName) `recoverWith` RepoHead
+                [] -> do
+                    parseFailure pos $ "'source-repository' needs one argument"
+                    pure RepoHead
+                _ -> do
+                    parseFailure pos $ "Invalid source-repository kind " ++ show args
+                    pure RepoHead
+            sr <- parseFields sourceRepoFieldDescrs warnUnrec (emptySourceRepo kind) fields
+            -- I want lens
+            let pd =  packageDescription gpd
+            let srs = sourceRepos pd
+            let gpd' = gpd { packageDescription = pd { sourceRepos = srs ++ [sr] } }
+            pure gpd'
+
+        | otherwise = do
+            parseWarning pos PWTUnknownSection $ "Ignoring section: " ++ show name
+            pure gpd
+
+    newSyntaxVersion :: Version
+    newSyntaxVersion = mkVersion [1, 2]
+
+    maybeWarnCabalVersion :: Bool -> PackageDescription -> ParseResult ()
+    maybeWarnCabalVersion newsyntax pkg
+      | newsyntax && specVersion pkg < newSyntaxVersion 
+      = parseWarning (Position 0 0) PWTNewSyntax $
+             "A package using section syntax must specify at least\n"
+          ++ "'cabal-version: >= 1.2'."
+
+    maybeWarnCabalVersion newsyntax pkg
+      | not newsyntax && specVersion pkg >= newSyntaxVersion 
+      = parseWarning (Position 0 0) PWTOldSyntax $
+             "A package using 'cabal-version: "
+          ++ displaySpecVersion (specVersionRaw pkg)
+          ++ "' must use section syntax. See the Cabal user guide for details."
+      where
+        displaySpecVersion (Left version)       = display version
+        displaySpecVersion (Right versionRange) =
+          case asVersionIntervals versionRange of
+            [] {- impossible -}           -> display versionRange
+            ((LowerBound version _, _):_) -> display (orLaterVersion version)
+
+    maybeWarnCabalVersion _ _ = return ()
+
+{-
+    handleFutureVersionParseFailure :: Version -> ParseResult a -> ParseResult GenericPackageDescription
+    handleFutureVersionParseFailure _cabalVersionNeeded _parseBody =
+        error "handleFutureVersionParseFailure"
+-}
+
+ {-
+      undefined (unless versionOk (warning message) >> parseBody)
+        `catchParseError` \parseError -> case parseError of
+        TabsError _   -> parseFail parseError
+        _ | versionOk -> parseFail parseError
+          | otherwise -> fail message
+      where versionOk = cabalVersionNeeded <= cabalVersion
+            message   = "This package requires at least Cabal version "
+                     ++ display cabalVersionNeeded
+    -}
+
+    checkForUndefinedFlags
+        :: GenericPackageDescription
+        -> ParseResult ()
+    checkForUndefinedFlags _gpd = pure ()
+{-
+        let definedFlags = map flagName flags
+        mapM_ (checkCondTreeFlags definedFlags) (maybeToList mlib)
+        mapM_ (checkCondTreeFlags definedFlags . snd) sub_libs
+        mapM_ (checkCondTreeFlags definedFlags . snd) exes
+        mapM_ (checkCondTreeFlags definedFlags . snd) tests
+
+    checkCondTreeFlags :: [FlagName] -> CondTree ConfVar c a -> PM ()
+    checkCondTreeFlags definedFlags ct = do
+        let fv = nub $ freeVars ct
+        unless (all (`elem` definedFlags) fv) $
+            fail $ "These flags are used without having been defined: "
+                ++ intercalate ", " [ n | FlagName n <- fv \\ definedFlags ]
+-}
+
+parseName :: Position -> [SectionArg Position] -> ParseResult String
+parseName pos args = case args of
+    [SecArgName _pos secName] ->
+         pure $ fromUTF8BS secName
+    [SecArgStr _pos secName] ->
+         pure secName
+    [] -> do
+         parseFailure pos $ "name required"
+         pure ""
+    _ -> do
+         -- TODO: pretty print args
+         parseFailure pos $ "Invalid name " ++ show args
+         pure ""
+
+-- | Parse a list of fields, given a list of field descriptions,
+--   a structure to accumulate the parsed fields, and a function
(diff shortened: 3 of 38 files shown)


Cabal/Distribution/Parsec/Lexer.x
```diff
@@ -0,0 +1,257 @@
...
+@oplike      = [ \, \. \= \< \> \+ \* \- \& \| \! \$ \% \^ \@ \# \? \/ \\ \~ ]+
+
+tokens :-
+
+<0> {
+  $bom   ;
+  ()     ;
+}
+
+<bol_section, bol_field_layout, bol_field_braces> {
```
ezyang: OK I give up, what does 'bol' stand for.

Cabal/Distribution/Parsec/Lexer.x
```diff
@@ -0,0 +1,257 @@
...
+
+}
+-- Various character classes
+
+$space           = \          -- single space char
+$digit           = 0-9        -- digits
+$alpha           = [a-z A-Z]  -- alphabetic characters
+$symbol          = [\= \< \> \+ \* \- \& \| \! \$ \% \^ \@ \# \? \/ \\ \~]
+$ctlchar         = [\x0-\x1f \x7f]
+$printable       = \x0-\x10ffff # $ctlchar   -- so no \n \r
```
ezyang: Well, not JUST no \n and \r; \t is excluded too right?

Cabal/Distribution/Lexer.x
```diff
@@ -0,0 +1,244 @@
+{
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Parsec.Lexer
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- Lexer for the cabal files.
```
ezyang: Is any of this code from a sample / modeled off of another module? If so, cite it.

Cabal/Distribution/Parsec/Types/Field.hs
```diff
@@ -0,0 +1,81 @@
...
+mkName ann bs = Name ann (B.map Char.toLower bs)
+
+getName :: Name a -> ByteString
+getName (Name _ bs) = bs
+
+-------------------------------------------------------------------------------
+-- Position
+-------------------------------------------------------------------------------
+
+data Position = Position
```
ezyang: Two points: (1) are you sure you just want position, and not source *spans* (row/col to row/col), and (2) would be convenient to have an unhelpful source position, e.g., if you use simpleParsec

Cabal/Distribution/Lexer.x
```diff
@@ -0,0 +1,244 @@
...
+
+@nl          = \n | \r\n | \r
+@name        = $nameextra* $namecore $nameextra*
+@string      = \" ( $instr | \\ $instresc )* \"
+@numlike     = $digit [$digit \.]*
+@oplike      = [ \, \. \= \< \> \+ \* \- \& \| \! \$ \% \^ \@ \# \? \/ \\ \~ ]+
+
+tokens :-
+
+<0> {
```
ezyang: FWIW, I'm pretty sure we have a check to make sure there is no leading BOM in Cabal files. So if we record this in the result of parse we can do the check purely rather than by re-reading the Cabal file and looking for a BOM. See `checkCabalFileBOM` in `Cabal/Distribution/PackageDescription/Check.hs`

phadej: For old parser we have `readUTF8File :: FilePath -> IO String`, but no we use `readFile :: FilePath -> IO ByteString` and parse that.

But we don't need to parse whole file into UTF8, so it's omitted for later steps.

Cabal/Distribution/Lexer.x
phadej: I got this from @dcoutts

Cabal/Distribution/Parsec/Lexer.x
phadej: IIRC: braces or line

Cabal/Distribution/Parsec/Types/Field.hs
phadej: 1. I'm quite sure that beginning of the token would be enough (the end of token is easy to calculate for all tokens, and also `Field`).
2. `simpleParsec` will go away, we will always have some Position.

ezyang: IIRC, some of the parsers are also used to parse command line flags. But I guess you are right it is the data structures that I am thinking of, not the tokens themselves.

Cabal/Distribution/Lexer.x
dcoutts: Yeah, I wrote the initial version some years ago (alex lexer + parsec parser). Not really based on anything except the old parser plus lots of experiments with parsing all the existing hackage .cabal files.

Cabal/Cabal.cabal
```diff
@@ -232,6 +232,10 @@ flag old-directory
   description:  Use directory < 1.2 and old-time
   default:      False
 
+flag parsec-debug
```
ezyang: Would be nice if there were a little comment saying what this does.

Cabal/Cabal.cabal
```diff
@@ -459,3 +485,21 @@ test-suite package-tests
   ghc-options: -Wall -rtsopts
   default-extensions: CPP
   default-language: Haskell98
+
+test-suite parser-tests
+  if !flag(parsec-debug)
+    buildable: False
```
ezyang: When I `cabal new-build Cabal` in our current tooling, do you know if the solver toggles `parsec-debug` on or off by  default? If it is off by default I am a bit annoyed, because then if I want to run parser tests I have to rebuild the library.

ezyang:
Don't forget to add parser-tests to Travis (it's all manual right now, sorry!)

Cabal/Cabal.cabal
```diff
@@ -251,11 +257,20 @@ library
     build-depends: directory >= 1.2 && < 1.3,
                    process   >= 1.1.0.2  && < 1.5
 
+  build-tools:
```
ezyang: alex needs to be manually installed in Travis since Cabal is too dumb to install build tools right now.

Cabal/Distribution/Lexer.x
```diff
@@ -0,0 +1,244 @@
...
+  $spacetab+; --TODO prevent or record leading tabs
+  $field_braces' $field_braces*    { toki TokFieldLine }
+  \{                { tok  OpenBrace  }
+  \}                { tok  CloseBrace }
+  @nl               { \_ _ _ -> adjustPos retPos >> setStartCode bol_field_braces >> lexToken }
+}
+
+{
+
+data Token = TokSym   !ByteString
```
ezyang: Sigh... I would love docs, but I also understand there is a long and venerated tradition of totally underdocumented lexer implementations.

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,330 @@
...
+
+import qualified Text.Parsec as Parsec
+
+-- ---------------------------------------------------------------
+-- Parsing
+
+-- | Given a parser and a filename, return the parse of the file,
+-- after checking if the file exists.
+readAndParseFile
+    :: (BS.ByteString -> ParseResult a)
```
ezyang: NB: Normally Verbosity comes first but here the parser comes first so that you can partially apply it.

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,330 @@
...
+    -> IO a
+readAndParseFile parser verbosity fpath = do
+    exists <- doesFileExist fpath
+    unless exists
+        (die $ "Error Parsing: file \"" ++ fpath ++ "\" doesn't exist. Cannot continue.")
+    bs <- BS.readFile fpath
+    case parser bs of
+        PR ws (Left errs) -> do
+            mapM_ (warn verbosity . showPWarning fpath) $ reverse ws
+            -- | TODO: die properly
```
ezyang: Is this TODO still applicable?

Cabal/Cabal.cabal
23Skidoo: `parsec-debug` is `default: False, manual: True`, so I'd be surprised if it was on by default.

Cabal/Cabal.cabal
phadej: I'll document it after I know myself :P

Cabal/Cabal.cabal
phadej: I'm not sure whether those test would survive in their current form (i.e. comparison to old parser doesn't make sense when there aren't old parser)

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: `readAndParseFile` isn't exported. Even it has a haddock :)

I'll reorganise the file, so it would be more obvious.

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: I plan to make parser accumulate all errors it could, so we'll need to print them all and die in different way.

phadej:
Btw this will force GHC to bundle mtl, text and transformers. @bgamari said it's doable, and we have to do that before this can be merged

Cabal/Cabal.cabal
phadej: And appveyor.. and osx... is there binaries for those platforms?

Cabal/Distribution/Lexer.x
dcoutts: Right if possible the lexer or parser should give us a warning if it encounters a BOM.

dcoutts:
> Btw this will force GHC to bundle mtl, text and transformers. @bgamari said it's doable, and we have to do that before this can be merged

But at least the ghc lib package doesn't have to depend on those :-)

Cabal/Distribution/Lexer.x
phadej: It is (we warn about tabs already)

phadej:
uh. `StructDiff` really helps. Seems my `Version` parser is broken :P

```diff
packageDescription package pkgVersion versionBranch : 
- [0,10]
+ [0,1]
packageDescription license : 
- UnspecifiedLicense
+ BSD3
packageDescription licenseFiles : 
- []
+ ["LICENSE"]
packageDescription maintainer : 
- ""
+ "danburton.email@gmail.com"
packageDescription author : 
- ""
+ "Dan Burton"
packageDescription synopsis : 
- ""
+ "Combinators for unorthodox function composition"
packageDescription category : 
- ""
+ "Data"
packageDescription specVersionRaw Right : 
- AnyVersion
+ UnionVersionRanges
packageDescription buildType : 
- Nothing
+ Just
condLibrary : 
- Nothing
+ Just
```

phadej:
@dcoutts, @ezyang, @dcoutts now the parser can parse very simple cabal files:

The `parser-tests` results into:
```
::: composition/0.1/composition.cabal
::: composition/0.2/composition.cabal
::: composition/0.2.1/composition.cabal
::: composition/1.0.0.0/composition.cabal
PError (Position 20 1) "Unknown section: \"source-repository\""
```

So now here is "working" proof-of-concept. Please review the design, and comment, before I'll do the rest of bulk work.

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,364 @@
...
+-}
+
+parseConfVar :: [SectionArg Position] -> ParseResult ConfVar
+parseConfVar = error "parseConfVar"
+
+-- | Parse a list of fields, given a list of field descriptions,
+--   a structure to accumulate the parsed fields, and a function
+--   that can decide what to do with fields which don't match any
+--   of the field descriptions.
+parseFields :: [FieldDescr a]      -- ^ descriptions of fields we know how to
```
phadej: unused, and will be removed actually (i.e. not implemented)

phadej: this will be useful for sections without `if-else-blocks`, i.e. `source-repository`

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,364 @@
...
+    -> UnrecFieldParser a  -- ^ How to parse unknown fields
+    -> (a -> c)            -- ^ Condition extractor
+    -> a                   -- ^ Initial value
+    -> [Field Position]    -- ^ Fields to parse
+    -> ParseResult (CondTree ConfVar c a)
+parseCondTree descs unread cond ini fields = do
+    x <- foldM go ini fields
+    return $ CondNode x (cond x) mempty
+  where
+    -- | TODO: change to take and return condnode
```
phadej: or `Pair a [CondTree ConfVar c a])` actually.

Cabal/Distribution/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,572 @@
...
+
+-- | TODO: remove position
+syntaxError :: Position -> String -> ParsecBS a
+syntaxError n s = fail s
+
+-------------------------------------------------------------------------------
+-- Unrecoginsed fields
+-------------------------------------------------------------------------------
+
+type UnrecFieldParser a = (BS.ByteString, [FieldLine Position]) -> a -> Maybe a
```
phadej: First parameter will be `FieldName`

Cabal/Cabal.cabal
ezyang: Well, `cabal install alex` will do the trick in a pinch. Won't even be that slow if cached. hvr's PPA has the stuff for Linux though.

ezyang:
Thanks @phadej, this looks great. Are there any particular parts which you were uncertain how to proceed and specifically want feedback?

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,364 @@
...
+            -> ParseResult a
+parseFields descrs unrec ini fields = return ini -- TODO!
+
+-- TODO: build-depends is a special case
+-- TODO: add post check
+parseCondTree
+    :: forall a c.
+       [FieldDescr a]      -- ^ Field descriptions
+    -> UnrecFieldParser a  -- ^ How to parse unknown fields
+    -> (a -> c)            -- ^ Condition extractor
```
phadej: This is a trick, as we parse `build-depends` for the section `a` anyway, we can then extract it for `CondTree`, is this ok?

Cabal/Distribution/Parsec/Class.hs
```diff
@@ -0,0 +1,254 @@
...
+import Distribution.Text (display)
+
+-------------------------------------------------------------------------------
+-- Class
+-------------------------------------------------------------------------------
+
+-- |
+--
+-- TODO: implementation details: should be careful about consuming trailing whitespace?
+-- Should we always consume it?
```
phadej: I'd like comment on this

phadej: For now `Parsec` class (is naming good btw?) is used only to parse fields, but it could (will?) be used for command line args as well, if we get rid of `ReadP` completely

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,52 @@
...
+
+import Distribution.Parsec.Types.FieldDescr
+import Distribution.Parsec.Types.Common (PError (..), PWarning (..), Position (..))
+import Distribution.Parsec.Class (ParsecBS)
+
+import qualified Data.ByteString as BS
+import qualified Text.Parsec as P
+
+-- | TODO: we could accumulate errors as well
+data ParseResult a = PR [PWarning] (Either PError a)
```
phadej: One idea would be to use `= PR [PWarning] [PError] (Maybe a)`, so we could accumulate "non-fatal" errors, as field parsing is quite independent.

phadej:
@ezyang I commented on some things which I'd definitely like some comments. Also you could go thru `File changed` searching for `TODO`, there are some questions as well.

Cabal/Distribution/Parsec/Types/ParseResult.hs
phadej: ... or strict state variant with better monoid if we care about performance a lot.

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: OK, well, put that in the comment, and have a reference to the function for parsing sections that *are* recursive. (Confusingly, UnrecFieldParser = unrecognized, not unrecursive!)

Cabal/Distribution/Parsec/Types/Field.hs
```diff
@@ -0,0 +1,70 @@
...
+import Data.ByteString (ByteString)
+
+import qualified Data.Char as Char
+import qualified Data.ByteString.Char8 as B
+
+-------------------------------------------------------------------------------
+-- Cabal file
+-------------------------------------------------------------------------------
+
+data Field ann
```
ezyang: So, I guess the convention is we associate the annotation with the name, except when there isn't a name?

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,364 @@
...
+            -> [Field Position]    -- ^ fields to be parsed
+            -> ParseResult a
+parseFields descrs unrec ini fields = return ini -- TODO!
+
+-- TODO: build-depends is a special case
+-- TODO: add post check
+parseCondTree
+    :: forall a c.
+       [FieldDescr a]      -- ^ Field descriptions
+    -> UnrecFieldParser a  -- ^ How to parse unknown fields
```
ezyang: Since ParseResult is our own monad, we could feed in the UnrecFieldParser as a reader field.

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,364 @@
...
+        return x
+    go x (Field (Name pos name) fieldLines) =
+        case Map.lookup name fieldParsers of
+            Nothing -> do
+                parseWarning pos $ "Unknown field: " ++ show name
+                return x
+            Just parser ->
+                runFieldParser (parser x) fieldLines
+
+    fieldParsers :: Map FieldName (a -> FieldParser a)
```
ezyang: Dumb question why is it a -> FieldParser a

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: Sorry, I don't understand this comment.

ezyang: Actually, I think I understand this a little more: the point is the conditional is embedded INSIDE the parsed structure. This is very strange and deserves quite a bit more explanation. The depends in the CondTree and in the actual struct are redundant. I wonder if this means that we should parse the build-depends normally, and then have a post-process step which turns them into Cond nodes. I guess it's OK to do it on the fly, but it needs to be explained, e.g., with an example saying how a cabal file parsers into a CondTree.

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: Whatever the most precise type is :)

Cabal/Distribution/Parsec/Class.hs
ezyang: Normally the tokenizer is responsible for removing whitespace / converting layout sensitivity into braces. If at all possible I would like this to be the case here. But I guess I missed what the Alex file is for!

Cabal/Distribution/Parsec/FieldDescr.hs
ezyang: Type synonym me :)

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: the `CondTree` isn't redundant, the `ConfVar` part is still required (the actual condition isn't embedded)

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: because we accumulate the structure. Alternatively we could have `exists f. (FieldParser f, f -> a -> a)`

Cabal/Distribution/Parsec/Class.hs
phadej: alex parsers only the outer structure of the cabal file, not the actual field contents (think json)

Cabal/Distribution/Parsec/Types/ParseResult.hs
ezyang: I think the Writer monad stack leak is fairly severe and we should strictify it. See http://stackoverflow.com/questions/25749126/a-stricter-control-monad-trans-writer-strict

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: Just goes to show how poorly I understand how `CondTree` works :)

Cabal/Distribution/Parsec/Types/Field.hs
phadej: Have to add note: it might feel that we don't need `IfElseBlock`, but it's separate form section as there could be follow-up "else". OTOH we could handle those in actual cabal file parser.

I'm not sure which solution is clearer. E.g. for config or BuildHooksInfo `IfElseBlock` is fully unnecessary, but we cannot know that statically...

Cabal/Distribution/Parsec/Class.hs
ezyang: In that case, I'm sorry to say I don't have any good guidance in this case. :(

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: `UnrecFieldParser` is different per each `a`, and it isn't passed "down".

Cabal/Distribution/Parsec/Types/Common.hs
```diff
@@ -0,0 +1,64 @@
...
+ --
+showPWarning :: FilePath -> PWarning -> String
+showPWarning fpath (PWarning pos msg) =
+  normalise fpath ++ ": " ++ show pos ++ msg
+
+-------------------------------------------------------------------------------
+-- Field parser
+-------------------------------------------------------------------------------
+
+data FieldParser a
```
ezyang: Would love a comment here.

Cabal/Distribution/Parsec/Class.hs
phadej: This is also something for me to consider. Some parsers are easier to write so they consume trailing whitespace, some aren't. Luckily they are pretty well isolated into this file, so it's easy to change

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: I'd love it if there was an explanation somewhere of why it would be really painful to do the parsers in a non-accumulating style. I think the reason is because you don't know a priori what all the fields are going to be, so a parser which returns just the value (rather than field: value pair) is just not that useful.

Cabal/Distribution/Parsec/Types/Common.hs
phadej: Would something like:
> `FieldParser` is a variant of  `Stream s Identity Char => Parsec s [PWarning] a` restricted to either ByteString (ascii) or String (utf8) streams

do?

ezyang: OK, so that tells me the what--but why?! :)

Cabal/Distribution/PackageDescription/Parsec.hs
ezyang: Oh if it's not propagated, then there's no point. Never mind! (I just kept seeing it in the signatures)

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: That: almost all fields are optional (in `~/.cabal/config` or `cabal.project` actually *all* fields are optional)
But more importantly, in this formulation it's easier to have close to "bi-directional" declarative spec of the fields.

This variant is quite similar to what already was in `Cabal`, so I went with it, as I'm sure this will work out.

ezyang:
It seems to me that the technical risk associated with this refactoring is low. :+1:

Cabal/Distribution/Parsec/Types/Common.hs
phadej: Because for some fields it isn't necessary to decode utf8. At this point I'm not sure though how much non-utf8 fields we have, so might be that this becomes a type synonym.

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: also there is `license-file` / `license-files` issue, i.e. two surface syntax fields mapping to one internal one.

phadej: FWIW, I'll write those comments done as notes in code.

Cabal/Distribution/PackageDescription/Parsec.hs
23Skidoo: Bikeshedding: s/Unrec/Unkn or Unknown/.

phadej:
```
package-tests: could not build shared Setup executable
  ran: '/opt/ghc/7.4.2/bin/ghc-7.4.2' '--make' '-package-conf=/home/travis/build/haskell/cabal/dist-newstyle/build/Cabal-1.25.0.0/package.conf.inplace' '-hide-package Cabal' '-package Cabal-1.25.0.0' '-O0' 'Setup.hs'
  output:
<command line>: cannot satisfy -package Cabal-1.25.0.0
    (use -v for more information)
```

travis fails. Why this could happen?

23Skidoo:
Well, try using `-v`, maybe it gives a clue.

ezyang:
GHC can't satisfy Cabal's parsec dependency, because parsec was installed to the Nix store, but the `package-test` runner assumes that `Cabal` is well-formed using only the inplace database. So I guess we have to make the invocation here more robust. `CABAL_PACKAGETESTS_DB_STACK` is probably of interest.

phadej:
I did quite a lot during weekend. At this point I'm able to parse `compo*` package files. First step `if-else-blocks`, after that parse will be able to in working state.

Modules start to be in their final shape also, I'm not expecting to move things around anymore.

phadej:
Now this is able to parse the whole of Hackage (agrees with readp-parser), except quite short list in `ParserTest.hs`

Performance isn't super fast compared `readp`, but at least not less!
```
% time dist/build/parser-tests/parser-tests parse-readp
dist/build/parser-tests/parser-tests parse-readp  105.50s user 3.12s system 98% cpu 1:49.73 total
% time dist/build/parser-tests/parser-tests parse-parsec
dist/build/parser-tests/parser-tests parse-parsec  79.06s user 1.31s system 99% cpu 1:20.93 total
```

Next steps:
- verify that we emit at least as much warnings as readp
- see if we can do something to failing cases

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -185,7 +187,8 @@ parseGenericPackageDescription' fs = do
         | name == "executable" = do
             -- TODO: make a combinator for this, repeated in tests and bench
             name' <- parseName pos args
-            exe <- parseCondTree executableFieldDescrs storeXFieldsExe (targetBuildDepends . buildInfo) emptyExecutable fields
+            -- Note: we don't parse the "executable" field here, hence the tail hack. Duncan 2010
```
phadej: @dcoutts the same note is in `D.PD.Parse.hs` (from 2010), I have no idea why `executable:` field is dropped?

ezyang:
Sweet. Any chance we can get Hackage parsing in CI? (Maybe in the bootstrap test?)

phadej:
Well, Hackage-parsing test can be done separately from this (if someone have spare time, please do). 
I'd prefer separate job, as handling index file needs `tar`

phadej:
I think I'll remove `IfElseBlock` from `Field`, it's not hard to deal directly in the later stage of the parser, would be simpler to keep outline parser stable. 

IIRC original @dcoutts version doesn't had it. I introduced it to test `readFields` implementation last year.

Any thoughts on that?

phadej:
```hs
readp warnings: 2616
parsec count:   1076
 - PWTGluedOperators : 6
 - PWTVersionTag : 53
 - PWTInvalidSubsection : 256
 - PWTUnknownField : 761
```

ezyang:
Let us know when you want us to review this again!

Cabal/Cabal.cabal
```diff
@@ -1,5 +1,5 @@
 name:          Cabal
-version:       1.25.0.0
+version:       2.0.0.0
```
23Skidoo: This is usually only done when the release branch is created. If you want to disambiguate between your branch and `master`, I suggest using something like `1.25.1.0`.

phadej: I want this temporarily, to see if I can force custom setups of test dependencies to be build with system `Cabal`

phadej:
@ezyang if Cabal 2.0 is due in late October, than no. This is not going to be ready then. Hopefully I'll make some progress during the hackathon after Haskell eXchange, but I rather opt-out and not stress with this too much.

ezyang:
OK. Since you are actively working on this I'll slot you in for 2.2, and then we'll bug you about this some more in a year ;)

ezyang:
Are you still blocking on the CI issues? Is there anything the project as a whole can do to help?

23Skidoo:
@phadej told me that this is now ready to go in.

BardurArantsson:
Can it please be rebased + (possibly squashed) first? Given the age of some of these commits it will wreak absolute havoc with the commit graph if it isn't. (Assuming I understand the gitk "commit graph" algorithm, which I may not.)

phadej:
@BardurArantsson it's rebased on Saturday, so I don't think there's anything changed in `master` worth rebasing for the sake of rebasing.

Also commits are also already squashed, two first ones are big, the rest are quite small and separate, which doesn't change a big picture (you can look at them separately). GitHub supports this nowadays quite well:

screen shot 2016-10-10 at 10 38 39

phadej:
Ah it seems to conflict with `travis-install.sh` atm. That's probably *stack CI* changes. Shouldn't be anything major. Will rebase when addressing comment issues.

phadej:
The diff is huge, especially two first commits. These two just add code, don't change anything.

# Cabal.Distribution.Parsec namespace

- `Types/Common.hs` common types: `Position`, `Perror`, `PWarning`
- `Types/Field.hs` the cabal-like file AST types: `Field` etc

- `Types/ParseResult.hs` a parse result for parsers from AST to Haskell types
- `Types/FieldDescr.hs` a helper structure to define parsers (and pretty-printers) for fields to make AST→Haskell parsers.

- `Lexer.x`: alex file for lexer
- `LexerMonad.hs`: helper definition to make a lexer
- `Parser.hs`: Parser for cabal-like AST

- `ConfVar.hs`: `parseConditionConfVar :: [SectionArg Position] -> ParseResult (Condition ConfVar)`
   parsing of the if-else block conditionals


- `Class.hs`: something which can be parsed from a stream of `Char`s
    ```hs
    class Parsec a where                                                            
        parsec :: P.Stream s Identity Char => P.Parsec s [PWarning] a
    ```

# Cabal.Distribution.PackageDescription

- `Parsec.hs` is a parser implementation for ByteString→GenericPackageDescription parser
- `Parsec/FieldDescr.hs` "instances" of `FieldDescr`s for sections in GPD.

# Other changes

- tests: `ParserTests.hs` basically compare `ReadP` and `Parsec` based resutls to each other
  - there is some stuff to mangle outputs so they compare structurally, because new parser is more precise
  - also few files are skipped (there is a list: `problematicFiles`), the reasons are explained. You can see them on the Hackage via: `DSTM/0.1/DSTM.cabal` -> `hackage.haskell.org/package/DSTM-0.1/DSTM.cabal`, i.e. prefix and first slash to dash.

- `StructDiff` / `DiffInstances` with `-fparsec-struct-diff` the test above outputs human readable diff. Useful for debugging.

- travis is changed so there is GHC 8.0.1 -fparsec job.

ping @ezyang

Cabal/Distribution/Parsec/Types/Common.hs
```diff
@@ -0,0 +1,85 @@
...
+
+data PError = PError Position String
+    deriving (Show)
+
+-- | We do classify warnings.
+--
+-- Different application may decide not to show some, or have fatal behaviour on others
+data PWarnType
+    = PWTOther                 -- ^ Unclassified warning
+    | PWTUTF
```
ezyang: -- ^ Invalid UTF encoding

Cabal/Distribution/Parsec/Types/Common.hs
```diff
@@ -0,0 +1,85 @@
...
+
+showPError :: FilePath -> PError -> String
+showPError fpath (PError pos msg) =
+  normalise fpath ++ ":" ++ showPos pos ++ ": " ++ msg
+
+-------------------------------------------------------------------------------
+-- Field parser
+-------------------------------------------------------------------------------
+
+type FieldParser = Parsec.Parsec String [PWarning]
```
ezyang: :: * -> *

Cabal/Distribution/Parsec/Types/Field.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE DeriveFunctor #-}
+module Distribution.Parsec.Types.Field (
```
ezyang: -- | Cabal-like file AST types ('Field', etc.)

Cabal/Distribution/Parsec/Types/Field.hs
```diff
@@ -0,0 +1,68 @@
...
+-- Name
+-------------------------------------------------------------------------------
+
+data Name ann  = Name       !ann !ByteString
+  deriving (Eq, Show, Functor)
+
+mkName :: ann -> ByteString -> Name ann
+mkName ann bs = Name ann (B.map Char.toLower bs)
+
+getName :: Name a -> ByteString
```
ezyang: s/a/ann/

Cabal/Distribution/Parsec/Types/Field.hs
ezyang: -- | A Cabal file consists of a series of fields (@foo: bar@) and sections (@library ...@).

ezyang: Is Field really a good name for this data type? Sections are not really fields.

Cabal/Distribution/Parsec/Types/Field.hs
```diff
@@ -0,0 +1,68 @@
...
+  deriving (Eq, Show, Functor)
+
+fieldAnn :: Field ann -> ann
+fieldAnn (Field (Name ann _) _)     = ann
+fieldAnn (Section (Name ann _) _ _) = ann
+
+data FieldLine ann  = FieldLine  !ann !ByteString
+  deriving (Eq, Show, Functor)
+
+data SectionArg ann
```
ezyang: I don't know what this is or what the constructors are.

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE CPP #-}
+module Distribution.Parsec.Types.ParseResult (
```
ezyang: -- | A parse result type for parsers from AST to Haskell types.

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
...
+    parseWarnings',
+    ) where
+
+import           Distribution.Compat.Prelude
+import           Prelude ()
+
+import           Distribution.Parsec.Types.Common
+                 (PError (..), PWarnType (..), PWarning (..), Position (..))
+
+-- | A monad with failure and accumulating errors
```
ezyang: s/errors/errors and warnings/

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
...
+
+instance Monad ParseResult where
+    return = pure
+    (>>) = (*>)
+    PR m >>= k = PR $ \s -> case m s of
+        (Nothing, s') -> (Nothing, s')
+        (Just x,  s') -> runPR (k x) s'
+
+-- | "Recover" the parse result, so we can proceed parsing.
+-- 'runParseResult' will still result 'Nothing', if there are recorded errors.
```
ezyang: s/still result/still result in/

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
...
+
+-- | A monad with failure and accumulating errors
+newtype ParseResult a = PR { runPR :: PRState -> (Maybe a, PRState) }
+
+data PRState = PRState ![PWarning] ![PError]
+
+emptyPRState :: PRState
+emptyPRState = PRState [] []
+
+runParseResult :: ParseResult a -> ([PWarning], [PError], Maybe a)
```
ezyang: -- | Destruct a 'ParseResult' into the emitted warnings and errors, and possibly the final result if there were no errors.

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
...
+
+parseWarning :: Position -> PWarnType -> String -> ParseResult ()
+parseWarning pos t msg = PR $ \(PRState warns errs) ->
+    (Just (), PRState (PWarning t pos msg : warns) errs)
+
+parseWarnings' :: [PWarning] -> ParseResult ()
+parseWarnings' newWarns = PR $ \(PRState warns errs) ->
+    (Just (), PRState (warns ++ newWarns) errs) 
+
+-- | Add an error, but not fail the parser yet.
```
ezyang: How do I fail the parser.

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,579 @@
...
+-- | This is /almost/ @'many' 'Distribution.Compat.Parsec.anyChar'@, but it
+--
+-- * trims whitespace from ends of the lines,
+--
+-- * converts lines with only single dot into empty line.
+--
+freeTextFieldParser :: FieldParser String
+freeTextFieldParser = dropDotLines <$ Parsec.spaces <*> many Parsec.anyChar
+  where
+    -- http://hackage.haskell.org/package/copilot-cbmc-0.1/copilot-cbmc.cabal
```
ezyang: Example package with dot lines

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,579 @@
...
+    dotToEmpty x | trim' x == "." = ""
+    dotToEmpty x                  = trim x
+
+    trim' = dropWhileEnd (`elem` (" \t" :: String))
+
+-------------------------------------------------------------------------------
+-- PackageDescription
+-------------------------------------------------------------------------------
+
+-- TODO: other-files isn't used in any cabla file on Hackage.
```
ezyang: s/cabla/cabal/

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,579 @@
...
+
+-------------------------------------------------------------------------------
+-- common FieldParsers
+-------------------------------------------------------------------------------
+
+-- | This is /almost/ @'many' 'Distribution.Compat.Parsec.anyChar'@, but it
+--
+-- * trims whitespace from ends of the lines,
+--
+-- * converts lines with only single dot into empty line.
```
phadej: In old parser? I don't know, probably not. @dcoutts ?

Cabal/Distribution/Parsec/Types/ParseResult.hs
```diff
@@ -0,0 +1,83 @@
...
+    (res, PRState warns errs)
+        -- If there are any errors, don't return the result
+        | null errs -> (warns, [], res)
+        | otherwise -> (warns, errs, Nothing)
+
+instance Functor ParseResult where
+    fmap f (PR pr) = PR $ \s -> case pr s of
+        (r, s') -> (fmap f r, s')
+
+-- | Note: this is not a Monad!
```
phadej: Outdated comment!

phadej:
@ezyang GitHub introduced "start review" &mdash; "finish review" process so you could don't flood my email. Please use it!

screen shot 2016-10-12 at 12 45 55

(ALso I guess, I can reply to review when all comments in it are addressed)

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
dcoutts: Don't know for sure in the old parser, you'd just have to try it. If so it'd probably be ".." or ". ."

phadej: Would it make sense to do this cleanup only if it's not `cabal-version >= 2.0`?

Cabal/Distribution/Parsec/ConfVar.hs
```diff
@@ -0,0 +1,139 @@
...
+
+        parseRangeOp (s,f) = P.try (f <$ oper s <*> version)
+        rangeOps = [ ("<",  earlierVersion),
+                     ("<=", orEarlierVersion),
+                     (">",  laterVersion),
+                     (">=", orLaterVersion),
+                     ("^>=", majorBoundVersion),
+                     ("==", thisVersion) ]
+
+    -- numbers are weird: SecArgNum (Position 65 15) "7.6.1"
```
ezyang: What does this comment mean?

Cabal/Distribution/Parsec/ConfVar.hs
```diff
@@ -0,0 +1,139 @@
...
+        _                         -> Nothing
+
+    oper o = tokenPrim $ \case
+        SecArgOther _ o' | o == o' -> Just ()
+        _                          -> Nothing
+
+    parens = P.between (oper "(") (oper ")")
+
+    tokenPrim = P.tokenPrim prettySectionArg updatePosition
+    updatePosition x _ _ = x
```
ezyang: Uh, how does this update position?

Cabal/Distribution/Parsec/Class.hs
```diff
@@ -0,0 +1,373 @@
...
+parsecToken :: P.Stream s Identity Char => P.Parsec s [PWarning] String
+parsecToken = parsecHaskellString <|> (P.munch1 (\x -> not (isSpace x) && x /= ',')  P.<?> "identifier" )
+
+parsecToken' :: P.Stream s Identity Char => P.Parsec s [PWarning] String
+parsecToken' = parsecHaskellString <|> (P.munch1 (not . isSpace) P.<?> "token")
+
+parsecFilePath :: P.Stream s Identity Char => P.Parsec s [PWarning] String
+parsecFilePath = parsecToken
+
+stdParse
```
ezyang: What's a stdParse

Cabal/Distribution/Compat/Parsec.hs
```diff
@@ -0,0 +1,71 @@
...
+    f '3' = Just 3
+    f '4' = Just 4
+    f '5' = Just 5
+    f '6' = Just 6
+    f '7' = Just 7
+    f '8' = Just 8
+    f '9' = Just 9
+    f _   = Nothing
+
+-- | Greedely munch characters while predicate holds.
```
ezyang: Greedily

Cabal/Distribution/Compat/Parsec.hs
```diff
@@ -0,0 +1,71 @@
...
+munch1
+    :: P.Stream s m Char
+    => (Char -> Bool)
+    -> P.ParsecT s u m String
+munch1 = some . P.satisfy
+
+-- | Greedely munch characters while predicate holds.
```
ezyang: Greedily

Cabal/Distribution/Compat/Parsec.hs
ezyang: And requiring at least one character.

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,566 @@
...
+-- Parsing
+
+-- | Helper combinator to do parsing plumbing for files.
+--
+-- Given a parser and a filename, return the parse of the file,
+-- after checking if the file exists.
+--
+-- Argument order is chosen to encourage partial application.
+readAndParseFile
+    :: (BS.ByteString -> ParseResult a)
```
ezyang: What's this ByteString argument

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,566 @@
...
+      where
+        displaySpecVersion (Left version)       = display version
+        displaySpecVersion (Right versionRange) =
+          case asVersionIntervals versionRange of
+            [] {- impossible -}           -> display versionRange
+            ((LowerBound version _, _):_) -> display (orLaterVersion version)
+
+    maybeWarnCabalVersion _ _ = return ()
+
+{-
```
ezyang: Can we delete this code?

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,566 @@
...
+                (P.errorMessages err)
+
+        parseFatalFailure epos $ msg ++ ": " ++ show str
+  where
+    p' = (,) <$ P.spaces <*> p <* P.spaces <* P.eof <*> P.getState
+
+data GPDS = Fields | Sections
+
+-- Note [Accumulating parser]
+parseGenericPackageDescription'
```
ezyang: Following the other naming convention, why isn't this called parsecGenericPackageDescription?

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -0,0 +1,566 @@
...
+        let epos = Position (row - 1 + P.sourceLine ppos) (col - 1 + P.sourceColumn ppos)
+        let msg = P.showErrorMessages
+                "or" "unknown parse error" "expecting" "unexpected" "end of input"
+                (P.errorMessages err)
+
+        parseFatalFailure epos $ msg ++ ": " ++ show str
+  where
+    p' = (,) <$ P.spaces <*> p <* P.spaces <* P.eof <*> P.getState
+
+data GPDS = Fields | Sections
```
ezyang: GPD... state?

ezyang:
It would also be nice if there were cross-reference notes for where code in the new parser is "duplicated" (obviously not direct copy paste, but of sufficient complexity and structured the same way, such that if you change one, you have to change the other.) Until we kill the old parser, of course...

phadej:
@ezyang I'd say that the best "notes" for keeping two parsers up to date is writing tests.

Also except from obvious `FieldDescr`, I don't actually know how old parser does `Field` → `GenericPackageDescription`

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,580 @@
...
+
+
+-------------------------------------------------------------------------------
+-- Utilities
+-------------------------------------------------------------------------------
+
+beginsWithX :: FieldName -> Bool
+beginsWithX bs = case BS.uncons bs of
+    Just (x, _)
+        | x == fromIntegral (ord 'x') -> True
```
phadej: Good catch. Fixed.

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+  do setLexerMode mode; x <- p; setLexerMode in_section; return x
+
+
+-----------------------
+-- Cabal file grammar
+--
+
+-- $grammar
+--
+-- @
```
phadej: added `CabalStyleFile` production, which is hopefully self-explanatory

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+--
+-- > '\n'? '{'
+--
+-- That is, an optional newline (and indent) followed by a @{@ token.
+-- In the @SectionLayoutOrBraces@ case you can see that this makes it
+-- not fully left factored (because @SecElems@ can start with a @\n@).
+-- Fully left factoring here would be ugly, and though we could use a
+-- lookahead of two tokens to resolve the alternatives, we can't
+-- conveniently use Parsec's 'try' here to get a lookahead of only two.
+-- So instead we deal with this case in the lexer by making a line
```
phadej: The code reference would be useful indeed...

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+--                         | line? ('\n' line)*
+fieldLayoutOrBraces :: IndentLevel -> Name Position -> Parser (Field Position)
+fieldLayoutOrBraces ilevel name =
+      (do openBrace
+          ls <- inLexerMode (LexerMode in_field_braces) (many fieldContent)
+          closeBrace
+          return (Field name ls))
+  <|> (inLexerMode (LexerMode in_field_layout)
+        (do l  <- option (FieldLine (Position 0 0) B8.empty) fieldContent
+                  --FIXME ^^ having to add an extra empty here is silly!
```
phadej: fixed with `optionMaybe`

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+-- TODO: For some reason alex parser cannot handle BOM, is it a bug?
+recodeStringUtf8 :: [Word8] -> (Maybe LexWarning, [Word8])
+recodeStringUtf8 (0xef : 0xbb : 0xbf : bytes) =
+    ( Just $ LexWarning LexWarningBOM (Position 1 1) "Byte-order mark found"
+    , encodeStringUtf8 (decodeStringUtf8 bytes)
+    )
+recodeStringUtf8 bytes =
+    (Nothing, encodeStringUtf8 (decodeStringUtf8 bytes))
+
+#ifdef CABAL_PARSEC_DEBUG
```
phadej: IIRC it requires `dbgText` in lexer too. There was some complications. Have to recheck and comment about them,

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+  where
+    parser = do
+        fields <- cabalStyleFile
+        ws     <- getLexerWarnings
+        pure (fields, maybeToList w ++ ws)
+
+    (w, s') = fmap B.pack . recodeStringUtf8 . B.unpack $ s
+    lexSt = mkLexState' (mkLexState s')
+
+-- TODO: For some reason alex parser cannot handle BOM, is it a bug?
```
phadej: done

Cabal/Distribution/Parsec/Parser.hs
```diff
@@ -0,0 +1,393 @@
...
+  where
+    pos       = errorPos perr
+    ls        = lines' (T.decodeUtf8With T.lenientDecode input)
+    errLine   = T.unpack (ls !! (sourceLine pos - 1))
+    indicator = replicate (sourceColumn pos) ' ' ++ "^"
+    errmsg    = showErrorMessages "or" "unknown parse error"
+                                  "expecting" "unexpected" "end of file"
+                                  (errorMessages perr)
+
+lines' :: T.Text -> [T.Text]
```
phadej: done

Cabal/Distribution/Parsec/ConfVar.hs
```diff
@@ -0,0 +1,139 @@
...
+                 orEarlierVersion, orLaterVersion, thisVersion,
+                 unionVersionRanges, withinVersion)
+
+-- | Parse @'Condition' 'ConfVar'@ from section arguments provided by parsec
+-- based outline parser.
+parseConditionConfVar :: [SectionArg Position] -> ParseResult (Condition ConfVar)
+parseConditionConfVar args = do
+    -- Warnings!
+    args' <- preprocess args
+    case P.runParser (parser <* P.eof) () "<condition>" args' of
```
phadej: it's not used, we reformat the error message. Added a comment

Cabal/Distribution/Parsec/ConfVar.hs
```diff
@@ -0,0 +1,139 @@
...
+        Left err -> do
+            let ppos = P.errorPos err
+            let epos = Position (P.sourceLine ppos) (P.sourceColumn ppos)
+            let msg = P.showErrorMessages
+                    "or" "unknown parse error" "expecting" "unexpected" "end of input"
+                    (P.errorMessages err)
+            parseFailure epos msg
+            pure $ Lit True
+
+-- This is a hack, as we have "broken" .cabal files on Hackage
```
phadej: commented

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: commented

Cabal/Distribution/Compat/Parsec.hs
phadej: fixed

Cabal/Distribution/Parsec/Class.hs
phadej: the name duncan came with ~10 years ago :)

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: they are "TODOs", there seems to not be tests for those, so didn't remember to implement

Cabal/Distribution/PackageDescription/Parsec/FieldDescr.hs
```diff
@@ -0,0 +1,580 @@
...
+
+libFieldDescrs :: [FieldDescr Library]
+libFieldDescrs =
+    [ listFieldWithSep vcat "exposed-modules" disp (parsecMaybeQuoted parsec)
+        exposedModules (\mods lib -> lib{exposedModules=mods})
+    , commaListFieldWithSep vcat "reexported-modules" disp parsec
+        reexportedModules (\mods lib -> lib{reexportedModules=mods})
+
+{-
+  , listFieldWithSep vcat "required-signatures" disp parseModuleNameQ
```
ezyang: Well, if we ran the parser tests on the Cabal files in PackageTests it would have been caught.

phadej: We do run them (this is how I caught `autogen-modules` e.g.). Or are backpack package tests in `cabal-install`?

ezyang: Ah, so the problem is we don't have ghc-head tests online yet, and the Backpack tests don't get run unless you are at ghc-head (whereas you want to run the parser on everything; that should work at least!)

phadej:
@ezyang I'd opt to postpone cross commenting (and clarifying alex codes) until we do cabal commentary https://github.com/haskell/cabal/issues/3990

I find it difficult to comment these issues inside the source, as stuff is spread (for good) into separate modules. Also, writing external commentary will definitely make me rethink naming a bit (e.g. what `bol_section` means, and is there a better name).

Cabal/Distribution/Parsec/Types/Field.hs
phadej: Well, that's what the old parser uses as well. I don't think `CabalAST` is much better name, if at all. I'm open to suggestions though

phadej:
@ezyang @dcoutts @hvr I call for (hopefully) final review round!

phadej:
Ok, I'll rebase this and fix CI and merge. Then continue with adding support to `cabal-install` so I can dogfood it myself.

phadej:
appveyor: hackage-security-0.5.2.2 depends on cryptohash-sha256-0.11.100.1 which failed
to install

DemiMarie:
Is there some reason that Parsec was chosen over a generated LALR parser using Happy?

PR #3603, opened 2016-07-23
Check that all internal libraries have a name.

23Skidoo:


ezyang:
Thanks, I had this patch but you beat me to it!

ezyang:
Note that it's actually pretty impossible for this happen unless you have a Custom script, because the parser, if a stanza is missing, will always put it in `library`. So there isn't actually a way for a user to (easily) trigger it.

23Skidoo:
What will happen if there are two non-internal `library` sections?

ezyang:
Presently, parse error, you cannot declare library twice.

23Skidoo:
Would be nice to make unnamed sublibraries non-representable, but it looks like that'd entail too much boilerplate.

PR #3607, opened 2016-07-23
Restore autogenModulesDir to BC-compatible signature

ezyang:
On the request of @thomie, pursuant to https://github.com/ekmett/distributive/issues/17#issuecomment-234711958

But you should still fix your code.


thomie:
Thanks @ezyang!

PR #3615, opened 2016-07-25
doc: Add brief description of setup-depends

bgamari:
Just as it says on the tin.


23Skidoo:
Merged, thanks!

phadej:
related, shouldn't `custom-setup` section be required for new uploads, or at least `cabal-check` notify about it?

PR #3620, opened 2016-07-25
Add more reexported module tests

ezyang:


23Skidoo:
I think you need to update `extra-source-files`. Otherwise LGTM.

ezyang:
Oh that's right... I fixed the bug on my Backpack branch but not on master. I guess these will wait till Backpack!

PR #3624, opened 2016-07-25
Test travis

phadej:
Doesn't fail with osx, thought locally it does...

### DIFF
--- a/travis-bootstrap.sh
+++ b/travis-bootstrap.sh
@@ -42,11 +42,11 @@ timed cabal update
 # issue, and it is easy enough to work around: clean first.
 
 echo Cabal
-(cd Cabal && timed cabal clean)
-(cd Cabal && timed cabal sdist)
-(cd Cabal && timed install_from_tarball)
+(cd Cabal && timed cabal clean) || exit $?
+(cd Cabal && timed cabal sdist) || exit $?
+(cd Cabal && timed install_from_tarball) || exit $?
 
 echo cabal-install
-(cd cabal-install && timed cabal clean)
-(cd cabal-install && timed cabal sdist)
-(cd cabal-install && timed install_from_tarball)
+(cd cabal-install && timed cabal clean) || exit $?
+(cd cabal-install && timed cabal sdist) || exit $?
+(cd cabal-install && timed install_from_tarball) || exit $?
--- a/travis-common.sh
+++ b/travis-common.sh
@@ -9,7 +9,7 @@ CABAL_VERSION="1.25.0.0"
 timed() {
     echo "\$ $*"
     start_time=$(date +%s)
-    $*
+    $* || exit $?
     end_time=$(date +%s)
     duration=$((end_time - start_time))
     echo "$* took $duration seconds."
--- a/travis-meta.sh
+++ b/travis-meta.sh
@@ -11,10 +11,10 @@
 #./Cabal/misc/gen-authors.sh > AUTHORS
 
 # Regenerate the 'extra-source-files' field in Cabal.cabal.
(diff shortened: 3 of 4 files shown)


ezyang:
Maybe you'll have better luck getting a more timely Travis build by asking Travis to do it on your home repo.

phadej:
I can wait.

phadej:
from log https://travis-ci.org/haskell/cabal/jobs/147287449

```sh
...
  RegisterAssumeDepsUpToDate:                   OK (2.17s)
    Initial build and copy
    Register q
    Register p
1 out of 77 tests failed (243.66s)
1 # <-- echo $?
```

Dunno why :/

ezyang:
Your second test is reasonable but `travis-common.sh` is supposed to `set -e`, and my local test shows this is supposed to work.

phadej:
it is. But shell scripting is total mystery to me.

Travis transforms `.travis.yml` parts into shell script, and I *don't* know if there is something top level which might interfere with our build shell scripts.

phadej:
Finally. It fails (only) on osx. I'll remove the revert, and push soon.

travis-bootstrap.sh
```diff
@@ -42,11 +42,11 @@ timed cabal update
 # issue, and it is easy enough to work around: clean first.
 
 echo Cabal
-(cd Cabal && timed cabal clean)
```
23Skidoo: Eww, shell sucks. Wish there was a nicer way to fix this.

phadej: Well, I have no idea why travis osx /bin/sh is /bin/bash, my osx has real `sh`, but I don't want to invest in this too much.

phadej: Alternatively we could use `bash` everywhere and `{ }` syntax, i.e. no subshells. But ah.

I was thinking of writing own shell scripting language just to make CI stuff easier, but the pain and frustration levels haven't risen enough yet.

23Skidoo: Well, you spent enough time on this already, so let's merge this patch in the current form. In the (distant) future maybe we can rewrite these scripts and `bootstrap.sh` in Haskell.

ezyang:
@phadej, thank you for taking time to find a workaround. We need to somehow document why this workaround is needed in the code, less a bash-hacker looks and says, "Why, of course this `||` business is unnecessary, exit codes get inherited."

Indeed, the behavior we see here has nothing to do with Mac OS X using bash to interpret the shell script or not (which is actually quite common.) Look at the following litmus test:

```
#!/bin/bash
set -e
foo() {
    false
}
(cd . && foo)
echo not printed
```

This does not print anything on bash 4.3.46. I have heard that Mac OS X runs a very old version of bash, so it is possible that the syntax that I am using is tickling a bug.

ezyang:
OK, I built a copy of bash-3.2 and verified that it indeed prints "not printed". So basically Mac OS X bash is too old :o)

23Skidoo:
@ezyang Thanks for investigating. I expanded the comment in `travis-script.sh` a bit.

tuncer:
Just to be clear, is the plan to turn this into a bash script (again)?

phadej:
@tuncer no. but on my OSX machine:
```
% /bin/sh --version            
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)
Copyright (C) 2007 Free Software Foundation, Inc.
```

tuncer:
@phadej, the only problem that can cause is that one accidentally uses non-POSIX feature. However, as long as the CI env has a POSIX sh as `/bin/sh`, I wouldn't expect regressions.

Issue #3626, opened 2016-07-26
ghc-paths-0.1.0.9 setup is broken


phadej:
```
cabal: Entering directory '/tmp/cabal-tmp-3198/ghc-paths-0.1.0.9'
[1 of 1] Compiling Main             ( /tmp/cabal-tmp-3198/ghc-paths-0.1.0.9/dist/setup/setup.hs, /tmp/cabal-tmp-3198/ghc-paths-0.1.0.9/dist/setup/Main.o )
/tmp/cabal-tmp-3198/ghc-paths-0.1.0.9/dist/setup/setup.hs:65:7: error:
    • Couldn't match type ‘(Maybe a0, [a1])’
                     with ‘[(ComponentName, BuildInfo)]’
      Expected type: IO HookedBuildInfo
        Actual type: IO (Maybe a0, [a1])
    • In a stmt of a 'do' block: return (Just (read str), [])
      In the expression:
        do { str <- readFile file;
             return (Just (read str), []) }
      In an equation for ‘readHook’:
          readHook _ _
            = do { str <- readFile file;
                   return (Just (read str), []) }
```

Job: https://travis-ci.org/ekmett/bytes/jobs/147342137

phadej:
Regarding this and other recent `setup-depends` & breaking change issues. We should setup a test-bench which tests various different packages with `setup: Custom` across different ghc and cabal-versions (three dimensional matrix).

It should be as "easy" as having a repository with travis-setup which installs all ghc/cabal-install versions (ghc-7.0 &mdash; ghc-head, cabal-install-1.10 &mdash; cabal-install-head) and has a job per package. We should figure out how to deal with common custom `Setup.lhs`

ezyang:
Ah yes, this one. I think I'm going to revert the new feature I wrote to generalize this to work for all components. Seems emphatically not worth it.

phadej:
I setup https://travis-ci.org/phadej/cabal-compatibility/builds/147350764

Seeps that only `new-build` on `1.24` honours `setup-depends`. It's this still the case?

phadej:
ping @ezyang @dcoutts @hvr

ezyang:
This is reverted on HEAD.

PR #3627, opened 2016-07-26
Revert "Generalize HookedBuildInfo to work with any type of component."

ezyang:
This reverts commit af7bb5371a1ac776f93ab075d7e945cefefe0928,
restoring 'HookedBuildInfo' backwards compatibility.


23Skidoo:
Context: #3626, https://github.com/ekmett/bytes/pull/31.

Since this is only a problem with HEAD (future 2.0), do we really need to revert it?

ezyang:
I don't think we should apply this BC-breaking change unless the rest of hooks gets redesigned. There's been muttering about it but I don't see any code. (Fortunately it's pretty easy to revert this revert.)

23Skidoo:
OK.

Issue #3623, opened 2016-07-25
Cabal/PackageTests/Configure fails on osx, but travis don't.


phadej:
See https://travis-ci.org/haskell/cabal/jobs/146958847

reproduced locally, I have `autoreconf` installed

phadej:
Uh. requiring `X` is a bit overkill.

ezyang:
The test uses X because that was the example in documentation https://www.haskell.org/cabal/users-guide/developing-packages.html#system-dependent-parameters

But there isn't any reason it can't be changed to something more portable. Suggestions?

phadej:
`zlib` comes to mind

23Skidoo:
`zilb-dev` is usually not installed by default.

phadej:
Well, it's needed to build Haskell `zlib`, so it shouldn't be far away.

23Skidoo:
Yeah, you're right, `zlib-dev` is probably installed by default on Travis given that we don't install it ourselves and `cabal-install` depends on `zlib`.

ezyang:
Fixed.

PR #3629, opened 2016-07-26
PackageTests/Configure: Use zlib instead of X11.

23Skidoo:
Fixes (part of) #3623.


phadej:
Thanks, LGTM :+1:

ezyang:
Yep, let's put it in.

phadej:
duh. Now I can verify that osx test would actually fail when there is something failing...

23Skidoo:
@phadej You revert this patch in your branch.

phadej:
have to do that

PR #3632, opened 2016-07-27
Don't test sdist in the main script.

23Skidoo:
See https://github.com/haskell/cabal/pull/3630#issuecomment-235438329.


ezyang:
:+1:

PR #3639, opened 2016-07-28
Don't pass -package-db and -package flags to --abi-hash.

ezyang:


23Skidoo:
@ezyang Let's merge this as well.

PR #3643, opened 2016-07-29
Add a 'status' command.

23Skidoo:
#2882, rebased on top of `master`. Also moved the branch to the main repo so that others could push to it.

Current status: needs more testing, there are apparently some bugs, e.g. https://github.com/haskell/cabal/pull/2882#issuecomment-203911073.

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -44,7 +44,6 @@ module Distribution.Simple.Configure (configure,
                                       getInstalledPackages,
                                       getInstalledPackagesMonitorFiles,
                                       getPackageDBContents,
-                                      checkPackageDBs,
                                       configCompiler, configCompilerAux,
                                       configCompilerEx, configCompilerAuxEx,
                                       ccLdOptionsBuildInfo,
@@ -1268,29 +1267,6 @@ getInstalledPackages verbosity comp packageDBs progdb = do
     HaskellSuite {} ->
       HaskellSuite.getInstalledPackages verbosity packageDBs progdb
     flv -> die $ "don't know how to find the installed packages for "
-              ++ display flv
-
--- | Check the consistency of the given package databases.
-checkPackageDBs :: Verbosity -> Compiler
-                -> PackageDBStack -- ^ The stack of package databases.
-                -> ProgramConfiguration
-                -> IO [(PackageDB, [String])]
-checkPackageDBs verbosity comp packageDBs progconf = do
-  when (null packageDBs) $
-    die $ "No package databases have been specified. If you use "
-       ++ "--package-db=clear, you must follow it with --package-db= "
-       ++ "with 'global', 'user' or a specific file."
-
-  debug verbosity "checking package-db..."
-  case compilerFlavor comp of
-    GHC   -> GHC.checkPackageDBs verbosity comp packageDBs progconf
-    -- GHCJS -> GHCJS.checkPackageDBs verbosity packageDBs progconf
-    -- JHC   -> JHC.checkPackageDBs verbosity packageDBs progconf
-    -- LHC   -> LHC.checkPackageDBs verbosity packageDBs progconf
-    -- UHC   -> UHC.checkPackageDBs verbosity comp packageDBs progconf
-    -- HaskellSuite {} ->
-    --   HaskellSuite.checkPackageDBs verbosity packageDBs progconf
-    flv -> die $ "don't know how to check the packages database for "
               ++ display flv
 
 -- | Like 'getInstalledPackages', but for a single package DB.
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -44,6 +44,7 @@ module Distribution.Simple.Configure (configure,
                                       getInstalledPackages,
                                       getInstalledPackagesMonitorFiles,
                                       getPackageDBContents,
+                                      checkPackageDBs,
                                       configCompiler, configCompilerAux,
                                       configCompilerEx, configCompilerAuxEx,
                                       ccLdOptionsBuildInfo,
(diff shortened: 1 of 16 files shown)


Cabal/Distribution/Simple/GHC.hs
```diff
@@ -297,6 +298,18 @@ getInstalledPackages verbosity comp packagedbs conf = do
...
+checkPackageDBs :: Verbosity -> Compiler -> PackageDBStack
+                -> ProgramConfiguration
+                -> IO [(PackageDB, [String])]
+checkPackageDBs verbosity comp packagedbs conf = do
+  checkPackageDbEnvVar
+  checkPackageDbStack comp packagedbs
+  sequenceA
+    [ do strs <- HcPkg.check (hcPkgInfo conf) verbosity packagedb
+         return (packagedb, strs)
+    | packagedb <- packagedbs ]
```
ezyang: A bit surprised this works, since I thought ghc-pkg needs to know about all the package databases under you to check validity (otherwise some package references could look dangling.)

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -42,6 +42,7 @@ module Distribution.Simple.Configure (configure,
                                       getInstalledPackages,
                                       getInstalledPackagesMonitorFiles,
                                       getPackageDBContents,
+                                      checkPackageDBs,
```
ezyang: Bikeshedding: I don't think there is any particular reason for this function to live in this module.

Cabal/doc/installing-packages.markdown
```diff
@@ -349,6 +349,33 @@ $ cabal --ignore-sandbox install text
...
+Flags can be used to print specific items only; `cabal status --all` will
+print the full summary.
+
+Example:
+
+~~~~~~~~~~~~~~~
+$ cabal status --compiler
+Configured compiler:
+  ghc-7.10.3
+~~~~~~~~~~~~~~~
```
ezyang: I'd add a note that the output is NOT intended to be machine parseable, and should not be considered stable between versions.

cabal-install/Distribution/Client/Status.hs
```diff
@@ -0,0 +1,417 @@
...
+
+status :: Verbosity -> GlobalFlags -> StatusFlags -> IO ()
+status verbosity globalFlags statusFlags = do
+  (useSandbox, config) <- loadConfigOrSandboxConfig
+    verbosity
+    (globalFlags { globalRequireSandbox = Cabal.Flag False })
+  (defaultComp, platform, defaultConf) <- configCompilerAux'
+                                        $ savedConfigureFlags config
+  let distPref = useDistPref defaultSetupScriptOptions
+  -- TODO: this currently only works if the cabal version used for configuring
```
ezyang: This TODO is actually quite difficult to fix. If the LBI is from a different version of Cabal library, as it might be in the case of a Custom setup script, we can't actually get our grubby hands on it. This problem plagues package-tests. Probably the only way to fix it is to have a different program which links against the relevant version of Cabal.

cabal-install/Distribution/Client/Status.hs
```diff
@@ -0,0 +1,417 @@
...
+  sandboxConfigExists <- doesFileExist sandboxConfigPath
+  (sandboxPath, sandboxDbPath) <- if sandboxConfigExists
+    then do
+      (sandboxPath, _pkgEnv) <- tryLoadSandboxConfig verbosity globalFlags
+      return $ ( Just sandboxPath
+               , Just $ sandboxPackageDBPath sandboxPath comp platform)
+    else return (Nothing, Nothing)
+  let dbs :: [PackageDB] -- TODO: it might be better to inspect the currently
+                         -- configured dbs instead of a static global,local,..
+      dbs = GlobalPackageDB
```
ezyang: Yes, use `withPackageDB buildConfig` here.

ezyang:
I think it would be far better for this program to quit if it can't read the LBI, rather than give possibly incorrect behavior. Better for a tool to give no info rather than bad info.

I'm not sure why the compiler output is broken. (Too bad there's no tests ;) If I were SPJ, I'd check in some munged output of this to make sure that when we change the output it's acceptable. Need an 'accept' command though.)

I'm very sorry that this (and the other PRs) died before we changed the branch access policy (you can see in the previous discussion @lspitzner asking if he can have write bits to maintainer this! Obviously he does now.)

Cabal/Distribution/Simple/Configure.hs
23Skidoo: Where do you think it should live?

ezyang: Well, a safe place to put it is cabal-install; this way it is not publically exported and can be refactored at will. If it's discovered that it's needed elsewhere we can subsequently move it.

23Skidoo:
> I think it would be far better for this program to quit if it can't read the LBI, rather than give possibly incorrect behavior. Better for a tool to give no info rather than bad info.

+1. Unreadable LBIs are a bottomless source of bugs and breakage.

23Skidoo:
Rebased, addressed all of @ezyang's comments except https://github.com/haskell/cabal/pull/3643#discussion_r72832894.

23Skidoo:
Rebased.

Mikolaj:
@23Skidoo: hi! what are your thought on the future of this command and PR? Would you like to work on that or perhaps comment on the design of a possible new version, drawing from many years of your experience with newer cabals?

jneira:
this looks like the `ide` command @fendor is working on here https://github.com/haskell/cabal/pull/7500
the pov of @23Skidoo would be invaluable there but I would close this if the ide command is merged finally

23Skidoo:
Feel free to close, I don't have the bandwidth to work on this at the moment.

jneira:
> Feel free to close, I don't have the bandwidth to work on this at the moment.

sad to read that, btw many many thanks for the huge amount of hard work in all areas you have put in this project. Your mood has helped to make progress in many issues an prs.

23Skidoo:
thanks!

PR #3644, opened 2016-07-30
One-component configure

ezyang:
Described in: https://github.com/ghc-proposals/ghc-proposals/pull/4

./Setup configure now takes an argument to specify a specific
component name that should solely be configured.

Most of the gyrations in Configure are all about making it so that
we can feed in internal dependencies via --dependency.  I dropped
the package name match sanity check to handle convenience library
package name munging.  I also had to make register/copy unconditionally
install internal libraries; otherwise you can't refer to them
from later builds.

Also a miscellaneous refactor: convenience libraries are printed with a
"header" stanza now (not really a stanza header).

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -94,6 +94,17 @@ extra-source-files:
   tests/PackageTests/Configure/include/HsZlibConfig.h.in
   tests/PackageTests/Configure/zlib.buildinfo.in
   tests/PackageTests/Configure/zlib.cabal
+  tests/PackageTests/ConfigureComponent/Exe/Bad.hs
+  tests/PackageTests/ConfigureComponent/Exe/Exe.cabal
+  tests/PackageTests/ConfigureComponent/Exe/Good.hs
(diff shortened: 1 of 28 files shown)


phadej:
As it's in *ghc-proposals*, we have to wait what it says before merging this?

ezyang:
Well, Cabal is not within the ghc-proposals mandate, and even if it were, there is currently no timeline when it will be within the mandate. So yes I would like to get commentary, but I also think we can informally decide when it's good enough to go in.

phadej:
As we added it there, let's wait ~2 weeks (the usual discussion time on proposals in e.g. libraries), otherwise it will look like "look we decided to do this way, now it's already done".

I like the idea of proposal, but as you posted it there, I'd give people opportunity to comment on it. Sorry that it will delay this PR.

ezyang:
Two weeks is fine. I am well used to the rebased branches on top of rebased branches workflow for extend periods of time ;)

ezyang:
Since this commit is rolled up in #3662 I'm going to close this; review that PR instead.

PR #3647, opened 2016-07-31
Undo new-build support for convenience libraries.

### LINKED ISSUE #3637: new-build: Repeating build fails
Repeating a new-build command twice fails on my project with the error message

```
improveInstallPlanWithUpToDatePackages: dep on non lib package
error, called at ./Distribution/Client/ProjectBuilding.hs:358:7 in
main:Distribution.Client.ProjectBuilding
```

the second time I run the build. The command I'm running is

```
cabal new-build cqrs-{core,example,memory,postgresql,testkit}
```

and the repository is current 'master' here: https://github.com/BardurArantsson/cqrs/tree/master

ezyang:
The previous approach I took, though correct, was quite
confusing.  If I refactor InstallPlan to operate on a
per-component basis, then we'll automatically get support
for convenience libraries, which will ultimately cleaner.
(While we won't be able to get rid of support for whole
package installs, it will be safe to assume packages
using convenience libraries also support one-shot
configure.)

I think this may fix #3637.

### DIFF
--- a/cabal-install/Distribution/Client/Install.hs
+++ b/cabal-install/Distribution/Client/Install.hs
@@ -54,13 +54,11 @@ import Control.Applicative
 import Data.Traversable
          ( traverse )
 #endif
-import Control.Exception ( assert )
 import Control.Monad
          ( filterM, forM_, when, unless )
 import System.Directory
          ( getTemporaryDirectory, doesDirectoryExist, doesFileExist,
-           createDirectoryIfMissing, removeFile, renameDirectory,
-           getDirectoryContents )
+           createDirectoryIfMissing, removeFile, renameDirectory )
 import System.FilePath
          ( (</>), (<.>), equalFilePath, takeDirectory )
 import System.IO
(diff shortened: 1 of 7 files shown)


hvr:
Just for the record, this doesn't fix #3637 yet, as with this patch applied to master, and trying to build `exe:cabal` with it:

```
$ cabal new-build exe:cabal
cabal-install-1.25.0.0
rebuildTarget: unexpected package status
error, called at ./Distribution/Client/ProjectBuilding.hs:676:23 in
main:Distribution.Client.ProjectBuilding
```

23Skidoo:
> If I refactor InstallPlan to operate on a
per-component basis, then we'll automatically get support
for convenience libraries, which will ultimately cleaner.

SGTM.

ezyang:
Actually, I am feeling a little regretful about this patch, because component-izing the new-build code was quite involved, and I'm not sure I want to do it again for the old codepath (install). So I think maybe I'll revert half this patch, keeping the old implementation for the legacy codepath (which hopefully is going away soon.)

23Skidoo:
> (which hopefully is going away soon.)

Well, it will still be around in the next release for sure.

ezyang:
Going to redo the patch to only drop this codepath for new-build.

PR #3650, opened 2016-07-31
Few very small refactors

phadej:
Found those while working on parsec parser

### DIFF
--- a/Cabal/Distribution/Types/GenericPackageDescription.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription.hs
@@ -7,6 +7,7 @@
 module Distribution.Types.GenericPackageDescription (
     GenericPackageDescription(..),
     Flag(..),
+    emptyFlag,
     FlagName(..),
     FlagAssignment,
     ConfVar(..),
@@ -63,6 +64,15 @@ data Flag = MkFlag
     deriving (Show, Eq, Typeable, Data, Generic)
 
 instance Binary Flag
+
+-- | A 'Flag' initialized with default parameters.
+emptyFlag :: FlagName -> Flag
+emptyFlag name = MkFlag
+    { flagName        = name
+    , flagDescription = ""
+    , flagDefault     = True
+    , flagManual      = False
+    }
 
 -- | A 'FlagName' is the name of a user-defined configuration flag
 newtype FlagName = FlagName String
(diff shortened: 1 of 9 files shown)


Cabal/Distribution/Types/GenericPackageDescription.hs
```diff
@@ -61,6 +65,9 @@ data Flag = MkFlag
 
 instance Binary Flag
 
+emptyFlag :: FlagName -> Flag
```
ezyang: Needs doc.

ezyang:
LGTM, but fix the build.

Cabal/Distribution/Types/GenericPackageDescription.hs
phadej: Which would be?

> `Flag`  with default parameters.

?

ezyang: Well, when I look at the code, I wonder, "Why is it True and then False?" And then I look at `Flag` and I see that the first is `flagDefault` and `flagManual`. At this point I can infer that the default settings for a flag are for it to be (1) turned on by default and (2) not manual (whatever that means...) The comment should explain something like that?

phadej: would it be clearer to use record syntax for the definition of `emptyFlag`?

phadej: to avoid boolean blindness, if that's what worries you

phadej: done

phadej: ping @ezyang

ezyang: OK. Could we maybe add Haddock like `-- | A 'Flag' initialized with default parameters (enabled by default, not manual, empty description)`?

phadej: I'll not repeat the code in the haddock. Comments are IMHO for explaining *why* definition exists or *how" it's meant to be used, not *what* it is (there is "source" for that).
In this case it should be irrelevant, and I unfortunately don't know why default parameters are what they are (but would rather comment that in code, not in haddock).

So it's just ` -- | A 'Flag' initialized with default parameters .`

Issue #3649, opened 2016-07-31
getShortPathName returns invalid response in Cygwin


NovaDenizen:
I installed Haskell Platform 8.0.1 into `C:\HPlat`, so gcc is at "c:\\HPlat\\8.0.1\\mingw\\bin\\gcc.exe"
```
$ uname -a
CYGWIN_NT-6.0 myhostnameI 2.5.2(0.297/5/3) 2016-06-23 14:29 x86_64 Cygwin

$ cygpath -d "c:\\HPlat\\8.0.1\\mingw\\bin\\gcc.exe"
C:\HPlat\80227D~1.1\mingw\bin\gcc.exe

$ cygpath -l -w 'C:\HPlat\80227D~1.1\mingw\bin\gcc.exe'
C:\HPlat\8.0.1\mingw\bin\gcc.exe

$ cd ~/dl/Cabal-1.24.0.0/
$ stack repl
...much deleted...
...stribution.Version Language.Haskell.Extension> getShortPathName "c:\\HPlat\\8.0.1\\mingw\\bin\\gcc.exe"
"c:\\HPlat\\80227D~1.1"
...stribution.Version Language.Haskell.Extension>:q

$ cmd /version
Microsoft Windows [Version 6.0.6002]
Copyright (c) 2006 Microsoft Corporation.  All rights reserved.

```

ezyang:
CC @23Skidoo who wrote this code.

Since cygpath computes short path correctly, we can look at its source code to see how to do it properly http://cygwin-ports.sourceforge.net/scan-build/utils/report-MWMmDX.html

From GetShortPath docs:

> The path that the lpszLongPath parameter specifies does not have to be a full or long path. The short form can be longer than the specified path.

So that makes the current implementation incorrect in some cases. But after reading the code I don't see anything else obviously wrong.

NovaDenizen:
Found the way to fix it from [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364989.aspx) 

> Passing NULL for this parameter and zero for cchBuffer will always return the required buffer size for a specified lpszLongPath.

NovaDenizen:
#3487 seems related too.

ezyang:
Thanks, I think it is indeed this bug!

ezyang:
I would close this but if we do a point release for 1.24.0.1 we should put this patch in.

23Skidoo:
Already cherry-picked this into 1.24, so closing.

PR #3657, opened 2016-08-01
Now allocates properly sized result buffer.

NovaDenizen:
See #3649


ezyang:
Thanks! I'll wait for Appveyor tests to go green and then I'll merge it.

(I checked to make sure we weren't committing an off-by-one-error. The docs are very confusing in this respect but I believe the code is right: querying for the needed buffer size will return enough space for the null byte, even though the success return does NOT include the null byte. Pshaw!)

ezyang:
@NovaDenizen I just realized that this code could have been made more robust in one slight way: we should check the return value of the second invocation and error if it is greater than the buffer size we passed in (this would have also caught the previous problem, instead of passing a half-constructed buffer back to the user.) Now that the code is right, I guess it's unnecessary though!

23Skidoo:
LGTM. Thanks, @NovaDenizen!

23Skidoo:
Also cherry-picked into 1.24.

PR #3664, opened 2016-08-03
Fix Haddock syntax in cabal-install.

ezyang:
If we were generating Haddock for cabal-install (we're not
currently) these would cause errors.  Make them stop causing
errors.


23Skidoo:
Merged, thanks!

PR #3669, opened 2016-08-04
stack.yaml: add optional Nix dependency zlib

ttuegel:
The external zlib package is required to build the zlib Haskell
library. This change will not affect Stack-using developers who do not
have Nix integration enabled globally. This change will not affect
non-Stack-using developers at all.


snoyberg:
LGTM

23Skidoo:
Merged, thanks!

PR #1046, opened 2012-09-21
Add way to exclude generated modules from sdists

lpotmann:
In one of my projects, I am machine-generating a number of modules, which other parts of the code then import. These need to be specified in the .cabal file, otherwise hard-to-debug linker errors happen. However, then they get included in sdist tarballs, or sdist chokes if they're not present, which is undesirable. This patch lets generated modules be excluded from sdists on the command line with a `--generated-module` option.

(An alternative (and perhaps more robust) approach would be to add a `generated-modules` field to cabal files, but that's only just occurred to me and this patch is sitting here!)

Issue #719, opened 2012-05-24
Support generated modules better


bos:
(Imported from [Trac #729](http://hackage.haskell.org/trac/hackage/ticket/729), reported by @dcoutts on 2010-08-21)

Currently Cabal's simple build system expects that all modules listed in the <tt>exposed-modules</tt> and <tt>other-modules</tt> have some original source file (possibly one that needs a preprocessor). There is a built-in hack for the generated Paths_pkgname module.

Adding other generated modules is rather tricky because of the assumption that an original source file can be found. This should not be too hard to fix for build, since we can just look for the generated source file in the build location, but it affects sdist more severely. There is no way for sdist to know if a file is generated or not, so if a source file is not found then currently it just fails. Unlike at build time, it's reasonable to run sdist in a clean build tree.

People can hack around it in the Setup.hs by filtering out the module name for various stages, but that is still tricky due to the combination of #410 and #627.

This affects some core ghc components and also darcs.

fmaste:
Closed with #3656 and #3670

PR #3670, opened 2016-08-05
Filter all autogen modules

### LINKED ISSUE #3656: Filter all autogen modules / c-sources
Currently when calling sdist the autogenerated module Paths_\* is ignored, this is a problem for packages with built type custom that generate other modules like this one because you get "Error: Could not find module: ..."

My quick fix was just not using sdist because I don't need to, not really an elegant solution but it worked until I added my package as source to a sandbox and discovered that when building on the sandbox sdist is called on my package!

My proposed solution is to list all modules on the autogen directory and filter them from exposed-modules and other-modules, not just Paths_\* like function filterAutogenModule on Distribution.Simple.SrcDist does. This is what I'm actually doing with a hook on my setup script.

fmaste:
This fixes #3656. From #719 this was already a problem and the idea of creating a new field was also discussed on #1046, so I decided to give it a try

I created a field 'autogen-modules' to list modules that are autogenerated, like Paths_*. This should contain modules that are either on 'exposed-modules' or 'other-modules' so we can filter those when needed


23Skidoo:
Can you please rebase this against master to get rid of the spurious merge commits? Also needs tests, docs, and a line in the changelog.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1110,6 +1139,29 @@ checkCabalVersion pkg =
         ++ "that specifies the dependencies of the Setup.hs script itself. "
         ++ "The 'setup-depends' field uses the same syntax as 'build-depends', "
         ++ "so a simple example would be 'setup-depends: base, Cabal'."
+
+  , check (specVersion pkg >= Version [1,25] []
+           && elem (autogenModuleName pkg) (allModuleNames pkg)
+           && not (elem (autogenModuleName pkg) (allModuleNamesAutogen pkg)) ) $
+      PackageBuildWarning $
+           "Packages using 'cabal-version: >= 1.25' and the autogenerated "
+        ++ "module Paths_* must include it also on the 'autoge-modules' field "
```
23Skidoo: s/autoge/autogen/

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1110,6 +1139,29 @@ checkCabalVersion pkg =
...
+        ++ "the module does not come with the package and is generated on "
+        ++ "setup. Modules built with a custom Setup.hs script also go here "
+        ++ "to ensure that commands like sdist don't fail."
+
+  , check (specVersion pkg < Version [1,25] []
+           && elem (autogenModuleName pkg) (allModuleNames pkg)
+           && not (elem (autogenModuleName pkg) (allModuleNamesAutogen pkg)) ) $
+      PackageDistSuspiciousWarn $
+           "From version 1.25 autogenerated modules are included on the "
+        ++ "'autoge-modules' field besides 'exposed-modules' and "
```
23Skidoo: Ditto.

23Skidoo:
Otherwise LGTM.

ezyang:
This patch looks good, thank you! I think you can reduce the duplication in `Check` by using `allBuildInfo` instead of duplicating the logic for each type of component. See for example `mentionedExtensions` in `Check` for an example that does this. Then you can get rid of the `libModuleAutogen` and variants (which IMO don't add much: just call `buildInfo` if you need it!)

Other notes: in a lot of places, the message reads `autogen-module` when it should read `autogen-modules`.

cabal-install/Distribution/Client/Init.hs
```diff
@@ -889,6 +889,12 @@ generateCabalFile fileName c =
                  Executable -> "Modules included in this executable, other than Main.")
               True
 
+     , fieldS "autogen-modules" (listField (otherModules c'))
```
ezyang: I don't think there's a need to include this field in the default `cabal init`; it's definitely a power-user case.

Cabal/Distribution/Types/TestSuite.hs
```diff
@@ -65,3 +66,11 @@ testModules test = (case testInterface test of
                      TestSuiteLibV09 _ m -> [m]
                      _                   -> [])
                 ++ otherModules (testBuildInfo test)
+
+-- | Get all the auto generated module names from a test suite.
+-- This are a subset of 'testModules'.
+testModulesAutogen :: TestSuite -> [ModuleName]
+testModulesAutogen test = (case testInterface test of
```
ezyang: This doesn't look right to me. In a library test suite, the declared module is generally written explicitly by the user. With this patch, you will stop including this module in the sdist. Bad!

Cabal/Distribution/Types/PackageDescription.hs
```diff
@@ -296,6 +299,18 @@ allBuildInfo pkg_descr = [ bi | lib <- allLibraries pkg_descr
   --FIXME: many of the places where this is used, we actually want to look at
   --       unbuildable bits too, probably need separate functions
 
+-- | As they appear on the cabal file, Paths_* module has no special treatment.
+-- It is not exluded if it already appears here.
```
ezyang: excluded

Cabal/Distribution/Types/PackageDescription.hs
```diff
@@ -296,6 +299,18 @@ allBuildInfo pkg_descr = [ bi | lib <- allLibraries pkg_descr
   --FIXME: many of the places where this is used, we actually want to look at
   --       unbuildable bits too, probably need separate functions
 
+-- | As they appear on the cabal file, Paths_* module has no special treatment.
```
ezyang: @Paths_*@ (so that it gets Haddock formatted correctly). Also, the Haddock should say what the function does, not a general remark about how Paths modules get handled. It's important for a user to know `allModuleNames` applies to all NON-autogenerated modules, this is not obvious without looking at the source. Also this seems to miss LibV09? I'm not sure this function is the right approach. Look at my comment at its call site.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1110,6 +1139,29 @@ checkCabalVersion pkg =
         ++ "that specifies the dependencies of the Setup.hs script itself. "
         ++ "The 'setup-depends' field uses the same syntax as 'build-depends', "
         ++ "so a simple example would be 'setup-depends: base, Cabal'."
+
+  , check (specVersion pkg >= Version [1,25] []
+           && elem (autogenModuleName pkg) (allModuleNames pkg)
```
ezyang: This made me double-take. It would be nice if we can rename `autogenModuleName` to `autogenPathsModuleName` (but keep the old function name around for BC; maybe deprecate it) so that it's clear that this is specifically about the `Paths` module.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -245,6 +246,12 @@ checkLibrary pkg lib =
       PackageDistInexcusable $
            "To use the 'required-signatures' field the package needs to specify "
         ++ "at least 'cabal-version: >= 1.21'."
+
+    -- check that all autogen-modules appear on other-modules or exposed-modules
+  , checkVersion [1,25] (not $ and $ map (flip elem (libModules lib)) (libModulesAutogen lib)) $
```
ezyang: I don't think these tests have the intended semantics you want. What you have here is, "If the Cabal version on your file is 1.25 or later, THEN check that every autogen-modules appears in libModules." That means if the Cabal version is lower, this check gets skipped.

ezyang:
OK, so let me summarize what I think are the blocking changes for this patchset are:

* As @23Skidoo mentions, there needs to be documentation. As part of documenting this, you need to be quite clear about what the intended semantics of `autogen-modules` are. Without introducing an exposed autogen modules field as well as an other autogen modules field, the only way to specify if the autogenerated module is exported or not is by duplicating it in exposed-modules/other-modules. So I think the behavior you want is this: if Cabal < 1.25, then we should ERROR if an autogenerated module is not included in exposed-modules/other-modules/main-is/etc (I think the coverage here in the code is incomplete; the end-all-be-all of what modules may show up is in `Distribution.Simple.SrcDist` and you may have to refactor that code to get at this source of modules.)  Importantly, the `allModuleNames` logic is just incorrect, because, a module name can be duplicated among a library and a test suite (doesn't happen often, but it's allowed!) and one might be autogenerated, while the other not.

* Relatedly, it is a little confusing that you have to specify autogenerated modules twice. No good way around it, as far as I can tell. (You could suggest that if Cabal >= 1.25, an autogen-modules that is not mentioned anywhere implies an other-modules entry, but this might be too much magic. No need to implement.)

* I would *like* it if you reduced the error message duplication in `Check`, but I understand if that's not a priority. Put that in the nice to have bin.

* Renaming `autogenModuleName` would be nice, but not essential for the patch

* Most important: a test showing the whole shebang works!

ezyang:
I gave you commit bits! So when you think the PR is ready and doesn't break anything, you can take pushing it to master into your own hands ;)

Cabal/Distribution/PackageDescription/Check.hs
fmaste: You are right, and also there's no need to check the version because 'autogen-modules' would only parse and contain something when version >= 1.25. I can just remove that version check

fmaste:
Some remarks and questions:

1. Now I return `PackageBuildWarning` without any version check when a module is on `autogen-modules` but not on `other-modules` or `exposed-modules`! **Should it be an error?** A `PackageBuildImpossible`?
2. When `Paths_PACKAGENAME` is mentioned on `other-modules` or `exposed-modules` but not on any `autogen-modules` it's a `PackageBuildWarning` for spec >= 1.25 and `PackageDistSuspiciousWarn` for lower versions with a recommendation to use 1.25. **I think the former is OK but don't know how you usually treat the latter case.** Maybe just an error when the spec version is lower than 1.25 and there's something on `autogen-modules`.
3. Removed `allModuleNames` as recommended.
4. I added function `libModulesAutogen` next to `libModules`, `exeModulesAutogen` next to `exeModules` and so on for `Benchmark` and `TestSuite`. It was mentioned that those are just a recordfield name away and not necessary, but so do `benchmarkModules` and `exeModules`. I think that all these help to be more explicit about the meaning of this new field, really helpful when reading the docs and using Cabal as a library for a custom build.
5. Adding to number 3, I tried to get all this in or from `Distribution.Simple.SrcDist` but the refactor was huge. Started making little changes on this direction but I decided to leave it for a future PR, also most of the functions there reside on IO monad when only with a `PackageDescription -> PackageDescription` suffices.
6. I renamed `autogenModuleName` to `autogenPathsModuleName` and deprecate it.
7. Just the docs missing.

ezyang:
1. It should be an error, unless you plan to implement a feature along the lines of, "If a module is in autogen-modules but not anywhere else, it is implicitly in other-modules." Because if the user commits the error, the build is just not going to work (the autogen module won't get linked in.)
2. So, if I understand your question correctly, you are wondering if we should give a warning at all if someone pre-1.25 mentions `Paths_*` without putting it in `autogen-modules`. If you asked me, I wouldn't warn at all: just keep `Paths_*` as a special-case. But I'm pretty conservative. I think it would also be pretty reasonable to make it a HARD ERROR to say `>= 1.25` and not put paths in autogen-modules. I'm less hot about warning in both cases, because you still have to stick the special case in anyway, so it's not like it's really helping users (besides letting them know about `autogen-modules`)
4. OK
5. OK, I don't blame you. I guess `libModules` is probably a reasonable approximation; it won't catch if you're autogenerating the `main-is` field but I seriously doubt anyone is actually doing that

Awesome work! The patch is really shaping up.

A thought: maybe if you can't find a file in `sdist`, the error message should mention that if the file is autogenerated, you should add it to `autogen-modules`?

Issue #3640, opened 2016-07-28
Apply codensity to Progress


ezyang:
CC @dcoutts, @grayjay, @kosmikus,

The `Progress` type in `Distribution.Client.Dependency.Types` is a great idea and more code in Cabal should use it. However, I am a bit worried about the asymptotic complexity with left associated binds.

Progress is an instance of the free monad over:

```
data ProgressF step fail a = Step step a
                                                | Fail fail
```

As a reminder, the free monad construction is:

```
data Free f done = Free (f (Free f a)) | Done done
```

Inlining `ProgressF step fail` into `f`, we have:

```
Free (ProgressF step fail) done = Step step (Free (ProgressF step fail) done) | Fail fail | Done done
```

which is the current definition of `Progress`.

It is known that free monads have trouble with left associativity, see http://comonad.com/reader/2011/free-monads-for-less/ and http://blog.ezyang.com/2012/01/problem-set-the-codensity-transformation/ so instead we want to apply `Codensity` to avoid this problem.

So, I would like to apply the `Codensity` transformation to `ProgressF`. However, there are some questions:

1. Does `Codensity` retain the incrementality that the original `Progress had?
2. Should we define `ProgressF`, `Free` and `Codensity`, bring in a dependency to get them, or inline all the data type declarations (just as `Progress` is the inlined version of `Free ProgressF`)

grayjay:
That sounds like a good idea, though I'm not very familiar with `Codensity`.  IIUC, it would preserve the incremental behavior.  I'm not sure how it would work with the existing uses of `Progress`, though.  We could split it into two types if necessary. How would you use it in Cabal?

[...]

PR #3674, opened 2016-08-07
Solver: Use difference lists to combine logs in 'Explore.exploreLog'.

grayjay:
This commit adds a data structure, `RetryLog`, which is like a difference list
for the `Progress` type, except that it only supports efficient appends at
failures. Since the solver continually appends logs and calls `tryWith` while
exploring the search tree, it is important for those operations to be efficient.
Afterwards, the solver converts the `RetryLog` back to a `Progress` so that it
can be processed with pattern matching in Log.hs and Message.hs.

I got the idea from #3640 (See https://github.com/haskell/cabal/issues/3640#issuecomment-236391147).  The new log type isn't a monad because bind would have had to manipulate failures instead of successes, which is inconsistent with the monad instance for `Progress`.  I also avoided modifying the `Progress` type, because it is convenient for pattern matching.  This change only affects the solver.

Here are some times.  I tested packages with different numbers of dependencies, because some of the log operations on master are O(N) in the depth of the tree.  I compiled cabal with GHC 8.0.1 and ran `cabal install --dry-run` on each package with GHC 7.10.3 on the path.  Each time is the average of three trials.

package | time (s) - master | time (s) - this branch | max search tree depth | speedup | result
-------------|------------------|------------------------|----------------|------------|-----------
`aeson`      | 1.166 | 1.130 | 71 | 1.032 | success
`phooey` and `hackage-server` | 6.810 | 6.311 | 102 | 1.079 | reached backjump limit
`yesod`      | 1.795 | 1.640 | 334 | 1.095 | success
`leksah`     | 2.756 | 2.551 | 665 | 1.080 | success

/cc @kosmikus @ezyang

### DIFF
--- a/cabal-install/tests/UnitTests/Distribution/Solver/Modular/RetryLog.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Solver/Modular/RetryLog.hs
@@ -1,0 +1,69 @@
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+module UnitTests.Distribution.Solver.Modular.RetryLog (
+  tests
+  ) where
+
+import Distribution.Solver.Modular.Message
+import Distribution.Solver.Modular.RetryLog
+import Distribution.Solver.Types.Progress
+
+import Test.Tasty (TestTree)
+import Test.Tasty.HUnit (testCase, (@?=))
+
+type Log a = Progress a String String
+
+tests :: [TestTree]
+tests = [
+    testCase "convert to and from RetryLog ending in failure" $
+        let lg = Step 1 (Step 2 (Step 3 (Fail "Error")))
+        in toProgress (fromProgress lg) @?= (lg :: Log Int)
+
+  , testCase "convert to and from RetryLog ending in success" $
+        let lg = Step 1 (Step 2 (Step 3 (Done "Result")))
+        in toProgress (fromProgress lg) @?= (lg :: Log Int)
+
+  , testCase "retry with failure" $
+        let log1 = fromProgress $ Step 1 (Step 2 (Fail "Error 1"))
+            log2 = fromProgress $ Step 3 (Step 4 (Fail "Error 2"))
+        in toProgress (retry log1 (const log2))
+           @?= (Step 1 (Step 2 (Step 3 (Step 4 (Fail "Error 2")))) :: Log Int)
+
+  , testCase "retry with success" $
+        let lg1 = fromProgress $ Step 1 (Step 2 (Done "Done"))
+            lg2 = fromProgress $ Step 3 (Step 4 (Fail "Error"))
+        in toProgress (retry lg1 (const lg2))
+           @?= (Step 1 (Step 2 (Done "Done")) :: Log Int)
+
+  , testCase "failWith" $
+        toProgress (failWith 1 "Error") @?= (Step 1 (Fail "Error") :: Log Int)
+
+  , testCase "succeedWith" $
+        toProgress (succeedWith 1 "Result")
+        @?= (Step 1 (Done "Result") :: Log Int)
+
+  , testCase "continueWith" $
+        let failure = Fail "Error"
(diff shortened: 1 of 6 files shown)


23Skidoo:
Looks OK to me.

ezyang:
The speed up is good, but I wonder if we shouldn't be seeing a bigger speedup. How did the memory usage compare? If I understand correctly, you have to force the difference list every time you backjump; it is too bad this cannot be avoided.

grayjay:
The solver should only force the difference list once at the end, when it calls `toProgress` in `backjumpAndExplore`.  Is it forcing the list somewhere else?

There is less allocation, but no reduction in total memory.  I ran it again with -s, with all three commits:

`cabal install --dry-run phooey hackage-server +RTS -s`

master:
```
   5,989,087,760 bytes allocated in the heap
   3,461,095,560 bytes copied during GC
     113,995,536 bytes maximum residency (28 sample(s))
       1,749,544 bytes maximum slop
             331 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0     11475 colls,     0 par    1.947s   1.948s     0.0002s    0.0007s
  Gen  1        28 colls,     0 par    1.942s   1.942s     0.0694s    0.1303s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    2.530s  (  2.947s elapsed)
  GC      time    3.889s  (  3.890s elapsed)
  EXIT    time    0.012s  (  0.012s elapsed)
  Total   time    6.432s  (  6.849s elapsed)

  Alloc rate    2,367,430,271 bytes per MUT second

  Productivity  39.5% of total user, 37.1% of total elapsed

gc_alloc_block_sync: 0
whitehole_spin: 0
gen[0].sync: 0
gen[1].sync: 0
```

this branch:
```
   5,385,345,672 bytes allocated in the heap
   3,038,809,800 bytes copied during GC
     114,326,352 bytes maximum residency (26 sample(s))
       1,924,128 bytes maximum slop
             332 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0     10319 colls,     0 par    1.777s   1.778s     0.0002s    0.0007s
  Gen  1        26 colls,     0 par    1.651s   1.651s     0.0635s    0.1231s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    2.415s  (  2.836s elapsed)
  GC      time    3.428s  (  3.429s elapsed)
  EXIT    time    0.012s  (  0.012s elapsed)
  Total   time    5.856s  (  6.278s elapsed)

  Alloc rate    2,230,254,649 bytes per MUT second

  Productivity  41.5% of total user, 38.7% of total elapsed

gc_alloc_block_sync: 0
whitehole_spin: 0
gen[0].sync: 0
gen[1].sync: 0
```

ezyang:
OK, I guess I misunderstood; I thought it was converting every time you backjumped.

cabal-install/tests/UnitTests/Distribution/Solver/Modular/RetryLog.hs
```diff
@@ -0,0 +1,69 @@
...
+import Distribution.Solver.Types.Progress
+
+import Test.Tasty (TestTree)
+import Test.Tasty.HUnit (testCase, (@?=))
+
+type Log a = Progress a String String
+
+tests :: [TestTree]
+tests = [
+    testCase "convert to and from RetryLog ending in failure" $
```
grayjay: I converted all of the tests to Quickcheck properties, except for the two that need to use `Message` as the step type.

grayjay:
Actually, I started getting a speedup of 1.44 (3.56s vs 2.47s) for yesod after I did a `cabal update`. The new aeson release caused the solver to spend more time resolving a conflict near level 177.

ezyang:
Ship it! It's probably worth spending some time considering how we can get perf stats for the solver into CI.

grayjay:
Thanks for the review! I agree that it would be nice to run benchmarks in CI.  As a simple solution, maybe we could solve for a few popular packages and compare times with the latest cabal release.

PR #3683, opened 2016-08-09
Fix preprocessors to put output in correct directory.

ezyang:


23Skidoo:
This breaks the test suite apparently.

23Skidoo:
LGTM.

PR #3693, opened 2016-08-15
Deprecate ProgramConfiguration et al.

23Skidoo:
This stuff has been actually deprecated for ages, in 2.0 we can
finally mark it as such. Next up: removing all uses from
Cabal/cabal-install code.

~~WIP, please don't merge yet.~~


ezyang:
OK, would be good if there was guidance on how to move to the new API written somewhere.

23Skidoo:
It's as simple as s/old/new/, we're just deprecating the synonyms.

23Skidoo:
The reason I didn't simply run a global search and replace is that I also want to change all instances of `conf` as a variable to `progdb`.

23Skidoo:
This should be now ready to go in.

ezyang:
OK, merge it when it's green.

PR #3695, opened 2016-08-15
Fix #3041, ifndef all CURRENT_PACKAGE_KEY and related macros.

### LINKED ISSUE #3041: Cabal should ifndef CURRENT_PACKAGE_KEY macro definitions
@mgsloan reports that Stack has to manually undef CURRENT_PACKAGE_KEY in order to avoid a slew of redefinition errors. We should ifndef these macro definitinos.

ezyang:
I also realized LOCAL_COMPONENT_ID is pessimal if we don't
depend on a library, so I removed it. (It has never been
in a real Cabal release.)

### DIFF
--- a/Cabal/Distribution/Simple/Build/Macros.hs
+++ b/Cabal/Distribution/Simple/Build/Macros.hs
@@ -37,6 +37,36 @@ import Distribution.Text
 -- * Generate cabal_macros.h
 -- ------------------------------------------------------------
 
+-- Invariant: HeaderLines always has a trailing newline
+type HeaderLines = String
+
+line :: String -> HeaderLines
+line str = str ++ "\n"
+
+ifndef :: String -> HeaderLines -> HeaderLines
+ifndef macro body =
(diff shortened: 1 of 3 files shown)


ezyang:
CC @mgsloan who requested this.

mgsloan:
Thanks, LGTM!

Cabal/Distribution/Simple/Build/Macros.hs
```diff
@@ -37,6 +37,36 @@ import Distribution.Text
...
+type HeaderLines = String
+
+line :: String -> HeaderLines
+line str = str ++ "\n"
+
+ifndef :: String -> HeaderLines -> HeaderLines
+ifndef macro body =
+    line ("#ifndef " ++ macro) ++
+    body ++
+    line "#endif"
```
23Skidoo: I'd also add a `"/*" ++ macro ++ "*/"` comment after `ifdef`.

23Skidoo:
LGTM. I guess this warrants a changelog note?

PR #3696, opened 2016-08-16
Make setEnv accept empty strings.

ezyang:
See https://ghc.haskell.org/trac/ghc/ticket/12494; we cargo-culted
some incorrect code from base. Lift the restriction.


23Skidoo:
LGTM. IIRC, this behaviour originated in the `setenv` package.

/cc @sol

PR #3702, opened 2016-08-17
Simple.SrcDist cleanup (fixes #3656)

### LINKED ISSUE #3656: Filter all autogen modules / c-sources

fmaste:
- The `PackageDescription` is not changed any more, just the pieces of it we need to (see `preprocessComponentSdist`).
- Created `filterAutogenModules` and `filterAutogenBuildInfo` functions so given a `PackageDescription` and `BuildInfo` returns filtered answers. Now is clear where to add an `autogen-c-sources` field.
- Created `sdistSourcesList` and `sdistSourcesDirectory` functions moving the logic out of frontend `sdist` where this last one just reads the flags and prints messages.
- `findSetupFile`, `findMainExeFile` and two pieces of code from `listPackageSources` where identical, extracted all on `findMainFile` (You can see through the commits how it is done step by step).
- Renamed not exported functions so now we have `findModulesFiles`, `findMainFile`, `findIncludeFile` and `findSetupFile` together.
- TODO: `verbosity` variable of `sdistSourcesList`, `sdistSourcesDirectory` goes to `listPackageSources` then `listPackageSourcesOrdinary` that calls `defaultPackageDesc verbosity` that ignores it. But `listPackageSources` is exported and is too much for a single PR.

There are a lot of commits because it's the only way to make small steps to figure all this out and test between every change. **Dont worry I will `merge --squash`!** but before you can see that I tried hard not to break anything.


Cabal/Distribution/Simple/SrcDist.hs
```diff
@@ -171,12 +171,12 @@ listPackageSourcesOrdinary verbosity pkg_descr pps =
     -- Library sources.
     fmap concat
     . withAllLib $ \Library { exposedModules = modules, libBuildInfo = libBi } ->
-     allSourcesBuildInfo libBi pps modules
+     allSourcesBuildInfo libBi pps (modules ++ (otherModules libBi))
```
ezyang: Maybe you undid this later, but is this an improvement?

Cabal/Distribution/Simple/SrcDist.hs
```diff
@@ -150,55 +163,55 @@ listPackageSourcesOrdinary :: Verbosity
...
   [
     -- Library sources.
     fmap concat
-    . withAllLib $ \Library { exposedModules = modules, libBuildInfo = libBi } ->
-     allSourcesBuildInfo libBi pps modules
+    . withAllLib $ \Library {exposedModules = ex, libBuildInfo = bi} -> do
+       let exposed = filterAutogenModules pkg_descr bi ex
+       let others = otherModules $ filterAutogenBuildInfo pkg_descr bi
+       moduleSrcs <- findModulesFiles (hsSourceDirs bi) pps (exposed ++ others)
+       return $ moduleSrcs ++ cSources bi ++ jsSources bi
```
ezyang: Surely `cSources bi ++ jsSources bi` can be put into a common function.

Cabal/Distribution/Simple/SrcDist.hs
```diff
@@ -150,55 +163,55 @@ listPackageSourcesOrdinary :: Verbosity
...
+--       Function defaultPackageDesc isn't using it internally.
   fmap concat . sequenceA $
   [
     -- Library sources.
     fmap concat
-    . withAllLib $ \Library { exposedModules = modules, libBuildInfo = libBi } ->
-     allSourcesBuildInfo libBi pps modules
+    . withAllLib $ \Library {exposedModules = ex, libBuildInfo = bi} -> do
+       let exposed = filterAutogenModules pkg_descr bi ex
+       let others = otherModules $ filterAutogenBuildInfo pkg_descr bi
```
ezyang: Warning! This function implies that everything autogen gets filtered from `BuildInfo` but actually this only handles modules.

ezyang:
I like the patch, though it would be nice to make sure it all hangs together (i.e. actually lets you implement the thing you want to implement.) Do you want to merge this sooner rather than later? I don't think `sdist` is a high churn area and we can block other patchsets to it in the meantime.

Cabal/Distribution/Simple/SrcDist.hs
fmaste: Yes, at that time it was in discussion where to put the autogen field (really just me trying that the field was global)

ezyang:
@fmaste Just let us know when you want people to look at your patch again!

fmaste:
Finished and waiting for someone else to have a look

23Skidoo:
@fmaste Can you please use rebase instead of merge to get rid of spurious merge commits?

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -253,6 +253,18 @@ checkLibrary pkg lib =
            "An 'autogen-module' is neither on 'exposed-modules' or "
         ++ "'other-modules'."
 
+    -- check that all autogen-c-sources appear on c-sources
+  , check
+      (not $ and $ map (flip elem
+        (cSources $ libBuildInfo lib)) (autogenCSources $ libBuildInfo lib)) $
+      PackageBuildImpossible $ "An 'autogen-c-sources' is not on 'c-sources'"
```
ezyang: This error would be better if (1) it explained WHY you had to duplicate them (it's for backwards compatibility with old Cabal) and (2) said that it was sufficient to just copy paste.

Cabal/Distribution/Simple/SrcDist.hs
```diff
@@ -149,55 +162,64 @@ listPackageSourcesOrdinary :: Verbosity
...
     -- Library sources.
     fmap concat
-    . withAllLib $ \Library { exposedModules = modules, libBuildInfo = libBi } ->
-     allSourcesBuildInfo libBi pps modules
+    . withAllLib $ \Library {exposedModules = ex, libBuildInfo = bi} -> do
+       let exposed = filterAutogenModules pkg_descr bi ex
+       let filteredBi = filterAutogenBuildInfo pkg_descr bi
+       moduleSrcs <- findModulesFiles (hsSourceDirs bi) pps
+                                      (exposed ++ otherModules filteredBi)
+       return $ moduleSrcs ++ cSources filteredBi ++ jsSources filteredBi
```
ezyang: You're still copy-pasting cSources/jsSources four times ;)

ezyang:
It looks pretty good. I didn't review each diff individually (I can do so if you like) so some of the changes look "big" to me but if there are bugs, we can iron it out.

I know in #3656 we ended up requiring users to duplicate autogenerated modules in `c-modules` and `autogen-c-modules`. Did we also implement something where if you forced a sufficiently new version of Cabal, you could avoid the duplication? I can't remember why we didn't do it, but looking at this patch I felt a pang of sympathy for people who think that Cabal descriptions are too wordy.

23Skidoo:
Looks reasonable to me as well. I guess we need documentation for `autogen-{c,js}-sources`?

ezyang:
@fmaste Feel free to squash and merge :)

fmaste:
> @fmaste Can you please use rebase instead of merge to get rid of spurious merge commits?
If I rebase I can't easily code at work, home or on the go at the same time. That's why I'm going to merge squash this PR

fmaste:
> I know in #3656 we ended up requiring users to duplicate autogenerated modules in c-modules and autogen-c-modules. Did we also implement something where if you forced a sufficiently new version of Cabal, you could avoid the duplication?

With modules we can't because on libraries we have `exposed-modules` and `other-modules`, that would mean having something like `exposed-autogen-modules` and `other-autogen-modules` there but `autogen-modules` on executables, benchmakrs and tests. **It's adding three new fields for the same thing were which one to use depends on the stanza or repeating modules names like we do now.**

> I can't remember why we didn't do it, but looking at this patch I felt a pang of sympathy for people who think that Cabal descriptions are too wordy.

Yes, right now this is horrible. The other option that I liked and had no cons was having all autogen modules on a global field much like `extra-source-files` but I was the only one that wanted that.

Cabal/Distribution/PackageDescription/Check.hs
fmaste: 1) The real reason why we repeat filenames is not for backward compatibility is to be consistent with how `autogen-modules` work, it would be awkward to explain that on an error message.

2) Do you mean a helper function inside this module that does `not $ and $ map (flip elem e)` or something like libModules and `libModulesAutogen` on `Distribution.Types.Library` ?

Cabal/Distribution/Simple/SrcDist.hs
fmaste: The `cSources filteredBi ++ jsSources filteredBi` code appears five times but is used on different `BuildInfo` structures. Five places seems to have the same 4-line code but there are three distinct cases that can't be extracted on a single functions without being unreadable.

BardurArantsson:
> If I rebase I can't easily code at work, home or on the go at the same time. That's why I'm going to merge squash this PR

You certainly can! :) There are at least two ways I can think of: a) Just create branches from the 'merge-candidate' branch for each separate location. That gives you a simple reference point and you can merge/rebase at will using the `--onto` flag of `git rebase`. b) You care forego the separate branches and still use the `--onto` strategy. All it requires is a bit of looking up commits in the git log (gitk is easier here). The idea is to just always transplanting any 'local' commits onto the rebased 'orgin' branch. It does require a little bit of practice though.

*However*, in this case you don't seem to have "logical unit" type commits with meaningful messages/descriptions anyway, so I'd say just squashing everything together probably wouldn't hurt.

ezyang:
@23Skidoo @BardurArantsson, I don't think we should tell people what their workflow should be, as long as it doesn't interfere with other people working on the project ;)

> With modules we can't because on libraries we have exposed-modules and other-modules, that would mean having something like exposed-autogen-modules and other-autogen-modules there but autogen-modules on executables, benchmakrs and tests. It's adding three new fields for the same thing were which one to use depends on the stanza or repeating modules names like we do now.

Yes, that was the reason! (I totally forgot about it.) It would be really good if this was recorded durably somewhere (maybe put it in the same wiki page that I want the error message to link to, see comments on code).

Maybe everything will be better with #3744 where we can just fill in this data automatically. Then it's OK to optimize for ease of consumption rather than ease of writing.

> The other option that I liked and had no cons was having all autogen modules on a global field much like extra-source-files but I was the only one that wanted that.

IIRC, I didn't like this plan because it didn't generalize to C sources/JS sources which are not associated with module names. So you'd need both `autogen-modules` and `autogen-files`, but then you also needed to respect the differing include path resolution correctly and then we just had a new field for each existing one. Does that sound familiar?

BardurArantsson:
@ezyang I certainly didn't mean to "tell". I was just offering a suggestion that (hopefully) could solve @fmaste's problem.

(But I *do* care about commit logs, so maybe I'm projecting some of that into the tone of my message. *Shrug*)

23Skidoo:
> I don't think we should tell people what their workflow should be, as long as it doesn't interfere with other people working on the project ;)

Spurious merge commits pollute history and make it harder to do code review, that's why we care about them.

ezyang:
Yes, but @fmaste has already stated that he will merge squash before pushing to master (IIUC, that will be a merge, then a git rebase -i which will result in a single commit with no merges) so no problem right :)

On September 16, 2016 6:40:14 PM GMT+09:00, Mikhail Glushenkov <notifications@github.com> wrote:
>> I don't think we should tell people what their workflow should be, as
>long as it doesn't interfere with other people working on the project
>;)
>
>Spurious merge commits pollute history and make it harder to do code
>review, that's why we care about them.
>
>-- 
>https://github.com/haskell/cabal/pull/3702#issuecomment-247559684

-- 

23Skidoo:
Yeah, it's no problem in this case, but in general merge commits make it harder to do code review, so it's nice to get rid of them before submitting a PR.

ezyang:
No, I don't see how merge commits necessarily cause problems at review time. For example, GitHub suppresses the diff if the merge is "clean". Very rarely git will mess up an "obvious" textual merge. But that's what looking at the final diff is for.

If the merge is dirty, well, it is true that it is "easier" to review patches if the various bits of the merge have been backported back. I do this for my own patchsets. But it is extremely tedious and time consuming. I wouldn't wish it on anyone else. For example, for this patchset, I reviewed it by just clicking on "Changed files" overall and looking at the changes.

fmaste:
> You certainly can! :) There are at least two ways I can think of ...

Everything is possible with git but no thanks, sounds like a recipe for disaster

> @23Skidoo @BardurArantsson, I don't think we should tell people what their workflow should be, as long as it doesn't interfere with other people working on the project ;)

As I'm doing some refactoring I do little non-destructive changes with a silly commit description just to identify them when looking at the history. At all time I test locally and if something fails I cherry pick and correct.

> Yeah, it's no problem in this case, but in general merge commits make it harder to do code review, so it's nice to get rid of them before submitting a PR.

Nobody else is working on this branch and the merge is not shown on the diff. After squashing all to master the result will be the same as rebasing and I can delete this branch. The only problem I see is that you have to skip merges if you are looking at every commit separately, but I doubt somebody is doing that with all this commits

ezyang:
@fmaste You're all good. Squash and push!

23Skidoo:
> For example, GitHub suppresses the diff if the merge is "clean".

It only recently started doing that. Not sure how well it works for "dirty" merges.

>  I do this for my own patchsets. But it is extremely tedious and time consuming. I wouldn't wish it on anyone else.

I sympathise, but this is essentially an argument for not removing merge commits at all.

> For example, for this patchset, I reviewed it by just clicking on "Changed files" overall and looking at the changes.

It works in this case, but in general not having everything rolled up in one giant commit and using logical commits instead of historical makes the reviewer's life easier.

23Skidoo:
@fmaste Are you going to update this PR? This is basically good to go, just needs to be squashed and have some minor conflicts resolved. Anything we can do to help?

ezyang:
@23Skidoo Maybe we should just squash and push ourselves.

phadej:
I assign this to @ezyang to *squash and push*

phadej:
This PR contains some good stuff, but it cannot be merged as is. `v2-sdist` doesn't (and will not) run `configure` beforehand.

phadej:
sdist code have changed, and some changes are there now.

PR #3704, opened 2016-08-17
Clarify what logical negation of impl with version ranges means

RyanGlScott:
I was recently [tripped up](https://github.com/haskell/vector/pull/134#issuecomment-240459112) since negating (`!`) an `impl` test with a version constraint didn't behave like I thought it did. Per @hvr's request, I'm documenting this behavior here.


hvr:
Thanks! :-)

ezyang:
Thanks, merged.

PR #3705, opened 2016-08-20
Implement Caret-style version range operator

hvr:
This implements a new syntactic sugar: The version range operator
`^>=` which is equivalent to `>=` intersected with an
automatically inferred major upper bound.

This new syntax is only allowed for `cabal-version: >=2.0`, and
allows to describe the most common use of version
bounds more conveniently:

    build-depends: foo ^>= 1.2.3.4,
                   bar ^>= 1

The declaration above is exactly equivalent to

    build-depends: foo >= 1.2.3.4 && < 1.3,
                   bar >= 1 && < 1.1

The `^`-symbol was chosen because it can serve as a mnemonic when the
`>` sign is rotated and interpreted as "less than ceiling"

Moreover, `^` appears to become a quasi-standard to denote morally
equivalent operator that way in other language ecosystems which similiar
to Haskell have adopted semantic versioning:

 - Node: https://nodesource.com/blog/semver-tilde-and-caret/
 - Bower: https://bower.io/docs/api/#install
 - PHP: https://getcomposer.org/doc/articles/versions.md#caret

Ruby, on the other hand, uses a Tilde operator (`~>`) for that
purpose (but with a less robust semantic):

 - https://blog.codeship.com/optimists-guide-pessimistic-library-versioning

And Python is currently planing to use an `~=` operator:

 - https://www.python.org/dev/peps/pep-0440/#compatible-release

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -107,6 +107,7 @@ extra-source-files:
   tests/PackageTests/CMain/Bar.hs
   tests/PackageTests/CMain/foo.c
   tests/PackageTests/CMain/my.cabal
+  tests/PackageTests/CaretOperator/my.cabal
   tests/PackageTests/Configure/A.hs
   tests/PackageTests/Configure/Setup.hs
   tests/PackageTests/Configure/include/HsZlibConfig.h.in
@@ -467,6 +468,7 @@ test-suite package-tests
     PackageTests.AutogenModules.Package.Check
     PackageTests.AutogenModules.SrcDist.Check
     PackageTests.BenchmarkStanza.Check
+    PackageTests.CaretOperator.Check
     PackageTests.TestStanza.Check
     PackageTests.DeterministicAr.Check
     PackageTests.TestSuiteTests.ExeV10.Check
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1054,6 +1054,18 @@ checkCabalVersion pkg =
            [ display (Dependency name (eliminateWildcardSyntax versionRange))
            | Dependency name versionRange <- depsUsingWildcardSyntax ]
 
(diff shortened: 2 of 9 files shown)


phadej:
LGTM. But I'll like to see tests. (which would break in parsec branch)

23Skidoo:
Would be nice to add a changelog note.

hvr:
I've changed the syntax from `^>` to `^>=` because it's more accurate/self-descriptive (since it's more evident that `^`+`>=` translates to `<` + `>=` than the original syntax) in my opinion.

@23Skidoo changelog entry added

@phadej finally brought myself to try to come up with a simple parser test... getting there longer than it should have...

23Skidoo:
Thanks! So I think this can be merged once Travis is green.

unclechu:
That’s a bit confusing since in other platforms `^` has a different meaning:

``` javascript
semver.toComparators('^1.2.3')
// [ [ '>=1.2.3-0', '<2.0.0-0' ] ]

// compare upper limit for ~
semver.toComparators('~1.2.3')
// [ [ '>=1.2.3-0', '<1.3.0-0' ] ]
```

I’d prefer to see `~>=` instead. But it seems it’s too late to change.

unclechu:
Or maybe I’m missing something?

gbaz:
> Or maybe I’m missing something?

In the haskell pvp (https://pvp.haskell.org/), which predates semver btw, the _first two_ components indicate major version, while in semver, only the first component indicates major version. So semantically the caret operator means the same thing in both cases -- picking the range up to the next major version. It just so happens that "major version" covers a different number of components.

fgaz:
Well, haskell packages use [PVP](pvp.haskell.org), which is a bit different from semver (and predates it).

In PVP, the first two numbers are major, so the meaning of `^>=` in cabal is the equivalent of what you wrote, just adjusted for the PVP.

unclechu:
@gbaz thanks for this clarification!

PR #3711, opened 2016-08-22
Fix (exe-)depends in plan.json

hvr:


ezyang:
OK looks fine, let's wait for build to pass. We should also squash the patches.

I'm not hot on the conditional. Can we just use the less efficient implementation in all cases? Not like this is performance sensitive.

hvr:
@ezyang yeah we can... I'd still leave the more efficient as a comment in there for when we can drop support for pre-containers-0.5

ezyang:
Sure. Just not as executable code.

Issue #462, opened 2012-05-24
Support relocatable packages


bos:
(Imported from [Trac #469](http://hackage.haskell.org/trac/hackage/ticket/469), reported by @dcoutts on 2009-01-20)

This is both useful and doable on both Windows and Unix systems.

We already have limited support for relocatable (also sometimes called prefix-independent) packages on Windows. Specifically, a relocatable package is often what people want when they want to prepare a package for redistribution, especially on Windows for use with an installer. What is required is that an exe and all supporting files (data files and potentially also other shared libs) exist together in one directory hierarchy and that that directory can be placed anywhere in the filesystem and the exe will still be able to find all its associated support files.

There are some restrictions on preparing a relocatable package. In particular the user must configure it such that all install directories for the various kinds of files (libdir, datadir etc) are relative to the <tt>$prefix</tt>.

Configuring a relocatable package should be something that is done explicitly. At the moment it is done simply by configuring it in the right way, but Cabal is never aware that the user is trying to construct a relocatable package. If it is explicit then we can take different actions if necessary (as it is on unix) and do additional checks.

So a configure option <tt>--relocatable</tt> or something should be used. Installing a relocatable package is somewhat of an exception because it involves using a prefix which would otherwise be empty. Creating an image for a relocatable package would use a empty <tt>$prefix</tt>.

The tricky issues that have to be addressed are:

<ul><li>data files for the program itself
</li><li>data files in library dependencies
</li><li>shared libraries
</li></ul>The mechanism for finding data files at runtime is the <tt>Paths_pkgname</tt> module that Cabal generates. On Windows this can make use of the Win32 api that lets a .exe program discover where it was run from. On unix there is no reliable equivalent but the next best alternative is to use wrapper scripts to set environment variables before running the real program.

This mechanism can work pretty well for data files that belong to the executable. It is rather harder for data files in dependent packages (see #459). One option would be to copy the data files belonging to dependent packages and install them along with the executable (checking for clashes). This would involve being able to locate the data files of installed libraries which would require extending the data stored for installed packages. Libraries that require data files are relatively rare (though this should be checked) so this feature could be punted for an initial version.

[...]

PR #3713, opened 2016-08-22
Document environment variables that can be used to change Paths values.

ezyang:
See also #462.

### DIFF
--- a/Cabal/doc/developing-packages.markdown
+++ b/Cabal/doc/developing-packages.markdown
@@ -2071,6 +2071,13 @@ getLibexecDir :: IO FilePath
 getSysconfDir :: IO FilePath
 ~~~~~~~~~~~~~~~
 
+The actual location of all these directories can be individually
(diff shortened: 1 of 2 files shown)


ezyang:
(Build failed because I forgot to ci skip)

Cabal/doc/developing-packages.markdown
```diff
@@ -2071,6 +2071,13 @@ getLibexecDir :: IO FilePath
 getSysconfDir :: IO FilePath
 ~~~~~~~~~~~~~~~
 
+The actual location of all these directories can be individually
+overridden at runtime using environment variables of the form
+`pkg_name_var`, where `pkg_name` is the name of the package with
+all hyphens converted into underscores, and `var` is either
+`bindir`, `libdir`, `datadir`, `libexedir` or `sysconfdir`.
```
hvr: It may be useful to give a concrete example for some well-known package name as to how such a var looks like for demonstration purposes

Issue #3475, opened 2016-06-04
Build sequencing when build taffybar


erikd:
I'm using ghc-7.10 and cabal-install-1.24.0.0 and I'm getting a strange build failure when building the "taffybar" package.

I have a git clone of taffybar and in that directory I do:

```
cabal sandbox init
cabal install --dependencies-only --enable-tests

```

This then fails with:

```
Configuring gtk2hs-buildtools-0.13.2.0...
Building gtk2hs-buildtools-0.13.2.0...
Installed xmonad-contrib-0.12
Installed gtk2hs-buildtools-0.13.2.0
Configuring cairo-0.13.3.0...
Configuring glib-0.13.4.0...
Failed to install cairo-0.13.3.0
Build log ( /md0/jenkins/workspace/xmonad-taffybar/.cabal-sandbox/logs/cairo-0.13.3.0.log ):
cabal: Entering directory '/tmp/cabal-tmp-17308/cairo-0.13.3.0'
cabal: Leaving directory '/tmp/cabal-tmp-17308/cairo-0.13.3.0'
Failed to install glib-0.13.4.0
Build log ( /md0/jenkins/workspace/xmonad-taffybar/.cabal-sandbox/logs/glib-0.13.4.0.log ):
cabal: Entering directory '/tmp/cabal-tmp-17308/glib-0.13.4.0'
cabal: Leaving directory '/tmp/cabal-tmp-17308/glib-0.13.4.0'
cabal: Error: some packages failed to install:
cairo-0.13.3.0 failed during the configure step. The exception was:
user error (The package 'cairo' requires Cabal library version -any && >=1.24
but no suitable version is installed.)
gio-0.13.3.0 depends on glib-0.13.4.0 which failed to install.
glib-0.13.4.0 failed during the configure step. The exception was:
user error (The package 'glib' requires Cabal library version -any && >=1.24
but no suitable version is installed.)
gtk-0.14.4 depends on cairo-0.13.3.0 which failed to install.
gtk-traymanager-0.1.6 depends on cairo-0.13.3.0 which failed to install.
pango-0.13.3.0 depends on cairo-0.13.3.0 which failed to install.
```

but if I then do:

```
cabal install glib-0.13.4.0 cabal install cairo-0.13.3.0
cabal install --dependencies-only --enable-tests
```

everything is fine. Cabal seems to get something wrong with the sequencing, I have reproduced this on two different machines, but with the same versions of ghc and cabal-install.

[...]

Issue #3436, opened 2016-05-15
Have to install twice package with 'custom-setup' stanza


garetxe:
I just uploaded a new version of the `gi-glib` package to hackage using the new `custom-setup` stanza, but `cabal install` fails to install it on the first try:
```
$ cabal sandbox init
Writing a default package environment file to
/home/inaki/tst/cabal.sandbox.config
Creating a new sandbox at /home/inaki/tst/.cabal-sandbox

$ cabal install gi-glib-0.0.2.2
[... work for a while ...]
Installed Cabal-1.24.0.0
[... work some more ...]
Configuring gi-glib-0.0.2.2...
Failed to install gi-glib-0.0.2.2
Build log ( /home/inaki/tst/.cabal-sandbox/logs/gi-glib-0.0.2.2.log ):
cabal: Entering directory '/tmp/cabal-tmp-23716/gi-glib-0.0.2.2'
cabal: Leaving directory '/tmp/cabal-tmp-23716/gi-glib-0.0.2.2'
cabal: Error: some packages failed to install:
gi-glib-0.0.2.2 failed during the configure step. The exception was:
user error (The package 'gi-glib' requires Cabal library version -any &&
>=1.24 but no suitable version is installed.)
```

Notice that `Cabal-1.24.0.0` was installed properly, but somehow the configure of `gi-glib` still fails. But repeating the command now suceeds:
```
$ cabal install gi-glib-0.0.2.2
Resolving dependencies...
Notice: installing into a sandbox located at /home/inaki/tst/.cabal-sandbox
Configuring gi-glib-0.0.2.2...
Building gi-glib-0.0.2.2...
Installed gi-glib-0.0.2.2
```

garetxe:
For reference, here is the `.cabal` file for the package (from https://hackage.haskell.org/package/gi-glib-0.0.2.2/gi-glib.cabal):
```
name:               gi-glib
version:            0.0.2.2
x-revision: 1
synopsis:           GLib bindings
description:        Bindings for GLib, autogenerated by haskell-gi.
homepage:           https://github.com/haskell-gi/haskell-gi
license:            LGPL-2.1
license-file:       LICENSE
author:             Will Thompson, Iñaki García Etxebarria and Jonas Platte
maintainer:         Iñaki García Etxebarria (garetxe@gmail.com)
category:           Bindings
build-type:         Custom
cabal-version:      >= 1.24

extra-source-files: GLib.overrides

custom-setup
[...]

PR #3723, opened 2016-08-28
Consistently use the Cabal version picked by the dependency solver.

### LINKED ISSUE #3433: cabal-install 1.24 passes bad options to custom Setup
NB: `base-noprelude-4.7.0.2` uses a `custom-setup` stanza

in this case, cabal-install-1.24 needs to talk to a custom `Setup.hs` compiled against Cabal-1.20:

```
$ cabal install -v2 -w /opt/ghc/7.8.4/bin/ghc base-noprelude 
Warning: /home/hvr/.cabal/config: 'root-keys' for repository
"hackage.haskell.org" non-empty, but 'secure' not set to True.
/home/hvr/.cabal/bin/alex --version
/home/hvr/.cabal/bin/c2hs --numeric-version
/home/hvr/.cabal/bin/cpphs --version
/usr/bin/gcc -dumpversion
/home/hvr/.cabal/bin/ghcjs --numeric-ghcjs-version
/home/hvr/.cabal/bin/ghcjs-pkg --version
looking for tool haddock near compiler in /opt/ghc/7.8.4/bin
found haddock in /opt/ghc/7.8.4/bin/haddock
/opt/ghc/7.8.4/bin/haddock --version
/home/hvr/.cabal/bin/happy --version
looking for tool hpc near compiler in /opt/ghc/7.8.4/bin
found hpc in /opt/ghc/7.8.4/bin/hpc
/opt/ghc/7.8.4/bin/hpc version
looking for tool hsc2hs near compiler in /opt/ghc/7.8.4/bin
found hsc2hs in /opt/ghc/7.8.4/bin/hsc2hs
/opt/ghc/7.8.4/bin/hsc2hs --version
/home/hvr/.cabal/bin/HsColour -version
/home/hvr/.cabal/bin/jhc --version
/opt/ghc/7.8.4/bin/ghc -c /tmp/1804289383846930886.c -o /tmp/16816927771714636915.o -ferror-spans
/usr/bin/ld -x -r /tmp/16816927771714636915.o -o /tmp/1957747793424238335.o
/usr/bin/pkg-config --version
/bin/tar --help
Reading available packages...
/usr/bin/pkg-config --list-all
/usr/bin/pkg-config --modversion sqlite3 blkid pixman-1 iso-codes gmodule-export-2.0 libevent_pthreads xaw7 gstreamer-interfaces-0.10 cairo-svg libdrm_intel recordproto nettle gnutls libpcreposix lapack-netlib lqr-1 atk-bridge-2.0 m17n-gui libelf cairo-tee xcb-render dbus-1 e2p dri gnome-icon-theme-symbolic xkbcommon dri3proto systemd udisks adwaita-icon-theme mobile-broadband-provider-info libbsd cairo-png wayland-cursor gstreamer-pbutils-0.10 QtXml pangoxft QtDBus xcursor libcurl xft libidn libxslt mircookie glu gstreamer-rtp-0.10 xtrans libexif gdk-pixbuf-2.0 'gtk+-x11-3.0' xf86vidmodeproto QtScriptTools lapack gnome-system-tools damageproto QtDeclarative libpulse kbproto expat videoproto cairo-xlib mit-krb5-gssapi banshee-lastfm-gui gstreamer-netbuffer-0.10 uuid valgrind fontutil xcb-dri2 xcb-dri3 panel gconf-2.0 randrproto pygoocanvas libxdot openal xrandr libgssglue 'Magick++' dleyna-server-1.0 wayland-egl bash-completion QtGui gdk-2.0 capi20 xcomposite libtasn1 dri2proto QtScript OpenEXR gstreamer-riff-0.10 icu-i18n libgphoto2 datrie-0.2 QtDesigner sm xorg-sgml-doctools cairo-ps p11-kit-1 libcdt libpulse-mainloop-glib python gstreamer-check-0.10 gdk-mir-3.0 protobuf-lite python2 mirclient gnome-icon-theme gmodule-2.0 gdk-pixbuf-xlib-2.0 ImageMagick gstreamer-dataprotocol-0.10 libdrm_radeon libdrm fixesproto libgdiplus libpulse-simple pango QtDesignerComponents banshee-mono-media m17n-flt geoclue-2.0 m17n-core banshee-hyena pangoft2 atspi-2 xineramaproto gdk-3.0 sane-backends compositeproto icu-uc libbsd-ctor xvmc ginn libpcre libpcre16 cairo-xcb-shm libthai glproto cairo-pdf gstreamer-sdp-0.10 libevent_openssl pangocairo gstreamer-0.10 lcms2 pm-utils gstreamer-app-0.10 gnome-mime-data-2.0 x11-xcb libpq gthread-2.0 libusb-1.0 'gtk+-broadway-3.0' vdpau gstreamer-base-0.10 libssl Qt3Support libpcrecpp cairo-xcb gl xorg-wacom gstreamer-plugins-base-0.10 menu QtXmlPatterns intel-gen4asm MagickWand egl libxml-2.0 mit-krb5 libcrypto QtTest libdrm_nouveau alsa xextproto cairo-gobject yelp-xsl gstreamer-controller-0.10 gmodule-no-export-2.0 wayland-client libsepol bdw-gc x11 libdrm_amdgpu xtst libgvpr gstreamer-fft-0.10 ddjvuapi 'gtk+-mir-3.0' libevent libpcre32 xcb-glx libpathplan QtCLucene QtCore xmu 'gtk+-unix-print-2.0' gstreamer-cdda-0.10 gdk-broadway-3.0 hogweed poppler-data libexslt dleyna-server-service-1.0 wayland-server ibus-table xi 'libcrypto++' banshee-services udev libR xbitmaps xt cairo-fc xv glib-2.0 cairo xorg-macros gdk-wayland-3.0 Wand icu-io cairo-ft banshee-hyena-data-sqlite xcb-sync banshee-lastfm gio-unix-2.0 MagickCore libquvi-scripts pygtksourceview-2.0 QtSql xfixes presentproto xcb-xfixes tomboy-addins 'gtk+-unix-print-3.0' banshee-thickclient tic gstreamer-tag-0.10 xinerama libbsd-overlay tinfo epoxy libpng m17n-shell usbutils xpm 'gtk+-wayland-3.0' gstreamer-rtsp-0.10 gdk-x11-2.0 blas-netlib xml2po librsvg-2.0 libomxil-bellagio xshmfence xcb-randr IlmBase banshee-musicbrainz freetype2 xext libjpeg libudev libgphoto2_port libtiff-4 xau harfbuzz xcb-present ice xdamage xkbcomp libselinux openssl atk xxf86vm cairo-script 'gtk+-2.0' ruby-1.9 banshee-hyena-gui dleyna-connector-dbus-1.0 python-2.7 QtHelp icu-le libffi wayland-scanner xkeyboard-config gnome-doc-utils xcb-shape banshee-collection-indexer libclc harfbuzz-icu pthread-stubs shared-mime-info QtOpenGL xrender gstreamer-net-0.10 banshee-core inputproto icu-lx 'ImageMagick++' libpng12 banshee-webbrowser xcb-shm xproto protobuf xcb libmpg123 dbus-glib-1 liblzma gdk-x11-3.0 libpcsclite gstreamer-audio-0.10 gstreamer-video-0.10 ncurses form fontconfig dbus-python cairo-xlib-xrender ruby-2.3 'gtk+-3.0' QtUiTools banshee-nowplaying ext2fs harfbuzz-gobject gstreamer-floatcast-0.10 lxappearance zlib xfce4-session-2.0 com_err 'ncurses++' libcgraph gobject-2.0 ruby m17n-db libotf gio-2.0 QtSvg d3d libgvc xdmcp QtNetwork notify-python 'gtk+-x11-2.0' blas renderproto
Choosing modular solver.
Resolving dependencies...
Ready to install base-noprelude-4.7.0.2
Extracting
/home/hvr/.cabal/packages/hackage.haskell.org/base-noprelude/4.7.0.2/base-noprelude-4.7.0.2.tar.gz
to /tmp/cabal-tmp-5991...
Updating base-noprelude.cabal with the latest revision from the index.
Configuring base-noprelude-4.7.0.2...
Waiting for install task to finish...
creating /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup
creating /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist
creating /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup
copy /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/Setup.hs to
/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup.hs
/opt/ghc/7.8.4/bin/ghc --make -fbuilding-cabal-package -odir /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup -hidir /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup -i -i/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2 -optP-include -optP/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup_macros.h -hide-all-packages -package-id Cabal-1.20.0.3-714a9efc373b208cfba0775acac262d0 -package-id base-4.7.0.2-bfd89587617e381ae01b8dd7b6c7f1c1 -package-id filepath-1.3.0.2-25a474a9272ae6260626ce0d70ad1cab /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup.hs -o /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup -threaded -ferror-spans
/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup configure
--verbose=2 --builddir=dist --ghc --with-compiler=/opt/ghc/7.8.4/bin/ghc
--prefix=/home/hvr/.cabal --bindir=/home/hvr/.cabal/bin
--libdir=/home/hvr/.cabal/lib
--libsubdir=x86_64-linux-ghc-7.8.4/base-noprelude-4.7.0.2-BnLIddom9I4GRVKz87SqFV
--libexecdir=/home/hvr/.cabal/libexec --datadir=/home/hvr/.cabal/share
--datasubdir=x86_64-linux-ghc-7.8.4/base-noprelude-4.7.0.2
--docdir=/home/hvr/.cabal/share/doc/x86_64-linux-ghc-7.8.4/base-noprelude-4.7.0.2
--htmldir=/home/hvr/.cabal/share/doc/x86_64-linux-ghc-7.8.4/base-noprelude-4.7.0.2/html
--haddockdir=/home/hvr/.cabal/share/doc/x86_64-linux-ghc-7.8.4/base-noprelude-4.7.0.2/html
--sysconfdir=/home/hvr/.cabal/etc --user
--ipid=base-noprelude-4.7.0.2-BnLIddom9I4GRVKz87SqFV
--dependency=base=base-4.7.0.2-bfd89587617e381ae01b8dd7b6c7f1c1
--disable-tests --exact-configuration --disable-benchmarks
--ghc-option=-ferror-spans
Redirecting build log to {handle:
/home/hvr/.cabal/logs/base-noprelude-4.7.0.2.log}
Failed to install base-noprelude-4.7.0.2
Build log ( /home/hvr/.cabal/logs/base-noprelude-4.7.0.2.log ):
cabal: Entering directory '/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2'
[1 of 1] Compiling Main             ( /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup.hs, /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/Main.o )
Linking /tmp/cabal-tmp-5991/base-noprelude-4.7.0.2/dist/setup/setup ...
unrecognized 'configure' option `--ipid=base-noprelude-4.7.0.2-BnLIddom9I4GRVKz87SqFV'
cabal: Leaving directory '/tmp/cabal-tmp-5991/base-noprelude-4.7.0.2'
World file is already up to date.
base-noprelude-4.7.0.2 failed during the configure step. The exception was:
ExitFailure 1
```

grayjay:
Previously, cabal always chose a Cabal library version to use for a package's
setup script that didn't take setup-depends into account (cabalLibVersion).
cabalLibVersion depended on the cabal-version field, the installed versions of
Cabal, etc., and it was used when setup dependencies were not chosen by the
dependency solver. When a package had a setup-depends, cabalLibVersion was later
ignored in favor of the version chosen by the dependency solver. However,
calculating the variable caused an error when the only suitable Cabal version
was installed during the same install command (issue #3436). cabalLibVersion was
also used to filter the configure flags (issue #3433).

This commit sets cabalLibVersion to the version chosen by the dependency solver,
when possible.

I added an integration test for issue #3436 in the second commit.  I also added two more custom-setup tests in the third commit, but they passed before this change.

EDIT: This fixes #3433, #3436, and #3475.

### DIFF
--- a/cabal-install/Distribution/Client/SetupWrapper.hs
+++ b/cabal-install/Distribution/Client/SetupWrapper.hs
@@ -389,6 +389,18 @@ externalSetupMethod verbosity options pkg bt mkargs = do
       Nothing    -> getInstalledPackages verbosity
                     comp (usePackageDB options') conf
 
+  -- Choose the version of Cabal to use if the setup script has a dependency on
+  -- Cabal, and possibly update the setup script options. The version also
+  -- determines how to filter the flags to Setup.
+  --
+  -- We first check whether the dependency solver has specified a Cabal version.
+  -- If it has, we use the solver's version without looking at the installed
+  -- package index (See issue #3436). Otherwise, we pick the Cabal version by
+  -- checking 'useCabalSpecVersion', then the saved version, and finally the
+  -- versions available in the index.
+  --
+  -- The version chosen here must match the one used in 'compileSetupExecutable'
+  -- (See issue #3433).
   cabalLibVersionToUse :: IO (Version, Maybe ComponentId
                              ,SetupScriptOptions)
   cabalLibVersionToUse =
--- a/cabal-install/Distribution/Client/SetupWrapper.hs
+++ b/cabal-install/Distribution/Client/SetupWrapper.hs
@@ -106,7 +106,7 @@ import Control.Applicative ( (<$>), (<*>) )
 import Data.Monoid         ( mempty )
 #endif
 import Control.Monad       ( when, unless )
-import Data.List           ( foldl1' )
+import Data.List           ( find, foldl1' )
 import Data.Maybe          ( fromMaybe, isJust )
 import Data.Char           ( isSpace )
 import Distribution.Client.Compat.ExecutablePath  ( getExecutablePath )
@@ -389,26 +389,33 @@ externalSetupMethod verbosity options pkg bt mkargs = do
       Nothing    -> getInstalledPackages verbosity
                     comp (usePackageDB options') conf
 
-  cabalLibVersionToUse :: IO (Version, (Maybe ComponentId)
+  cabalLibVersionToUse :: IO (Version, Maybe ComponentId
                              ,SetupScriptOptions)
   cabalLibVersionToUse =
-    case useCabalSpecVersion options of
-      Just version -> do
+    case find (hasCabal . snd) (useDependencies options) of
+      Just (unitId, pkgId) -> do
+        let version = pkgVersion pkgId
         updateSetupScript version bt
-        writeFile setupVersionFile (show version ++ "\n")
-        return (version, Nothing, options)
-      Nothing  -> do
-        savedVer <- savedVersion
-        case savedVer of
-          Just version | version `withinRange` useCabalVersion options
-            -> do updateSetupScript version bt
-                  -- Does the previously compiled setup executable still exist
-                  -- and is it up-to date?
-                  useExisting <- canUseExistingSetup version
-                  if useExisting
-                    then return (version, Nothing, options)
-                    else installedVersion
-          _ -> installedVersion
+        writeSetupVersionFile version
+        return (version, Just unitId, options)
+      Nothing ->
+        case useCabalSpecVersion options of
+          Just version -> do
+            updateSetupScript version bt
+            writeSetupVersionFile version
+            return (version, Nothing, options)
+          Nothing  -> do
+            savedVer <- savedVersion
+            case savedVer of
+              Just version | version `withinRange` useCabalVersion options
+                -> do updateSetupScript version bt
+                      -- Does the previously compiled setup executable still exist
+                      -- and is it up-to date?
+                      useExisting <- canUseExistingSetup version
+                      if useExisting
+                        then return (version, Nothing, options)
+                        else installedVersion
+              _ -> installedVersion
     where
       -- This check duplicates the checks in 'getCachedSetupExecutable' /
       -- 'compileSetupExecutable'. Unfortunately, we have to perform it twice
@@ -424,13 +431,20 @@ externalSetupMethod verbosity options pkg bt mkargs = do
           (&&) <$> setupProgFile `existsAndIsMoreRecentThan` setupHs
                <*> setupProgFile `existsAndIsMoreRecentThan` setupVersionFile
 
+      writeSetupVersionFile :: Version -> IO ()
+      writeSetupVersionFile version =
+          writeFile setupVersionFile (show version ++ "\n")
+
+      hasCabal (PackageIdentifier (PackageName "Cabal") _) = True
+      hasCabal _                                           = False
+
       installedVersion :: IO (Version, Maybe InstalledPackageId
                              ,SetupScriptOptions)
       installedVersion = do
         (comp,    conf,    options')  <- configureCompiler options
         (version, mipkgid, options'') <- installedCabalVersion options' comp conf
         updateSetupScript version bt
-        writeFile setupVersionFile (show version ++ "\n")
+        writeSetupVersionFile version
         return (version, mipkgid, options'')
 
       savedVersion :: IO (Maybe Version)
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -23,6 +23,13 @@ Extra-Source-Files:
   -- Do NOT edit this section manually; instead, run the script.
   -- BEGIN gen-extra-source-files
   tests/IntegrationTests/common.sh
+  tests/IntegrationTests/custom-setup/Cabal-99998/Cabal.cabal
+  tests/IntegrationTests/custom-setup/Cabal-99998/CabalMessage.hs
+  tests/IntegrationTests/custom-setup/Cabal-99999/Cabal.cabal
+  tests/IntegrationTests/custom-setup/Cabal-99999/CabalMessage.hs
+  tests/IntegrationTests/custom-setup/custom-setup/Setup.hs
+  tests/IntegrationTests/custom-setup/custom-setup/custom-setup.cabal
+  tests/IntegrationTests/custom-setup/installs_Cabal_as_setup_dep.sh
   tests/IntegrationTests/custom/custom_dep.sh
   tests/IntegrationTests/custom/custom_dep/client/B.hs
   tests/IntegrationTests/custom/custom_dep/client/Setup.hs
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -27,8 +27,14 @@ Extra-Source-Files:
   tests/IntegrationTests/custom-setup/Cabal-99998/CabalMessage.hs
   tests/IntegrationTests/custom-setup/Cabal-99999/Cabal.cabal
   tests/IntegrationTests/custom-setup/Cabal-99999/CabalMessage.hs
+  tests/IntegrationTests/custom-setup/custom-setup-without-cabal-defaultMain/Setup.hs
+  tests/IntegrationTests/custom-setup/custom-setup-without-cabal-defaultMain/custom-setup-without-cabal-defaultMain.cabal
+  tests/IntegrationTests/custom-setup/custom-setup-without-cabal/Setup.hs
+  tests/IntegrationTests/custom-setup/custom-setup-without-cabal/custom-setup-without-cabal.cabal
   tests/IntegrationTests/custom-setup/custom-setup/Setup.hs
   tests/IntegrationTests/custom-setup/custom-setup/custom-setup.cabal
+  tests/IntegrationTests/custom-setup/custom_setup_without_Cabal_doesnt_allow_Cabal_import.sh
+  tests/IntegrationTests/custom-setup/custom_setup_without_Cabal_doesnt_require_Cabal.sh
   tests/IntegrationTests/custom-setup/installs_Cabal_as_setup_dep.sh
   tests/IntegrationTests/custom/custom_dep.sh
   tests/IntegrationTests/custom/custom_dep/client/B.hs
--- a/cabal-install/tests/IntegrationTests/custom-setup/Cabal-99998/Cabal.cabal
+++ b/cabal-install/tests/IntegrationTests/custom-setup/Cabal-99998/Cabal.cabal
@@ -1,0 +1,8 @@
+name: Cabal
+version: 99998
+build-type: Simple
+cabal-version: >= 1.2
+
+library
+  build-depends: base
+  exposed-modules: CabalMessage
--- a/cabal-install/tests/IntegrationTests/custom-setup/Cabal-99998/CabalMessage.hs
+++ b/cabal-install/tests/IntegrationTests/custom-setup/Cabal-99998/CabalMessage.hs
@@ -1,0 +1,3 @@
+module CabalMessage where
+
+message = "This is Cabal-99998"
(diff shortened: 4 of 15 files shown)


cabal-install/Distribution/Client/SetupWrapper.hs
```diff
@@ -389,26 +389,33 @@ externalSetupMethod verbosity options pkg bt mkargs = do
       Nothing    -> getInstalledPackages verbosity
                     comp (usePackageDB options') conf
 
-  cabalLibVersionToUse :: IO (Version, (Maybe ComponentId)
+  cabalLibVersionToUse :: IO (Version, Maybe ComponentId
                              ,SetupScriptOptions)
   cabalLibVersionToUse =
-    case useCabalSpecVersion options of
-      Just version -> do
+    case find (hasCabal . snd) (useDependencies options) of
```
ezyang: I'd appreciate a comment laying out the logic here, and a reference to the issues that arise when you don't do it correctly. (Yes you can blame, but right now it's not even obvious something nontrivial is going on here.)

ezyang:
LGTM.

23Skidoo:
/cc @edsko 

I think this should also go into the 1.24 branch.

cabal-install/Distribution/Client/SetupWrapper.hs
23Skidoo: Doesn't the solver pass the Cabal version to us as `useCabalVersion`? Can we just that instead?

grayjay: @23Skidoo The [comment](https://github.com/haskell/cabal/blob/c4fc3bde3b2690063a63f65de9787f82bc9bef26/cabal-install/Distribution/Client/SetupWrapper.hs#L134) over the `useCabalVersion` field says that it should be used when `useDependenciesExclusive` is not set, so it isn't always set by the dependency solver.  I used `useDependencies` because SetupWrapper already uses it to check whether the solver has picked a Cabal version: https://github.com/haskell/cabal/blob/c4fc3bde3b2690063a63f65de9787f82bc9bef26/cabal-install/Distribution/Client/SetupWrapper.hs#L629.

23Skidoo: @grayjay I think the comment refers to the fact that if `useDependenciesExclusive` is set, then we should just use those dependencies exclusively. If you look at the code, in both places we set `useDependencies`, we also set `useCabalVersion`.

23Skidoo: I'm just a bit worried that the already complicated logic in this file is becoming more complicated, so I'm thinking about how we can make it simpler.

grayjay: The code still needs to determine whether the solver chose the Cabal version, though, which would be hard to determine from `useCabalVersion`.  If the solver chose a Cabal version then we could use `useCabalVersion`.  But I didn't want to change the behavior in the case that the solver didn't choose any setup dependencies or chose setup dependencies that didn't include Cabal.

23Skidoo: OK, let's merge your patch as-is and refactor the logic some other time.

grayjay: @ezyang I pushed a new commit with comments.

ezyang:
Thanks @grayjay, the docs are much appreciated.

Reading them, I wonder: how does the solver decide which version of Cabal to pick? Clearly it tries to respect setup dependencies, but do any of the other parameters feed into it? Based on what I've seen, I don't think so, but I could be wrong?

cabal-install/Distribution/Client/SetupWrapper.hs
grayjay: @23Skidoo I agree that the logic is becoming complicated.  We could separate out the old and new code paths, as suggested here https://github.com/haskell/cabal/pull/3094#issuecomment-177631386.  Then it would be easier to see which `SetupScriptOptions` apply in each case.

grayjay:
@ezyang AFAIK, the solver only considers the setup dependencies.  It ignores `cabal-version`, so it is possible for the solver to choose a version of Cabal that later refuses to configure the package.  We should probably add a check for packages with a setup-depends and cabal-version that are inconsistent, or take the intersection of the ranges before dependency solving.

ezyang:
Yeah, it seems to me applying `cabal-version` when solving for `Custom` seems like the right thing to do. Doesn't have to be for this patch but something to keep in mind.

ezyang:
OK, so I was trying to document `cabal-lib-version` and I realized that I have no idea what this is supposed to do. What are the intended semantics of this field; e.g., when would I put it in my `cabal.project`?

grayjay:
@ezyang Do you mean `cabal-version`? It is the version of the Cabal specification used by the .cabal file, and it's documented in https://github.com/haskell/cabal/blob/master/Cabal/doc/developing-packages.markdown.

ezyang:
Nope, I'm referring to the `--cabal-lib-version` flag, which gets set to `configCabalVersion`

23Skidoo:
IIRC, `--cabal-lib-version` is aimed at developers and is used to force setup scripts to be compiled against a different Cabal version that'd be picked up automatically. So you can test e.g. that cabal-install doesn't pass unsupported flags to scripts compiled against old versions of Cabal.

ezyang:
OK, but that doesn't answer all my questions:

1. Does it apply only if I have `Custom`, or for `Simple` as well?
2. If there is a `setup-depends` field, does it override the setup depends field? Or not?
3. Is there a way to override this for just one package? At the moment, it's a global field.

Does anyone actually use this? Maybe we should drop it.

grayjay:
> Yeah, it seems to me applying cabal-version when solving for Custom seems like the right thing to do. Doesn't have to be for this patch but something to keep in mind.

I opened an issue, #3751.

Should I merge this and create a PR for 1.24?

23Skidoo:
> Does it apply only if I have `Custom`, or for `Simple` as well?

I think it does, in that case it forces the external setup method.

> If there is a `setup-depends` field, does it override the setup depends field? Or not?

It predates the introduction of `setup-depends`, so I'm not sure what actually happens. I'd expect an error.

> Is there a way to override this for just one package? At the moment, it's a global field.

I don't think so. Normally you use it in a single package directory anyway.

> Does anyone actually use this? Maybe we should drop it.

I haven't used it much lately. I think that one can now use a package with with a setup dependency on Cabal for the same purposes.

grayjay:
@ezyang, @23Skidoo Thanks for the code review!

PR #3754, opened 2016-09-01
stack.yaml/nix: add required build and test tools

ttuegel:
Adding more tools for building with Stack on Nix.


23Skidoo:
Our CI doesn't test Stack support (yet), so you could've used `[ci skip]`.

ttuegel:
> Our CI doesn't test Stack support (yet), so you could've used [ci skip].

D'oh.

Issue #3756, opened 2016-09-01
Package name and user guide documentation


jaccokrijnen:
The [user guide](https://www.haskell.org/cabal/users-guide/developing-packages.html#kinds-of-package-cabal-vs-ghc-vs-system) says

> Cabal package names can use letters, numbers and hyphens, but not spaces. 

But I cannot create a new package named `reproduce-1234`

ezyang:
Thanks. The specific restriction is that any hyphen separated bit must NOT be all numbers, because then it's ambiguous with version numbers. Want to submit a doc patch?

jaccokrijnen:
Sure, can you point me to the right repo/directory?

jaccokrijnen:
Found it

PR #3757, opened 2016-09-01
Update developing-packages.markdown

jaccokrijnen:
Document illegal package names ending in a hyphen and numbers


jaccokrijnen:
Issue #3756

ezyang:
Factual note: foo-123-bar is not allowed either (all numbers disallowed anywhere, not just at the end.)

ezyang:
LGTM.

PR #3765, opened 2016-09-03
Fix  `simpleUsage` example in docs.

erantapaa:


ezyang:
Thank you for the PR!

PR #3769, opened 2016-09-03
Include program args in PackageHash

arybczak:


ezyang:
Good catch. I'll wait for CI to green and merge.

Issue #3768, opened 2016-09-03
Intent to refactor: Add call stacks to our debugging output


ezyang:
Have you ever gotten a log message from Cabal and wondered where it came from? Me too! Here's how we're going to fix it:

1. Add a compat shim for GHC.Stack using http://stackoverflow.com/questions/36369252/is-there-a-way-to-syntactically-simulate-empty-constraints-with-ghc-7-8-1
2. Change Verbosity to support extra "trace this specifically." Add a field for annotating debug messages with where they came from
3. Change implementation of debug/info/etc to print the caller filename/lineno if "annotate location of debug messages" is enabled
4. Profit

This is an intent to refactor.

PR #3777, opened 2016-09-05
Added TODO: eradicateNoParse comments to all uses of bare 'read'

NovaDenizen:
When 'read' is used on invalid input, it raises a "no parse" exception that is
typically very unhelpful to the user.  It is generally more desirable to either
handle the error or produce a more helpful error message.

This actually happens sometimes.  When a user creates a detailed-0.9 test
suite, and that test suite crashes or is killed so it can't write its result
log, the bare read in Distribution.Simple.Test.LibV09.runTest fails and cabal
fails with "no parse".

I could have just written a patch to fix that, but it seemed like a better idea
to find all uses of bare read and see what we can do for all of them.

I have grepped through the repository and I think I've found all uses of bare
'read' that might raise the dreaded 'no parse' exception.  The plan is to
separately patch each of them by either replacing with an alternate Read
function (readMaybe is the most obvious candidate), proving the read will
always succeed, or establishing that producing a 'no parse' exception is the
most desirable behavior.

### DIFF
--- a/Cabal/Distribution/Simple/InstallDirs.hs
+++ b/Cabal/Distribution/Simple/InstallDirs.hs
@@ -366,7 +366,7 @@ type PathTemplateEnv = [(PathTemplateVariable, PathTemplate)]
 -- | Convert a 'FilePath' to a 'PathTemplate' including any template vars.
 --
 toPathTemplate :: FilePath -> PathTemplate
-toPathTemplate = PathTemplate . read
+toPathTemplate = PathTemplate . read -- TODO: eradicateNoParse
 
 -- | Convert back to a path, any remaining vars are included
 --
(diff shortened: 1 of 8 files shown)


NovaDenizen:
..or am I just being too formal here and I should just go ahead and do all of them n ow?  I'm not quite sure what the preferred workflow would be.

23Skidoo:
I guess #3768 is related? Otherwise I'm +1 on replacing calls to `read` with something that throws a nicer exception.

> ..or am I just being too formal here and I should just go ahead and do all of them n ow?

You can do it the way you want.

ezyang:
Just go ahead and fix it! Also, not saying you should block on it, but #3768 might be relevant to your interests.

23Skidoo:
@NovaDenizen I've merged this for now to cut down on the number of open PRs.

Issue #2842, opened 2015-09-26
--independent-goals can produce an invalid install plan


grayjay:
Here is a test case.  It is very fragile, because the failure completely depends on the order that the solver chooses to solve for variables.
https://github.com/grayjay/cabal/commit/bca1eda6851dabc6ae8563ac9315264647c582c3

    targets: D, E, F
    constraints: 
      stanzas A test (unknown source)
      stanzas B test (unknown source)
      stanzas C test (unknown source)
      stanzas C test (unknown source)
      stanzas D test (unknown source)
      stanzas E test (unknown source)
      stanzas F test (unknown source)
    preferences: 
    strategy: PreferLatestForSelected
    [__0] trying: 0.D-1.0.0 (user goal)
    [__1] next goal: 0.C (dependency of 0.D-1.0.0)
    [__1] rejecting: 0.C-2.0.0 (conflict: 0.D => 0.C==1.0.0)
    [__1] trying: 0.C-1.0.0
    [__2] trying: 0.A-1.0.0 (dependency of 0.D-1.0.0)
    [__3] trying: 1.E-1.0.0 (user goal)
    [__4] next goal: 1.C (dependency of 1.E-1.0.0)
    [__4] rejecting: 1.C~>0.C-1.0.0 (conflict: 1.E => 1.C==2.0.0)
    [__4] trying: 1.C-2.0.0
    [__5] trying: 1.B-1.0.0 (dependency of 1.E-1.0.0)
    [__6] trying: 2.F-1.0.0 (user goal)
    [__7] trying: 2.B~>1.B-1.0.0 (dependency of 2.F-1.0.0)
    [__8] trying: 2.C~>1.C-2.0.0 (dependency of 2.B-1.0.0)
    [__9] next goal: 2.A (dependency of 2.F-1.0.0)
    [__9] rejecting: 2.A~>0.A-1.0.0 (dependencies not linked: cannot merge {*1.C-2.0.0,2.C-2.0.0} and {*0.C-1.0.0}; conflict set: 0.C, 1.C, 2.A, 2.B, 2.C)
    [__9] rejecting: 2.A-1.0.0 (multiple instances)
    [__9] fail (backjumping, conflict set: 0.A, 0.C, 1.C, 2.A, 2.B, 2.C)
    [__8] trying: 2.C~>0.C-1.0.0
    [__9] next goal: 2.A (dependency of 2.F-1.0.0)
    [__9] trying: 2.A~>0.A-1.0.0
[...]

PR #3778, opened 2016-09-05
Update some dependency solver comments

grayjay:
I removed the comments about --independent-goals not being ready, because #2842 was fixed, and @kosmikus said that the other comment was added before setup dependencies were finished.  I tried the option just now, and it seemed to work.  It seems like we don't really need it with new-build, though.


ezyang:
All for doc fixes.

PR #3779, opened 2016-09-05
Deduplicate build targets by storing them in a set.

grayjay:
The solver can handle duplicate targets by linking them together, but the log
is more complex, and the result is the same as if the targets were deduplicated
beforehand.

Duplicate targets can occur in a sandbox (issue #3203).


ezyang:
LGTM. Is there an easy test we can add?

grayjay:
Thanks.  I'll add a unit test that checks that each target only appears once in the solver log.

ezyang:
Merge when green.

grayjay:
Thanks!

PR #3782, opened 2016-09-06
Fix #3775.

grayjay:
This commit undoes a change from c0a4860202 that filtered out build-depends
entries that matched the names of executable components.

Issue #3775, opened 2016-09-04
internal error: could not construct a valid install plan


hvr:
**Edited @ezyang**: Repros on several versions of GHC, no base bound needed.

With current HEAD:

```
name:                bug
version:             0
build-type:          Simple
cabal-version:       >=1.10

library
  build-depends:       base, servant-server == 0.7.*
  default-language:    Haskell2010
```

results in the following internal error (same with `cabal install --dry` btw):

```
$ cabal new-build --dry

Resolving dependencies...
internal error: could not construct a valid install plan.
The proposed (invalid) plan contained the following problems:
Package wai-app-static-3.1.6.1 has an invalid configuration, in particular:
the package has a dependency warp >=3.0.11 && <3.3 but no package has been
selected to satisfy it.

Proposed plan:
PreExisting Cabal-1.24.0.0 (Cabal-1.24.0.0)
PreExisting array-0.5.1.1 (array-0.5.1.1)
PreExisting base-4.9.0.0 (base-4.9.0.0)
PreExisting binary-0.8.3.0 (binary-0.8.3.0)
PreExisting bytestring-0.10.8.1 (bytestring-0.10.8.1)
PreExisting containers-0.5.7.1 (containers-0.5.7.1)
PreExisting deepseq-1.4.2.0 (deepseq-1.4.2.0)
PreExisting directory-1.2.6.2 (directory-1.2.6.2)
PreExisting filepath-1.4.1.0 (filepath-1.4.1.0)
PreExisting ghc-boot-th-8.0.1 (ghc-boot-th-8.0.1)
PreExisting ghc-prim-0.5.0.0 (ghc-prim-0.5.0.0)
PreExisting integer-gmp-1.0.0.1 (integer-gmp-1.0.0.1)
PreExisting pretty-1.1.3.3 (pretty-1.1.3.3)
PreExisting process-1.4.2.0 (process-1.4.2.0)
PreExisting rts-1.0 (rts)
PreExisting template-haskell-2.11.0.0 (template-haskell-2.11.0.0)
PreExisting time-1.6.0.1 (time-1.6.0.1)
PreExisting transformers-0.5.2.0 (transformers-0.5.2.0)
PreExisting unix-2.7.2.0 (unix-2.7.2.0)
Configured aeson-0.11.2.1
Configured ansi-terminal-0.6.2.3
Configured ansi-wl-pprint-0.6.7.3
Configured appar-0.1.4
Configured async-2.1.0
[...]

PR #3783, opened 2016-09-06
Add support for executables to DSL, and failing test for #3775

ezyang:
CC @grayjay. The commit which fixes this failing test case is #3782

### DIFF
--- a/cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
@@ -4,6 +4,7 @@ module UnitTests.Distribution.Solver.Modular.DSL (
     ExampleDependency(..)
   , Dependencies(..)
   , ExTest(..)
+  , ExExe(..)
   , ExPreference(..)
   , ExampleDb
   , ExampleVersionRange
@@ -21,6 +22,8 @@ module UnitTests.Distribution.Solver.Modular.DSL (
   , withSetupDeps
   , withTest
   , withTests
+  , withExe
+  , withExes
   , runProgress
   ) where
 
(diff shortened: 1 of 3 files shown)


cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
```diff
@@ -194,6 +200,15 @@ withTests ex tests =
                             | ExTest name deps <- tests]
   in ex { exAvDeps = exAvDeps ex <> testCDs }
 
+withExe :: ExampleAvailable -> ExExe -> ExampleAvailable
+withExe ex test = withExes ex [test]
```
grayjay: test/exe

cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
```diff
@@ -249,13 +265,18 @@ exAvSrcPkg ex =
...
+                                    ++ concatMap snd executables
              , C.condLibrary = Just (mkCondTree
                     (extsLib exts <> langLib mlang <> pcpkgLib pcpkgs <> buildtoolsLib exes)
                                                      disableLib
                                                      (Buildable libraryDeps))
              , C.condSubLibraries = []
-             , C.condExecutables = []
+             , C.condExecutables =
+                 -- Executables not presently disableable
+                 let mkTree = mkCondTree mempty id . Buildable
```
grayjay: I think it would be safer to error here if Buildable isn't implemented.

cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs
```diff
@@ -1,4 +1,7 @@
 {-# LANGUAGE RecordWildCards #-}
+-- | This is a set of unit tests for the dependency solver,
+-- which uses the solver DSL ("UnitTests.Distribution.Solver.Modular.DSL")
+-- to more conveniently create package databases to run the solver tests on.
```
grayjay: :+1:

grayjay:
Thanks for adding this!  LGTM, modulo comments.

grayjay:
Oops!  I forgot about the quickcheck tests.  Sure, you can leave `id` with a comment.

grayjay:
@ezyang When I made another change that caused the Quickcheck tests to fail, I realized that they now need to prevent tests and exes from having the same name.  I made a PR: #3819

PR #3798, opened 2016-09-06
D.C.SrcDist: pass builddir to the setup script.

### LINKED ISSUE #3794: cabal sdist seems to ignore the --builddir option when Build-Type for a package is Custom
I am trying to run `cabal sdist` in a directory for which I do not have write permissions.

If the `Build-Type` in the `.cabal` file is `Simple` then it is sufficient to use `--output-directory` to create the output somewhere I do have write permissions.

If the `Build-Type` is `Custom`, then `cabal sdist` will attempt to build the `Setup.hs`. However, it will fail because it can not write to the `dist` directory. I would imagine that the `--builddir` option would fix that issue -- but it still fails.

```
 $ cabal sdist --builddir=/private/var/folders/_b/wj7ns2190gd3hbz9w0f3m17c0000gn/T/nix-build-cabalsdist.drv-0 --output-directory /nix/store/8li5jw9wbkndvdlpbhi5wzw2x5wynjr8-cabalsdist 
creating /nix/store/8li5jw9wbkndvdlpbhi5wzw2x5wynjr8-cabalsdist
creating dist/setup
creating dist
cabal: dist: permission denied
```

23Skidoo:
Fixes #3794.

### DIFF
--- a/cabal-install/Distribution/Client/SrcDist.hs
+++ b/cabal-install/Distribution/Client/SrcDist.hs
@@ -90,6 +90,7 @@ sdist flags exflags = do
     distPref        = fromFlag (sDistDistPref flags)
     tmpTargetDir    = fromFlagOrDefault (srcPref distPref) (sDistDirectory flags)
     setupOpts       = defaultSetupScriptOptions {
+      useDistPref     = distPref,
       -- The '--output-directory' sdist flag was introduced in Cabal 1.12, and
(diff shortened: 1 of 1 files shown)


ezyang:
OK sure. Since it's the legacy codepath I'm OK if we don't add a test.

23Skidoo:
Also cherry-picked into 1.24.

PR #3818, opened 2016-09-10
Add `reconfigure` command

### LINKED ISSUE #2214: I shouldn't have to manually re-run `cabal configure`

ttuegel:
Fixes #2214 by adding a `reconfigure` command and invoking it whenever necessary. The last configure flags used are saved in a version-independent format so it should never be necessary for the user to reconfigure manually.

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -18,6 +18,7 @@ module Distribution.Client.Setup
     ( globalCommand, GlobalFlags(..), defaultGlobalFlags
     , RepoContext(..), withRepoContext
     , configureCommand, ConfigFlags(..), filterConfigureFlags
+    , configPackageDB', configCompilerAux'
     , configureExCommand, ConfigExFlags(..), defaultConfigExFlags
                         , configureExOptions
     , buildCommand, BuildFlags(..), BuildExFlags(..), SkipAddSourceDepsCheck(..)
@@ -70,11 +71,12 @@ import Distribution.Utils.NubList
 import Distribution.Solver.Types.ConstraintSource
 import Distribution.Solver.Types.Settings
 
-import Distribution.Simple.Compiler (PackageDB)
-import Distribution.Simple.Program
-         ( defaultProgramDb )
+import Distribution.Simple.Compiler ( Compiler, PackageDB, PackageDBStack )
+import Distribution.Simple.Program (ProgramDb, defaultProgramDb)
 import Distribution.Simple.Command hiding (boolOpt, boolOpt')
 import qualified Distribution.Simple.Command as Command
+import Distribution.Simple.Configure
+       ( configCompilerAuxEx, interpretPackageDbFlags )
 import qualified Distribution.Simple.Setup as Cabal
 import Distribution.Simple.Setup
          ( ConfigFlags(..), BuildFlags(..), ReplFlags
@@ -94,6 +96,7 @@ import Distribution.Package
          ( PackageIdentifier, packageName, packageVersion, Dependency(..) )
 import Distribution.PackageDescription
          ( BuildType(..), RepoKind(..) )
+import Distribution.System ( Platform )
 import Distribution.Text
          ( Text(..), display )
 import Distribution.ReadE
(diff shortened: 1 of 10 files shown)


cabal-install/Distribution/Client/Setup.hs
```diff
@@ -416,6 +419,18 @@ filterConfigureFlags flags cabalLibVersion
     -- Cabal < 1.3.10 does not grok the '--constraints' flag.
     flags_1_3_10 = flags_1_10_0 { configConstraints = [] }
 
+configPackageDB' :: ConfigFlags -> PackageDBStack
```
23Skidoo: While you're at it, would be nice to add Haddock comments here to make it more clear what's the difference between these two and the normal versions.

cabal-install/changelog
```diff
@@ -22,6 +22,7 @@
 	verbose GHC output (#3540, #3671).
 	* Changed the default logfile template from
 	'.../$pkgid.log' to '.../$compiler/$libname.log' (#3807).
+	* Added a new command, 'cabal reconfigure' (#2214).
```
23Skidoo: A bit too laconic for my taste :-)

cabal-install/Main.hs
```diff
@@ -1171,21 +998,23 @@ reportAction reportFlags extraArgs globalFlags = do
 runAction :: (BuildFlags, BuildExFlags) -> [String] -> Action
 runAction (buildFlags, buildExFlags) extraArgs globalFlags = do
   let verbosity   = fromFlagOrDefault normal (buildVerbosity buildFlags)
+  (useSandbox, config) <- loadConfigOrSandboxConfig verbosity globalFlags
+  distPref <- findSavedDistPref config (buildDistPref buildFlags)
   let noAddSource = fromFlagOrDefault DontSkipAddSourceDepsCheck
                     (buildOnly buildExFlags)
 
   -- reconfigure also checks if we're in a sandbox and reinstalls add-source
```
23Skidoo: The comment about the useSandbox check is out of date.

cabal-install/Main.hs
```diff
@@ -258,6 +256,7 @@ mainWorker args = topHandler $
       , regularCmd runCommand runAction
       , regularCmd initCommand initAction
       , regularCmd configureExCommand configureAction
+      , regularCmd reconfigureCommand reconfigureAction
```
23Skidoo: Do we really want to make `reconfigure` a non-hidden command? IMO it'll be hard for normal users to figure out what the difference is and when to use it instead of `configure`.

ttuegel: @23Skidoo Do you think there is some documentation I could add to clarify? The command is something I've often wanted; if it's hidden, I don't really see what the point of having it is at all.

23Skidoo: @ttuegel Yes, it'd be nice with some more documentation describing the use cases. In my understanding, since reconfiguring normally happens automatically, users should almost never have to run `reconfigure` manually.

ezyang: Based on conversation with @ttuegel, I think one reason you would call `reconfigure` directly is to modify the set of configuration parameters slightly, while still preserving the old set.

23Skidoo: Thanks, this makes sense. Can we add this example to `cabal help reconfigure` and maybe to the manual?

PR #3819, opened 2016-09-10
Validate InstallPlan in solver quickcheck tests, and update tests to handle executables.

grayjay:
This PR ensures that randomly-generated ComponentDeps don't have tests and exes with the same name.  It also evaluates the solver result to force install plan validation.  Previously, the tests only evaluated enough of the result to determine whether the solver found an install plan, but that install plan could be invalid.


ezyang:
Nice catch, LGTM.

grayjay:
Thanks!

PR #3820, opened 2016-09-11
Add a new field to solver Done nodes, and simplify Explore.exploreLog.

grayjay:
This PR is part of #2917.  The new field is only used for some refactoring in this PR, but #2917 uses it to store the install plan scores.

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Tree.hs
+++ b/cabal-install/Distribution/Solver/Modular/Tree.hs
@@ -67,7 +67,7 @@ data Tree a =
 
     -- | We failed to find a solution in this path through the tree
   | Fail (ConflictSet QPN) FailReason
-  deriving (Eq, Show, Functor)
+  deriving (Eq, Show)
 
 -- | A package option is a package instance with an optional linking annotation
 --
--- a/cabal-install/Distribution/Solver/Modular/Tree.hs
+++ b/cabal-install/Distribution/Solver/Modular/Tree.hs
@@ -35,19 +35,25 @@ type Weight = Double
 
 -- | Type of the search tree. Inlining the choice nodes for now. Weights on
 -- package, flag, and stanza choices control the traversal order.
+--
+-- The tree can hold additional data on 'Done' nodes (type 'd') and choice nodes
+-- (type 'c'). For example, during the final traversal, choice nodes contain the
+-- variables that introduced the choices, and 'Done' nodes contain the
+-- assignments for all variables.
+--
 -- TODO: The weight type should be changed from [Double] to Double to avoid
 -- giving too much weight to preferences that are applied later.
-data Tree a =
+data Tree d c =
     -- | Choose a version for a package (or choose to link)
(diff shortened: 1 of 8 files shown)


cabal-install/Distribution/Solver/Modular/Tree.hs
```diff
@@ -109,74 +109,74 @@ data FailReason = InconsistentInitialConstraints
   deriving (Eq, Show)
 
 -- | Functor for the tree type.
-data TreeF a b =
-    PChoiceF    QPN a                    (WeightedPSQ [Weight] POption b)
-  | FChoiceF    QFN a WeakOrTrivial Bool (WeightedPSQ [Weight] Bool    b)
-  | SChoiceF    QSN a WeakOrTrivial      (WeightedPSQ [Weight] Bool    b)
-  | GoalChoiceF                          (PSQ (Goal QPN) b)
-  | DoneF       RevDepMap
+data TreeF a b c =
```
ezyang: Tell me what the parameters mean!

cabal-install/Distribution/Solver/Modular/Tree.hs
```diff
@@ -37,17 +37,17 @@ type Weight = Double
 -- package, flag, and stanza choices control the traversal order.
 -- TODO: The weight type should be changed from [Double] to Double to avoid
 -- giving too much weight to preferences that are applied later.
-data Tree a =
+data Tree a b =
```
ezyang: Tell me what these parameters mean.

ezyang:
I don't really know what is going on algorithmically, but I can at least ask for docs on the types ;)

cabal-install/Distribution/Solver/Modular/Tree.hs
grayjay: I wasn't sure how to document the parameters, since they don't really have a meaning and just allow different types of data to be stored on the nodes between tree traversals.  #2917 uses `a` for install plan scores, so the type is `()` before the scoring traversal and a type that contains the score afterwards.  I could say that `a` is the type of extra values stored on `Done` nodes, and `b` is the type of values stored on choice nodes.  Would that be clearer, or were you looking for something else?

ezyang: One way to solve this is to point out a few places where we specialize, and describe what the nodes are. Commentary about which type of node the value is stored on also seems helpful; you could even name it `done`, `choice` e.g. to give readers the right idea.

grayjay: Thanks.  I'll update it when I have time.

grayjay: I added comments describing the type variables for `Tree` and `TreeF`.  I also renamed the types `c` and `d`, for choice and done nodes, respectively.  I think that the one-letter names are more readable when they appear several times in a single type signature.

ezyang:
Lovely, thank you.

PR #3828, opened 2016-09-12
Print profiling status when displaying install plan.

ezyang:
CC @grayjay, @hvr

### DIFF
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -78,11 +78,14 @@ import           Distribution.Package
 import qualified Distribution.PackageDescription as PD
 import           Distribution.PackageDescription (FlagAssignment)
 import           Distribution.Simple.Setup (HaddockFlags)
+import qualified Distribution.Simple.Setup as Setup
+import           Distribution.Simple.Command (commandShowOptions)
 
 import           Distribution.Simple.Utils (die, notice, noticeNoWrap, debug)
 import           Distribution.Verbosity
 import           Distribution.Text
 
+import qualified Data.Monoid as Mon
 import qualified Data.Set as Set
 import qualified Data.Map as Map
 import           Data.Map (Map)
@@ -92,6 +95,7 @@ import           Data.Maybe
 import           Data.Either
 import           Control.Exception (Exception(..))
 import           System.Exit (ExitCode(..), exitFailure)
+import qualified System.Process.Internals as Process (translate)
 #ifdef MIN_VERSION_unix
 import           System.Posix.Signals (sigKILL, sigSEGV)
 #endif
@@ -420,6 +424,7 @@ printPlan :: Verbosity -> ProjectBuildContext -> IO ()
 printPlan verbosity
           ProjectBuildContext {
             elaboratedPlan,
+            elaboratedShared,
             pkgsBuildStatus,
             buildSettings = BuildTimeSettings{buildSettingDryRun}
           }
@@ -458,6 +463,7 @@ printPlan verbosity
             " (" ++ maybe "custom" display (compComponentName comp) ++ ")"
             ) ++
       showFlagAssignment (nonDefaultFlags elab) ++
+      showConfigureFlags elab ++
       let buildStatus = pkgsBuildStatus Map.! installedUnitId elab in
       " (" ++ showBuildStatus buildStatus ++ ")"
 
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -483,6 +484,14 @@ printPlan verbosity
     showFlagValue (f, False)  = '-' : showFlagName f
     showFlagName (PD.FlagName f) = f
 
+    showProfiling elab =
+        (if elabProfLib elab
+            then " profile"
+            else "") ++
+        (if elabProfExe elab
+            then " exe-profile"
+            else "")
```
grayjay: What do the different combinations of "profile" and "exe-profile" mean?

ezyang: It's horrible forever.

"profile" means that we will pass `--enable-lib-profiling` to configure, and "exe-profile" means we'll pass `--enable-exe-profiling` to configure. Whether or not these flags *do* anything is a separate matter. ;)

grayjay: Okay, now I understand.  I found it confusing at first, especially when it listed "exe-profile" for a library-only package.  Would it make sense to just list "profile" for either flag, or only show the profiling information at a higher verbosity?

ezyang: Yes it's bad.

Here's another thought; what if the output log message output the flags that we would pass to the configure script (maybe not all of them, but some of the important ones)? That would make things clearer.

grayjay: Yeah, showing the full flag names would be clearer.

ezyang:
OK totally reworked the patch to print configure flags. Read the new commit message.

23Skidoo:
Fails to build on 7.6 and 7.8 due to missing Data.Monoid import.

ezyang:
@grayjay, the new output looks like this:

```
Resolving dependencies...
 - base-orphans-0.5.4 (lib) --enable-library-profiling (requires build)
 - transformers-compat-0.5.1.4 (lib) --enable-library-profiling (requires build)
 - tagged-0.8.5 (lib) --enable-library-profiling (requires build)
 - distributive-0.5.0.2 (lib:distributive) --enable-library-profiling (requires build)
 - test-0.1.0.0 (lib) --enable-profiling (cannot read state cache)
```

How does this look?

hvr:
@ezyang *bikeshedalert* imho the verbatim flags like `--enable-library-profiling` are a bit too verbose for my taste... specifically the `enable` and the `profiling` part seems overly verbose; the interesting information imho is that profiling is e.g. enabled for only libraries, executables or everything... including the profiling level requested I assume?

So maybe something like

```
Resolving dependencies...
 - base-orphans-0.5.4 (lib) (prof: lib[default]) (requires build)
 - transformers-compat-0.5.1.4 (lib) (prof: lib[default]) (requires build)
 - tagged-0.8.5 (lib) (prof: lib[default]) (requires build)
 - distributive-0.5.0.2 (lib:distributive) (prof: lib[default]) (requires build)
 - test-0.1.0.0 (lib) (prof: lib[default] exe[default]) (cannot read state cache)
```

?

or one could merge the profiling information into the component name:


```
Resolving dependencies...
 - base-orphans-0.5.4 (lib[prof:default]) (requires build)
 - transformers-compat-0.5.1.4 (lib[prof:default]) (requires build)
 - tagged-0.8.5 (lib[prof:default]) (requires build)
 - distributive-0.5.0.2 (lib:distributive[prof:default]) (requires build)
 - test-0.1.0.0 (lib[prof:default]) (cannot read state cache)
```

(where default is the profiling level selected via `--library-*detail=level`)

btw, why does `distributive-0.5.0.2 (lib:distributive)` name a qualified `lib:` component name rather than merely `(lib)` like the other entries?

ezyang:
I spent a while agonizing over the format before I came up with the flag idea. The reason was that I wanted the output to be *obvious* in meaning. Looking at the other pieces: enabled components coincide with the targets you pass on command line (except lib, but we promised to fix that), and the configure flags are part of the user visible interface to Setup and cabal configure/install. But a specialized syntax for indicating profiling doesn't.

Given the choice between a specialized syntax and dropping this patch, I think I'd rather drop the patch. What do you think?

23Skidoo:
> Given the choice between a specialized syntax and dropping this patch, I think I'd rather drop the patch. What do you think?

Please don't, I also think that's `--enable-library-profiling` is more clear than `[prof:default]`. The second can be read as either "profiling: default detail level" or "profiling setting: default (i.e. not enabled)".

grayjay:
@ezyang I like the new format.  It's very clear, but not too cluttered.  Thanks!

PR #3830, opened 2016-09-12
Basic cabal lexer for docs

sopvop:
A really basic cabal lexer, but makes code samples a little bit nicer.

I've only converted `nix-local-build.rst` so far.

caballex


23Skidoo:
Merged, thanks!

hvr:
@sopvop currently we have in `conf.py`:

```py
# Syntax highlighting
highlight_language = 'haskell'
```

does this mean we can switch that to `'cabal'` grammar now (most code blocks are either cabal.config or `.cabal` examples, which all follow the same basic grammar)?

PR #3831, opened 2016-09-12
More docs highlighting changes

sopvop:
Also makes 'cabal' a default highlighter.

I think a much prettier highlighter can be made by simplifying the yaml highlighter from pygments, but IMO it's more work than anyone would care to do.


23Skidoo:
@sopvop BTW, since our CI currently doesn't build docs, you can use `[ci skip]` in the commit message body to skip unnecessary builds.

23Skidoo:
Merged, thanks!

Issue #3791, opened 2016-09-06
Miscellaneous Documentation Improvements


linearray:
This issue is to collect a few todos regarding the new documentation:
- [ ] The GHC documentation defines `object types` for documenting recurring elements, such as command line flags. We should explore which object types or other Sphinx features (directives, roles…) would make sense for Cabal.
- [ ] The sphinx search is still sub-par, case in point: "build-depends". It appears that code blocks are not indexed. Perhaps we need a custom Sphinx search adapter for that.
- [ ] The meta-docs (README.md in Cabal/doc) should include steps for Ubuntu/Debian users.
- [ ] There are a few instances where the converted documentation includes formatting in inline literals, an obvious contradiction. Need to find a way to work around that.
- [ ] A few paragraphs have "Strikeout" prepended to them. Find out what to do with that.
- [ ] All pages that consist solely of a TOC, should indicate that in the filename (e.g. `toc-config-and-install.rst`)
- [ ] Lastly, once all or most of the other improvements are complete, it would not hurt to actually read through all the documentation, to spot rendering bugs and because at least some of the information appears to be outdated.

hvr:
The indexing issue for `build-depends` would be somewhat handled by addressing the first issue, c.f.

```py
    app.add_object_type('ghc-flag', 'ghc-flag',
                        objname='GHC command-line option',
                        parse_node=parse_flag,
                        indextemplate='pair: %s; GHC option',
                        doc_field_types=[
                            Field('since', label='Introduced in GHC version', names=['since']),
                            Field('default', label='Default value', names=['default']),
                            Field('static')
                        ])
```

which implies an index-entry template; we want something like that in the context of Cabal at least for `stanza` and `field` and `flag`. 

/cc @ezyang

ezyang:
@linearray, would you like to drive this effort? If so I'll milestone this for 2.0 :)

sopvop:
Simple `app,add_object_type(...)` is not enough for package fields because of duplicates. For example both test and executable section have `main-is` field.
One workaround is to add `pkg-exec-field`, `pkg-test-field` etc. but that leads to a lot of different directives with different ref roles. sad.
The `.. option::` directive works around this by examining state written by `.. program::` directive, similar trick can be used here too, but that requires some code.

About adding more meta-data to .. option:: directive now used to describe setup interface -
the ObjectDescription directive in sphinx does not require adding fields explicitly for simple stuff, you can actually do

~~~.rst
.. option:: --allow-newer

[...]

PR #3836, opened 2016-09-13
Make docs less noisy and referenceable

sopvop:
Makes installing-packages.rst less noisy by replacing most of monospace
markup with `.. option::`. This makes it possible to reference options
from other parts of docs and adds them to index.

Index is no longer empty so it is added to nav menu via template
override in _template folder

[preview](https://sopvop.github.io/cabal/installing-packages.html#programs-used-for-building)
and [index](https://sopvop.github.io/cabal/genindex.html)

### DIFF
--- a/Cabal/doc/conf.py
+++ b/Cabal/doc/conf.py
@@ -15,7 +15,7 @@ version = "1.25"
 
 extensions = ['sphinx.ext.extlinks']
 
-templates_path = ['.templates']
+templates_path = ['_templates']
 source_suffix = '.rst'
 source_encoding = 'utf-8-sig'
 master_doc = 'index'
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -210,7 +210,7 @@ Building and installing a user package
     $ runhaskell Setup.hs install
 
 The package is installed under the user's home directory and is
-registered in the user's package database (``--user``).
+registered in the user's package database (:option:`setup configure --user`).
 
 Installing packages from Hackage
 --------------------------------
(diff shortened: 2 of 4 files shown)


Cabal/doc/installing-packages.rst
```diff
@@ -716,37 +744,53 @@ Paths in the simple build system
...
-+------------------------------+-------------------------------------------------------------+---------------------------+
-| ``--sysconfdir``             | ``$prefix\etc``                                             | ``$prefix/etc``           |
-+------------------------------+-------------------------------------------------------------+---------------------------+
-| ``--htmldir``                | ``$docdir\html``                                            | ``$docdir/html``          |
-+------------------------------+-------------------------------------------------------------+---------------------------+
-| ``--program-prefix``         | (empty)                                                     | (empty)                   |
-+------------------------------+-------------------------------------------------------------+---------------------------+
-| ``--program-suffix``         | (empty)                                                     | (empty)                   |
-+------------------------------+-------------------------------------------------------------+---------------------------+
+.. list-table:: Default installation paths
```
sopvop: A questionable change. But because the table is so wide I think this makes source a bit easier to read and edit.

Cabal/doc/conf.py
```diff
@@ -15,7 +15,7 @@
 
 extensions = ['sphinx.ext.extlinks']
 
-templates_path = ['.templates']
+templates_path = ['_templates']
```
hvr: :-( I like my temp folders hidden....

Cabal/doc/installing-packages.rst
hvr: On smaller window it makes more sense to have the unix column first, as the windows column as one lengthy entry which shifts the unix column out of view otherwise

hvr:
This looks like a good improvement to me, and we should merge it as an intermediate improvement. 

However, I'd like to investigate improving evne more by using semantic markup as outlined in https://github.com/haskell/cabal/issues/3791#issuecomment-244934762

Cabal/doc/conf.py
sopvop: That is a source code folder and not temp/config folder. And naming it `_templates` is standard for sphinx docs.

23Skidoo:
@sopvop BTW, if you connect your GitHub account to your readthedocs.io account, we can add you to cabal.readthedocs.io as admin.

Cabal/doc/conf.py
hvr: fair enough

23Skidoo:
Merged, thanks!

PR #3842, opened 2016-09-14
Even more docs beautification

sopvop:
Now search is [working](https://sopvop.github.io/cabal/search.html?q=build-depends&check_keywords=yes&area=default) for package.cabal fields.

Much more internal references

Some stuff looks [ugly](https://sopvop.github.io/cabal/developing-packages.html#pkg-field-custom-setup-setup-depends)

shinx cabal domain is not documented.

Should this be merged right away, or maybe it is better to push more stuff to this PR first?


ezyang:
The Python script has gotten pretty meaty, so lets get some docs for it.

sopvop:
I'm not happy with the way readthedocs theme renders cabal arguments here
cabalfields

Maybe it would be better to add special directive, which would render like `Default value` fields?
Bikeshedding is welcome

sopvop:
@ezyang I'll document python and write a readme for cabal domain once I've finished with it.

ezyang:
@sopvop, I think "why comments" (as @geofft eloquently describes in https://www.reddit.com/r/rust/comments/5008nf/a_collaborative_effort_to_rewrite_the_kernel_of_redox_with_focus_on_correct/d70sbiv?context=3) would be useful even before the code is "finished", and be unlikely to need to be changed even if you're rewriting code. Things like references to the relevant Sphinx documentation for APIs used, and a high-level description of what we want the "Cabal domain" to do. You don't have to docblock every function (yet, anyway), but a top-level docblock would be useful today.

sopvop:
@ezyang You are right. Now that I've figured out how I want it all to work, I'll add docs in next commit.
In this one I've added fancy [reference](https://sopvop.github.io/cabal/cabal-projectindex.html) for fields and flags.
Now I need to deal with cabal `setup` options and it should be good enough.

Also all of this requires proofreading, my English is far from perfect and I'm doing a lot of this in a hurry, because I might not have time next week.

ezyang:
Nifty! At the end of day, I'd personally prefer it if you merged these doc improvements to master than let it languish on a branch. ;) So if you run out of time, let's just merge it. (But I would like docs! Even short little breadcrumbs that could be expanded.)

sopvop:
I've added some docs to cabaldomain.py and reworked how some directives work to simplify usage.
python code needs more cleanup, and custom `.. option::` directives are not implemented yet.

I may find some time to work on this next week, but I think this can be merged for now.

ezyang:
Thank you, the docs are much appreciated. I'm not terribly familiar with Sphinx, so the immediate comparison between domains and object flew over my head. At least link to http://www.sphinx-doc.org/en/stable/domains.html and http://www.sphinx-doc.org/en/stable/glossary.html#term-object . And maybe even before you start comparing, you should just say what the module does "we add support for the following ReST directives, and also collating them together into an index" or something like that.

PR #3855, opened 2016-09-17
Add examples to help text for reconfigure command

ttuegel:
@23Skidoo We discussed improving the help text for the reconfigure command. Is this what you had in mind?

IIUC, the man page is generated from the command-line help text, so this should update the man page, too.

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -508,11 +508,18 @@ reconfigureCommand
     { commandName         = "reconfigure"
     , commandSynopsis     = "Reconfigure the package if necessary."
     , commandDescription  = Just $ \pname -> wrapText $
-         "Run `configure` with the most recently used flags and append FLAGS. "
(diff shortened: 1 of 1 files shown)


ezyang:
Looks OK to me.

Question: If I do `cabal reconfigure --ghc-option=-blah`, does `-blah` get appended to the pre-existing GHC options or does it overwrite them?

23Skidoo:
>Question: If I do `cabal reconfigure --ghc-option=-blah`, does `-blah` get appended to the pre-existing GHC options or does it overwrite them?

IIUC, this will be equivalent to calling `configure --ghc-option=-foo --ghc-option=-bar`, i.e. the former.

ttuegel:
> IIUC, this will be equivalent to calling configure --ghc-option=-foo --ghc-option=-bar, i.e. the former.

That's right; calling `reconfigure` with any flag is exactly the same as calling `configure` with the original command line, plus `reconfigure`'s flags appended to the end.

The snide answer is: `reconfigure` does whatever the flag's `Monoid` instance does. :smiley:

PR #3876, opened 2016-09-20
Add helper for Generic NFData instances

hvr:
I needed the `NFData` instance for `Dependency`... which made me realise that
writing the implementation for the also needed `VersionRange` instance by hand
was too tedious, so here we are with a `genericRnf` compat helper...


ezyang:
Looks fine, but tell me again why you need this? ;)

hvr:
@ezyang for the same reason we have those other `NFData` instances... convenience during debugging/benchmarking/profiling/forcing lazy-eval/squeezing out potential space leaks... and avoiding orphan instances :-)

23Skidoo:
Merged, thanks!

PR #3877, opened 2016-09-20
Whitelist branches for travis to build

phadej:


phadej:
FWIW, I'd rather not have any dev branches in this repository, but this will prevent some builds.

Something like that would be good for appveyor too, but I'm not familiar with its configuration

ezyang:
OK. The classic reason to push branches to the Cabal repo is so that other people can edit them. Things are better with the "allow maintainers to edit PRs" so maybe we can ask people not to push branches, or always qualify them with a prefix.

23Skidoo:
Merged, thanks!

23Skidoo:
@ezyang Usually when we do that, we also create a PR. If there's a long-lived feature branch without a corresponding PR, it can be whitelisted.

ezyang:
Well, you're right, we shouldn't build them anyway :)

PR #3879, opened 2016-09-21
Don't solve for executables in legacy code path.

### LINKED ISSUE #3875: HEAD: Strange warning about "unknown package" (related to build-tools?)
With cabal-install HEAD (@hvr's 1.25+git20160918.1.c9e046e~xenial) I get this warning when I call configure:

```
# cabal configure
Resolving dependencies...
Warning: solver failed to find a solution:
trying: ingest-0.1.0.0 (user goal)
unknown package: ingest-alex-exe.alex (dependency of ingest-0.1.0.0)
fail (backjumping, conflict set: ingest, ingest-alex-exe.alex)
Trying configure anyway.
Configuring ingest-0.1.0.0...
```

This `ingest-alex-exe.alex` dependency is really odd; I have no idea where that is coming from. The only mention of alex in my `.cabal` file is

```
  build-tools:         alex >= 3.1 && < 3.2
```

(and there are some `.x` files in the project).

ezyang:
There is a bug in `cabal configure`'s invocation of the solver in
Distribution/Client/Configure.hs:

    standardInstallPolicy
        installedPkgIndex
        (SourcePackageDb mempty packagePrefs)
        [SpecificSourcePackage localPkg]

We can see that the solver is given an EMPTY source package database.
This is because we assume that everything you need from cabal configure
is taken from the installed package index.

But this is NOT true for executables, which never have an entry in the
installed package index. So we SHOULD NOT solve for
executables in the legacy codepath, since there isn't anything useful we
can do with the info anyway.  This gets toggled using a new solver
parameter SolveExecutables.

I didn't bother with a test because this will be obsoleted by
nix-local-build.

Fixes #3875

### DIFF
--- a/cabal-install/Distribution/Client/Configure.hs
+++ b/cabal-install/Distribution/Client/Configure.hs
@@ -39,6 +39,7 @@ import Distribution.Package (PackageId)
 import Distribution.Client.JobControl (Lock)
 
 import qualified Distribution.Solver.Types.ComponentDeps as CD
+import           Distribution.Solver.Types.Settings
 import           Distribution.Solver.Types.ConstraintSource
 import           Distribution.Solver.Types.LabeledPackageConstraint
 import           Distribution.Solver.Types.OptionalStanza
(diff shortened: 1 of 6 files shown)


23Skidoo:
Merged, thanks!

PR #3895, opened 2016-09-25
Refactor cabal-install with custom Prelude

hvr:
This helps avoid 95% of CPP uses, and also helps reduce a bit of the `import`-boilerplate
for `Data.Maybe` et al.

But most importantly, this should reduce the risk of wasting time when a PR works for
GHC 8.0.1, but then Travis complains that some pre-AMP `import` needs to be added
for e.g. GHC 7.8.4

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -315,6 +315,7 @@ library
     Distribution.Compat.Exception
     Distribution.Compat.Graph
     Distribution.Compat.Internal.TempFile
+    Distribution.Compat.Prelude
     Distribution.Compat.ReadP
     Distribution.Compat.Semigroup
     Distribution.Compat.Stack
@@ -415,7 +416,6 @@ library
     Distribution.Compat.Binary
 
   other-modules:
-    Distribution.Compat.Prelude
     Distribution.Compat.CopyFile
     Distribution.Compat.GetShortPathName
     Distribution.Compat.MonadFail
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -315,7 +315,7 @@ library
     Distribution.Compat.Exception
     Distribution.Compat.Graph
     Distribution.Compat.Internal.TempFile
-    Distribution.Compat.Prelude
+    Distribution.Compat.Prelude.Internal
     Distribution.Compat.ReadP
     Distribution.Compat.Semigroup
     Distribution.Compat.Stack
@@ -419,6 +419,7 @@ library
     Distribution.Compat.CopyFile
     Distribution.Compat.GetShortPathName
     Distribution.Compat.MonadFail
+    Distribution.Compat.Prelude
     Distribution.GetOpt
     Distribution.Lex
     Distribution.Simple.GHC.Internal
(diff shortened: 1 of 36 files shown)


phadej:
`D.C.Prelude` exports `for_` and `traverse_`, should we favour then and not `forM_` and `mapM_`?

hvr:
@phadej I'm not against it, but I've ran out of steam... so I'll leave this to a future refactoring by somebody else...  :-)

Cabal/Cabal.cabal
```diff
@@ -315,6 +315,7 @@ library
     Distribution.Compat.Exception
     Distribution.Compat.Graph
     Distribution.Compat.Internal.TempFile
+    Distribution.Compat.Prelude
```
ezyang: So, when the Distribution.Compat.Prelude was originally added @phadej specifically didn't export it because he didn't want any clients to depend on it. I'm pretty ambivalent about exposing it; I definitely don't want Custom setups depending on it unless we commit it as an API. But I understand why it makes your job here a bit easier. Not sure what to do!

phadej: I agree on this comment. I'd rather depend on `base-compat` for `cabal-install`, which would solve most of this PR goals.

hvr: I actually started the refactoring using `base-compat`, but then realised we'd end up with an additional dependency to keep track of in bootstrap.sh as well as redundant/duplicate maintenance work keeping `Distribution.Compat.Client.Prelude` and `Distribution.Compat.Prelude` in sync in order to have a common vocabulary between `Cabal` and `cabal-install` which are essentially the same project spread across two packages.

hvr: btw, if we want to discourage `Setup.hs` using that module, we could add a reexporting exposed  module `Distribution.Compat.Prelude.Internal`, and hide its documentation in Haddock....

ezyang: I'm not that found of Haddock-hiding, myself ;) And it doesn't help much once a snippet of code starts getting cargo culted around. I guess we could just slap a prominent comment saying that the API is NOT stable and should not be relied upon.

hvr: @ezyang one could also use a module-level warning (in the exposed `.Internal` module; Cabal's own code would still use the non-.Internal one) so people really notice the prominent comment

ezyang: There isn't a good way to suppress it from cabal-install's build though :(

hvr: @ezyang we only need to import that `.Internal` module once, in the `Distribution.Client.Compat.Prelude` module, so we can tolerate a single `{-# OPTIONS_GHC -Wno-deprecations #-}` in there, can't we?

ezyang: Yes, I think that's fine, assuming that the module reexport drops the warning (I can't imagine it doesn't?!) Do we have a plan? (If this works out, we could probably do this to other modules in Cabal too!)

hvr: @ezyang I'll create a commit showcasing this  and will add it to this PR asap

hvr: @ezyang see 99d4d4795c6ad6edbf38dd4a47a6d423259d471d

PR #3906, opened 2016-09-28
Fix reexported-modules display mangling.

ezyang:
sep allows for horizontal separation, which is absolutely
not what we want.  Use hcat instead.  Thanks @stepcut
for reporting.

I'd like this to merge to 1.24 too, if the bug is present there.

### DIFF
--- a/Cabal/Distribution/InstalledPackageInfo.hs
+++ b/Cabal/Distribution/InstalledPackageInfo.hs
@@ -181,11 +181,11 @@ data ExposedModule
 
 instance Text ExposedModule where
     disp (ExposedModule m reexport) =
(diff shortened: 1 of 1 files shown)


stepcut:
I don't thinks this patch is quite right. Won't `hcat` join the `Doc` pieces with out a space between them? Perhaps `hsep` would be better? That is just a guess on my part -- I have not confirmed it.

ezyang:
Oops, I changed too many seps. Ugh.

ezyang:
We should merge this to 1.24, @23Skidoo

23Skidoo:
Merged into 1.24 in #3928.

PR #3918, opened 2016-10-01
Handle the case when min-version binary macro is not defined.

ezyang:
Fixes bootstrapping by GHC build system.

CC @hvr

### DIFF
--- a/Cabal/Distribution/Utils/ShortText.hs
+++ b/Cabal/Distribution/Utils/ShortText.hs
@@ -25,6 +25,23 @@ import Data.String (IsString(..))
 # endif
 #endif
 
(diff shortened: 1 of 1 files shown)


hvr:
@ezyang that's surprising... I assumed that if `MIN_VERSION_bytestring` is defined, so is `MIN_VERSION_binary` :confused:

hvr:
after more discussion on IRC, this is an effect of how GHC's buildsystem bootstraps Cabal, which doesn't rely on using `binary` as a package. So I guess this hack/workaround is fine for now. @ezyang maybe add a short comment pointing out why this hack is needed? :)

PR #3921, opened 2016-10-01
Make `AbiHash` opaque & `ShortText`-backed

hvr:


ezyang:
I suppose it is worth remarking that if you really wanted to, you could represent GHC ABI hashes as a 128-bit number for MOAR EFFICIENCY.

ezyang:
LGTM.

PR #3924, opened 2016-10-02
Don't solve for executables in legacy install codepath.

### LINKED ISSUE #3912: Assertion failure in D.C.Install
With current master, when I do:

```
$ cabal get log
$ cd log-0.5.3
$ cabal sandbox init
$ cabal install --only-dep
```
I get the following assertion failure:

```
$ cabal install --only-dep
Resolving dependencies...
/home/refold/code/haskell/log-0.5.3/.cabal-sandbox
Configuring base64-bytestring-1.0.0.1...
Configuring base-orphans-0.5.4...
[...]
Building nats-1.1.1...
Installed nats-1.1.1
Configuring network-2.6.3.1...
Configuring old-locale-1.0.0.7...
Assertion failed
assert, called at ./Distribution/Client/Install.hs:1427:34 in
main:Distribution.Client.Install
```

This does not seem to happen with 1.24 (I checked that I compiled with assertions on).

ezyang:
In 9e99b3f49911028edc1c2be8152002a6a3b7ad3c I turned of executable
solving for legacy configure, but I forgot to turn it off for legacy
install too, which lead to assertion failures.

Fixes #3912.

### DIFF
--- a/cabal-install/Distribution/Client/Install.hs
+++ b/cabal-install/Distribution/Client/Install.hs
@@ -107,6 +107,7 @@ import Distribution.Client.JobControl
 
 import qualified Distribution.Solver.Types.ComponentDeps as CD
 import           Distribution.Solver.Types.ConstraintSource
+import           Distribution.Solver.Types.Settings
 import           Distribution.Solver.Types.LabeledPackageConstraint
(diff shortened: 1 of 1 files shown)


23Skidoo:
Thanks!

PR #3925, opened 2016-10-02
GHC 8.1 compatibility fixes.

ezyang:
It seems that GHC HEAD is now unable to infer the correct
type for "f . g" when g has a higher-rank type.  So eta-expand
all such occurrences to keep the build working.


ezyang:
GHC tracking bug: https://ghc.haskell.org/trac/ghc/ticket/12652

PR #3928, opened 2016-10-02
Fix reexported-modules display mangling.

ezyang:
sep allows for horizontal separation, which is absolutely
not what we want.  Use hcat instead.  Thanks @stepcut
for reporting.


23Skidoo:
Merged, thanks!

PR #3937, opened 2016-10-04
Make the initial '~/.cabal/config' comment more obvious.

23Skidoo:
Some new users think that the `-- Cabal library version: 1.25.0.0` string is a setting they should uncomment.


ezyang:
OK.

PR #3939, opened 2016-10-04
Lower library compatibility version to 1.25, since 1.25 is dev series…

ezyang:
… for 2.0.

Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>

PR #3940, opened 2016-10-04
Lower implicit setup-depends bound

bgamari:
As long as the tree contains Cabal-1.25 the implicit setup-depends bound
should be Cabal<1.25 to ensure that we don't attempt to build custom
Setup.hs's against the new Cabal, which is at this point thoroughly
incompatible with 1.24 and earlier.

This bound should likely be again lifted to Cabal<2.0 once the in-tree
Cabal version has been bumped to 2.0.


bgamari:
This is a duplicate of #3939.

PR #3944, opened 2016-10-06
Pass -this-component-id to GHC when necessary.

ezyang:
Previously GHC inferred it off of -this-unit-id but we've
modified things so that GHC makes no assumptions about the
format of instantiated unit ids. In that case, pass it in
explicitly!

### DIFF
--- a/Cabal/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/Distribution/Simple/Program/GHC.hs
@@ -90,6 +90,10 @@ data GhcOptions = GhcOptions {
   -- is.  This only gets set when 'ghcOptInstantiatedWith' is non-empty
   ghcOptThisComponentId :: Flag ComponentId,
 
+  -- | How the requirements of the package being compiled are to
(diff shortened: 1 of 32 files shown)


Cabal/Distribution/Simple/Program/GHC.hs
```diff
@@ -83,6 +84,12 @@ data GhcOptions = GhcOptions {
   -- (we need to handle backwards compatibility.)
   ghcOptThisUnitId   :: Flag String,
 
+  -- | GHC doesn't make any assumptions about the format of
+  -- definite unit ids, so when we are instantiating a package it
+  -- needs to be told explicitly what the component being instantiated
+  -- is.  This only gets set when 'ghcOptInstantiatedWith' is non-empty
+  ghcOptThisComponentId :: Flag ComponentId,
+
   ghcOptInstantiatedWith :: [(ModuleName, OpenModule)],
```
23Skidoo: This is a good opportunity to add a Haddock comment to `ghcOptInstantiatedWith` as well.

PR #3949, opened 2016-10-08
Fix Stack build

quchen:
The hackage-security dependency in stack.yaml (0.5.2.1) is incompatible with what cabal-install requires (>=0.5.2.2). Bumping the version to that lower bound resolves the dependency issue.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -12,7 +12,7 @@ extra-deps:
 - cryptohash-sha256-0.11.100.1
 - ed25519-0.0.5.0
 - exceptions-0.8.2.1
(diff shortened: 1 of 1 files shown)


phadej:
let's see what travis says

quchen:
@phadej Travis might fail due to #3950, but that seems to be entirely unrelated.

phadej:
I see. @quchen do you want to add stack based travis job?

quchen:
@phadej Could try to do that, yes!

23Skidoo:
CI doesn't test building with Stack at all right now, so I'm merging this.

Issue #3570, opened 2016-07-17
Print final conflict set when dependency solver fails after an exhaustive search


grayjay:
@kosmikus suggested showing this message at default verbosity because it is often more relevant than the summarized log.

EDIT:
The new message could go here: https://github.com/haskell/cabal/blob/c116df4742264641f281e9480731cdd8313b309a/cabal-install/Distribution/Solver/Modular/Log.hs#L86  This is where the solver prints a message after an exhaustive search.  It already has access to the final conflict set, `cs`.

I would add a unit test to Solver.hs. Here is an example of a test that checks for a specific error message:
https://github.com/haskell/cabal/blob/c116df4742264641f281e9480731cdd8313b309a/cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs#L45

ezyang:
@grayjay, if PR is welcome, could you edit your message to say where the code is that you would update to make this change, and how you would test it? Thanks!

grayjay:
I realized that the hardest part of this issue is coming up with helpful wording.  The task isn't as well-defined as I thought, so I removed the label.

grayjay:
Added in #3960.

Issue #941, opened 2012-05-24
Incorrect error messages for non-existing dependencies


bos:
(Imported from [Trac #951](http://hackage.haskell.org/trac/hackage/ticket/951), reported by guest on 2012-05-15)

When the build-depends-field contains non-existing packages, cabal-install sometimes does not report that these packages could not be found but complains about other constraints being unsolvable.

The attached cabal-file contains a dependency for a non-existing package called "bogus". This is what I get:

<pre class="wiki">$ cabal install --only-dependencies --dry-run
Resolving dependencies...
cabal: Could not resolve dependencies:
next goal: foo (user goal)
rejecting: foo-1.0 (global constraint requires ==0.1.0.0)
trying: foo-0.1.0.0
trying: heist-0.8.0 (dependency of foo-0.1.0.0)
trying: transformers-0.3.0.0 (dependency of heist-0.8.0)
next goal: mtl (dependency of heist-0.8.0)
rejecting: mtl-2.1.1, 2.1 (conflict: heist =&gt; mtl&gt;=2.0 &amp;&amp; &lt;2.1)
rejecting: mtl-2.0.1.0, 2.0.0.0 (conflict: transformers==0.3.0.0, mtl =&gt;
transformers==0.2.*)
rejecting: mtl-1.1.1.1, 1.1.1.0, 1.1.0.2, 1.1.0.1, 1.1.0.0, 1.0 (conflict:
heist =&gt; mtl&gt;=2.0 &amp;&amp; &lt;2.1)
</pre>When removing "bogus" from the build-depends, cabal-install is able to come up with a working install plan.

Used versions:

<pre class="wiki">$ cabal --version
cabal-install version 0.14.0
using version 1.14.0 of the Cabal library
</pre>

grayjay:
I have an idea for improving the summary in cases like this.  We could run the solver once to try to find a solution.  Then, if it fails, we could run it again, but have it prefer goals from the first run's top-level conflict set.  The solver would use the log from the second run, up to the first error, as the summary.

The errors from the second run could be more relevant, because they are likely to come from the final conflict set, which explains why the whole tree has no solution.  In this example, the conflict set is probably {bogus, foo}.  So the error message would show the solver trying `foo` and then `bogus`.

grayjay:
See #4139 for another example of a misleading first error message.

grayjay:
[...]

PR #3960, opened 2016-10-08
Print final conflict set when search is exhaustive

fmthoma:
Pull request for #3570.

The final confilct set is printed along with the conflict count of each conflict, if available.

Open for discussion:
- The original error message is thrown away, we currently only show the conflilct set. Should we restore it and print both?
- The conflict count is not actually a useful metric. Should we expose it, or should we rescale it to something more useful to the user?

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Explore.hs
+++ b/cabal-install/Distribution/Solver/Modular/Explore.hs
@@ -68,8 +68,6 @@ backjump (EnableBackjumping enableBj) var initial xs =
 
 type ConflictSetLog = RetryLog Message (ConflictSet QPN, ConflictMap)
 
-type ConflictMap = Map (Var QPN) Int
-
 getBestGoal :: ConflictMap -> P.PSQ (Goal QPN) a -> (Goal QPN, a)
 getBestGoal cm =
   P.maximumBy
@@ -174,7 +172,4 @@ backjumpAndExplore :: EnableBackjumping
                    -> CountConflicts
                    -> Tree d QGoalReason -> Log Message (Assignment, RevDepMap)
 backjumpAndExplore enableBj countConflicts =
-    toLog . exploreLog enableBj countConflicts . assign
-  where
-    toLog :: RetryLog step fail done -> Log step done
-    toLog = toProgress . mapFailure (const ())
+    toProgress . exploreLog enableBj countConflicts . assign
--- a/cabal-install/Distribution/Solver/Modular/Log.hs
+++ b/cabal-install/Distribution/Solver/Modular/Log.hs
@@ -34,7 +34,7 @@ data Exhaustiveness = Exhaustive | BackjumpLimitReached
 logToProgress :: Maybe Int -> Log Message a -> Progress String String a
 logToProgress mbj l = let
                         es = proc (Just 0) l -- catch first error (always)
-                        ms = useFirstError (proc mbj l)
+                        ms = proc mbj l
                       in go es es -- trace for first error
                             (showMessages (const True) True ms) -- run with backjump limit applied
   where
@@ -52,19 +52,6 @@ logToProgress mbj l = let
     proc (Just n) (Step x@(Failure _  Backjump) xs) = Step x (proc (Just (n - 1)) xs)
     proc mbj'     (Step x                       xs) = Step x (proc mbj'           xs)
 
-    -- Sets the conflict set from the first backjump as the final error in case of a
-    -- non-exhaustive search.
-    useFirstError :: Progress Message (Exhaustiveness, ConflictSet QPN, ConflictMap) b
-                  -> Progress Message (Exhaustiveness, ConflictSet QPN, ConflictMap) b
-    useFirstError = replace Nothing
-      where
-        replace _       (Done x)                              = Done x
-        replace _       (Fail (Exhaustive,           cs, cm)) = Fail (Exhaustive, cs, cm)
-        replace cs'     (Fail (BackjumpLimitReached, cs, cm)) = -- Prefer first error over later error.
-                                                                Fail (BackjumpLimitReached, fromMaybe cs cs', cm)
-        replace Nothing (Step x@(Failure cs Backjump) xs)     = Step x $ replace (Just cs) xs
-        replace cs'     (Step x                       xs)     = Step x $ replace cs' xs
-
     -- The first two arguments are both supposed to be the log up to the first error.
     -- That's the error that will always be printed in case we do not find a solution.
     -- We pass this log twice, because we evaluate it in parallel with the full log,
@@ -73,15 +60,16 @@ logToProgress mbj l = let
     --
     -- The third argument is the full log, ending with either the solution or the
     -- exhaustiveness and first conflict set.
-    go :: Progress Message a b
-       -> Progress Message a b
+    go :: Progress Message (Exhaustiveness, ConflictSet QPN, ConflictMap) b
+       -> Progress Message (Exhaustiveness, ConflictSet QPN, ConflictMap) b
        -> Progress String (Exhaustiveness, ConflictSet QPN, ConflictMap) b
        -> Progress String String b
-    go ms (Step _ ns) (Step x xs)           = Step x (go ms ns xs)
-    go ms r           (Step x xs)           = Step x (go ms r  xs)
-    go ms _           (Fail (exh, cs, cm))  = Fail $
+    go ms (Step _ ns)        (Step x xs)           = Step x (go ms ns xs)
+    go ms r                  (Step x xs)           = Step x (go ms r  xs)
+    go ms (Step _ ns)        r                     = go ms ns r
+    go ms (Fail (_, cs', _)) (Fail (exh, cs, cm))  = Fail $
         "Could not resolve dependencies:\n" ++
-        unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs) True) False ms) ++
+        unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs') True) False ms) ++
         case exh of
             Exhaustive ->
                 "After searching the rest of the dependency tree exhaustively, "
@@ -92,4 +80,9 @@ logToProgress mbj l = let
                 "change with --max-backjumps or try to run with --reorder-goals).\n"
               where currlimit (Just n) = "currently " ++ show n ++ ", "
                     currlimit Nothing  = ""
-    go _  _           (Done s)              = Done s
+    go _  _                  (Done s)              = Done s
+    go _  (Done _)           (Fail _)              = Fail $
+        -- Should not happen: Second argument is the log up to first error,
+        -- third one is the entire log. Therefore it should never happen that
+        -- the second log finishes with 'Done' and the third log with 'Fail'.
+        "Could not resolve dependencies; something strange happened."
--- a/cabal-install/Distribution/Solver/Modular/Log.hs
+++ b/cabal-install/Distribution/Solver/Modular/Log.hs
@@ -84,8 +84,8 @@ logToProgress mbj l = let
         unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs) True) False ms) ++
         case exh of
             Exhaustive ->
-                "Dependency tree exhaustively searched.\n" ++
-                "I've had most trouble fulfilling the following goals: "
+                "After searching the rest of the dependency tree exhaustively, "
+                ++ "these were the goals I've had most trouble fulfilling: "
                 ++ CS.showCSWithFrequency cm cs
             BackjumpLimitReached ->
                 "Backjump limit reached (" ++ currlimit mbj ++
--- a/cabal-install/Distribution/Solver/Modular/Log.hs
+++ b/cabal-install/Distribution/Solver/Modular/Log.hs
@@ -39,25 +39,25 @@ logToProgress mbj l = let
     -- and ignores repeated backjumps. If proc reaches the backjump limit, it truncates
     -- the 'Progress' and ends it with the last conflict set. Otherwise, it leaves the
     -- original success result or replaces the original failure with 'Nothing'.
-    proc :: Maybe Int -> Log Message b -> Progress Message (Bool, ConflictSet QPN) b
+    proc :: Maybe Int -> Log Message b -> Progress Message (Bool, ConflictSet QPN, ConflictMap) b
     proc _        (Done x)                          = Done x
-    proc _        (Fail (cs, _cm))                  = Fail (True, cs)
+    proc _        (Fail (cs, cm))                   = Fail (True, cs, cm)
     proc mbj'     (Step x@(Failure cs Backjump) xs@(Step Leave (Step (Failure cs' Backjump) _)))
       | cs == cs'                                   = Step x (proc mbj'           xs) -- repeated backjumps count as one
-    proc (Just 0) (Step   (Failure cs Backjump)  _) = Fail (False, cs)
+    proc (Just 0) (Step   (Failure cs Backjump)  _) = Fail (False, cs, mempty)
     proc (Just n) (Step x@(Failure _  Backjump) xs) = Step x (proc (Just (n - 1)) xs)
     proc mbj'     (Step x                       xs) = Step x (proc mbj'           xs)
 
     -- Sets the conflict set from the first backjump as the final error, and records
     -- whether the search was exhaustive.
-    useFirstError :: Progress Message (Bool, ConflictSet QPN) b
-                  -> Progress Message (Bool, ConflictSet QPN) b
+    useFirstError :: Progress Message (Bool, ConflictSet QPN, ConflictMap) b
+                  -> Progress Message (Bool, ConflictSet QPN, ConflictMap) b
     useFirstError = replace Nothing
       where
         replace _       (Done x)                          = Done x
-        replace cs'     (Fail (exh, cs))                  = -- 'Nothing' means backjump limit not reached.
+        replace cs'     (Fail (exh, cs, cm))              = -- 'Nothing' means backjump limit not reached.
                                                             -- Prefer first error over later error.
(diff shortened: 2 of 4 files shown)


cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -40,25 +40,25 @@ logToProgress mbj l = let
     -- and ignores repeated backjumps. If proc reaches the backjump limit, it truncates
```
ezyang: This comment needs updating.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -40,25 +40,25 @@ logToProgress mbj l = let
...
-    useFirstError :: Progress Message (Maybe (ConflictSet QPN)) b
-                  -> Progress Message (Bool, Maybe (ConflictSet QPN)) b
+    useFirstError :: Progress Message (Bool, ConflictSet QPN) b
+                  -> Progress Message (Bool, ConflictSet QPN) b
     useFirstError = replace Nothing
       where
         replace _       (Done x)                          = Done x
-        replace cs'     (Fail cs)                         = -- 'Nothing' means backjump limit not reached.
+        replace cs'     (Fail (exh, cs))                  = -- 'Nothing' means backjump limit not reached.
                                                             -- Prefer first error over later error.
```
ezyang: This comment needs updating (something about doing something specifically when you exhaust search space?)

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -72,17 +72,16 @@ logToProgress mbj l = let
...
-                                              "Could not resolve dependencies:\n" ++
-                                              unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs) True) False ms) ++
-                                              (if exh then "Dependency tree exhaustively searched.\n"
-                                                      else "Backjump limit reached (" ++ currlimit mbj ++
-                                                               "change with --max-backjumps or try to run with --reorder-goals).\n")
-                                                  where currlimit (Just n) = "currently " ++ show n ++ ", "
-                                                        currlimit Nothing  = ""
+    go ms _           (Fail (exh, cs))      = Fail $ "Could not resolve dependencies:\n" ++ if exh
+                                                then "Dependency tree exhaustively searched.\n" ++
+                                                         "Final conflict set is: " ++ CS.showCS cs
```
ezyang: Not something you need to do for this patchset, but someday we should explain to mere mortals what a conflict set is :)

ezyang:
Since this is just user visible business, it doesn't look that easy to automated test. But it would be nice if the commit messages included some sample output, i.e., "What would I expect to see with this new change?" For example, something like in da388222533c52018fe94ad7951cb2b46540dec8

fmthoma:
@ezyang Thank you for your review comments! I have addressed them in my updates. Also I have changed the message to a slightly less technical »I've had most trouble fulfilling the following goals: …«.

ezyang:
Thanks! Really helpful. CI failure is real though; try building locally with -Werror.

I wonder if we should print the conflict counts in non-verbose mode. They don't seem that useful for someone who doesn't know what they are.

fmthoma:
@ezyang: I had this discussion with @kosmikus as well. On one hand, the absolute numbers do not provide any useful metric. On the other hand, their relative values do provide a tendency how severe a conflict is (there is a difference between `package1 (10000), package2 (9999)` and `package1 (10000), package2 (3)` that is not neglectable). So here are some suggestions:
* Print the numbers as they are
* Defer them to verbose mode
* Normalize them to something more useful (e.g. to percentage of all conflicts)
* Use them to further reduce the output, e.g. cut off all packages that contribute less than 1..10% of all conflicts.

The warning should be gone now, I overlooked a `Control.Applicative` import.

ezyang:
I'm OK with leaving the counts, I just want to open up a discussion.

Actually, while we're at it, why don't we talk about what the conflict counts mean in the first place. My understanding is they is they specify how many times a package shows up in a "conflict set", which is a set of packages that have incompatible constraints that cause a solution to be unsat. Is this accurate?  This makes me wonder: the solver is supposed to bail out early when it encounters a conflict. So the actual conflict counts are going to be heavily dependent on the order the solver attempted to solve packages in.

fmthoma:
> So the actual conflict counts are going to be heavily dependent on the order the solver attempted to solve packages in.

Yes, this shows up for example in #3961 where the counts are heavily dependent on how we accumulate them while backjumping. So these numbers are not an exact science, but they give a good indication nevertheless.

I personally would probably prefer from a user perspective to have percentages (»This package was involved in about n% of all conflicts«) and a sensible cutoff.

phadej:
@grayjay use

```hs
import Prelude ()         
import Distribution.Client.Compat.Prelude
```

(related to `Control.Applicative`)

cabal-install/Distribution/Solver/Modular/Explore.hs
```diff
@@ -67,8 +67,6 @@ backjump (EnableBackjumping enableBj) var initial xs =
```
grayjay: I just realized that the line above updates the conflict map even when `countConflicts` is false.  It should probably be something like `if countConflicts then (cs, updateCM initial cm) else (cs, cm)` to avoid creating a thunk from the conflict map when it isn't used.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -20,11 +22,13 @@ import qualified Distribution.Solver.Modular.ConflictSet as CS
 -- Represents the progress of a computation lazily.
 --
 -- Parameterized over the type of actual messages and the final result.
-type Log m a = Progress m () a
+type Log m a = Progress m (ConflictSet QPN, ConflictMap) a
 
 messages :: Progress step fail done -> [step]
 messages = foldProgress (:) (const []) (const [])
 
+data Exhaustive = Exhaustive | NotExhaustive
```
grayjay: I think `Exhaustiveness` would be clearer.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -39,26 +43,27 @@ logToProgress mbj l = let
...
+    -- Sets the conflict set from the first backjump as the final error in case of a
+    -- non-exhaustive search.
+    useFirstError :: Progress Message (Exhaustive, ConflictSet QPN, ConflictMap) b
+                  -> Progress Message (Exhaustive, ConflictSet QPN, ConflictMap) b
     useFirstError = replace Nothing
       where
         replace _       (Done x)                          = Done x
-        replace cs'     (Fail cs)                         = -- 'Nothing' means backjump limit not reached.
+        replace _       (Fail (Exhaustive,    cs, cm))    = Fail (Exhaustive, cs, cm)
+        replace cs'     (Fail (NotExhaustive, cs, cm))    = -- Backjump limit not reached.
```
grayjay: Should this be "Backjump limit reached"?

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -72,17 +77,21 @@ logToProgress mbj l = let
...
-                                              "Could not resolve dependencies:\n" ++
-                                              unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs) True) False ms) ++
-                                              (if exh then "Dependency tree exhaustively searched.\n"
-                                                      else "Backjump limit reached (" ++ currlimit mbj ++
-                                                               "change with --max-backjumps or try to run with --reorder-goals).\n")
-                                                  where currlimit (Just n) = "currently " ++ show n ++ ", "
-                                                        currlimit Nothing  = ""
+    go ms _           (Fail (exh, cs, cm))  = Fail $
+        "Could not resolve dependencies:\n" ++
+        unlines (messages $ showMessages (L.foldr (\ v _ -> v `CS.member` cs) True) False ms) ++
```
grayjay: It looks like `cs` is the first conflict set when the backjump limit is reached and the last conflict set when the search is exhaustive.  Shouldn't this line always filter the short log using the first conflict set?  The final conflict set might not contain the first error, which would cause the error to be filtered out.

Another option is to remove the filtered log altogether, to avoid showing two possibly unrelated errors.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -72,17 +77,21 @@ logToProgress mbj l = let
...
+        case exh of
+            Exhaustive ->
+                "Dependency tree exhaustively searched.\n" ++
+                "I've had most trouble fulfilling the following goals: "
+                ++ CS.showCSWithFrequency cm cs
```
grayjay: I wonder if we should only show the counts at higher verbosity, since there are cases where they could be misleading.  For example, the solver could immediately stop after finding a missing dependency, which would give the missing dependency a count of 1, even though it is part of a minimal conflict set.

cabal-install/Distribution/Solver/Modular/Log.hs
fmthoma: Indeed, I'll the `data Exhaustive` to `data Exhaustiveness = Exhaustive | BackjumpLimitReached` so the comment is now unnecessary.

fmthoma:
@grayjay, @ezyang: I think it makes sense to show the conflict counts only on a higher verbosity level. However, this requires either passing the verbosity to the solver (which is kind of an intrusive change), or to change the return type of the solver (which is also quite intrusive). I am working on this, but I would rather suggest we merge this pull request without this change, and I open another one with the conflict counts only on verbose.
@kosmikus: Do you agree with @grayjay and @ezyang to show the counts only in verbose mode?

fmthoma:
@grayjay, @ezyang: I think it makes sense to show the conflict counts only on a higher verbosity level. However, this requires either passing the verbosity to the solver (which is kind of an intrusive change), or to change the return type of the solver (which is also quite intrusive). I am working on this, but I would rather suggest we merge this pull request without this change, and I open another one with the conflict counts only on verbose.
@kosmikus: Do you agree with @grayjay and @ezyang to show the counts only in verbose mode?

23Skidoo:
> I would rather suggest we merge this pull request without this change, and I open another one with the conflict counts only on verbose.

Fine with me.

cabal-install/Distribution/Solver/Modular/Explore.hs
fmthoma: Good catch, changed it in the PR.

cabal-install/Distribution/Solver/Modular/Log.hs
grayjay: The solver always prints `ms`, which is the log to the first error.  It truncates the log on the line `es = proc (Just 0) l -- catch first error (always)` above.

> BTW, what is the original rationale anyway for showing the first rather than the last error in case the backjump limit is reached?

According to the discussion in #941, the first error was chosen because the solver has made fewer compromises at that point, and the conflicts are more likely to be relevant.  Also, I don't think that there is an obvious place to start printing the summarized log leading up to the final error.  The log wouldn't always include the top of the tree.

cabal-install/Distribution/Solver/Modular/Explore.hs
grayjay: @fmthoma Actually, the code that I linked to above isn't enough to prevent space leaks.  I had to fix it in https://github.com/haskell/cabal/pull/4122/commits/41cd78641926bfa05f9aa6d8e75e56be0f8ec18c.

`getFirstGoal` doesn't evaluate the conflict map unless there is more than one goal, so it is possible for a thunk to develop even with --count-conflicts.  Using `seq` is more reliable.

cabal-install/Distribution/Solver/Modular/Log.hs
fmthoma: Okay, so to summarize:
* We always print the first error encountered.
* In case the backjump limit is reached, we print the first conflict set (which coincides with the first error).
* For an exhaustive search, we print the final conflict set (which may not coincide with the first error).

So this could be solved
1. either by choosing to print the first rather than the final conflict set also in the exhaustive case (what would this imply?),
2. or by printing the error only in the non-exhaustive case and thus avoiding the problem (in fact, in an early version of the PR, @kosmikus and I had printed the error message only in the non-exhaustive case, but we changed it),
3. or by printing the last error in the non-exhaustive case.

My understanding of the issue is still somewhat incomplete, but it seems to me that using the reasoning from #941 also applies to the exhaustive case, so option (1) looks to me like the soundest one?

grayjay: I like option (2).

> 1. either by choosing to print the first rather than the final conflict set also in the exhaustive case (what would this imply?),

I think that the first conflict set isn't very useful, because it comes from the first node where the solver backjumps, which is near the leaves of the tree.  It contains the same information as the first error.  However, the final conflict set is assigned to the root of the tree (in the exhaustive case), and it summarizes all conflicts in the tree.  It means that any solution to the dependency problem would need to assign a value that isn't available in the search tree to at least one of the variables in the conflict set.

The solver's conflict sets differ from the ones in [Modular lazy search for Constraint Satisfaction Problems] (http://dx.doi.org/10.1017/S0956796801004051) because they can also contain variables that are below the current node.  That difference makes the final conflict set more useful.

> 3 . or by printing the last error in the non-exhaustive case.

I'm not sure how this makes the behavior more consistent, or did you mean the exhaustive case?

fmthoma: Yes, I meant the exhaustive case :-) (fixed it in my comment above.)

Then let's go with option (2), I will do the change.

fmthoma: Pushed the change. Unfortunately, the failing tests point out a problem with this approach: We don't know _why_ a package causes a problem (which might be trivial in case of a package conflict, but in the case of a missing unknown package, that information is quite valuable).

fmthoma: How about a message conveying that the error does not need to be in line with the final conflict set:

```
> cabal install --dry-run cabal-install-1.24.0.0 cabal-install-1.22.9.0
Resolving dependencies...
next goal: cabal-install (user goal)
rejecting: cabal-install-1.24.0.0 (constraint from user target requires
==1.22.9.0)
rejecting: cabal-install-1.22.9.0 (constraint from user target requires
==1.24.0.0)
rejecting: cabal-install-1.22.8.0, cabal-install-1.22.7.0,
cabal-install-1.22.6.0, cabal-install-1.22.5.0, cabal-install-1.22.4.0,
cabal-install-1.22.3.0, cabal-install-1.22.2.0, cabal-install-1.22.0.1,
cabal-install-1.22.0.0, cabal-install-1.20.2.0, cabal-install-1.20.1.0,
cabal-install-1.20.0.6, cabal-install-1.20.0.5, cabal-install-1.20.0.4,
cabal-install-1.20.0.3, cabal-install-1.20.0.2, cabal-install-1.20.0.1,
cabal-install-1.20.0.0, cabal-install-1.18.2.0, cabal-install-1.18.1.0,
cabal-install-1.18.0.8, cabal-install-1.18.0.7, cabal-install-1.18.0.6,
cabal-install-1.18.0.5, cabal-install-1.18.0.4, cabal-install-1.18.0.3,
cabal-install-1.18.0.2, cabal-install-1.18.0.1, cabal-install-1.18.0,
cabal-install-1.16.1.0, cabal-install-1.16.0.2, cabal-install-1.16.0.1,
cabal-install-1.16.0, cabal-install-0.14.1, cabal-install-0.14.0,
cabal-install-0.10.2, cabal-install-0.10.0, cabal-install-0.8.2,
cabal-install-0.8.0, cabal-install-0.6.4, cabal-install-0.6.2,
cabal-install-0.6.0, cabal-install-0.5.2, cabal-install-0.5.1,
cabal-install-0.5.0, cabal-install-0.4.0 (constraint from user target requires
==1.22.9.0)
After searching the rest of the dependency exhaustively, these were the goals
I had most trouble fulfilling: cabal-install (49)
```

grayjay: That looks good.  Should it be "the rest of the dependencies"?

grayjay: That looks good.  So I think the only thing left is changing the summarized log so that it is always filtered with the first conflict set.

fmthoma: Please have a look at the last commit. Now the summarized log should be filtered with the first conflict set. This involved removing `useFirstError` and changing the traversal of the log in the `go` function, making `go` partial again, so I hope I didn't break anything.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -70,19 +60,29 @@ logToProgress mbj l = let
     --
     -- The third argument is the full log, ending with either the solution or the
     -- exhaustiveness and first conflict set.
```
grayjay: This is now the final conflict set.

cabal-install/Distribution/Solver/Modular/Log.hs
```diff
@@ -20,48 +21,37 @@ import qualified Distribution.Solver.Modular.ConflictSet as CS
...
-                  -> Progress Message (Bool, Maybe (ConflictSet QPN)) b
-    useFirstError = replace Nothing
-      where
-        replace _       (Done x)                          = Done x
-        replace cs'     (Fail cs)                         = -- 'Nothing' means backjump limit not reached.
-                                                            -- Prefer first error over later error.
-                                                            Fail (isNothing cs, cs' <|> cs)
-        replace Nothing (Step x@(Failure cs Backjump) xs) = Step x $ replace (Just cs) xs
-        replace cs'     (Step x                       xs) = Step x $ replace cs' xs
-
```
grayjay: Nice!

23Skidoo:
@grayjay @fmthoma Feel free to merge once the CI is green if you think that this PR is ready to go in.

grayjay:
It's green, so I'll merge it.  Thanks, @fmthoma!

PR #3961, opened 2016-10-08
Fix conflict counting bug

fmthoma:
Encountered during the implementation of #3570.

We used to miss a lot of conflicts during counting in `logBackjump`.
Counting initial conflicts here drastically reduces search time in some
cases.
Not sure whether counting the current conflicts (`cs`) may be even
better than counting the initial conflicts (`initial`) ⇒ Benchmarking by @kosmikus?

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Explore.hs
+++ b/cabal-install/Distribution/Solver/Modular/Explore.hs
@@ -60,7 +60,10 @@ backjump (EnableBackjumping enableBj) var initial xs =
           | otherwise                            = f (csAcc `CS.union` cs) cm'
 
     logBackjump :: ConflictSet QPN -> ConflictMap -> ConflictSetLog a
-    logBackjump cs cm = failWith (Failure cs Backjump) (cs, cm)
+    logBackjump cs cm = failWith (Failure cs Backjump) (cs, updateCM initial cm)
+                                   -- 'intial' instead of 'cs' here ---^
+                                   -- since we do not want to double-count the
+                                   -- additionally accumulated conflicts.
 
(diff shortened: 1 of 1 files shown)


ezyang:
Thanks for the fix, @fmthoma ! I'm not too familiar with conflict counting (@grayjay, maybe you can comment?) but is there any possible solver test we could add to test this fix?

grayjay:
@ezyang Conflict counting is just a heuristic to improve performance, so it would be hard to test.  This PR just applies it more consistently.  I'll merge it when the build passes.

kosmikus:
@ezyang This is a performance-related fix, not a correctness one. While it may change solver results (because it can change goal order, and therefore affect which of many possible install plans is ultimately found), it should not affect correctness at all, and seems difficult (not impossible) to test for. I'll think about it, but I think we shouldn't delay the merge for this.

cabal-install/Distribution/Solver/Modular/Explore.hs
```diff
@@ -60,7 +60,10 @@ backjump (EnableBackjumping enableBj) var initial xs =
           | otherwise                            = f (csAcc `CS.union` cs) cm'
 
     logBackjump :: ConflictSet QPN -> ConflictMap -> ConflictSetLog a
-    logBackjump cs cm = failWith (Failure cs Backjump) (cs, cm)
+    logBackjump cs cm = failWith (Failure cs Backjump) (cs, updateCM initial cm)
```
fmthoma: We should benchmark `updateCM initial cm` versus `updateCM cs cm`. At least in the examples I tried, the latter gives even better results in terms of performance. On the other hand, @kosmikus has reservations that using `cs` here leads to double-counting some conflicts.

PR #3962, opened 2016-10-08
Add Distribution.Compat.DList

phadej:
In new parser there are few places where it can be used, so we can win few percents of performance...

### DIFF
--- a/Cabal/Distribution/Compat/DList.hs
+++ b/Cabal/Distribution/Compat/DList.hs
@@ -1,0 +1,36 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Compat.DList
+-- Copyright   :  (c) Ben Gamari 2015-2019
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-dev@haskell.org
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- A very simple difference list.
+module Distribution.Compat.DList (
+    DList,
+    runDList,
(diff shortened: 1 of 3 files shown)


ezyang:
LGTM. But one question: exposed or hidden? :)

phadej:
maybe it can be hidden for now

Cabal/Distribution/Compat/DList.hs
```diff
@@ -0,0 +1,36 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Compat.DList
+-- Copyright   :  (c) Ben Gamari 2015-2019
```
23Skidoo: Should be 2016 instead of 2019? Though if it was up to me, I'd just ditch copyright headers from all modules in favour of the AUTHORS file (which is referred to from `{Cabal,cabal-install}.cabal`).

phadej:
changed to be hidden

Cabal/Distribution/Compat/DList.hs
phadej: copied from `Cabal/Distribution/Lex.hs`

Yeah, attributing file to a single person nowadays is bending the reality quite a lot

23Skidoo: OK, let's merge, I can tackle the removal of `copyright` headers separately at some point in the future.

23Skidoo:
Merged, thanks!

Issue #3951, opened 2016-10-08
Add Stack-based CI


quchen:
I think Cabal and Stack should be buildable with each other, but since #3949 was not detected by CI, there seems to be no CI test for building Cabal with Stack.

23Skidoo:
Patches welcome! If you do this, please use a separate matrix entry for Stack.

hvr:
Generally a good idea! But it ought to be marked as `allow_failures`

quchen:
I can’t assign myself, but I’ll try working on it later.

phadej:
Seems I cannot add you as a contributor, but @23Skidoo or @hvr probably can.

23Skidoo:
> Seems I cannot add you as a contributor, but @23Skidoo or @hvr probably can.
Already done!

quchen:
Status: `stack build` works, `stack test` still fails.

phadej:
@quchen on stack test is problematic because on env vars, try `--no-run-tests` and run them manually...

quchen:
@phadej Yes, `stack test` does not work so well yet. The PR is for `stack build` only for now.

PR #3966, opened 2016-10-09
Add CI target for building Cabal using Stack

quchen:
Resolves haskell/cabal#3951

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,15 +58,18 @@ matrix:
      os: osx
    - env: GHCVER=8.0.1 SCRIPT=bootstrap
      os: osx
-  allow_failures:
-   - env: GHCVER=head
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -58,9 +58,18 @@ matrix:
...
+     os: linux
+     addons:
+        apt:
+            packages:
+                - libgmp-dev
+
   allow_failures:
    - env: GHCVER=head
      sudo: required
+   - env: GHCVER=via-stack SCRIPT=stack STACKAGE_RESOLVER=lts
```
23Skidoo: I think we can just remove the `allow_failures` section since we're not testing with GHC HEAD.

23Skidoo:
Merged, thanks!

PR #3973, opened 2016-10-11
Show Version returning "mkVersion [...]"

phadej:
With tests,

as an extra improved `NubList` properties (it had a typo, so added few properties there as well)


23Skidoo:
Looks OK.

PR #3980, opened 2016-10-14
Reimpelement gen-extra-source-files in Haskell

phadej:
- No need to add other-modules by hand
- Works on OSX (.sh didn't)
- Found missed files!


Cabal/misc/gen-extra-source-files.hs
```diff
@@ -0,0 +1,106 @@
...
+import System.Environment                    (getArgs, getProgName)
+import System.FilePath                       (takeExtension, takeFileName)
+import System.Process                        (readProcess)
+
+import qualified Distribution.ModuleName as ModuleName
+import qualified System.IO               as IO
+
+main' :: FilePath -> IO ()
+main' fp = do
+    -- Raad cabal file, so we can determine test modules
```
ezyang: Read

phadej:
meta test passes now.

PR #3982, opened 2016-10-14
Store interfaces files separately from library object files 

christiaanb:
These two flags indicate where the two distinct aspects of a
Haskell library end up.

`--binlibsubdir`: The subdirectory of `--libdir` where the binary library files (.so/.a/.dll/.dylib) get installed.
`--hidir`: The directory where the interface files (.hi) get installed.

The reason we want to do this is because we want dynamic libraries
to all end up in a shared directory to reduce start-up times of
the run-time linker. However, we still want the .hi to end up
in one directory per package.

We cannot repurpose --libsubdir to take over the role of what
--commonlibdir is doing because then we would run into trouble
with Setup.hs files build against an older Cabal. If we were
to repurpose --libsubdir, then all the object and interface
files would end up in a single shared directory for all
packages when using an older Cabal.

After several attempts (#3955,#3968,#3979), this should be the definitive solution for: https://ghc.haskell.org/trac/ghc/ticket/12479

EDIT: `--binlibsubdir` was originally called `--commonsubdir` when this PR was first created.

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -897,9 +897,9 @@ installDirsOptions =
       libdir (\v flags -> flags { libdir = v })
       installDirArg
 
-  , option "" ["commonlibdir"]
+  , option "" ["binlibsubdir"]
       "subdirectory of libdir in which binary libraries are installed"
-      commonlibdir (\v flags -> flags { commonlibdir = v })
+      binlibsubdir (\v flags -> flags { binlibsubdir = v })
       installDirArg
 
   , option "" ["hidir"]
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -897,9 +897,14 @@ installDirsOptions =
       libdir (\v flags -> flags { libdir = v })
       installDirArg
 
-  , option "" ["libsubdir"]
-      "subdirectory of libdir in which libs are installed"
-      libsubdir (\v flags -> flags { libsubdir = v })
+  , option "" ["commonlibdir"]
+      "subdirectory of libdir in which the object files of libs are installed"
+      commonlibdir (\v flags -> flags { commonlibdir = v })
+      installDirArg
+
+  , option "" ["hidir"]
+      "installation directory for library interface files"
+      hidir (\v flags -> flags { hidir = v })
       installDirArg
 
   , option "" ["libexecdir"]
@@ -935,6 +940,13 @@ installDirsOptions =
   , option "" ["sysconfdir"]
       "installation directory for configuration files"
       sysconfdir (\v flags -> flags { sysconfdir = v })
+      installDirArg
+
+  , option "" ["libsubdir"]
+      ("subdirectory of libdir in which libs are installed." ++
+       "Only has an effect on Setup files build against Cabal < 1.25"
+      )
+      libsubdir (\v flags -> flags { libsubdir = v })
       installDirArg
   ]
   where
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -898,7 +898,7 @@ installDirsOptions =
       installDirArg
 
   , option "" ["commonlibdir"]
-      "subdirectory of libdir in which the object files of libs are installed"
+      "subdirectory of libdir in which binary libraries are installed"
       commonlibdir (\v flags -> flags { commonlibdir = v })
       installDirArg
 
@@ -944,7 +944,7 @@ installDirsOptions =
 
   , option "" ["libsubdir"]
       ("subdirectory of libdir in which libs are installed." ++
-       "Only has an effect on Setup files build against Cabal < 1.25"
+       "Only has an effect on Setup files built against Cabal < 1.25"
       )
       libsubdir (\v flags -> flags { libsubdir = v })
       installDirArg
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -600,13 +600,22 @@ package:
     ``$version``, ``$compiler``, ``$os``, ``$arch``, ``$abi``,
     ``$abitag``
 
+.. option:: --hidir=dir
+
+    Interface files (.hi) of libraries are installed here.
+
+    In the simple build system, *dir* may contain the following path
+    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$commonlibdir``,
+    ``$pkgid``, ``$pkg``, ``$version``, ``$compiler``, ``$os``,
+    ``$arch``, ``$abi``, ``$abitag``
+
 .. option:: --libexecdir=dir
 
     Executables that are not expected to be invoked directly by the user
     are installed here.
 
     In the simple build system, *dir* may contain the following path
-    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$libsubdir``,
+    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$commonlibdir``,
     ``$pkgid``, ``$pkg``, ``$version``, ``$compiler``, ``$os``,
     ``$arch``, ``$abi``, ``$abitag``
 
@@ -615,7 +624,7 @@ package:
     Architecture-independent data files are installed here.
 
     In the simple build system, *dir* may contain the following path
-    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$libsubdir``,
+    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$commonlibdir``,
     ``$pkgid``, ``$pkg``, ``$version``, ``$compiler``, ``$os``,
     ``$arch``, ``$abi``, ``$abitag``
 
@@ -624,21 +633,20 @@ package:
     Installation directory for the configuration files.
 
     In the simple build system, *dir* may contain the following path
-    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$libsubdir``,
+    variables: ``$prefix``, ``$bindir``, ``$libdir``, ``$commonlibdir``,
     ``$pkgid``, ``$pkg``, ``$version``, ``$compiler``, ``$os``,
     ``$arch``, ``$abi``, ``$abitag``
 
 In addition the simple build system supports the following installation
 path options:
 
-.. option:: --libsubdir=dir
+.. option:: --commonlibdir=dir
 
     A subdirectory of *libdir* in which libraries are actually
     installed. For example, in the simple build system on Unix, the
-    default *libdir* is ``/usr/local/lib``, and *libsubdir* contains the
-    package identifier and compiler, e.g. ``mypkg-0.2/ghc-6.4``, so
-    libraries would be installed in
-    ``/usr/local/lib/mypkg-0.2/ghc-6.4``.
+    default *libdir* is ``/usr/local/lib``, and *commonlibdir* contains the
+    ABI, e.g. ``x86_64-linux-8.0.1``, so libraries would be installed in
+    ``/usr/local/lib/x86_64-linux-8.0.1``.
 
     *dir* may contain the following path variables: ``$pkgid``,
     ``$pkg``, ``$version``, ``$compiler``, ``$os``, ``$arch``, ``$abi``,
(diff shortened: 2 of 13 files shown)


ilovezfs:
@christiaanb seems to work
```
==> Summary
🍺  /usr/local/Cellar/git-annex/6.20161012: 105 files, 87.4M, built in 21 minutes 39 seconds
```
and
```
All 281 tests passed (192.81s)
```

ilovezfs:
This appears to fail with new-build https://gist.github.com/ilovezfs/ff5539ab4b6b16ab9a832ff7c9476284

christiaanb:
@ilovezfs what you mean by "it fails on new-build" is "this patch doesn't solve the problems reported in GHC issue 12497 when using new-build"? Or that things are actually broken by this patch that used to work before?

ilovezfs:
I mean that this PR is an incomplete solution to GHC 12479 https://ghc.haskell.org/trac/ghc/ticket/12479 not that it causes a regression.

Is there no hope of a common solution that will address non-new-build and new-build alike?

If there is a prospect for a common solution, then it should be understood that the change in this PR may need to be backed out in favor of that common solution at some point in the near future.

If there is no such prospect, then no harm, no foul in treating them as separate problems even though GHC issue 12479 is the underlying cause for both.

CC @cartazio @uniqmartin

23Skidoo:
> the change in this PR may need to be backed out in favor of that common solution at some point in the near future.

If it's true, I'd prefer to get it right the first time, since backing this change out will result

phadej:
IMHO `new-build` is more or less `cabal-install` issue (or even `ghc-pkg?`). It's a bit like trying to fix https://github.com/commercialhaskell/stack/issues/2577 also in this PR.

I'd like to hear @dcoutts or someone who knows on how new-build store works, to comment on this.

christiaanb:
@phadej: yes these flags allow you to place library objectfiles (.so/.a/.dll/.dylib) in another directory than the interface files (.hi). That's all this PR basically does:
* Add flags so we can put .hi files somewhere else than .so/.a/.dll/.dylib flags
* By default, put .so/.a/.dll/.dylib in a shared directory (per package database).

So yes, to fix new-build and stack with regards to  https://ghc.haskell.org/trac/ghc/ticket/12479 should just be to teach them these new flags, and make sure that library object files end up in a shared directory.

ezyang:
Yeah, the fix for new-build should not be too difficult. Just passing the new flag.

cartazio:
1) do we want to put the object code in these locations, or merely symlinks
(ignoring the complication that symlinks don't exist on windows I think?)

2) I thought Duncan had a concern with having different dylib builds put
into a shared directory

3) possibly doesn't matter: but I seem to recall that many file systems
have ad hoc limits to the number of files/folders stored in single
directory?  Or would that require several orders of magnitude more dynamic
libraries than all of hackage built by a particular ghc version?

23Skidoo:
> Or would that require several orders of magnitude more dynamic libraries than all of hackage built by a particular ghc version?

Not exactly, but yes. See [here](http://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory) and [here](http://avid.force.com/pkb/articles/en_US/troubleshooting/en250749). HFS+ and FAT32 have the lowest limits on the number of files per directory (32K and 65K, respectively), for other filesystems it is practically unlimited.

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -936,6 +941,13 @@ installDirsOptions =
       "installation directory for configuration files"
       sysconfdir (\v flags -> flags { sysconfdir = v })
       installDirArg
+
+  , option "" ["libsubdir"]
+      ("subdirectory of libdir in which libs are installed." ++
+       "Only has an effect on Setup files build against Cabal < 1.25"
```
ezyang: s/build/built/

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -897,9 +897,14 @@ installDirsOptions =
       libdir (\v flags -> flags { libdir = v })
       installDirArg
 
-  , option "" ["libsubdir"]
-      "subdirectory of libdir in which libs are installed"
-      libsubdir (\v flags -> flags { libsubdir = v })
+  , option "" ["commonlibdir"]
+      "subdirectory of libdir in which the object files of libs are installed"
```
ezyang: The object files aren't installed here, right? Only the .so libraries.

Cabal/doc/installing-packages.rst
```diff
@@ -690,6 +698,22 @@ path options:
     ``$pkg``, ``$version``, ``$compiler``, ``$os``, ``$arch``, ``$abi``,
     ``$abitag``
 
+.. option:: --libsubdir=dir
+
+    For use with Setup.hs files build against a version of Cabal prior to 1.25.
```
ezyang: s/build/built/

Cabal/doc/installing-packages.rst
```diff
@@ -690,6 +698,22 @@ path options:
...
+    For later versions of Cabal, this flag is basically deprecated, and you
+    should use ``--commonlibdir=dir``.
```
ezyang: With later versions of Cabal, you should prefer :option:\`--commonlibdir\` instead.

Cabal/doc/installing-packages.rst
```diff
@@ -624,21 +633,20 @@ package:
...
 
     A subdirectory of *libdir* in which libraries are actually
     installed. For example, in the simple build system on Unix, the
-    default *libdir* is ``/usr/local/lib``, and *libsubdir* contains the
-    package identifier and compiler, e.g. ``mypkg-0.2/ghc-6.4``, so
-    libraries would be installed in
-    ``/usr/local/lib/mypkg-0.2/ghc-6.4``.
+    default *libdir* is ``/usr/local/lib``, and *commonlibdir* contains the
+    ABI, e.g. ``x86_64-linux-8.0.1``, so libraries would be installed in
+    ``/usr/local/lib/x86_64-linux-8.0.1``.
```
ezyang: It is recommended that a single, common directory to be used to store all installed libraries (as opposed to using \`\`$pkgid\`\` or similar variables to create a directory per installed library), as this helps reduce the size of rpath in executables built against dynamic libraries.  See https://github.com/haskell/cabal/pull/3982 for more details.

dcoutts:
> We cannot repurpose --libsubdir to take over the role of what --commonlibdir is doing because then we would run into trouble with Setup.hs files build against an older Cabal. If we were to repurpose --libsubdir, then all the object and interface files would end up in a single shared directory for all packages when using an older Cabal.

I don't understand this bit.

Surely the situation is that in older Cabal lib versions the `$libdir/$libsubdir` is used for both library files {a,so,dynlib,dll} and also for .hi files. Nothing we can do now can change that. But in new versions we can control them independently. To do that we have to introduce at minimum one new flag, e.g. `$hidir` for the .hi files. But why would we need to introduce /two/ new flags?

Isn't building Setup.hs against older Cabal lib versions a red herring, since that doesn't let us control them independently anyway. We know what version of the Cabal Setup.hs CLI we're talking to, so we can set the flags accordingly.

What am I missing?

dcoutts:
So as I see it, there's a few parts to this change:

 * Introducing a `--hidir` flag, independent from `--libdir` / `--libsubdir` (but defaulting to `$libdir/$libsubdir` so that there is no change in behaviour yet). This on its own is I think completely uncontroversial since it just provides more control without changing anything if it's not actively used.

One or both of:

 * Changing the default in Cabal for `$libdir` and `$hidir`
 * Changing the default in cabal-install/stack for `--libdir` and making use of `--hidir`.

The advantage of changing the default in Cabal is that it affects Linux distros that install libraries globally, for those Linux distros that don't override things already and just go with the defaults.

The disadvantage of changing the default in Cabal is that it adds extra version-dependent behaviour changes, that downstream tools need to be aware of and allow for or override (e.g. cabal-install already overrides all the flags in new-build at least, though it obviously does not yet override new flags like `--hidir`). In principle the file layout should be fully controlled by the callers, cabal-install/stack and not leave any of it to changing defaults in Cabal.

So an advantage of only changing the defaults in cabal-install/stack is that it may be simpler to do there, since they have to do it anyway.

dcoutts:
This is to help me understand...

So the original layout scheme was:
```
libdir=$prefix/lib
libsubdir=$abi/$libname
includedir=$libdir/$libsubdir/include
```
The .hi files and the .a .so .dll .dynlib files all get installed under `$libdir/$libsubdir`.

The `$abi` is like `x86_64-linux-ghc-7.10.1`, while the `$libname` is things like `tar-0.4.2.1-3r9cNWLCzNmJa5oTzjaqTP`.

Note that the libname is the package id plus a hash that is unique locally to a linked app/lib. The hash includes the names and versions of all deps, but no more. It is not a nix hash including sources and configuration.

The layout proposed in this patch (and please correct me if I'm wrong) is:
```
libdir=$prefix/lib
libsubdir=$abi/$libname
commonlibdir=$abi
hidir=$libname
includedir=$hidir/include
```
The .hi files get installed under `$hidir`. The .a .so .dll .dynlib files all get installed under `$libdir/$commonlibdir`.

So this introduces both `commonlibdir` and `hidir` and changes the default location for both .hi files and .a .so .dll .dynlib files. The .hi files move from `$prefix/lib/$abi/$libname` to `$prefix/lib/$libname`, ie dropping the `$abi` (it's not clear if this is intentional). The .a .so .dll .dynlib files move from `$prefix/lib/$abi/$libname` to `$prefix/lib/$abi`, i.e. dropping the `$libname` (which is the original motivation of the patch). The `$libsubdir` and `--libsubdir` remains but they are now not used for anything at all.

So a few things I'm concerned about:

 * I'm still not clear why we should introduce a new `$commonlibdir` and leave `$libsubdir` unused.
 * The hidir should be *explicitly* relative to something like `$libdir`, not implicitly. The tradition (going back to classic gnu `./configure`) is that the default for all these variables are relative to each other, and ultimately relative to `$prefix`. The only reason we have `$libsubdir` is as a hack because distros are used to using `--libdir=/usr/lib` and we needed a way to implicitly re-interpret that. But we don't need to do that for any other new vars, so we can use the normal explicitly-relative default scheme.
 * We still don't have any way to control the location of dynamic libs vs static libs. The scheme we want (I think) still places the static libs in the versioned dirs. This is relevant because we do not currently (afaik) use the `$libname` hash in the names of the static libs.

christiaanb:
So let's assume we only add `--hidir` (which is what I did in #3968), and chose this new layout scheme:

```
libdir=$prefix/lib
libsubdir=$abi
hidir=$libdir/$abi/$libname
```

This all works fine for `cabal-install` for `Setup.hs` files that can be build against Cabal-1.25.  But then, we have a _custom_ `Setup.hs` that can only be build against an older version of `Cabal` that does not know `--hidir`. What is going to happen then? Well `cabal-install` runs:

```
./Setup configure --libdir=$prefix/lib --libsubdir=$abi
```

And so what happens is that the `Setup` will put _all_ library files `.hi` and the `.a/.so/.dll/.dylib` files in `$prefix/lib/$abi`. Now, you might say: well, if we can detect that `Setup` does not know the `--hidir` flag, then we can just revert to the old install layout of `--libsubdir=$abi/$libname`.
But, we can't! Because a user might have _explicitly_ set `libsubdir=$abi`. And detecting whether the user explicitly set `libsubdir` to be the same as the `Cabal-1.25` default is just extremely fragile.

And that is the reason I had to add another flag, `--commonlibdir` (name up for bikeshedding) to determine where the binary files (`.a/.dll/.so/.dylib`) go in `Cabal-1.25` while still keeping around `--libsubdir` for pre-1.25 `Cabal`.

christiaanb:
@dcoutts 

The proposed layout is:

```
libdir=$prefix/lib
libsubdir=$abi/$libname
commonlibdir=$abi
hidir=$libdir/$abi/$libname
includedir=$hidir/include
```

christiaanb:
@dcoutts Currently in this PR, `--hidir` can be set to a directory that is completely independent of `--libdir`. However, by default, `--hidir` is set to `$libdir/$abi/$libname`.

But I'm not really sure of your comments: is this something you find desirable? or do you want `--hidir` to always be _implicitly_ be prefixed by `$libdir`?

cartazio:
Could we not use symlinks to collect all the libs Into a flattened
directory? At least outside of windows?

christiaanb:
@cartazio If we did, where would be "store" this location so that GHC and Cabal know where to find this symlinked directory? Would we update the package database with an extra field?

We need to know the location of the symlinked libraries so that GHC and Cabal can tell the linker which RPATHs to add.

dcoutts:
> But I'm not really sure of your comments: is this something you find desirable? or do you want --hidir to always be implicitly be prefixed by $libdir?

I was wrong actually. The `hidir` is exactly following the scheme I meant. It's `hidir=$libdir/$abi/$libname` as you said.

dcoutts:
Reopened #3979 since we're pursuing that approach. So I'll take the liberty of closing this PR.

PR #3983, opened 2016-10-14
Backport "Store interfaces files separately from library object files #3982" to 1.24

christiaanb:
These two flags indicate where the two distinct aspects of a
Haskell library end up.

--commonlibdir: The subdirectory of --libdir where the object
                library files (.so/.a/.dll/.dylib) get installed.
--hidir:        The directory where the interface files (.hi)
                get installed.

The reason we want to do this is because we want dynamic libraries
to all end up in a shared directory to reduce start-up times of
the run-time linker. However, we still want the .hi to end up
in one directory per package.

We cannot repurpose --libsubdir to take over the role of what
--commonlibdir is doing because then we would run into trouble
with Setup.hs files build against an older Cabal. If we were
to repurpose --libsubdir, then all the object and interface
files would end up in a single shared directory for all
packages when using an older Cabal.

This PR backports #3982 to the 1.24 branch in order to solve https://ghc.haskell.org/trac/ghc/ticket/12479

### DIFF
--- a/Cabal/Distribution/Simple/InstallDirs.hs
+++ b/Cabal/Distribution/Simple/InstallDirs.hs
@@ -80,7 +80,17 @@ data InstallDirs dir = InstallDirs {
         prefix       :: dir,
         bindir       :: dir,
         libdir       :: dir,
-        libsubdir    :: dir,
+        commonlibdir :: dir,
+        hidir        :: dir,
+        libsubdir    :: dir, -- This field is basically deprecated by the
+                             -- introduction of commonlibdir and hidir. However,
+                             -- We must keep it so that we can still work with
+                             -- Setup executables build against an older version
+                             -- of Cabal. For the same reason, we cannot
+                             -- simply use libsubdir as the commonlibdir,
+                             -- because then Setup's build against an older
+                             -- Cabal would put .hi files in the same directory
+                             -- as the library object files
         dynlibdir    :: dir,
         libexecdir   :: dir,
         includedir   :: dir,
@@ -110,6 +120,8 @@ combineInstallDirs combine a b = InstallDirs {
     prefix       = prefix a     `combine` prefix b,
     bindir       = bindir a     `combine` bindir b,
     libdir       = libdir a     `combine` libdir b,
+    commonlibdir = commonlibdir a `combine` commonlibdir b,
+    hidir        = hidir a      `combine` hidir b,
     libsubdir    = libsubdir a  `combine` libsubdir b,
     dynlibdir    = dynlibdir a  `combine` dynlibdir b,
     libexecdir   = libexecdir a `combine` libexecdir b,
@@ -125,9 +137,9 @@ combineInstallDirs combine a b = InstallDirs {
 
 appendSubdirs :: (a -> a -> a) -> InstallDirs a -> InstallDirs a
 appendSubdirs append dirs = dirs {
-    libdir     = libdir dirs `append` libsubdir dirs,
+    libdir     = libdir dirs `append` commonlibdir dirs,
     datadir    = datadir dirs `append` datasubdir dirs,
-    libsubdir  = error "internal error InstallDirs.libsubdir",
+    commonlibdir = error "internal error InstallDirs.commonlibdir",
     datasubdir = error "internal error InstallDirs.datasubdir"
   }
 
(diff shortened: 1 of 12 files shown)


ilovezfs:
This appears to fail with `new-build` https://gist.github.com/ilovezfs/ff5539ab4b6b16ab9a832ff7c9476284

phadej:
I think it's unfortunate that 1.24 `new-build` won't work on Sierra, but it shouldn't block this going in.

phadej:
ping @dcoutts about `new-build` and where to place dylibs in the store. Should we make a separate issue of it?

23Skidoo:
Do I understand correctly that `new-build` doesn't currently work on Sierra & 1.24 anyway?

ilovezfs:
No, new-build works just fine on Sierra, as long as you don't run into the bug that this PR is fixing for non-new-build.

ilovezfs:
Half a loaf is better than no loaf, but it would be good if a comprehensive solution shipped before 8.0.2 so that Sierra isn't rendered a second class GHC citizen.

The question is whether there is a possibility of a single fix that will address the problem for both non-new-build and new-build. If so, this PR may end up having to be backed out.

If there is no reasonable prospect of having a common fix, then there's a case for "pretending" like they're different problems, though they actually aren't, and shipping this PR.

23Skidoo:
For `new-build`, we currently recommend using cabal-install HEAD, since it's more stable. Will this fix break HEAD as well?

phadej:
@23Skidoo this doesn't.

Yet the store layout is decided only in `cabal-install`, so we can fix it with a point release.
I hope that flags here are enough to make it work.

ilovezfs:
>For new-build, we currently recommend using cabal-install HEAD, since it's more stable. Will this fix break HEAD as well?

The bug I referred to is not caused by this PR. So no it won't break HEAD or anything else.

christiaanb:
OK, just to be absolutely 100% clear: This PR does not break anything.

What is going on, is that this PR does _currently_ not solve the issues mentioned in https://ghc.haskell.org/trac/ghc/ticket/12479 when we use `new-build`. It _does_, however, solve the issues mentioned in https://ghc.haskell.org/trac/ghc/ticket/12479 when we use "old" `build` / `install`.

I believe that the changes in this PR to `Cabal` (the library) are sufficient to _eventually_ solve the issues as reported in https://ghc.haskell.org/trac/ghc/ticket/12479 when we use  `new-build`. However, since this PR is holding up the GHC 8.0.2 release, I think this issue with `new-build` should be solved at a later time in another PR.

The tests on this PR are all green, so I'm just waiting for the reviewers to give the go-ahead for a merge.

christiaanb:
Also, regardless of whether this PR solves https://ghc.haskell.org/trac/ghc/ticket/12479, it allows us to place library object files (.dll/.so/.a/.dylib) in a directory seperate from interface files `.hi`. And this might be desirable for reason other than solving https://ghc.haskell.org/trac/ghc/ticket/12479. For example, on Linux, startup times can be significantly reduced by limiting the number of RPATHs. One way to do this is by placing all the dynamic libraries in a single directory, which is what this PR does.

Cabal/Distribution/Simple/InstallDirs.hs
```diff
@@ -80,7 +80,17 @@ data InstallDirs dir = InstallDirs {
         prefix       :: dir,
         bindir       :: dir,
         libdir       :: dir,
-        libsubdir    :: dir,
+        commonlibdir :: dir,
+        hidir        :: dir,
+        libsubdir    :: dir, -- This field is basically deprecated by the
```
23Skidoo: s/basically deprecated/deprecated/

Cabal/Distribution/Simple/InstallDirs.hs
```diff
@@ -80,7 +80,17 @@ data InstallDirs dir = InstallDirs {
...
+                             -- introduction of commonlibdir and hidir. However,
+                             -- We must keep it so that we can still work with
```
23Skidoo: s/We/we/

Cabal/doc/installing-packages.markdown
```diff
@@ -568,6 +576,20 @@ In addition the simple build system supports the following installation path opt
     _suffix_ may contain the following path variables: `$pkgid`, `$pkg`,
     `$version`, `$compiler`, `$os`, `$arch`, `$abi`, `$abitag`
 
+`--commonlibdir=`_dir_
+:   For use with Setup.hs files build against a version of Cabal prior to 1.24.1.
+    For later versions of Cabal, this flag is basically deprecated, and you
```
23Skidoo: > For later versions of Cabal, this flag is basically deprecated

I think you were talking about `libsubdir` here.

23Skidoo:
Closing in favour of #3979, context: https://github.com/haskell/cabal/pull/3982#issuecomment-254606843.

PR #3986, opened 2016-10-16
Add Travis job to build cabal with solver debug flags enabled.

grayjay:
The job should detect issues like #3967.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,9 @@ matrix:
    - env: GHCVER=8.0.1 SCRIPT=script DEPLOY_DOCS=YES
      sudo: required
      os: linux
+   - env: GHCVER=8.0.1 SCRIPT=solver-debug-flags
+     sudo: required
+     os: linux
    - env: GHCVER=8.0.1 SCRIPT=bootstrap
      sudo: required
      os: linux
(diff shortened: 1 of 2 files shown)


ezyang:
Re sandbox, is this fundamentally a problem with how new-build works (i.e., would new-build work if we did it with HEAD?) Can we do something about the deps?

grayjay:
The dependencies' setup scripts don't compile with Cabal HEAD, so we would have to force an older Cabal version.  new-build at HEAD succeeds, because the dependencies lack custom-setup stanzas, and new-build adds a default setup dependency on Cabal < 1.25 (cabal-install 1.24 only constrains it to 2.0).  I think that the best solution would be to use qualified constraints (#3502) in cabal.project, once they are available.

ezyang:
Heh, oh the pains of being Cabal (maybe this behavior makes sense for other people, but it definitely doesn't make sense for us!)

grayjay:
Yeah, Cabal definitely seems like a special case.

phadej:
(FWIW, similar issue as compiling with `-fparsec-struct-diff`... cannot do that with 1.24 `new-build`)

grayjay:
> (FWIW, similar issue as compiling with -fparsec-struct-diff... cannot do that with 1.24 new-build)

Maybe we should apply #3939 to 1.24.

23Skidoo:
> Maybe we should apply #3939 to 1.24.

Done: 7ba7489c673e87a9a22ab892ede3faacafba8f9a.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #3989, opened 2016-10-17
Re-enable Travis testing other versions of GHC.

ezyang:
This increases our test coverage for using Cabal with GHC 7.0, 7.2
and HEAD.

### DIFF
--- a/Cabal/tests/PackageTests/Tests.hs
+++ b/Cabal/tests/PackageTests/Tests.hs
@@ -42,8 +42,11 @@ tests config = do
   -- Test that Cabal parses 'test' sections correctly
   tc "TestStanza"       PackageTests.TestStanza.Check.suite
 
-  -- Test that Cabal parses '^>=' operator correctly
-  tc "CaretOperator"    PackageTests.CaretOperator.Check.suite
+  -- Test that Cabal parses '^>=' operator correctly.
+  -- Don't run this for GHC 7.0/7.2, which doesn't have a recent
+  -- enough version of pretty. (But this is pretty dumb.)
+  tc "CaretOperator" . whenGhcVersion (>= mkVersion [7,3])$
+    PackageTests.CaretOperator.Check.suite
 
   -- Test that Cabal determinstically generates object archives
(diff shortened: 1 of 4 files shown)


BardurArantsson:
Not sure, but https://github.com/haskell/cabal/commit/d36a2d7c6c62d7155b1889d264caa0b87877f4c0 might be relevant to bring back?

ezyang:
It's a good thought, but building with GHC HEAD is broken due to broken hackage-security, and I don't feel like enabling it while it is broken.

Cabal/tests/PackageTests/Tests.hs
```diff
@@ -42,8 +42,11 @@ tests config = do
   -- Test that Cabal parses 'test' sections correctly
   tc "TestStanza"       PackageTests.TestStanza.Check.suite
 
-  -- Test that Cabal parses '^>=' operator correctly
-  tc "CaretOperator"    PackageTests.CaretOperator.Check.suite
+  -- Test that Cabal parses '^>=' operator correctly.
+  -- Don't run this for GHC 7.0/7.2, which doesn't have a recent
+  -- enough version of pretty. (But this is pretty dumb.)
+  tc "CaretOperator" . whenGhcVersion (>= mkVersion [7,3])$
```
phadej: do we have incorrect lowerbound?

ezyang: No, it's just that these tests run independently of the Hackage index, so the dep solver is only able to solve for boot packages. In 7.2/7.0 the boot version is lower than the caret lower bound.

23Skidoo:
This is marked green, but Travis tests [actually didn't run](https://travis-ci.org/haskell/cabal/builds/168185848).

ezyang:
It was green; I accidentally triggered a rebuild and then cancelled it.

23Skidoo:
OK, it's just that GitHub only shows the AppVeyor check as succeeded. I guess this can go in then?

ezyang:
Yes, though if you want to feel better about it, restart the build ;)

23Skidoo:
Nope, I trust you.

PR #3992, opened 2016-10-17
Add explicit osx image to GHC 7.8.4 job

### LINKED ISSUE #3991: Mac OS X GHC 7.8 CI failure: Cannot cabal install happy
All of our Mac OS X GHC 7.8 CI builds are failing because the cache was recently cleared out, and we are now tickling a latent bug: "cabal install happy" no longer works because happy is built with a Custom setup which is built with the system Cabal which is built against the old version of unix which has the "ar: permission denied" error.

I'm not sure what the easiest/most robust way of fixing this. There isn't an easily accessible non-buggy copy of the Cabal library on our OS X image which the Custom setup could link against. There isn't an obvious pre-built happy binary we can use. I suppose we could patch and install a custom copy of happy with the problem fixed?

Maybe we should just drop Mac OS X 7.8 support?

phadej:
Hopefully resolves #3991

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ matrix:
    # https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version
    - env: GHCVER=7.8.4 SCRIPT=script
      os: osx
+     osx_image: xcode7.1
    - env: GHCVER=7.10.3 SCRIPT=script
      os: osx
    - env: GHCVER=8.0.1 SCRIPT=script


phadej:
It works, just runs out of time :P

ezyang:
:-1: XCode 7.1 support is going away on October 31st: https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/

phadej:
We can use `osx_image: xcode6.4`, as the xcode version is irrelevant. Their documentation haven't pointed that fact.

ezyang:
OK let's do that.

PR #3997, opened 2016-10-18
Workaround GHC #11214 by filtering JavaScriptFFI

hvr:
Unfortunately, "native" GHC advertises support for `JavaScriptFFI` even though
it doesn't support it. See also https://ghc.haskell.org/ticket/11214 for respective
bug.

However, in order to properly declare that packages require `JavaScriptFFI` support
via `other-extensions` we need to fixup the list of extensions fed to the cabal solver.

This patch does something similiar to the workaround we added some time ago to
filter out TemplateHaskell for older GHCs which didn't properly advertise
`TemplateHaskell` availability (c.f. 9f68eb4455808ea9dd9adc5a59b5ebee52a02de8)

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -151,16 +151,21 @@ configure verbosity hcPath hcPkgPath conf0 = do
 
   ghcInfo <- Internal.getGhcInfo verbosity implInfo ghcProg
   let ghcInfoMap = Map.fromList ghcInfo
+      extensions = -- workaround https://ghc.haskell.org/ticket/11214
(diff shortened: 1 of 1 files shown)


hvr:
@23Skidoo this patch should be cherry-picked into the upcoming 1.24 point release as well

I've tested the patch locally with `ghcjs` to make sure it does the right thing in the presence of ` other-extensions: JavaScriptFFI`

/cc @luite

ezyang:
OK looks fine. A bit annoying that we can't test.

23Skidoo:
Merged, thanks! Also cherry-picked into 1.24 (0bae3176a68730487edeb59d935ec1ed06822ac1).

PR #4001, opened 2016-10-18
Don't register inplace if we're in per-component mode.

ezyang:
Generally, the the runner (e.g. new-build) will know
better about the specific package db that the package
should be registered inplace into.  So don't register
into our distdir; it's not useful.


ezyang:
AppVeyor failure is spurious

Issue #4009, opened 2016-10-19
`cabal new-build`  & {alex,happy}


hvr:
**UPDATE**: Partially (specifically for the case when `alex`/`happy` are to be installed into the nix store because they were requested via `build-tools`), this turns out to be a regression introduced by 8dc39db73dc40e3ca84116f6915ff8d1cc78f584

`new-build` currently isn't able to self-bootstrap `happy` and `alex` (which are two special cases in Haskell's ecosystem and for it's imho legitimate to add special logic to `new-build`), and this is causing problems for the `new-build` based Travis scripts which don't install `alex`/`happy` via binary packages anymore:

```
$ cabal get happy-1.19.5
Unpacking to happy-1.19.5/
$ cd happy-1.19.5/
$ cabal new-build
Resolving dependencies...
In order, the following will be built (use -v for more details):
 - happy-1.19.5 (exe:happy) (first run)
[1 of 1] Compiling Main             ( /tmp/happy-1.19.5/dist-newstyle/build/x86_64-linux/ghc-8.0.1.20161014/happy-1.19.5/setup/setup.hs, /tmp/happy-1.19.5/dist-newstyle/build/x86_64-linux/ghc-8.0.1.20161014/happy-1.19.5/setup/Main.o )

/tmp/happy-1.19.5/dist-newstyle/build/x86_64-linux/ghc-8.0.1.20161014/happy-1.19.5/setup/setup.hs:21:1-8: warning: [-Wtabs]
    Tab character found here, and in 39 further locations.
    Please use spaces instead.
Linking /tmp/happy-1.19.5/dist-newstyle/build/x86_64-linux/ghc-8.0.1.20161014/happy-1.19.5/setup/setup ...
Configuring happy-1.19.5...
Preprocessing executable 'happy' for happy-1.19.5...
setup: The program 'happy' is required but it could not be found
```

and similiarly for `alex`:

```
$ cabal get alex-3.2.1
Unpacking to alex-3.2.1/
$ cd alex-3.2.1/
$ cabal new-build
Resolving dependencies...
In order, the following will be built (use -v for more details):
 - alex-3.2.1 (exe:alex) (first run)
[...]

PR #4012, opened 2016-10-20
Fix install of packages bundling generated source in dist/

hvr:
Packages like `alex` include pre-generated lexer/parser source
in their source tarball as e.g.

  dist/build/alex/alex-tmp/Scan.hs
  dist/build/alex/alex-tmp/Parser.hs

to avoid having to have alex already installed before building
alex...

Unfortunately, 8dc39db73dc40e3ca84116f6915ff8d1cc78f584 broke
packages relying on this accidental feature by changing where the
distdir points to. This patch partly reverts that commit and
therefore addresses the regression aspect of #4009.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -749,7 +749,7 @@ withTarballLocalDirectory verbosity distDirLayout@DistDirLayout{..}
         -- compatibility we put the dist dir within it
         -- (i.e. tmp/src2345/foo-1.0/dist/).
         --
-        -- Unfortunately. a few custom Setup.hs scripts do not respect
+        -- Unfortunately, a few custom Setup.hs scripts do not respect
         -- the --builddir flag and always look for it at ./dist/ so
         -- this way we avoid breaking those packages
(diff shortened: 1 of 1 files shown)


ezyang:
Ahahahaha, so that's what it is. LGTM.

cabal-install/Distribution/Client/ProjectBuilding.hs
```diff
@@ -744,16 +744,21 @@ withTarballLocalDirectory verbosity distDirLayout@DistDirLayout{..}
...
       case buildstyle of
-        -- In this case we make a couple temp dirs, unpack the tarball to one
-        -- and build and install it from the other. We avoid nesting the
-        -- builddir under the tarball src dir to keep path name lengths down.
+        -- In this case we make a temp dir (e.g. tmp/src2345/), unpack
+        -- the tarball to it (e.g. tmp/src2345/foo-1.0/), and for
+        -- compatibility we put the dist dir within it
+        -- (i.e. tmp/src2345/foo-1.0/dist/).
+        --
+        -- Unfortunately. a few custom Setup.hs scripts do not respect
```
23Skidoo: s/./,/

PR #4019, opened 2016-10-21
Hard-code unit id of Cabal to test, and use -hide-all-packages.

### LINKED ISSUE #4018: package-tests on old GHC spazzes out trying to build Setup.hs
The problem is that we don't pass `-hide-all-packages`, so everything from the store database just gets slammed into GHC, a pile of shadowing takes place and when the dust settles, more often than not, 1.25.0.0 will have gotten hidden.

Passing `-hide-all-packages` should solve the problem.

ezyang:
Fixes #4018.


ezyang:
OK, uh, I guess this isn't going to work.

PR #4023, opened 2016-10-22
Do an early check for unbuildable deps, and give good error.

ezyang:
When a library/executable is unbuildable, the solver may still
return a reference to it.  Handle this gracefully in ProjectPlanning
with a nice error message, rather than an assert failure.

Fixes #3978.

Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>

Issue #3784, opened 2016-09-06
We should sanity check GenericPackageDescriptions created by the solver DSL before we ship it to the solver


ezyang:
While adding executable support to the solver DSL in #3783, I forgot to add executables to the condExecutables subtree of the generic package description. This lead to considerable head-scratching when I was unable to reproduce the error.

There were two ways this could have been easier to debug:
1. If there was some way to print out the GPD output by the DSL, I could have just printed them out and understood the problem immediately (no executable sections.)
2. If there was some sort of sanity check on the GPD, it would have realized that the ComponentDeps recorded dependencies for an executable which wasn't anywhere in the tree.

This doesn't seem like a pants on fire problem but it would be good to fix eventually. Maybe if it bites me again! (CC @grayjay)

grayjay:
Was the GPD invalid, or was it just inconsistent with the `DSL.ExampleAvailable`?

ezyang:
I am not sure (because I didn't know how to print it); but I think it was invalid because the condition tree should have contained some dependencies were not in the actual BuildInfo.

grayjay:
6dc05349888fc234e4fb3771271cfc050d04883a runs package checks on the GenericPackageDescription produced by the solver DSL.

PR #4028, opened 2016-10-22
Solver DSL improvements

grayjay:
Three commits:

#### Update solver unit tests for Buildable field.

This commit modifies the tests so that they disable executables instead of
libraries.  Disabling executables with `Buildable: False` is a more realistic
test case.

See #3988.

#### Refactor 'UnitTests.Distribution.Solver.Modular.DSL.exAvSrcPkg'.

Previously, the solver DSL ignored some types of dependencies when they appeared
in executables or under flags. This commit uses one function, mkBuildInfoTree,
to create all BuildInfos, so that any dependency can be used in any location.

#### Run package checks on the GenericPackageDescription created by the solver DSL.

The checks might detect some errors in the DSL. The commit also adds non-default
values to some fields to make the checks pass, such as adding a default-language
to each component.

Related to #3784

/cc @ezyang 
I don't know if you need the first commit for #3988, since #4023 was merged, but I still think it's an improvement.

### DIFF
--- a/cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 -- | DSL for testing the modular solver
 module UnitTests.Distribution.Solver.Modular.DSL (
     ExampleDependency(..)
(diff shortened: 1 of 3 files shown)


cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
```diff
@@ -246,59 +251,100 @@ type ExampleDb = [Either ExampleInstalled ExampleAvailable]
...
+                  let mkTree = mkCondTree defaultTest mkTest . mkBuildInfoTree . Buildable
+                      mkTest bi = mempty { C.testBuildInfo = bi }
+                  in map (\(t, deps) -> (t, mkTree deps)) testSuites
+              , C.condBenchmarks  = []
+              }
+            }
+        pkgCheckErrors =
+          let ignore = ["Unknown extensions:", "Unknown languages:"]
+          in [ err | err <- C.checkPackage (packageDescription package) Nothing
+             , not $ any (`isPrefixOf` C.explanation err) ignore ]
```
grayjay: I filtered out these two warnings because some unit tests test that the solver allows unknown extensions/languages when the compiler supports them.  I don't know whether the unit tests or package checks are correct.

ezyang:
I haven't looked too closely, but any refactoring of the solver DSL is fine by me. But the quickchecks are failing!

cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
ezyang: Cabal hardcodes a list of known extensions and warns if it sees something it doesn't understand.

I think the reason we hardcode is because some extensions require special handling: e.g., TemplateHaskell. But the vast majority of extensions don't get any special handling. This raises an interesting question: for an extension we've never seen before, should we assume that it does or does not need special handling? If the former, the warning is appropriate; if the latter, we really ought to only define extensions which we do handle specially. But maybe this is too much for a simple PR like this.

grayjay: I don't know.  If most extensions don't need special handling, and we can get their names directly from the compiler, then it seems like it would be simpler to not define all of them.

I'll just add a comment to the line above for now.

grayjay:
@ezyang Thanks.  The main reason that I cc'ed you is that the change relates to #3784 and #3988.

The tests pass now.  I just needed to disallow unconstrained dependencies on `base` in the quickcheck tests so that the package checks pass.

ezyang:
Go for it :)

grayjay:
Thanks!  I'll merge it now.

PR #4029, opened 2016-10-23
Fix CPP usage

erikd:
The code had a mixtire of `#ifdef mingw32_HOST_OS` and `#if`. The later
works but is not really correct. GHC HEAD has just got a new warning flag
`-Wcpp-undef` which will warn on `#if` used with an undefined identifier.
Since we want to turn that on in the GHC build system we need to fix cabal
first.


erikd:
The Appveyor build failure had nothing to do with my patch.

erikd:
Thanks @dcoutts !

PR #4030, opened 2016-10-23
Revert "D.PD.Check: extra-doc-files only allowed when Cabal >= 1.18."

ezyang:
This causes filepath to not validate, which means GHC build doesn't
work. Annoying. Let's not for 8.0 series.

This reverts commit 0b458361e7a1af485e74354bad74b8ae02e0f5f4.

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -904,13 +904,6 @@ checkCabalVersion pkg =
         ++ "different modules then list the other ones in the "
         ++ "'other-languages' field."
 
-  , checkVersion [1,18]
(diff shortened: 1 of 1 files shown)


23Skidoo:
Can't we fix filepath instead?

ezyang:
Well, we did. And now filepath has to cut a new release so that GHC 8.0.2 can ship it.

I guess the point more is that, for bugfix releases we should really avoid gratuitously breaking things for users. This patch breaks things for no discernible gain. It is certainly correct; which is why it's a good patch, but maybe it doesn't make sense for 1.24.

23Skidoo:
> I guess the point more is that, for bugfix releases we should really avoid gratuitously breaking things for users.

Sure, let's merge if it's 1.24 only.

PR #4034, opened 2016-10-25
Force gen-extra-source-files to use pre-installed library.

### LINKED ISSUE #4007: gen-extra-source-files.hs is slow
```
ezyang@sabre:~/Dev/cabal-backpack/Cabal$ time misc/gen-extra-source-files.hs Cabal.cabal

real	0m14.212s
user	0m13.524s
sys	0m0.580s
ezyang@sabre:~/Dev/cabal-backpack/Cabal$ runghc --version
runghc 7.10.3
ezyang@sabre:~/Dev/cabal-backpack/Cabal$ time misc/gen-extra-source-files.hs 
Error too few arguments!
Usage: gen-extra-source-files.hs FILE
  where FILE is Cabal.cabal or cabal-install.cabal

real	0m13.731s
user	0m13.032s
sys	0m0.608s
```

Maybe this is just runghc is slow? Anyway it's kind of annoying.

CC @phadej

ezyang:
Without a package qualified import, runhaskell will attempt to
load all of Cabal from source, because the files happen to
be named in the same way as the import requests.  That leads
to very slow runtimes.

Fixes #4007.

CC @phadej

### DIFF
--- a/Cabal/misc/gen-extra-source-files.hs
+++ b/Cabal/misc/gen-extra-source-files.hs
@@ -1,14 +1,19 @@
 #!/usr/bin/env runhaskell
+{-# LANGUAGE PackageImports #-}
+
+-- NB: Force an installed Cabal package to be used, NOT
+-- some local files which have these names (as would be
+-- the case if we were in the Cabal source directory.)
+import "Cabal" Distribution.PackageDescription
+import "Cabal" Distribution.PackageDescription.Parse (ParseResult (..), parsePackageDescription)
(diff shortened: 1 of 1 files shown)


23Skidoo:
Or we can just call it from the top of tree.

ezyang:
That sounds like a workaround, whereas a package qualified import fixes the problem properly.

23Skidoo:
Merged, thanks!

PR #4048, opened 2016-10-28
Enable other GHCs with Parsec (for Backpack coverage.)

ezyang:
Signed-off-by: Edward Z. Yang ezyang@cs.stanford.edu

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
    - env: GHCVER=8.0.1 SCRIPT=solver-debug-flags
      sudo: required
      os: linux
-   - env: GHCVER=8.0.1 SCRIPT=script PARSEC=YES
+   - env: GHCVER=8.0.1 SCRIPT=script PARSEC=YES TEST_OTHER_VERSIONS=YES
      os: linux
      sudo: required
    - env: GHCVER=8.0.1 SCRIPT=bootstrap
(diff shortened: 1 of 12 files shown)


ezyang:
Merged.

phadej:
Whst happened to Travis file?

ezyang:
Uhh, not much? https://github.com/haskell/cabal/commit/20bed0639ac7dd50149a98f5dd1a0cda09f9e5b4

.travis.yml
```diff
@@ -19,65 +19,9 @@ branches:
 # lines listings versions you don't need/want testing for.
 matrix:
   include:
-   - env: GHCVER=8.0.1 SCRIPT=meta BUILDER=none
```
phadej: @ezyang yo merged this commit.

phadej:
Ah, you pushed directly to master. Sorry. It's just very confusing. I look at PR see, it's merged, but its changes are *wrong*. 

I think we should *protect* master branch and disallow direct pushes. There is still way to admin's to go around that limitation, but at least you'll stop and think.

Especially this all was confusing as mention-bot pinged me, and no-one said I don't need to do anything.

ezyang:
Sorry, I'll make sure the PR is up-to-date before merging.

The irritating thing about our Travis setup is it's too dumb to know that I made a trivial change that doesn't actually need to be rebuilt; just a waste of queue (and impedes other people who are more interested in their builds.) So a direct push is appealing because it avoids the useless Travis build. Would be far better if we could do something else, like how homu works.

PR #4054, opened 2016-10-29
Move PackageTests to new cabal-testsuite package.

ezyang:
The key idea is that once we put PackageTests in its own package, we can
ensure that its Setup.hs is compiled with the same version of Cabal
library as package-tests is compiled with.  This means that LBI reading
will _always succeed_ which is quite nice.

Signed-off-by: Edward Z. Yang ezyang@cs.stanford.edu


23Skidoo:
Nice!

23Skidoo:
Let's merge this ASAP because of high conflict potential.

PR #4055, opened 2016-10-29
[WIP] Implement External Custom Setup

Ericson2314:
Implementation of #4047

### DIFF
--- a/Cabal/Distribution/Package.hs
+++ b/Cabal/Distribution/Package.hs
@@ -18,6 +18,7 @@
 
 module Distribution.Package (
         -- * Package ids
+        UnqualComponentName, unUnqualComponentName, mkUnqualComponentName,
         PackageName, unPackageName, mkPackageName,
         PackageIdentifier(..),
         PackageId,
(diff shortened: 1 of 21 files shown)


Cabal/Distribution/Package.hs
```diff
@@ -303,7 +313,20 @@ instance Text Dependency where
              Parse.skipSpaces
              return (Dependency name ver)
 
+instance Text ExeDependency where
+  disp (ExeDependency name exe ver) =
+    (disp name <<>> Disp.text ":" <<>> Disp.text exe) <+> disp ver
+
+  parse = do name <- parse
+             _ <- Parse.char ':'
+             exe <- Parse.munch1 isAlphaNum
```
ezyang: I think the executable name format is more flexible than AlphaNum. Safest is to parse as if it were a PackageName.

Cabal/Distribution/Parsec/Class.hs
```diff
@@ -115,6 +115,14 @@ instance Parsec Dependency where
         ver  <- parsec <|> pure anyVersion
         return (Dependency name ver)
 
+instance Parsec ExeDependency where
+    parsec = do
+        name <- lexemeParsec
+        _    <- P.char ':'
+        exe  <- P.munch1 isAlphaNum
```
ezyang: Don't forget to update here too when you're done.

Cabal/Distribution/Types/TestSuite.hs
```diff
@@ -43,11 +44,12 @@ instance Semigroup TestSuite where
         testInterface = combine  testInterface,
         testBuildInfo = combine  testBuildInfo
     }
-        where combine   field = field a `mappend` field b
-              combine' f = case (f a, f b) of
-                        ("", x) -> x
-                        (x, "") -> x
-                        (x, y) -> error "Ambiguous values for test field: '"
+        where combine  field = field a `mappend` field b
+              combine' field = case ( unUnqualComponentName $ field a
```
ezyang: I'd much prefer it if you could add a function like "isNullUnqualComponentName" to factor this type of test. It would be more efficient too since unUnqualComponentName requires serializing to String.

ezyang:
@Ericson2314 Should we close this PR for now?

Ericson2314:
Sure. It will look completely different rebased, and the point won't be external custom setups anyways.

PR #4062, opened 2016-10-30
Make 'FlagName' opaque & `ShortText`-backed

hvr:


hvr:
This PR is done as far as I'm concerned; however, as future work, there's possible extensions/refactorings I can think of:

 - make lowercase normalisation part of `mkFlagName`, as it appears that cabal flagnames are supposed to be case-insensitive from what I saw
 - validate input to `mkFlagName` (e.g. leading `-` or `+`, whitespaces etc) to enforce invariant

ezyang:
Thanks for the patch!

Issue #4049, opened 2016-10-28
Add foreign-library to LD_LIBRARY_PATH during cabal exec


abooij:
Thanks for the work on the new `foreign-library` section. I have started using it for one of my projects, and it seems to work!

My use case is to reimplement a C library's ABI, and run existing binaries written in C against the Haskell version. At development time, this means building my haskell code as a foreign library, adding the relevant directory to `LD_LIBRARY_PATH`, and then running the existing binaries.

It would be convenient if cabal could add the haskell foreign library to `LD_LIBRARY_PATH`. I suppose the `cabal exec` command could be the right place to do this. So instead of

```
$ cabal build
$ LD_LIBRARY_PATH=dist/build/myproject/ /usr/bin/existing-binary
```

one would instead simply:

```
$ cabal exec /usr/bin/existing-binary
```

ezyang:
The motivation seems reasonable. The big question, however, is WHICH paths should get added to LD_LIBRARY_PATH? After all, a foreign library could itself depend on (1) dynamic Haskell libraries that were installed user-wide, or (2) a different foreign library. All of these paths need to be added to LD_LIBRARY_PATH for things to "work" (although I guess we do bake in RPATHs to make sure that things can be found without LD_LIBRARY_PATH).

`cabal run`, as opposed to `cabal exec`, DOES twiddle the `LD_LIBRARY_PATH`, but because it's associated with an actual Haskell package, we know exactly what things need to be added. In your suggested `cabal exec`, this is not known; is the suggestion then that every foreign library of the package be added? (What if you're in a sandbox?) That actually makes a good bit of sense, as `cabal exec` is supposed to add the local sandbox bin directory to your PATH, so similarly, it should add the local sandbox lib directory (doesn't exist yet) to your LD_LIBRARY_PATH. Though, this doesn't help much in the non-sandbox case where PATH doesn't get twiddled (maybe it should: add to the PATH the directory of each executable build product; similarly, add to LD_LIBRARY_PATH the directory of each library build product).

@abooij does this make sense? Would you be interested in implementing this?

abooij:
@ezyang That does make sense. And you are able to make more precise what I was somewhat worried about when I proposed this -- thanks for that.

We discussed this a bit on IRC and I can try to work on this.

ezyang:
OK, I added you to the Cabal project and assigned you this ticket.

ezyang:
@abooij Would you like an architectural overview on this ticket?

PR #4064, opened 2016-10-30
Add sandbox lib dir to environment during cabal exec

abooij:
See #4049

Adding the sandbox lib dir means that foreign libraries are found by random executables after installing those foreign libraries, even if they are installed to the sandbox. Without this, LD_LIBRARY_PATH needs to be set manually if using sandboxes.

### DIFF
--- a/cabal-install/Distribution/Client/Exec.hs
+++ b/cabal-install/Distribution/Client/Exec.hs
@@ -95,8 +95,9 @@ sandboxEnvironment verbosity sandboxDir comp platform programDb =
         unless exists $ warn verbosity $ "Package database is not a directory: "
                                            ++ sandboxPackagePath
         let ldPath = case os of
-                       OSX -> "DYLD_LIBRARY_PATH"
-                       _   -> "LD_LIBRARY_PATH"
+                       OSX     -> "DYLD_LIBRARY_PATH"
+                       Windows -> "PATH"
+                       _       -> "LD_LIBRARY_PATH"
         currentLibraryPath <- lookupEnv ldPath
         let newLibraryPath = case currentLibraryPath of
                                Nothing -> sandboxDir </> "lib"
--- a/cabal-install/Distribution/Client/Exec.hs
+++ b/cabal-install/Distribution/Client/Exec.hs
@@ -29,10 +29,11 @@ import Distribution.Simple.Program.Run (programInvocation, runProgramInvocation)
 import Distribution.Simple.Program.Types ( simpleProgram, ConfiguredProgram(..) )
 import Distribution.Simple.Utils       (die, warn)
 
-import Distribution.System    (Platform)
+import Distribution.System    (Platform(..), OS(..))
 import Distribution.Verbosity (Verbosity)
 
 import System.Directory ( doesDirectoryExist )
+import System.Environment (lookupEnv)
 import System.FilePath (searchPathSeparator, (</>))
 
 
@@ -81,6 +82,7 @@ sandboxEnvironment verbosity sandboxDir comp platform programDb =
       GHCJS -> env GHCJS.getGlobalPackageDB ghcjsProgram "GHCJS_PACKAGE_PATH"
       _     -> die "exec only works with GHC and GHCJS"
   where
+    (Platform _ os) = platform
     env getGlobalPackageDB hcProgram packagePathEnvVar = do
         let Just program = lookupProgram hcProgram programDb
         gDb <- getGlobalPackageDB verbosity program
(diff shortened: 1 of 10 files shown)


23Skidoo:
LGTM.

abooij:
Perhaps this should take `ldPath` to be `PATH` for `OS=Windows`?

ezyang:
@abooij Yes, that seems quite reasonable. But it would be good if we checked if that worked. If you added a test for this functionality to the cabal-install test suite, Appveyor would check for you. Would you like to do that? Otherwise feel free to go ahead and merge.

abooij:
I tried to write a test but it turns out I don't have easy access to "cabal exec" so I can't figure out an easy way to test this.
Apologies for all the force-pushing, every time I thought I had fixed it... I'll be more organised.

ezyang:
@abooij Wrong place for the test! Don't put it in Cabal/tests/; put it in cabal-install/tests/IntegrationTests; cabal exec is a cabal-install only feature.

Don't worry about force pushing; that's what PRs are for :)

abooij:
@ezyang: I see. How would you suggest to organise this? Start a new "foreign" test category? Or extend one of the existing ones?

ezyang:
Put it in exec!

abooij:
Okay, so now we need to compile a C program in the `cabal-install` test suite. How am I supposed to obtain a valid version of gcc?

ezyang:
Hmm, that is a good question.

Go look at `cabal-install/tests/IntegrationTests.hs`, and fine these lines:

```
  (cabal, _) <- requireProgram normal cabalProgram (setProgramSearchPath programSearchPath defaultProgramDb)
  (ghcPkg, _) <- requireProgram normal ghcPkgProgram defaultProgramDb
  baseDirectory <- canonicalizePath $ "tests" </> "IntegrationTests"
  -- Set up environment variables for test scripts
  setEnv "GHC_PKG" $ programPath ghcPkg
  setEnv "CABAL" $ programPath cabal
```

Try requiring `gccProgram` and then setup an environment variable for it. That should be sufficient.

abooij:
Okay, I've spent a some time trying to get this to work but it is too hairy for the benefit we're supposed to get. Can we leave it as is?

ezyang:
This is a fairly obscure corner of Cabal, and so I fear that if we don't have a test, if someone breaks it later we won't notice (and that will be a PITA for you, downstream user of the feature!) Let me have a go at writing the test case.

abooij:
The remaining failing test is irrelevant, right?
```
All 8 tests passed (51.59s)
/Users/travis/build/haskell/cabal/dist-newstyle/build/cabal-install-1.25.0.0/build/integration-tests2/integration-tests2 took 52 seconds.
whole job took 2406 seconds so far.
Job taking over 40 minutes. Terminating


The command "./travis-${SCRIPT}.sh" exited with 1.
```

ezyang:
Yep; I restarted that build.

ezyang:
@abooij I pushed a rebased patch with the test suite. When it's green we'll merge.

abooij:
Well, that's a pretty dramatic failure on OSX. I suppose it was important to test this after all.
```
    T4049.sh (ignoring stdout+stderr):                                                FAIL (8.28s)
      Unexpected exit status.
      
      Exit status: ExitFailure 139
      Working directory: /var/folders/vy/rcv48w3j4w79llzf_x6qnvw40000gn/T/cabal-install-test-13245
      
      <stdout> was:
      
      
      <stderr> was:
      
      UseLib.c:5:3:
           warning: implicit declaration of function 'myForeignLibInit' is invalid in C99 [-Wimplicit-function-declaration]
            myForeignLibInit();
            ^
      
      UseLib.c:6:3:
           warning: implicit declaration of function 'sayHi' is invalid in C99 [-Wimplicit-function-declaration]
            sayHi();
            ^
      
      UseLib.c:7:3:
           warning: implicit declaration of function 'myForeignLibExit' is invalid in C99 [-Wimplicit-function-declaration]
            myForeignLibExit();
            ^
      3 warnings generated.
      ./common.sh: line 5: 13697 Segmentation fault: 11  "$CABAL" $CABAL_ARGS "$@"
```

There is also one remaining failure on linux which is due to a missing check in the test for GHC >=7.8

But Windows is green!

ezyang:
I guess the first thing we can try is add in explicit declarations for the functions, squashing the implicit declaration warnings.

abooij:
I don't think that will help. The code in `cabal-testsuite` works fine, after all. And if the cause of the error is a bunch of compiler warnings, why is there a mention of a segfault?

abooij:
I added explicit declarations, avoiding the compiler warnings. It didn't help.

The patch itself seems to do its job, however: running `./Uselib` _without_ `DYLD_LIBRARY_PATH` set results in a different error:
```
dyld: Library not loaded: @rpath/libmyforeignlib.dylib
  Referenced from: /private/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/cabal-install-test-6591/T4049/UseLib
  Reason: image not found
```
What else could cause the segfault? The RTS not being found? I don't know how to proceed. What's especially confusing is that the almost identical code in cabal-testsuite works.

abooij:
And that last parenthetical remark, it seems, was the problem. Even a simple hello world program in C produces a segfault if compiled with GHC.

abooij:
@ezyang The test now passes.

ezyang:
Awesome! Let's merge!

23Skidoo:
Merged, thanks! Great job @abooij and @ezyang!

PR #4065, opened 2016-10-30
Add test for foreign libs to depend on Haskell libs

abooij:
This is supposed to test if a foreign library can depend on Haskell code in the same project but in a different binary object.

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -1806,8 +1806,8 @@ checkForeignLibSupported comp platform flib = go (compilerFlavor comp)
   where
     go :: CompilerFlavor -> Maybe String
     go GHC
-      | compilerVersion comp < mkVersion [7,4] = unsupported [
-        "Building foreign libraires is only supported with GHC >= 7.4"
+      | compilerVersion comp < mkVersion [7,8] = unsupported [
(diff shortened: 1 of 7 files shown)


ezyang:
Sorry, this PR got swizzled by the package-tests move. Just transplant the changes into the new cabal-testsuite directory, or tell me if you want me to do it.

ezyang:
Lovely, nice new test which uncovers some bugs in the foreign library implementation.

1. Run `update-cabal-files.sh`, that will fix the first CI failure
2. For some reason foreign libraries with an external Haskell library dep don't work for GHC 7.6 and earlier. I'm happy to bump the support window of foreign libraries to 7.8+, but then `checkForeignLibSupported` in  `Cabal/Distribution/Simple/Configure.hs` has to be updated, and the test suites configured to not run in that case.

abooij:
I made the modifications that you proposed: rebased, ran `update-cabal-files.sh`, and bumped to 7.8 in `Configure.hs` and `Check.hs`.

ezyang:
Pinging @edsko: If you need foreign libraries to work on pre GHC 7.8, you better check this bug out!

edsko:
No, pre 7.8 support for limited anyway (https://github.com/haskell/cabal/pull/2540#issuecomment-96016726). Fine with me to just regard as unsupported.

PR #4066, opened 2016-10-31
parsec: Fix inconsistency and allow parse quoted-versions for build-tools

Ericson2314:
Is this an intentional inconsistency with the old parser? Compare with the endpoints of https://github.com/haskell/cabal/blob/master/Cabal/Distribution/ParseUtils.hs#L630-L658

CC @phadej


phadej:
FWIW, I'd rather disallow quoting the version range in old parser. There are no single use of it on Hackage, and it seems weird to allow:
```
build-depends: foo ">= 1.2.3"
```

Ericson2314:
@phadej great!

Ericson2314:
Repuposed this to disallow across the board instead.

PR #4067, opened 2016-10-31
Make more dependency types

Ericson2314:
This is a follow-up to https://github.com/haskell/cabal/pull/4057, so until that is dealt with this diff will also be humongous and unreviewable.
 
This PR makes separate types for different types of dependencies:
 - `Dependency` is now just for regular (usually library) dependencies on haskell packages
 - `LegacyExeDependecy` for legacy `build-tools` dependencies 
 - `PkgconfigDependecy` for pkg-config dependencies

Additionally, there is a newtype `PkgconfigName` for pkg-config library names. For Cabal's uses alone this wouldn't be worth it, but cabal-install already distinguishes these libraries widely with `PackageName`. Continuing to use `PackageName` seemed inconsistent with the newly-rigorous use of `Dependecy`, while going back to `String` seemed like a step backwards.

Plain `String` is used for legacy built tool names, as those can mean quite different things, and very little code was using `PackageName` and not `String` for them.

Note I had to make some orphan instances (module-wise, not package-wise) in order to avoid a `Distribution.ParserUtils` <-> `Distribution.Package` cyclic dep. This can go away once the old parser is removed.

### DIFF
--- a/Cabal/Distribution/Package.hs
+++ b/Cabal/Distribution/Package.hs
@@ -23,6 +23,7 @@ module Distribution.Package (
         packageNameToUnqualComponentName, unqualComponentNameToPackageName,
         PackageIdentifier(..),
         PackageId,
+        PkgconfigName, unPkgconfigName, mkPkgconfigName,
 
         -- * Package keys/installed package IDs (used for linker symbols)
         ComponentId, unComponentId, mkComponentId,
@@ -43,6 +44,8 @@ module Distribution.Package (
 
         -- * Package source dependencies
         Dependency(..),
+        LegacyExeDependency(..),
+        PkgconfigDependency(..),
         thisPackageVersion,
         notThisPackageVersion,
         simplifyDependency,
@@ -59,7 +62,7 @@ import Distribution.Compat.Prelude
 import Distribution.Utils.ShortText
 
 import Distribution.Version
-         ( Version, VersionRange, anyVersion, thisVersion
+         ( Version, VersionRange, thisVersion
          , notThisVersion, simplifyVersionRange
          , nullVersion )
 
@@ -69,7 +72,7 @@ import Distribution.Compat.ReadP
 import Distribution.Text
 import Distribution.ModuleName
 
-import Text.PrettyPrint ((<+>), text)
+import Text.PrettyPrint (text)
 
 -- | An unqualified component name, for any kind of component.
 --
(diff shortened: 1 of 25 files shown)


ezyang:
CI failures are real. If you try compiling with -Werror that should help you catch the warnings (CI will fail unless you are warnings-free.)

ezyang:
CC @phadej especially on https://github.com/haskell/cabal/pull/4067/commits/b729cffac4a86dca670cc3a807b93a25c205a6ba ; @Ericson2314 spotted an inconsistency with the Parsec parser and the old parser: whether or not quoted strings in build-tools are allowed.

Cabal/Distribution/ParseUtils.hs
```diff
@@ -19,6 +19,7 @@
 -- This module is meant to be local-only to Distribution...
 
 {-# OPTIONS_HADDOCK hide #-}
+{-# LANGUAGE Rank2Types #-}
```
ezyang: OK, excellent use of rank 2 types :)

Cabal/Distribution/Package.hs
```diff
@@ -138,6 +141,39 @@ unPackageName = unUnqualComponentName . packageNameToUnqualComponentName
 mkPackageName :: String -> PackageName
 mkPackageName = PackageName . mkUnqualComponentName
 
+-- | A pkgconfig library name
```
ezyang: But what is a pkgconfig library name?! Answer: "a valid argument to the pkg-config utility".

Cabal/Distribution/Package.hs
```diff
@@ -317,19 +353,24 @@ mkLegacyUnitId = newSimpleUnitId . mkComponentId . display
 data Dependency = Dependency PackageName VersionRange
                   deriving (Generic, Read, Show, Eq, Typeable, Data)
 
-instance Binary Dependency
+-- | Describes a legacy `build-tools`-style dependency on an executable
```
ezyang: It is "legacy" because we do not know if the build-tool referred to, refers to a pkg-config executable, or an internal executable (thus it is stringly typed.)

Cabal/Distribution/Package.hs
```diff
@@ -317,19 +353,24 @@ mkLegacyUnitId = newSimpleUnitId . mkComponentId . display
...
+--
+data LegacyExeDependency = LegacyExeDependency
+                           String  -- ^ name of one of the hardcoded understood tools
```
ezyang: Not always a hardcoded tool, right?

Cabal/Distribution/Package/TextClass.hs
```diff
@@ -0,0 +1,56 @@
+-- | *Dependency Text instances moved from Distribution.Package
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
ezyang: OK for now. Let's refactor it away later.

Ericson2314:
Ah to be I opened up a separate PR for the parser commit so @phadej can ignore this.

Ericson2314:
> Distribution/Package.hs:407:28: parse error on input ‘VersionRange’

There's a haddock error but not sure what's wrong. Would `-- ^` be required for every field?

ezyang:
I guess so. Add it and see?

Ericson2314:
Ah, need a record for per-field documetation

Ericson2314:
No point documenting per-variant if there's only one, so I removed the `-- ^`.

Ericson2314:
@ezyang also despite the not-invalidated diffs aboved I think I fixed everything you mentioned if you want to toggle your review :).

PR #4068, opened 2016-10-31
[ci skip] Commentary for configureRequiredProgram.

ezyang:


23Skidoo:
Nice, let's merge.

PR #4074, opened 2016-11-01
Foreign library versioning

abooij:
See #4052 

- [x] FOREIGNLIBFIELDS
- [x] add senity checks for this: for now only supported on linux
- [x] ADDSONAME
- [x] INSTALLSONAME
- [x] document new field and code
- [x] write a test
- [x] implement new ideas in #4052

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -875,7 +875,7 @@ exeTargetName exe = unUnqualComponentName (exeName exe) `withExt` exeExtension
 -- than the target OS (but this is wrong elsewhere in Cabal as well).
 flibTargetName :: LocalBuildInfo -> ForeignLib -> String
 flibTargetName lbi flib =
-    case (platformOS (hostPlatform lbi), foreignLibType flib) of
+    case (os, foreignLibType flib) of
       (Windows, ForeignLibNativeShared) -> nm <.> "dll"
       (Windows, ForeignLibNativeStatic) -> nm <.> "lib"
       (Linux,   ForeignLibNativeShared) -> "lib" ++ nm <.> versionedExt
@@ -886,20 +886,24 @@ flibTargetName lbi flib =
     nm :: String
     nm = unUnqualComponentName $ foreignLibName flib
 
-    platformOS :: Platform -> OS
-    platformOS (Platform _arch os) = os
+    os :: OS
+    os = let (Platform _ os') = hostPlatform lbi
+         in os'
 
-    -- If a foreign lib foo has lib-version-info 5:1:2, it should be built
-    -- as libfoo.so.3.2.1
+    -- If a foreign lib foo has lib-version-info 5:1:2 or
+    -- lib-version-linux 3.2.1, it should be built as libfoo.so.3.2.1
+    -- Libtool's version-info data is translated into library versions in a
+    -- nontrivial way: so refer to libtool documentation.
     versionedExt :: String
-    versionedExt = case foreignLibVersionInfo flib of
-                     Nothing -> "so"
-                     Just v  -> "so" <.> libVersionNumberShow v
+    versionedExt =
+      let nums = foreignLibVersion flib os
+      in foldl (<.>) "so" (map show nums)
 
 -- | Name for the library when building.
 --
--- If the `lib-version-info` field of a foreign library target is set, we
--- need to incorporate that version into the SONAME field.
+-- If the `lib-version-info` field or the `lib-version-linux` field of
+-- a foreign library target is set, we need to incorporate that
+-- version into the SONAME field.
 --
 -- If a foreign library foo has lib-version-info 5:1:2, it should be
 -- built as libfoo.so.3.2.1.  We want it to get soname libfoo.so.3.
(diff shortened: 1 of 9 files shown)


abooij:
@ezyang The code is now ready for review. What's missing is documentation: I'll write something in Cabal/doc/developing-packages.rst. Anywhere else? Is the code documented clearly?

After review I can clean up the history a little bit. Or just make it one big commit?

ezyang:
You're not done until there's documentation ;) I think something nestled with the existing foreign library docs would make sense.

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -1489,8 +1493,23 @@ installFLib verbosity lbi targetDir builtDir _pkg flib =
...
+        else installOrdinaryFile   verbosity src dst
+      -- Now install appropriate symlinks if library is versioned
+      when (isJust (foreignLibELFVersion flib)) $ do
+          let (Platform _ os) = hostPlatform lbi
+          when (os /= Linux) $ die
+            "Can't install foreign-library symlink on non-Linux OS"
+#ifndef mingw32_HOST_OS
+          -- createSymbolicLink file1 file2 creates a symbolic link
+          -- named file2 which points to the file file1.
+          createSymbolicLink name (dstDir </> flibBuildName lbi flib)
```
abooij: I will add a comment to that effect.

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -873,6 +874,43 @@ flibTargetName lbi flib =
...
+--
+-- This method allows to adjust the name of the library at build time
+-- such that the correct soname can be set.
+flibBuildName :: LocalBuildInfo -> ForeignLib -> String
+flibBuildName lbi flib
+  | (platformOS (hostPlatform lbi), foreignLibType flib) ==
+    (Linux, ForeignLibNativeShared)
+  = case foreignLibELFVersion flib of
+      Nothing -> "lib" ++ nm <.> "so"
+      Just v  -> "lib" ++ nm <.> "so" <.> show (head (versionNumbers v))
```
abooij: While I agree, it couldn't have resulted in errors, since a user can only set `v` to a non-null version.

Cabal/Distribution/Types/ForeignLib.hs
```diff
@@ -42,16 +53,81 @@ data ForeignLib = ForeignLib {
     }
     deriving (Generic, Show, Read, Eq, Typeable, Data)
 
+data LibVersionInfo = LibVersionInfo Int Int Int deriving (Data, Eq, Generic, Typeable)
```
abooij: I will refer to libtool more clearly here.
I can also add these accessors. There is also `libVersionInfoCRA` which gives you a tuple of the three numbers, but I can see how that is not sufficient and/or clear enough.

abooij:
@ezyang Thanks for your review.

> So.... I am seriously tempted to ditch libtool numbers, and just use the good, old, simple, "here is the version number that will get put in the SONAME", and tell users to update the major if they break ABI-compatibility, and minor if they add a new interface. [...] But I am open to being convinced.

Well, the advantage of version-info over version numbers is that it is more likely we will be able to parse version-info into the data required for non-Linux OSes. For example, we can't reliably deduce an age from major.minor.build (as people will just be randomly bumping version numbers). But it is important that we know both "current" and "current-age" on OSX. So while I agree with your worries about version-info, if there is any potential future wish for OSX support, this is the more general solution.

I can add documentation on how to set these values, if that'd help (basically repeating the three golden rules of version-info).

> Honestly, we should probably have an option to compute the patchlevel automatically from the Cabal version.

No! ABI versions are NOT equal to package versions, and this is already an important distinction made in C. This *does* cause problems.  And besides, in my use case this is absolutely undesirable.

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -878,20 +878,20 @@ flibTargetName lbi flib =
     platformOS :: Platform -> OS
     platformOS (Platform _arch os) = os
 
-    -- If a foreign lib foo has elf-version 3.2.1, it should be built
+    -- If a foreign lib foo has lib-version-info 5:1:2, it should be built
     -- as libfoo.so.3.2.1
```
abooij: I will add [an appropriate reference to] an explanation.

ezyang:
> No! ABI versions are NOT equal to package versions, and this is already an important distinction made in C. This does cause problems. And besides, in my use case this is absolutely undesirable.

Even the revision level? According to the libtool guidance, you are supposed to bump revision whenever there is a source code change. I guess, maybe not if the C library didn't change (even though the hs library did)?

abooij:
(Also as discussed on IRC:)
For example: there's no reason why the ABI changes with every release of the package. So then there's no reason to even bump the "revision" field. And one might want to make a major change to the haskell code without modifying the ABI (in which case you would update the package version but not the ABI version), or vice versa (in which case you would increase the major version of the ABI, but only make an incremental version bump for the Haskell package). ABI versions are, to some extent, independent of release versions.

ezyang:
This looks reasonable.

abooij:
I wrote some documentation. So this work now seems ready. Final feedback welcome.

ezyang:
Ship it!

PR #4075, opened 2016-11-02
Add version number to file monitor cache.

### LINKED ISSUE #4059: Excessive memory usage while "Resolving dependencies..."
Activity monitoring while `cabal new-build` in cabal repository root. Unfortunately I only have 8gb of RAM! 

bildschirmfoto 2016-10-29 um 19 18 33

ezyang:
I'm actually not very happy with this patch, because it
doesn't actually solve any real problems (the next time
we drastically change the binary format, anyone tracking
cabal-install HEAD is going to end up in trouble in
any case, because cabal-install version is not a fine-grained
enough verison, you also need build date.)

Fixes #4059.


ezyang:
I've got a better patch.

PR #4078, opened 2016-11-02
Add fingerprint of Generic representation when serializing.

### LINKED ISSUE #4059: Excessive memory usage while "Resolving dependencies..."

ezyang:
The idea is we can use Rep to get a full, structural representation
of a type, and the fingerprint it using Typeable.  This gives
us a very concise way of fingerprinting our Binary representation.

This patch is not completely correct; the fingerprint needs
to be overridable when someone writes a custom Binary instance.
But this should be "good enough" in practice; we're not using
these fingerprints to check anything security critical.

TODO: Not sure if I have tagged all the call-sites which could
profit from this.

Fixes #4059.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -235,6 +235,7 @@ library
     Distribution.Utils.NubList
     Distribution.Utils.ShortText
     Distribution.Utils.Progress
+    Distribution.Utils.BinaryWithFingerprint
     Distribution.Verbosity
     Distribution.Version
     Language.Haskell.Extension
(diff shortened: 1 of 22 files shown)


23Skidoo:
Seems to fail on GHC < 7.10.

ezyang:
My plan is to just drop < GHC 7.10 support (so, we just won't add fingerprints in that case.)

PR #4082, opened 2016-11-04
No LambdaCase, resolve #4079

### LINKED ISSUE #4079: LambdaCase in Parsec is not supported by GHC 7.4
Self-explanatory. CC @phadej

CI doesn't catch because we don't build with Parsec except on 8.0.

phadej:


23Skidoo:
Merged, thanks!

PR #4083, opened 2016-11-04
Add more fields to plan.json schema

hvr:
This adds the following new fields:

 - `.compiler-id`
 - `.os`
 - `.arch`
 - `.install-plan[].pkg-name`
 - `.install-plan[].pkg-version`
 - `.install-plan[].pkg-src-sha256`

Having separate "pkg-{name,version}" fields is desirable because the "id"-field
is actually a UnitId which is supposed to be an opaque id which may not
always be easily convertible into a PackageId via string operations.

The "pkg-src-sha256" is generally useful to uniquely identify a
source-tarball (when the package-id is not exact enough).

Finally, the compiler-id/os/arch information is needed to determine the
values of the respective impl()/os()/arch() predicates used in
conditionals in .cabal files.

Here's an example demonstrating the new fields:

```js
    {
      "cabal-version": "1.25.0.0",
      "cabal-lib-version": "1.25.0.0",
      "compiler-id": "ghc-8.0.1",
      "os": "linux",
      "arch": "x86_64",
      "install-plan": [
	{
	  "type": "configured",
	  "id": "StateVar-1.1.0.4-048d5d25d5813f17f152f9e766fe3ea0ef6f5317439aa115f3fc24e53a0e3c17",
	  "pkg-name": "StateVar",
	  "pkg-version": "1.1.0.4",
	  "flags": {},
	  "style": "global",
	  "pkg-src-sha256": "7ad68decb5c9a76f83c95ece5fa13d1b053e4fb1079bd2d3538f6b05014dffb7",
	  "depends": [
	    "base-4.9.0.0",
	    "stm-2.4.4.1-2c7789312d2396a91db173e1ae8cf3a9af3dd18de42bb1021c4978fbab49c191",
	    "transformers-0.5.2.0"
	  ],
	  "exe-depends": [],
	  "component-name": "lib"
	},
        {
          "type": "pre-existing",
          "id": "rts",
          "pkg-name": "rts",
          "pkg-version": "1.0",
          "depends": []
        },
	{
	  "type": "configured",
	  "id": "hackage-matrix-builder3-0.3-inplace-matrix-lib",
	  "pkg-name": "hackage-matrix-builder3",
	  "pkg-version": "0.3",
	  "component-name": "lib:matrix-lib"
	  // ...
	}
	// ...
    }
```

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanOutput.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanOutput.hs
@@ -19,6 +19,7 @@ import           Distribution.Client.ProjectPlanning.Types
 import           Distribution.Client.ProjectBuilding.Types
 import           Distribution.Client.DistDirLayout
 import           Distribution.Client.Types (confInstId)
+import           Distribution.Client.PackageHash (showHashValue)
 
 import qualified Distribution.Client.InstallPlan as InstallPlan
 import qualified Distribution.Client.Utils.Json as J
@@ -27,12 +28,13 @@ import qualified Distribution.Simple.InstallDirs as InstallDirs
 import qualified Distribution.Solver.Types.ComponentDeps as ComponentDeps
 
 import           Distribution.Package
+import           Distribution.System
 import           Distribution.InstalledPackageInfo (InstalledPackageInfo)
 import qualified Distribution.PackageDescription as PD
 import           Distribution.Compiler (CompilerFlavor(GHC))
 import           Distribution.Simple.Compiler
                    ( PackageDBStack, PackageDB(..)
-                   , compilerVersion, compilerFlavor )
+                   , compilerVersion, compilerFlavor, showCompilerId )
 import           Distribution.Simple.GHC
                    ( getImplInfo, GhcImplInfo(supportsPkgEnvFiles)
                    , GhcEnvironmentFileEntry(..), simpleGhcEnvironmentFile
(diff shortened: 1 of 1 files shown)


hvr:
and here's an example where `id` contains less information than `package-id`

```js
    {
      "type": "pre-existing",
      "id": "rts",
      "package-id": "rts-1.0",
      "depends": []
    }
```

ezyang:
Modulo name bikeshedding looks good to me. (For example, we might consider splitting package-id into name/version, which then matches the format in the installed package database.)

hvr:
@ezyang I don't mind splitting it up (it may have minor benefits for simpler tooling (think e.g. `jq`) for which it's more awkward to split up names); shall we do it? I'd split it up as e.g.

```json
    {
      "type": "pre-existing",
      "id": "rts",
      "package-name": "rts",
      "package-version": "1.0",
      "depends": []
    }
```

Ericson2314:
A cursory glance makes me think this is probably good for my planned Nix stuff, so +1 from me.

ezyang:
Sure, let's do it.

hvr:
@ezyang I've updated the patch; I opted for `pkg-` as the fields started looking a bit noisy with a written out "`package-`". I also added os/arch which I forgot to add in the original patch.

ezyang:
lgtm but make the build green!

PR #4084, opened 2016-11-04
Correctly record internal dependencies in compLibDependencies, report…

ezyang:
…ed by @hvr.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning.hs
@@ -1252,7 +1252,8 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
             -- compNonSetupDependencies are defined when we define
             -- 'elab'.
             compLibDependencies =
-                concatMap (elaborateLibSolverId mapDep) external_lib_dep_sids
+                -- concatMap (elaborateLibSolverId mapDep) external_lib_dep_sids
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1252,7 +1252,8 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
             -- compNonSetupDependencies are defined when we define
             -- 'elab'.
             compLibDependencies =
-                concatMap (elaborateLibSolverId mapDep) external_lib_dep_sids
+                -- concatMap (elaborateLibSolverId mapDep) external_lib_dep_sids
```
hvr: do you plan to keep that as a comment there?

hvr:
I tried this locally and this fix appears to do what I expect it to; so lgtm!

23Skidoo:
Do we have a ticket about the problem this is supposed to fix?

hvr:
@23Skidoo not afaik, I was talking to @ezyang on IRC about `plan.json` and wondered why it was missing those intra-pkg refs... :-)

ezyang:
I'm working on an improved version of the patch which tightens up ./Setup configure checks to detect when we aren't feeding enough `--dependency` flags, as was the case here.

hvr:
@ezyang what's the status on this? does it need more testing?

ezyang:
Just needs someone to review the latest patch :)

PR #4086, opened 2016-11-05
Improve error output when package fails to build.

### LINKED ISSUE #4081: BuildOutcomes should save information about components
You have a package `foo` defining a single library and multiple executables that depend on that library. If the library fails to build, you get the following message with `new-build`:

```
$ cabal new-build
[...]
cabal: Failed to build foo-1.0 (which is required by foo,
foo and others).
Failed to build foo-1.0.
Failed to build foo-1.0.
Failed to build foo-1.0.
Failed to build foo-1.0.
```

Obviously, it'd be more user-friendly if it told you which specific exes failed to build instead of printing the package name multiple times.

ezyang:
New error message is:

cabal: Failed to build p-1.0 (which is required by exe:e4 from p-1.0, exe:e3
from p-1.0 and others).

Fixes #4081.

TODO: Use these utility functions in other appropriate places.


ezyang:
CC @23Skidoo who reported.

23Skidoo:
Thanks! Testing this right now.

23Skidoo:
This patch is an improvement, but it doesn't fully fix the issue:
```
cabal: Failed to build foo-1.0 (which is required by
exe:foo-test from foo-1.0, exe:foo-server from
foo-1.0 and others).
Failed to build foo-1.0.
Failed to build foo-1.0.
Failed to build foo-1.0.
Failed to build foo-1.0.
```
It looks like `renderFailureSummary` in `D.C.ProjectOrchestration` also needs updating.

ezyang:
(The new patch fixes that bug, I think, @23Skidoo !)

BardurArantsson:
Merging as it's a definite improvement; other fixes can be done as follow-ups.

PR #4091, opened 2016-11-08
Implement +nowrap verbosity flag to suppress log-msg line-wrapping

hvr:
Unfortunately, propagating the `Verbosity` to the `topHandler`
proves to be more complicated.

This is mostly a problem for `die` invocations which are translated
into user-error exceptions (which in turn then cause callstacks
to be printed; are we abusing `die`?)


ezyang:
I guess reworking just dieMsg for now is OK. LGTM.

hvr:
@ezyang btw, the confusing part about `dieMsg` is that it doesn't terminate program execution...

23Skidoo:
We should rename it then, to e.g. `noticeAlways` or `shout`.

23Skidoo:
Merged, thanks!

PR #4099, opened 2016-11-11
fix comment typo of security

juhp:


23Skidoo:
Merged, thanks! In the future, please use `[ci skip]` for documentation/comment-only commits to avoid triggering CI.

juhp:
Thank you!

Sure will do that :+1:

Issue #2965, opened 2015-12-19
[RFC] Plugin dependencies


ezyang:
Pursuant to https://ghc.haskell.org/trac/ghc/ticket/11244, it would be preferable for users to stop putting plugin dependencies in `build-depends` and a new field.

I propose to introduce a new field, `ghc-plugins`, which has the same format as `build-depends`, but adds dependencies on plugins. Packages specified here:

1. Do not get put in the `depends` in the installed package info (so they don't get linked in when you build an executable),
2. Are passed to GHC 8.0 as `-plugin-package-id` (with `-hide-all-plugin-packages`), which forces GHC to only use them to determine scoping of plugin modules, e.g. `-fplugin ModName`). Previous versions of GHC, they get passed in as normal `-package-id` flags for backwards compatibility

For the purposes of dependency resolution, they should be treated similarly to `build-tools`.

23Skidoo:
Sounds sensible.

/cc @dcoutts

ezyang:
OK, one minor complication here: if a user put a plugin dependency in `build-depends`, we want this to keep working. But this will stop working if Cabal unconditionally passes `-hide-all-plugin-packages` to GHC 8.0 or later. Perhaps use of `-hide-all-plugin-packages` should be toggleable, or maybe based on the version of Cabal dependency declared? (So, if a user requires a recent enough version of Cabal, then we assume that they correctly used `ghc-plugins`)

23Skidoo:
> or maybe based on the version of Cabal dependency declared? (So, if a user requires a recent enough version of Cabal, then we assume that they correctly used ghc-plugins)

I think that's better than adding yet another obscure switch. We'll need a version check for `ghc-plugins` anyway. We can also hardcode the list of known GHC plugins released on Hackage and warn if a GHC plugin is found in `build-depends`.

ezyang:
I previously claimed that `plugin-depends` should be implemented similarly to `build-tools`. But I looked more carefully at `build-tools` and actually, these are not handled at all by the solver (#220). So `build-tools` is not a good way of implementing this.

Instead, we need `plugin-depends` to be MERGED with `build-depends` when we are doing dependency solving; from the dependency solver's perspective, there's no distinction between these two fields. (Actually, if we wanted to be a bit more clever, each plugin dependency should be treated like a setup dependency, since it's OK if a plugin uses foo-0.1 but the main library uses foo-0.2: they won't actually interact.)

[...]

PR #4104, opened 2016-11-12
[WIP] implement tool-depends

### LINKED ISSUE #3708: New `tool-depend` goals to complement `build-tools`
When you write `build-tools: alex`, Cabal checks that an executable of this name is installed. It is jointly the responsibility of the package manager, e.g., `cabal-install`, to arrange for an executable named `alex` to be installed.

In a recent commit (c0a4860202393882d2fd0f4de253c3af3a092fe2) we added a special case in cabal-install to handle "well known" build-tools like alex and happy, building and installing them automatically when they are requested in `build-tools`. However, there remains a more general problem of letting users ask for an executable, which is known to live in a particular package. `build-tools` as currently specified cannot really handle this, because it is specified to only specify executable names (not package names.) So it seems we really need a new field to give this information. Here is the proposal:
- We introduce a new `tool-depends` field, which consists of `pkg:exename version-constraints` specifications, e.g., `tool-depends: gtk2hs-buildtools:gtk2hs2hs < 2.0`. This means that we must build the `gtk2hsc2hs` executable from `gtk2hs-buildtools` before building this package. In principle cabal-install need only build the `gtk2hsc2hs` executable, although it's also permissible to just build all the executables in the package (if, e.g., the Setup script doesn't support `per-component Cabal <https://github.com/ezyang/ghc-proposals/blob/master/proposals/0000-componentized-cabal.rst>`_). If qualification is omitted it is assumed that you need ALL executables from the package.
- `build-tools` refers exclusively to non-packaged executables, e.g. provided by the system. However, for backwards compatibility, we introduce an internal mapping hard-coded into cabal-install which maps some `build-tools` fields to package names. The intended interpretation is that `build-tools: happy > 0.4` actually elaborates to `tool-depends: happy:happy > 0.4`. ~~This mapping will be configurable using the `build-tool-packages` field in a `cabal-project` and also modifiable by command line argument. (Names can be bikeshedded of course.)~~
### Examples:
- `tool-depends: mytoolpkg`: cabal-install will install all executables defined by a package named `mytoolpkg`, and ensure they are available prior to building this package.
- `tool-depends: mytoolpkg:mytool`: cabal-install will install the executable `mytool` from `mytoolpkg` before building this package
- `build-tools: happy`: this is equivalent to writing `tool-depends: happy:happy`.
- `build-tools: someprog`: as before this does not effect solver behavior; there just simply needs to be a `someprog` in the path.

---

_Note: this was originally conceived by @ezyang as part of #220 (which has been addressed); however, since this represents an extension of the `.cabal` format and hasn't yet been implemented it's been split out into its own issue._

Ericson2314:
Implements much of #3708

Note the the exe component name is currently mandatory, unlike as specified in the original issue.


Cabal/Distribution/Package.hs
```diff
@@ -395,12 +396,20 @@ mkLegacyUnitId = newSimpleUnitId . mkComponentId . display
...
+                     VersionRange
+                     deriving (Generic, Read, Show, Eq, Typeable, Data)
+
 -- | Describes a legacy `build-tools`-style dependency on an executable
 --
 -- It is "legacy" because we do not know what the build-tool referred to. It
 -- could refer to a pkg-config executable (PkgconfigName), or an internal
 -- executable (UnqualComponentName). Thus the name is stringly typed.
---
+--g
```
ezyang: oops

Cabal/Distribution/Package.hs
```diff
@@ -395,12 +396,20 @@ mkLegacyUnitId = newSimpleUnitId . mkComponentId . display
 data Dependency = Dependency PackageName VersionRange
                   deriving (Generic, Read, Show, Eq, Typeable, Data)
 
+-- | Describes a dependency on an executable from a package
+--
+data ExeDependency = ExeDependency
+                     PackageName
+                     UnqualComponentName -- name of executable component of package
+                     VersionRange
+                     deriving (Generic, Read, Show, Eq, Typeable, Data)
```
ezyang: It is possible that you may want a helper function that takes an ExeDependency and gives a ComponentName (since the UnqualComponentName is only "implicitly" qualified.) I don't know if you'd actually use it.

Cabal/Distribution/Package/TextClass.hs
```diff
@@ -27,6 +27,18 @@ instance Text Dependency where
              Parse.skipSpaces
              return (Dependency name ver)
 
+instance Text ExeDependency where
+  disp (ExeDependency name exe ver) =
+    (disp name <<>> Disp.text ":" <<>> disp exe) <+> disp ver
+
+  parse = do name <- parse
+             _ <- Parse.char ':'
+             exe <- parse
```
ezyang: OK, so right now you are forced to specify the executable name. That's fine (although our users might complain :)

Cabal/Distribution/Package.hs
Ericson2314: :)

Cabal/Distribution/FooBar.hs
```diff
@@ -0,0 +1,56 @@
...
+    whitelist = [ "hscolour", "haddock", "happy", "alex", "hsc2hs", "c2hs"
+                , "cpphs", "greencard"]
+    whiteMap  = Map.fromList $ flip map whitelist $ \n ->
+      (n, ExeDependency (mkPackageName n) (mkUnqualComponentName n) reqVer)
+
+-- | Get everything from "tool-depends", along with entries from "build-tools"
+-- that we know how to desugar.
+--
+-- This should almost always be used instead of just accessing the
+-- `toolDepends` field directly.
```
ezyang: Ok, if that's true there better be a comment on toolDepends telling people about it, or rename that field to something to discourage people from using it directoy.

Cabal/Distribution/FooBar.hs
```diff
@@ -0,0 +1,56 @@
...
+                       -> BuildInfo
+                       -> [ExeDependency]
+getAllToolDependencies pkg bi =
+  toolDepends bi ++ mapMaybe (desugarBuildTool pkg) (buildTools bi)
+
+-- | Does the given executable dependency map to this current package?
+--
+-- This is a tiny function, but used in a number of places. Note that the
+-- version bounds and components of the package are unchecked. This is because
+-- we sanitize exe deps so that the matching name implies these other
```
ezyang: Where does this sanitizing happen.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -524,13 +525,42 @@ checkFields pkg =
         ++ "for example 'tested-with: GHC==6.10.4, GHC==6.12.3' and not "
         ++ "'tested-with: GHC==6.10.4 && ==6.12.3'."
 
-  , check (not (null buildDependsRangeOnInternalLibrary)) $
+  , check (not (null depInternalLibraryWithExtraVersion)) $
```
ezyang: I didn't read this code carefully. Good tests would tickle all of these cases.

ezyang:
Looking good. Obviously you still need to do the cabal-install bits.

Cabal/Distribution/Package/TextClass.hs
Ericson2314: Besides laziness, I figured a blanket depend-on-all-the-exes dependency might complicate making the solver use deps on specific components. Then again we may need a blanket package dep anyways.

Cabal/Distribution/Package.hs
Ericson2314: Sure, doesn't hurt to add it now. Good hint on unqual vs qual for anyone working with the code base.

Cabal/Distribution/FooBar.hs
Ericson2314: `PackageDescription.Check` does sanitation happen in places besides that?

ezyang: I think the overall codebase doesn't use "sanitization" as the term here. Furthermore, not all checks in `PackageDescription.Check` are necessarily run all the time: a package can still be built even if some of the checks fail. It is better to say that there is some invariant that is expected to hold; in that case the invariant description should be at the data type definition.

Ericson2314:
How should I write the (presumably system) tests for those tests? I see the cabal files in `PackageTests`, but not sure where the code verifying stdout / stderr is.

23Skidoo:
You may want to wait until #4095 is merged.

Ericson2314:
Hmm good idea

Ericson2314:
Hmm, so `InternalLibrary4` explicitly tests that version bounds can trigger the use of an external library instead of internal one. This is in direct conflict with the `buildDependsRangeOnInternalLibrary` test this patch expands---it says version bounds don't matter because the internal library is always used. I don't really care which semantics we go with, but we should be consistent for all 3 of "build-tools", "build-depends", and "tool-depends".

ezyang:
@Ericson2314 We got rid of this feature in af24cefe12adf0f0e1745b70db45507ae5b45d06. Internal library should always be used.

Cabal/Distribution/FooBar.hs
Ericson2314: I added some longer documentation to the relevant fields in the top commit. I'll also grep at the end to make sure nothing is lurking.

Cabal/Distribution/PackageDescription/Check.hs
Ericson2314: Whew! A lot of copy pasta but tested everything.

Ericson2314:
Ok things left:
 - cabal-install's init
 - cabal-install's tests
 - Real name for `Distribution.FooBar` :D

Cabal/Distribution/Backpack/ComponentsGraph.hs
```diff
@@ -55,9 +56,8 @@ toComponentsGraph enabled pkg_descr =
     -- The dependencies for the given component
     componentDeps component =
          [ CExeName toolname
-         | LegacyExeDependency name _ <- buildTools bi
-         , let toolname = mkUnqualComponentName name
-         , toolname `elem` map exeName (executables pkg_descr) ]
+         | (ExeDependency _ toolname _)
+            <- getAllInternalToolDependencies pkg_descr bi ]
```
ezyang: Good.

Cabal/Distribution/Types/BuildInfo.hs
```diff
@@ -30,8 +30,21 @@ import Language.Haskell.Extension
...
+        -- Unless use are very sure what you are doing, use the function in
+        -- `Distribution.FooBar` rather than accessing this field directly.
+        buildTools        :: [LegacyExeDependency],
+        -- | Haskell tools needed to build this bit
+        --
+        -- This field is better than "build-tools" because it allows one to
+        -- precisely specify an executable in a package.
+        --
+        -- Unless use are very sure what you are doing, use the function in
+        -- `Distribution.FooBar` rather than accessing this field directly.
```
ezyang: But... which function/!

ezyang:
Lookin' good!

ezyang:
NB: this shouldn't be merged until "Ok things left:" list is done.

hvr:
@ezyang what's the first todo item about? if it's about `cabal init` I don't think it's essential for this PR

ezyang:
Looking at the source code, it doesn't look like the new-build code has been adjusted for tool-depends yet. But maybe no changes were needed; I don't know if it works or not; the tests would be able to say so.

Ericson2314:
So my test `BuildDependsExtra` that fails in https://travis-ci.org/haskell/cabal/jobs/181701243 passed locally when run by hand. I just rebased this in hopes it goes away, but I doubt it will as it survived a rebase before.

hvr:
Btw, one commit message mentions

> Now for all three of "build-tools", "tool-depends", and "build-inputs":

what are `build-inputs`?

Ericson2314:
@hvr `s/build-inputs/build-depends` forgot to fix that commit message. (`buildInputs` is a roughly-analogous nixpkgs ideom, and sometimes i get the jargon jumbled.)

Ericson2314:
`build-exe-depends` sounds better to me. But, I want to eventually support build-time-only TH deps; is `build-lib-depends` properly distinguished from `build-depends`?

> A subtle semantic issue is also whether a benchmark or test depending on some `build-tool` properly tracks the dependency of the test's/benchmark's run-time or rather only its build-time.

I don't know what you mean.

hvr:
Can you provide a simple semi-concrete example/use-case for built-time only TH-deps?

As for the test/benchmark thing,

consider e.g.

```
test-suite foo
    tool-depends: some-data-generator:bar
    main-is: ...
```

Now in order to compile `test:foo` you wouldn't require some-data-generator:bar` yet; only when you actually execute via `cabal new-test foo`. Specifically, it may happen that:

 - `cabal new-build test:foo` triggers `some-data-generator:exe:bar` to be build needlessly (or at least, too early, possibly creating a bottleneck in cabal's job parallelism)
 - if `foo` is already built, but `some-data-generator` was removed from the nix-store for some reason, `cabal new-test foo` may fail to ensure that `exe:bar` is built (I may be wrong here though as I'm not sure if cabal's logic wouldn't try to rebuild it in any case)

Ericson2314:
I understood `tool-depends` to always a compiler-time-only dependency on an executable. Shouldn't run-time-only executable dependencies be specified in their own field?

Ericson2314:
> Can you provide a simple semi-concrete example/use-case for built-time only TH-deps?

Sorry this answer depends on a big other project, but with a properly-phased TH and cross compiling (https://mail.haskell.org/pipermail/ghc-devs/2016-January/011066.html, http://blog.ezyang.com/2016/07/what-template-haskell-gets-wrong-and-racket-gets-right/), it would be very normal use packages at compiler time that do to IO, threading, etc cannot be built it run time. And for packages that can be built for both, there is still no reason to do the extra work of doing so.

hvr:
> Shouldn't run-time-only executable dependencies be specified in their own field?

yes, exactly... and how would you name that field (if the name `tool-depends` is already taken)? :-)

Ericson2314:
> yes, exactly... and how would you name that field (if the name tool-depends is already taken)? :-)

Whew! I think the "also" through me off: I didn't realize you were still talking about users being confused rather than a subtle semantic issue for us all to wrap our heads around. Yeah so ideally `build-` would mean `build-time-only`, but then the `build-` in `build-depends` doesn't match that at all. `build-depends` is IMO a horribly undescriptive name all around, but we are stuck with it.

I generally prefer "build-time" to "compiler-time" as the latter strikes me as over precise, but maybe `compile-` is the best we can do:

 - `compile-exe-depends`: compile-time-only exe dependencies, `tool-depends` currently
 - `compile-lib-depends`: compile-time-only lib dependencies (i.e. next-gen TH if we get that, GHC plugins)
 - `run-lib-depends`: better name for `build-depends` (I personally consider static linking the build-time aggregation of run-time deps, but I understand if others disagree or find this confusing)
 - `run-exe-depends`: kind of a smell as its so much better to use a library than shell out, but would be your hypothetical testing use-case.

I'm also confused why the verb ("depends") was chosen instead of the noun ("dependencies", "deps" for short) but this is not important.
----
Anyone have any ideas on why that test isn't passing? I really want to get this merged and fixing the names is just a sed away.

ezyang:
Hi @Ericson2314, I figured out why this test is failing on CI but not locally: for `setupAndCabalTest`, in CI we test both that `./Setup configure` works, but also `cabal configure` works. The test is failing when you try to run `cabal configure`, it successfully depsolves but then install plan validation fails. Does this make sense?

If you want to run the test locally, just run `cabal-tests --with-cabal=/path/to/locally/built/cabal setup.test.hs --skip-setup-tests` and that will ONLY run the cabal test.

hvr:
@Ericson2314 I generally agree with your classification (btw, what about a s/run/install/ rename?). And yesterday @ezyang mentioned that there's another dimension to consider (and/or get headaches about): host/target deps when cross-compiling. :-))

ezyang:
First, I find compile-time/run-time dependency somewhat confusing for libraries, because for any library you always need it a compile time. The more salient difference is whether or not you needed the dep compiled for the host, or the target.

Second, here are the three primary constraints on our naming scheme:

1. We must preserve `build-depends` as the name of host+target library dependencies for Template Haskell and plugin-using packages (build-depends brings modules into scope for both TH and normal compilation; it is also traditionally used to bring plugins into scope, see #2965).
2. We need a name for "host only" executable dependencies, because that is what this commit is about.
3. We need a name for "host only" library dependencies. This would be immediately useful for plugins (#2965) and also eventually useful for TH_ONLY packages.

Do we need a way of specifying "target only" library dependencies? I think that is the tail wagging the dog: for all Haskell packages which do NOT use Template Haskell (and there are many), `build-depends` effectively specifies target only dependencies. So if you introduce a new, blessed, field for target only lib dependencies, you are also effectively saying that all of these packages should change. I don't think that's right: the onus should be on *Template Haskell* using packages to change their usage and become more accurate. In fact, you could nearly directly interpret build-depends in the following way: if TemplateHaskell is listed in other-extensions, build-depends is host+target, otherwise, build-depends is target only. (And then we can talk about new syntax for disabling this interpretation when a TH using package has specified things precisely.)

So, to summarize, here is my suggestion:

1. `build-depends` should denote target lib dependencies, except if TemplateHaskell is in other-extensions, in which case it's target+host lib dependencies (and eventually, if we get host-only TH, there should be a way to force build-depends to be target only.)
2. I was pretty fond of `plugin-depends` for host lib dependencies, but it doesn't adequately convey the fact that it is also applicable for Template Haskell. But I like the idea of sticking to this nomenclature, because it's the nomenclature GHC adopted (`-plugin-package`) and there isn't much reason for us to "not" consider TH as a kind of plugin.
3. So, what's wrong with `tool-depends`? The primary objection is that it might be a compile-time tool, or a run-time tool. So let's call it `build-tool-depends` (and call the other thing `runtime-tool-depends`?) In the case of tools, the host/target distinction is pretty obvious; compile time tools run on host, run time tools run on target.

Ericson2314:
@hvr yeah kinda like what @ezyang says, and despite what I said before, I believe more in host vs target than build-time vs run-time.

Ericson2314:
@ezyang hmm so I tried the test with an old `cabal` executable (the one I used to bootstrap, few days old master). It also fails with the same internal error on `cabal-testsuite/PackageTests/InternalVersions/BuildDepends*/`. Cabal-install must already not be happy using `foo:lib:bar` for `build-depends: bar` in `foo`'s default library.

ezyang:
Uh, I never claimed that the `foo:lib:bar` syntax in `build-depends` worked?

It's definitely possible that the test just doesn't work with `cabal`; in that case, switch the top-level call to `setupTest`.

Ericson2314:
@ezyang Sorry I was unnecessarily unclear there. I did not mean I used the `foo:lib:bar` syntax, but just meant to precisely identifier the component I'd expect cabal-install to use for the plan. See https://github.com/haskell/cabal/blob/7ba095bf0e599119de3585c4f0e07c40b637c362/cabal-testsuite/PackageTests/InternalVersions/BuildDependsBad/build-depends-bad-version.cabal#L11 for what I wrote.

I think I am going to change this to be more traditional: internal exe depends on default lib. That still suffices to check the version bound logic, and will stay correct going forward as the current plan last I checked is to use `mixin:` for internal lib deps anyways.

Ericson2314:
Ok, at least locally making the test more traditional worked! Assuming Travis agrees that means what's left is:

 - [ ] `sed -e 's/tool-depends/build-tool-depends/g $(git ls-files)`. Conveniently, the new module is already called `Distribution.Simple.BuildToolDepends`!
 - [ ] Add end-to-end cabal-install test with build tool that actually does something. Do we need one with an internal exe for cabal alone too?
-  [ ] cabal-install's init, if that needs to be done in this PR.

[N.B. In response to an earlier comment, I had original grepped for `build-tools` to see anything else that should be updated, and the remaining matches seemed to all be init-related. I don't think there is any new-build code that needs to be updated.]

Ericson2314:
Ok if test pass all that remains is the init code!

Ericson2314:
...and they don't pass. In https://travis-ci.org/haskell/cabal/jobs/184091606:

1. The exe isn't on the PATH for some reason,
2. Due to some inferred version requirement my "extranous version bound" warning is spuriously going off.

ezyang:
For the spurious warning, see #4169

As for the exe not being in the path, you're going to have to debug that. I can point you at the logic that plumbs this too. Maybe we can refactor it so that whatever bug you did, won't happen in the future.

Ericson2314:
@ezyang thanks, again! Glad to hear #4169's desired fix is orthogonal to my changes :).

For the path issue, actually I don't think this is a regression: I just never tested things manually nor editted any such logic, and just hoped things would keep on working because `build-tools` worked. So yes, if you could point me to the current logic for the PATH that would be helpful.

Also I modeled my end-to-end test from `cabal-install/tests/IntegrationTests/new-build/BuildToolsPath` which seems to copied exactly from `cabal-testsuite/PackageTests/BuildToolsPath`. Since (as I now know), package tests also get run with cabal-install, should the new-build version be made more different--e.g. using separate packages with one of the special-cased names as opposed to an internal external executable? At the very least, that would be a useful test if indeed my PATH error is nothing new.

ezyang:
Let me first clarify something: I'm pretty sure `build-tools` on one of the hardcoded package names DOES work on HEAD (we have a test for it). When I look at your most recent CI log, the error for your newly constructed test is:

```
    BuildToolsExternal.sh (ignoring stdout+stderr):                     FAIL (1.73s)
      Unexpected exit status.
      
      Exit status: ExitFailure 1
      Working directory: /tmp/cabal-install-test-21100
      
      <stdout> was:
      Resolving dependencies...
      
      
      <stderr> was:
      cabal: Unknown build target 'build-tools-external:hello-world'.
      The project has no package 'build-tools-external'.
      Perhaps you meant the package 'build-tool-depends'?
```

...because you are cd'ing into the wrong directory. For the record, the pre-existing build-tools test is `cabal-install/tests/IntegrationTests/new-build/external_build_tools.sh`

In any case, why is the new code not working? The crux of the issue is probably the binding of `compExeDependencies` in `Distribution/Client/ProjectPlanning.hs`. It seems some work has to be done here.

1. This code assumes that exe_deps0 from SolverPackage is exactly the list of packages, for which we want to bring ALL the executables into scope. The subsequent code which refers to this variable assumes that we just want to suck in all the executables (in particular `external_exe_dep_sids`, which is in turn used by `compExeDependencies`). This is clearly not what you want with `build-tool-depends`, because you have qualified a specific executable here. So probably what you actually need to do is grovel in `buildToolDepends` for package names, look those package names up in `exe_deps0` to find the `SolverId`, and then rewrite `elaborateExeSolverId` so that it can select a specific executable based on the name you have there (right now it just returns all of the executables of the package)
2. You'll also need to update `elaborateSolverToPackage`, which also knows about external deps (factor out into a function?!)

Ericson2314:
> because you are cd'ing into the wrong directory

Ok I admit I didn't manually test the script and rather manually ran it's one-line, because all the scripts seem to lack an executable bit. Should the executable bit be set?

> For the record, the pre-existing build-tools test is cabal-install/tests/IntegrationTests/new-build/external_build_tools.sh

Wow.... it was completely in sight and I missed it.

I'll fix all that and then look at project planning. Thanks again for all your guidance, @ezyang!

Ericson2314:
Ok, docs are one also all that remains is `cabal init`, and making the PATH more fine-grained a la https://github.com/haskell/cabal/pull/4104#issuecomment-267484861 . OK to do those in future PRs?

ezyang:
Yep. Anything in particular you want reviewed before merging?

hvr:
let's get this merged soon before the PR starts to "conflict-rot" :-)

ezyang:
@Ericson2314 It would be great if we could have a changelog entry too!

phadej:
I can't wait till this gets to hvr's ppa repository so I can try it out wild!

hvr:
@phadej ...I got the hint ;-)

Ericson2314:
@ezyang OK I will do that in the PR that https://github.com/Ericson2314/cabal/tree/build-tool-depends-glob turns into.

ezyang:
Thanks!

PR #4107, opened 2016-11-13
Dedicated modules for `*Dependency` types and `UnqualComponentName`

Ericson2314:
Also gets rid of `Distribution.Packages.TextClass`


Ericson2314:
So AppVeyor worked every time but this, and the failure is upstream, so either upstream needs to be fixed or we need to tighten a bound--both of which are orthogonal to this.
I feel bad about merging but this is big and can bit-rot quick so gonna merge anyways.

PR #4110, opened 2016-11-14
Solver: Fix space leak in 'addlinking' (issue #2899).

grayjay:
The solver creates linked nodes by copying existing unlinked nodes. Previously,
the subtrees shared data, which caused a space leak. This commit combines the
"build" and "addLinking" tree traversals so that the nodes are copied before
they are expanded into full trees.

This should prevent sharing more reliably than my previous PR, #3530.

I'd like to add a regression test, but I'm not sure how to do it without significantly lengthening the build.  I wrote a test with the solver DSL (https://github.com/grayjay/cabal/commit/1f1b95d7b4d836ac79a60f943bb36e0ac6ef41e8) that fails when I limit the heap size.  Do you think we should add another test suite that we run with a limited heap?  I can't add the test to the existing unit test suite, because some of the quickcheck tests already cause spikes in memory usage.

EDIT: The diff for the main commit isn't aligned well, so here is a diff with different options: https://rawgit.com/grayjay/9d842a8c8c1b04b1a1f0d8eca4a8a2ad/raw/467ffdf7b523272a5a7118f940af9e7f2634a0eb/cabal-pr-4110.html

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -1,4 +1,5 @@
-module Distribution.Solver.Modular.Builder (buildTree) where
+{-# LANGUAGE ScopedTypeVariables #-}
+module Distribution.Solver.Modular.Builder (buildTree, addLinking) where
 
 -- Building the search tree.
 --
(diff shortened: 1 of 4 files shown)


23Skidoo:
> Do you think we should add another test suite that we run with a limited heap?

This is a good idea IMO.

ezyang:
If you want to merge this without the test I'm OK with that.

grayjay:
@ezyang Thanks.  I thought it would be easier to develop the test while the fix is still easy to revert, though.

I rebased on top of #4122, so the first four commits are from that PR.

grayjay:
Rebased onto master

23Skidoo:
I think that this should also go in.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #4115, opened 2016-11-15
Add `--index-state` support to `cabal get`

hvr:
With this `cabal get` gains the ability to

 a) unpack and update with the .cabal revision which was
    active at the given index-state, as well as

 b) allow to unpack the most recent package version
    (while satisfying imposed version constraints) as of
    the given index-state

The `--index-state` flag can be combined with `--pristine` if only b) is
desired.

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -2274,6 +2274,15 @@ The ``get`` command supports the following options:
     Fork the package's source repository using the appropriate version
     control system. The optional argument allows to choose a specific
     repository kind.
+``--index-state`` *[HEAD\|@<unix-timestamp>\|<iso8601-utc-timestamp>]*
+    Use source package index state as it existed at a previous time. Accepts
+    unix-timestamps (e.g. ``@1474732068``), ISO8601 UTC timestamps (e.g.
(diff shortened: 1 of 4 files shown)


23Skidoo:
May be also worth adding `--x-hackage-revision=REVNO`.

hvr:
@23Skidoo Yeah, but that's a bit harder to implement (and then there's the question whether to trust the x-revision field or rather the tar-entry ordering in the .tar.gz stream; there's a few unfortunate cases where it doesn't coincide) :-) 

in any case, it would go beyond the scope of this PR as it would require some additional refactoring in the index-reading code

23Skidoo:
> (and then there's the question whether to trust the x-revision field or rather the tar-entry ordering in the .tar.gz stream; there's a few unfortunate cases where it doesn't coincide)

IMO it's best to fail loudly in that case and point the user to `--index-state`.

> in any case, it would go beyond the scope of this PR as it would require some additional refactoring in the index-reading code

Yep, implementing `--x-hackage-revision` is of course not a requirement for merging this PR. But can you please add a changelog note and expand the `cabal get` section in `developing-packages.rst`?

hvr:
@23Skidoo I've updated changelog & manual; anything else?

23Skidoo:
Merged, thanks!

PR #4116, opened 2016-11-16
Move functions for creating solver HUnit test cases into a new module.

grayjay:
This commit moves the utility functions from UnitTests.Distribution.Solver.Modular.Solver to a new module, UnitTests.Distribution.Solver.Modular.DSL.TestCaseUtils.

I'm planning to use these functions in a performance test for #4110.


grayjay:
Thanks!

PR #4122, opened 2016-11-20
Add test suite with two basic tests for solver space leaks.

grayjay:
The new test suite uses the solver DSL and is built with `-with-rtsopts=-M4M`.  I think this is a better way to check for space leaks than using the cabal executable, because it isolates the solver and uses simpler packages.  Here's an example where I introduced a space leak by removing the bang patterns on the Int parameter in https://github.com/haskell/cabal/blob/e496b09eb3e9d15ca3380155a85c9a5c7a3587d5/cabal-install/Distribution/Solver/Modular/Message.hs#L54:

Solver DSL (There are three tests, because I used PR #4110.)
solver-dsl-tests-with-space-leak

cabal executable
cabal-with-space-leak

I also had to fix two space leaks to make the tests pass.  One occurs when the solver always has exactly one choice for the next goal.  The other is caused by a bug in GHC 7.6.3.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Explore.hs
+++ b/cabal-install/Distribution/Solver/Modular/Explore.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 module Distribution.Solver.Modular.Explore
     ( backjump
@@ -60,7 +61,7 @@ backjump (EnableBackjumping enableBj) var initial xs =
           | otherwise                            = f (csAcc `CS.union` cs) cm'
 
     logBackjump :: ConflictSet QPN -> ConflictMap -> ConflictSetLog a
-    logBackjump cs cm = cm `seq` failWith (Failure cs Backjump) (cs, updateCM initial cm)
+    logBackjump cs !cm = failWith (Failure cs Backjump) (cs, updateCM initial cm)
                                    -- 'intial' instead of 'cs' here ---^
                                    -- since we do not want to double-count the
                                    -- additionally accumulated conflicts.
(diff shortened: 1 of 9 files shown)


cabal-install/Distribution/Solver/Modular/Explore.hs
```diff
@@ -60,7 +60,7 @@ backjump (EnableBackjumping enableBj) var initial xs =
           | otherwise                            = f (csAcc `CS.union` cs) cm'
 
     logBackjump :: ConflictSet QPN -> ConflictMap -> ConflictSetLog a
-    logBackjump cs cm = failWith (Failure cs Backjump) (cs, updateCM initial cm)
+    logBackjump cs cm = cm `seq` failWith (Failure cs Backjump) (cs, updateCM initial cm)
```
fmthoma: I'd consider a bang pattern instead of `seq` for readability.

Also, in `exploreLog` below the `cm` is only updated when conflict counting is enabled, while it is always updated here. As you suggested in #3570, I think the `if countConflicts…` would make sense here as well. (Or, if you think that `seq`ing `cm` is already sufficient, remove the `if countConflicts` in `exploreLog` as well, which would IMHO improve readability).

grayjay: > I'd consider a bang pattern instead of seq for readability.

fixed

> Also, in exploreLog below the cm is only updated when conflict counting is enabled, while it is always updated here. As you suggested in #3960, I think the if countConflicts… would make sense here as well. (Or, if you think that seqing cm is already sufficient, remove the if countConflicts in exploreLog as well, which would IMHO improve readability).

I just made a minimal change to avoid conflicts with your PR, but you're right: it's simpler to always update the map.

grayjay:
@fmthoma Thanks for the review.  I applied your suggestions in the last two commits.

> Have you considered also limiting the stack size? Some space leaks tend to only build up the stack, but not the heap, and these would not be discovered here. (Example: foldr (+) 0 [1..])

Good idea.  It seems to work really well, because the test framework catches the exception and continues running the other tests.  Maybe we should also apply it to the other test suites.

> And just for curiosity: Why did you disable backjumping for your test cases? Would it make sense to include additional test cases where backjumping is enabled?

I thought it would be better to test with backjumping, but I didn't know how to ensure that the solver doesn't find the result immediately.  Then the tests wouldn't catch space leaks that only appear after a large area of the search tree has been explored.  Turning off backjumping forces the solver to explore the whole tree.

We could probably do something like count the number of nodes that the solver has searched and assert that it is high enough.  Then we would know when we needed to make the tests more difficult.  I think that should go in another PR, though.

fmthoma:
> Good idea. It seems to work really well, because the test framework catches the exception and continues running the other tests. Maybe we should also apply it to the other test suites.

The idea [isn't mine](http://neilmitchell.blogspot.de/2015/09/detecting-space-leaks.html), but yes, I think limiting the stack to a reasonable, but not overly small size for all tests is a good idea to detect space leaks early on. The only drawback is that when some test fails because of a newly introduced space leak, it may not be immediately obvious.

>  I think that should go in another PR, though.

Agreed :-)

grayjay:
Thanks!

PR #4124, opened 2016-11-20
cabal-install integration-tests replacement omnibus patch

ezyang:
**UPDATE.** This branch has been commandeered into a bigger integration-tests replacement patch

----

These flags allow users to configure the location of
cabal.project and dist-newstyle, respectively.  Primary
motivation is for the test suite, to allow non-clobbering
inplace testing.

Bikeshed points:

- GlobalFlags or InstallFlags?  I picked GlobalFlags as it
  seemed more appropriate from a semantic perspective, but
  the downside is that you must now specify the flag
  before new-build, i.e., cabal --dist-dir=... new-build.
  However, if we ever support a Nix-style command that
  doesn't take the full complement of InstallFlags, GlobalFlags
  is the only place it could go.

- Flag names?  I think --project-file is good but perhaps
  someone has a better idea for --dist-dir.

- Is this in duplication with globalConfigFile?  I have
  no idea what that parameter does, so I don't think so.

It's not too easy to make the value of 'dist-dir' saveable
to the cabal.project file, as we'd prefer to load directly
from the cache if cabal.project is not updated.

CC @hvr


hvr:
As pointed out on IRC, there's already an option in `cabal install` named

```
    --builddir=DIR                   The directory where Cabal puts generated
                                     build files (default dist)
```

ezyang:
It's not entirely clear to me that reusing `--builddir` here is the right thing to do, since `--builddir` needs to have different, distinct values for every subcall to Cabal. Could be error-prone.

23Skidoo:
From the user's point of view, what would be the diffrerence between `--dist-dir` and `--builddir`? We must explain it well to our users if we are to preserve both.

ezyang:
I'm OK with these not being user facing.

A `--builddir` has a specific layout: you can find the executable named foo at `$builddir/build/foo/foo`. A `--dist-dir` guarantees a JSON plan file at `$distdir/cache/plan.json` but otherwise the format is unspecified.

hvr:
Otoh, weren't we going to rename the default `dist-newstyle` to `dist` anyway?

23Skidoo:
> I'm OK with these not being user facing.

Then I guess we should hide them in `--help` output?

ezyang:
Based on #4127, I've been convinced that we should use `--builddir` to control the location of `dist-newstyle`. Note that `--builddir` is a flag that is separately represented in every flags data structure (e.g., `ConfigFlags`, `InstallFlags`, `CopyFlags`; see `****DistPref`) so some care must be taken. We cannot convert it into a global flag as that will break backwards compatibility with scripts that expect to be able pass `--builddir` after the subcommand (global flags must be specified before the subcommand.) Possibly for consistencies sake this means that `--project-file` should move out to (though I'm not sure to where.) I do not know if we can refactor dist-pref so that it doesn't show up in every flag settings individually.

hvr:
@ezyang is there any chance to enhance the option parser to have it accept global flags even after the subcommand? Then we could also make verbosity a global flag... (unless there's a reason not to)

ezyang:
We'd have to make more breaking API changes in this case, as GlobalFlags is not passed into many of the subcommand functions (e.g., `configure` only takes `ConfigFlags`.

ezyang:
Retiring this PR for now.

ezyang:
GREEEEEEEEEEEEEEN.

Issue #4123, opened 2016-11-20
Incompatible API change in Cabal 1.24.1.0 / PVP Violation


hvr:
It appears as if Cabal 1.24.1.0 introduced a non-backward compatible change;

I noticed this because when `cabal-install-1.24.0.0` is  compiled with `Cabal-1.24.1.0`  one  ends up with the following compile error:

```
[106 of 106] Compiling Main             ( Main.hs, /tmp/matrix-worker/1479637119/dist-newstyle/build/x86_64-linux/ghc-7.10.3/cabal-install-1.24.0.0/build/cabal/cabal-tmp/Main.o )

Main.hs:904:35:
    Couldn't match type ‘Cabal.HaddockTarget’ with ‘Bool’
    Expected type: Flag Bool
      Actual type: Flag Cabal.HaddockTarget
    In the second argument of ‘($)’, namely
      ‘haddockForHackage haddockFlags’
    In the first argument of ‘when’, namely
      ‘(fromFlagOrDefault False $ haddockForHackage haddockFlags)’

Main.hs:1106:69:
    Couldn't match expected type ‘Cabal.HaddockTarget’
                with actual type ‘Bool’
    In the first argument of ‘Flag’, namely ‘True’
    In the ‘haddockForHackage’ field of a record
```

Further investigation reveals that the culprit appears to be a0b495661853c273a10e50b91559524d9bf8aeee

which changed the type-signatures in `Distribution.Simple.BuildPaths`:

```
-hscolourPref :: FilePath -> PackageDescription -> FilePath
+hscolourPref :: HaddockTarget -> FilePath -> PackageDescription -> FilePath
```

```
-haddockPref :: FilePath -> PackageDescription -> FilePath
+haddockPref :: HaddockTarget -> FilePath -> PackageDescription -> FilePath
```

This is a serious breach of API contract, especially since `Setup.hs` scripts are still rather fragile to such kinds of API breakages.

IMO, we should undo this API breakage, and release a Cabal 1.24.1.1 (& cabal-install-1.24.0.2) ASAP, and blacklist-deprecate Cabal 1.24.1.0 

/cc @bgamari since depending on the resolution of this issue, GHC 8.0.2, may need to update to version different from Cabal 1.24.1.0

hvr:
I edited in a

    build-depends: Cabal < 1.24.1.0 || > 1.24.1.0

[...]

PR #4125, opened 2016-11-20
Cherry-pick #4094 onto 1.24.

grayjay:
I added an entry to the changelog, but I'm not sure if the version is correct.  Will there be a Cabal 1.24.1.1?  I also had to copy the test for this issue in a separate commit.


23Skidoo:
> Will there be a Cabal 1.24.1.1?

Very likely, given #4123.

grayjay:
Thanks!

PR #4131, opened 2016-11-24
Solver: Avoid removing goal choices from the tree when applying heuristics.

grayjay:
Previously, the solver applied goal-ordering heuristics by removing all
non-preferred goal choices when there was at least one preferred choice. This
left fewer goals for later steps, such as conflict counting, to work with.

This commit changes the heuristics so that they only sort the goals. I didn't
change the `preferBaseGoalChoice` heuristic, though, because it made the solver
slower in some cases. I also think that it is safe to always choose the version
of base first, because there is usually only one choice anyway.

I reversed the order of the heuristics, because sorting gives the most weight to
the last step, and pruning gives the most weight to the first step. The solver's
behavior should be unchanged with --no-count-conflicts.

/cc @kosmikus

This change is related to #3488.

I was surprised by how little the change seemed to affect the solver, once I reverted the change to `preferBaseGoalChoice`.  The biggest change that I noticed was with `--reorder-goals --no-count-conflicts`.  It that case, --reorder-goals prefers goals with 0, 1, or 2 choices (https://github.com/haskell/cabal/blob/f43c06cd451372a43e9c13a361a3bdb31656bfcd/cabal-install/Distribution/Solver/Modular/Solver.hs#L168).  Since the function `deferWeakFlagChoices` no longer prunes the flag choices, --reorder-goals moves them higher up in the tree.  Do you think it makes sense for --reorder-goals to only prefer goals with 0 or 1 choices, so that it doesn't have the effect of preferring all flags?

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/PSQ.hs
+++ b/cabal-install/Distribution/Solver/Modular/PSQ.hs
@@ -4,7 +4,6 @@ module Distribution.Solver.Modular.PSQ
     , casePSQ
     , cons
     , degree
-    , delete
     , dminimumBy
     , length
     , lookup
@@ -17,13 +16,11 @@ module Distribution.Solver.Modular.PSQ
     , map
     , mapKeys
     , mapWithKey
-    , mapWithKeyState
     , maximumBy
     , minimumBy
     , null
     , prefer
     , preferByKeys
-    , preferOrElse
     , snoc
     , sortBy
     , sortByKeys
@@ -67,15 +64,6 @@ mapKeys f (PSQ xs) = PSQ (fmap (first f) xs)
 
 mapWithKey :: (k -> a -> b) -> PSQ k a -> PSQ k b
 mapWithKey f (PSQ xs) = PSQ (fmap (\ (k, v) -> (k, f k v)) xs)
-
-mapWithKeyState :: (s -> k -> a -> (b, s)) -> PSQ k a -> s -> PSQ k b
-mapWithKeyState p (PSQ xs) s0 =
-  PSQ (F.foldr (\ (k, v) r s -> case p s k v of
-                                  (w, n) -> (k, w) : (r n))
-               (const []) xs s0)
-
-delete :: Eq k => k -> PSQ k a -> PSQ k a
-delete k (PSQ xs) = PSQ (snd (S.partition ((== k) . fst) xs))
 
 fromList :: [(k, a)] -> PSQ k a
 fromList = PSQ
@@ -145,15 +133,6 @@ prefer p (PSQ xs) =
     (pro, con) = S.partition (p . snd) xs
   in
     if S.null pro then PSQ con else PSQ pro
-
--- | Variant of 'prefer' that takes a continuation for the case
--- that there are none of the desired elements.
-preferOrElse :: (a -> Bool) -> (PSQ k a -> PSQ k a) -> PSQ k a -> PSQ k a
-preferOrElse p k (PSQ xs) =
-  let
-    (pro, con) = S.partition (p . snd) xs
-  in
-    if S.null pro then k (PSQ con) else PSQ pro
 
 -- | Variant of 'prefer' that takes a predicate on the keys
 -- rather than on the values.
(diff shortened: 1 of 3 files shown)


ezyang:
Hi @grayjay,

Thanks for your solver contributions, it is always much appreciated. When you say the change "didn't affect the solver much", what test cases did you use it on? Maybe some logs pasted here / in the commit message would be useful.

grayjay:
@ezyang I meant that I expected this change to improve performance in many cases by giving the "conflict counting" heuristic more goals to choose from. However, I wasn't able to find any cases where the solver made different choices compared with master, except when I also used --reorder-goals.  I spent a while looking, because it was hard to believe.

I searched for examples first by comparing the verbose logs from this branch and master for a few packages with many dependencies.  Then I compared runtimes with master when solving for each package on Hackage (with GHC 8.0.1), in order to find packages with very large differences in runtime.  Surprisingly, I didn't see any where the change was over 50% in either direction.  I reran ~10 with the biggest difference and found two where the difference was more than noise.  I compared their logs, but they were also unchanged.  Both were slower than master.  I profiled solving for `grapefruit-ui-gtk`, which was the slowest (18% slower than master, with a runtime of ~20 seconds).  I found that this branch spent about twice as much time in `Explore.getBestGoal`.  That makes sense, because `getBestGoal` now has more goals to choose from.  I didn't look into why the change had such a big impact on that particular package.

I also tried running the solver on `grapefruit-ui-gtk` with --reorder-goals and no backjump limit.  This branch finished in about 67 seconds, and I stopped master after ~8 minutes.

Then I compared runtime for another long-running combination of packages to test the overhead of this change when the solver makes the same choices as master.  I ran `cabal install --dry-run yesod phooey --max-backjumps -1` with GHC 8.0.1 and took the average of 4 runs.  master ran for 8.44 seconds, and this branch ran for 8.50 seconds, which is a difference of less than 1%.

Even though this change doesn't improve performance now, I think it's worthwhile for simplifying the interactions between goal-ordering heuristics, and working towards #3488.

ezyang:
Put that in the commit message! Sounds good to me :)

grayjay:
@ezyang Thanks.  I updated the commit message.

I also added a commit that changes --reorder-goals so that it only prefers goals with 0 or 1 choices.

23Skidoo:
LGTM, so merging.

@kosmikus, feel free to revert if you disagree.

grayjay:
Thanks!

kosmikus:
I certainly support this change, but I'm also surprised that @grayjay claims there aren't any substantial benefits. In the past, when I was looking at examples which were slow despite count-conflicts, it almost always seemed due to the dropping of goals that could then not be preferred later.

I will try to look at this a bit more.

grayjay:
It looks like this PR works better in combination with #4147, which was merged later.  I compared master and a branch where I started with master and then reverted this PR, using the same steps I described in #4147.  Here are the packages from Hackage that had a significant difference in runtime:

package | master (seconds) | #4131 reverted from master (seconds) | ratio
------------|----------|---------------------|------
CabalSearch |	2.34 |	6.55 |	2.80
grapefruit-examples |	3.00 |	4.46 |	1.49
grapefruit-records |	4.23 |	5.01 |	1.18
grapefruit-ui |	4.19 |	4.87 |	1.16
grapefruit-ui-gtk |	7.26 |	13.79 |	1.90
hack-handler-simpleserver |	1.97 |	3.52 |	1.79
hack-middleware-clientsession |	1.79 |	2.52 |	1.41
hamusic |	2.43 |	4.26 |	1.76
haskore |	1.57 |	4.90 |	3.13
haskore-synthesizer |	2.24 |	7.18 |	3.21
ms |	8.35 |	7.56 |	0.91
react-haskell |	17.18 |	27.68 |	1.61
shpider |	1.73 |	4.56 |	2.64

grapefruit-records, grapefruit-ui-gtk, and react-haskell hit the backjump limit, so faster may not be better.  shpider hit the backjump limit only when the PR was reverted.

I ran `cabal update` since the last time I tested this PR, so that could explain some of the difference (I should have looked at the index state.  This time it was `index-state(hackage.haskell.org) = 2016-12-03T17:22:05Z`).  To check for that possibility, I also compared the -v3 log for all four combinations of reverting #4131 and #4147 for three packages (CabalSearch, grapefruit-examples, and hack-handler-simpleserver).  In all three cases, all combinations were different, except that reverting both PRs was the same as just reverting #4147.  So, at least for those packages, #4131 had no effect without #4147.

PR #4132, opened 2016-11-25
Disable environment files due to #4010.

ezyang:
As a fix for #4010 is nowhere in sight, I'm taking the initiative to disable this feature until we have a fix.


hvr:
@dcoutts says "I'm ok with disabling it 'til we fix it"

PR #4135, opened 2016-11-26
Attempt to fix issue 4133

abooij:
It's only the last patch that does something, but it's based on @ezyang's work in order to see if it did anything.

DO NOT MERGE THIS - primarily this was the easiest way to run the tests on Windows.

### DIFF
--- a/cabal-install/Distribution/Client/Exec.hs
+++ b/cabal-install/Distribution/Client/Exec.hs
@@ -52,7 +52,7 @@ exec verbosity useSandbox comp platform programDb extraArgs =
     case extraArgs of
         (exe:args) -> do
             program <- requireProgram' verbosity useSandbox programDb exe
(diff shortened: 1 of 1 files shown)


abooij:
CC @23Skidoo

abooij:
The Windows test ran as expected (see https://ci.appveyor.com/project/23Skidoo/cabal/build/%232713%20(master) ).

So I've now based this patch on master. If the lights are green this should be mergeable.

23Skidoo:
Code looks somewhat hairy, but the complexity seems to be intrinsic, and (thankfully) confined to the Exec module.

ezyang:
I don't think it's intrinsic complexity: I think the Program API may have made some bad life choices. But it's not altogether clear what the better design is so I'm OK with a workaround for now.

PR #4147, opened 2016-11-30
Solver: Add missing call to 'simplifyVar'.

grayjay:
`simplifyVar` removes flag names from the `Var` type, so that all flags within a
package are treated as one during backjumping. A more complete fix would involve
creating a `SimpleVar` type.

The bug caused the conflict counting heuristic to never prefer flag goals. Flag
variables in the tree's goals had the flags' original names, and the flag
variables in the conflict map did not have names, so they could never be equal.

Since this fix changes the goal order, I wanted to test for an unexpected large
negative impact on solver runtime. I ran the solver on all packages on Hackage
individually with GHC 8.0.1 and looked for differences of more than 10% between
master and the branch. There were twelve packages. I reran those packages three
times and found ten with a significant difference in runtime. Here are the
average runtimes. None of them hit the backjump limit.

package               |          master (seconds)  | branch (seconds) |  ratio
-----------------------|--------------------------------|------------------------|-----
clash-ghc              |          2.60     |         3.99       |        1.54
hack-middleware-clientsession  |  8.22      |        2.54       |        0.31
hackage-server          |         1.46      |        1.85      |         1.26
hamusic                 |         5.47        |      4.55         |      0.83
haskore-synthesizer      |       10.13       |       7.64        |       0.75
language-gcl             |        2.54       |       2.03       |        0.80
ms               |               36.98       |       8.02        |       0.22
pipes-cereal-plus      |          1.51       |       1.66      |         1.10
thorn                |            8.28       |       3.08         |      0.37
wai-handler-devel        |        1.72      |       1.91    |           1.11

I looked at the diff in the -v3 log for several of them and saw that the solver
was making some flag choices earlier, as expected.

This isn't much of an improvement, but it at least looks like a safe change.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Explore.hs
+++ b/cabal-install/Distribution/Solver/Modular/Explore.hs
@@ -74,7 +74,7 @@ getBestGoal :: ConflictMap -> P.PSQ (Goal QPN) a -> (Goal QPN, a)
 getBestGoal cm =
   P.maximumBy
     ( flip (M.findWithDefault 0) cm
-    . (\ (Goal v _) -> v)
+    . (\ (Goal v _) -> simplifyVar v)
     )
 
 getFirstGoal :: P.PSQ (Goal QPN) a -> (Goal QPN, a)


23Skidoo:
Looks good to me.

/cc @kosmikus

fmthoma:
The change looks sensible to me.

However, I think the problem is a mismacht in the public APIs between `ConflictSet` and `ConflictMap`. For the `ConflictSet`, the public API works with bare `Var`s, but internally with simplified `Var`s. It looks to me like a more thorough solution would be to do a similar thing for the `ConflictMap`, so that external calls to `simplifyVar` should not even be necessary.

grayjay:
Thanks, @23Skidoo and @fmthoma .

> However, I think the problem is a mismatch in the public APIs between ConflictSet and ConflictMap. For the ConflictSet, the public API works with bare Vars, but internally with simplified Vars. It looks to me like a more thorough solution would be to do a similar thing for the ConflictMap, so that external calls to simplifyVar should not even be necessary.

Is this in comparison to my PR or my suggestion that we create a SimpleVar type?  I suggested a SimpleVar type because I think it would be more type safe when we have functions that can get variables out of a ConflictSet or ConflictMap.  For example, ConflictSet has a `toList` function with type `ConflictSet qpn -> [Var qpn]`.  I think it would be better to indicate that the variables in the list no longer have flag names.  We could create a ConflictMap type that contains SimpleVars.

I wanted to leave the bigger change for later, since it would conflict with other PRs.

23Skidoo:
> I wanted to leave the bigger change for later, since it would conflict with other PRs.

Sure, that's fine. Let's merge this PR right now then.

23Skidoo:
Merged, thanks!

fmthoma:
@grayjay 
> Is this in comparison to my PR or my suggestion that we create a SimpleVar type? I suggested a SimpleVar type because I think it would be more type safe when we have functions that can get variables out of a ConflictSet or ConflictMap.

This is in comparison to your PR. Creating a `SimpleVar` type would make this problem more obvious (which is why I am in favor of your suggestion), since `ConflictSet` and `ConflictMap` would work with `SimpleVar`s internally and with `Var`s on the outside.

kosmikus:
Thanks for this. This seems like it's definitely the right thing to do.

grayjay:
This change was reverted in #4153.

grayjay:
This PR isn't needed after #4562.

PR #4149, opened 2016-12-02
Nix integration

ttuegel:
These commits add Nix integration to `cabal-install`. If the global `--enable-nix` option is specified and a `shell.nix` file exists, `cabal` will run the affected commands in a `nix-shell` instance.

### DIFF
--- a/Cabal/doc/nix-integration.rst
+++ b/Cabal/doc/nix-integration.rst
@@ -1,0 +1,40 @@
+Nix Integration
+===============
+
+`Nix <http://nixos.org/nix/>`_ is a package manager popular with some Haskell developers due to its focus on reliability and reproducibility. ``cabal`` now has the ability to integrate with Nix for dependency management during local package development.
+
+Enabling Nix Integration
+------------------------
+
+To enable Nix integration, simply pass the ``--enable-nix`` global option when you call ``cabal``. To use this option everywhere, edit your ``$HOME/.cabal/config`` file to include:
+
+.. code-block:: cabal
+
+    nix: True
+
+If the package (which must be locally unpacked) provides a ``shell.nix`` file, this flag will cause ``cabal`` to run most commands through ``nix-shell``. The following commands are affected:
+
+- ``cabal configure``
+- ``cabal build``
+- ``cabal repl``
+- ``cabal install`` (only if installing into a sandbox)
+- ``cabal haddock``
+- ``cabal freeze``
+- ``cabal gen-bounds``
+- ``cabal run``
+
+If the package does not provide a ``shell.nix``, ``cabal`` runs normally.
+
+Creating Nix Expressions
+------------------------
+
+The Nix package manager is based on a lazy, pure, functional programming language; packages are defined by expressions in this language. The fastest way to create a Nix expression for a Cabal package is with the `cabal2nix <https://github.com/NixOS/cabal2nix>`_ tool. To create a ``shell.nix`` expression for the package in the current directory, run this command:
+
+.. code-block:: console
+
+    $ cabal2nix --shell ./. >shell.nix
+
+Further Reading
+----------------
+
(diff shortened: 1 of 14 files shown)


23Skidoo:
Nice!

Needs a note in the changelog, though. Will `new-build`be supported as well?

k0001:
From `nix-shell`'s man page:

> `nix-shell` defaults to `shell.nix` if it exists, and `default.nix` otherwise.

Shouldn't `cabal` try and do the same so that the behavior of `cabal` is more predictable to `nix-shell` users?

ttuegel:
> Also, would it be possible to set up some kind of CI for this stuff?

@23Skidoo Maybe? I'm not sure what you have in mind, could you elaborate?

ttuegel:
> Shouldn't cabal try and do the same so that the behavior of cabal is more predictable to nix-shell users?

@k0001 Well, I thought about that. The fact that `nix-shell` respects `default.nix` is mostly a historical accident. Once there were no `shell.nix` files, but more and more packages need different expressions for building versus developing. This distinction is usually represented by having different `shell.nix` and `default.nix` files. In fact, the need to distinguish was mostly driven by Haskell packages! I imagine most Nix users will set `nix: True` in their `~/.cabal/config` and I don't want them to get into trouble if a package doesn't provide `shell.nix` and its `default.nix` isn't suitable for `nix-shell`. But as you pointed out, Nix users may also be confused when they encounter a `default.nix` that *is* suitable for `nix-shell`, but gets ignored. What do you think?

Cabal/doc/nix-integration.rst
```diff
@@ -0,0 +1,35 @@
...
+- ``cabal freeze``
+- ``cabal gen-bounds``
+- ``cabal run``
+
+If the package does not provide a ``shell.nix``, ``cabal`` runs normally.
+
+Creating Nix Expressions
+------------------------
+
+The Nix package manager uses a functional language to describe packages, known as expressions. The fastest way to create a Nix expression for a Cabal package is with the ``cabal2nix`` tool. To create a ``shell.nix`` expression for the package in the current directory, run this command:
```
ttuegel: @grayjay That's not quite what I meant. Does the updated version read better?

grayjay: Yes, I think that's clearer.

k0001:
I find it useful to define the behavior both for nix-shell and nix-build in
a single default.nix file. But  anyway, seeing as this is the default
behavior for nix-shell, I think supporting it in cabal as well would lead
to more predictable results. At worst, if your default.nix is not nix-shell
aware, you'll get some kind of error in the command line right away.

ttuegel:
> At worst, if your default.nix is not nix-shell aware, you'll get some kind of error in the command line right away.

I can live with that!

ttuegel:
@23Skidoo I'm getting timeouts on Travis. What is the etiquette for merging in that case?

ezyang:
It doesn't look like you made any changes that should have made the build go slower; the OS X machines are just inexplicably 20-50% slower. Unfortunately, I don't see any reports on Travis CI.  I went ahead and restarted the builds in hopes that it was just a temporary slowdown. If not, we'll have to play some games to get it down to acceptable level (maybe finally make cabal-install-lib so we don't have to build the cabal-install libraries three times to test them!)

ezyang:
Fixed!

cstrahan:
Hey, this looks pretty great!

One thing I'm curious about: I'm almost always in a Nix shell, so I imagine I'd have to either open a fresh shell or invoke cabal as `env -u IN_NIX_SHELL cabal ...`. Assuming my reasoning is correct, might it make more sense to use some other heuristic - perhaps an env var specifically for this purpose (e.g. `CABAL_IN_NIX_SHELL`)?

ttuegel:
> perhaps an env var specifically for this purpose (e.g. CABAL_IN_NIX_SHELL)?

That's a good idea. `cabal` should not care if it is initially executed inside a Nix shell; we only want to distinguish between "inside a shell started by cabal" and not.

PR #4151, opened 2016-12-03
Migrate ConflictSet qpn ↦ ConflictSet

fmthoma:
This is a cleanup after #3960.

The parameter in `ConflictSet qpn` is only ever used with `QPN`, so the parametricity does not really make sense. Also, with #3960 the `ConflictMap` is exposed a bit more, and although `ConflictSet` and `ConflictMap` are very similar in their usage, one is parameterized over `qpn` and the other is not.


kosmikus:
Yes, looks good to me as well.

23Skidoo:
Merged, thanks!

PR #4152, opened 2016-12-04
Solver: Add a 'SimpleVar' type for representing variables in conflict sets.

grayjay:
The solver currently treats all flags in a package as the same variable during
backjumping, in order to simplify the backjumping logic. It does that by
removing the flag names from variables when they are added to conflict sets.

The 'SimpleVar' type should prevent bugs related to mixing up variables with
flag names and variables without flag names.

I also removed a redundant call to 'simplifyVar'.

See #4147.
/cc @kosmikus @fmthoma

PR #4153, opened 2016-12-05
Revert "Solver: Add missing call to 'simplifyVar'."

grayjay:
This reverts commit adc1fe96e06a652f798c7f1ed62b60b9809df383.

See #4152.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Explore.hs
+++ b/cabal-install/Distribution/Solver/Modular/Explore.hs
@@ -72,7 +72,7 @@ getBestGoal :: ConflictMap -> P.PSQ (Goal QPN) a -> (Goal QPN, a)
 getBestGoal cm =
   P.maximumBy
     ( flip (M.findWithDefault 0) cm
-    . (\ (Goal v _) -> simplifyVar v)
+    . (\ (Goal v _) -> v)
     )
 
 getFirstGoal :: P.PSQ (Goal QPN) a -> (Goal QPN, a)


fmthoma:
Should be safe to merge since it just reverts to old behavior.

Can you give a short description of the bug this introduced?

ezyang:
The bug seems to be tracked at #4152.

grayjay:
It looks like the solver doesn't add enough variables to the conflict set when there is a test suite dependency under a flag, and the solver chooses the flag before it chooses to enable tests.  #4131 and #4147 probably exposed the bug because they let the solver choose flags earlier.  I wanted to revert my PR first, since I'm not sure when I'll have time to look into this further.  It's probably not an easy fix.

fmthoma:
Could you compare the behavior under your test case when changing the `updateCM initial cm` to `updateCM cs cm` or even `cm` in [Explore.hs#L64](https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Solver/Modular/Explore.hs#L64)? (This was changed from `cm` to `updateCM initial cm` in #3961.)

grayjay:
@fmthoma I'm not sure I understand.  Do you think #3961 is related?  I don't think it caused the quickcheck failure.  I reran the tests on this PR, this time with `--quickcheck-tests 1000000`, and didn't get any failures.  However, I can reproduce the failure on master fairly easily.

I also think that #3961 couldn't have introduced the bug because, before #4147, --count-conflicts couldn't prefer flags, and the bug seems to only occur when flags are chosen earlier.

fmthoma:
At least I have the suspicion that it #3961 may influence the behavior. OTOH, the conflict counts should not influence correctness of the result (am I correct?), only the order in which the tree is traversed, so even if #3961 has some influence, it probably is not the cause.

But let's merge this to fix the bug for now.

grayjay:
> OTOH, the conflict counts should not influence correctness of the result (am I correct?), only the order in which the tree is traversed, so even if #3961 has some influence, it probably is not the cause.

Yes, the solver should work correctly with any goal order.  After the bug is fixed, it might be worth adding a quickcheck test that runs the solver with a random goal order.

PR #4160, opened 2016-12-07
Update 'bootstrap.sh' and the upper bound on 'time'.

23Skidoo:
Same as #4159, but for 1.24.


23Skidoo:
@bgamari This is the final fix that I want to apply before cutting 1.24.2.

PR #4164, opened 2016-12-07
First create then rename symlinks to fix #4156

### LINKED ISSUE #4156: Installing foreign libraries with versions twice fails
First time:
```
$ cabal install
Resolving dependencies...
Notice: installing into a sandbox located at /tmp/ForeignLibs/.cabal-sandbox
Configuring my-foreign-lib-0.1.0.0...
Building my-foreign-lib-0.1.0.0...
Installed my-foreign-lib-0.1.0.0
```

Second time:
```
$ cabal install
Resolving dependencies...
my-foreign-lib-0.1.0.0 (reinstall)
Warning: Note that reinstalls are always dangerous. Continuing anyway...
Notice: installing into a sandbox located at /tmp/ForeignLibs/.cabal-sandbox
Configuring my-foreign-lib-0.1.0.0...
Building my-foreign-lib-0.1.0.0...
Failed to install my-foreign-lib-0.1.0.0
Build log ( /tmp/ForeignLibs/.cabal-sandbox/logs/ghc-8.0.1/my-foreign-lib-0.1.0.0-1SzQjJT6J8GK3tvm2Ul0XO.log ):
cabal: Entering directory '.'
Configuring my-foreign-lib-0.1.0.0...
Preprocessing library my-foreign-lib-0.1.0.0...
Preprocessing foreign library 'myforeignlib' for my-foreign-lib-0.1.0.0...
Linking dist/dist-sandbox-20a9337b/build/myforeignlib/libmyforeignlib.so ...
Preprocessing foreign library 'versionedlib' for my-foreign-lib-0.1.0.0...
Linking dist/dist-sandbox-20a9337b/build/versionedlib/libversionedlib.so.5 ...
Installing library in
/tmp/ForeignLibs/.cabal-sandbox/lib/x86_64-linux-ghc-8.0.1/my-foreign-lib-0.1.0.0-1SzQjJT6J8GK3tvm2Ul0XO
Installing foreign library myforeignlib in /tmp/ForeignLibs/.cabal-sandbox/lib
Installing foreign library versionedlib in /tmp/ForeignLibs/.cabal-sandbox/lib
cabal: libversionedlib.so.5.4.3: already exists
cabal: Leaving directory '.'
my-foreign-lib-0.1.0.0-1SzQjJT6J8GK3tvm2Ul0XO failed during the final install
step. The exception was:
ExitFailure 1
```

I'm pretty sure the problem is that in 8744e30, in `Cabal/Distribution/Simple/GHC.hs`, line 1526 and 1527, the `createSymbolicLink` fails if the link already exists.

What is an OK thing to do here? First remove a file if it exists, then create what we want? I don't think there is a variant of `createSymbolicLink` that overwrites existing files. Should I first create the link under a different name, and then rename it?

abooij:


ezyang:
Thanks. Can we test? Also, wouldn't it be better to create a temporary directory, make the symlink there, and then rename it in?

abooij:
> Can we test? 

Yes, by attempting to install twice. Is that a desirable test case? Where'd you want to place it? I think there are many many features prone to this sort of bug. Basically, all cabal commands should be idempotent.

> Also, wouldn't it be better to create a temporary directory, make the symlink there, and then rename it in?

What problem does that solve?

ezyang:
> Yes, by attempting to install twice. Is that a desirable test case? Where'd you want to place it? I think there are many many features prone to this sort of bug. Basically, all cabal commands should be idempotent.

Yes, that sounds good. Probably you can just adjust the existing test for foreign libs to install twice?

> What problem does that solve?

If an async exception strikes between creating the temporary symlink and renaming, it will get left behind. If a symlink of that name already lives in the install directory, this op will fail.

abooij:
> If an async exception strikes between creating the temporary symlink and renaming, it will get left behind. If a symlink of that name already lives in the install directory, this op will fail.

That is still a problem when we create the symlink in a subdirectory, no?

ezyang:
If it's a temporary directory as per mkdtemp() (I think we have a lib function for doing this) then (1) the directory will be guaranteed to be empty, and (2) it will automatically get removed upon failure.

abooij:
I implemented your suggestions.

ezyang:
Thanks, looks good! Note that setup_install [] goes ahead and reconfigures/rebuilds before reinstalling, so if you want to make it slightly faster, I imagine just calling setup "copy" [] again at the end would also be sufficient to tickle the bug.

abooij:
I can't get the OSX tests to pass. The jobs have timed out about 5 times now. I'm giving up. Feel free to merge this.

23Skidoo:
CI is green now.

23Skidoo:
Merged, thanks!

PR #4171, opened 2016-12-17
adding ghc-prof-options to fprof warning

### LINKED ISSUE #3718: Confusing or wrong warning about profiling option -auto-all for executables
I have a package containing executables with the field:

```
  GHC-Prof-Options: -auto-all
```

`cabal check` warns me:

```
* 'ghc-options: -fprof*' profiling flags are typically not appropriate for a
distributed library package. These flags are useful to profile this package,
but when profiling other packages that use this one these flags clutter the
profile output with excessive detail. If you think other packages really want
to see cost centres from this package then use '-fprof-auto-exported' which
puts cost centres only on exported functions. Alternatively, if you want to
use this, make it conditional based on a Cabal configuration flag (with
'manual: True' and 'default: False') and enable that flag during development.
```

I was confused because it talks about GHC-Options, although the problem occurs in a GHC-Prof-Options and it does not mention a line number. It complains about `-fprof*` type flags although I have no flag of this form. I guess that `-auto-all` is an alias for an `-fprof` flag. Then the message talks about other packages that are bothered by my flags. However, I added the flag only to executables. So no other package is affected.
Maybe the warning is still valid, but at least it confuses me.

accraze:
This PR contains an update to cabal check:

Since `-auto-all` is an alias for `-fprof` flag, then the `PackageDistSuspicious` warning should mention `ghc-prof-options` as well.

fixes #3718

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1539,7 +1539,7 @@ checkDevelopmentOnlyFlagsBuildInfo bi =
                "-fprof-cafs", "-fno-prof-count-entries",
                "-auto-all", "-auto", "-caf-all"] $
       PackageDistSuspicious $
-           "'ghc-options: -fprof*' profiling flags are typically not "
+           "'ghc-options/ghc-prof-options: -fprof*' profiling flags are typically not "
         ++ "appropriate for a distributed library package. These flags are "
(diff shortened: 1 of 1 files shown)


ezyang:
LGTM. CI failing is not your fault.

accraze:
cool, thanks @ezyang and @23Skidoo!

PR #4172, opened 2016-12-18
Unsilence curl to figure out why https download is failing

ezyang:


gbaz:
We just rolled over the haskell.org certs -- ppl seem to think they check out, but there may be some missing intermediate keys or something?

ezyang:
Yes it's missing intermediate keys. I posted to ghc-devs, see  https://www.sslshopper.com/ssl-checker.html#hostname=www.haskell.org

gbaz:
i guess we really do need to fix that bundling? cc @relrod @thoughtpolice

Issue #4154, opened 2016-12-05
new-build dependency solver failure with GHC 7.0


hsenag:
I'm running

    The Glorious Glasgow Haskell Compilation System, version 7.0.4
    cabal-install version 1.24.0.1
    compiled using version 1.24.1.0 of the Cabal library 

This command-line works fine with a clean sandbox:

    cabal install HTTP-4000.3.3 --constraint 'time==1.1.2.3' --constraint 'entropy<0.2.2.4' \
            -f-warp-tests --constraint 'HUnit<1.4' --enable-tests

If I unpack HTTP-4000.3.3 and run this command-line in the unpacked folder, it fails in the dependency solver after several minutes:

    cabal new-build --constraint 'time==1.1.2.3' --constraint 'entropy<0.2.2.4' \
            -f-warp-tests --constraint 'HUnit<1.4' --enable-tests --max-backjumps=-1

with

    Resolving dependencies...
    cabal: Could not resolve dependencies:
    trying: time-setup.Cabal~>entropy-setup.Cabal-1.20.0.4 (dependency of time-1.1.2.3)
    trying: time-setup.time~>entropy-setup.time-1.1.2.3 (dependency of time-1.1.2.3)
    Dependency tree exhaustively searched.

I'm not sure how to interpret the error message - the first line suggests it's trying to link the `Cabal` versions used for `time` and `entropy`'s setup programs, but I'm not sure what the second line is about at all.

ezyang:
One thing that pops out to me: GHC 7.0.4 is really old. When I run your repro steps with cabal-install HEAD, I get this:

```
ezyang@sabre:~/Dev/labs/T4154/HTTP-4000.3.3$ cabal new-build --constraint 'time==1.1.2.3' --constraint 'entropy<0.2.2.4'         -f-warp-tests --constraint 'HUnit<1.4' --enable-tests --max-backjumps=-1 -w ghc-7.0.4
Resolving dependencies...
cabal: Could not resolve dependencies:
trying: HTTP-4000.3.3 (user goal)
trying: random-1.0.0.3/installed-751... (dependency of test-framework-0.8.1.1)
next goal: time (dependency of HTTP-4000.3.3)
rejecting: time-1.2.0.3/installed-349..., time-1.7, time-1.6.0.1, time-1.6,
time-1.5.0.1, time-1.5, time-1.4.2, time-1.4.1, time-1.4.0.2, time-1.4.0.1,
[...]

PR #4176, opened 2016-12-19
Solver: Check for cycles after every step.

grayjay:
Previously, the solver only checked for cycles after it had already found a
solution. That reduced the number of times that it performed the check in the
common case where there were no cycles. However, when there was a cycle, the
solver could spend a lot of time searching subtrees that already had a cyclic
dependency and therefore could not lead to a solution. This is part of
#3824.

Changes in this commit:
- Store the reverse dependency map on all choice nodes in the search tree, so
  that 'detectCyclesPhase' can access it at every step.
- Check for cycles incrementally at every step. Any new cycle must contain the
  current package, so we just check whether the current package is reachable
  from its neighbors.
- If there is a cycle, we convert the map to a graph and find a strongly
  connected component, as before.
- Instead of using the whole strongly connected component as the conflict set,
  we select one cycle. Smaller conflict sets are better for backjumping.
- The incremental cycle detection automatically fixes a bug where the solver
  filtered out the message about cyclic dependencies when it summarized the full
  log. The bug occurred when the failure message was not immediately after the
  line where the solver chose one of the packages involved in the conflict. See
  #4154.

I tried several approaches and compared performance when solving for
packages with different numbers of dependencies. Here are the results. None of
these runs involved any cycles, so they should have only tested the overhead of
cycle checking. I turned off assertions when building cabal.

Index state: index-state(hackage.haskell.org) = 2016-12-03T17:22:05Z
GHC 8.0.1

Runtime in seconds:

Packages        |            Search tree depth  | Trials |  master  | This PR |  #1 |     #2
------------------|----------------------------------|---------|------------|------------|------|-------
yesod            |           343     |            3  |      2.00  |   2.00  |    2.13 |   2.02
yesod gi-glib leksah     |   744  |               3    |    3.21  |   3.31   |   4.10  |  3.48
phooey              |        66      |            3  |      3.48  |   3.54 |     3.56  |  3.57
Stackage nightly snapshot  | 6791   |             1    |    186   |   193  |     357  |   191

Total memory usage in MB, with '+RTS -s':

Packages                   |                     Trials |  master  |  This PR  | #1  |   #2
----------------------------|------------------------|------------|-------------|------|--------
yesod                                     |      1    |     189  |     188    |   188   |  198
yesod gi-glib leksah                |            1    |     257    |   257    |   263   |  306
Stackage nightly snapshot       |                1    |    1288   |   1338   |   1432 |  12699

#1 - Same as master, but with cycle checking (Data.Graph.stronglyConnComp) after
     every step.
#2 - Store dependencies in Distribution.Compat.Graph in the search tree, and
     check for cycles containing the current package at every step.

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -70,7 +70,7 @@ extendOpen qpn' gs s@(BS { rdeps = gs', open = o' }) = go gs' o' gs
     go g o (ng@(OpenGoal (Simple (Dep _ qpn _) c) _gr) : ngs)
       | qpn == qpn'       = go                            g               o  ngs
           -- we ignore self-dependencies at this point; TODO: more care may be needed
-      | qpn `M.member` g  = go (M.adjust ((c, qpn'):) qpn g)              o  ngs
+      | qpn `M.member` g  = go (M.adjust (addIfAbsent (c, qpn')) qpn g)   o  ngs
       | otherwise         = go (M.insert qpn [(c, qpn')]  g) (cons' ng () o) ngs
           -- code above is correct; insert/adjust have different arg order
     go g o (   (OpenGoal (Simple (Ext _ext ) _) _gr) : ngs) = go g o ngs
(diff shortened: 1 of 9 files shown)


ezyang:
LGTM, thank you!

grayjay:
Thanks!

PR #4185, opened 2016-12-21
Mark `template-haskell` non-upgradable (again)

### LINKED ISSUE #4098: Consider making `template-haskell` non-reinstallable again
This is a reminder ticket to investigate under which conditions it is allowed to reinstall `template-haskell`, and consequently decide whether we should disallow it sometimes or always.

hvr:
65e9b88bc849e76040ed7947591ea7172cd02db5 made template-haskell non-upgradable

65ae95c191b19ec1884a4d80afbdaba9c3d477d1 made it upgradable again

...and this commit makes it non-upgradable again

While there may be a few cases when `template-haskell` can be
reinstalled, it's generally not advisable, and causes more harm than the
benefit in the few cases where we'd gain install-plan
flexibility (e.g. when only template-haskell's API is needed for
compile, but no actual TH evaluation takes place).

Closes #4098

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -348,8 +348,14 @@ dontUpgradeNonUpgradeablePackages params =
         (PackageConstraintInstalled pkgname)
         ConstraintSourceNonUpgradeablePackage
       | Set.notMember (mkPackageName "base") (depResolverTargets params)
-      , pkgname <- map mkPackageName [ "base", "ghc-prim", "integer-gmp"
-                                     , "integer-simple" ]
+      -- If you change this enumeration, make sure to update the list in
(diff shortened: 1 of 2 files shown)


23Skidoo:
OK. Do we want this backported to 1.24?

hvr:
@23Skidoo probably; there is a slight risk that it may outlaw working install-plans (e.g. those which don't actually evaluate TH code using the reinstalled template-haskell lib), but it would reduce the risk of users being confused about the weird error messages they get when this causes errors due to GHC being confused.

So I'd suggest to backport this to the next 1.24 point rls of cabal-install

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 1.24 (141effc6c413ac9ef9f1c7e150eb61916efccf8c).

PR #4193, opened 2016-12-30
Nix integration improvements

ttuegel:
The Nix expressions for Cabal and cabal-install are updated. Nix expressions are instantiated in a "fake" Nix shell to cope with environment impurities introduced by newer versions of cabal2nix. cabal looks at the `CABAL_IN_NIX_SHELL` environment variable instead of `IN_NIX_SHELL`; it was pointed out that cabal need not care if it is running inside `nix-shell` because we only care if we are running in a shell created by cabal.

### DIFF
--- a/Cabal/default.nix
+++ b/Cabal/default.nix
@@ -1,20 +1,18 @@
 { mkDerivation, array, base, binary, bytestring, containers
-, deepseq, directory, exceptions, filepath, old-time, pretty
-, process, QuickCheck, regex-posix, stdenv, tagged, tasty
-, tasty-hunit, tasty-quickcheck, time, transformers, unix
+, deepseq, directory, filepath, pretty, process, QuickCheck, stdenv
+, tagged, tasty, tasty-hunit, tasty-quickcheck, time, unix
 }:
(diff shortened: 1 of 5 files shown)


laura-dietz:
Does this not require some changes in documentation?

ttuegel:
It does, but I need it to build first! :smiley: For some reason Travis won't run on my personal branch anymore...

grayjay:
> For some reason Travis won't run on my personal branch anymore...

I think it is the branch whitelist: https://github.com/haskell/cabal/blob/19d9738ff6e52ec033c1d76f4d9e6f180f50face/.travis.yml#L7-L15

23Skidoo:
Branches with corresponding PRs shouldn't be affected by the whitelist...

ttuegel:
Travis does run on the PR, it just doesn't run on my personal clone anymore (unless I open a PR). That's why I opened this PR before it was really ready. It would be nice if the Travis web interface would indicate that the whitelist is blocking a branch... The branch still appeared in the web interface, so I was very confused why Travis would not build it.

Issue #2735, opened 2015-07-26
Monoid instance for CondTree


hvr:
When experimenting with the Cabal solver, I've defined the following orphan `Monoid` instance:

```haskell
instance (Monoid c, Monoid a) => Monoid (CondTree v c a) where
    mempty = CondNode mempty mempty mempty

    mappend cn1 cn2 =
        CondNode { condTreeData        = mappend (condTreeData        cn1) (condTreeData        cn2)
                 , condTreeConstraints = mappend (condTreeConstraints cn1) (condTreeConstraints cn2)
                 , condTreeComponents  = mappend (condTreeComponents  cn1) (condTreeComponents  cn2)
                 }

    mconcat cns =
        CondNode { condTreeData        = mconcat (map condTreeData        cns)
                 , condTreeConstraints = mconcat (map condTreeConstraints cns)
                 , condTreeComponents  = mconcat (map condTreeComponents  cns)
                 }
```

Does this instance make sense to be added to `Cabal`?

/cc @kosmikus

phadej:
This make sense as `c` is often (always?) `[Dependency]` which we treat as a conjunction of dependencies.

ezyang:
Yes, this seems correct.

PR #4201, opened 2017-01-02
Split Condition/CondTree out of GenericPackageDescription.

ezyang:

### DIFF
--- a/Cabal/Distribution/Types/CondTree.hs
+++ b/Cabal/Distribution/Types/CondTree.hs
@@ -9,6 +9,13 @@ module Distribution.Types.CondTree (
     CondBranch(..),
     condIfThen,
     condIfThenElse,
+    mapCondTree,
+    mapTreeConstrs,
+    mapTreeConds,
+    mapTreeData,
+    extractCondition,
+    simplifyCondTree,
+    ignoreConditions,
 ) where
 
 import Prelude ()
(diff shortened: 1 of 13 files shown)


hvr:
Btw, since you're looking at this part of the code: any opinion on #2735?

Cabal/Distribution/Types/CondTree.hs
```diff
@@ -0,0 +1,25 @@
...
+module Distribution.Types.CondTree (
+    CondTree(..),
+) where
+
+import Prelude ()
+import Distribution.Compat.Prelude
+
+import Distribution.Types.Condition
+
+data CondTree v c a = CondNode
```
23Skidoo: Can you please add a Haddock comment here while you're at it?

cabal-install/Distribution/Client/JobControl.hs
```diff
@@ -1,3 +1,4 @@
+--
```
23Skidoo: Committed by mistake?

Cabal/Distribution/Types/CondTree.hs
```diff
@@ -16,10 +19,30 @@ import Distribution.Types.Condition
...
-    , condTreeComponents  :: [( Condition v
-                              , CondTree v c a
-                              , Maybe (CondTree v c a))]
+    , condTreeComponents  :: [CondBranch v c a]
     }
     deriving (Show, Eq, Typeable, Data, Generic, Functor, Foldable, Traversable)
 
 instance (Binary v, Binary c, Binary a) => Binary (CondTree v c a)
+
+data CondBranch v c a = CondBranch
```
23Skidoo: Would be nice with a Haddock comment here as well.

23Skidoo:
LGTM modulo minor comments.

cabal-install/Distribution/Client/JobControl.hs
ezyang: Yep.

PR #4205, opened 2017-01-03
Track source files of dependencies so we rebuild as necessary.

### LINKED ISSUE #4202: cabal new-build does not rebuild package if dep changed on separate run
Steps to reproduce:
1. Create a project with two packages `p` and `q`, `q` depends on `p`
2. `cabal new-build p q`
3. Modify the first package
4. `cabal new-build p`
5. `cabal new-build q`

Expected behavior: q rebuilds on step 5
Actual behavior: cabal reports "Up to date"

The problem seems to be the logic in `checkPackageFileMonitorChanged`:

```
      MonitorUnchanged () _
          -- The configChanged here includes the identity of the dependencies,
          -- so depsBuildStatus is just needed for the changes in the content
          -- of dependencies.
        | any buildStatusRequiresBuild depsBuildStatus -> do
            regChanged <- checkFileMonitorChanged pkgFileMonitorReg srcdir ()
            let mreg = changedToMaybe regChanged
            return (Left (BuildStatusBuild mreg BuildReasonDepsRebuilt))
```

This tests if any dependencies changed. But this information is populated on a per-run basis. If the dependency finished building successfully in a previous run, it will not have reported as requiring build, and then we'll decide nothing needs building.

Correct fix would be to somehow monitor the dep's outputs (now entirely sure what the correct file to monitor is) so that the change persists across runs.

CC @dcoutts

ezyang:
The fix is a bit of a hack but it seems to work.

Fixes #4202.


hvr:
@ezyang does this by any chance fix https://github.com/haskell/cabal/issues/3179#issuecomment-250902556 ?

DanielG:
This seems to be to aggressive, with this commit `cabal new-build` always seems shell out to `ghc --make` regardless if something changed or not. At least I don't think that's intended?

ezyang:
@DanielG Nope, that's too aggressive. Back to the drawing board!

ezyang:
@DanielG I tried running with this commit on a simple project and I did not see a spurious recompile. What project were you testing on; Cabal?

ezyang:
It looks like the problem is that we're adding deps on monitor files even when they don't exist, in the case of Custom packages. Shouldn't be hard to fix.

DanielG:
@ezyang yes I was building cabal in the source repo.

ezyang:
(CC @DanielG, in case you'd like to test it out.)

PR #4207, opened 2017-01-05
New 'cabal-install' command: 'outdated'

### LINKED ISSUE #2139: outdated subcommand like as cab's
I think `cab outdated` command is very usefull.
Wouldn't you support this feature?

23Skidoo:
Fixes #2139.

~~WIP (so don't merge), items still left to implement:~~

  * [x] docs
  * [x] tests
  * [x] changelog note


ezyang:
Does this work with new-build freeze files?

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -0,0 +1,134 @@
...
+    Left _        -> die "finalizePD failed"
+    Right (pd, _) -> do
+      let bd = buildDepends pd
+      debug verbosity
+        "Reading the list of dependencies from the package description"
+      return bd
+
+-- | Find all outdated dependencies.
+listOutdated :: [Dependency] -> PackageIndex UnresolvedSourcePackage
+              -> S.Set PackageName -> S.Set PackageName
```
ezyang: What do these two sets do.

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -0,0 +1,134 @@
...
+  return deps
+
+-- | Read the list of dependencies from the package description.
+depsFromPkgDesc :: Verbosity -> Compiler  -> Platform -> IO [Dependency]
+depsFromPkgDesc verbosity comp platform = do
+  cwd  <- getCurrentDirectory
+  path <- tryFindPackageDesc cwd
+  gpd  <- readPackageDescription verbosity path
+  let cinfo = compilerInfo comp
+      epd = finalizePD [] defaultComponentRequestedSpec
```
ezyang: Use of defaultComponentRequestedSpec means that you will never check bounds on tests and benchmarks.

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -0,0 +1,134 @@
...
+  let outdatedDeps = listOutdated deps pkgIndex ignore minor
+  notice verbosity ("Outdated dependencies: "
+                     ++ intercalate ", "
+                        (map (\(d, v) -> display d
+                               ++ " (latest: " ++ display v ++ ")") outdatedDeps))
+  if (exitCode && (not . null $ outdatedDeps))
+    then exitFailure
+    else return ()
+
+-- | Read the list of dependencies from the freeze file.
```
ezyang: So, I'm wondering, what if a hard-coded dep in the freeze file is out of date, but actually it's the latest version of the package which actually has a consistent build plan. Will we report outdated in this case?

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -0,0 +1,134 @@
...
+depsFromFreezeFile :: Verbosity -> IO [Dependency]
+depsFromFreezeFile verbosity = do
+  cwd        <- getCurrentDirectory
+  userConfig <- loadUserConfig verbosity cwd Nothing
+  let ucnstrs = map fst . configExConstraints . savedConfigureExFlags $ userConfig
+      deps    = mapMaybe (packageConstraintToDependency . userToPackageConstraint)
+                ucnstrs
```
ezyang: I can't tell if I should expect these deps to look like foo == 0.3.1.0, bar == 0.2.1.0 or something more relaxed. Comment would help.

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -0,0 +1,134 @@
...
+  where
+    isOutdated :: Dependency -> Maybe (Dependency, Version)
+    isOutdated dep
+      | depPkgName dep `S.member` ignore = Nothing
+      | otherwise                      =
+          let this   = map packageVersion $ lookupDependency pkgIndex dep
+              latest = lookupLatest dep
+          in (\v -> (dep, v)) `fmap` isOutdated' this latest
+
+    isOutdated' :: [Version] -> [Version] -> Maybe Version
```
ezyang: OK, I hope sortedness of lookupDependency output is actually an invariant we can rely on.

23Skidoo:
> Does this work with new-build freeze files?

Nope, but wouldn't be hard to add.

cabal-install/Distribution/Client/Outdated.hs
23Skidoo: Yes, as you can see this code doesn't run the solver so it has no way of seeing that. This is a potential future improvement.

cabal-install/Distribution/Client/Outdated.hs
23Skidoo: This code is simply converting the list of `UserConstraint`s from the freeze file to `[Dependency]` so that we can look them up in the package index. So it doesn't rely on them being in any special form.

cabal-install/Distribution/Client/Outdated.hs
23Skidoo: A set of packages to ignore and a set of packages for which major version bumps should be ignored.

23Skidoo:
Rebased, added a changelog note.

23Skidoo:
Added tests, docs, new-style freeze file support, fixed a bug in `--minor` mode. This is now ready to go in.

cabal-install/Distribution/Client/Outdated.hs
23Skidoo: Changed `last` to `maximum` just in case.

cabal-install/Distribution/Client/Outdated.hs
23Skidoo: Changed to `ComponentRequestedSpec True True`.

Cabal/doc/developing-packages.rst
```diff
@@ -1147,6 +1147,57 @@ For example, given the following dependencies specified in
...
+configured to act on the freeze file (both old- and new-style) and
+ignore major (or all) version bumps on Hackage for a subset of
+dependencies.
+
+Example:
+
+.. code-block:: console
+
+    $ cabal outdated
+    Outdated dependencies: HTTP ==4000.3.3 (latest: 4000.3.4), HUnit ==1.3.1.1
```
phadej: Can this be on each line separately, now the word wrapping is unfortunate: *(latests: 1.5.0.0)* drops to the next line

phadej: Also an example of "nothing to update" output would be great.

Cabal/doc/developing-packages.rst
```diff
@@ -1147,6 +1147,57 @@ For example, given the following dependencies specified in
...
+    ListLike ==4.2.1 (latest: 4.5), QuickCheck ==2.8.2 (latest:
+    2.9.2), aeson ==0.11.2.0 (latest: 1.1.0.0)
+
+    $ cabal outdated --ignore=HTTP --minor=HUnit,ListLike
+    Outdated dependencies: HUnit ==1.3.1.1 (latest: 1.3.1.2),
+    QuickCheck ==2.8.2 (latest: 2.9.2), aeson ==0.11.2.0 (latest: 1.1.0.0)
+
+The following flags are supported by the ``outdated`` command:
+
+``--freeze-file``
```
phadej: Could we add `--exit-code`, from `git diff --help`:
```
       --exit-code
           Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.
           Disable all output of the program. Implies --exit-code.
```

Also `--simple-output` (to be machine parseable), e.g. outputting just `HUnit-1.5.0.0 ListLike-4.5 QuickCheck-2.9.2` etc.

23Skidoo: It already supports `--exit-code`, I just forgot to document it. Thanks for the reminder. Will add `--simple-output`.

23Skidoo: Oh, actually I didn't, `--exit-code` is documented. `--quiet` is a good idea.

phadej: Sorry for not spotting it

23Skidoo:
Rebased, implemented `--quiet` and `--simple-output`, expanded documentation.

23Skidoo:
And now a test suite failure that I cannot reproduce locally...

ezyang:
`outdated_freeze.sh`?  What if you ported the test to `cabal-testsuite`? It's where everything should go eventually, and that might smooth over some difference and make it reproducible.

23Skidoo:
Moved the test case to `cabal-testsuite`, hopefully it works now.

@ezyang Amazing job on the new test suite, BTW.

23Skidoo:
Test suite passes locally.

23Skidoo:
Should be good to merge now, just blocked on the OS X build bot.

23Skidoo:
OS X build bots keep timing out, that's quite annoying. One of them finally passed now, waiting on the last one...

23Skidoo:
CI is green, merging.

Cabal/doc/developing-packages.rst
```diff
@@ -1147,6 +1147,83 @@ For example, given the following dependencies specified in
...
+``--new-freeze-file``
+    Read dependency version bounds from the new-style freeze file
+    (``cabal.project.freeze``) instead of the package description file.
+``--simple-output``
+    Print only the names of outdated dependencies, one per line.
+``--exit-code``
+    Exit with a non-zero exit code when there are outdated dependencies.
+``-q, --quiet``
+    Don't print any output. Implies ``-v0`` and ``--exit-code``.
+``--ignore`` *PACKAGENAMES*
```
jonathanjouty: @23Skidoo Pedantry... shouldn't this be `*[PACKAGENAMES]*`?

23Skidoo: Actually no, that form isn't supported, because ignoring everything doesn't make sense (at least until a whitelist feature is added).

Issue #3938, opened 2016-10-04
Strange OS X failures on Travis with GHC 7.8.4


23Skidoo:
Our CI recently started failing on OS X with GHC 7.8.4. There's a large number of permissions failures, all of them looking like this:

```
    Command /Users/travis/build/haskell/cabal/Cabal/tests/Setup configure -v --distdir dist-test/./dist --global --with-ghc /Users/travis/.ghc-install/7.8.4/bin/ghc --disable-optimization '--prefix=/Users/travis/build/haskell/cabal/Cabal/tests/PackageTests/BenchmarkStanza/dist-test/usr' '--package-db=clear' '--package-db=global' '--package-db=user' '--package-db=/Users/travis/build/haskell/cabal/dist-newstyle/build/Cabal-1.25.0.0/package.conf.inplace' failed.
    Output:
    Warning: my.cabal: A package using section syntax must specify at least
    'cabal-version: >= 1.2'.
    Configuring BenchmarkStanza-0.1...
    Dependency base -any: using base-4.7.0.2
    Setup: /usr/bin/ar: permission denied
```

The problem is probably not on our side, but this is still annoying, since it keeps happening.

/cc @phadej

23Skidoo:
The first failing commit was https://github.com/haskell/cabal/commit/c6929fc334d1aad1fbb9867f1828060b3cf760f2, which passed when it was a PR: https://github.com/haskell/cabal/pull/3929.

ezyang:
I think that commit is pretty unlikely to be causing the failure? What is more likely the culprit is Travis's update to the OS X image: https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/

ezyang:
See also #2653.

ezyang:
Which suggests that if we add `constraint: unix>=2.7.1.0` that should solve the problem... except that we need our bootstrapping cabal to be updated :(

23Skidoo:
>  except that we need our bootstrapping cabal to be updated :(

Hmm, I thought [the one we were using](https://www.haskell.org/cabal/release/cabal-install-1.24.0.0/cabal-install-1.24.0.0-x86_64-apple-darwin-yosemite.tar.gz) had this bug fixed.

PR #4208, opened 2017-01-05
Corrected version lower bound on 'directory' dependency.

robjhen:
Needed for the use of System.Directory.makeAbsolute in Distribution.Client.Nix.

I have not modified bootstrap.sh, as it did not appear to contain any reference to the directory package. Have I missed anything here?

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -372,7 +372,7 @@ executable cabal
       build-depends: directory >= 1.1 && < 1.2, old-time >= 1 && < 1.2,
                      process   >= 1.0.1.1  && < 1.1.0.2
     else
-      build-depends: directory >= 1.2 && < 1.4,
+      build-depends: directory >= 1.2.2.0 && < 1.4,
                      process   >= 1.1.0.2  && < 1.5
 
     -- NOTE: you MUST include the network dependency even when network-uri


23Skidoo:
`directory` is a boot library, so it comes with GHC and the bootstrap script doesn't need to install it. Given that GHC 7.8.4 comes with directory 1.2.1.0, we may want to add `makeAbsolute` to a `.Compat` module instead. Not sure why CI didn't catch this.

(Plus there's all that `old-directory`/`old-time` business, not sure whether that still works -- maybe we can remove those flags for 2.0?)

/cc @ttuegel

ttuegel:
> Given that GHC 7.8.4 comes with directory 1.2.1.0, we may want to add makeAbsolute to a .Compat module instead. Not sure why CI didn't catch this.

It's needed for `cabal-install`, but not `Cabal`. You're quite right, we should add a `Compat` module for `System.Directory`.

hvr:
@23Skidoo As for why the CI might not have caught this: I haven't looked at the travis buildlog, but I see there's a `constraints: unix >= 2.7.1.0` in the cabal.project file, so that's likely going to force a reinstall of `directory` anyway for GHC 7.8. If that's the case, we may not need to bother with a `.Compat` module either.

robjhen:
@hvr As I understand it, cabal.project is only used with 'cabal new-build' is that right? I think this is an issue when using the old-style 'cabal build'. CI uses new-build which is why this got through. Assuming we are still supporting old-style build, I'm thinking that probably we do need a Compat module.

@ttuegel How necessary is the use of `makeAbsolute` in Distribution.Client.Nix? Is there any easy way to rewrite the code so as to do without it, or would a Compat module be the more straightforward option?

23Skidoo:
We have that constraint in `cabal.project` because otherwise the test suite wouldn't pass (#3938), but `cabal-install`/`Cabal` should still build with `directory < 1.2.1.0`.

ttuegel:
> Is there any easy way to rewrite the code so as to do without it

I think so; let me give it a go...

23Skidoo:
Closing in favour of #4221.

PR #4211, opened 2017-01-06
Preliminary refactoring in preparation for issue #3502

robjhen:
I've removed some duplicated code and moved some utility functions into more central locations.

I've also added an 'all on one line' pretty-printing style to Distribution.Text called 'flatStyle'. I'll be needing it later to help me eliminate commonality between the debug printing code of 'PackageConstraint' and the pretty printing code of 'UserConstraint'.

I have more code to come, but would you be happy to review these changes first as a standalone pull request?


ezyang:
LGTM!

Cabal/Distribution/Compat/ReadP.hs
```diff
@@ -297,6 +298,11 @@ skipSpaces =
   skip (c:s) | isSpace c = do _ <- get; skip s
   skip _                 = do return ()
 
+skipSpaces1 :: ReadP r ()
+-- ^ Like 'skipSpaces' but succeeds only if there is at least one
```
23Skidoo: BTW, why do Haddock comments in this file use this strange format? Everywhere else we use

```
-- | Foo bar baz [...]
quux :: Foo -> Bar -> [...]
```

This is not something you need to fix, I'm just wondering aloud.

robjhen: As I understand it, Distribution.Compat.ReadP is a compatibility module which was verbatim copied some time ago from Text.ParserCombinators.ReadP in the base package. It uses a different code formatting style from the main cabal code. Incidentally, what is the reason why we need this compatibility module rather than just have cabal depend on ReadP from base as a library?

ezyang: @robjhen In general, Compat modules exist to work around time-limited problems in base libraries. So at some point in time in the past (git log to find out) ReadP was not available in the base package for versions of base we wanted to support. That version may have gone out of our support window, in which case it's OK to drop the module!

PR #4219, opened 2017-01-07
Qualified constraints (issue #3502)

robjhen:
The main changes are:

1. Updated the syntax of qualifiers in PackagePath.hs to use colon as a separator rather than hyphen.
2. Changed PackageName to QPN in PackageConstraint.
3. Added 'UserQualifier' into UserConstraint.

The only difference to the overall behaviour of cabal due to this patch is that qualified constraints will be parsed on the command line. Currently the solver etc. just ignores these qualifiers, however everything should be all set now to make the solver act on them.

One last thing: I'm going to add some new unit tests for the UserConstraint parser, but if it's okay I'll leave that till the next submission to stop this one getting too big.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -23,7 +23,9 @@ module Distribution.Client.Dependency (
     resolveWithoutDependencies,
 
     -- * Constructing resolver policies
+    PackageProperty(..),
     PackageConstraint(..),
+    unqualified,
     PackagesPreferenceDefault(..),
     PackagePreference(..),
 
@@ -346,7 +348,7 @@ dontUpgradeNonUpgradeablePackages params =
   where
     extraConstraints =
       [ LabeledPackageConstraint
-        (PackageConstraintInstalled pkgname)
+        (PackageConstraint (unqualified pkgname) PackagePropertyInstalled)
         ConstraintSourceNonUpgradeablePackage
       | Set.notMember (mkPackageName "base") (depResolverTargets params)
       -- If you change this enumeration, make sure to update the list in
@@ -477,7 +479,8 @@ addSetupCabalMinVersionConstraint :: Version
 addSetupCabalMinVersionConstraint minVersion =
     addConstraints
       [ LabeledPackageConstraint
-          (PackageConstraintVersion cabalPkgname (orLaterVersion minVersion))
+          (PackageConstraint (unqualified cabalPkgname)
+                             (PackagePropertyVersion $ orLaterVersion minVersion))
           ConstraintSetupCabalMinVersion
       ]
   where
@@ -583,8 +586,9 @@ applySandboxInstallPolicy
         (thisVersion (packageVersion pkg)) | pkg <- otherDeps ]
 
   . addConstraints
-      [ let pc = PackageConstraintVersion (packageName pkg)
-                 (thisVersion (packageVersion pkg))
+      [ let pc = PackageConstraint
+                 (unqualified $ packageName pkg)
+                 (PackagePropertyVersion $ thisVersion (packageVersion pkg))
         in LabeledPackageConstraint pc ConstraintSourceModifiedAddSourceDep
       | pkg <- modifiedDeps ]
 
@@ -925,7 +929,8 @@ resolveWithoutDependencies (DepResolverParams targets constraints
     packageVersionConstraintMap =
       let pcs = map unlabelPackageConstraint constraints
       in Map.fromList [ (name, range)
-                      | PackageConstraintVersion name range <- pcs ]
+                      | PackageConstraint
+                          (Q _ name) (PackagePropertyVersion range) <- pcs ]
 
     packagePreferences :: PackageName -> PackagePreferences
     packagePreferences = interpretPackagesPreference targets defpref prefs
--- a/cabal-install/Distribution/Client/Targets.hs
+++ b/cabal-install/Distribution/Client/Targets.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -408,7 +409,7 @@ data PackageTarget pkg =
      -- to be resolved to the right case-sensitive name.
    | PackageTargetNamedFuzzy PackageName [PackageProperty] UserTarget
    | PackageTargetLocation pkg
-  deriving Show
+  deriving (Show, Functor, Foldable, Traversable)
 
 
 -- ------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


23Skidoo:
AppVeyor failure seems to be genuine: 

```
tests\UnitTests\Distribution\Client\ProjectConfig.hs:567:10: error:
    Duplicate instance declarations:
      instance Arbitrary UserConstraint
        -- Defined at tests\UnitTests\Distribution\Client\ProjectConfig.hs:567:10
      instance Arbitrary UserConstraint
        -- Defined at tests\UnitTests\Distribution\Client\ProjectConfig.hs:577:10
```

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -346,7 +348,7 @@ dontUpgradeNonUpgradeablePackages params =
   where
     extraConstraints =
       [ LabeledPackageConstraint
-        (PackageConstraintInstalled pkgname)
+        (PackageConstraint (unqualified pkgname) PackagePropertyInstalled)
```
robjhen: Yes I see your point. This may require a little thought, so I'll make a new PR after this one. Would it make sense instead of expanding `PackageConstraint` to expand `Qualifier` by adding a new constructor? I see two obvious choices:

1. Let 'Unqualified' mean top-level dependency and add a new constructor called e.g. `Any`.
2. Let 'Unqualified' mean applies everywhere and add a new constructor `Toplevel`.

Option 1 seems somewhat confusing semantically, so maybe option 2 is better - what do you think?

cabal-install/Distribution/Client/Targets.hs
```diff
@@ -744,35 +733,42 @@ readUserConstraint str =
       ++ "either a version range, 'installed', 'source' or flags"
```
robjhen: agreed

cabal-install/Distribution/Solver/Types/PackageConstraint.hs
```diff
@@ -1,46 +1,75 @@
...
+dispPackageProperty (PackagePropertyFlags flags) = dispFlagAssignment flags
+dispPackageProperty (PackagePropertyStanzas stanzas) =
+  Disp.hsep $ map (Disp.text . showStanza) stanzas
+
+-- | A package constraint consists of a package plus a property
+-- that must hold for that package.
+data PackageConstraint = PackageConstraint QPN PackageProperty
   deriving (Eq, Show, Generic)
 
 instance Binary PackageConstraint
```
robjhen: Yes, I think you're right - I'll look into that.

cabal-install/Distribution/Solver/Types/PackagePath.hs
```diff
@@ -27,7 +37,15 @@ data Namespace =
     --
     -- For now we just number these (rather than giving them more structure).
   | Independent Int
-  deriving (Eq, Ord, Show)
+  deriving (Eq, Ord, Show, Generic)
+           
+instance Binary Namespace
+
+-- | Pretty-prints a namespace. The result is either empty or
+-- ends in a period, so it can be prepended onto a package name.
```
robjhen: agreed

robjhen:
Thanks very much! I'm going to merge this PR now, and then I'll create a new PR addressing the points you've made.

Regarding your question is there anything else that needs to be done, I think the top priority is to add in the ability to distinguish between 'any' and 'toplevel' qualifiers, as you pointed out. Probably best not to modify the solver until the any vs. toplevel distinction is done. After that I'll see if I can think of anything else that we may have missed.

23Skidoo:
Do we need any documentation for this or is this purely an internal thing?

ezyang:
Documentation for the new syntax would be very helpful. I am a little worried that it will look similar but be subtly different from the current component qualification syntax in new-build.

cabal-install/Distribution/Client/Dependency.hs
robjhen: On second thoughts, having looked at how `Qualifier` is used in the solver, I think my last suggestion was incorrect. So, if I understand you, I should modify `PackageConstraint` to support a wildcard or 'any' qualifier as an extra alternative, but leave the `Qualifier` and `QPN` datatypes as they are?

23Skidoo:
We can still change the syntax, the feature is not in any released version.

robjhen:
@23Skidoo @ezyang Would you mind giving me a link to the code or info concerning the component qualification syntax, as I am not familiar with that feature. I'd like to compare it with the `UserConstraint` syntax. Yes you are right, once this is finalized we will need documentation for qualified constraints on the command line.

ezyang:
http://cabal.readthedocs.io/en/latest/nix-local-build.html#cabal-new-build

robjhen:
Thanks.

Also, just a quick question to check I've understood the semantics of `Exe` qualifiers. Consider the following qualified package name:
```
Q (PackagePath DefaultNamespace
               (Exe (mkPackageName "foo") (mkPackageName "happy")))
  (mkPackageName "containers")
```
In the current provisional syntax this is pretty-printed as:
```
foo:happy:exe.containers
```
As I understand it, this means the following:
- I want to build some package `foo`.
- The `happy` build tool is needed to build `foo`.
- The `containers` library is a dependency of `happy`.

So, for example, specifying the following constraint:
```
foo:happy:exe.containers ==0.4
```
means that when building `foo`, you must use a `happy` executable that has been linked against version 0.4 of the `containers` package.

My question is: have I understood this correctly?

ezyang:
Yep.

cabal-install/Distribution/Client/Dependency.hs
grayjay: > So, if I understand you, I should modify PackageConstraint to support a wildcard or 'any' qualifier as an extra alternative, but leave the Qualifier and QPN datatypes as they are?

Yes, the solver still needs QPN to refer to a single instance of a package in many places.  It might be worth adding a new type for the scope of the constraint, since we'll probably need other scopes later, such as "all setup dependencies". For example:

```haskell
data PackageConstraint = PackageConstraint ConstraintScope PackageProperty

data ConstraintScope =
  Qualified QPN
  | AnyQualifier PackageName
  | AnySetupQualifier PackageName
```

> Let 'Unqualified' mean applies everywhere and add a new constructor Toplevel.

I think Toplevel is a lot clearer than the current use of Unqualified.  (I'm not suggesting a change for these PRs, though.)

grayjay:
> Do we need any documentation for this or is this purely an internal thing?

The changelog should also mention that the meaning of constraints that use the old syntax has changed. They applied to all dependencies in 1.24, but now they only apply to top-level dependencies.

robjhen:
See PR #4228.

@grayjay I've added a 'ConstraintScope' data type as you suggested. I've also renamed 'Unqualified' to 'Toplevel'.

hvr:
Looking at https://github.com/haskell/cabal/pull/4219#issuecomment-271366798, can somebody explain to me the meaning of the `.`s and `:`s symbols in the path

```
foo:happy:exe.containers ==0.4
```

?

Is `happy` the executable name or is it the package name? In the former case it looks a bit confusing to me that `:exe` is postfixed to the executable name, as in the goal syntax it's the other way round (i.e. `happy:exe:happy` denotes the executable component; and the first `happy` occurence before the `:exe` is the package name there)

When do I use `:` delimiter, and when do I use a `.` delimiter?

ezyang:
@hvr We are having discussion about this in #4236. I think the syntax definitely needs to change.

PR #4221, opened 2017-01-09
Remove makeAbsolute

ttuegel:
Replaces #4208. We do not actually care about canonicity of paths, but
makeAbsolute is not available in older versions of directory.

### DIFF
--- a/Cabal/shell.nix
+++ b/Cabal/shell.nix
@@ -1,0 +1,1 @@
+(import ../.).Cabal.env
--- a/cabal-install/Distribution/Client/Nix.hs
+++ b/cabal-install/Distribution/Client/Nix.hs
@@ -17,11 +17,11 @@ import Control.Applicative ((<$>))
 import Control.Exception (bracket, catch)
 import Control.Monad (filterM, when, unless)
 import System.Directory
-       ( createDirectoryIfMissing, doesDirectoryExist, doesFileExist
-       , makeAbsolute, removeDirectoryRecursive, removeFile )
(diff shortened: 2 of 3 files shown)


23Skidoo:
CI failure is due to `-Werror`.

23Skidoo:
Specifically,

```
[ 99 of 132] Compiling Distribution.Client.Nix ( Distribution\Client\Nix.hs, dist\build\cabal\cabal-tmp\Distribution\Client\Nix.o )
Distribution\Client\Nix.hs:23:1: warning: [-Wunused-imports]
    The import of `<.>' from module `System.FilePath' is redundant
<no location info>: error: 
Failing due to -Werror.
```

ttuegel:
What's the procedure for a timed-out build?

phadej:
@ttuegel it still caches the deps, so restart usually lets the job pass. Restarted it (OSX boxes are slow).

ezyang:
Sometimes Travis machines run a lot slower than normal, in which case a restart helps. But if we are consistently hitting the time limit we'll need to slice up the jobs some more.

ttuegel:
It hit the timeout on Darwin again. Are you sure it's caching dependencies? The log doesn't look that way, or maybe I don't understand what it's supposed to cache.

ezyang:
Yeah, I guess we are just over the limit. Slicin' time.

23Skidoo:
OK, I think I'll just merge this. If this works on OS X with 7.10.3 and on Linux with 8.0.1, it should work on OS X with 8.0.1 as well.

PR #4225, opened 2017-01-11
Echo stack yaml

phadej:


ezyang:
OK. Should we make Stack builds mandatory so we catch this in CI next time?

phadej:
@ezyang, well, not many seems to use `stack` for development as `stack.yaml` happened to bitrot. So maybe at this point it's more of additional chore than additional value?

Issue #4209, opened 2017-01-06
Add flag to cabal to specify requiredInstall packages in modular solver.


robinp:
In the modular solver at https://github.com/haskell/cabal/blob/2146ea148fc52ee97549b89c1e9bb18c90610670/cabal-install/Distribution/Solver/Modular/Solver.hs#L134, the list of wired-in packages is hardcoded.

When one wants to rebuild one of these packages (see [mail thread about rebuilding `base`](https://mail.haskell.org/pipermail/haskell-cafe/2017-January/125904.html)), the given package needs to be commented out, and cabal rebuilt locally.

It would be nice if the list of packages came from a flag (defaulting to the current set?), which users could override. Or, if it's needed for some reason, keep the list hardcoded but let some packages be removed with a flag.

grayjay:
@robinp I rebased my branch from #3987 and made a PR.  The flag always ignores the whole list of non-installable libraries, because cabal already has "installed" constraints.  For example, `cabal new-build --constraint="base installed" --install-base-libraries` would allow cabal to install template-haskell but not base.

grayjay:
Fixed by #4227.

PR #4227, opened 2017-01-12
Add flag to allow the dependency solver to install any package (issue #4209).

grayjay:
--install-base-libraries makes cabal ignore the list of packages that cannot be
installed or upgraded.

I haven't added documentation yet.

### DIFF
--- a/Cabal/doc/nix-local-build.rst
+++ b/Cabal/doc/nix-local-build.rst
@@ -1710,6 +1710,21 @@ Most users generally won't need these.
 
     The command line variant of this field is ``--(no-)strong-flags``.
 
+.. cfg-field:: install-base-libraries: boolean
+               --install-base-libraries
+               --no-install-base-libraries
+    :synopsis: Allow cabal to install or upgrade any package.
+
+    :default: False
+
+    By default, the dependency solver doesn't allow ``base``,
+    ``ghc-prim``, ``integer-simple``, ``integer-gmp``, and
+    ``template-haskell`` to be installed or upgraded. This flag
+    removes the restriction.
+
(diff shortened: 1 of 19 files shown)


grayjay:
@23Skidoo Thanks.  I added a line to the changelog and documented the flag in Cabal/doc/nix-local-build.rst.  Is there anything else I should update?

23Skidoo:
I think that's all. Let's merge!

grayjay:
Thanks!

hvr:
Sorry to chime in so late: do we consider `--install-base-libraries` a good name for the flag? I would have expected something like `--allow-reinstall=*` or something to be more in line with `--allow-{newer,older}` and avoid overloading the term "`base`" (we also refer to such libraries sometimes as boot libraries or maybe even core libraries; it's all rather vague)

23Skidoo:
Sure, the name can be changed.

grayjay:
@hvr Thanks for the feedback.  I had a hard time coming up with a descriptive name, so I would be happy to change it to something better.  What about `--allow-boot-library-installs`?  I didn't want to use reinstall, because we already have `--avoid-reinstalls` and `--force-reinstalls`, which use reinstall to mean installing an existing version again.

23Skidoo:
> What about `--allow-boot-library-installs`?

I like it. This is a flag that is useful only in very special circumstances, so a long descriptive name is perfectly OK and makes it unlikely people will use it by mistake.

hvr:
`--allow-boot-library-installs` sounds more consistent to me; sounds good to me!

grayjay:
Thanks.  I made a PR: #4240

PR #4232, opened 2017-01-13
Add whitelist hspec-discover

phadej:


ezyang:
LGTM.

Ericson2314:
Err, shouldn't we move everyone to `build-tool-depends`?

Ericson2314:
Also is `hspec-discover`'s exe also called `hspec-discover`? The whitelist currently assumes the exe within package has the same name.

phadej:
It is

Issue #1493, opened 2013-09-09
build Setup.hs with local compiler when cross-compiling


rwbarton:
I have a script that invokes cabal with `--with-ghc=...`, `--with-ghc-pkg=...`, `--with-gcc=...`, etc. for cross-compiling to Android.  It works great for most packages, but fails for a package with Build-type: Custom, because it builds `Setup.hs` with the cross-compiler specified on the command-line, and then obviously it can't run `dist/setup/setup` because it's an ARM executable.

I can work around the issue by doing an ordinary cabal configure with no `--with-ghc=...` options first, then doing a cabal install with my cross-compiler script, because the second invocation uses the cached `dist/setup/setup` from the first one.  But that's a hack, obviously, and annoying to do when the package I'm installing is a dependency of the one I actually want.

What I would like is a separate set of options for specifying the program configuration to use to build `Setup.hs`; or at least an option to tell cabal to just use the ones on my path.  I found that using `defaultSetupScriptOptions` for `setupScriptOptions` in `configure` and `performInstallations` seemed to do the trick for me, to give an idea of what I want to happen, though I haven't tested the resulting libraries.  I couldn't determine whether there is already a way to achieve the same effect through command-line options.

sheyll:
I will make an experimental version on the fork I have in my github account, and notify you here.
Maybe what's there works, but I have not test-covered or implemented all affected parts yet, go ahead and try.
To answer your question: yes. 
Happy new year, too.

acw:
Hey, what's the status on this? Is there anything we can do to accelerate this getting into mainline?

sheyll:
hi
I will look into it in the next couple of days, I have a fork on github
that contains the current state if anyone is interested.
I remember that I added some unit test and todo comments to be able to pick
up the work where I had to stop.
Basically it was working, but the design approach taken did not feel too
good.

Wbr Sven
Am 09.05.2015 02:13 schrieb "Adam Wick" <notifications@github.com>:

Ericson2314:
In https://github.com/haskell/cabal/issues/3882 @ezyang and I talked about turning Setup.hs into its own component? That probably dovetails with this nicely so setup components can be solved separately against the host compile / host-specific packager?

ezyang:
Setup as a component itself could help, but I am not sure how it relates to @sheyll's patch. @sheyll I just added you to the Cabal team so that you can assign yourself this ticket.

ezyang:
[...]

PR #4236, opened 2017-01-15
Qualified constraints: documentation and unit tests (issue #3502)

robjhen:
In this PR:
- Non-upgradeable package constraints now use `ScopeAnyQualifier`.
- Added documentation for qualified constraints. I've also refactored the documentation a bit as there was duplication of information between the description of the `--constraint` command line option and the `constraints` field of cabal.project (let me know if any objections to my moving things around).
- Added unit tests for parsing qualified constraints.

@grayjay You asked me before if there is anything else you'll need to do aside from enforcing the constraints in `Distribution.Solver.Modular.Preference`. Mainly just the following:
- I would advise briefly checking all instances where `PackageConstraint` / `UserConstraint` values are constructed and pattern-matched (you can just grep for these identifiers). Where they are constructed, check that the correct `ConstraintScope` has been applied (I have currently set them all to `scopeTopLevel` except for the non-upgradeable packages one that is set to `ScopeAnyQualifier`). Where they are pattern-matched, check if you need to handle the qualifier in any way (currently qualifiers are just being ignored at these places). Two cases of pattern matching that are particularly worth inspecting are `packageVersionConstraintMap` in `Distribution.Client.Dependency` and `pcName` in `Distribution.Solver.Modular`.
- If that's all okay, I'll leave it to you to add the changelog entry once the solver part is done!

### DIFF
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -1255,18 +1255,75 @@ Miscellaneous options
 
 .. option:: --constraint=constraint
 
-    Restrict solutions involving a package to a given version range. For
-    example, ``cabal install --constraint="bar==2.1"`` will only
-    consider install plans that do not use ``bar`` at all, or ``bar`` of
-    version 2.1.
+    Restrict solutions involving a package to given version
+    bounds, flag settings, and other properties. For example, to
+    consider only install plans that use version 2.1 of ``bar``
+    or do not use ``bar`` at all, write:
 
-    As a special case, ``cabal install --constraint="bar -none"``
-    prevents ``bar`` from being used at all (``-none`` abbreviates
-    ``> 1 && < 1``); ``cabal install --constraint="bar installed"``
-    prevents reinstallation of the ``bar`` package;
-    ``cabal install --constraint="bar +foo -baz"`` specifies that the
-    flag ``foo`` should be turned on and the ``baz`` flag should be
-    turned off.
+    ::
+        $ cabal install --constraint="bar == 2.1"
+
+    Version bounds have the same syntax as ``build-depends``. As
+    a special case, the following prevents ``bar`` from being
+    used at all:
+
+    ::
+       
+        # Note: this is just syntax sugar for '> 1 && < 1', and is
+        # supported by build-depends.
+        $ cabal install --constraint="bar -none"
+
+    You can also specify flag assignments:
+
+    ::
+
+        # Require bar to be installed with the foo flag turned on and
+        # the baz flag turned off.
+        $ cabal install --constraint="bar +foo -baz"
+
+    To specify multiple constraints, you may pass the
+    ``constraint`` option multiple times.
+
+    There are also some more specialized constraints, which most people
+    don't generally need:
+
+    ::
+
+        # Require that a version of bar be used that is already installed in
+        # the global package database.
+        $ cabal install --constraint="bar installed"
+
+        # Require the local source copy of bar to be used.
+        # (Note: By default, if we have a local package we will
+        # automatically use it, so it will generally not be necessary to
+        # specify this.)
+        $ cabal install --constraint="bar source"
+
+        # Require that bar have test suites and benchmarks enabled.
+        $ cabal install --constraint="bar test" --constraint="bar bench"
+
+    By default, constraints only apply to build dependencies
+    (``build-depends``), build dependencies of build
+    dependencies, and so on. Constraints normally do not apply to
+    dependencies of the ``Setup.hs`` script of any package
+    (``setup-depends``) nor do they apply to build tools
+    (``build-tool-depends``) or the dependencies of build
+    tools. To explicitly apply a constraint to a setup or build
+    tool dependency, you can add a qualifier to the constraint as
+    follows:
+
+    ::
+
+        # Example use of the 'setup' qualifier. This constraint
+        # applies to package bar when it is a dependency of the
+        # Setup.hs script of package foo.
+        $ cabal install --constraint="foo:setup.bar == 1.0"
+
+        # Example use of the 'exe' (executable build tool)
+        # qualifier. This constraint applies to package baz when it
+        # is a dependency of the build tool bar being used
+        # build package foo.
+        $ cabal install --constraint="foo:bar:exe.baz == 1.0"
 
 .. option:: --preference=preference
 
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -1261,6 +1261,7 @@ Miscellaneous options
     or do not use ``bar`` at all, write:
 
     ::
+
         $ cabal install --constraint="bar == 2.1"
 
     Version bounds have the same syntax as ``build-depends``. As
@@ -1321,7 +1322,7 @@ Miscellaneous options
 
         # Example use of the 'exe' (executable build tool)
         # qualifier. This constraint applies to package baz when it
-        # is a dependency of the build tool bar being used
+        # is a dependency of the build tool bar being used to
         # build package foo.
         $ cabal install --constraint="foo:bar:exe.baz == 1.0"
 
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -361,7 +361,7 @@ dontUpgradeNonUpgradeablePackages params =
   where
     extraConstraints =
       [ LabeledPackageConstraint
-        (PackageConstraint (scopeToplevel pkgname) PackagePropertyInstalled)
+        (PackageConstraint (ScopeAnyQualifier pkgname) PackagePropertyInstalled)
         ConstraintSourceNonUpgradeablePackage
       | Set.notMember (mkPackageName "base") (depResolverTargets params)
       -- If you change this enumeration, make sure to update the list in
--- a/Cabal/doc/nix-local-build.rst
+++ b/Cabal/doc/nix-local-build.rst
@@ -613,68 +613,21 @@ Solver configuration options
 The following settings control the behavior of the dependency solver:
 
 .. cfg-field:: constraints: constraints list (comma separated)
-               --constrant="pkg >= 2.0"
+               --constraint="pkg >= 2.0"
     :synopsis: Extra dependencies constraints.
 
-    Add extra constraints to the version bounds, flag settings, and
-    other properties a solver can pick for a package. For example, to
-    only consider install plans that do not use ``bar`` at all, or use
-    ``bar-2.1``, write:
-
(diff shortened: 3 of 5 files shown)


ezyang:
Thanks, we should definitely merge this.

The docs made it a lot clearer to me about what kind of syntax bikeshedding we might want to do for constraints. I think the current syntax is close; in particular, `foo:setup.bar == 2.0` seems fine. So it's just the executable syntax that seems bad. The big problem is that the syntax is not parallel: `foo:setup` says, "the setup component of the foo package" (this is true both with new-build syntax and here), but `foo:bar:exe` says, "the bar `build-tools` dependency of ALL components in foo". So, if `x.y` means, "the y package referenced by entity x", then to say something like "the baz referenced by the executable bar referenced by the build-tools of foo", we actually want a syntax like `foo.barpkg:exe:bar.baz`. Except I don't entirely like this because there is this new `barpkg` thing and maybe if you `build-tools-depend` on multiple executables from the same package you want to end up picking the same version (but maybe not?!)

In part, this is a bit related to the underlying model of how much flexibility the solver should have in choosing the dependencies for executables. One way to design the system is to be *as* flexible as possible: then, it should be possible to single out how we are depsolving a particular executable from a particular build-tools-dependency of a particular package. But maybe that's too much and a user who is actually wanting to exert some control here only cares about, "well, I just want to change what dep alex is using." In that case, the syntax should simplify to handle that case.

Does this make sense?

ezyang:
So, let me give a concrete suggestion. It should be possible to say `exe:alex.alexdep >= 2.0`, or even `pkg:exe:subexe.subexedep`, to apply a constraint to ALL versions of `alex` or `subexe` (from `pkg`) which are in our dependency plan. And maybe that should be the only user visible knob available, for now.

Cabal/doc/installing-packages.rst
```diff
@@ -1255,18 +1255,75 @@ Miscellaneous options
...
-    ``cabal install --constraint="bar +foo -baz"`` specifies that the
-    flag ``foo`` should be turned on and the ``baz`` flag should be
-    turned off.
+    Restrict solutions involving a package to given version
+    bounds, flag settings, and other properties. For example, to
+    consider only install plans that use version 2.1 of ``bar``
+    or do not use ``bar`` at all, write:
+
+    ::
+        $ cabal install --constraint="bar == 2.1"
```
grayjay: This needs another newline after the "::" to form a code block.

Cabal/doc/installing-packages.rst
```diff
@@ -1255,18 +1255,75 @@ Miscellaneous options
...
+
+        # Example use of the 'setup' qualifier. This constraint
+        # applies to package bar when it is a dependency of the
+        # Setup.hs script of package foo.
+        $ cabal install --constraint="foo:setup.bar == 1.0"
+
+        # Example use of the 'exe' (executable build tool)
+        # qualifier. This constraint applies to package baz when it
+        # is a dependency of the build tool bar being used
+        # build package foo.
```
grayjay: to build

grayjay:
I like the idea of trying to make the syntax more similar to new-build's targets.  We may also want to make it consistent with the implication constraints from @dcoutts's package-collections branch: https://github.com/haskell/cabal/commit/657448b5c9fc4555d8acda95fe4e8b570cef135b#diff-2a65a1064ba8104a25e44b665f4aabf5R193

> So, let me give a concrete suggestion. It should be possible to say exe:alex.alexdep >= 2.0, or even pkg:exe:subexe.subexedep, to apply a constraint to ALL versions of alex or subexe (from pkg) which are in our dependency plan. And maybe that should be the only user visible knob available, for now.

So this syntax would apply to all non-lib components?  It seems similar to independent goals qualifiers, which allow different targets to have different dependency versions.

robjhen:
@ezyang @grayjay If/when you need to change the syntax, it should be easy to do. Just update the following four sections of code:
1. Pretty-printing: `dispQualifier` in `Distribution.Solver.Types.PackagePath` (and potentially also `dispConstraintScope` in `Distribution.Solver.Types.PackageConstraint`, although that doesn't currently affect user constraints).
2. Parsing: `Text` instance of `UserConstraint` in `Distribution.Client.Targets`.
3. Parsing unit test: `exampleConstraints` in `UnitTests.Distribution.Client.Targets`.
4. Documentation: one paragraph on qualified constraints in the `--constraint` option section of `doc/installing-packages.rst`.

The syntax that we've got currently was never intended as final, and I agree it could be improved, particularly the `exe` qualifier as you pointed out.

ezyang:
@grayjay How do the independent goal qualifiers look like? :) Oh, those are the like, `goal1.blah` things? Yes, something like that, even though internally we have something more flexible.

grayjay:
> @grayjay How do the independent goal qualifiers look like? :) Oh, those are the like, goal1.blah things? Yes, something like that, even though internally we have something more flexible.

Yes, the independent goals are numbered internally, but I imagine they would look similar to your suggested syntax if we ever exposed them.

> In part, this is a bit related to the underlying model of how much flexibility the solver should have in choosing the dependencies for executables. One way to design the system is to be as flexible as possible: then, it should be possible to single out how we are depsolving a particular executable from a particular build-tools-dependency of a particular package.

I meant to add that the current syntax for build tool dependencies, `pkg1:pkg2:exe.pkg3`, captures everything that the solver uses internally now, but with per-component solving, that will change.  I think that the solver should have maximum flexibility when choosing dependencies for separate executables.  Internally, the solver will probably need a component (of any type) instead of `pkg1`, and an executable component instead of `pkg2`.  I think that your idea would fit well with that design.  We don't want users to always have to specify everything, like `pkg1:ctype1:comp1-pkg2:exe:comp2.pkg3`.  We can just expose the second half of the internal qualifier until there is a use case that requires more targeted constraints.  We could also simplify it further, like you suggested, i.e., `pkg2:exe.pkg3` or `exe:comp2.pkg3`.

Ericson2314:
Hmm, found this. I wrote up a plan for cross compiling involving changing the qualifier type to a list: https://github.com/haskell/cabal/issues/1493#issuecomment-272965698 . I figure this is quite relevant to bike shedding syntax.

Ericson2314:
Ok, so if understand this stuff right, qualifying comes from "encapulating" edges in the dependency DAG. If we think about solving per-package, these encapuslating edges are the single connection between between otherwise-disconnected subgraphs.

Solving per-component is weirder as each non-setup component would have and edge to the setup component coming from the same package, though technically those exes could be configured separately so not the same node restoring single connection property? Let's assume that for the following so we have the single-connection property.

Today we do some "pruning" (I talk about this in the linked comment), but a qualification is a chain of these encapsulating edges that must be traversed to reach the package in question. [Formally, `N` is the set of nodes, `D` is the set of dependency edges, and `E` is the subset of encapsulating edges, then `N / (D \ E)≡` (`N` quotiented by conflating any two nodes that can be reached via non-encapsulating edges) is a tree.] The pruning would be quotienting the space of qualifications in some fashion so the search space for the solver is less big

Now if we do want to be less flexible, or the single-edge property doesn't hold, that's means we don't have a tree, so qualifications do not uniquely reach a node. I guess we have to whittle down the space of connections then---which incidentally this pruning might do.

Anyways, back to the syntax, it might make sense to again forget about the is pruning, and except that syntactically different constraints may in fact conflict. Maybe I'm wrong and actually that's a horrible design, but for the sake of argument let me try to give a grammar for the full unadulterated paths-as-qualifiers for components. 

```bnf
Component ::= 'exe' ':' UnqualComponentName
              -- or, like elsewhere, `$pkg:lib:$pkg` is the default lib?
           |  'lib' ':' UnqualComponentName?
              -- Allow an optional trailing semicolon for consistency with `$pkgs:lib:` ?
           |  'setup'

PackageComponent ::= Package ':' Component

Edge ::= PackageComponent '->' Package ':' 'exe' ':' UnqualComponentName
      |  PackageComponent '->' Package ':' 'setup'
      |  PackageComponent '->' 'base' -- do we even want to expose this?

Bound ::= -- ...Just like today

Qualifier ::= ('(' Edge ')' '.')*

Constraint ::= Qualifier Package Bound
```

grayjay:
@Ericson2314 The solver previously allowed unlimited qualifiers, but they were removed in #3220 because of a problem with cyclic dependencies.  If we need more qualifiers, there might be other ways to prevent the list from growing too long, such as a fixed bound on the length.

> Edge ::= PackageComponent '->' Package ':' 'exe' ':' UnqualComponentName
      |  PackageComponent '->' Package ':' 'setup'
      |  PackageComponent '->' 'base' -- do we even want to expose this?

I like the use of arrows here.

grayjay:
I made a pull request (#4252) to disallow the `exe` qualifier until we decide on syntax.  I left `setup`, because, even if we decide to handle setup scripts more similarly to other build tools, I think it would be nice to also be able to use the shorter syntax.  For example, `pkg1:setup.pkg2` could mean both `pkg1:lib->pkg1:setup.pkg2` and `pkg1:exe:exe1->pkg1:setup.pkg2` if `pkg1` had a library and an executable.  What do you think?  Is there any reason to also wait to expose `setup`?

Cabal/doc/installing-packages.rst
```diff
@@ -1255,18 +1255,76 @@ Miscellaneous options
...
+
+        $ cabal install --constraint="bar == 2.1"
+
+    Version bounds have the same syntax as ``build-depends``. As
+    a special case, the following prevents ``bar`` from being
+    used at all:
+
+    ::
+       
+        # Note: this is just syntax sugar for '> 1 && < 1', and is
```
hvr: Btw, why do we use this arbitrary `> 1 && < 1` contradiction instead of the more canonical `< 0` constraint?

ezyang: Patches accepted :) But `-none` is literally implemented this way:

```
noVersion :: VersionRange
noVersion = IntersectVersionRanges (LaterVersion v) (EarlierVersion v)
  where v = mkVersion [1]
```

We could change this as long as it is actually impossible to have negative version numbers :)

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -361,7 +361,7 @@ dontUpgradeNonUpgradeablePackages params =
   where
     extraConstraints =
       [ LabeledPackageConstraint
-        (PackageConstraint (scopeToplevel pkgname) PackagePropertyInstalled)
+        (PackageConstraint (ScopeAnyQualifier pkgname) PackagePropertyInstalled)
```
grayjay: `scopeTopLevel` was only added in #4219 (originally `unqualified`).  Before that, every constraint applied to all occurrences of the package, so the constraints above always prevented `base`, etc. from being installed.

PR #4239, opened 2017-01-16
Extra commentary on build-tools BC.

ezyang:

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -1716,10 +1716,16 @@ system-dependent values for these fields.
     Cabal can make sure that specified programs are built and on the ``PATH`` before building the component in question.
     It will always do so for internal dependencies, and also do so for external dependencies when using Nix-style local builds.
 
+    :pkg-field:`build-tool-depends` was added in Cabal 2.0, and it will
+    be ignored (with a warning) with old versions of Cabal.  See
+    :pkg-field:`build-tools` for more information about backwards
+    compatibility.
+
 .. pkg-field:: build-tools: program list
    :deprecated:
 
-    Deprecated in favor of :pkg-field:`build-tool-depends`.
+    Deprecated in favor of :pkg-field:`build-tool-depends`, but see
+    below for backwards compatibility information.
 
     A list of Haskell programs needed to build this component.
     Each may be followed by an optional version bound.
@@ -1735,11 +1741,19 @@ system-dependent values for these fields.
     an executable in the package will override any of the hard-coded packages with the same name,
     and a hard-coded package will override any executable on the ``PATH``.
 
-    In the first two cases, the list entry is desguared into a :pkg-field:`build-tool-depends` entry.
+    In the first two cases, the list entry is desugared into a :pkg-field:`build-tool-depends` entry.
     In the first case, the entry is desugared into a :pkg-field:`build-tool-depends` entry by prefixing with ``$pkg:``.
     In the second case, it is desugared by looking up the package and executable name in a hard-coded table.
     In either case, the optional version bound is passed through unchanged.
     Refer to the documentation for :pkg-field:`build-tool-depends` to understand the desugared field's meaning, along with restrictions on version bounds.
+
+    If backwards-compatibility is important, you may opt to use
+    :pkg-field:`build-tool-depends` for cases (1) and (2), as they
+    are supported by old versions of Cabal.  Case (3) should not
+    be relied upon, as support for taking executables from ``PATH``
+    was not added until Cabal 2.0, and you'll get errors with old
+    versions of Cabal if you put an unrecognized executable into
+    :pkg-field:`build-tools`.
 
 .. pkg-field:: buildable: boolean
 


Ericson2314:
Oh huh, to be clear I thought the PATH lookup was older. Also, it's that third case which is currently borked on until I make that next PR (which should be soon, promise!).

23Skidoo:
Looks like you forgot `[ci skip]`.

Cabal/doc/developing-packages.rst
```diff
@@ -1716,10 +1716,16 @@ system-dependent values for these fields.
...
+    be ignored (with a warning) with old versions of Cabal.  See
+    :pkg-field:`build-tools` for more information about backwards
+    compatibility.
+
 .. pkg-field:: build-tools: program list
    :deprecated:
 
-    Deprecated in favor of :pkg-field:`build-tool-depends`.
+    Deprecated in favor of :pkg-field:`build-tool-depends`, but see
+    below for backwards compatibility information.
```
phadej: can this be a link? I'm not very familiar with rst, but I guess it supports linking to anchors? (the section is long below)

Ericson2314: There is another deprecation notice I modeled this on. It would be cool if `pkg-field` itself did the linking, but that may be too much to ask (it is used in the defining heading too).

Cabal/doc/developing-packages.rst
```diff
@@ -1735,12 +1741,20 @@ system-dependent values for these fields.
     an executable in the package will override any of the hard-coded packages with the same name,
     and a hard-coded package will override any executable on the ``PATH``.
 
-    In the first two cases, the list entry is desguared into a :pkg-field:`build-tool-depends` entry.
+    In the first two cases, the list entry is desugared into a :pkg-field:`build-tool-depends` entry.
```
Ericson2314: Oops! :) thanks for fixing this!

Cabal/doc/developing-packages.rst
```diff
@@ -1735,12 +1741,20 @@ system-dependent values for these fields.
...
     In the first case, the entry is desugared into a :pkg-field:`build-tool-depends` entry by prefixing with ``$pkg:``.
     In the second case, it is desugared by looking up the package and executable name in a hard-coded table.
     In either case, the optional version bound is passed through unchanged.
     Refer to the documentation for :pkg-field:`build-tool-depends` to understand the desugared field's meaning, along with restrictions on version bounds.
 
+    If backwards-compatibility is important, you may opt to use
```
Ericson2314: How is this possible? `build-tool-depends` is brand new? Do we want cabal-install to desugar it for old `Cabal`?

Cabal/doc/developing-packages.rst
```diff
@@ -1735,12 +1741,20 @@ system-dependent values for these fields.
...
+    :pkg-field:`build-tool-depends` for cases (1) and (2), as they
+    are supported by old versions of Cabal.  Case (3) should not
+    be relied upon, as support for taking executables from ``PATH``
```
Ericson2314: Do we even want to add this feature then? It's sort of a hacky feature, and do we really want to overload build-tools even more? `foreign-build-tool-depends` might be a nice name for it.

Cabal/doc/developing-packages.rst
```diff
@@ -1716,10 +1716,16 @@ system-dependent values for these fields.
     Cabal can make sure that specified programs are built and on the ``PATH`` before building the component in question.
     It will always do so for internal dependencies, and also do so for external dependencies when using Nix-style local builds.
 
+    :pkg-field:`build-tool-depends` was added in Cabal 2.0, and it will
```
Ericson2314: nit nit: Should we start doing the 1 sentance per line thing? :) I did that with the documentation I wrote.

Cabal/doc/developing-packages.rst
ezyang: The idea is that, with old Cabal, there was no way to actually request the executable be installed, so it was always a case of "user has to manually install executable before you try installing this." So putting the dep in `build-tool-depends` (where it will be ignored) is no worse than the status quo. (And they can put it in `build-depends` if they want to do that horrible hack which only works with `cabal install`)

Ericson2314: Oh, so rather than PATH lookup, this checks the ghc db? I had assumed this workaround was PATH lookup.

ezyang: Erm, perhaps I didn't explain clearly. Let's suppose that you're using an external program via `{-# OPTIONS_GHC -F -pgmF my-custom-preprocessor #-}`. GHC relies on being able to invoke `my-custom-preprocessor` from the PATH it's invoked with. So how do we get it into the PATH?

In old Cabal, cabal-install won't do this for you unless you `cabal install my-custom-preprocessor`. One way to cause this command to be run is if you `build-depends: my-custom-preprocessor` and you have a dummy library in the package; then it will get installed. Or you can ask the user to install it manually.

Cabal/doc/developing-packages.rst
ezyang: Are you referring to the PATH feature? Unfortunately I can't really remove it, because it's legal to put an internal exe in build-tools, and when I implemented per-component build I needed this to keep working.

ezyang:
OK I edited and merged.

Cabal/doc/developing-packages.rst
Ericson2314: Gotcha

Cabal/doc/developing-packages.rst
Ericson2314: Oh wow, haha. Glad we're beyond this world!

Issue #4087, opened 2016-11-05
Per-component dependency solving


grayjay:
I don't think there is a clear design for component-based solving yet, so I wanted to create an issue where we can discuss it.  #1575 addresses circular dependencies in test suites but is moving towards a different solution.

Related issues:
#779, #3978, https://github.com/haskell/cabal/issues/3492#issuecomment-225391004 - solver doesn't reject configurations that require unbuildable/non-existent components
#1575 - cabal can't handle cycles between packages that are not cycles between components
#2725 - cabal requires dependencies of executables that aren't being built
#3662 - more per-component support
#3263 - fine-grained dependencies

grayjay:
Here are my thoughts, so far.  At a minimum, the solver should keep track of dependencies between packages' components.  (I think that it currently tracks dependencies from components to packages.)  Then it can stop requiring dependencies of components that aren't needed, and ensure that required components are buildable.  The solver should still ensure that all components within a package are consistent (have the same version and flag assignments), unless they have different qualifiers.  Maybe we can add a flag or use private dependencies to relax that restriction later.

ezyang:
@grayjay, thanks for making this ticket. Your plan seems eminently reasonable. Go for it!

grayjay:
Here is a first pass at a design:

### Summary
- This design isn't 100% component-based: We keep package variables and add one
  boolean variable for each component.  However, the nodes of the dependency
  graph are components.
- We transform packages' dependency trees to handle all combinations of setup
  Cabal version, Buildable field value, component enabled/disabled, and
  `--enable-tests`/`benchmarks`.  This design would also fix #3881.
- We add several search tree transformations to enforce new component-related
  constraints.


## Tree structure
- Per-component solving requires some changes to the variables used by the
  solver, which means that we also need to change the goals, as well as the
  choice nodes created for those goals.
### Variables
- Types of variables (including those that already exist):
  - **Package**:
    - The instance chosen for a specific package, as before.
  - **Flag**:
    - The boolean value for a specific flag, as before.
  - **Flag-like** (I'm not sure whether these should be implemented as flag
      variables, stanza variables, or something new):
    - Each has two possible values.
    - They are used in conditionals to control dependencies, just like flags,
      except that they aren't in the `.cabal` file.  They need to be added when
      the solver converts the package indexes to the solver-specific index
      format.
    - Three types of flag-like variables per package:
      - **Tests/Benchmarks enabled**:
        - Two variables per package, just like current test and benchmark
[...]

PR #4241, opened 2017-01-17
[WIP] Add glob deps

### LINKED ISSUE #4217: Only build needed executables in a package
Once https://github.com/haskell/cabal/pull/4055 is merged, `build-tool-depends` will be implemented, but somewhat shoddily in that on the cabal-install side the exact needed exe dependency will be thrown away and all built instead. This is extra unneeded work, and also changes the PATH in potentially unexpected ways.

Additionally, #3708 in the original issue proposed there also be wildcard deps, for depending on all exes in the package as the implementation currently does, which #4055 did not implement. Parsing and handling those would be a good thing to do in conjunction with this, lest the wildcard case in solving and building be deleted and then reimplemented.

Ericson2314:
This is really really WIP, but I wanted to ask whether my top commit, adding an extra field to exe deps in the solver, will help route that information to ProjectPlanning for the plan in  https://github.com/haskell/cabal/pull/4104#issuecomment-267484861, or whether I'm barking up the wrong tree.

#4217

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -571,6 +571,9 @@ configure (pkg_descr0', pbi) cfg = do
          ++ intercalate "," unsupportedFLibs
 
     -- Configure certain external build tools, see below for which ones.
+    --
+    -- TODO: HACK we're currently ignoring wildcard deps as we don't know
+    -- precisely which executables the resolved package will contain.
     let requiredBuildTools = do
           bi <- enabledBuildInfos pkg_descr enabled
           -- First, we collect any tool dep that we know is external. This is,
@@ -581,7 +584,7 @@ configure (pkg_descr0', pbi) cfg = do
           -- 2. `build-tool-depends` that aren't from the current package.
           let externBuildToolDeps =
                 [ LegacyExeDependency (unUnqualComponentName eName) versionRange
-                | buildTool@(ExeDependency _ eName versionRange)
+                | buildTool@(ExeDependency _ (Just eName) versionRange)
                   <- getAllToolDependencies pkg_descr bi
                 , not $ isInternal pkg_descr buildTool ]
           -- Second, we collect any build-tools entry we don't know how to
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -635,7 +635,7 @@ checkFields pkg =
 
     depMissingInternalExecutable =
       [ dep
-      | dep@(ExeDependency _ eName _) <- internalExeDeps
+      | dep@(ExeDependency _ (Just eName) _) <- internalExeDeps
       , not $ eName `elem` internalExecutables
       ]
 
--- a/Cabal/Distribution/Parsec/Class.hs
+++ b/Cabal/Distribution/Parsec/Class.hs
@@ -141,7 +141,7 @@ instance Parsec ExeDependency where
     parsec = do
         name <- lexemeParsec
         _    <- P.char ':'
-        exe  <- lexemeParsec
+        exe  <- (Just <$> lexemeParsec) <|> (P.char '*' >> pure Nothing)
         ver  <- parsec <|> pure anyVersion
         return (ExeDependency name exe ver)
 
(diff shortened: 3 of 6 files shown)


grayjay:
There is a related issue for refactoring the dependency solver to solve for components instead of packages: #4087

Ericson2314:
@grayjay thanks! Yeah, I view this `role` as a bridge towards that---metadata ferried along by the solver for now, but actually inspected by it in the future.

23Skidoo:
CI failure is due to `-Werror`:

```
Distribution/Simple/Build.hs:33:1: Warning:
    The import of ‘Distribution.Types.ExeDependency’ is redundant
      except perhaps to import instances from ‘Distribution.Types.ExeDependency’
    To import instances alone, use: import Distribution.Types.ExeDependency()
```

Ericson2314:
@23Skidoo thanks! Though heh isn't quite ready to be merged anyways :)

ezyang:
The top commit seems fine to me. What happens if I depend on multiple executables from the same package; does it show up twice?

Ericson2314:
@ezyang hmm probably? I haven't looked at the solving code in much detail---I mainly just changed the definition of `Dep` and then dealt with the fallout as ghc fed it to me. Perhaps it should take a list of exes and have Monoid instance I can use somewhere?

ezyang:
@Ericson2314 My suggestion is to take a look at the output you get after the change. (Hard to get this output? Add some logging!)

Having thought about this more, it may not be necessary to propagate executable dependencies from the solver; ProjectPlanning can recompute them from scratch. Take a look at how we handle exe_deps0 in elaborateSolverToComponents: the first thing we do run elaborateExeSolverId/elaborateExePath on them, which essentially looks up the package that defined the executable(s) and then adds them as deps. With the information you have from the package description, it would be a simple matter to filter this list down to the true dependencies.

(BTW, I think this code is probably in need of some refactoring!)

Ericson2314:
I added this to the 2.0 milestone, because without it build-tool-depends isn't so well implemented.

Ericson2314:
Ok, did things just in ProjectPlanning as @ezyang described (I still have my old commit in another branch for future work :)).

All that's left here is https://github.com/haskell/cabal/pull/4241/files#diff-a6de9cdd16372d3e6c4ec16d4d1703a1R575, I need to route the expanded wildcard back to cabal somehow. Does this mean adding flags to Cabal?

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -570,19 +570,31 @@ configure (pkg_descr0', pbi) cfg = do
...
+          -- 2. `build-tool-depends` that aren't from the current package.
+          let externBuildToolDeps =
+                [ LegacyExeDependency (unUnqualComponentName eName) versionRange
+                | buildTool@(ExeDependency _ (Just eName) versionRange)
+                  <- getAllToolDependencies pkg_descr bi
+                , not $ isInternal pkg_descr buildTool ]
+          -- Second, we collect any build-tools entry we don't know how to
+          -- desugar. We'll never have no idea how to build them, so we just
+          -- hope they are already on the PATH.
+          let unknownBuildTools =
```
ezyang: I just realized that I may have communicated the situation here a little poorly. It is true that Setup was modified to pick up unknown build-tools from PATH, but in fact, all the testing code we have does not seem to actually exercise this case, as it is always assumed the build-tools refers to an internal dep. So a test here would be appreciated.

Ericson2314: Will do.

Ericson2314:
@ezyang Yes, definitely tests, but I do need the route the expanded wildcard back to Cabal too, right?

ezyang:
I guess so (although it's not altogether clear what that looks like; I don't think Cabal has any way of conveniently finding out what the wildcard will glob).

Ericson2314:
I think the bug-fix portion of this should at least make 2.0 if not wildcards, so I'm going to factor those apart.

ezyang:
Hello @Ericson2314, what would you like to be done on this PR specifically?

Ericson2314:
@ezyang It's in limbo until someone has a good idea for routing the glob expansion back to Cabal. Not a priority for 2.0 I'd think unless @hvr differs (it's back-compat with explicit `:*` syntax).

ezyang:
OK. Maybe this presents the case for extending the database to also include executables, or maybe teaching Cabal how to interact with Nix in some way.

gbaz:
#4217 looks like it was resolved without this pr. can this be closed?

jneira:
closing optimistically after no feedback for the previous comment, thanks all!

PR #4244, opened 2017-01-18
WIP: new-test

phadej:
I have no idea what I have done, so all comments are more than welcome.

### DIFF
--- a/cabal-install/Distribution/Client/CmdTest.hs
+++ b/cabal-install/Distribution/Client/CmdTest.hs
@@ -1,0 +1,83 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
+-- | cabal-install CLI command: new-test
+--
+module Distribution.Client.CmdTest (
+    testCommand,
+    testAction,
+  ) where
+
+import Distribution.Client.ProjectOrchestration
+import Distribution.Client.ProjectPlanning
+         ( PackageTarget(..) )
+import Distribution.Client.BuildTarget
+         ( readUserBuildTargets )
+
+import Distribution.Client.Setup
+         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags )
+import Distribution.Simple.Setup
+         ( HaddockFlags, fromFlagOrDefault )
+import Distribution.Verbosity
+         ( normal )
+
+import Distribution.Simple.Command
+         ( CommandUI(..), usageAlternatives )
+import Distribution.Simple.Utils
+         ( wrapText )
+import qualified Distribution.Client.Setup as Client
+
+testCommand :: CommandUI (ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags)
+testCommand = Client.installCommand {
+  commandName         = "new-test",
+  commandSynopsis     = "Perform new-build and run tests",
+  commandUsage        = usageAlternatives "new-test" [ "[FLAGS] TARGET" ],
+  commandDescription  = Just $ \_ -> wrapText $
+        "Build and run test targets",
+  commandNotes        = Just $ \_pname ->
+        "Examples:\n"
+   }
+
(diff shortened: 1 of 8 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1877,13 +1887,17 @@ elaboratePackageTargets ElaboratedConfiguredPackage{..} targets =
                       $ targets
         buildHaddocks = HaddockDefaultComponents `elem` targets
 
-     in (buildTargets, replTargets, buildHaddocks)
+     in (buildTargets, testTargets, replTargets, buildHaddocks)
   where
     --TODO: need to report an error here if defaultComponents is empty
     elaborateBuildTarget  BuildDefaultComponents    = pkgDefaultComponents
     elaborateBuildTarget (BuildSpecificComponent t) = [t]
```
phadej: I guess I have to include test targets here as well.

ezyang:
Big question from me is how this relates to #4179.

phadej:
I wasnt' aware of #4179. And it's looks like bitrotten, and too big, I don't like refactors and new features in single diff. FWIW, I don't see #4179 going in soon enough. This on other hand is ~200 lines diff.

23Skidoo:
I'd like @dcoutts to weigh in before merging.

phadej:
FWIW, I use this with `servant` repository. The output is a bit misleading, but e.g .`cabal new-test --enable tests servant` does there exactly what I want.

phadej:
... i.e. I'm totally OK with @dcoutts eventually refactoring ProjectPlanning etc, but this is valuable already today.

cabal-install/Distribution/Client/CmdTest.hs
```diff
@@ -0,0 +1,88 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
+-- | cabal-install CLI command: repl
```
ezyang: typo

ezyang:
Irrelevant of anything, this needs tests. In this case, I would advise just finding the tests in cabal-testsuite which exercise Setup test, cabal.project'izing them and then running `cabal new-test` to run them. Let me know if you need more guidance on this.

phadej:
@ezyang does the last commit looks somewhat right?

ezyang:
Looks good, but put it in a separate file (current convention is `cabal.test.hs`) so that it can be run in parallel with the setup variant.

phadej:
I'll wait till travis says it's otherwise ok.

phadej:
Separated the tests, travis is green.

ezyang:
Is this non WIP now?

phadej:
Well, it's not "ready, but it's usable.

cabal-install/Distribution/Client/CmdTest.hs
```diff
@@ -0,0 +1,83 @@
...
+  commandUsage        = usageAlternatives "new-test" [ "[FLAGS] TARGET" ],
+  commandDescription  = Just $ \_ -> wrapText $
+        "Build and run test targets",
+  commandNotes        = Just $ \_pname ->
+        "Examples:\n"
+   }
+
+-- | The @test@ command is very much like @build@. It brings the install plan
+-- up to date, selects that part of the plan needed by the given or implicit
+-- test arget(s) and then executes the plan.
```
ezyang: arget?

phadej:
I'd rather add `new-test` as hidden command. Unfortunately I don't have time to polish it in near future (next month or so), especially if @dcoutts refactor will redo it anyway.

ezyang:
That's OK by me. At least, reading the patch, I don't think the *semantics* are wrong, so even if it's buggy it may still be fine.

It's possible that @dcoutts' refactor won't actually affect this, because actually 'cabal new-run' works quite differently form 'cabal new-test', and from reading this test it does seem like new-test will work fine in this realm.

PR #4250, opened 2017-01-20
Clarify our policy on cabal-install support window.

23Skidoo:
It wasn't clear what "three-year support window for cabal-install" means re: self-upgrade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,11 +147,13 @@ Conventions
   buildable out-of-the-box with the dependencies that shipped with GHC
   for at least five years.  The Travis CI checks this, so most
   developers submit a PR to see if their code works on all these
-  versions of GHC.  cabal-install must also be buildable on all
+  versions of GHC.  `cabal-install` must also be buildable on all
   supported GHCs, although it does not have to be buildable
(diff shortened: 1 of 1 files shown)


23Skidoo:
Related: https://github.com/haskell/cabal/pull/4235#issuecomment-273510258.

/cc @Ericson2314

ezyang:
OK. But if this is a guarantee we will give, we should probably CI it...

23Skidoo:
I can add a check that `cabal install Cabal/ cabal-install/` works with 1.18. Or do you want to check every released major version?

ezyang:
That sounds enough!

23Skidoo:
> But how are we going to get our grubby hands on a cabal-install-1.18 binary? Maybe we can just download it.

I'll need to upload one to https://www.haskell.org/cabal/release/cabal-install-1.18.1.0/

phadej:
Isn't `apt-get install cabal-install-1.18` from https://launchpad.net/~hvr/+archive/ubuntu/ghc enough?

The `cabal-install-1.18` for OS X is a bit problem anyway.

23Skidoo:
I guess that'll work too.

PR #4257, opened 2017-01-22
Enforce qualified constraints in the dependency solver.

grayjay:
This PR is a continuation of #4236.  I made two changes to `ConstraintScope` before enforcing the constraints in the solver:
* I removed the `Namespace` from the `ScopeQualified` constructor.  This change doesn't affect the behavior of the solver, because we currently don't use `Namespace` anywhere.
* I added a `ScopeAnySetupQualifier` constructor to allow constraints to apply to all setup scripts.  The new constructor is only used internally in this PR.

This PR also adds dependencies with version ranges to the solver DSL.  I ended up not needing the version ranges for these unit tests, but I left them because I think they could be useful.

/cc @robjhen @kosmikus 

EDIT:

/cc @dcoutts @23Skidoo @ezyang @Ericson2314 @hvr @phadej 

Here is a summary of the changes, from the changelog:

* Added qualified constraints for setup dependencies. For example,
--constraint="setup.bar == 1.0" constrains all setup dependencies on
bar, and --constraint="foo:setup.bar == 1.0" constrains foo's setup
dependency on bar (part of #3502).
* Non-qualified constraints, such as --constraint="bar == 1.0", now
only apply to top-level dependencies. They don't constrain setup or
build-tool dependencies. The new syntax --constraint="any.bar == 1.0"
constrains all uses of bar.

### DIFF
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -1315,6 +1315,12 @@ Miscellaneous options
 
     ::
 
+        # Example use of the 'any' qualifier. This constraint
+        # applies to package bar anywhere in the dependency graph.
+        $ cabal install --constraint="any.bar == 1.0"
+
+    ::
+
         # Example use of the 'setup' qualifier. This constraint
         # applies to package bar when it is a dependency of the
         # Setup.hs script of package foo.
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -1321,9 +1321,14 @@ Miscellaneous options
 
     ::
 
-        # Example use of the 'setup' qualifier. This constraint
-        # applies to package bar when it is a dependency of the
-        # Setup.hs script of package foo.
+        # Example uses of 'setup' qualifiers.
+
+        # This constraint applies to package bar when it is a
+        # dependency of any Setup.hs script.
+        $ cabal install --constraint="setup.bar == 1.0"
+
+        # This constraint applies to package bar when it is a
+        # dependency of the Setup.hs script of package foo.
         $ cabal install --constraint="foo:setup.bar == 1.0"
 
     ..  TODO: Uncomment this example once we decide on a syntax for 'exe'.
--- a/cabal-install/Distribution/Client/CmdFreeze.hs
+++ b/cabal-install/Distribution/Client/CmdFreeze.hs
@@ -13,7 +13,7 @@ import Distribution.Client.ProjectConfig
          , commandLineFlagsToProjectConfig, writeProjectLocalFreezeConfig
          , findProjectRoot, getProjectFileName )
 import Distribution.Client.Targets
-         ( UserQualifier(..), UserConstraint(..) )
+         ( UserQualifier(..), UserConstraintScope(..), UserConstraint(..) )
 import Distribution.Solver.Types.PackageConstraint
          ( PackageProperty(..) )
 import Distribution.Solver.Types.ConstraintSource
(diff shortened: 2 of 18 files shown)


grayjay:
Since the release is coming up, I wanted to figure out what changes relating to qualified constraints we should include.  I can see three options that wouldn't take long to implement:

1. Disable the parsing of setup qualified constraints so that no changes go into 2.0.  We'll be able to choose the syntax later, possibly after we have component-based solving.
2. Merge this pull request so that we have these changes in 2.0:
    * new setup qualified constraints, such as `foo:setup.bar == 1.0`
    * Non-qualified constraints, such as `bar == 1.0`, only apply to top-level dependencies.
    * There is no longer a way to constrain build tool dependencies, or all setup dependencies, which means that a user can't prevent cabal from installing a particular package.
3. Merge this pull request and add a new syntax for constraints that apply to all dependencies.  The previous PRs added that type of constraint but only used it internally.  They are currently displayed as `any.bar == 1.0` in the log.

I prefer 1 and 3, because I don't want to take away the ability to constrain any dependency.

23Skidoo:
@grayjay 
> I prefer 1 and 3, because I don't want to take away the ability to constrain any dependency.

+1 from me. If we can implement 3 in time, then let's go with 3, otherwise we can postpone merging this PR.

ezyang:
OK, I'm going to mark this as action-required for now. @grayjay, are you on point for (3)?

grayjay:
Yes, I'll try to implement (3) this weekend.  I'll add it to this PR.

grayjay:
I exposed the constraints that apply to all dependencies, as well as constraints that only apply to setup dependencies.  The current syntax is:

`--constraint="setup.Cabal >= 1.24"`  - constrain Cabal in all setup scripts
`--constraint="any.Cabal >= 1.24"`  - constrain all uses of Cabal

Ericson2314:
@grayjay this looks great! But say there is a package called `setup` or `any`? 😛

grayjay:
@Ericson2314 Thanks for reviewing the syntax.  Maybe we should reserve the package names "any" and "setup" to allow more flexibility for future changes to constraint syntax.

Ericson2314:
@grayjay well I'm thinking we can use my arrow notation with * and ** wildcards which can express everything, and then we won't need anything behind what we currently have as short-hands.

If we can reserve package names at this point, that's cool, though.

ezyang:
I went ahead and merged this. Hopefully we can reserve any/setup in the near future; I don't actually know how to go about doing this.

23Skidoo:
Someone who's familiar with `hackage-server` code needs to step in.

grayjay:
Thanks!

> @grayjay well I'm thinking we can use my arrow notation with * and ** wildcards which can express everything, and then we won't need anything behind what we currently have as short-hands.

I'm not sure I understand.  Are you suggesting using asterisks instead?  Then we would have something like `*.Cabal` or `**.Cabal` instead of `any.Cabal`.

Ericson2314:
@grayjay err nah I wasn't trying to subtly imply that. I was thinking if all single constraint path edges can be written `(component -> component)`, `*` could be a wildcard within the component, and `**` could be a wildcard over 0 or more edges. That does many `any` and `**` overlap, but `any` may be nice to have as an English term that mirrors `setup`.

hvr:
>> Hopefully we can reserve any/setup in the near future; I don't actually know how to go about doing this.

> Someone who's familiar with `hackage-server` code needs to step in.

@23Skidoo  @ezyang  Fyi, since yesterday, Hackage rejects package names such as "all", "setup", "any" or "none"

Ericson2314:
Great!

grayjay:
Thanks!

PR #4262, opened 2017-01-24
Fix #4217

### LINKED ISSUE #4217: Only build needed executables in a package

Ericson2314:
I'd like to instead change the solver to route the used
components through (baby steps towards per-component
solving), rather than reconstruct it post-hoc from the
Package Description, but this gets the job done for now.

FYI, I have bad luck running the cabal-install tests locally---so I'm just going to rely on CI for the moment.


ezyang:
What problems are you having running the test suite? (Are you passing `--with-cabal`?)

Ericson2314:
Arg---forgot to pull before `git commit --amend`ing. Knock on wood, but this last one should be it.

Ericson2314:
Only a timeout on Travis.

ezyang:
OK LGTM, CI failures seem spurious.

Issue #3535, opened 2016-07-11
How do I haddock with new-build?


ezyang:
**Summary.** There are a few ways to do it:
1. The most straightforward approach is to introduce Haddock as an extra "target" that can be applied to a component. There is a major downside to this approach, which is that you have to rebuild the _libraries_ when you want to rebuild the Haddocks (because they count as separate "packages" from the perspective of the store). As an addendum to this, you could have some configuration option to stop transitively building Haddock at some point, to avoid thrashing the store (in case you only care about Haddocks for inplace.)
2. A better solution is to somehow treat the Haddock documentation as a separate component that can be installed separately.

---

I see code to do it (c.f. HaddockDefaultComponents), but I don't see anywhere which constructs this?

ezyang:
CC @dcoutts

ezyang:
We can work around the problem by running `cabal act-as-setup -- haddock --builddir=dist-newstyle` (or if you have a Custom build, invoke your setup script directly.)

acfoltzer:
This workaround seems to require having run `cabal configure` already, unfortunately.

ezyang:
Err, just to double check, did you run `cabal act-as-setup -- haddock`, and not `cabal haddock`? (The emailed copy of this ticket has the wrong instructions; I edited it to fix it.)

hamishmack:
The `--builddir` needs to point to the subdirectory for the package:

    cabal new-build
    cabal act-as-setup -- haddock --builddir=dist-newstyle/build/package-name-0.1

Also keep in mind that if your `cabal.project` file is in a parent directory `dist-newstyle` will also be in that directory.

ezyang:
Oh yes, thank you for the correction @hamishmack !

acfoltzer:
Ah, yeah, I should have tried that based on the `copy` workaround. That does indeed work without needing a `cabal configure`.

And as soon as I get this workaround going, I realize that what is more important to us for this feature is to lift this command to the metaproject level. It would be great to run `cabal new-haddock` in a directory with `cabal.project` and watch `dist-newstyle` fill with the docs from all the packages.

jaccokrijnen:
So how can I run haddock on installed packages in my global store?

ezyang:
@jaccokrijnen We recently merged a draft new-haddock build but I don't know if it rebuilds things in the global store with Haddock.

gbaz:
I think this can be closed as we do have a v2-haddock working now.

PR #4265, opened 2017-01-26
Fix #4155 with pkg:libcomp in build-depends

### LINKED ISSUE #4155: Reconsider how to specify dependencies on internal libraries

### LINKED ISSUE #4206: Multiple public libraries in a package
**Motivation.** A common pattern with large scale Haskell projects is to have a large number of tighty-coupled packages that are released in lockstep. One notable example is amazonka; as pointed out in https://github.com/haskell/cabal/issues/4155#issuecomment-270126748 every release involves the lockstep release of 89 packages. Here, the tension between the two uses of packages are clearly on display:

1. A package is a *unit of code*, that can be built independently. amazonka is split into lots of small packages instead of one monolithic package so that end-users can pick and choose what code they actually depend on, rather than bringing one gigantic, mega-library as a dependency of the library.

2. A package is the mechanism for distribution, something that is ascribed a version, author, etc. amazonka is a tightly coupled series of libraries with a common author, and so it makes sense that they want to be distributed together.

The concerns of (1) have overriden the concerns of (2): amazonka is split into small packages which is nice for end-users, but means that the package maintainer needs to upload 89 packages whenever they need to do a new version.

The way to solve this problem is to split apart (1) and (2) into different units. The package should remain the mechanism for distribution, but a package itself should contain multiple libraries, which are independent units of code that can be built separately.

In the Cabal 1.25 cycle, we've added two features which have steadily moved in the direction of making multiple public libraries possible:

* Convenience libraries (#269) mean that we already have Cabal-file level syntax support for defining multiple libraries. Note that these libraries are only accessible inside the package, so to a certain extent, the only thing that would need to be changed is making it possible to refer to these libraries.

* Per-component build (#3064) makes it easy to build each internal library of a package separately, without having to reconfigure a package and then rebuild. This means that, from the perspective of new-build, building multiple libraries from a package is "just as separate" as building multiple packages, and I imagine Stack would be interested in taking advantage of this Setup.hs feature.

So the time is ripe for multiple public libraries.

**Proposal.**
First off, I want to say that for the 2.0 release cycle, I do not think
we should add support for the feature below. However, what I do want to
do is make sure that the design for convenience libraries (which is new)
is forwards compatible with this (see also #4155)

We propose the following syntactic extensions to a Cabal file:

1. First, `build-depends` shall accept the form `pkgname:libname` whereever
`pkgname` was previously accepted.  Thus, the following syntax is now supported:

```
build-depends: amazonka:appstream, amazonka
```

   `amazonka` refers to the "public" library of amazonka (the contents of the `library` stanza with no name), while `amazonka:appstream` refers to `library appstream` inside `amazonka`.  A version range associated with sub-library dependency is a version constraint on the package containing that dependency; e.g., `amazonka:appstream >= 2.0` will force us to pick a version of the `amazonka` package that is greater than or equal to 2.0.

2. We add a new field to `library` stanzas, `public`, which indicates whether or not the library is available to be depended upon.  By default, sub-libraries are NOT public.

NEXT, we need the following modifications to the `Setup.hs` interface:

The `--dependency` flag previously took `pkgname=componentid`; we now augment this to accept strings of the form `pkgname:libname=componentid`, specifying what component should be used for the `libname` of `pkgname`.

**Explanation.**
The primary problem is determining a syntax and semantics for dependencies on sub-libraries of a package. This is actually a bit of a tricky problem, because the `build-depends` field historically serves two purposes: (1) it specifies what libraries are brought into scope, and (2) it specifies version constraints on the packages that we want to bring in. The obvious syntax (using a colon separator between package name and library name) is something like this:

```
build-depends: amazonka:appstream, amazonka:elb
```

But we now have to consider: what is the semantics of a *version-range* applied to one of these internal libraries? E.g., as in:

```
build-depends: amazonka:appstream >= 2.0,
               amazonka:elb >= 3.0
```

Does having separate version ranges even make sense? Because the point of putting all libraries in the same package is to ensure that they are tightly coupled, it doesn't make sense to consider the version range on a library; only on a package. So the `build-depends` above should be considered as levying the combined constraint `>= 2.0 && >= 3.0` to the `amazonka` package as a whole.

This causes a "syntax" problem where, if you want to depend only on sub-libraries of a package, there is no obvious place to put the version bound on the entire package itself.  One way to solve this problem is to add support for the following syntax `amazonka:{appstream,elb} >= 2.0`; now there is an obvious place to put the version range.

**Downsides.** Prior to solving #3732, there will be some loss of expressivity if a number of packages are combined into a single package: cabal-install's dependency solver will solve for the dependencies of ALL the libraries (even if you're only actually interested in using some of them.)  This is because the solver always solves for all the components of a package, whereas it won't solve for the dependencies of a package that you don't depend on.

**Prior art.** This is a redux of https://github.com/haskell/cabal/issues/2716 Here is what has changed since then:

1. The motivation has been substantially improved; last time the motivation involved some Backpack/internal code readability hand-waving; now we specifically identify some existing, lockstep packages which would benefit from this. The proposal has nothing to do with Backpack and stands alone.

2. The previous proposal suggested use of dashes for namespace separation; this proposal uses colons, and the fact that the package must be explicitly specified in `build-depends` means that it is easy to translate a new-style `build-depends` into an old-style list of `Dependency`, which means tooling keeps working.

3. The previous proposal attempted to be backwards compatible. This proposal is not: you'll need a sufficiently recent version of Cabal library to work with it.

CC @mgsloan, @snoyberg, @hvr, @Ericson2314, @23Skidoo, @dcoutts, @edsko

Ericson2314:
Working on finishing @ezyang's branch. I also cleaned up the various build-dep related fields in the process. Everything builds but tests fail.

@ezyang note I renamed `BuildDependency` to `LibDependency` to match `ExeDependency`. I suspect when componentization is complete we might be able to remove plain `Dependency` 😃.


Ericson2314:
Hmm OK it builds now---fixing cabal-install was way easier than I at first thought. Waiting to see what CI says.

ezyang:
@Ericson2314 Tell me when you want review.

Ericson2314:
@ezyang Sure will do. Step one is how can I best run haddock locally? I think that is causing the failure.

ezyang:
It's pretty annoying. Workaround here: https://github.com/haskell/cabal/issues/3535#issuecomment-234411057

Ericson2314:
```
Distribution/Backpack/ConfiguredComponent.hs:116:21: error:
    parse error on input ‘-- *package* exists, but doesn't have this’
```
@ezyang that's something that you wrote :). More importantly, it looks fine to me?

ezyang:
OH. This is a dumb Haddock thing: you're not allowed to have an asterisk at the beginning of a `--` because Haddock will try to interpret as a section header. MY BANE.

Ericson2314:
Sorry to have been slow about this. Hopefully the Haddock error goes away now.

Ericson2314:
Removing the WIP in hopes tests pass! :)

Ericson2314:
Hmm, I do not understand the Travis error.

Ericson2314:
Locally tested, this is getting pretty close to being working, though I can probably shrink the diff a bit (gratuitous changes to `PackageIndex` for example).

23Skidoo:
Remilestoned.

Ericson2314:
I won't be back for a few more days, but how should we accommodate the syntax landed in 2.0? Warning? Deprecation?

ezyang:
I'm not sure. We may need to support it permanently.

Ericson2314:
Oh when I said "deprecate", I meant ban it for min version > 2.0, but support it for min version = 2---deem it legacy syntax but not actually break existing packages.

I've love to make release `2.0.1` and doublethink this into a bugfix, but I'm not holding out hope for such devilry :).

Ericson2314:
Copying old Travis links for comparison (cause I find them hard to find after  the fact)
- travis: https://travis-ci.org/haskell/cabal/builds/226543191
- appveyor: https://ci.appveyor.com/project/23Skidoo/cabal/build/%233424%20(master)

kmicklas:
@Ericson2314 and I rebased this on master. We will try to finish it before it bitrots this time :)

Ericson2314:
Is it possible to get this into 2.1? We'd like to ban legacy-style sub library dependencies as soon as possible, so that they are are only allowed with `cabal-version >= 2.0` projects.

ezyang:
You have no objections from me, please use your discretion.

Ericson2314:
This will depend on refactors that @phadej doesn't want for 2.2, so changed milestone.

phadej:
Does @ezyang or @Ericson2314 plan to come back to this?

phadej:
I assume no.

PR #4274, opened 2017-01-27
Fix #4270: make detailed-0.9 test suites work with dynamic exes

### LINKED ISSUE #4270: Test suite type "detailed-0.9" broken with dynamic executables
Example:

```
/code git clone https://github.com/josefs/STMonadTrans/
Cloning into 'STMonadTrans'...
remote: Counting objects: 138, done.
remote: Total 138 (delta 0), reused 0 (delta 0), pack-reused 138
Receiving objects: 100% (138/138), 18.14 KiB | 0 bytes/s, done.
Resolving deltas: 100% (53/53), done.
/code cd STMonadTrans/
/c/STMonadTrans (master=) cabal sandbox init
Writing a default package environment file to
/code/STMonadTrans/cabal.sandbox.config
Creating a new sandbox at /code/STMonadTrans/.cabal-sandbox
c⏎                                                                                                                                                                                                                                             /c/STMonadTrans (master=) cabal install --only-dependencies
Warning: The package list for 'hackage.haskell.org' is 21 days old.
Resolving dependencies...
Notice: installing into a sandbox located at /code/STMonadTrans/.cabal-sandbox
Configuring mtl-2.2.1...
Building mtl-2.2.1...
Installed mtl-2.2.1
/c/STMonadTrans (master=) cabal configure --enable-tests --enable-shared --enable-executable-dynamic
Warning: The package list for 'hackage.haskell.org' is 21 days old.
Resolving dependencies...
Configuring STMonadTrans-0.3.4...
/c/STMonadTrans (master=) cabal test
Preprocessing library STMonadTrans-0.3.4...
[1 of 2] Compiling Control.Monad.ST.Trans.Internal ( Control/Monad/ST/Trans/Internal.hs, dist/build/Control/Monad/ST/Trans/Internal.o )
[2 of 2] Compiling Control.Monad.ST.Trans ( Control/Monad/ST/Trans.hs, dist/build/Control/Monad/ST/Trans.o )
Preprocessing test suite 'foo' for STMonadTrans-0.3.4...
[1 of 1] Compiling Test             ( test/Test.hs, dist/build/STMonadTrans-0.3.4-1RpplpBlcPJKj6etqjuGLV-foo.test/Test.o )
[1 of 1] Compiling Main             ( dist/build/fooStub/fooStub-tmp/fooStub.hs, dist/build/fooStub/fooStub-tmp/Main.dyn_o )
Linking dist/build/fooStub/fooStub ...
Running 1 test suites...
Test suite foo: RUNNING...
Prelude.read: no parse
```

It does work if I omit the `--enable-executable-dynamic`

bennofs:
The detailed-0.9 test suite type compiles the test module into a shared library,
so for running the tests, we need to include the directory where the shared
library is stored in LD_LIBRARY_PATH.

### DIFF
--- a/Cabal/Distribution/Simple/Test/LibV09.hs
+++ b/Cabal/Distribution/Simple/Test/LibV09.hs
@@ -37,7 +37,7 @@ import qualified Control.Exception as CE
 import System.Directory
     ( createDirectoryIfMissing, doesDirectoryExist, doesFileExist
     , getCurrentDirectory, removeDirectoryRecursive, removeFile
-    , setCurrentDirectory )
+    , setCurrentDirectory, makeAbsolute )
 import System.Exit ( ExitCode(..), exitWith )
 import System.FilePath ( (</>), (<.>) )
 import System.IO ( hClose, hGetContents, hPutStr )
@@ -94,7 +94,8 @@ runTest pkg_descr lbi clbi flags suite = do
                                   let (Platform _ os) = LBI.hostPlatform lbi
                                   paths <- LBI.depLibraryPaths
                                              True False lbi clbi
-                                  return (addLibraryPath os paths shellEnv)
+                                  cpath <- makeAbsolute $ LBI.componentBuildDir lbi clbi
+                                  return (addLibraryPath os (cpath : paths) shellEnv)
                                 else return shellEnv
                 createProcessWithEnv verbosity cmd opts Nothing (Just shellEnv')
                                      -- these handles are closed automatically
--- a/cabal-testsuite/Test/Cabal/Prelude.hs
+++ b/cabal-testsuite/Test/Cabal/Prelude.hs
@@ -573,6 +573,16 @@ hasSharedLibraries = do
     shared_libs_were_removed <- ghcVersionIs (>= mkVersion [7,8])
     return (not (buildOS == Windows && shared_libs_were_removed))
 
+-- | Check if the GHC that is used for compiling package tests has
+-- a shared library of the cabal library under test in its database.
+--
+-- An example where this is needed is if you want to dynamically link
+-- detailed-0.9 test suites, since those depend on the Cabal library unde rtest.
+hasCabalShared :: TestM Bool
+hasCabalShared = do
+  env <- getTestEnv
+  return (testHaveCabalShared env)
+
 ghcVersionIs :: WithCallStack ((Version -> Bool) -> TestM Bool)
 ghcVersionIs f = do
     ghc_program <- requireProgramM ghcProgram
--- a/cabal-testsuite/PackageTests/Regression/T4270/Lib.hs
+++ b/cabal-testsuite/PackageTests/Regression/T4270/Lib.hs
@@ -1,0 +1,4 @@
+module Lib where
+
+foo :: Bool
+foo = True
--- a/cabal-testsuite/PackageTests/Regression/T4270/T4270.cabal
+++ b/cabal-testsuite/PackageTests/Regression/T4270/T4270.cabal
@@ -1,0 +1,20 @@
+name: T4270
+version: 0.1
+cabal-version: >= 1.2
+license: BSD3
+author: Benno Fünfstück
+stability: stable
+category: PackageTests
+build-type: Simple
+cabal-version: >= 1.9.2
+
+description: Check type detailed-0.9 test suites with dynamic linking.
+
+library
+  exposed-modules: Lib
+  build-depends: base
+
+test-suite test
+  type: detailed-0.9
+  test-module: Test
+  build-depends: base, Cabal, T4270
--- a/cabal-testsuite/PackageTests/Regression/T4270/Test.hs
+++ b/cabal-testsuite/PackageTests/Regression/T4270/Test.hs
@@ -1,0 +1,17 @@
+module Test where
+
+import Lib
+
+import Distribution.TestSuite
+
+tests :: IO [Test]
+tests = return [Test bar]
+  where
+    bar = TestInstance
+        { run = return $ Finished run
+        , name = "test"
+        , tags = []
+        , options = []
+        , setOption = \_ _ -> Right bar
+        }
+    run = if foo then Pass else Fail "should pass"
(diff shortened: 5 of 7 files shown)


ttuegel:
Do we need to do the same thing for `exitcode-stdio-1.0` tests?

23Skidoo:
@ttuegel Mayhaps. Does the fix look good to you?

@bennofs Can you please add a regression test?

bennofs:
@ttuegel I believe not since a normal executable component will not build any shared library that needs to be linked against

ttuegel:
> I believe not since a normal executable component will not build any shared library that needs to be linked against

@bennofs What if it links against the package's library component?

ezyang:
From AppVeyor failure, test needs to be disabled on platforms that don't support dynamic linking e.g. Windows. Use `skipUnless =<< hasSharedLibraries`

Cabal/Distribution/Simple/Test/LibV09.hs
```diff
@@ -94,8 +94,10 @@ runTest pkg_descr lbi clbi flags suite = do
                                   let (Platform _ os) = LBI.hostPlatform lbi
                                   paths <- LBI.depLibraryPaths
                                              True False lbi clbi
-                                  return (addLibraryPath os paths shellEnv)
+                                  cpath <- canonicalizePath $ LBI.componentBuildDir lbi clbi
+                                  return (addLibraryPath os (cpath : paths) shellEnv)
                                 else return shellEnv
+                print shellEnv'
```
ezyang: I don't think you wanted to keep this

Cabal/Distribution/Simple/Test/LibV09.hs
```diff
@@ -94,8 +94,10 @@ runTest pkg_descr lbi clbi flags suite = do
                                   let (Platform _ os) = LBI.hostPlatform lbi
                                   paths <- LBI.depLibraryPaths
                                              True False lbi clbi
-                                  return (addLibraryPath os paths shellEnv)
+                                  cpath <- canonicalizePath $ LBI.componentBuildDir lbi clbi
```
ezyang: Are you sure you want canonicalizePath and not makeAbsolute?

bennofs:
@ttuegel That is already handled correctly, since we do add the library dirs for all dependencies. The `LibV09` case just wasn't handled correctly because for this particular case, we need the libdir of the *current* component as well (since that contains the shared lib that is loaded by the test stub), not just of the dependencies.

bennofs:
Should `hasSharedLibraries` be false when the `Cabal` library itself is not build with shared libraries? This is the reason why it's failing on travis right now: for some build configs on travis, we only build the static Cabal library so the test cannot run.

ezyang:
@bennofs You're right! So I guess we also need a  "has Cabal shared library" parameter.  Unfortunately, I don't actually know how to test for this... maybe do a `./configure` style probe? (Don't do it on every test, only when requested; see `withDelay` for an example of how we've been coding this sort of thing.)

bennofs:
This PR actually seems to be a duplicate of #3527

bennofs:
@ezyang I was looking at the test code, and found a function called hasCabalGorGhc in Prelude.hs. What's that function for?

ezyang:
hasCabalForGhc is a different function, related to what we do when we test the Cabal library, passing `--with-ghc` for a different version of GHC. In this situation, we can't use certain features like detailed test suites, because the version of GHC we are compiling the test with doesn't have the new version of Cabal.

bennofs:
I squashed some of the commits, travis build should be fine now.

cabal-testsuite/Test/Cabal/Prelude.hs
```diff
@@ -573,6 +578,31 @@ hasSharedLibraries = do
     shared_libs_were_removed <- ghcVersionIs (>= mkVersion [7,8])
     return (not (buildOS == Windows && shared_libs_were_removed))
 
+-- | Check if the GHC that is used for compiling package tests has
+-- a shared library of the cabal library under test in its database.
+--
+-- An example where this is needed is if you want to dynamically link
+-- detailed-0.9 test suites, since those depend on the Cabal library unde rtest.
+hasCabalShared :: TestM Bool
+hasCabalShared = do
```
ezyang: This is an interesting implementation of `hasCabalShared`. At the very least, I would add a note that it is a relatively expensive check, because it involves banging on GHC, so you really only want to call this once, if at all possible.

Assuming that this is the best way to go about doing this, I have some comments:

* First, are you sure that `runProgramM ghcProgram` isn't enough to invoke GHC?  I believe that GHC should already be configured in `testProgramDb`; the configuration was done in `Test/Cabal/Monad.hs` so I don't think we should have to do it again. If this doesn't work, there is probably a different bug somewhere. Is the problem that `runM` calls `requireSuccess` whereas you need to get a boolean out instead? Maybe we can refactor those functions so you get a version that raises exceptions, and one that doesn't. This is related to #4292 where we wanted to handle `fails` more systematically.

* Second, the invocation of GHC doesn't attempt to control what packages are in scope, which means that you could end up in a situation where there are two `Cabal` packages in scope. I guess GHC is preferring the latest version so it's working correctly in this case, but it would be more robust to explicitly specify which Cabal library we want to use, and the test would stop working the moment someone uses a GHC that comes with a newer version of Cabal.

Is this check the way we want to go about it? I'll admit, I can't think of very many other good choices. One possibility is to look at the `LocalBuildInfo` when we `mkScriptEnv` and check if dynamic compilation was enabled (adding a new field to `ScriptEnv` saying if we were building with dynamic or not.) This seems a lot simpler and efficient!

ezyang:
Sorry about the yak shaving on `hasCabalShared`; I know I'm the one who suggested the configure style check, but I only just realized that maybe the LBI has the info we want. If that method works, I think it's my preferred choice. But even if it doesn't, I hope we can simplify the code in `hasCabalShared` a bit.

bennofs:
Using the `LocalBuildInfo` sounds like a good idea indeed, I will have to try that.

bennofs:
Windows build failure is due to #4005

bennofs:
@ezyang using the LocalBuildInfo was a really good idea, there's now much less code.

ezyang:
@bennofs Good! I restarted the Windows build, if it passes let's ship it!

ezyang:
Failing -Werror; import changes need to be undone.

bennofs:
@ezyang fixed that and squashed the changes into the add-regression-test commit.

ezyang:
Good work!

PR #4279, opened 2017-01-28
Fix deprecation warnings from containers

erikd:
This also bumps the lower bound on containers to `>= 0.5` which
should be fine since that was released in May 2012. This change
should work therefore fine with the last 4 major GHC releases.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -131,6 +131,7 @@ library
     Distribution.Compat.Exception
     Distribution.Compat.Graph
     Distribution.Compat.Internal.TempFile
+    Distribution.Compat.Map.Strict
     Distribution.Compat.Prelude.Internal
     Distribution.Compat.ReadP
     Distribution.Compat.Semigroup
--- a/Cabal/Distribution/Compat/Map/Strict.hs
+++ b/Cabal/Distribution/Compat/Map/Strict.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE CPP #-}
+
+module Distribution.Compat.Map.Strict
+    ( module X
+#if MIN_VERSION_containers(0,5,0)
+#else
+    , insertWith
+#endif
+    ) where
+
+#if MIN_VERSION_containers(0,5,0)
+import Data.Map.Strict as X
+#else
+import Data.Map as X hiding (insertWith, insertWith')
+import qualified Data.Map
+
+insertWith :: Ord k => (a -> a -> a) -> k -> a -> Map k a -> Map k a
+insertWith = Data.Map.insertWith'
+#endif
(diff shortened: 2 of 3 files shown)


phadej:
We do test `Cabal` agains 7.4.2. ping @ezyang

erikd:
Yeah, but the last 4 releases is 7.6, 7.8, 7.10 and 8.0. Also, 8.2 is just around the corner.

ezyang:
We've had this discussion before; the current stance is, if it's not too much work to keep 7.4 support going, let's keep it. E.g., if 7.4 support is making our life hard, drop it, but otherwise, don't.

I don't know how compelling this particular patch is; I'll let you guys make a judgment on it.

erikd:
My reason for submitting this patch is that this deprecation warning was coming up when building GHC from git with `-Werror`. I have also submitted patches for [hoopl](https://github.com/haskell/hoopl/pull/36) and [ghc itself](https://phabricator.haskell.org/D3036).

23Skidoo:
How hard would it be to preserve GHC 7.4 support? If we just need a new `.Compat` module, let's add one.

erikd:
I'll see if I can come up with a .Compat module.

erikd:
Just to make sure I get this right, we create a new module `Distribution.Compat.Map` that compiles with both containers `0.4.*` and `0.5.*` and then use that instead of `Data.Map` everywhere else or only where needed?

ezyang:
I don't think it has to be that major. In this case, we just need un-deprecated versions of the functions we can access, right?

erikd:
This issue is a little more subtle than just a name change. The code in cabal before this PR uses `insertWith'` from `Data.Map`, with the prime indicating that this version is more strict that the unprimed version. My initial change, which was only compatible with `containers >= 0.5`  was to switch to `Data.Map.Strict`. However `containers < 0.5` doesn't have `Data.Map.Strict`.

I therefore provide `Distribution.Compat.Map` which chooses `Data.Map.Strict` when `containers >= 0.5` is available. For `containers < 0.5`  we use`Data.Map` and export the strict version `insertWith'` in place of the less strict one.

Still not sure this is the best solution but at least Travis is happy.

erikd:
@ezyang Happy with this solution?

ezyang:
Shouldn't Strict be in the module name, in this case? (containers strict and lazy maps have historically been the same underlying type; it was just a matter of swapping names out.)

erikd:
@ezyang You mean `s/Distribution.Compat.Map/Distribution.Compat.Map.Strict/` ?

ezyang:
Yep!  (If you don't mind!)

erikd:
Done! Hope travis agrees :).

ezyang:
Good. One thing I'm a little afraid of is if someone imports this expecting all the functions to be strict but actually only insertWith is strict, and that leads to a leak, but I guess we can cross that bridge when we get there.

erikd:
I looked at the possibility of importing this instead of `Data.Map` across the whole project, but that would have touched about 50% of the files.

ezyang:
Yeah, not worth it, honestly.

Issue #4277, opened 2017-01-27
Shorter Travis build times


ezyang:
Travis takes too long to run; esp Mac OS X on GHC 8.0.1 which is running up against the limit.  What can we do?  I see two main ways we can win a lot:

1. cabal-install library.  Currently building cabal-install takes 740s (12min), but this far bigger than it should be: we actually have to build many modules FOUR times (134+70+72+119) because the test suites can't share a library between themselves. Primary objection to this is that we don't really want external clients to start depending on this API. I don't know a good way around this.

2. Saving build products.  Currently building takes 1143s (19min) while tests take 740s (12min).  The tests are embarassingly parallel, so they could all be run in parallel (matrix style), IF ONLY we had a way of efficiently sharing the build products.

(1) is easy to do.  (2) is a bit more difficult, for a few reasons:

- Travis doesn't have any way of blocking builds, waiting for others
  to happen.  This has been requested at https://github.com/travis-ci/travis-ci/issues/249
  but Travis is not planning to implement it.  So we need to manually
  implement this (more on this below).

- We need to store the build products somewhere, so that we can grab
  them when we do the second build.  Preferably somewhere free.  There
  are a number of candidates so we'd have to pick one.

So, here is a straw proposal:

1. We configure Travis on the main Cabal repo to ONLY build executables.
   When the build finishes, the Travis script pushes the binaries to a different,
   Cabal binaries repository.

2. This Cabal binaries repository has, itself, a Travis script, which
   takes the binaries and runs the CI tests on them.  It also has an
   EXTRA Travis webhook to report the result of these runs to the
   ORIGINATING issue.

The biggest technical issue is probably whether or not GitHub can deal
with a repository that is repeatedly having binaries pushed to it.
One possiblity: make a branch per binary, and then regularly clear out the
branches.

Timings were based on https://travis-ci.org/haskell/cabal/jobs/195667947

23Skidoo:
> we actually have to build many modules FOUR times (134+70+72+119) because the test suites can't share a library between themselves.

One thing we can do here is add a special `cabal-install.cabal` file for CI only that has those modules exposed as a library. We'll have to be careful to keep both `.cabal` files in sync, though.

> One possiblity: make a branch per binary, and then regularly clear out the
branches.

Or just use `commit --amend` and `push -f`. This is how haddocks deployment to `cabal-website/gh-pages` works already.

phadej:
[...]

PR #4292, opened 2017-02-02
cabal-testsuite: check return code of `runghc Setup.hs`

bennofs:
This PR fixes a bug in the cabal-testsuite, where the result of calling `runghc Setup.hs`
wasn't checked for failure.

I also wrote a regression test for this, however that test depends on the fact that failing
tests throw `ErrorCall`, and I'm not sure if it's good to rely on that in a test. I haven't
found any other way to check whether a test fails though, since `fails` only checks that all
assertions fail, while in this case I want to check that there actually is an assertion. In
other words, `fails (return ())` succeeds, which is not what we need in this case.

### DIFF
--- a/cabal-testsuite/Test/Cabal/Prelude.hs
+++ b/cabal-testsuite/Test/Cabal/Prelude.hs
@@ -143,11 +143,14 @@ setup' cmd args = do
             if buildType (packageDescription pdesc) == Just Simple
                 then runM (testSetupPath env) full_args
                 -- Run the Custom script!
-                else liftIO $ runghc (testScriptEnv env)
-                                (Just (testCurrentDir env))
-                                (testEnvironment env)
-                                (testCurrentDir env </> "Setup.hs")
-                                full_args
+                else do
+                  r <- liftIO $ runghc (testScriptEnv env)
+                                       (Just (testCurrentDir env))
+                                       (testEnvironment env)
+                                       (testCurrentDir env </> "Setup.hs")
+                                       full_args
+                  record r
+                  requireSuccess r
     -- This code is very tempting (and in principle should be quick:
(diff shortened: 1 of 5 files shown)


ezyang:
Thank you, nice catch.

It is indeed not good to rely on ErrorCall, but in this case it's pretty harmless because we defined `assertFailure`, so it would be an easy matter to update the test if there is a problem. I'd definitely put a comment saying that it's a hack though.

Looking at how the bug arose, maybe we (I?) should refactor the test prelude so that everything that fails does so with exceptions, so that you can't forget to handle them.

23Skidoo:
Looks like we've now got a failing test on Windows.

bennofs:
@23Skidoo yeah, but that seems pretty unrelated to my changes? That test seems to be instable in general, I've got a really weird travis failure with that test previously (with some garbage unicode characters appearing in the output for the package name).

23Skidoo:
Yes, your change just exposed a test that was already failing. It looks like it was @ezyang who wrote that test, so I'll defer to him the decision whether to disable the test or fix it.

bennofs:
@23Skidoo well, that's an integration test for cabal-install, which shouldn't even depend on `cabal-testsuite`, I thought? How could that test be affected by this change?

23Skidoo:
My mistake. I restarted the build, maybe it'll pass this time.

We should open a ticket about the flaky test, here's the log: https://gist.github.com/23Skidoo/01d15a0c3c1f0582c5b6da57b1c815e6

ezyang:
The Windows flaking should go away if you rebuild, that's the known #4005 problem (we could probably also suppress it here by passing `-j1` to new-build)

Travis ran out of time. Let's fix #4277

PR #4301, opened 2017-02-07
Retry when running Appveyor to work around failures.

ezyang:

### DIFF
--- a/appveyor-retry.cmd
+++ b/appveyor-retry.cmd
@@ -1,0 +1,21 @@
+@echo off
+rem Source: https://github.com/appveyor/ci/blob/master/scripts/appveyor-retry.cmd
+rem initiate the retry number
+set retryNumber=0
+set maxRetries=3
+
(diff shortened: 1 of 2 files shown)


23Skidoo:
Yes please.

appveyor-retry.cmd
```diff
@@ -0,0 +1,20 @@
+@echo off
+rem initiate the retry number
+set retryNumber=0
+set maxRetries=500
```
23Skidoo: Seems a bit high.

ezyang: How about 5? (I literally just copy pasted the code from someone else)

ezyang: Oh, the original script has just 3; that seems good to me.

23Skidoo: Yep, somewhere in that ballpark. `travis_retry` retries three times by default.

PR #4302, opened 2017-02-07
Fix strict reading of package index

sergv:
I was having issue with this piece in cabal-install-1.24.0.2 when using local package repository. The issue was manifesting itself as error from `hSeek` being called on a closed handle.

The issue was caused by misplaced `seq`s that were not having any effect. This resulted in `readFile` call at  https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/IndexUtils.hs#L273 to
end before actual data was ready, which was causing `hSeek` error later.

I believe `master` also has this problem since even though `srcpkgs` argument of type `Map` is forced through bang pattern, the map itself is lazy. I believe `seq`ing against monadic action is appropriate here to force the values.

However, this issue was biting cabal built from branch `1.24`. Should I submit this change there as well?

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE GADTs #-}
 
 -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ezyang:
Hi @sergv, nice catch, and thank you fro the PR. Did you have a simple test case which reproduces the problem? (If yes, it would be great if we could add it, but if it's difficult never mind.)

One important comment, you should use `evaluate` here, not `seq`: see https://ghc.haskell.org/trac/ghc/ticket/5129#comment:6 for more background on the matter.

sergv:
Hi @ezyang, thanks for suggesting `evaluate`, I didn't knew that.

23Skidoo:
Lazy I/O strikes again...

23Skidoo:
> Should I submit this change there as well?

I can't guarantee that there will be another 1.24 release, but you're welcome to.

ezyang:
@sergv If you could rebase and then force push your branch, that will clear up the CI failures.

ezyang:
Was merged in a separate PR #4358

PR #4305, opened 2017-02-08
Make cabal-install a library.

ezyang:
TODO: We should put an admonition somewhere that the library is
unstable; basically, we reserve the right to break backwards
compatibility at any point in time; but I didn't know where
the best place to put it was.


ezyang:
One downside of this patchset is that we are no longer testing with asserts. But maybe we should just turn on asserts when we build the binaries in the first place.

23Skidoo:
/cc @dcoutts 

@ezyang Are you OK with not merging this until the 2.0 branch is created?

BardurArantsson:
Could we get an explanation for the rationale if it *isn't* to be able to depend on it?

(Regardless of warnings you slap on things, once the cat's out of the bag, it's *very* hard to stuff it back in. If previous experience is anything to go by -- like it or not -- people *will* start depending on cabal-install and suddenly it's *your* fault if anything breaks. Doesn't matter if you had great big warnings on everything.)

23Skidoo:
@BardurArantsson Making CI build times shorter is the main reason.

23Skidoo:
Re: where to put the warning, I suggest the synopsis field, preferably in bold, so it's right in your face when you're accessing http://hackage.haskell.org/package/cabal-install

BardurArantsson:
I would go so far as to suggest that, in addition to doc warnings, cabal-install (the library) would have to be compiled with some flag to not fail to compile as a dependency without extra work. WDYT?

ezyang:
I'm not sure how we'd go about implementing that. One possibility is to make the library non-buildable without an explicit flag. But we still need the executable to be buildable, so we'd have to copy paste all of the deps / macro stuff to the executable stanza. Might not be too bad.

ezyang:
Here's an updated variant of the patch which only builds the library with the lib flag. I adjusted `cabal.project` so that developers using new-build automatically build with this flag; people still using sandboxes/old-style (e.g., Appveyor) need to provide it explicitly.

BardurArantsson:
@ezyang Fantastic! I guess all that's needed then is to add an admonition to the generated docs? (I'm assuming nobody's actually going to be uploading to Hackage, but can't be too careful! :))

(Only a little nitpick, but I'd have named the flag something *really* crazy like "yes-i-really-understand-that-im-doing-something-wrong-by-building-as-a-library", but I understand why you went with a shorter name :).)

Looking forward to seeing the build time gainz! :)

ezyang:
I don't really want to have to add a header to each module in cabal-install. Are you suggesting we put it in the description maybe?

BardurArantsson:
Oh, yes, just in the overall description, sorry for being unclear. Given the flag I think the danger should (now) be minimal that anyone makes the intention of making it a library :).

ezyang:
I went to modify the description, but actually, aren't we asking for trouble if we mention that there's a library? Ideally, we just don't want the module listing to show up on Hackage. I don't know if we can actually make that happen.

BardurArantsson:
Isn't there a flag which marks a package as "deprecated"? Maybe that's an option, but then again if it causes warnings when installing (or just makes it un-installable), then maybe not.

I kind of buy the argument that "the less attention, the better", given the existence of the flag... maybe just make the flag description even more in-your-face about "DO NOT USE THIS", etc.? :)

ezyang:
Unfortunately, it looks like Hackage isn't smart enough to respect the `buildable` field, see: https://hackage.haskell.org/package/ezyang-unbuildable-library-test-1.0/candidate so I'll just modify synopsis.

ezyang:
I just realized there is one important downside to this patch: if you are building the cabal executable without the lib flag with new-build, recompilation won't work (because I didn't relist all of the modules in the executable stanza in this case). We can fix this by listing every module twice but I kind of don't want to. Maybe this is fine because no new-build developer will build without the lib flag? What do people think.

ezyang:
OK I decided I don't care.

ezyang:
(Merged).

PR #4307, opened 2017-02-08
Change our build process to upload build products for further testing.

ezyang:
See travis/README.md for the executive summary of how this works


travis/id_rsa
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
23Skidoo: Um, shouldn't this file be encrypted?

23Skidoo: Relevant links:
    * https://developer.github.com/guides/managing-deploy-keys/
    * https://docs.travis-ci.com/user/encrypting-files/

See also how pushing to the `cabal-website` repo is implemented.

ezyang: It's purposely not encrypted so that PRs can take advantage of the infrastructure. PRs don't get access to encrypted files.

23Skidoo: OK, so this means that we're giving the whole wide world write access to https://github.com/haskell-pushbot/cabal-binaries
Guess there's no way around this in this scheme.

ezyang: Yeah, unfortunately not :(

ezyang:
Agh, webhook not working. Investigating.

ezyang:
Aha! sake-bot is not actually installed on the cabal repository yet. Pinging @hvr

travis/id_rsa
23Skidoo: OK. Just hoping it won't fill up with spam about Viagra and Cialis.

ezyang:
And it's installed! :)

travis-script.sh
```diff
@@ -107,21 +107,6 @@ if [ "x$CABAL_INSTALL_ONLY" != "xYES" ] ; then
       timed cabal new-build $jobs Cabal Cabal:unit-tests
     fi
 
-    # NB: the '|| exit $?' workaround is required on old broken versions of bash
```
23Skidoo: Do we use this workaround anywhere else? If so, we should preserve the comment.

travis-script.sh
```diff
@@ -200,3 +157,6 @@ unset CABAL_BUILDDIR
 
 # Check what we got
 ${CABAL_INSTALL_BDIR}/build/cabal/cabal --version
+
+# If this fails, we WANT to fail, because the tests will not be running then
+(timed ./travis/upload.sh) || exit $?
```
23Skidoo: Looks like we do.

travis/README.md
```diff
@@ -0,0 +1,81 @@
...
+This folder contains scripts for running CI on Travis.
+
+The most unusual thing about our Travis setup is that, after we finish
+building our project, we upload the build products to Git (via the
+haskell-pushbot account) to a separate repository to do testing.
+There are two reasons we do this:
+
+1. On our slowest configuration (GHC 8 on Mac OS X), the time to
+   build and run tests was easily bumping up against the Travis time
+   limit.  By uploading our build products to a separate account
```
23Skidoo: >  By uploading our build products to a separate account

You accidentally a sentence.

travis-script.sh
ezyang: We do, I'll add it back.

PR #4308, opened 2017-02-09
Fix GHC bootstrapping

bgamari:
GHC's stage0 build doesn't provide MIN_VERSION_* macros. We need to take care to
provide appropriate version check macros by hand in GHC's build system.

### DIFF
--- a/Cabal/Distribution/Compat/Graph.hs
+++ b/Cabal/Distribution/Compat/Graph.hs
@@ -83,13 +83,21 @@ module Distribution.Compat.Graph (
     nodeValue,
 ) where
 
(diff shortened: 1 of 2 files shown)


ezyang:
Thanks. We should figure out how to test this in CI so we don't break it again.

23Skidoo:
LGTM.

23Skidoo:
Is there a way to disable those macros?

ezyang:
Probably the right way to go about doing this is to copy paste the GHC command line that GHC uses to bootstrap ghc-cabal, and see if that builds the modules correctly.

PR #4318, opened 2017-02-13
Improve "Preprocessing library" messages.

ezyang:
Before:

    Preprocessing library for reflex-backpack-0.5.0...

After:

    Preprocessing library 'host' instantiated with
      Reflex.Host.Sig = reflex-backpack-0.5.0-inplace-spider:Reflex.Spider.Backpack
      Reflex.Sig = reflex-backpack-0.5.0-inplace-spider:Reflex.Spider.Backpack
    for reflex-backpack-0.5.0...

(The bug was that in single-component mode, we never output the
library name.  I fixed that, and added more output.)


ezyang:
This PR is subsumed by #4322.

PR #4323, opened 2017-02-13
Make LogProgress abstract.

ezyang:


ezyang:
I'm a bit flummoxed by the error since `Distribution.Compat.Prelude` should bring `Applicative` into scope...

ezyang:
Subsumed by #4324

PR #4330, opened 2017-02-15
DeriveRead for BuildFlags

metaleap:
Just a humble, simple and "free" request to add `deriving (Read)` to `BuildFlags` -- simplifies things slightly for a handful of current/future tool writers / Setup.hs integrations where you dump the buildflags (and other Hook args) in a temp file, invoke your tool (or `--frontend`) to then read that dump in again to get Cabal's rich project/lib/pkg contextual info-collection about the current build.

Of course `StandaloneDeriving` works fine too for this for me right now, but hey, maybe you guys won't mind adding the `Read`, if not no biggie =)

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -1631,7 +1631,7 @@ data BuildFlags = BuildFlags {
     -- UserHooks stop us from passing extra info in other ways
     buildArgs :: [String]
   }
-  deriving (Show, Generic)
+  deriving (Read, Show, Generic)
 
 {-# DEPRECATED buildVerbose "Use buildVerbosity instead" #-}
 buildVerbose :: BuildFlags -> Verbosity


23Skidoo:
OK, I guess. We already have `Read` instances for a bunch of core types.

ezyang:
Note that Read instances are *completely* forwards and backwards incompatible, so using this instance is only a good idea if you know exactly what version of Cabal is outputting it and you're using the same version.

A more robust way to serialize and reload BuildFlags would be to dump the command line format, and then use Cabal's command line parser to reparse it out.

23Skidoo:
Merged, thanks!

metaleap:
@ezyang yeah this is just essentially "my custom setup.hs shows (dumps) and my custom ghc `--frontend` plugin reads (loads)", within milliseconds of each other usually. Both cases "run in" the exact same library which has its dependency on `Cabal-a.b.c.d` --- I'll see if I run into issues.. (docs suggest something related to `ProgramDb` deep in `LocalBuildInfo` might not restore from the `show`-serialized representation, but nevermind for now)

Issue #3933, opened 2016-10-03
Misleading warning "the --global flag is deprecated" when using sandbox


edsko:
Perhaps this is a non-issue if we're going to deprecate sandboxes altogether, but right now doing a "traditional" install of a package into a sandbox yields a warning such as

```
# cabal install async
Warning: the --global flag is deprecated -- it is generally considered a bad
idea to install packages into the global store
Resolving dependencies...
Notice: installing into a sandbox located at
/home/edsko/wt/util/ghc-events-analyze/examples/blogpost/.cabal-sandbox
Configuring stm-2.4.4.1...
Building stm-2.4.4.1...
Installed stm-2.4.4.1
Configuring async-2.1.0...
Building async-2.1.0...
Installed async-2.1.0
```

This is misleading because of course I'm not actually installing anything into the global package store here.

23Skidoo:
I've also seen this. May be as easy as making sandboxes use `--user` instead of `--global`, though I vaguely remember that there was some issue with `--user` that made me change sandboxes to use the latter.

ezyang:
I updated the milestone because I don't think this would be a blocker for 2.0.

edsko:
Maybe it's just the wrong way around?

```
# cabal install ./ghcjs.cabal 
Resolving dependencies...
Configuring base16-bytestring-0.1.1.6...
Configuring byteable-0.1.1...
Configuring base-orphans-0.5.4...
....
```

(no warning), but

```
# cabal sandbox init
Writing a default package environment file to
/opt/ghcjs/HEAD-8.0/cabal.sandbox.config
Creating a new sandbox at /opt/ghcjs/HEAD-8.0/.cabal-sandbox

# cabal install ./ghcjs.cabal 
Warning: the --global flag is deprecated -- it is generally considered a bad
idea to install packages into the global store
Resolving dependencies...
Notice: installing into a sandbox located at
/opt/ghcjs/HEAD-8.0/.cabal-sandbox
Configuring ansi-terminal-0.6.2.3...
Configuring base-orphans-0.5.4...
...
```

though possibly there is something more subtle going on. Anyway, just thought I'd mention this.

23Skidoo:
Will try to look into this today.

edsko:
Well it's not a priority :) I'm just recording issues as I notice them :)

23Skidoo:
Better sooner than later.

rwbarton:
[...]

PR #4334, opened 2017-02-17
issue #3933: skip the --global deprecation warning when installing to a sandbox

dmwit:
Previously `cabal install foo` in a sandbox would say

even though the user didn't actually specify `--global`. That's because sandbox configs specify `--global` by default in their local configs for technical reasons. So this patch just skips the warning if we're going to install into a sandbox.


23Skidoo:
Thanks! I'll look at the code tomorrow.

ezyang:
(Travis failure was due to 2.0 fallout)

23Skidoo:
LGTM, merging.

23Skidoo:
Arghh, this broke the testsuite.

PR #4337, opened 2017-02-18
Accept tests megapatch

ezyang:
If possible, I'd like this patchset to go on the 2.0 branch.

* Big pile of bug fixes. These are all in separate patches.
* Backpack mix-in linking error messages are much clearer.
* Verbosity extra flags, like +callstack, now get propagated to Cabal. There's a hack to drop this info when the Custom setup doesn't support it, see 08a564e for more
* A new `die'` function which takes Verbosity, and can emit callstacks. This patch touched a lot of code (which is why I want it in 2.0)
* An "expect test" framework for cabal-testsuite, where we save the recorded output directly as the golden version. There are lots of goodies, see the README in cabal-testsuite for more details


BardurArantsson:
Do the patches make sense individually? If so, I'll try to at least do a brief review today.

ezyang:
Yep. Some of them are "refactor in preparation for" but they should all be self-contained.

Cabal/Distribution/Backpack/DescribeUnitId.hs
```diff
@@ -0,0 +1,61 @@
...
+--
+setupMessage' :: Text a => Verbosity
+             -> String            -- ^ Operation being done (capitalized), on:
+             -> PackageIdentifier -- ^ Package
+             -> ComponentName     -- ^ Component name
+             -> Maybe [(ModuleName, a)] -- ^ Instantiation, if available.
+                                        -- Polymorphic to take
+                                        -- 'OpenModule' or 'Module'
+             -> IO ()
+setupMessage' verbosity msg pkgid cname mb_insts = withFrozenCallStack $ do
```
BardurArantsson: The "prime" naming convention is from your other work on taking a `Verbosity` flag, right?

Cabal/Distribution/Simple/Haddock.hs
```diff
@@ -209,24 +210,29 @@ haddock pkg_descr lbi suffixes flags' = do
            warn (fromFlag $ haddockVerbosity flags)
              "Unsupported component, skipping..."
            return ()
+        -- Don't print this message if nothing is going to happen...
```
BardurArantsson: Not sure the comment makes sense

Cabal/Distribution/Utils/LogProgress.hs
```diff
@@ -15,21 +17,40 @@ import Distribution.Utils.Progress
...
+data LogEnv = LogEnv {
+        le_verbosity :: Verbosity,
+        le_context   :: [Doc]
+    }
 
 -- | The 'Progress' monad with specialized logging and
 -- error messages.
-newtype LogProgress a = LogProgress (Progress LogMsg Doc a)
-    deriving (Functor, Applicative, Monad)
+newtype LogProgress a = LogProgress { unLogProgress :: LogEnv -> Progress LogMsg Doc a }
```
BardurArantsson: Not a big deal, but isn't there a Reader(T) somewhere from a library that `Cabal` already depends on?

Cabal/Distribution/Utils/LogProgress.hs
```diff
@@ -63,11 +63,15 @@ runLogProgress verbosity (LogProgress m) =
 
 -- | Output a warning trace message in 'LogProgress'.
 warnProgress :: Doc -> LogProgress ()
-warnProgress s = LogProgress $ \_ -> stepProgress (LogMsg normal (text "Warning:" <+> s))
+warnProgress s = LogProgress $ \env ->
+    when (le_verbosity env >= normal) $
```
BardurArantsson: Heh, "le verbosity" sounds like a meme of some kind :).

Cabal/Distribution/Backpack/LinkedComponent.hs
```diff
@@ -185,18 +191,19 @@ toLinkedComponent verbosity db this_pid pkg_map ConfiguredComponent {
 
     -- linked_shape0 is almost complete, but it doesn't contain
     -- the actual modules we export ourselves.  Add them!
-    let reqs = modScopeRequires linked_shape0
-        -- check that there aren't pre-filled requirements...
+    let reqs = Map.keysSet (modScopeRequires linked_shape0)
+                `Set.union` Set.fromList src_reqs
```
BardurArantsson: Is the union here an effective change in behavior? If so, maybe it should be mention in commit msg.

Cabal/Distribution/Backpack/LinkedComponent.hs
```diff
@@ -205,7 +212,13 @@ toLinkedComponent verbosity db this_pid pkg_map ConfiguredComponent {
                           modScopeProvides =
                             Map.unionWith (++)
                               local_exports
-                              (modScopeProvides linked_shape0)
+                              (modScopeProvides linked_shape0),
+                          -- TODO: test that requirements aren't already
+                          -- in scope
+                          modScopeRequires =
+                            Map.unionWith (++)
```
BardurArantsson: Ditto

Cabal/Distribution/Backpack/ModuleScope.hs
```diff
@@ -58,29 +57,28 @@ import qualified Data.Set as Set
...
     modScopeProvides :: ModuleProvides,
-    modScopeRequires :: Set ModuleName
+    modScopeRequires :: ModuleRequires
     }
 
 -- | Every 'Module' in scope at a 'ModuleName' is annotated with
 -- the 'PackageName' it comes from.
 type ModuleProvides = Map ModuleName [ModuleSource]
+-- | INVARIANT: entries for ModuleName m, have msrc_module is OpenModuleVar m
+type ModuleRequires = Map ModuleName [ModuleSource]
```
BardurArantsson: Nice. Maybe even better to have a newtype, perhaps in a follow-up to make both types newtypes? I realize that such newtypes aren't exactly idiomatic in Cabal :).

Cabal/Distribution/Backpack/UnifyM.hs
```diff
@@ -108,31 +139,67 @@ instance Applicative (UnifyM s) where
...
-          Left err -> return (Left err)
-          Right f'' -> do
+          Nothing -> return Nothing
+          Just f'' -> do
               x' <- x r
               case x' of
-                  Left err -> return (Left err)
-                  Right x'' -> return (Right (f'' x''))
+                  Nothing -> return Nothing
+                  Just x'' -> return (Just (f'' x''))
```
BardurArantsson: Isn't this `case` just a liftM/fmap?

Cabal/Distribution/Backpack/UnifyM.hs
```diff
@@ -108,31 +139,67 @@ instance Applicative (UnifyM s) where
...
 
 instance Monad (UnifyM s) where
     return = pure
     UnifyM m >>= f = UnifyM $ \r -> do
         x <- m r
         case x of
-            Left err -> return (Left err)
-            Right x' -> unUnifyM (f x') r
+            Nothing -> return Nothing
+            Just x' -> unUnifyM (f x') r
```
BardurArantsson: ditto

Cabal/Distribution/Backpack/UnifyM.hs
```diff
@@ -108,31 +139,67 @@ instance Applicative (UnifyM s) where
...
 liftST :: ST s a -> UnifyM s a
-liftST m = UnifyM $ \_ -> fmap Right m
+liftST m = UnifyM $ \_ -> fmap Just m
+
+addErr :: MsgDoc -> UnifyM s ()
+addErr msg = do
+    env <- getUnifEnv
+    -- TODO: Perhaps it would be wise to defer running these
+    -- actions until the very end?
+    -- ctx <- sequence (unify_ctx env)
```
BardurArantsson: The comment seems to contradict the fact that the next line is commented out.

Cabal/Distribution/Backpack/UnifyM.hs
```diff
@@ -318,7 +391,9 @@ convertUnitIdU' stk uid_u = do
         UnitIdThunkU uid -> return (DefiniteUnitId uid)
         UnitIdU u cid insts_u ->
             case lookupMooEnv stk u of
-                Just _i -> error "convertUnitIdU': mutual recursion" -- return (UnitIdVar i)
+                Just _i ->
+                    failWith (text "Unsupported mutually recursive unit identifier")
+                    -- return (UnitIdVar i)
```
BardurArantsson: Remove the commented bit?

Cabal/Distribution/Backpack/UnifyM.hs
```diff
@@ -390,16 +481,29 @@ convertInclude (ComponentInclude {
...
-                      case req_rns of
-                        DefaultRenaming -> return []
-                        -- Not valid here, but whatever
-                        HidingRenaming _ -> error "Cannot use hiding in requirement renaming"
-                        ModuleRenaming rns -> return rns
+
+    req_rename_list <-
+      case req_rns of
+        DefaultRenaming -> return []
+        -- Not valid here, but whatever
```
BardurArantsson: Move comment to next line just to make it absolutely clear which case it applies to?

Cabal/Distribution/Backpack/DescribeUnitId.hs
ezyang: It's not really much of a naming convention. In this case, the prime means, "and takes more arguments".

Cabal/Distribution/Simple/Haddock.hs
ezyang: It's saying that we shouldn't print this message if no Haddocking will occur. Previously, the setup message was unconditionally printed. Now we put the action in smsg and call it inside the cases.

Cabal/Distribution/Utils/LogProgress.hs
ezyang: Cabal doesn't depend on transformers (yet). When parsec becomes default it will but I don't think that GHC's build system has been adjusted to handle this yet.

Cabal/Distribution/Backpack/LinkedComponent.hs
ezyang: I guess it is, but for a reason more subtle than reqs including the 'sigantures' field (in the old code, it was mixed in). I'll mention it.

Cabal/Distribution/Backpack/LinkedComponent.hs
ezyang: This one isn't a change in behavior: previously the local requirements were linked in during mixin linking so they were already in linked_shape0 ;)

Cabal/Distribution/Backpack/ModuleScope.hs
ezyang: Perhaps, but in a later patch :) (It's possible that the punning helps save a bit of code.)

Cabal/Distribution/Backpack/UnifyM.hs
ezyang: Unfortunately not, because it's IO (Maybe a) so an fmap will use the IO instance, not the MaybeT IO instance. But we can't depend on MaybeT, so it's done by hand...

Cabal/Distribution/Backpack/UnifyM.hs
ezyang: Same deal.

Cabal/Distribution/Backpack/UnifyM.hs
ezyang: Yes, this comment is stale, I'll remove it.

Cabal/Distribution/Backpack/UnifyM.hs
ezyang: Well, it's what you would replace the line with if you added support for mutual recursion :)

Cabal/Distribution/Backpack/UnifyM.hs
ezyang: OK.

ezyang:
Repushed with updates for comments.

ezyang:
Because this commit is so huge, I'm going to go ahead and merge it. But please feel free to add more comments, I will address them ASAP.

23Skidoo:
OK, sure. I'll try to take a look today and merge it into 2.0.

23Skidoo:
Oh, I see you already did it. Good.

PR #4338, opened 2017-02-18
Kill bad transformers version harder.

ezyang:


ezyang:
This patch doesn't work because the SetupCustomExplicitDeps test in integration-tests2 tries to build an older version of Cabal which, apparently, depends on transformers. Maybe it will be easiest just to wait for Hackage bounds to be fixed correctly.

ezyang:
I implemented `cabal update --index-state` to stab transformers with more prejudice. @hvr you might be interested.

PR #4339, opened 2017-02-18
Update process bounds.

Mistuke:
I'd like to bump the version of `process` in the GHC tree, but need to bump it in Cabal first.

Bumped the rest just to be consistent with versions of `process` used.

https://github.com/haskell/process/blob/master/changelog.md

This is being bumped for a new `process` API in Windows that is off by default.


ezyang:
CI is busted right now so I'll go ahead and merge!

Mistuke:
Thanks!

PR #4341, opened 2017-02-18
Support for --index-state on cabal update

ezyang:
CC @hvr

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -32,6 +32,8 @@ module Distribution.Client.IndexUtils (
   parsePackageIndex,
   updateRepoIndexCache,
   updatePackageIndexCacheFile,
+  writeIndexTimestamp,
+  currentIndexTimestamp,
   readCacheStrict, -- only used by soon-to-be-obsolete sandbox code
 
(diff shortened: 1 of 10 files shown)


cabal-install/Distribution/Client/IndexUtils.hs
```diff
@@ -262,7 +262,7 @@ getSourcePackagesAtIndexState verbosity repoCtxt mb_idxState = do
             return ()
         IndexStateTime ts0 -> do
             when (isiMaxTime isi /= ts0) $
-                warn verbosity ("Requested index-state " ++ display ts0
+                info verbosity ("Requested index-state " ++ display ts0
```
ezyang: So, my mental model of `--index-state` is that you are saying, "Give me the state of the index at time X." In this case, the date I pass in might not correspond to any timestamp explicitly mentioned in the index. But this doesn't feel like a warning case to me: the time of last update might be different from the time I gave, but what's wrong with that?

23Skidoo: OK. I was thinking of the case where the requested timestamp is invalid, e.g. in the future or deep past, so the command doesn't do what the user expects.

ezyang: Yes, I suppose that is something we could warn about, but that's a different warning!

PR #4342, opened 2017-02-19
Allow the solver to toggle manual flags to match constraints that have any qualifier.

### LINKED ISSUE #4299: Cannot find solution with +parsec on Cabal, with cabal-install HEAD
As of a1d9e88d32eaa9b28ed973e3f6dfe3aa74993fda, I can no longer depsolve Cabal successfully with `cabal new-build`, with the `+parsec` flag (i.e. running `cabal new-build --constraint "Cabal +parsec"`). Here's the v3 log: http://lpaste.net/377459545791791104. The error I get in the end is:

```
trying: cabal-testsuite-1.25.0.0 (user goal)
trying: attoparsec-0.13.0.1/installed-13e... (dependency of
cabal-testsuite-1.25.0.0)
next goal: aeson (dependency of cabal-testsuite-1.25.0.0)
rejecting: aeson-1.1.0.0 (conflict: attoparsec =>
scientific==0.3.4.4/installed-ad7..., aeson => scientific>=0.3.4.7 && <0.4)
rejecting: aeson-1.0.2.1, aeson-1.0.2.0, aeson-1.0.1.0, aeson-1.0.0.0,
aeson-0.11.3.0, aeson-0.11.2.1, aeson-0.11.2.0, aeson-0.11.1.4,
aeson-0.11.1.3, aeson-0.11.1.2, aeson-0.11.1.1, aeson-0.11.1.0,
aeson-0.11.0.0, aeson-0.9.0.1, aeson-0.9.0.0, aeson-0.8.1.1, aeson-0.8.1.0,
aeson-0.8.0.2, aeson-0.7.0.6, aeson-0.7.0.4, aeson-0.6.2.1, aeson-0.6.2.0,
aeson-0.6.1.0, aeson-0.6.0.2, aeson-0.6.0.1, aeson-0.6.0.0, aeson-0.5.0.0,
(conflict: cabal-testsuite => aeson==1.1.*)
goals I've had most trouble fulfilling: cabal-testsuite:setup.Cabal,
cabal-testsuite, Cabal, cabal-testsuite:setup.Cabal-1.25.0.0:flag,
Cabal-1.25.0.0:flag
```

CC @grayjay

grayjay:
Three commits:

#### Allow the solver to toggle manual flags to match constraints that have any qualifier.

This fixes #4299. The change gives the dependency solver the flexibility to link
dependencies when the user has only set a manual flag on one of them.
Previously, the solver would force the constrained dependency to have the
flag value from the constraint and force the unconstrained dependency to have
the default flag value. In cases where the single instance restriction required
the dependencies to be linked, the solver couldn't find a solution.

Qualified constraints can still be used to force different dependencies on a
package to use different flag values. For example,
`--constraint 'pkg +flag' --constraint 'pkg2:setup.pkg -flag'` turns the flag on
for the top-level dependency and off for the setup dependency.

I also stored flag default values in the search tree to simplify the code.

#### Add manual flags to the solver DSL.

Tests can now declare flags before using them, in order to specify non-default
values for the fields in `D.Types.GenericPackageDescription.Flag`.

####  Add tests for manual flags.

_______________

@23Skidoo Do you think we could add this to the release branch?  I wanted to fix #4299 because it's a regression.

@ezyang What do you think of this solution?

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Cycles.hs
+++ b/cabal-install/Distribution/Solver/Modular/Cycles.hs
@@ -23,21 +23,21 @@ detectCyclesPhase = cata go
   where
     -- Only check children of choice nodes.
     go :: TreeF d c (Tree d c) -> Tree d c
-    go (PChoiceF qpn rdm gr     cs)                   =
-        PChoice qpn rdm gr     $ fmap (checkChild qpn) cs
-    go (FChoiceF qfn@(FN (PI qpn _) _) rdm gr w m cs) =
-        FChoice qfn rdm gr w m $ fmap (checkChild qpn) cs
-    go (SChoiceF qsn@(SN (PI qpn _) _) rdm gr w   cs) =
-        SChoice qsn rdm gr w   $ fmap (checkChild qpn) cs
-    go x                            = inn x
+    go (PChoiceF qpn rdm gr                         cs) =
+        PChoice qpn rdm gr     $ fmap (checkChild qpn)   cs
+    go (FChoiceF qfn@(FN (PI qpn _) _) rdm gr w m d cs) =
+        FChoice qfn rdm gr w m d $ fmap (checkChild qpn) cs
+    go (SChoiceF qsn@(SN (PI qpn _) _) rdm gr w     cs) =
+        SChoice qsn rdm gr w   $ fmap (checkChild qpn)   cs
+    go x                                                = inn x
 
     checkChild :: QPN -> Tree d c -> Tree d c
-    checkChild qpn x@(PChoice _  rdm _     _) = failIfCycle qpn rdm x
-    checkChild qpn x@(FChoice _  rdm _ _ _ _) = failIfCycle qpn rdm x
-    checkChild qpn x@(SChoice _  rdm _ _   _) = failIfCycle qpn rdm x
-    checkChild qpn x@(GoalChoice rdm       _) = failIfCycle qpn rdm x
-    checkChild _   x@(Fail _ _)               = x
-    checkChild qpn x@(Done       rdm _)       = failIfCycle qpn rdm x
+    checkChild qpn x@(PChoice _  rdm _       _) = failIfCycle qpn rdm x
+    checkChild qpn x@(FChoice _  rdm _ _ _ _ _) = failIfCycle qpn rdm x
+    checkChild qpn x@(SChoice _  rdm _ _     _) = failIfCycle qpn rdm x
+    checkChild qpn x@(GoalChoice rdm         _) = failIfCycle qpn rdm x
+    checkChild _   x@(Fail _ _)                 = x
+    checkChild qpn x@(Done       rdm _)         = failIfCycle qpn rdm x
 
     failIfCycle :: QPN -> RevDepMap -> Tree d c -> Tree d c
     failIfCycle qpn rdm x =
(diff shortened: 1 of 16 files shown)


cabal-install/Distribution/Solver/Modular/Preference.hs
```diff
@@ -228,40 +229,66 @@ enforcePackageConstraints pcs = trav go
...
+-- for a single package (with different qualifiers), and flag constraints for
+-- manual flag x only apply to some of those goals. In that case, we allow the
+-- unconstrained goals to use the default value for x OR any of the values in
+-- the constraints on x (even though the constraints don't apply), in order to
+-- allow the unconstrained goals to be linked to the constrained goals. See
+-- https://github.com/haskell/cabal/issues/4299.
+--
+-- This function does not enforce any of the constraints, since that is done by
+-- 'enforcePackageConstraints'.
+enforceManualFlags :: M.Map PN [LabeledPackageConstraint] -> Tree d c -> Tree d c
```
ezyang: The key change!

23Skidoo:
Sure, I can cherry-pick this into 2.0.

cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
```diff
@@ -166,16 +169,23 @@ data ExampleDependency =
   | ExPkg (ExamplePkgName, ExamplePkgVersion)
   deriving Show
 
+data ExFlag = ExFlag {
```
23Skidoo: A Haddock comment here would be nice.

23Skidoo:
Also cherry-picked into 2.0.

cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL.hs
grayjay: I made a PR: #4345

cabal-install/Distribution/Solver/Modular/Cycles.hs
```diff
@@ -23,21 +23,21 @@ detectCyclesPhase = cata go
...
+    go x                                                = inn x
 
     checkChild :: QPN -> Tree d c -> Tree d c
-    checkChild qpn x@(PChoice _  rdm _     _) = failIfCycle qpn rdm x
-    checkChild qpn x@(FChoice _  rdm _ _ _ _) = failIfCycle qpn rdm x
-    checkChild qpn x@(SChoice _  rdm _ _   _) = failIfCycle qpn rdm x
-    checkChild qpn x@(GoalChoice rdm       _) = failIfCycle qpn rdm x
-    checkChild _   x@(Fail _ _)               = x
-    checkChild qpn x@(Done       rdm _)       = failIfCycle qpn rdm x
+    checkChild qpn x@(PChoice _  rdm _       _) = failIfCycle qpn rdm x
```
grayjay: What type of selector?  A function that returns a `Maybe`?

ezyang: Yep.

ezyang:
Aha! I guess I misunderstood how linking works; I thought it was semantically meaningful, probably due to its use for handling the single instance restriction. So then, would it be accurate to say that if we got rid of the single instance restriction, this patch wouldn't be necessary either?

grayjay:
Yes, that was one of my suggestions in #4299.  Maybe we can remove the single instance restriction on master and then revert this behavior change.  I'm not sure how hard that would be.  A few sections of code assume that there is only one instance of each package, and we might need to change how the solver decides when to build the inplace copies of packages.

ezyang:
Oops, I missed that! Maybe it is worth a comment saying that this change can be reverted if single instance is dropped. If the semantics here cause us trouble in the future, we'll know what to change!

grayjay:
Good idea.  I made a PR: #4354

PR #4346, opened 2017-02-20
Control expensive assertion in Distribution.Solver.Modular.Linking with a build flag.

### LINKED ISSUE #4258: Expensive assertion in Distribution.Solver.Modular.Linking
The assertion at https://github.com/haskell/cabal/blob/6d993e21bb39f6a2b6f22dc5043b4a4db65b4c35/cabal-install/Distribution/Solver/Modular/Linking.hs#L138 makes the solver run slowly when there are many setup dependencies, especially with --reorder-goals.  Solving for jsaddle-warp with new-build is a good example (See https://github.com/haskell/cabal/issues/4192#issuecomment-269940713).  When I ran `cabal new-build jsaddle-warp --dry-run` with GHC 8.0.1, it took almost nine minutes with the assertion and about eight seconds without.

I didn't see any obvious ways to improve the performance.  However, even if we minimized the effect of assertions on runtime, assertions could still affect performance measurements during development.  It's easy to accidentally compare a cabal executable with assertions enabled to one without assertions.

I'm not sure of the best solution.  Here are some ideas:

- Remove `ghc-options: -fno-ignore-asserts` from cabal.project and then enable assertions for unit tests in CI.
- Add another build flag to control just the expensive assertions, and turn them on in CI.  Other assertions would be enabled in cabal.project.
- Print a warning when assertions are enabled.

grayjay:
I added a function, `debugAssert`, that wraps `assert` and only calls it when
the build flag `debug-assertions` is enabled.  The flag defaults to false. I
only replaced one call to `assert` so far (in Distribution.Solver.Modular.Linking)
in order to resolve #4258.
__________________________

I enabled `debug-assertions` in a new GHC 8.0.1 build job, because I figured that it is important to continue testing with the default flag settings in all of the other matrix entries.  I could set `-fdebug-assertions` in one of the existing jobs instead, if it's not worth creating a separate job.

I can also make a PR for the 2.0 branch.

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -189,7 +189,7 @@ Flag network-uri
   description:  Get Network.URI from the network-uri package
   default:      True
 
-Flag debug-assertions
+Flag debug-expensive-assertions
   description:  Enable expensive assertions for testing or debugging
   default:      False
   manual:       True
@@ -414,8 +414,8 @@ library
     else
       build-depends: unix >= 2.5 && < 2.8
 
-    if flag(debug-assertions)
-      cpp-options: -DDEBUG_ASSERTIONS
+    if flag(debug-expensive-assertions)
+      cpp-options: -DDEBUG_EXPENSIVE_ASSERTIONS
 
     if flag(debug-conflict-sets)
       cpp-options: -DDEBUG_CONFLICT_SETS
@@ -506,8 +506,8 @@ executable cabal
         else
           build-depends: unix >= 2.5 && < 2.8
 
-        if flag(debug-assertions)
-          cpp-options: -DDEBUG_ASSERTIONS
+        if flag(debug-expensive-assertions)
+          cpp-options: -DDEBUG_EXPENSIVE_ASSERTIONS
 
         if flag(debug-conflict-sets)
           cpp-options: -DDEBUG_CONFLICT_SETS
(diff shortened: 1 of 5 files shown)


cabal-install/cabal-install.cabal
```diff
@@ -189,6 +189,11 @@ Flag network-uri
   description:  Get Network.URI from the network-uri package
   default:      True
 
+Flag debug-assertions
```
23Skidoo: Bikeshedding: what about introducing a flag `assertions` and call this one `expensive-assertions` to make it more clear what's going on? Then I guess `debugAssert` could be renamed to `expensiveAssert`. What do you think?

23Skidoo: Or, given that we already have `debug-tracetree`, we could also have `debug-assertions` for `ghc-options: -fno-ignore-asserts` and `debug-expensive-assertions` for expensive ones.

grayjay: I like the names `debug-expensive-assertions` and `expensiveAssert`.  I think that it would be simpler to use `ghc-options: -fno-ignore-asserts` to enable the usual assertions in cabal.project, though, because we could set the same flag for all packages: https://github.com/haskell/cabal/blob/9348682e939758a35d11534e6a723b799cbada53/cabal.project#L11

23Skidoo:
Everything looks good to me, merging.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #4347, opened 2017-02-20
Bump upper bound on time

bgamari:
This is necessary for GHC 8.2.


23Skidoo:
> @bgamari changed the base branch to `haskell:2.0` from `haskell:master`

Didn't know one could do that!

bgamari:
Indeed, I think this is a fairly new (and badly needed) feature.

23Skidoo:
I've also updated the upper bound on `process`, since 9d4607835102fb4634bbe1077aa7ad82a79fc276 apparently didn't make it into 2.0.

PR #4351, opened 2017-02-21
Bump upper bound on process

bgamari:
This is necessary for GHC 8.2.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -83,7 +83,7 @@ library
     deepseq    >= 1.3 && < 1.5,
     filepath   >= 1.3 && < 1.5,
     pretty     >= 1.1 && < 1.2,
-    process    >= 1.1.0.1 && < 1.6,
+    process    >= 1.1.0.1 && < 1.7,
     time       >= 1.4 && < 1.9
 
   if flag(old-directory)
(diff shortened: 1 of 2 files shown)


23Skidoo:
You should also update `cabal-install.cabal`.

23Skidoo:
I canceled the CI build because `process-1.6` is not on Hackage.

bgamari:
I didn't bump `cabal-install` since I haven't tested that `cabal-install` actually builds. I think we should keep these two bumps independent.

Also, it would probably be best to hold off a bit on merging this: I'm waiting to hear back from @snoyberg on whether the he intends to bump the `process` version number back down to 1.5 before the next release.

snoyberg:
I definitely don't intend to bump back down to 1.5; though the latest change to the process package was to an internal module only, it's still a breaking change, so I'm going to be proceeding with version 1.6.

23Skidoo:
@bgamari 
> I didn't bump `cabal-install` since I haven't tested that `cabal-install` actually builds.

If it doesn't build with this change, CI will catch it. But I'd wait until `process-1.6` is on Hackage.

snoyberg:
It's on Hackage now.

bgamari:
Can we merge this please?

23Skidoo:
Can we also bump `cabal-install`'s dependency on `process` at the same time? From what I remember of looking at the changes in `process-1.6`, it should work. If it doesn't, CI should catch any failures.

bgamari:
I've updated the `cabal-install` dependency and it appears to pass CI.

PR #4352, opened 2017-02-21
Fix use of deprecated functions from Data.Map

bgamari:
This removes on last use of `Data.Map`, which has been long deprecated.

### DIFF
--- a/Cabal/Distribution/Types/DependencyMap.hs
+++ b/Cabal/Distribution/Types/DependencyMap.hs
@@ -12,7 +12,7 @@ import Distribution.Types.Dependency
 import Distribution.Version
 import Distribution.Package
 
-import qualified Data.Map as Map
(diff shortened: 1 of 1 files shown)


bgamari:
This is necessary for GHC to bootstrap the `2.0` branch.

23Skidoo:
Not sure that this will compile on all versions of GHC we want to support, you may need CPP.

ezyang:
I went ahead and merged this; if it breaks master we'll try and fix it.

bgamari:
Thanks!

23Skidoo:
This broke the build on GHC 7.4: https://travis-ci.org/haskell/cabal/builds/203966869

Also needs to be backported into `master`.

23Skidoo:
> Also needs to be backported into `master`.

Se #4355.

PR #4364, opened 2017-03-03
Break up Distribution.Package -- only classes and reexports remain

Ericson2314:
Pretty unobjectionable cleanup :).

I think this will be used by #4265, in which case we should backport this to 2.0 too :/.


23Skidoo:
OK. Cherry-picked into `2.0`.

Ericson2314:
Thanks @23Skidoo!

PR #4366, opened 2017-03-03
Test suite porting megapatch

ezyang:
This patch set does a number of things:

* Ports the rest of integration-tests tests into the new cabal-testsuite format, finally deleting integration-tests!
* Implements some new features that were necessary for this (notably, withSourceCopy, which makes a duplicate of the source before running the tests)
* Fixes a bunch of bugs (getIPID choking when ghc-pkg has a warning, PATH_MAX problems on Windows, parallelism bug, Appveyor losing stderr logs)
* Fixes a registration bug (detailed below)

----

I noticed this bug when porting a (dead code) test case
for new-build on executables to the new test suite; I
noticed that if you do a fresh new-build on a project
with lib and executable, specifying to only build the
executable, AND the executable doesn't depend on the lib,
new-build incorrectly decides to try to register the lib
even though no lib was actually built.

This is because "needs registration" was computed pre-pruning.
I turned elabNeedsRegistration into a function that looks
at the post-pruning elabBuildTargets to determine if we
are actually going to build a library or not.


ezyang:
Merged.

PR #4373, opened 2017-03-04
Clean up PreExistingComponent

Ericson2314:
It no longer hacks around build-depends: that code has been moved to Backpack.Configure, and will be removed altogether with LibDependency

@ezyang I suppose as the foundation this should be reviewed carefully?

### DIFF
--- a/Cabal/Distribution/Backpack/ConfiguredComponent.hs
+++ b/Cabal/Distribution/Backpack/ConfiguredComponent.hs
@@ -25,6 +25,7 @@ import Distribution.Types.ComponentId
 import Distribution.Types.PackageId
 import Distribution.Types.PackageName
 import Distribution.Types.Mixin
+import Distribution.Types.ComponentName
 import Distribution.Types.UnqualComponentName
 import Distribution.Types.ComponentInclude
 import Distribution.Package
@@ -85,16 +86,16 @@ dispConfiguredComponent cc =
 -- and library/executable dependencies are known.  The primary
 -- work this does is handling implicit @backpack-include@ fields.
 mkConfiguredComponent
-    :: PackageId
+    :: PackageDescription
     -> ComponentId
-    -> [(PackageName, (ComponentId, PackageId))]
+    -> [((PackageName, ComponentName), (ComponentId, PackageId))]
     -> [ComponentId]
     -> Component
     -> ConfiguredComponent
-mkConfiguredComponent this_pid this_cid lib_deps exe_deps component =
+mkConfiguredComponent pkg_decr this_cid lib_deps exe_deps component =
     ConfiguredComponent {
         cc_cid = this_cid,
-        cc_pkgid = this_pid,
+        cc_pkgid = package pkg_decr,
         cc_component = component,
         cc_public = is_public,
         cc_internal_build_tools = exe_deps,
@@ -108,7 +109,7 @@ mkConfiguredComponent this_pid this_cid lib_deps exe_deps component =
     -- from @lib_deps@.
     explicit_includes
         = [ let (cid, pid) =
-                    case Map.lookup name deps_map of
+                    case Map.lookup (fixFakePkgName pkg_decr name) deps_map of
                         Nothing ->
                             error $ "Mix-in refers to non-existent package " ++ display name ++
                                     " (did you forget to add the package to build-depends?)"
@@ -128,7 +129,7 @@ mkConfiguredComponent this_pid this_cid lib_deps exe_deps component =
     -- @backpack-include@ is converted into an "implicit" include.
     used_explicitly = Set.fromList (map ci_id explicit_includes)
     implicit_includes
-        = map (\(pn, (cid, pid)) -> ComponentInclude {
+        = map (\((pn, _), (cid, pid)) -> ComponentInclude {
                                 ci_id = cid,
                                 -- See above ci_pkgid
                                 ci_pkgid = pid { pkgName = pn },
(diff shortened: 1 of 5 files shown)


Cabal/Distribution/Backpack/ConfiguredComponent.hs
```diff
@@ -189,23 +198,23 @@ toConfiguredComponent'
...
+                (lib_map, exe_map) component
     in if use_external_internal_deps
         then cc { cc_public = True }
         else cc
   where
     this_cid = computeComponentId deterministic ipid_flag cid_flag (package pkg_descr)
                 (componentName component) (Just (deps, flags))
-    deps = [ cid | (cid, _) <- Map.elems external_lib_map ]
+    deps = [ cid | ((dep_pn, _), (cid, _)) <- Map.toList lib_map
+                 , dep_pn /= packageName pkg_descr ]
```
ezyang: The old code probably wasn't right (if I'm in per-component mode, I should hash over the "internal" deps I feed in), but I don't think it's too bad for this code to be wrong because cabal-install doesn't use it anyway.

Cabal/Distribution/Backpack/ConfiguredComponent.hs
```diff
@@ -215,14 +224,7 @@ extendConfiguredComponentMap cc (lib_map, exe_map) =
...
     lib_map'
-      = case cc_name cc of
-          CLibName ->
-            Map.insert (pkgName (cc_pkgid cc))
-                       (cc_cid cc, cc_pkgid cc) lib_map
-          CSubLibName str ->
-            Map.insert (unqualComponentNameToPackageName str)
-                       (cc_cid cc, cc_pkgid cc) lib_map
-          _ -> lib_map
+      = Map.insert (pkgName (cc_pkgid cc), cc_name cc) (cc_cid cc, cc_pkgid cc) lib_map
```
ezyang: Yay!

Cabal/Distribution/Backpack/ConfiguredComponent.hs
```diff
@@ -267,3 +269,12 @@ newPackageDepsBehaviourMinVersion = mkVersion [1,7,1]
 newPackageDepsBehaviour :: PackageDescription -> Bool
 newPackageDepsBehaviour pkg =
    specVersion pkg >= newPackageDepsBehaviourMinVersion
+
+fixFakePkgName :: PackageDescription -> PackageName -> (PackageName, ComponentName)
```
ezyang: I had to read the call sites to figure out what this was doing. I guess it's going away soon though...

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -1208,21 +1208,24 @@ selectDependency pkgid internalIndex installedIndex requiredDepsMap
...
+        -- If we know the exact pkg to use, then use it.
+        Just pkginstance -> Right pkginstance
+        -- Otherwise we just pick an arbitrary instance of the latest version.
+        Nothing -> case PackageIndex.lookupDependency installedIndex dep' of
+          []   -> Left $ case is_internal of
+            Just cname -> DependencyMissingInternal dep_pkgname
+                   (computeCompatPackageName (packageName pkgid) cname)
+            Nothing -> DependencyNotExists dep_pkgname
+          pkgs -> Right $ head $ snd $ last pkgs
+      -- Fix metadata that may be stripped by old ghc-pkg
```
ezyang: Hmmm... do you rely on this? I'm worried that not all instances of the IPI floating around will be filled in correctly.

Cabal/Distribution/Backpack/ConfiguredComponent.hs
Ericson2314: Well, should we fix it now that we are changing it anyways? I can make it a separate commit in this PR.

Cabal/Distribution/Backpack/ConfiguredComponent.hs
Ericson2314: Code taken from you of course :).

Cabal/Distribution/Backpack/ConfiguredComponent.hs
Ericson2314: I'll add a comment while I change things

Cabal/Distribution/Simple/Configure.hs
Ericson2314: I do rely on it, but fairly directly. (These, and these alone, get "immediately" turned into `PreConfiguredComponent`s.) Perhaps we should live with it for now? Down the road perhaps I can convert them even more immediately.

Ericson2314: Ah yes, let's just make this return `PreExistingComponent` in the external case. That will clear things up.

Cabal/Distribution/Backpack/ConfiguredComponent.hs
ezyang: I don't think it matters.

Cabal/Distribution/Simple/Configure.hs
ezyang: OK Good :)

Ericson2314:
What do I need to do to cherry-pick this for 2.0?

ezyang:
Just checkout 2.0, cherry-pick, push. If you want to be prim and proper, PR first.

23Skidoo:
> Just checkout 2.0, cherry-pick, push. 

Um, it's okay in this case, but in general I'd prefer not to encourage this behaviour.

PR #4382, opened 2017-03-08
Distinguish between true package names, and munged package names

Ericson2314:
There should be little-to-no functional changes with this commit


cabal-install/Distribution/Client/Types.hs
```diff
@@ -151,6 +162,10 @@ instance Package ConfiguredId where
 instance Package (ConfiguredPackage loc) where
   packageId cpkg = packageId (confPkgSource cpkg)
 
+instance HasMungedId (ConfiguredPackage loc) where
```
ezyang: Has munged id... of what?

Cabal/Distribution/InstalledPackageInfo.hs
```diff
@@ -73,17 +75,19 @@ import Data.Set (Set)
...
-        -- of the package associated with this library, and most tooling
-        -- assumes that this field uniquely identifies any package
-        -- that a user might interact with in a single GHC session.
-        -- However, with convenience libraries, it's possible for there
-        -- to be multiple libraries associated with a package ID; to
-        -- keep backwards compatibility with old versions of GHC,
-        -- 'sourcePackageId' actually stores a *munged* version of the
-        -- package identifier that also incorporates the component name.
-        -- The /real/ package name is stored in 'sourcePackageName'.
-        sourcePackageId   :: PackageId,
```
ezyang: At the very least, we need a compatibility definition of sourcePackageId from this module. You can DEPRECATE it if you like.

Cabal/Distribution/InstalledPackageInfo.hs
```diff
@@ -188,7 +199,7 @@ instance IsNode InstalledPackageInfo where
 emptyInstalledPackageInfo :: InstalledPackageInfo
 emptyInstalledPackageInfo
    = InstalledPackageInfo {
-        sourcePackageId   = PackageIdentifier (mkPackageName "") nullVersion,
+        sourceMungedId   = MungedId (mkMungedName "") nullVersion,
```
ezyang: Isn't MungedId abstract?

Cabal/Distribution/Simple/Build/Macros.hs
```diff
@@ -73,7 +76,8 @@ generate :: PackageDescription -> LocalBuildInfo -> ComponentLocalBuildInfo -> S
 generate pkg_descr lbi clbi =
   "/* DO NOT EDIT: This file is automatically generated by Cabal */\n\n" ++
   generatePackageVersionMacros
-    (package pkg_descr : map snd (componentPackageDeps clbi)) ++
+    (computeCompatPackageId (package pkg_descr) CLibName
```
ezyang: This is a no-op right, because CLibName always is the real package name.

Cabal/Distribution/Simple/GHC/IPIConvert.hs
```diff
@@ -9,31 +9,31 @@
...
 import qualified Distribution.License as Current
 
 import Distribution.Version
 import Distribution.ModuleName
 import Distribution.Text
 
-data PackageIdentifier = PackageIdentifier {
-    pkgName    :: String,
-    pkgVersion :: Version
+data MungedId = MungedId {
```
ezyang: This isn't right. The reason why PackageIdentifier is redefined here is because the conversion function needs the Read instance (see below) to be compatible with the output of a Show instance from an old version of Cabal. If you change the data structure as you've done here, you break the Read instance!

Cabal/Distribution/Types/MungedId.hs
```diff
@@ -0,0 +1,53 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+module Distribution.Types.MungedId
+  ( MungedId(..)
```
ezyang: Let's make this abstract please.

ezyang:
There are also general BC considerations which I've mentioned to @Ericson2314 on IRC.

cabal-install/Distribution/Client/Types.hs
Ericson2314: Should I everywhere rename `MungedId` to `MungedPackageId`? I don't think this is any more or less clear than the type itself.

Cabal/Distribution/Backpack/PreExistingComponent.hs
```diff
@@ -46,9 +48,10 @@ ipiToPreExistingComponent ipi =
     PreExistingComponent {
         pc_pkgname = case Installed.sourcePackageName ipi of
             Just n -> n
-            Nothing -> pkgName $ Installed.sourcePackageId ipi,
+            Nothing -> mkPackageName $ unMungedName
+              $ mungedName $ Installed.sourceMungedId ipi,
```
Ericson2314: Heh sure

Cabal/Distribution/InstalledPackageInfo.hs
Ericson2314: Ah, that would solve most problems. Should it recover original name or just be straight synonym?

Cabal/Distribution/Types/MungedId.hs
Ericson2314: Oh oops that was never intended.

cabal-install/Distribution/Client/Types.hs
ezyang: At your discretion!

Cabal/Distribution/InstalledPackageInfo.hs
ezyang: Well, it depends on whether or not sourcePackageId returns a PackageId or a MungedId ;)

Cabal/Distribution/Simple/Build/Macros.hs
Ericson2314: Yeah, it's a no-op. I like to convert in this way this as a warning sign that the code might not be very component-aware.

Cabal/Distribution/Simple/GHC/IPIConvert.hs
Ericson2314: Can we change the type name as long as the constructor stays the same? I'm trying that.

Ericson2314:
Ok fixed everything but the opacity. Whatever is done there I think is best to do in a second commit (though in this PR is fine.)

To put some things into context, I just thought of an example why we hypothetically might want to change `PackageId` and `MungedId` in practice: to disambiguate between packages' origin in the context of new-build projects using local and git overrides. It's unclear how these other package sources interact with qualified goals---perhaps packages with the same name and version from different origins can exist in the same plan.

I guess this would be an argument for opacity if we can default new fields in a back-compat way (for construction). Perhaps they *both* should be opaque? The consistency would make me feel better about it, I think.

Cabal/Distribution/Backpack/PreExistingComponent.hs
Ericson2314: Well hmm, I moved the functionality to `Distribution.InstalledPackageInfo`, and give a it a quick comment, but perhaps the fact that `ghc-pkg` can strip this metadata so care must be taken should also be mentioned.

Cabal/Distribution/Simple/GHC/IPIConvert.hs
ezyang: Not only does the constructor have to be, but each record selector has to be the same.

```
ezyang@sabre:~$ runghc T.hs
MkT {f = 2}
ezyang@sabre:~$ cat T.hs
data T = MkT {f :: Int }
    deriving (Show)

main = print (MkT 2)
```

ezyang:
> I guess this would be an argument for opacity if we can default new fields in a back-compat way (for construction).

It depends on how you setup the smart constructor. One reasonably forwards compatible scheme is to have the constructor take package name, library name, and version number. Then you can switch representation between (MungedName, Version) versus (PackageName, LibraryName, Version).

> Perhaps they both should be opaque? The consistency would make me feel better about it, I think.

The difference is making PackageId opaque is a (gasp) BC-breaking change ;)

Issue #4203, opened 2017-01-03
Don't implement allow-newer/allow-older by modifying PackageDescription


ezyang:
It seems cleaner to hook it straight into the dep solver, rather than try to ninja it in by package description modification. If you do it by modification, you have to worry about making sure the "correct" package description is used anywhere you want to look at deps, and sometimes you get spurious warnings because it looks like you modify build-depends which affects our source level checks, but actually we didn't.

The lack of cleanliness here is hurting me because I want BuildInfo to reflect the written syntax as closely as possible, but I have to keep `targetBuildDepends` around as a real field if people are editing it.

Previously @kosmikus requested this in #3466.

Ericson2314:
This would be in the same spirit as removing cabal-install's other extra bounds which cause https://github.com/haskell/cabal/issues/4169 right?

ezyang:
Yep, I think so, thanks for posting the reference!

23Skidoo:
The problem here is that we won't be able to implement `./Setup.hs configure --alow-newer` in this way, and IIRC that interface is used by Stack (it was added per @snoyberg's request).

hvr:
@23Skidoo relevant: https://github.com/haskell/cabal/issues/3581

23Skidoo:
OK, if Stack doesn't need that interface, then we can remove it.

Ericson2314:
Because of https://github.com/haskell/cabal/pull/4527, this only covers other cases of modifying bounds in situ like #4169, and #4383 if its merged, right?

23Skidoo:
#4527 only removed the `--allow-{newer,older}` command-line options from setup scripts, the main issue this ticket is about is still there.

Ericson2314:
@23Skidoo Ah the code in Cabal is still there and not dead via something else?

23Skidoo:
Yes, it's spread across `lib:Cabal` and `cabal-install`. See [`D.C.Dependency.removeBounds`](https://github.com/haskell/cabal/blob/3401b6ba570a0b92979af95400a1749c74faffe1/cabal-install/Distribution/Client/Dependency.hs#L436), which uses [`D.PD.transformAllBuildDepends`](https://github.com/haskell/cabal/blob/66fafcae6f15abc8efe74c191c3c028aa378ebd2/Cabal/Distribution/PackageDescription/Configuration.hs#L616).

Ericson2314:
Thanks!

PR #4383, opened 2017-03-09
Fix #2066: buildDepends is just an accessor function, not a field

### LINKED ISSUE #2066: PackageDescription.buildDepends should be nuked

Ericson2314:
Prerequisite for #4265


Ericson2314:
@ezyang this and munge (#4382) have a trivial conflict---I made this off master and submitted it now cause I figure it poses fewer problems (BC, opacity), and can get merged first.

I hope by likewise splitting out more pieces of #4265, I can finally get that one working.

ezyang:
I'd like to see other opinions on the approach taken here, which is to replace the `buildDepends` computation with a modification of the `BuildInfo` in each component to incorporate the computed dependency information that otherwise went in `buildDepends` (whatever that is; I'm not exactly sure what is being computed here.)

Some prior discussion on the subject is here: https://github.com/haskell/cabal/pull/3466#issuecomment-230550326 where @kosmikus comments on the fact that allow-older operates by modifying the package description to change the version bounds. In the end we accepted the modification code, because (a) we needed it somehow in the Setup support for allow-older, and (b) it wasn't convenient to modify the solver to do it otherwise. But in the case of `buildDepends` there is a clear alternative, which is to compute the dep info as an auxiliary data structure and then it pass it around where it is needed.

Ericson2314:
What's all this "quick brown fix" stuff in the test. We're not a font renderer!

ezyang:
It's testing that we don't deadlock when the test executable emits a lot of output.

Ericson2314:
@ezyang this is passing except for some appveyor thing which looks unrelated. I agree this is not the most elegant fix, but might be expedient for 2.0?

23Skidoo:
AppVeyor failure doesn't looks spurious to me. Restarted the AppVeyor build.

Ericson2314:
@23Skidoo indeed it doesn't look spurious, but I didn't touch that file or anything related either? More spooky CI!

23Skidoo:
Maybe a dependency changed or some CPP'ed code bitrotted.

ezyang:
Build failure looks triggered by release of optparse-applicative-0.13.2.0, which appears to be BC breaking. If you add an upper bound that should solve the problem.

As for whether to merge, I think I'd prefer this to be a package deal with LibDependency changes

Ericson2314:
Sounds good

Ericson2314:
@ezyang https://github.com/pcapriotti/optparse-applicative/compare/0.13.1...0.13.2 looks back-compat to me, you sure about that?

ezyang:
Yeah, looks like I'm wrong and the actual problem is a missing Data.Semigroup import...

Ericson2314:
@ezyang I feel like the right solution would be `Distribution.Compat.Prelude`, except that's a private module.

ezyang:
What I find really perplexing is that I can't repro your bug. :/

Ericson2314:
To double check, you tried on 7.4? Yeah I'm confused why master and other PRs don't break with this.

Ericson2314:
Perhaps waiting for #4265 was not the right approach as I (with @kmicklas's help) are only getting back to it a year later. I'd like to fix #4203 too, but I don't think this digs that hole much deeper. As soon as there as any mechanism to add/override extra constraints, I'd hope it can easily be used for all such things.

Cabal/Distribution/Types/Component.hs
```diff
@@ -62,6 +63,14 @@ componentBuildInfo :: Component -> BuildInfo
 componentBuildInfo =
   foldComponent libBuildInfo foreignLibBuildInfo buildInfo testBuildInfo benchmarkBuildInfo
 
+lensBuildInfo :: (BuildInfo -> BuildInfo) -> (Component -> Component)
```
Ericson2314: We now depend on some Lenz stuff so maybe this can go elsewhere / use type synonym in the type?

Cabal/Distribution/PackageDescription/Configuration.hs
```diff
@@ -351,18 +351,21 @@ overallDependencies enabled (TargetSet targets) = mconcat depss
...
+      | any ((== n) . fst) comps ->
+        userBug $ "There exist several components with the same name: '" ++ display n ++ "'"
+      | otherwise -> (mb_lib, (n, redoBD lensBuildInfo c) : comps)
+    (PDNull, x) -> x  -- actually this should not happen, but let's be liberal
+    where
+      redoBD :: ((BuildInfo -> BuildInfo) -> (a -> a)) -> (a -> a)
+      redoBD bd_lens = bd_lens $ \bi -> bi { targetBuildDepends = fromDepMap depMap }
+
+      lensLibBD :: (BuildInfo -> BuildInfo) -> (Library -> Library)
+      lensLibBD f = \l -> l { libBuildInfo = f $ libBuildInfo l }
```
phadej: we have that lens. it's `over L.buildInfo`. We also have `targetBuildDepends` (which can apply directly on `Library`):

https://github.com/haskell/cabal/blob/736c6f421488050893eb6efa55c76e7ba17d346e/Cabal/Distribution/Types/BuildInfo/Lens.hs#L182

Cabal/Distribution/Types/Component.hs
phadej: Yes, we have: https://github.com/haskell/cabal/blob/736c6f421488050893eb6efa55c76e7ba17d346e/Cabal/Distribution/Types/Component.hs#L41

Cabal/Distribution/PackageDescription/Configuration.hs
phadej: BTW, it's not `Lens`, it's `Setter` ;)

Ericson2314: 😲😆

Ericson2314:
I can't figure out why the parsing tests would fail for one GHC but not another. But maybe its just spurious? 8.2.2 certainly worked fine locally.

Ericson2314:
Ah, its cause I need to rebase

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1369,7 +1369,7 @@ checkCabalVersion pkg =
       _                   -> False
 
     versionRangeExpressions =
-        [ dep | dep@(Dependency _ vr) <- buildDepends pkg
+        [ dep | dep@(Dependency _ vr) <- allBuildDepends pkg
```
ezyang: This makes my BC senses tingle. What's the BC story for custom setups that call `buildDepends`. Can we preserve the old semantics under the old name?

Cabal/Distribution/Types/PackageDescription.hs
```diff
@@ -398,6 +389,13 @@ enabledBuildInfos pkg enabled =
 -- * Utils
 -- ------------------------------------------------------------
 
+allBuildDepends :: PackageDescription -> [Dependency]
```
ezyang: Need docs, same with the other function below.

Cabal/Distribution/PackageDescription/Configuration.hs
```diff
@@ -517,38 +518,25 @@ flattenPackageDescription
...
+    flattenLib (n, t) = libFillInDefaults $ (fst $ ignoreConditions t)
+      { libName = Just n, libExposed = False }
+    flattenFLib (n, t) = flibFillInDefaults $ (fst $ ignoreConditions t)
+      { foreignLibName = n }
+    flattenExe (n, t) = exeFillInDefaults $ (fst $ ignoreConditions t)
+      { exeName = n }
+    flattenTst (n, t) = testFillInDefaults $ (fst $ ignoreConditions t)
+      { testName = n }
+    flattenBm (n, t) = benchFillInDefaults $ (fst $ ignoreConditions t)
+      { benchmarkName = n }
```
ezyang: What happened in this diff?

Cabal/Distribution/PackageDescription/Configuration.hs
```diff
@@ -351,18 +353,18 @@ overallDependencies enabled (TargetSet targets) = mconcat depss
...
+    (Lib _, (Just _, _)) -> userBug "Only one library expected"
+    (Lib l, (Nothing, comps)) -> (Just $ redoBD l, comps)
+    (SubComp n c, (mb_lib, comps))
+      | any ((== n) . fst) comps ->
+        userBug $ "There exist several components with the same name: '" ++ display n ++ "'"
+      | otherwise -> (mb_lib, (n, redoBD c) : comps)
+    (PDNull, x) -> x  -- actually this should not happen, but let's be liberal
+    where
+      redoBD :: L.HasBuildInfo a => a -> a
+      redoBD = set L.targetBuildDepends $ fromDepMap depMap
```
ezyang: What does this do?

ezyang:
Well, I guess it looks plausible; we'll see how the tests go.

Are you planning to fix the spurious warnings? What's the game plan there?

Cabal/Distribution/PackageDescription/Check.hs
phadej: https://github.com/23Skidoo/all-custom-setups says `buildDepends` it isn't really used anywhere, `pandoc-1.6` is from 2010

```
~/all-custom-setups$ git grep   -e buildDepends origin/all-custom-setups
origin/all-custom-setups:open-pandoc-1.5.1.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.0.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.1.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.6.0.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.6/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
ogre@bigtest:~/all-custom-setups$ git grep   -e buildDepends origin/all-custom-setups
origin/all-custom-setups:open-pandoc-1.5.1.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.0.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.1.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.5/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.6.0.1/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
origin/all-custom-setups:pandoc-1.6/Setup.hs:  let highlightingSupport = any isHighlightingKate $ buildDepends pkg
```

Cabal/Distribution/PackageDescription/Configuration.hs
Ericson2314: It's been a while, but pretty sure this is the problematic bit via your choice a year ago. The difference between the new version and the old version is now we updated the components' `targetBuildDepends` with the `depMaps` of each target, whereas before those just updated the top-level build depends at https://github.com/haskell/cabal/pull/4383/files#diff-12c8f4f2a1b901f8cbc37a619afafcb7L450.

Cabal/Distribution/PackageDescription/Configuration.hs
Ericson2314: I removed the accumulation of deps and just did the flattening of components. Because of the removal of https://github.com/haskell/cabal/pull/4383/files#diff-12c8f4f2a1b901f8cbc37a619afafcb7L520, we no longer need that accumulation.

Cabal/Distribution/PackageDescription/Check.hs
Ericson2314: Yeah per the big comment at https://github.com/haskell/cabal/pull/4383/files#diff-d02dbf8d6911c6afb8c8cb25ebc6ea9cL127, anything that uses it seems bound to be incorrect anyways.

Ericson2314:
@ezyang happy to fix #4203 and the spurious warnings next if that's what you prefer. I don't think it should be too hard?

Cabal/Distribution/Types/PackageDescription.hs
phadej: FYI, in other PR I made `allBuildInfo` return all `BuildInfo`, also non-buildable. I don't know if they should be filtered there.

Ericson2314: `enabledBuildDepends` below does that.

PR #4386, opened 2017-03-10
ProjectPlanning refactor

ezyang:
* toConfiguredComponents and friends are now monadic.  This means
  we can report a user-friendly error when we fail to find a
  dependency in the dependency map.  I had to rejigger a bit
  of the logic in ProjectPlanning since we were knot-tying through
  this function, but it all worked out.  This means
  that unbuildable_external_lib_deps is no more.

* cc_internal_build_tools is no more; instead it's cc_exe_deps,
  which tracks ALL dependencies.  It also comes with a PackageId
  so we can build ConfiguredId cabal-install side.  This change
  propagates all the way to 'ReadyComponent'

* ProjectPlanning: Instead of recomputing dependencies from scratch,
  we instead use the ElaboratedConfiguredPackages we just finished
  making to build the ComponentDeps.

* ProjectPlanning now constructs a skeletal setupComponent.  This
  is used to setup the above with correct setup dependencies.
  In principle this component might also be used for building, but
  lots of functionality isn't written in yet.

* filterExeMapDep is no more; it's all handled by Cabal now.

* The ConfiguredComponentMap now handles both libraries and
  executables in one data structure.  This is nice.

* compSetupDependencies is no more, because elaborated components
  never have custom setup.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning/Types.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning/Types.hs
@@ -406,10 +406,9 @@ elabOrderExeDependencies =
 -- these are the executables we must add to the PATH before we invoke
 -- the setup script.
 elabExeDependencies :: ElaboratedConfiguredPackage -> [ComponentId]
-elabExeDependencies elab =
+elabExeDependencies elab = map confInstId $
     case elabPkgOrComp elab of
-        -- TODO: pkgExeDependencies being ConfiguredId is slightly awkward
-        ElabPackage pkg    -> map confInstId (CD.nonSetupDeps (pkgExeDependencies pkg))
+        ElabPackage pkg    -> CD.nonSetupDeps (pkgExeDependencies pkg)
         ElabComponent comp -> compExeDependencies comp
 
 -- | This returns the paths of all the executables we depend on; we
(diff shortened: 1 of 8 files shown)


cabal-install/Distribution/Client/ProjectPlanning/Types.hs
```diff
@@ -428,8 +427,10 @@ elabExeDependencyPaths elab =
 elabSetupDependencies :: ElaboratedConfiguredPackage -> [ConfiguredId]
 elabSetupDependencies elab =
     case elabPkgOrComp elab of
-        ElabPackage pkg    -> CD.setupDeps (pkgLibDependencies pkg)
-        ElabComponent comp -> compSetupDependencies comp
+        ElabPackage pkg -> CD.setupDeps (pkgLibDependencies pkg)
+        -- TODO: Custom setups not supported for components yet.  When
```
Ericson2314: Well I assume we (will eventually) want them to instead have a special exe dep on a setup component, which in turn track its deps normally? So this isn't commit back?

ezyang: Yes, but I think this chunk of code needs to be rewritten in that regime which is why I put the TODO in.

PR #4389, opened 2017-03-12
Filter out internal deps when forming package ElaboratedConfiguredPackage.

### LINKED ISSUE #4388: Dependency cycle when building cabal master using cabal master
I built cabal from the master branch (40318a77f2bacfc8b82b466ff72d344d954b9602) using cabal 1.24. Now, I am trying to rebuild cabal at the same commit using the previously built cabal. However, I’m seeing the following error:

```
internal error in InstallPlan.fromSolverInstallPlanWithProgress: 
The following packages are involved in a dependency cycle cabal-testsuite-2.1.0.0-inplace
The following packages are involved in a dependency cycle cabal-install-2.1.0.0-inplace

  error, called at ./Distribution/Client/InstallPlan.hs:261:25 in cabal-install-2.1.0.0-inplace:Distribution.Client.InstallPlan
```

ezyang:
Prior to this patch, we were unconditionally sticking all dependencies
in a "whole package' ElaboratedConfiguredPackage, which resulted in
cycles when there were internal dependencies on the library in question,
and broken deps with convenience libraries.  Happily, the fix is simple,
although I did have to add ConfiguredId to the "extra" dependency fields
to make it easier to tell if a dependency was external or internal.

I added an extra test to make sure that we give a good error message
if a package requires per-component builds (apparently, internal
libraries require per-component builds!)

Fixes #4388.


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1227,38 +1227,63 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
...
             dieProgress $
                 hang (text "Dependency cycle between the following components:") 4
                      (vcat (map (text . componentNameStanza) cns))
       where
-        eligible g
+        -- You are eligible to per-component build if this list is empty
+        why_not_per_component g
+            = cuz_custom ++ cuz_spec ++ cuz_length ++ cuz_flag
+          where
+            cuz reason = [text reason]
```
23Skidoo: I like the name :-)

PR #4394, opened 2017-03-13
Don't put VQuiet in VFlags; that's reserved for user-toggles.

### LINKED ISSUE #4393: setup: Can't parse verbosity silent
`cabal-install` seems to pass `-v silent` to Custom Setup.hs too.

ezyang:
Since VQuiet is set programmatically, it's inappropriate
for it to be propagated to the command line.  But unfortunately,
if it was set, we were accidentally using the fancy flag format.
This patch moves VQuiet to its own field in Verbosity so we
don't get confused.

Fixes #4393.


hvr:
@ezyang this reminds me slightly of https://github.com/haskell/cabal/pull/4103#discussion_r87724761

PR #4396, opened 2017-03-14
Fix Backpack tests

ezyang:
I recently was running Backpack tests manually, and noticed that they didn't work! As it turned out, our CI setup for testing with ghc-head had been broken, and so the tests weren't actually being run at all. Ooops.

When I turned the tests back on, I noticed a lot of failures. These were due to #4382, which did not correctly handle picking up internal libraries when `--dependency` is not explicitly specified (the proximate problem was that PackageIndex was indexed on *un*-munged PackageName, but we really do need an index from PackageName + LibName when we're testing if package names are defined or not.) The most expedient thing was to undo the change for now, hoping to merge it back once this problem is fixed. CC @Ericson2314.

There were some miscellaneous small test drift, but the rest of it didn't have any major problems.


ezyang:
Subsumed by #4397

PR #4397, opened 2017-03-14
Resurrect MungedId patch, and a pile of bugfixes

ezyang:
I recently was running Backpack tests manually, and noticed that they didn't work! As it turned out, our CI setup for testing with ghc-head had been broken, and so the tests weren't actually being run at all. Ooops.

When I turned the tests back on, I noticed a lot of failures. These were due to #4382, which did not correctly handle picking up internal libraries when --dependency is not explicitly specified (the proximate problem was that PackageIndex was indexed on un-munged PackageName, but we really do need an index from PackageName + LibName when we're testing if package names are defined or not.) The most expedient thing was to undo the change for now, hoping to merge it back once this problem is fixed.

There were some miscellaneous small test drift, but the rest of it didn't have any major problems.

That was the first five patches. THEN, I resurrected @Ericson2314's patch, with a huge pile of refactoring to make everything work.

The big ideas:

* There's a new AnnotatedId type, which is any identifier (like
  ComponentId), but also with a PackageId and ComponentName.
  It's a bit like ConfiguredId, but ConfiguredId is specialized
  for ComponentId.  There's a conversion function
  annotatedIdToConfiguredId.

* We adopt a totally new strategy for handling MungedPackageId in
  InstalledPackageInfo.  Rather than store the munged package
  identifier in IPI, we NEVER store it, instead computing it
  from the package name and library name (which are stored.)
  There is now code to PARSE a munged package name into these
  two components, so that when we read 'name' we get the
  correct info.  The resulting code is splendidly clear.

* Some places where we took a ComponentName (notably
  computeCompatPackageName) we now take 'Maybe UnqualComponentName'.
  This is more accurate, because compatibility package names are
  only computed for libraries, not other components.  Some code
  in Distribution.Backpack.ReadyComponent is partial now,
  but the invariants should hold up.

* A number of places where MungedId was applied, I undid them.
  Most notable are macro generation (that should be PACKAGES,
  not components) and mkLegacyUnitId (since REALLY old style
  unit identifiers, we won't support internal libraries anyway.)

* Many functions in PackageIndex are monomorphized to
  InstalledPackageInfo.  Fortunately cabal-install's usage
  still works.

* Distribution/Client/SolverPlanIndex.hs, not used by anyone,
  got the axe.

* Dependency solving has a hack to solve the problem of how to
  interpret installed internal libraries in the package database.
  The basic idea is that, although in most cases where we used
  a MungedId, we say it explicitly, in the SOLVER, we munge
  *installed package names* only, and keep the type PackageName.
  It's a hack, but the alternative is to write a lot more code.
  Note that is MORALLY PN is indeed a PackageName, since we really
  are solving for honest to goodness packages, not components!
  See Note [Index conversion with internal libraries] for more
  details.

* ConfiguredId now records the ComponentName.  This is quite important,
  because we need to use it to figure out how we are going to phrase
  --dependency.  In fact, it is a miracle that this worked at all
  (and it only worked because of a very skeevy update to the PackageId
  in the creation of ConfiguredComponent).  Irritatingly, this must
  be a Maybe ComponentName, because a ConfiguredId MIGHT refer to
  a setup component. Very distressing.

There's some final extra bugfixes at the end as I was bludgeoning the test suite into passing.


Cabal/Distribution/Backpack/MixLink.hs
```diff
@@ -35,12 +36,23 @@ mixLink scopes = do
...
- = text "build-depends:" <+> disp (usrc_pkgname src)
+ = text "build-depends:" <+> pp_pn
  | otherwise
- = text "mixins:" <+> disp (usrc_pkgname src) <+> disp (usrc_renaming src)
+ = text "mixins:" <+> pp_pn <+> disp (usrc_renaming src)
+ where
+  pp_pn =
+    -- NB: This syntax isn't quite the source syntax, but it
+    -- should be clear enough.  To do source syntax, we'd
+    -- need to know what the package we're linking is.
```
Ericson2314: This will become the source syntax with my changes, right? :)

Cabal/Distribution/Backpack/ReadyComponent.hs
```diff
@@ -128,16 +134,29 @@ rc_depends rc = ordNub $
...
+    computeCompatPackageId
+        (rc_pkgid rc)
+        (case rc_component rc of
+            CLib lib -> libName lib
+            _ -> error "rc_munged_id: not library")
 
 instance Package ReadyComponent where
     packageId = rc_pkgid
 
-instance HasMungedPackageId ReadyComponent where
```
Ericson2314: Ah good to avoid non-totality in instances 👍

Cabal/Distribution/Simple/PackageIndex.hs
```diff
@@ -138,7 +143,7 @@ data PackageIndex a = PackageIndex {
   --
   -- FIXME: Clarify what "preference order" means. Check that this invariant is
   -- preserved. See #1463 for discussion.
-  packageIdIndex :: !(Map PackageName (Map Version [a]))
+  packageIdIndex :: !(Map (PackageName, Maybe UnqualComponentName) (Map Version [a]))
```
Ericson2314: What about `Map PackageName (Map (Maybe UnqualComponentName) (Map Version [a]))`? This matches what you ended up on for `Backpack.ConfiguredComponent` (and what I did in my branch :D).

Ericson2314: I can push that (if I haven't already) if that would help.

Cabal/Distribution/Simple/PackageIndex.hs
```diff
@@ -435,12 +464,42 @@ lookupPackageName index name =
...
 -- satisfying the version range constraint.
 --
+-- This does NOT work for internal dependencies, DO NOT use this
+-- function on those; use 'lookupInternalDependency' instead.
+--
 -- INVARIANT: List of eligible 'IPI.InstalledPackageInfo' is non-empty.
 --
 lookupDependency :: InstalledPackageIndex -> Dependency
                  -> [(Version, [IPI.InstalledPackageInfo])]
 lookupDependency index (Dependency name versionRange) =
```
Ericson2314: Can't this be a tiny shim around `lookupInternalDependency` with `Nothing`?

Cabal/Distribution/Simple/PackageIndex.hs
ezyang: I think that variant is OK too, but it turned out that there weren't really any cases where you wanted to find all components of a PackageName; and this version required less code changes. Notice that in all the functions below, we generally default to looking for the public library.

Actually, in your other PR (still open, but probably a little obsolete now), you did the Map as `Map PackageName (Map Version (Map (Maybe UnqualComponentName) [a])))`. I tried using *this* version first, but it's actually not the right way to go about doing things, because in a lot of ops you just want the public library...

Maybe a three level map makes sense if we decide to handle the solver more correctly than the hack I have right now, but it still doesn't make too much sense, because you still need to work out the relationships between components with the same package identifier.

Cabal/Distribution/Backpack/MixLink.hs
ezyang: Yep.

Cabal/Distribution/Simple/PackageIndex.hs
ezyang: Yes you're right.

Cabal/Distribution/Simple/PackageIndex.hs
Ericson2314: Oh wait that's what I meant write. My goal was eventually that prefiltering of the index that I mentioned on IRC, but that can be done with this too.

Cabal/Distribution/Simple/PackageIndex.hs
Ericson2314: Also perhaps that can mean using only the internal on in configure to save a little code as I did before. I bring it up because with LibDependency it will work that way anyways.

ezyang: Well, this will all get rewritten when LibDependency lands.

ezyang:
CI failures are legit but I probably won't have time to debug until after my defense.

Ericson2314:
Oh! When is that?

Ericson2314:
Oh, I guess as a maintainer I can push to this?

ezyang:
Thursday! And yes you can!

Ericson2314:
Good luck!

23Skidoo:
@ezyang Good luck with your defense!

cabal-install/Distribution/Solver/Modular/IndexConversion.hs
```diff
@@ -90,6 +96,33 @@ convIP idx ipi =
...
+--      - If we drop the internal packages, now p-0.1 has a
+--        dangling dependency on an "installed" package we know
+--        nothing about. Oops.
+--
+-- An expedient hack is to put p-internal into cabal-install's
+-- index as a MUNGED package name, so that it doesn't conflict
+-- with anyone else (except other instances of itself).  But
+-- yet, we ought NOT to say that PNs in the solver are munged
+-- package names, because they're not; for source packages,
+-- we really will never see source package names.
```
grayjay: I don't really understand this sentence.  Should it say "never see munged source package names"?

ezyang: Yes, it should say munged.

PR #4398, opened 2017-03-15
Use a more common type of conflict in the solver basic memory usage test.

grayjay:
The test previously caused the solver to backtrack because of a missing package.
Missing packages are handled as a special case in the solver, so the test didn't
exercise as much code as it could have.  This commit causes the solver to
backtrack with a version conflict, which is a more common type of failure.
Other tests continue to test the missing package case.


ezyang:
OK LGTM :)

PR #4410, opened 2017-03-20
Miscellaneous improvements needed for hackage-server

ezyang:
* Add Internal versions ShortText/Version so that we can accurately report the heap size of them
* Reexport Dependency from Distribution.Package, as it was exported previously
* Add showVersion function for better BC with Data.Version


Cabal/Distribution/Utils/ShortText/Internal.hs
```diff
@@ -0,0 +1,107 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+
+module Distribution.Utils.ShortText.Internal
```
hvr: Please use the WARNING scheme we use for the other `.Internal`-modules, see `Distribution.Compat.Prelude.Internal`

Cabal/Distribution/Version/Internal.hs
```diff
@@ -0,0 +1,1087 @@
...
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- Exports the 'Version' type along with a parser and pretty printer. A version
+-- is something like @\"1.3.3\"@. It also defines the 'VersionRange' data
+-- types. Version ranges are like @\">= 1.2 && < 2\"@.
+
+module Distribution.Version.Internal (
```
hvr: Please use the WARNING scheme we use for the other `.Internal`-modules, see `Distribution.Compat.Prelude.Internal`

Cabal/Distribution/Utils/ShortText/Internal.hs
ezyang: Let me see if I understand correctly, because `Distribution.Compat.Prelude.Internal` is setup slightly differently than the situation here.

In the Prelude case, we have `Distribution.Compat.Prelude` which is NOT exposed, and then we have `Distribution.Compat.Prelude.Internal` which is exposed, but no one in Cabal library actually uses.

The situation is a little different for `Distribution.Utils.ShortText.Internal`, since it contains the real definitions for the types in question, which then are reexported (with some things hidden) by `Distribution.Utils.ShortText`. So, is the suggestion to have `Distribution.Utils.ShortText` suppress warnings deprecations, or is the suggestion to have a NOT exposed `Distribution.Utils.ShortText.ReallyInternal` with no warnings, `Distribution.Utils.ShortText.Internal` with warnings but not used by anyone, and have `Distribution.Utils.ShortText` import the not exposed things?

ezyang:
This patchset is low priority.

23Skidoo:
@ezyang Can you please update this PR so that we can get it merged.

hvr:
There was also the suggestion discussed on IRC to expose a helper function so we don't have to expose internals at all. The motivation for exposing the internals was iirc so that hackage-server can compute the heap-size, which we can also do w/o /seeing/ the internal representation of e.g. ShortText.

PR #4411, opened 2017-03-20
Add PATH to curl to slightly speed up build.

ezyang:


ezyang:
Huh, I guess it doesn't work. I got this from  http://help.appveyor.com/discussions/problems/6312-curl-command-not-found#comment_42195491 ; maybe someone can tell what I did wrong.

PR #4413, opened 2017-03-21
2.0 service update

ezyang:
This megapushset contains most of the major changes from HEAD, which basically we want all of for 2.0.


23Skidoo:
I'm a bit ill to do a review right now, but please go ahead with this.

Ericson2314:
Whew! Glad you somehow found the time to do this, @ezyang.

PR #4422, opened 2017-03-23
Use default flags rather than mempty.

ezyang:
In particular, the Cmd* modules don't mappend in the defaults
before using the flags, so ensuring they're setup correctly from
the start helps avoid errors.

This default flag business is extremely dodgy and I wish we had a better
story for it.


23Skidoo:
Merged, thanks!

PR #4424, opened 2017-03-26
Tweak target selection semantics for indefinite packages.

ezyang:
Suppose you have a Backpack package foo-indef, which has
some signatures; furthermore, elsewhere in your install plan
ths packages gets instantiated. When you type
'cabal new-build foo-indef', what should get built?

Previously: foo-indef, as well as all of its instantiations,
get built.

Now: only the indefinite foo-indef is typechecked.

This is what you want!


ezyang:
subsumed by #4425

PR #4425, opened 2017-03-26
Some bugfixes I noticed from Backpacking

ezyang:
Three bugs:

* `mkPackageIndex` still wasn't quite right, resulting in assert failures when using internal libraries in some situations. I think I've finally nailed it.
* When I say `cabal new-build foo-indef`, I only want to typecheck an indefinite package, not also build all of its instantiations. This is now done.
* When I say `build-depends: str-sig, str-string`, I'm saying, "please verify str-string matches str-sig." But this check will only be done if we build the instantiated signature package, which we were not doing previously.


ezyang:
This should be backported.

Issue #4439, opened 2017-04-07
cabal does not respect ghc-options for -pgmc, -pgml etc


nh2:
[In this place](https://github.com/haskell/cabal/blob/eebf36b708fd865a7ef0f717b92b7a002fdf6aa1/Cabal/Distribution/Simple/GHC/Internal.hs#L194-L200)

```
               hPutStrLn testchnd "int foo() { return 0; }"
               hClose testchnd; hClose testohnd
               runProgram verbosity ghcProg
                          [ "-hide-all-packages"
                          , "-c", testcfile
                          , "-o", testofile
                          ]
```

cabal calls ghc just like that, without respecting the C compiler and linker that the user has set in `--ghc-options` and `--with-gcc`/`with-ld`.

To reproduce, simply `sudo mv /usr/bin/gcc /usr/bin/gcc.backup` and try to run `cabal configure --with-gcc=clang --with-ld=clang -v3`. You will get:

```
("/home/niklas/bin/ghc",["-c","/tmp/1804289383846930886.c","-o","/tmp/16816927771714636915.o"])
ghc: could not execute: /usr/bin/gcc
```

In this instance, cabal ignored `--with-gcc`.

--

Next, make gcc available again (`sudo mv /usr/bin/gcc.backup /usr/bin/gcc`), complete the `configure`, then move gcc away again and try to compile a

```
library
  hs-source-dirs: src
  exposed-modules: Lib
  build-depends: base
  ghc-options: "-pgmP clang" "-pgmc clang" "-pgma clang" "-pgml clang"
```

You will get when `Linking...` the library:

```
[...]

PR #4441, opened 2017-04-07
Try update-alternatives ld

phadej:


phadej:
Anyone have idea what's happening
```
/usr/bin/ld: --hash-size=31: unknown option
```
?

alexbiehl:
GHC checks `ld` and passes `--hash-size=32` to it if it has `GNU ld` in its version info. 
(Cf. https://github.com/ghc/ghc/commit/71a194d8ca2efd075a5c000be0b378c8706ca0b3#diff-183a20ee6313d2c3600ef96e732176a3R660). `gold` is not passed any additional options.

phadej:
@alexbiehl, ok so it's expected that `gold` doesn't work with GHC pre 7.8.x. Thanks

phadej:
It seems that the test isn't that much faster with gold, @nh2 do you have any ideas why so? Is it related to #4439 ?

ezyang:
We should verify that gold is actually being used for the tests. But perhaps the issue is that gold shines when linking large projects, and the test suite is basically always doing very small things.

23Skidoo:
`-pgml` shouldn't play a role here, since we don't use it.

23Skidoo:
I think I'll merge this, because it's a small change, but feel free to revert if you disagree.

PR #4459, opened 2017-04-24
Test Trusty Travis builds.

ezyang:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 # We specify language: c, so it doesn't default to e.g. ruby
 language: c
 
+dist: trusty
+
 sudo: false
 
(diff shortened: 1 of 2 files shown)


23Skidoo:
@ezyang The build is green, what have we learned here?

ezyang:
That... it works. We can switch over whenever, I think.

23Skidoo:
@ezyang Rationale for switching is a more up-to-date version of `gold`, right?

ezyang:
The immediate reason I tested it out was because Travis sent a message saying that the current default dist is getting EOL'ed by Ubuntu.

23Skidoo:
OK. Anyway, let's merge.

PR #4460, opened 2017-04-24
Output old config before rewriting `cabal.project.local`

### LINKED ISSUE #3943: Save old cabal.project.local when new-configure overwrites it
Something that I think is a little distressing about new-configure is that it blows away the old `cabal.project.local`, even if you've tweaked it by hand and it has some sort of substantial changes. To prevent accidents like this from happening, I think we should save the old local file if we overwrite it. There are two ways to do this: (1) just move it to some backup file e.g. `cabal.project.local-old`, or (2) print out the file contents when we override it.

pranitbauva1997:
The new-configure blows away the old `cabal.project.local`, even if
you've tweaked it by hand and it has some sort of substantial changes.
To prevents accidents like this from happening, we should print the old
local file if we overwrite it.

This fixes #3943 .

### DIFF
--- a/cabal-install/Distribution/Client/CmdConfigure.hs
+++ b/cabal-install/Distribution/Client/CmdConfigure.hs
@@ -5,6 +5,9 @@ module Distribution.Client.CmdConfigure (
     configureAction,
   ) where
 
+import System.Directory
+import Control.Monad
+
 import Distribution.Client.ProjectOrchestration
 import Distribution.Client.ProjectConfig
          ( writeProjectLocalExtraConfig )
@@ -19,7 +22,7 @@ import Distribution.Verbosity
 import Distribution.Simple.Command
          ( CommandUI(..), usageAlternatives )
 import Distribution.Simple.Utils
-         ( wrapText )
+         ( wrapText, notice )
 import qualified Distribution.Client.Setup as Client
 
 configureCommand :: CommandUI (ConfigFlags, ConfigExFlags
@@ -83,7 +86,12 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     -- Write out the @cabal.project.local@ so it gets picked up by the
-    -- planning phase.
+    -- planning phase. If old config exists, then print the contents
+    -- before overwriting
+    exists <- doesFileExist "cabal.project.local"
+    when exists $ do
+        notice verbosity "'cabal.project.local' file already exists. Now overwriting it."
+        copyFile "cabal.project.local" "cabal.project.local~"
     writeProjectLocalExtraConfig (distDirLayout baseCtx)
                                  cliConfig
 
--- a/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/NewConfigure.cabal
+++ b/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/NewConfigure.cabal
@@ -1,0 +1,11 @@
+name:                NewConfigure
+version:             0.1.0.0
+author:              Foo Bar
+maintainer:          cabal-dev@haskell.org
+build-type:          Simple
+cabal-version:       >=1.10
+
+library
+  exposed-modules:   Foo
+  build-depends:     base
+  default-language:  Haskell2010
--- a/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/Setup.hs
+++ b/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/Setup.hs
@@ -1,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
--- a/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.out
+++ b/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.out
@@ -1,0 +1,5 @@
+# cabal new-configure
+'cabal.project.local' file already exists. Now overwriting it.
+Resolving dependencies...
+In order, the following would be built:
+ - NewConfigure-0.1.0.0 (lib) (first run)
--- a/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.project
+++ b/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.project
@@ -1,0 +1,1 @@
+packages: .
--- a/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.test.hs
+++ b/cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.test.hs
@@ -1,0 +1,6 @@
+import Test.Cabal.Prelude
+
+main = cabalTest $
+    withSourceCopy $ do
+        r <- cabal' "new-configure" []
+        assertOutputContains "Now overwriting it" r


cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -83,9 +85,18 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
...
-    -- planning phase.
-    writeProjectLocalExtraConfig (distDirLayout baseCtx)
-                                 cliConfig
+    -- planning phase. If old config exists, then print the contents
+    -- before overwriting
+    exists <- doesFileExist "cabal.project.local"
+    if exists
+      then do
+        contents <- readFile "cabal.project.local"
+        putStr contents
```
ezyang: 1. Why not putStrLn?
2. Maybe worth outputting a header saying what the output is supposed to be?

cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -83,9 +85,18 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+        writeProjectLocalExtraConfig (distDirLayout baseCtx)
+                                      cliConfig
+      else
+        writeProjectLocalExtraConfig (distDirLayout baseCtx)
```
ezyang: Move these out of the conditional so that they are not duplicated. Once you do this, you can use 'when' above to avoid the conditional with empty branch.

cabal-install/Distribution/Client/CmdConfigure.hs
pranitbauva1997: 1. I think the `cabal.project.local` file would already contain a newline character at the end. Thus two of those would be redundant. Although this isn't necessary. I think I should probably use putStrLn

2. Yeah! That would be better.

ezyang: Although most tools encourage a trailing newline in file, it's by no means guaranteed. If the trailing newline is missing, you'll get weird output. Also, you might want the extra newline anyway to make the output look better. Hard to say without seeing the output.

cabal-install/Distribution/Client/CmdConfigure.hs
pranitbauva1997: Yes. This seems better!

pranitbauva1997:
@ezyang Where would be a good place to put the tests in?

ezyang:
Take a look at https://github.com/haskell/cabal/tree/master/cabal-testsuite

pranitbauva1997:
@ezyang Please review. I have added the tests.

cabal-testsuite/PackageTests/NewConfigure/NewConfigure.cabal
```diff
@@ -0,0 +1,12 @@
+name:                NewConfigure
```
ezyang: Could you move this test to a subdirectory of `NewConfigure`? There may be other tests for `new-configure` we will want to add, and if you put it in `NewConfigure` directly you've used up the entire namespace!

cabal-testsuite/PackageTests/NewConfigure/cabal.project.local
```diff
@@ -0,0 +1,26 @@
+verbose: verbose +nowrap +markoutput
```
ezyang: Do you really need this complicated of an initial `cabal.project.local`? Surely one or two items is enough. (There is a reason to keep it small: if any of these parameters change format / are deprecated, then you will need to update this test.)

cabal-testsuite/PackageTests/NewConfigure/setup.test.hs
```diff
@@ -0,0 +1,4 @@
+import Test.Cabal.Prelude
+
+main = setupAndCabalTest $
+    setup "configure" []
```
ezyang: I'm not sure what the point of this test script is; you're trying to test `new-configure`, not regular configure? Also, this test is redundant because we already have tests for configure; no point in making the test suite run longer! (If you delete this, make sure you delete the `out` files for this test as well.)

cabal-testsuite/PackageTests/NewConfigure/cabal.test.hs
```diff
@@ -0,0 +1,5 @@
+import Test.Cabal.Prelude as P
+import qualified Test.Cabal.Prelude as P
+
+main = cabalTest . recordMode DoNotRecord $
```
ezyang: DoNotRecord is bad right? You want to record, and then have an out file showing that when you new-configure, some output is put to stdout.

cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -83,7 +86,14 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     -- Write out the @cabal.project.local@ so it gets picked up by the
-    -- planning phase.
+    -- planning phase. If old config exists, then print the contents
+    -- before overwriting
+    exists <- doesFileExist "cabal.project.local"
+    when exists $ do
+        putStrLn "file: cabal.project.local already exists. Now overwriting it."
```
ezyang: `file:`?

pranitbauva1997: To signify that `cabal.project.local` is a file. I can rewrite it as `"cabal.project.local" file already exists. Now overwriting it.`

cabal-testsuite/PackageTests/NewConfigure/cabal.project.local
pranitbauva1997: Okay. I ran the test, and whatever the output was, I just committed it. I can keep this simple.

cabal-testsuite/PackageTests/NewConfigure/cabal.test.hs
pranitbauva1997: Yeah, my bad.

cabal-testsuite/PackageTests/NewConfigure/setup.test.hs
pranitbauva1997: I will just remove this.

cabal-testsuite/PackageTests/NewConfigure/cabal.project.local
ezyang: Yeah, I don't think we care about what new-configure actually outputs (which, looking here, is quite a lot of things!) We just care that the old contents of cabal.project.local were preserved.

cabal-testsuite/PackageTests/NewConfigure/NewConfigure.cabal
pranitbauva1997: Yes

pranitbauva1997:
@ezyang The output of the file is not showing inside "BEGIN CABAL OUTPUT" thus it is problematic to test. Any which ways I switched to storing the old config in `cabal.project.local~` because that one had more upvotes.

ezyang:
Sweet :)

For future reference, if you `putStrLn`, it will indeed NOT get marked. You have to use one of the info/warning/etc family of functions to have the output show up.

cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -83,7 +86,13 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     -- Write out the @cabal.project.local@ so it gets picked up by the
-    -- planning phase.
+    -- planning phase. If old config exists, then print the contents
+    -- before overwriting
+    exists <- doesFileExist "cabal.project.local"
+    when exists $ do
+        putStrLn "'cabal.project.local' file already exists. Now overwriting it."
```
hvr: As this looks like a logging message (and also the one below), it should go via the logging API, possibly with `notice`-log-level, rather than bypassing the logging layer and write directly to stdout.

pranitbauva1997: Cool. Used this in the current version. Thanks!

pranitbauva1997:
@ezyang @hvr Thanks for all of your help! I think I have now completed work on this (after taking up lots of CI time).

Changes:
 * Store the previous contents in `cabal.project.local~`.
 * Use `withSourceCopy` for `cabal.project.local`.
 * Rephrase the warning
 * Use `notice` to display the warning so that it even shows up in the BEGIN CABAL OUTPUT.
 * Move the test to a sub-directory.
 * Kept `cabal.project.local` simple (actually empty).
 * Remove the redundant setup test.

cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -83,7 +86,13 @@ configureAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     -- Write out the @cabal.project.local@ so it gets picked up by the
-    -- planning phase.
+    -- planning phase. If old config exists, then print the contents
+    -- before overwriting
+    exists <- doesFileExist "cabal.project.local"
+    when exists $ do
+        notice verbosity "'cabal.project.local' file already exists. Now overwriting it."
+        contents <- readFile "cabal.project.local"
```
alexbiehl: We should really be using `copyFile` here. @hvr mentioned possible interference with `LOCALE` for `readFile`/`writeFile`.  http://hackage.haskell.org/package/directory-1.3.1.1/docs/System-Directory.html#v:copyFile

pranitbauva1997: Thanks! I wasn't aware of the `LOCALE` issues with `readFile / writeFile`. I have updated the PR with `copyFile`.

hvr:
@ezyang maybe we should merge this quickly before we find yet another problem with what's essentially a trivial 4-line patch :-)

23Skidoo:
@pranitbauva1997 Thanks for your contribution! In the future, please don't forget to add a changelog note when you're making user-visible changes.

PR #4465, opened 2017-04-27
configure: remove redundancy

pranitbauva1997:
Remove redundant do, bracket and $. Though it is a minor change,
avoiding warnings is always better.

I am not sure whether such minor changes are accepted. I was trying to read and understand the Configure.hs code but these kept coming up on my editor as warnings.


ezyang:
My personal philosophy is not to formatting refactor unless it is getting in the way of a change I want to make, but I think it's pretty legit to make cleanup changes like this when you're trying to understand the code in question.

PR #4468, opened 2017-04-28
Add a release note about Distirbution.Package changes.

### LINKED ISSUE #4467: Removal of Dependency type not mentioned in changelog
It seems that version 2.0 of the Cabal library has removed the `Dependency` type which used to be in `Distribution.Package`, but this change is not noted in the [changelog](https://github.com/haskell/cabal/blob/master/Cabal/changelog).

Would be really nice if this was mentioned in the changelog along with suggestions for working around its removal.

23Skidoo:
Fixes #4467.

[ci skip]


ezyang:
Subsumed by #4469. There shouldn't be any user-visible difference in Distribution.Package post this patch.

PR #4466, opened 2017-04-27
Cherry-pick fixes from master into 2.0

23Skidoo:
@hvr, please ping me if I have forgotten something.

PR #4469, opened 2017-04-28
Re-export Dependency from Distribution.Package.

### LINKED ISSUE #4404: Distribution.Package doesn't export Dependency anymore
That's wrong. It should.

erikd:
Fixes #4404.


erikd:
Related to #4467 .

Ericson2314:
👍Sorry about that breakage occurring in the first place.

23Skidoo:
Also incorporated into #4466/2.0.

PR #4470, opened 2017-04-28
More accurate "registration needed?" calculation for packages.

### LINKED ISSUE #4450: new-build is not registering inplace Custom packages with executables to project-wide database
Steps to reproduce:
1. Create a Custom package with a library and an executable that depends on the library
2. Ask cabal new-build to build the executable

Expected result: the library is installed into dist-newstyle/packagedb

Actual result: it is not

To make matters worse, if you subsequently request the library be built, it still is not registered.

This breaks cabal-testsuite if you "build it incorrectly" at the beginning. I don't know when this regression was introduced, or if it has always been broken.

Appears to affect 2.0 branch.

ezyang:
Previously, we stated that registration was needed if any build target
was a library.  Actually, this is not enough: if we have a build target
on an executable which in turn depends on the library, we must register
it.

To actually get this information, I had to refactor ComponentsGraph to
return the actual *graph*, so I could do a reverse closure on it, giving
me the set of components that actually depend on the library.  That info
then gets plumbed through ElaboratedPackage and then used to determine
if registration is necessary.

Fixes #4450, but it's possible there is another variant of the bug
that occurs if the executable to be built does NOT depend on the
librayr.

### DIFF
--- a/Cabal/Distribution/Backpack/ComponentsGraph.hs
+++ b/Cabal/Distribution/Backpack/ComponentsGraph.hs
@@ -1,8 +1,10 @@
 -- | See <https://github.com/ezyang/ghc-proposals/blob/backpack/proposals/0000-backpack.rst>
 module Distribution.Backpack.ComponentsGraph (
     ComponentsGraph,
-    dispComponentsGraph,
-    toComponentsGraph,
+    ComponentsWithDeps,
+    mkComponentsGraph,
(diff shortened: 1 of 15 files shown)


23Skidoo:
Tests fails with:

> `UNEXPECTED FAIL: PackageTests/NewBuild/T4450/cabal.test.hs`

PR #4480, opened 2017-04-30
Adds cabal doctest

### LINKED ISSUE #2327: doctest integration for Cabal
New summary by @ezyang.

The goal of this task is to add a new command `Setup doctest`, which runs the doctest command on all of the modules of each component in a package, in much the same way `Setup haddock` runs `haddock` on all the modules of each component in a package.

There are a number of steps to implementing this:

1. You have to add the boilerplate for setting doctest up as a new command. Grepping for occurrences of haddock should give you an idea for what you have to edit; the key files are `Cabal/Distribution/Simple/Setup.hs`, adding a case to `defaultMainHelper` in `Cabal/Distribution/Simple.hs`, and a new file to actually contain the logic. I'd recommend against putting too many flags into `DoctestFlags` when starting off.

2. You need to add `doctest` to the program database, so that you can invoke it (and end-users can customize which binary to use, extra options they want to pass, etc.) The relevant logic is in Cabal/Distribution/Simple/Program/Builtin.hs You should create create something like doctestProgram.

3. The overall logic of the driver should be something like this:

```
 withAllComponentsInBuildOrder pkg_descr lbi $ \component clbi -> do
   componentInitialBuildSteps (flag haddockDistPref) pkg_descr lbi clbi verbosity
   preprocessComponent pkg_descr component lbi clbi False verbosity suffixes
   -- Do the actual invocation of doctest
```

   Between component, lbi and clbi, you will have all the information necessary to invoke doctest. You can look at https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Haddock.hs for an example of all of the parts being put together.

4. How should doctest actually be invoked? In theory, doctest accepts flag identical to GHC, so you should be able to reuse Cabal's logic for invoking GHC here. In practice, this is not true, see https://github.com/sol/doctest/issues/155 so you will need to apply some workarounds. It might be good to get doctest to fix these discrepancies to solve problems. Still, it is probably wise to reuse the bulk of Cabal's existing logic in https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/GHC.hs#L509 . You may also want to consult https://github.com/phadej/cabal-doctest/blob/master/src/Distribution/Extra/Doctest.hs for "flags that doctest is known to work with". I don't necessarily recommend actually using the code from this module.

----

Original description below:

@sol suggested this would make more sense here than the doctest repository. 

A `cabal doctest` or some manner of Cabal auto-magic is needed.

Should do discovery and running of doctests automatically.

It's _extremely_ hard for people to get doctests working with a Cabal project. It took be 3 days, 10-12 hours apiece just to get doctests working for [Bloodhound](https://github.com/bitemyapp/bloodhound). Now I realize I'm a fucking moron, but I was also getting help in IRC from people like @ekmett who uses doctests for `lens` and he had no idea what was causing things to break either.

The things that make doctests hard to use fall into three categories:
1. Cabal project/sandbox awareness
2. Custom Setup.hs for doctest discovery and execution
3. Insanely finicky doctest syntax and interactions with Haddock

This notion fixes 1 and 2, but not 3. However, if users have the reasonable belief that 1 and 2 are okay and should be working, they'll spend less time trying to diagnose the wrong things when they've messed up number 3.

I filed this issue long after having gotten Bloodhound's doctest integration working because I ran into somebody I know on IRC struggling for multiple days with this again.

I'm tired of having to tell people to copy-paste what I have in Bloodhound to make it work and people hate having to cargo-cult this anyway. Doctests have caused problems for the [NICTA course](https://github.com/NICTA/course) as well.

angerman:
This adds cabal doctest logic, modelled after cabal haddock.

Resolves #2327


angerman:
Note, that this PR is likely incomplete, and I'd like to get some feedback early on about the code so far.  I'm certain there's likely some logic missing for full doctest support. Please also note that this PR depends on https://github.com/sol/doctest/pull/156

angerman:
However, the current state of affairs is that:
```
distributive-0.5.2 $ ../cabal/dist-newstyle/build/cabal-install-2.1.0.0/build/cabal/cabal doctest --with-doctest=../doctest/dist/build/doctest/doctest
Preprocessing library for distributive-0.5.2..
Examples: 1  Tried: 1  Errors: 0  Failures: 0
```
already works.

angerman:
Pinging @phadej as well, who implemented <https://github.com/phadej/cabal-doctest> and might know the innards of `doctest` wrt cabal much better than I do.

Cabal/Distribution/Simple/Program/Builtin.hs
```diff
@@ -309,6 +311,16 @@ hscolourProgram = (simpleProgram "hscolour") {
         _         -> ""
   }
 
+doctestProgram :: Program
+doctestProgram = (simpleProgram "doctest") {
```
phadej: Ok for now; but we should compile doctest against used ghc (and in fact also haddock)

phadej: I'd leave a todo.

angerman: Could (and more likely should?) cabal-install depend on doctest and haddock then, and use their binaries installed alongside cabal, unless explicitly overwritten?

angerman: On second thought, that might actually not be sufficient, as cabal-install could be compiled by a different ghc than the one it is used with, unless I'm mistaken.

Cabal/Distribution/Simple.hs
```diff
@@ -290,6 +292,21 @@ hscolourAction hooks flags args = do
                  (getBuildConfig hooks verbosity distPref)
                  hooks flags' args
 
+doctestAcion :: UserHooks -> DoctestFlags -> Args -> IO ()
```
ezyang: s/Acion/Action/

Cabal/Distribution/Simple/Program/Builtin.hs
ezyang: @angerman That's correct. In general, we don't have this problem for haddock because it is distributed with GHC, so we can use the same detection code that finds the correct ghc-pkg for a ghc to also get the correct haddock. For now, I'm OK with requiring people to do some footwork to get the correct doctest if they're using a different ghc, but we should test and make sure we get a reasonable error message in this case (similar to what happens when ghc and ghc-pkg versions mismatch.)

ezyang: I think teaching cabal-install how to compile correct versions of haddock and doctest is worthwhile, and fits into the new-build model (we need a simulated build-tools-depends) but it is definitely out of scope for this ticket. Maybe we should make a ticket for it.

Cabal/Distribution/Simple/Program/Builtin.hs
```diff
@@ -309,6 +311,16 @@ hscolourProgram = (simpleProgram "hscolour") {
...
+    programFindLocation = \v p -> findProgramOnSearchPath v p "doctest"
+  , programFindVersion  = findProgramVersion "--version" $ \str ->
+         -- "doctest version 0.11.2"
```
ezyang: Hmm... we kind of really want the GHC version too, so you can check for compatibility. Not sure how to fix the API to make this possible.

cabal-install/main/Main.hs
```diff
@@ -761,6 +763,14 @@ haddockAction haddockFlags extraArgs globalFlags = do
       createTarGzFile dest docDir name
       notice verbosity $ "Documentation tarball created: " ++ dest
 
+doctestAcion :: DoctestFlags -> [String] -> Action
```
ezyang: s/Acion/Action/

Cabal/Distribution/Simple/Doctest.hs
```diff
@@ -49,7 +59,100 @@ doctest :: PackageDescription
...
+           -> DoctestArgs
+           -> (([String],[FilePath]) -> IO a)
+           -> IO a
+renderArgs verbosity args k = do
+  -- inject the "--no-magic" flag, to have a rather bare
+  -- doctest invocation, and disable doctests automagic discovery heuristics.
+  k (["--no-magic"], argTargets args)
+
+-- -----------------------------------------------------------------------------
+-- TODO: move somewhere else (this is copied from Haddock.hs!)
```
ezyang: Yes please do :)

Cabal/Distribution/Simple/Doctest.hs
```diff
@@ -62,36 +72,102 @@ doctest pkg_descr lbi suffixes doctestFlags = do
...
+
+mkDoctestArgs :: Verbosity
+              -> FilePath
+              -> LocalBuildInfo
+              -> ComponentLocalBuildInfo
+              -> [FilePath]
+              -> BuildInfo
+              -> IO DoctestArgs
+mkDoctestArgs verbosity tmp lbi clbi inFiles bi = do
+  let vanillaOpts = (componentGhcOptions normal lbi bi clbi (buildDir lbi))
```
ezyang: Since there is modest duplication here with existing code (which is not easy to remove, so don't) I'd like forward and backward pointers between the code, as a reminder that if someone changes one you might have to change the other.

Cabal/Distribution/Simple/Doctest.hs
```diff
@@ -62,36 +72,102 @@ doctest pkg_descr lbi suffixes doctestFlags = do
...
+        { ghcOptDynLinkMode = toFlag GhcDynamicOnly
+        , ghcOptFPic        = toFlag True
+        , ghcOptHiSuffix    = toFlag "dyn_hi"
+        , ghcOptObjSuffix   = toFlag "dyn_o"
+        , ghcOptExtra       = toNubListR (hcSharedOptions GHC bi)}
+  opts <- if withVanillaLib lbi
+          then return vanillaOpts
+          else if withSharedLib lbi
+          then return sharedOpts
+          else die' verbosity $ "Must have canilla or shared lirbaries "
```
ezyang: s/canilla/vanilla/

angerman:
I'm fine with this a non-hidden command. Someone would need to find it anyway. And more visibility means more bug reports :-)

I'm perfectly fine with `-Wall`. If you want to compile your code, with `-Wall` why shouldn't the tests run with something else? Point being `ghc-options:` could contain arbitrary information. I'm not a fan of some magic argument drops from `ghc-options:`. If this turns out to be a real issue, I'd rather add `cabal doctest [more ghc options]`, so that one would pass `-Wno-xyz` and `-package` there. I just don't know how to grab arbitrary arguments there and just feed them into doctest. Pointers?

We'll have to wait for sol/doctest#156 before merging this though. Especially as the version bump will likely lead to `0.12`, and we'd need to adapt the version here then.

angerman:
We'll now ignore `ghc-options:`, and provide a `doctest-ghc-options:` field for when you need to specify ghc options when running `cabal doctest`. See ekmett/distributive#30.

phadej:
What is the scope of this PR (I.e. what should be possible after this). Some feature creep is already happening there; without discussing the design.

I'm against /just/ introducing `doctest-ghc-options`, the own `doctest` stanza feels to be better long-term goal. Adding fields which are soon removed isn't good for users.

If the lightweight support isn't possible (even simple cases doesn't work out) then lets's merge what we have to do proper RFC. @RyanGlScott will hopefully comment about problems he run into with Emmett packages

RyanGlScott:
I think I would would need to a see a test case to gauge how well I like the proposed implementation of this feature, since I can't tell from the code what this will end up looking like in practice.

Also, I'd prefer the name `doctest-options` (instead of `doctest-ghc-options`) for congruity with `haddock-options`.

angerman:
@phadej, @RyanGlScott: this PR is mostly here to get this ball rolling, #2327 has been started over two years ago, and I believe there needs to be some progress. My motivation might be different from others, as I'm primarily focused on getting rid of `build-type: Custom`, which in my eyes is an escape hatch, and should rarely be needed. It also complicates cross compilation. Of course we can try to teach cabal about different compilers (host, target, ...; c.f. #1493), which will eventually lead to full support for `build-type: Custom`.

In any case, do I believe that it should be as trivial as `cabal doctest` to run doctest over the library(s) and executable(s) defined in the `.cabal` file.

After giving the `doctest-ghc-options:` some thought, I'll scrap them again, in favor of just passing whatever is passed to `cabal doctest` to further down to doctest, which in turn would pass those on to ghci, in addition to the flags that cabal computes for the relevant component. The proper way seems to not pass `ghc-options:` when running `cabal doctest`.

So what would this change for a package?
- No more `build-type: Custom` just to get `doctest`.
- No more custom `Setup.hs`, for anything `doctest` related.
- instead of running `cabal test doctests`, just run `cabal doctest`.

ezyang:
Note that `haddock-options` isn't actually a thing at the Cabal level. It works from command line because haddock is part of the program database, and all programs in the program db get support for `--foo-options` automatically. This means, with the implementation today, `--doctest-options` is already a thing and can let you feed some custom doctest options as necessary.

angerman:
So with the latest changes this is what we get (e.g. for `distributive`):
(these all assume that doctest is installed with PR sol/doctest#156)
```
distributive $ cabal doctest --help
Run doctest tests.

Usage: cabal doctest [FLAGS]

Requires the program doctest, version 0.12.

Flags for doctest:
 -h --help                 Show this help text
 -v --verbose[=n]          Control verbosity (n is 0--3, default verbosity
                           level is 1)
    --builddir=DIR         The directory where Cabal puts generated build files
                           (default dist)
    --with-doctest=PATH    give the path to doctest
    --PROG-option=OPT      give an extra option to PROG (no need to quote
                           options containing spaces)
    --doctest-options=OPTS give extra options to doctest
```
```
distributive $ cabal doctest
Preprocessing library for distributive-0.5.2..
Examples: 1  Tried: 1  Errors: 0  Failures: 0
```
```
distributive $ cabal doctest --doctest-options=-Wall
Preprocessing library for distributive-0.5.2..
### Failure in src/Data/Distributive.hs:84: expression `distribute [(+1),(+2)] 1'
expected: [2,3]
 but got:
          <interactive>:23:1: warning: [-Wtype-defaults]
              • Defaulting the following constraints to type ‘Integer’
                  (Show t0) arising from a use of ‘print’ at <interactive>:23:1-24
                  (Num t0) arising from a use of ‘it’ at <interactive>:23:1-24
              • In a stmt of an interactive GHCi command: print it
          [2,3]
Examples: 1  Tried: 1  Errors: 0  Failures: 1
```

@RyanGlScott would this work as a test case for you?

23Skidoo:
This seems to lack documentation and a changelog note. OK for now, since it's expected to be buggy.

hvr:
@sol, I'm a bit confused, are you saying that the current behaviour doesn't match

http://cabal.readthedocs.io/en/latest/developing-packages.html#buildtoolsmap

(i.e. as tools like `runghc` and `doctest` are not not documented being part of that legacy mapping?)

> The version specified for an unknown build tool is silently ignored

we could warn that a version was specified for unknown `build-tools`; but other than that we have little choice but to ignore it for tools for which we don't know how to infer the tool's version

> Cabal does not error out if an unknown build tool is not on the PATH

if this is in fact the case, this sounds like a bug/regression to me

cc @Ericson2314

sol:
> I'm a bit confused, are you saying that the current behaviour doesn't match
>
> http://cabal.readthedocs.io/en/latest/developing-packages.html#buildtoolsmap

I don't think the documentation is wrong.  As I understand it there is merely some undocumented behavior.

As an example, here are two things that you can't infer from the documentation:

- You can specify a version constraint for `doctest` and `Cabal >= 2` will fail if that version constraint is not met by a pre-installed `doctest` executable
- You can list `ghc` under `build-tools` and `Cabal < 2` will not error out due to an unknown build tool

> we could warn that a version was specified for unknown `build-tools`

I'm still not sure if that is desirable.  As mentioned earlier, I could imagine that the version constraint is used by third party tools (specifically `cabal2nix`) to install system build tools. @peti do you have any input on this?

> we have little choice but to ignore it for tools for which we don't know how to infer the tool's version

Yes, I'm not suggesting to change anything here.

>> Cabal does not error out if an unknown build tool is not on the PATH

> if this is in fact the case, this sounds like a bug/regression to me

I opened https://github.com/haskell/cabal/issues/5464 for that.

peti:
> I could imagine that the version constraint is used by third party tools (specifically cabal2nix) to install system build tools.

We currently don't use that version information in Nix. When we convert a Cabal file into a Nix expression, all dependencies become inputs of the build function and whoever calls that function must provide in whatever versions they deem appropriate. So the build expression itself does not know which version are legal or not; it simply tries to build whatever configuration the upper layer sets up.

Now, this might change at some point in the future, but I don't think it will be any time soon.

Maybe it's a good idea to have Cabal warn about the fact unsupported tools cannot be version-checked Cabal. The problem I see is that `cabal check` warnings are considered errors by many CI setups, so if that warning were implemented we'd essentially force many people to remove the version information, which doesn't feel like a good idea. If you know the version constraints, then by all means put them in the Cabal file! All we need is people to be aware that they aren't enforced (yet) when they do that.

Issue #4295, opened 2017-02-05
custom-setup solving problem, is `Cabal` version shared across all `Setup.hs`?


phadej:
Happens only with GHC 7.8.4 and older.

With the code in https://github.com/phadej/cabal-doctest/commit/742c2d009d33d2f00fd9cb010382c5a397b7f590
Only if I add a `cabal-doctests >=1.0.1` bound to `example.cabal`'s setup-depends, I'll get a solver error https://travis-ci.org/phadej/cabal-doctest/jobs/198593113

This is very unfortunate, as I don't see why there can't be valid plan (`Cabal` for `ComponentSetup` should interfere with `ghc` / `doctest` etc part of the `--enable-tests` plan)

```
example:setup.Cabal-1.2.1, example:setup.Cabal-1.1.6 (constraint from minimum
version of Cabal used by Setup.hs requires >=1.20)
rejecting: example:setup.Cabal-1.24.1.0 (dependencies not linked: cannot make
example:setup.Cabal canonical member of
{*Cabal-1.18.1.5/installed-647...,example:setup.Cabal-1.18.1.5/installed-647...};
conflict set: Cabal, example:setup.Cabal, example:setup.cabal-doctest)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: example:setup.Cabal (6609),
example:setup.cabal-doctest (6567), Cabal (3617), doctest (3451),
example-1:test (3447), example (2798), example:setup.base (1432),

phadej:
cc @grayjay

phadej:
I have no idea where from the `Cabal >= 1.20` bound comes from. Feel free to ask me on IRC for more details.

ezyang:
new-build requires 1.20 or later because 1.18 setup does not support `--dependency`, so it's not possible for us to request specific packages. 80de7ff4fa5cc5c1a5b1e620787750e08f91325b and #4051

grayjay:
[...]

PR #4481, opened 2017-05-01
Uncomment --independent-goals flag.

grayjay:
--independent-goals could be useful for using local packages as setup or
build-tool dependencies in new-build.  See #4295 for an example of a dependency
problem that requires --independent-goals.  The flag is off by default.

I also had to update the constraints that cabal uses to force all build targets to be built with local packages.  They were too restrictive before, because they applied to all top-level dependencies.  Now they only apply to specific build targets.

@dcoutts Is there any problem with exposing this flag in new-build?  I didn't really understand the comments explaining why some of the flags were commented out in new-build, so I didn't know whether they still applied.  I marked them with Github comments.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig/Types.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig/Types.hs
@@ -185,11 +185,11 @@ data ProjectConfigShared
        projectConfigCountConflicts    :: Flag CountConflicts,
        projectConfigStrongFlags       :: Flag StrongFlags,
        projectConfigAllowBootLibInstalls :: Flag AllowBootLibInstalls,
-       projectConfigPerComponent      :: Flag Bool
+       projectConfigPerComponent      :: Flag Bool,
+       projectConfigIndependentGoals  :: Flag IndependentGoals
 
        -- More things that only make sense for manual mode, not --local mode
        -- too much control!
-     --projectConfigIndependentGoals  :: Flag IndependentGoals,
      --projectConfigShadowPkgs        :: Flag Bool,
      --projectConfigReinstall         :: Flag Bool,
      --projectConfigAvoidReinstalls   :: Flag Bool,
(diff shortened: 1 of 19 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -898,8 +898,7 @@ planPackages verbosity comp platform solver SolverSettings{..}
 
         setMaxBackjumps solverSettingMaxBackjumps
 
-        --TODO: [required eventually] should only be configurable for custom installs
-   -- . setIndependentGoals solverSettingIndependentGoals
+      . setIndependentGoals solverSettingIndependentGoals
```
grayjay: here

cabal-install/Distribution/Client/ProjectConfig/Types.hs
```diff
@@ -360,10 +360,10 @@ data SolverSettings
        solverSettingCountConflicts    :: CountConflicts,
        solverSettingStrongFlags       :: StrongFlags,
        solverSettingAllowBootLibInstalls :: AllowBootLibInstalls,
-       solverSettingIndexState        :: Maybe IndexState
+       solverSettingIndexState        :: Maybe IndexState,
+       solverSettingIndependentGoals  :: IndependentGoals
        -- Things that only make sense for manual mode, not --local mode
        -- too much control!
-     --solverSettingIndependentGoals  :: IndependentGoals,
```
grayjay: here

cabal-install/Distribution/Client/ProjectConfig/Types.hs
```diff
@@ -185,11 +185,11 @@ data ProjectConfigShared
        projectConfigCountConflicts    :: Flag CountConflicts,
        projectConfigStrongFlags       :: Flag StrongFlags,
        projectConfigAllowBootLibInstalls :: Flag AllowBootLibInstalls,
-       projectConfigPerComponent      :: Flag Bool
+       projectConfigPerComponent      :: Flag Bool,
+       projectConfigIndependentGoals  :: Flag IndependentGoals
 
        -- More things that only make sense for manual mode, not --local mode
        -- too much control!
-     --projectConfigIndependentGoals  :: Flag IndependentGoals,
```
grayjay: here

ezyang:
Given the thought you've put into this, and the tests, I'm pretty OK with merging this.

23Skidoo:
Merged, thanks!

grayjay:
Thanks!

PR #4482, opened 2017-05-01
Add an integration test for choosing to build the local version of a package.

grayjay:
While I was working on #4481 I noticed that we only test this behavior for internal libraries (PackageTests/BuildDeps/InternalLibrary3/setup.test.hs).


23Skidoo:
Merged, thanks!

PR #4488, opened 2017-05-02
Improve error display in top level handler.

ezyang:
I noticed that in some cases we weren't adding a trailing
newline to error messages arising from "error"; this
patch adds a few newlines where there were none to solve
the problem.

### DIFF
--- a/Cabal/Distribution/Simple/Utils.hs
+++ b/Cabal/Distribution/Simple/Utils.hs
@@ -369,6 +369,8 @@ annotateIO verbosity = modifyIOError f
           . withMetadata NeverMark VerboseTrace verbosity
           $ ioeGetErrorString ioe
 
+
+{-# NOINLINE topHandlerWith #-}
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Simple/Utils.hs
```diff
@@ -369,6 +369,8 @@ annotateIO verbosity = modifyIOError f
           . withMetadata NeverMark VerboseTrace verbosity
           $ ioeGetErrorString ioe
 
+
+{-# NOINLINE topHandlerWith #-}
```
23Skidoo: Why is this needed?

ezyang: It makes GHC recompile less when I change Distribution.Simple.Utils :)

23Skidoo: Maybe add a comment to that effect?

23Skidoo:
Merged, thanks!

PR #4491, opened 2017-05-03
Some linting [WIP]

fgaz:


23Skidoo:
`liftM -> liftA` change seems to cause problems on older GHCs.

fgaz:
> liftM -> liftA change seems to cause problems on older GHCs.

I suspected that, so I wanted to create a pr on my own repo to trigger a private build, but I accidentally selected the wrong owner, hence the added [WIP]. Oh well. I'll proceed here now.

Anyway, seems to be an import problem, but once fixed it'll probably fail again due to AMP.

If that's the case i'll fix some other warnings

fgaz:
I'll close this and work in my own fork for now, but first: why we don't export Distribution.Compat.Prelude so we can use it in the tests too? Is there a way to expose it for the tests only?

23Skidoo:
So that people wouldn't depend on it, I think. I guess it's okay to expose it if you really need it.

23Skidoo:
There's `Distribution.Compat.Prelude.Internal` that you can use.

23Skidoo:
See e.g. Cabal/tests/UnitTests/Distribution/Utils/NubList.hs.

Issue #4462, opened 2017-04-25
Make more efficient use of space in cabal-binaries


ezyang:
Today, I got a message from GitHub saying that cabal-binaries had gotten too big. (For reference, we inaugurated this repo on February 2, so that's about two months of binaries.) To fix the immediate problem, I ran this script to delete every branch besides master from cabal-binaries:

```
git ls-remote origin | cut -f 2 | cut -d "/" -f 3,4 | grep -v master | grep -v HEAD | xargs -n 1 git push --delete origin -v
```

In the future, it would be wise to have cabal-binaries clean up after itself. Probably a good way to do this is to have travis-test.sh DELETE its own branch when it finishes successfully. Failed branches will eventually creep us back up to the limit but then we can just delete everything again.

ezyang:
I received an update that even with nightly old branch pruning, cabal-binaries was taking up too much space. I want to write down some of the updates in developments here.

**Findings of fact.**

* The size of a single build (remember that every PR/branch spins up ~10 builds for various configurations) previously took about 1.2G of space. This is substantially reduced when put into Git; when I pull a single branch from cabal-binaries today, `.git` takes only ~160M of space. When uncompressed, the bulk of space usage is taken up by `dist-newstyle`, which takes a whopping 650M; this is basically all due to object and interface files. We also send over `.cabal`, which accounts for 200M uncompressed.

* The reason why we send all of `dist-newstyle` and `.cabal` is because `cabal-testsuite` needs to be able to run the GHC interpreter on test files which themselves import the Cabal library. Thus, there needs to be interface files for Cabal and all of its deps available for this to work. (Additionally, some Custom build-type tests link against the built version of Cabla and also need the interface files available.) None of the other test suites in Cabal have this requirement.

* The reason why we transmit our build products to another Travis instance is because otherwise, our builds take too long on Mac OS X. It was decided to perform all tests in the second round, because this was the "logical" way to go about doing things.

* GitHub would prefer our repository to stay under 200M. In contrast, GitHub LFS offers 1G storage free, with 1G of bandwidth a month. Amazon AWS offers 5G free storage and 15G free bandwidth a month. So clearly sending 1G, or even 200M of data is completely untenable.

**The current plan.**

[...]

PR #4494, opened 2017-05-04
 Greatly reduce the amount of build product we upload.

ezyang:
Greatly reduce the amount of build product we upload.

See #4462 for the gory details.

Main things about this commit:

- New 'monolithic' flag on cabal-install, which combines
  all of the tests into a single binary.  It's not very
  much code, and you don't pay for any of it on a release
  build.  I quite like it.  The one downside is that
  we can't also pull in Cabal test suites this way.

- Env vars got moved into travis-common.sh

- travis-script.sh now runs the cabal-tests tests, because
  we aren't sending enough build product over to do them
  on the second Travis run

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -130,6 +130,18 @@ flag lib
   Default:      False
   manual:       True
 
+-- When we do CI, we build our binaries on one machine, and then
+-- ship them to another machine for testing.  Because we use
+-- static linking (since it makes this sort of redeploy MUCH
+-- easier), if we build five executables, that means we
+-- need to ship ALL the Haskell libraries five times.  That's
+-- a waste of space!  A better strategy is to statically link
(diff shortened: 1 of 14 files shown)


ezyang:
This is ready for review.

cabal-install/cabal-install.cabal
```diff
@@ -519,7 +563,7 @@ Test-Suite memory-usage-tests
   type: exitcode-stdio-1.0
   main-is: MemoryUsageTests.hs
   hs-source-dirs: tests
-  ghc-options: -Wall -fwarn-tabs "-with-rtsopts=-M4M -K1K"
+  ghc-options: -Wall -fwarn-tabs "-with-rtsopts=-M4M -K1K" -main-is MemoryUsageTests
```
grayjay: It looks like the Travis script no longer uses these RTS options to run the test suite.

PR #4499, opened 2017-05-05
Make cabal --host aware.

angerman:
This allows to invoke cabal with --target=`aarch64-none-linux-android`, and
will result in the `aarch64-none-linux-android-ghc` being picked. Also the
dist dir will be adjusted. And `--cross-compile` be passed to hsc2hs.

This has been tested by configuring, building and installing the distributive package
(with modified `build-type: Simple`) with `--target=aarch64-apple-darwin14`.

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -206,7 +206,7 @@ instance BooleanFlag Bool where
 data GlobalFlags = GlobalFlags {
     globalVersion        :: Flag Bool,
     globalNumericVersion :: Flag Bool
-  } deriving (Generic)
+  } deriving (Generic, Show)
 
 defaultGlobalFlags :: GlobalFlags
 defaultGlobalFlags  = GlobalFlags {
@@ -376,6 +376,7 @@ data ConfigFlags = ConfigFlags {
     configHcFlavor      :: Flag CompilerFlavor, -- ^The \"flavor\" of the
                                                 -- compiler, such as GHC or
                                                 -- JHC.
+    configTarget        :: Flag String,   -- ^the target prefix. (e.g. ghc -> TARGET-ghc)
     configHcPath        :: Flag FilePath, -- ^given compiler location
     configHcPkg         :: Flag FilePath, -- ^given hc-pkg location
     configVanillaLib    :: Flag Bool,     -- ^Enable vanilla library
@@ -458,6 +459,7 @@ instance Eq ConfigFlags where
     && equal configProgramArgs
     && equal configProgramPathExtra
     && equal configHcFlavor
+    && equal configTarget
     && equal configHcPath
     && equal configHcPkg
     && equal configVanillaLib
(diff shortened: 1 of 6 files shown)


angerman:
This is supposed to make obsolete:
- https://github.com/zw3rk/ghc-scripts/blob/master/aarch64-apple-darwin14-cabal
- https://github.com/zw3rk/ghc-scripts/blob/master/common-cross-cabal

ezyang:
This definitely needs docs.

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -376,6 +376,7 @@ data ConfigFlags = ConfigFlags {
     configHcFlavor      :: Flag CompilerFlavor, -- ^The \"flavor\" of the
                                                 -- compiler, such as GHC or
                                                 -- JHC.
+    configTarget        :: Flag String,   -- ^the target prefix. (e.g. ghc -> TARGET-ghc)
```
ezyang: So, it's probably my fault that you edited this file, but given that the actual logic for configTarget lives solely in cabal-install, UNLESS you are planning on teaching Cabal the library how to interpret `--target` (you didn't mention anything about that) the flags don't belong here. Fortunately for you, cabal-install defines its own additional flags which you can pass on top; either `ConfigExFlags`/`BuildExFlags` or perhaps `InstallFlags` would be a better place to put them.

cabal-install/main/Main.hs
```diff
@@ -332,39 +332,74 @@ wrapperAction command verbosityFlag distPrefFlag =
...
+        d' = toFlag ("dist" </> target) <> d
+        c' = ("--host="++target):c
+
 configureAction :: (ConfigFlags, ConfigExFlags)
                 -> [String] -> Action
 configureAction (configFlags, configExFlags) extraArgs globalFlags = do
-  let verbosity = fromFlagOrDefault normal (configVerbosity configFlags)
-  (useSandbox, config) <- updateInstallDirs (configUserInstall configFlags)
+
+  let verbosity = fromFlagOrDefault normal (configVerbosity configFlags')
```
ezyang: NB: This modification (and others) apply to the *legacy* cabal-install codepaths `cabal configure` and `cabal build`. Thus, if you use `cabal new-build`, the modifications you've applied here won't apply at all.

angerman:
I'd rather do it properly than hack it. I simply could not find the right point to make these changes, thus I ended up doing the transformation rather early on. Any guidance would be welcome!

Ericson2314:
@angerman I have some prior thoughts / tiny work on how to do this. We should take when I am back. Also this should be --host I'm pretty sure as it's from the package's, not compiler's, perspective.

angerman:
@Ericson2314 we should! And yes, you are right, this should be `--host`. I would still like this to be done properly in Cabal.

angerman:
With #5018, I'm tempted to try some command line flag sugar for this. Such that
`--host=x` simply desugars into `--with-PROG=x-PROG` and `--cross-compiler` to `hsc2hs-options`, as well as the `--configure` flags.

angerman:
@gbaz I agree with @ezyang that there is a better way to do this. The whole --target (and subsequently prefixed target-triple) is what we need *today* for GHC, as it's simply not multi-target. However I can see this limitation changing sooner rather than later. And at that point this approach (target-triple prefixing) would not make much sense anymore, instead cabal would need to pass --target to the tools. (could be solved by having target-triple shell scripts as stand ins, but still ugly).

For now this whole ordeal of passing the right arguments is mostly solved using `nix`, which I'm personally not very happy with, but it's workable and does a lot of other lifting for toolchains around the compiler.

Thus, my firm believe is we should revisit this once GHC is multi-target capable.

gbaz:
Ok, thanks! As such I'm going to close this PR, and in the future a modified approach can be revisited.

Issue #4496, opened 2017-05-05
cabal init allows a license that cabal install does not understand


Smattr:
**Summary.** For example, a license containing a space:

    $ cabal init
    ...
    Please choose a license:
    ...
      14) Other (specify)
    Your choice? [default: BSD3] 14
    Please specify? hello world
    ...
    $ cabal install
    cabal: tmp.cabal:9: Parse of field 'license' failed.

Despite the promising sounding title, this does not appear to be a duplicate of #3098 that talks about license-file, not license. It is possible this is related to #1074 if `cabal install` is calling `cabal check`.

**Instructions from @ezyang.** First, follow the instructions in the README at https://github.com/haskell/cabal for setting up your dev environment.

As you can see from `instance Parsec License`, the set of permitted characters for a free-form license text is one or more alpha-numeric characters. To fix this bug, you will have to modify `cabal-install/Distribution/Client/Init.hs` to check if the other license has valid syntax and reject it / ask the user for another license if it does not. 

To see an example of an existing prompt which will reprompt if the text is invalid, see `getPackageName`.

If you want to go further beyond this ticket, look at `getLanguage` and make sure it doesn't also suffer from the same error.

A test for this fix is not necessary, but you should run the resulting cabal binary and check that the correct behavior occurs.

If you want a nice follow up task, look at #1074, decide what to do, and fix it!

**What you will learn.** How to test if a string is alphanumeric in Haskell; how the prompt mechanism in `cabal init` works. This is a nice and easy task to get started with the cabal-install build system as well.

ezyang:
Thanks for the very clear report @Smattr. I've turned this into a newcomer task for someone new to Haskell who wants to dip their toe in the Cabal project.

Smattr:
Thanks, @ezyang, and nice write up of an action plan! Gives the reader a very detailed idea of what they're in store for if they take the bait.

2016rshah:
I would like to work on this! I think I know how to do it (thanks to this excellently written issue) but I am having trouble setting up the development environment. I forked the repository and cloned it on my local machine. Then I tried to run `stack build` which failed with 

```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for cabal-install-2.1.0.0(+parsec):
    echo-0.1.2 must match >=0.1.3 && <0.2 (latest applicable is 0.1.3)

In the dependencies for cabal-testsuite-2.1.0.0:
[...]

PR #4504, opened 2017-05-06
checks if license provided is alphanumeric. #4496 WIP

colinwahl:
This PR attempts to solve issue #4496 .  This is my first contribution, and I would love feedback!  It builds locally and seems to fix the problem on my local machine.

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -352,7 +352,20 @@ getLanguage flags = do
                   (Just Haskell2010) display True)
           ?>> return (Just Haskell2010)
 
-  return $ flags { language = maybeToFlag lang }
+  invalidOtherLanguage <- if invalidLanguage lang
+                             then putStrLn invalidOtherLanguageMsg >> return True
+                             else return False
+
+  if invalidOtherLanguage
+    then getLanguage flags
+    else return $ flags { language = maybeToFlag lang }
+
+  where
+    invalidLanguage (Just (UnknownLanguage t)) = any (not . isAlphaNum) t
+    invalidLanguage _ = False
+
+    invalidOtherLanguageMsg = "\nThe language must be alphanumeric. " ++
+                              "Please enter a different language."
 
 -- | Ask whether to generate explanatory comments.
 getGenComments :: InitFlags -> IO InitFlags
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/Init.hs
```diff
@@ -207,8 +207,22 @@ getLicense flags = do
          ?>> fmap (fmap (either UnknownLicense id))
                   (maybePrompt flags
                     (promptList "Please choose a license" listedLicenses (Just BSD3) display True))
-  return $ flags { license = maybeToFlag lic }
+
+  invalidOtherLicense <- if isLicenseInvalid lic
```
ezyang: Stylistic note: You don't need this extra bind: you can just move the `promptInvalidOtherLicenseMsg` printing to the `invalidOtherLicense` test. (P.S. The reason `getPackageName` has this bind is because it actually needs to ask the user to find out if the boolean will be true or false. In your case, it's always True, so it's not necessary.

cabal-install/Distribution/Client/Init.hs
```diff
@@ -207,8 +207,22 @@ getLicense flags = do
...
+
+  if invalidOtherLicense
+    then getLicense flags
+    else return $ flags { license = maybeToFlag lic }
+
   where
+    isLicenseInvalid (Just (UnknownLicense t)) = any (not . isAlphaNum) t
+    isLicenseInvalid _ = False
+
+    promptInvalidOtherLicenseMsg = "\nThe license must be alphanumeric. " ++
```
ezyang: It might be worth noting that if a license name has many words in it, the convention is to camel case it (e.g., PublicDomain.) This gives the user better actionable advice.

cabal-install/Distribution/Client/Init.hs
colinwahl: Makes sense! I changed it.

cabal-install/Distribution/Client/Init.hs
colinwahl: I have updated the error message!

colinwahl:
@ezyang I have made the changes you suggested.  Since you only commented on the `getLicense` changes, I just wanted to double check - are the changes to `getLanguage` also approved?  If so, I will merge once all tests pass.

ezyang:
Great, thank you!

PR #4510, opened 2017-05-07
Improve exe `scope` docs

DanielG:
I clarified that $libexecsubdir is a sub-directory of $libexecdir, added some motivation for the subdir change to the changelog and added an example to the docs.


23Skidoo:
Merged, thanks!

PR #4520, opened 2017-05-16
Remove some redundant LANGUAGE pragmas and set other-extensions

hvr:
This way, cabal's solver detects that GHC 7.4.2 doesn't satisfy
other-extensions and informs us that a compiler is needed which
supports the LambdaCase extension.


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, RecordWildCards, NamedFieldPuns, RankNTypes #-}
+{-# LANGUAGE RecordWildCards, NamedFieldPuns, RankNTypes #-}
 {-# LANGUAGE ViewPatterns #-}
```
BardurArantsson: Given that e.g. ViewPatterns is now listed in other-extensions, shouldn't these also be removed? Or am I missing something?

hvr:
@23Skidoo btw, this PR is against master; but before releasing cabal-install 2.0, you should check that the cabal file excludes older GHC versions which would fail to build it

cabal-install/Distribution/Client/ProjectPlanning.hs
hvr: @23Skidoo that's btw a very common misconception... it's too late now, but I wonder if `depend-extensions` would have been a better name.

23Skidoo:
@hvr Via a minimum bound on `base`?

hvr:
@23Skidoo yeah; that would be the simplest way to achieve this (in a pragmatic way - from a purist POV it's not principled of course :-) ).

hvr:
This exposes a problem in the Travis job entry `GHCVER=7.4.2 SCRIPT=script CABAL_LIB_ONLY=YES TEST_OTHER_VERSIONS=YES`, as `cabal-install` isn't compatible w/ GHC 7.4.2:

```
The command "rm -rf dist-newstyle" exited with 0.
$ ./travis-${SCRIPT}.sh -j1
$ cabal update
Config file path source is default config file.
Config file /home/travis/.cabal/config not found.
Writing default configuration to /home/travis/.cabal/config
cabal update took 5 seconds.
whole job took 5 seconds so far.
----
/opt/happy/1.19.5/bin/happy
$ cabal unpack hackage-repo-tool-0.1.1
Unpacking to hackage-repo-tool-0.1.1/
cabal unpack hackage-repo-tool-0.1.1 took 0 seconds.
whole job took 5 seconds so far.
----
$ cabal new-build Cabal Cabal:unit-tests
Resolving dependencies...
next goal: cabal-install (user goal)
rejecting: cabal-install-2.1.0.0 (conflict: requires LambdaCase)
```

ezyang:
Oh yeah, I guess previously we were saved by the fact that we only ever tried to build the library. But we always solved for cabal-install and so now it is failing.

Easy "fix" is to remove cabal-install directory prior to doing rest of CI.

hvr:
@23Skidoo fyi, http://hackage.haskell.org/package/cabal-install-2.0.0.0/revisions/

23Skidoo:
@hvr Can you also apply that fix to the 2.0 branch?

hvr:
@23Skidoo well, in order to apply this to the 2.0 we'd first have to resolve the test-suite issue that blocked this PR going in... otherwise the testsuite for 2.0 will just start failing :-)

cabal-install/Distribution/Client/ProjectPlanning.hs
23Skidoo: IMO `depend-extensions` would be even more confusing.

23Skidoo:
> Btw, shall we also consider `cabal-version:>=1.18` for `cabal-install.cabal`?

OK, sure, though it'd nice to find out why the setup script segfaults.

23Skidoo:
> well, in order to apply this to the 2.0 we'd first have to resolve the test-suite issue that blocked this PR going in... 

Maybe we can have a separate `cabal.project` file for the library-only configuration?

ezyang:
> OK, sure, though it'd nice to find out why the setup script segfaults.

Segfault?

23Skidoo:
@ezyang @hvr tried to run `cabal install cabal-install-2.0.0.0` with cabal-install 1.16 and it failed to build. Quoting the e-mail @hvr sent me:

>```
> $ PATH=/opt/ghc/7.6.3/bin:$PATH /opt/cabal/1.16/bin/cabal build
> Warning: cabal-install.cabal: Ignoring unknown section type: custom-setup
> [1 of 1] Compiling Main             ( Setup.hs, dist/setup/Main.o )
> Linking ./dist/setup/setup ...
> setup: Run the 'configure' command first.
> ```
>
> But if I change `build-type` to Simple, it works fine.
> 
> Looking deeper, it seems that `./dist/setup/setup configure` just
> crashes (exit-code 11 / SIGSEGV)... and thus no configure state is
> saved; but it's not clear to me why...

gbaz:
for a code cleanup pass, i think this is way too bitrotted to rescue :-/

Mikolaj:
I think we can return in a few ears, set Haskell2021 and use run stylish-haskell to remove all spurious pragmas.

PR #4522, opened 2017-05-17
Filter `+timestamp` out when invoking Cabal-2.0.0 custom-setups

hvr:
This is is a follow-up to #4518

### DIFF
--- a/cabal-install/Distribution/Client/SetupWrapper.hs
+++ b/cabal-install/Distribution/Client/SetupWrapper.hs
@@ -365,7 +365,7 @@ runSetup verbosity setup args0 = do
 -- verbosity applies to ALL commands.
 verbosityHack :: Version -> [String] -> [String]
 verbosityHack ver args0
-    | ver >= mkVersion [1,25] = args0
+    | ver >= mkVersion [2,1]  = args0
     | otherwise = go args0
   where
     go (('-':'v':rest) : args)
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/SetupWrapper.hs
```diff
@@ -365,7 +365,7 @@ runSetup verbosity setup args0 = do
 -- verbosity applies to ALL commands.
 verbosityHack :: Version -> [String] -> [String]
 verbosityHack ver args0
-    | ver >= mkVersion [1,25] = args0
+    | ver >= mkVersion [2,1]  = args0
```
23Skidoo: Don't you also need to update `filterConfigureFlags`, like in eb0593ed7690e402c288011dc77964e404bb78d9?

hvr: @23Skidoo now that you mention it... yes that makes sense... I wonder how I can trigger that code-path... is that old-build?

23Skidoo: FWICT, every time we use `filterConfigureFlags`, we pass the result to the setup wrapper, so maybe the code in eb0593ed7690e402c288011dc77964e404bb78d9 is not strictly needed.

hvr: I asked @ezyang, and he told me

> I know why SetupWrapper has the hack: it's because verbosity flags apply to commands that are not configure
> but filterConfigureFlags is configure only 
> So, I think everything will still work even if you don't filterConfigureFlags. But it's "good style", I suppose

I'll update this PR to have `filterConfigureFlags` follow the good style... :)

hvr: @23Skidoo  Done, `filterConfigureFlags` adapted as per 51d3a2d48ae5b6dda56f3fdd29b37ff4cc6cd1d5 - everytime I have to touch `filterConfigureFlags` I get confused... :-)

PR #4526, opened 2017-05-19
More accurate -Wmissing-home-modules version check.

23Skidoo:
Relevant background on `-Wmissing-home-modules`: https://github.com/haskell/cabal/issues/1746#issuecomment-272691469.

### DIFF
--- a/Cabal/Distribution/Simple/GHC/ImplInfo.hs
+++ b/Cabal/Distribution/Simple/GHC/ImplInfo.hs
@@ -68,7 +68,7 @@ ghcVersionImplInfo ver = GhcImplInfo
   , flagPackageConf      = v <  [7,5]
   , flagDebugInfo        = v >= [7,10]
   , supportsPkgEnvFiles  = v >= [8,0,1,20160901] -- broken in 8.0.1, fixed in 8.0.2
-  , flagWarnMissingHomeModules = v >= [8,2,1]
+  , flagWarnMissingHomeModules = v >= [8,2]
   }
   where
     v = versionNumbers ver


ezyang:
I trust you on the version number :) LGTM

Cabal/Distribution/Simple/GHC/ImplInfo.hs
```diff
@@ -68,7 +68,7 @@ ghcVersionImplInfo ver = GhcImplInfo
   , flagPackageConf      = v <  [7,5]
   , flagDebugInfo        = v >= [7,10]
   , supportsPkgEnvFiles  = v >= [8,0,1,20160901] -- broken in 8.0.1, fixed in 8.0.2
-  , flagWarnMissingHomeModules = v >= [8,2,1]
+  , flagWarnMissingHomeModules = v >= [8,2,0,20170201]
```
hvr: You could have written `[8,2]` as well, as @bgamari branched off GHC 8.2 in march, and the patch was merged to GHC master on Jan 20 2017; so the earliest GHC 8.2 snapshots had the version number `8.2.0.201703xx`.

hvr:
I'm currently testing this, and I'm quite excited about this as finally there's a way to detect when you have incomplete `other-modules` even though the sdist is complete (this happens if you're careless with testsuites or executables, and GHC then picks up a module via hs-source-dirs rather than going through the intra-package library build-dep, in the process messing up cabal new-build's new file monitoring...)

Unfortunately, it's just this use-case which that doesn't work yet, as it seems that `-Wmissing-home-modules` isn't passed to the compilation of non-library targets... :-/

and when I pass it explicitly via `ghc-options: -Wmissing-home-modules` GHC complains about `Main`... *sigh*

Let's see if we can fix this in GHC before 8.2.1 final hits... in the current form, this has little benefit for me, as I already have workarounds/tooling to detect missing home modules in library stanzas, it's the non-library stanzas that still cause me problems... :-/

hvr:
Ok, so I've identified to problems.

Consider an executable with a `Main` module, which imports a module `Dummy`; now there's two problems:

 1.  GHC's `-Wmissing-home-modules` gets confused if the `Main` module is prefixed by a folder; e.g. if you say `ghc --make ./Main.hs` it adds `Main` to the set of home modules. However, if you say `ghc --make src/Main.hs`, then `Main` doesn't enter the set. I'm going to file this one upstream at GHC asap. This could be workarounded in cabal, if we instead let GHC locate `Main.hs` via the usual search-paths algorithm, i.e. invoke via  `ghc -isrc --make Main`

 2. Cabal doesn't pass the list of `other-extensions` to the `ghc --make` invocation for executables. Is there any reason why we can't do this at least for those GHCs which support `-Wmissing-home-modules`?

23Skidoo:
Yep, I also noticed that it doesn't work for executables currently. That's what I was trying to use it for :-(.

23Skidoo:
I suggest we merge this and open a new issue for `-Wmissing-home-modules` not working for exes.

Yuras:
IIRC it was intentional to make it work with libraries only, because cabal doesn't list all modules in command line. Though I could be wrong.

23Skidoo:
@hvr 

> Cabal doesn't pass the list of `other-extensions` to the `ghc --make` invocation for executables.

Sure you mean `other-modules`? Yes, I think we should do that.

hvr:
@23Skidoo yeah, I meant `other-modules` of course... I'm doing too many things at once :-)

23Skidoo:
Yep, that'd be nice.

hvr:
I know... which is why that warning would initially be rather a case for opt-in via `ghc-options` probably

hvr:
Upstream GHC ticket filed at https://ghc.haskell.org/trac/ghc/ticket/13727

PR #4527, opened 2017-05-19
Remove --allow-{newer,older} support from Cabal

### LINKED ISSUE #3581: RFC remove allow-older and allow-newer support from Setup.hs
_Shortly:_ revert changes made in  https://github.com/haskell/cabal/issues/3163 & https://github.com/haskell/cabal/pull/3165, as well as followed `--allow-older` changes. 

_Motivation:_ Such functionality isn't necessary (not used by anyone) to be provided by `Cabal`/`Setup.hs`<sup>&dagger;</sup>, and results into duplicate code (starting from argument parsing etc). `--exact-configuration` should be enough for everyone to force "I know what I'm doing". For the situations were "I almost know what I do, help me with rest", one could/should use external tools (e.g. `cabal-install` could have `cabal solver` command to produce configuration file).

_Optional:_
- Add `--exact-configuration` to `cabal-install`
- Add `cabal solver` (we now use `cabal configure -v2 --dry-run`, but it feels semantically wrong).

_Alternative:_ leave `--allow-newer` in `Cabal`, only remove followed `--allow-older` changes. IMHO this would lead to confusion. `--allow-older` hasn't hit Hackage AFAIK.

_Note:_ this is proposal to remove them only from `Cabal`-the-library, `cabal-install` will continue to have them.

<sup>&dagger;</sup> AFAICS stack opted to use `--exact-configuration` https://github.com/commercialhaskell/stack/issues/1579#issuecomment-168098272

ping @hvr, @ezyang, @23Skidoo, @dcoutts

hvr:
This is a preparatory refactoring needed for future work such as #4203.

I've refrained from doing additional cleanups in order to keep this a
refactoring that mostly moves around blocks of code mostly
unchanged (except for whitespace), and make it easier to review.

This feature was originally implemented because its lack was complained
about by Stack/Stackage developers. However, after it got implemented it
was never really being used; what's more, it's causing us overhead for
no benefit as well as blocking us improving the implementation via the
likes of #4203.

Closes #3581


hvr:
@23Skidoo Travis is green! Do you want to review this PR or can I merge it rightaway?

PR #4529, opened 2017-05-19
GHC 8.2.1 Compatibility

hvr:
This PR makes `cabal` compatible with GHC 8.2.1's libraries.

TODO: add GHC 8.2.1 to Travis-config


hvr:
@ezyang @23Skidoo unfortunately, the `cabal-testsuite` exposed a couple of problems w/ GHC 8.2.1RC2:

```
UNEXPECTED FAIL: 
PackageTests/Backpack/Includes2/setup-internal.test.hs 
PackageTests/Backpack/Includes2/setup-per-component.test.hs 
PackageTests/InternalLibraries/setup-gen-script.test.hs 
PackageTests/InternalLibraries/setup-per-component.test.hs 
PackageTests/InternalLibraries/setup-gen-pkg-config.test.hs 
PackageTests/InternalLibraries/Executable/setup-static.test.hs 
PackageTests/InternalLibraries/Library/setup.test.hs
```

(from https://travis-ci.org/haskell/cabal/jobs/234280545)

ezyang:
Yes, this is that upstream bug.

hvr:
@ezyang fair enough, in the mean-time a new GHC 8.2.1 snapshot has been build by the PPA which includes http://git.haskell.org/ghc.git/commit/8054a7435dc15bc45c04adabe098a78ff6d6a7eb but Travis now runs into a time out (the job gets killed after 50 minutes);

see e.g. https://travis-ci.org/haskell/cabal/jobs/234280545

ezyang:
The timeout is probably because we're building deps for the first time; they'll subsequently be cached. So if you keep restarting the build eventually it will succeed.

Test failure needs more investigatoin.

hvr:
@ezyang the problem is that the dependency are not cached if the job is killed by timeout. So I now temporarily workarounded this by `exit 1` shortly after everything has been built. This is exactly what Travis' new build stages would allow to do w/o such hacks!

ezyang:
Yes, those bugs are known: https://ghc.haskell.org/trac/ghc/ticket/13703

Should be fixed on next RC.

hvr:
@ezyang Why do you think the next RC would fix this? The current PPA ghc-8.2.1 snapshot is at ghc/ghc@c0b82c3826c8e1b26d198f050baa9d5077370247 (the build job used `8.2.0.20170522+git.0.c0b82c3-2~14.04`) which contains the supposed fix ghc/ghc@8054a7435dc15bc45c04adabe098a78ff6d6a7eb

23Skidoo:
Should we file a new GHC bug report/reopen 13703?

ezyang:
No. Instead, if we apply https://github.com/haskell/cabal/pull/4554 to this PR, I think things should be green.

hvr:
I've rebased this PR and added #4554 on top... now I'm curious :-)

23Skidoo:
The failure is due to timeout.

23Skidoo:
@ezyang There's one unexpected failure still: 
```
UNEXPECTED FAIL: PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.test.hs
```

ezyang:
@23Skidoo I pushed a fix to the other PR.

23Skidoo:
Yay, it passes now!

23Skidoo:
Cherry-picked relevant commits manually.

Issue #4528, opened 2017-05-19
-Wmissing-home-modules check doesn't work for executable components


23Skidoo:
We'd like to make the `-Wmissing-home-modules` check work for executables.

@hvr identified the following issues that need to be solved for it to work:

  * `-Wmissing-home-modules` gets confused when the main module is specified on the command line using a relative path. This needs to be fixed in GHC, upstream issue: https://ghc.haskell.org/trac/ghc/ticket/13727
  * Cabal doesn't pass the list of `other-modules` to `ghc --make` when building executables. This should be solved on the Cabal side.

See also discussion in https://github.com/haskell/cabal/pull/4526.

23Skidoo:
Fixed by #4534, but still needs a test case, and that is blocked by #4529. So leaving open for now.

hvr:
Points in the test-space might want to test/verify:

 - `main-is: main.c`
 - foreign libs
 - internal/convenience libraries
 - tests/executable/benchmark stanzas
 - autogenerated modules
 - combination of intra-package lib-dependency + overlapping hs-source-dirs (picking up modules directly already part of the library)
 - ...

hvr:
@23Skidoo I (or rather @alanz) just found a minor problem:

We need to cope with questionable cabal files such as http://hackage.haskell.org/package/happy-1.19.5/happy.cabal  which mention the main module twice;  via `main-is: Main.lhs`  *and* `other-modules: Main`.

So one simple hack would be to filter out modules named `Main` from other-modules for executables (and assuming that `ghc-options: -main-is ...` is used to infrequently that we don't need to bother);

...and also we should emit a warning when Main is both via main-is and via other-modules.

23Skidoo:
What happens now with such .cabal files?

hvr:
@23Skidoo GHC complains about multiple modules providing `main`; I'm already on it to implement a workaround (and I'll try to take into account `ghc-options`) ... PR coming later today!

hvr:
@23Skidoo See PR #4539

hvr:
From what I can see, this was already resolved in the released Cabal-2.0.0.2 (since #4539, #4533, #4534 were all backported and were part of the 2.0.0.2 release)

PR #4534, opened 2017-05-23
Pass -Wmissing-home-modules to all component types

hvr:
i.e. not only library components, but also executables, tests and benchmarks.

(With GHC-8.2.1 snapshot 8.2.0.20170521 or earlier, this emits a harmless warning about `Main`;
the next upcoming snapshot build of GHC-8.2.1 will already omit the spurious warning)

This addresses the other half of #4528

### DIFF
--- a/Cabal/Distribution/Simple/GHC/Internal.hs
+++ b/Cabal/Distribution/Simple/GHC/Internal.hs
@@ -316,12 +316,7 @@ componentGhcOptions verbosity implInfo lbi bi clbi odir =
         _ -> [],
       ghcOptNoCode          = toFlag $ componentIsIndefinite clbi,
       ghcOptHideAllPackages = toFlag True,
-      ghcOptWarnMissingHomeModules = case clbi of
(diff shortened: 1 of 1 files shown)


hvr:
The PPA snapshot `ghc-8.2.1_8.2.0.20170522+git.0.c0b82c3` which includes the upstream fix ghc/ghc@72eade6d7cf47f0738a58f560085acc952248d2c is already building; give it about 2-3 hours to become available to Travis.

23Skidoo:
Travis is green. Do we need a test case?

hvr:
@23Skidoo Well, at some point yes; I can think of a couple of constellations we could make into a testcase. However, This is all blocked on #4529 ...

So I'd just merge this one, and maybe create a reminder ticket to write testcases after GHC 8.2.1 got added to the test-matrix.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.0.

PR #4536, opened 2017-05-23
Refactor ProjectConfig Arbitrary instances

bitrauser:
In response to #4044, begin using record syntax instead of positional arguments.

I think this can be improved so I am happy for any kind of feedback and/or tips, especially since a part still uses positional arguments, but this change should already make it a bit clearer.


ezyang:
Great! You should merge when the build is green. OOC, why did you change the maps to fmaps?

bitrauser:
I thought those were an inconsistency, will change them back if I am mistaken.

23Skidoo:
Since `fmap` is overloaded, it makes you stop and think about which `Functor` instance is used each time when you're reading the code. So I prefer using `map` in monomorphic context.

bitrauser:
That does make sense, I was under the impression that fmap is generally preferred.

PR #4539, opened 2017-05-25
Workaround invalid .cabal files with Main modules in other-extensions

hvr:
`other-modules` is intended for non-main modules (which is also stated
in the cabal manual). Unfortunately, a few packages, such as `happy` do
incorrectly specify `other-modules: Main` for their executables, thereby
causing GHC to complain about duplicate Main modules.

The workaround implemented here is to filter out the main module
name (while taking into account `ghc-options: -main-is ...`) from the
`other-modules` passed to GHC, and emit a warning in the process.

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -103,6 +103,8 @@ import Distribution.Types.UnqualComponentName
 import Distribution.Utils.NubList
 import Language.Haskell.Extension
 
+import Control.Monad (msum)
+import Data.Char (isLower)
 import qualified Data.Map as Map
 import System.Directory
          ( doesFileExist, getAppUserDataDirectory, createDirectoryIfMissing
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Simple/GHC.hs
```diff
@@ -962,11 +964,56 @@ gbuildModDefFiles (GReplExe  _)     = []
...
+-- | Decode argument to '-main-is'
+--
+-- Returns 'Nothing' if argument set only the function name.
+--
+-- This code has been stolen/refactored from GHC's DynFlags.setMainIs function
+decodeMainIsArg :: String -> Maybe ModuleName
+decodeMainIsArg arg
+  | not (null main_fn) && isLower (head main_fn)
+                        -- The arg looked like "Foo.Bar.baz"
+  = Just (ModuleName.fromString main_mod)
```
23Skidoo: Do we actually check that `main_mod` is a valid module name? The way I see it, it could be something like `Foo.b$r.baz`, and this code would still accept it. I'd use `simpleParse` instead of `ModuleName.fromString` here.

Another possible refactoring is to add a type `QualifiedName = QualifiedName { qualModName :: ModuleName, qualName :: String }` that could potentially be also useful elsewhere, then this function would become just ``msum [ qualModName `fmap` simpleParse arg, simpleParse arg ]`` (assuming that `QualifiedName` has a `Text` instance).

hvr: Originally I started with doint it properly, but then I noticed that GHC's logic was different; so instead stole exactly the exact same logic GHC uses (which is as you correctly note imperfect). If I had done it properly, we'd have a different logic than the faulty one GHC uses. TLDR: I tried to be bug-compatible to GHC's faulty `-is-main` parser... :-)

23Skidoo: All right, let's merge then.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.0.

hvr:
@23Skidoo I'll try to remember to write some source-code comment to point out why the code is "broken" the way it is, next time I hack on cabal

23Skidoo:
@hvr I did that already: https://github.com/haskell/cabal/commit/420d1197647543cf4ba4f770a14a5e946887ce0d

PR #4543, opened 2017-05-28
Override flags WIP

colinwahl:
This refactors `FlagAssignment` to be a Data.Map.  

I am running this to check if it passes all tests.

STILL WIP

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1519,7 +1519,7 @@ checkPackageVersions pkg =
     -- open upper bound. To get a typical configuration we finalise
     -- using no package index and the current platform.
     finalised = finalizePD
-                              [] defaultComponentRequestedSpec (const True)
+                              Map.empty defaultComponentRequestedSpec (const True)
                               buildPlatform
                               (unknownCompilerInfo
                                 (CompilerId buildCompilerFlavor nullVersion)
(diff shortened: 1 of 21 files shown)


ezyang:
Test failure on AppVeyor looks spurious, all you need to do is update the output (changing the data structure a map probably affected how we compute the hash.)

hvr:
Does this have a measurable effect on heap allocations or performance? I'm worried that for such small maps, `Data.Map` may actually be worse than `[(k,v)]`.

ezyang:
(Let us know if you need guidance on how to check for allocation changes.)

colinwahl:
That was exactly what I was going to ask! I am not sure how to check for allocation changes.  Guidance would be appreciated!

23Skidoo:
Test runs with `+RTS -sstderr` before and after?

hvr:
to add to @23Skidoo: also make sure you have asserts disabled (they slow down the solver quite a lot), and also try to solve for more complex install-plans which have lots of backjumps.

colinwahl:
Okay, I figured out how to profile for memory.  I will try to find more complex install-plans to test with after work tomorrow.  Anywhere in particular I should look for these install plans, or should I try to put one together myself? Also, how do disable asserts?  Is that a flag on `cabal configure`?  Thanks in advance for the help

23Skidoo:
To disable *all* assertions, remove all instances of `-fno-disable-asserts` from `cabal.project` (assuming you're using `new-build`). Expensive assertions are disabled by default, but you can add `cabal-install -debug-expensive-assertions` to the list of constraints in `cabal.project` to be extra sure.

I've also added a warning that is shown when assertions are enabled (#4552).

PR #4547, opened 2017-06-01
Move Distribution.Client.Compat.Prelude to Distribution.Compat.Prelude

BardurArantsson:
The idea here is to break the dependency that has crept in from the
"Solver" into the "Client". (If there's ever going to be a separate
solver then the dependency would become a big problem.)

----

### DIFF
--- a/cabal-install/Distribution/Solver/Compat/Prelude.hs
+++ b/cabal-install/Distribution/Solver/Compat/Prelude.hs
@@ -1,0 +1,19 @@
+-- to suppress WARNING in "Distribution.Compat.Prelude.Internal"
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
+
+-- | This module does two things:
+--
+-- * Acts as a compatiblity layer, like @base-compat@.
+--
+-- * Provides commonly used imports.
+--
+-- This module is a superset of "Distribution.Compat.Prelude" (which
+-- this module re-exports)
+--
+module Distribution.Solver.Compat.Prelude
+  ( module Distribution.Compat.Prelude.Internal
+  , Prelude.IO
+  ) where
+
+import Prelude (IO)
+import Distribution.Compat.Prelude.Internal hiding (IO)
--- a/cabal-install/Distribution/Solver/Modular/Linking.hs
+++ b/cabal-install/Distribution/Solver/Modular/Linking.hs
@@ -5,7 +5,7 @@ module Distribution.Solver.Modular.Linking (
   ) where
 
 import Prelude ()
-import Distribution.Client.Compat.Prelude hiding (get,put)
+import Distribution.Solver.Compat.Prelude hiding (get,put)
 
 import Control.Exception (assert)
 import Control.Monad.Reader
--- a/cabal-install/Distribution/Solver/Modular/Log.hs
+++ b/cabal-install/Distribution/Solver/Modular/Log.hs
@@ -4,7 +4,7 @@ module Distribution.Solver.Modular.Log
     ) where
 
 import Prelude ()
-import Distribution.Client.Compat.Prelude
+import Distribution.Solver.Compat.Prelude
 
 import Data.List as L
 
--- a/cabal-install/Distribution/Solver/Modular/Preference.hs
+++ b/cabal-install/Distribution/Solver/Modular/Preference.hs
@@ -17,7 +17,7 @@ module Distribution.Solver.Modular.Preference
     ) where
 
 import Prelude ()
-import Distribution.Client.Compat.Prelude
+import Distribution.Solver.Compat.Prelude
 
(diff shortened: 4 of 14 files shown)


phadej:
:-1:. `Cabal` the library also have (non exposed) `Distribution.Compat.Prelude` which is a subset. This change will only induce confusion.

`Distribution.Something.Compat.Prelude` would be ok, where `Something` is a place to share the code between `Client` and `Solver`, but not `Cabal` the lib.

BardurArantsson:
Ah, right, that's probably true.

I was also planning to move one of the Utils.Assert modules (for the same reason), so I guess I'll have to come up some sensible "middle module" name anyway...

Any suggestions for the `Something`?

hvr:
The reason `Distribution.Client.Compat.Prelude` was picked originally was that there was already a precedent for that, as it's not the only `Compat` module in the `cabal-install` owned namespace:

```
        Distribution.Client.Compat.ExecutablePath
        Distribution.Client.Compat.FileLock
        Distribution.Client.Compat.FilePerms
        Distribution.Client.Compat.Prelude
        Distribution.Client.Compat.Process
        Distribution.Client.Compat.Semaphore
```

Those `Distribution.Client.Compat.*` modules should be ideally self-contained and not depend on anything else in the `Distribution.Client.*` namespace. So moving only the `.Prelude` w/o also renaming the other modules would be inconsistent. And one would need to be careful to stay out the way of `Cabal`'s owned module namespace to avoid confusion... this is not trivial :-)

BardurArantsson:
I don't foresee any issue with moving all of them -- in fact, I think it's preferable (for the reasons you state).

(... and yes, it would have to be a 3rd package, but there's no way around that -- other than just duplicating the tiny bit of code in the `Client.Compat.Prelude` package. After all it gets most of its functionality from `Distribution.Compat`. Some deal with the `Assert` package, I think.)

Anyway, the main interest right now is to prevent the clean separation between Solver and `cabal-install` from bitrotting accidentally. I'm aiming to add a check *somewhere* in the build infrastructure which ensures that this won't happen again (at least accidentally).

hvr:
@BardurArantsson btw, one thing I've been thinking about is to have an easy way to transform `cabal-install.cabal` into the variant which uses internal convenience libs for the solver and stuff; I think we can script that if we put in place a few markers in the `cabal-install.cabal` file. That way we could have CI enforce the module-barriers.

BardurArantsson:
Sounds like an interesting idea... but I'm not sure exactly what you mean. Are you thinking *something* like \#ifdef or similar?

hvr:
@BardurArantsson `#ifdefs` in their crudest forms; but you can encode the ifdefs also in the script doing the transformation; it just needs enough information to be able to recognize how to rearrange the `cabal-install.cabal` into one w/ internal libs.

ezyang:
To me, it doesn't seem worth doing overly many gyrations to get this clean separation; GHC's inability to do transparently recursive module loops exerts a reasonable pressure to prevent Client.Prelude from getting unruly dependencies. 

But, if we must, how about `Distribution.Solver.Compat.Prelude`, which is a clone of `Distribution.Client.Compat.Prelude`? The custom prelude is kind of idiosyncratic to each project, so if solver ever did get separated out, it might very well be reasonable for solver to do something different.

23Skidoo:
> But, if we must, how about `Distribution.Solver.Compat.Prelude`, which is a clone of `Distribution.Client.Compat.Prelude`?

Yep, I think we can live with some duplication here.

BardurArantsson:
@ezyang I like your suggestion, I'll do that. (It really is only about 20 LoC.)

Plus, I can't come up with a name for a "middle-layer" compat module thingy, so there's that :)

I'll try to redo this PR next week as per @ezyang 's suggestion.

BardurArantsson:
AppVeyor failure looked spurious, Travis failure was real... retrying...

BardurArantsson:
Alright, looks good to go. Anyone wish to review? :)

cabal-install/Distribution/Solver/Compat/Prelude.hs
```diff
@@ -0,0 +1,19 @@
...
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
+
+-- | This module does two things:
+--
+-- * Acts as a compatiblity layer, like @base-compat@.
+--
+-- * Provides commonly used imports.
+--
+-- This module is a superset of "Distribution.Compat.Prelude" (which
+-- this module re-exports)
```
BardurArantsson: Just a note: I've left this module in place (instead of just switching to Dist.Compat.Prelude directory) comment because figure it might grow a bit over time.

23Skidoo:
LGTM.

23Skidoo:
Merged, thanks!

PR #4549, opened 2017-06-04
Make new-haddock --haddock-for-hackage generate doc tarball

### LINKED ISSUE #4521: `new-haddock --haddock-for-hackage` doesn't generate doc tarball
#4412 added support for `--for-hackage`; one important feature was left out though, the generation of the doc tarball that `cabal haddock --for-hackage` would generate.

This is one of last pieces blocking #3837

cocreature:
fixes #4521

This seems to work but I’m not sure if it is the right way to do
this. It also places the doc tarball in `dist-newstyle` instead of in
`dist`. While that seems reasonable for `new-haddock` this means that
currently (since there is not yet a `new-dist`) the doc tarball will
live in a different directory than the source tarball. I’m also not
sure if it should go in some subdirectory of new-dist instead.

It should probably also have an integration test but I haven’t yet
looked into what the best way to test this is.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -65,6 +65,7 @@ import           Distribution.Package hiding (InstalledPackageId, installedPacka
 import qualified Distribution.PackageDescription as PD
 import           Distribution.InstalledPackageInfo (InstalledPackageInfo)
 import qualified Distribution.InstalledPackageInfo as Installed
+import           Distribution.Simple.BuildPaths (haddockDirName)
 import qualified Distribution.Simple.InstallDirs as InstallDirs
 import           Distribution.Types.BuildType
 import           Distribution.Simple.Program
@@ -1061,7 +1062,8 @@ buildInplaceUnpackedPackage :: Verbosity
 buildInplaceUnpackedPackage verbosity
                             distDirLayout@DistDirLayout {
                               distTempDirectory,
-                              distPackageCacheDirectory
+                              distPackageCacheDirectory,
+                              distDirectory
                             }
                             BuildTimeSettings{buildSettingNumJobs}
                             registerLock cacheLock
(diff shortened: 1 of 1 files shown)


23Skidoo:
Does new-haddock work on multi-package projects? Try also testing with a package that has `build-type: Custom`.

cabal-install/Distribution/Client/ProjectBuilding.hs
```diff
@@ -1166,8 +1168,15 @@ buildInplaceUnpackedPackage verbosity
 
         -- Haddock phase
         whenHaddock $
-          annotateFailureNoLog HaddocksFailed $
-          setup haddockCommand haddockFlags []
+          annotateFailureNoLog HaddocksFailed $ do
+            setup haddockCommand haddockFlags []
+            let haddockTarget = elabHaddockForHackage pkg
+            when (haddockTarget == Cabal.ForHackage) $ do
+              let dest = distDirectory </> name <.> "tar.gz"
```
23Skidoo: I think it should be `$dist-newstyle/doc/$tarball-name`. Or maybe even `$distBuildDirectory/doc/$tarball-name`.

cocreature:
> Does new-haddock work on multi-package projects?

Yes it does. With this PR one tarball will be created per project (I tested this).

> Try also testing with a package that has build-type: Custom.

I’ve tested this with `llvm-hs` which has a moderately complicated `Setup.hs` and it seems to work just fine. Are you thinking of anything in particular that could cause problems here?

cabal-install/Distribution/Client/ProjectBuilding.hs
cocreature: I don’t really like the `doc` subdirectory. It already has `docs` in the tarball name and eventually I think it would make sense if the tarballs that need to be uploaded all end up in the same directory.

23Skidoo:
> With this PR one tarball will be created per project (I tested this).

So in this case `--for-hackage` will create something that can't be uploaded to Hackage :-). IMO creating one tarball per package makes more sense, but I guess we can fix that later...

cocreature:
> So in this case --for-hackage will create something that can't be uploaded to Hackage :-). IMO creating one tarball per package makes more sense, but I guess we can fix that later...

Oh sorry, I intended to say one tarball per package and for some reason thought of packages as projects.

cabal-install/Distribution/Client/ProjectBuilding.hs
23Skidoo: Same place where `sdist` puts the source tarball, you mean. OK, makes sense.

23Skidoo:
OK, I think this can be merged.

23Skidoo:
Merged, thanks!

PR #4552, opened 2017-06-05
Print a warning when assertions are enabled.

### LINKED ISSUE #4377: Print a warning when cabal is built with assertions enabled
This idea is from #4258.  It would be useful to be able to check whether cabal was built with assertions enabled when measuring performance.  We could print the information with `--version`, or warn at the start of expensive commands, such as `new-build` or `install`.

23Skidoo:
Fixes #4377.

### DIFF
--- a/cabal-install/main/Main.hs
+++ b/cabal-install/main/Main.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP, ScopedTypeVariables #-}
 
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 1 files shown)


23Skidoo:
Test failures seem to be caused by changes in output, i.e. all those
```
+Warning: this is a debug build with assertions enabled.
```
diffs. 

@ezyang 
Should I just fix the expected output or disable the warning during test suite runs (e.g. by checking an env var)?

ezyang:
Waaay late to the party.

This warnings should not show up in the expect files. I'm a little confused why they are showing up, though, since you're outputting them directly with `putStrLn`, so they shouldn't have been "marked" for output...

gbaz:
If @23Skidoo doesn't want to pick this up anymore, I think it would be good if someone else did. This is a useful feature!

jneira:
I've opened a new pr cherry picking this one (so commit authoring is preserved), thanks for the pr!

PR #4554, opened 2017-06-06
Fix test suite wobbliness on GHC 8.2.

ezyang:
Here were the root causes:

- Some tests involving Custom setpu showed MORE output
  (UseLocalPackageForSetup) when run on GHC 8.2.  This is because GHC
  8.2 ships a recent enough version of Cabal to know how to emit
  markers, which means we have started picking up the output.  I hacked
  up these tests to not accept this output, but a more correct thing to
  do is figure out how to NOT request marking of a Setup script which is
  not the inplace install.  This was a little tricky so I bailed.

- GHC 8.2 no longer emits "It is a member of the hidden package".
  This broke CustomWithoutCabalDefaultMain.  Not sure if this
  is a GHC regression but it's pretty harmless.

- While I was at it, I fixed an inexhaustive pattern match in
  cabal-testsuite (though perhaps poorly; I couldn't figure out
  what the new constructor does.)


23Skidoo:
It looks like there is still one remaining failure, see https://github.com/haskell/cabal/pull/4529#issuecomment-306598031.

23Skidoo:
#4529 has passed, merging!

PR #4563, opened 2017-06-14
Solver: Refactor Validate.hs.

grayjay:
This just moves some functions and types into Validate.hs in preparation for
#4391.

______________________________________________________

I wanted to make separate PRs with the refactoring changes from #4562, to avoid conflicts.  This is the first commit from #4562.


ezyang:
LGTM.

PR #4564, opened 2017-06-14
Refactor goal sorting in the solver DSL.

grayjay:
Now DSL.exResolve takes a goal-ordering function, which is more flexible
than a list of sorted goals.

____________________________________________________

This is the second commit from #4562.


ezyang:
LGTM.

PR #4575, opened 2017-06-25
Enhance `--allow-{newer,older}` syntax

hvr:
This extends the capabilities of `--allow-{newer,older}` to allow for more
fine-grained scoping to control more precisely which packages and constraints
a relaxation is applied to. See updated documentation for more details.

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -73,6 +73,7 @@ import Distribution.Client.Types
          ( SourcePackageDb(SourcePackageDb)
          , UnresolvedPkgLoc, UnresolvedSourcePackage
          , AllowNewer(..), AllowOlder(..), RelaxDeps(..), RelaxedDep(..)
+         , RelaxDepScope(..), RelaxDepMod(..)
          )
 import Distribution.Client.Dependency.Types
          ( PreSolver(..), Solver(..)
@@ -90,10 +91,6 @@ import Distribution.PackageDescription.Configuration
          ( finalizePD )
 import Distribution.Client.PackageUtils
          ( externalBuildDepends )
-import Distribution.Version
-         ( Version, mkVersion
-         , VersionRange, anyVersion, thisVersion, orLaterVersion, withinRange
-         , simplifyVersionRange, removeLowerBound, removeUpperBound )
 import Distribution.Compiler
          ( CompilerInfo(..) )
 import Distribution.System
@@ -108,6 +105,7 @@ import Distribution.Text
          ( display )
 import Distribution.Verbosity
          ( normal, Verbosity )
+import Distribution.Version
 import qualified Distribution.Compat.Graph as Graph
 
 import           Distribution.Solver.Types.ComponentDeps (ComponentDeps)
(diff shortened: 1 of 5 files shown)


cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -452,35 +444,68 @@ removeLowerBounds (AllowOlder allowNewer)    params =
...
-relaxPackageDeps _ RelaxDepsNone gpd = gpd
-relaxPackageDeps vrtrans RelaxDepsAll  gpd = PD.transformAllBuildDepends relaxAll gpd
+relaxPackageDeps _ RelaxDepsNone gpd = gpd -- subsumed by no-op case in 'removeBounds'
+relaxPackageDeps relKind RelaxDepsAll  gpd = PD.transformAllBuildDepends relaxAll gpd
   where
-    relaxAll = \(Dependency pkgName verRange) ->
-      Dependency pkgName (vrtrans verRange)
-relaxPackageDeps vrtrans (RelaxDepsSome allowNewerDeps') gpd =
+    relaxAll (Dependency pkgName verRange) =
+        Dependency pkgName (removeBound relKind RelaxDepModNone verRange)
```
23Skidoo: I wonder if we should allow something like `--allow-newer=^all` to relax only caret dependencies everywhere. Probably not.

23Skidoo: BTW, the `^` thing conflicts with `zsh` shell expansion too:

```
$ echo ^all              
appveyor-retry.cmd appveyor.yml AUTHORS Cabal cabal-install cabal.project cabal.project.travis 
cabal-tests.log cabal-testsuite default.nix dist dist-newstyle ghc-packages id_rsa_cabal_website.aes256.enc LICENSE README.md stack.yaml TAGS travis travis-bootstrap.sh travis-common.sh travis-deploy.sh 
travis-install.sh travis-meta.sh travis-script.sh travis-solver-debug-flags.sh travis-stack.sh update-cabal-files.sh
$ echo foo:^all
zsh: no matches found: foo:^all
```

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -452,35 +444,68 @@ removeLowerBounds (AllowOlder allowNewer)    params =
...
+        | Just relMod <- Map.lookup depName depsToRelax =
+            Dependency depName (removeBound relKind relMod verRange)
+        | otherwise = d -- no-op
+
+-- | Internal helper for 'relaxPackageDeps'
+removeBound :: RelaxKind -> RelaxDepMod -> VersionRange -> VersionRange
+removeBound RelaxLower RelaxDepModNone = removeLowerBound
+removeBound RelaxUpper RelaxDepModNone = removeUpperBound
+removeBound relKind RelaxDepModCaret =
+    foldVersionRange'
```
hvr: I had the very same thought when I implemented that function!  :-)

cabal-install/Distribution/Client/Dependency.hs
hvr: Well, I do plan to support `--allow-newer=all`, `--allow-newer=^all`, `--allow-newer=foo:^all`, etc. via a future PR. That's a perfectly valid use-case  when I want to quickly see if there's overly constraining upper bounds in a specific package.

I'm not sure what to do about `zsh` getting in the way... which unquoted syntax does it _not_ try to interpret?  ;-)

cabal-install/Distribution/Client/Types.hs
```diff
@@ -409,20 +412,55 @@ data RelaxDeps =
...
+               RelaxDepModCaret -> Disp.char '^' Disp.<> disp dep
 
-  parse = scopedP Parse.<++ normalP
+  parse = RelaxedDep <$> scopeP <*> modP <*> parse
     where
-      scopedP = RelaxedDepScoped `fmap` parse <* Parse.char ':' <*> parse
-      normalP = RelaxedDep       `fmap` parse
+      -- "greedy" choices
+      scopeP =           (pure RelaxDepScopeAny  <* Parse.char '*' <* Parse.char ':')
+               Parse.<++ (pure RelaxDepScopeAny  <* Parse.string "all:")
```
hvr: I'll rename

cabal-install/Distribution/Client/Dependency.hs
23Skidoo: I think we can keep using `^`, it's not a very common operation, and zsh users can use quotes.

PR #4578, opened 2017-06-28
get last update timestamp from XX-index.timestamp's modification time

### LINKED ISSUE #4444: Cabal reports index is out of date based on index time, not last update time
**New summary by @ezyang.** When you run `cabal install` and it has been a while since you have run `cabal update`, you get this useful message:

```
Warning: The package list for 'next-hackage' is 16 days old.
```

This message is computed by looking at the most recently updated package in the package index, and then emitting a message if it was from too long ago. Unfortunately, there is a bug in how we perform this check.

**Steps to reproduce.**

1. Configure the `next-hackage` index using the instructions on http://next.hackage.haskell.org:8080/
2. Run `cabal update`
3. Look at the files in `~/.cabal/packages/next-hackage`, and manually reset their file modification times to some point far in the past (e.g., more than Cabal's update threshold.)
4. Run `cabal install`. You should get an out of date warning.
5. Run `cabal udpate`

Expected result: No more out of date warning
Actual result: The warning persists

**How to fix.**

1. First, you need to find where this warning is being emitted. If you grep for the message string you can find it.
2. You'll find the message is emitted in `cabal-install/Distribution/Client/IndexUtils.hs`, according to the conditional here:

```
    isOldThreshold = 15 --days
    warnIfIndexIsOld dt = do
      when (dt >= isOldThreshold) $ case repo of
        RepoRemote{..} -> warn verbosity $ errOutdatedPackageList repoRemote dt
        RepoSecure{..} -> warn verbosity $ errOutdatedPackageList repoRemote dt
        RepoLocal{..}  -> return ()

```

3. By tracing the code flow, we see `warnIfIndexIsOld =<< getIndexFileAge repo`, where getIndexFileAge looks at the file modification time. This means that the problem is likely that `cabal update` is not updating the modification time on a file when it updates and sees no update is necessary.

4. The code for updating on `cabal update` is in `cabal-install/Distribution/Client/Update.hs` in `updateRepo`. You will need to update this code to update the file mod time EVEN when there were no updates. You should at least handle the `RepoSecure` case but it may also be worth fixing `RepoRemote`

5. Check and make sure that the bug has gone away. It's not necessary to write a test (we don't really have infrastructure for mucking about the clock in cabal-install.)

**What you will learn.** How to build and develop Cabal. How to setup external repositories in Cabal. How index updates work in Cabal.

----

I have `next-hackage` configured, which changes rarely.

If I do a configure, I get the following warning

```
Warning: The package list for 'next-hackage' is 16 days old.
```

despite having just done one.

Perhaps there should be a local record of when last an update was done, with a grace period on the warning based on that.

tvestelind:
The reason we cannot the timestamp of the last update it from
XX-index.tar is because it's modification time is not updated unless
there is a newer version available in the repository.

On the other hand, XX-index.timestamp's modification time is always
updated, even if there are no new updates. That means that the
modifiation time of file will correspond to the last update attemp. So,
to decide whether to warn about having an old index we can use the last
modification time of that file.

One could also have the modification time of XX-index.tar updated even
thought there aren't any updates, but to do that I think we'd need to
modify checkForUpdates in Hackage.Security.Client.

Signed-off-by: Tomas Vestelind <tomas.vestelind@gmail.com>

This fixes #4444 and I tested this by following the steps to reproduce described in that issue. But to summarize, I did:
1. Add next-hackage to my `.cabal/config`
2. Built the latest cabal-install from master
3. Ran `./dist/build/cabal/cabal update` (I'll exclude the `./dist/.../` part from now one, it's just to show that I really used what I build)
4. Went to `cabal/packages/next-hackage` and used `touch` to set the modification date on all files to one year ago
5. Ran `cabal install` and verified that I got the error `Warning: The package list for 'next-hackage' is 365 days old.`
6. Ran `cabal update && cabal install` and verified that the error was still there.
7. Implemented my suggested solution
8. Ran `cabal update && cabal install` and verified that the error was gone.

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -17,7 +17,7 @@
 -- Extra utils related to the package indexes.
 -----------------------------------------------------------------------------
 module Distribution.Client.IndexUtils (
-  getIndexFileAge,
+  getIndexTimestampFileAge,
   getInstalledPackages,
   Configure.getInstalledPackagesMonitorFiles,
   getSourcePackages,
@@ -308,7 +308,7 @@ readRepoIndex :: Verbosity -> RepoContext -> Repo -> IndexState
               -> IO (PackageIndex UnresolvedSourcePackage, [Dependency], IndexStateInfo)
 readRepoIndex verbosity repoCtxt repo idxState =
   handleNotFound $ do
-    warnIfIndexIsOld =<< getIndexFileAge repo
+    warnIfIndexIsOld =<< getIndexTimestampFileAge repo
     updateRepoIndexCache verbosity (RepoIndex repoCtxt repo)
     readPackageIndexCacheFile verbosity mkAvailablePackage
                               (RepoIndex repoCtxt repo)
@@ -355,9 +355,9 @@ readRepoIndex verbosity repoCtxt repo idxState =
       ++ "' is " ++ shows (floor dt :: Int) " days old.\nRun "
       ++ "'cabal update' to get the latest list of available packages."
 
--- | Return the age of the index file in days (as a Double).
-getIndexFileAge :: Repo -> IO Double
-getIndexFileAge repo = getFileAge $ indexBaseName repo <.> "tar"
+-- | Return the age of the index timestamp file in days (as a Double).
+getIndexTimestampFileAge :: Repo -> IO Double
+getIndexTimestampFileAge repo = getFileAge $ indexBaseName repo <.> "timestamp"
 
 -- | A set of files (or directories) that can be monitored to detect when
 -- there might have been a change in the source packages.


23Skidoo:
This is nice, but doesn't this only work for for secure repos? In the non-secure case we don't have the `.timestamp` file:

```
$ ls ~/.cabal/packages/hackage.scrive.com/ 
00-index.cache        consumers/    happstack-static-routing/
00-index.tar          curl/         kontrakcja-templates/
00-index.tar.gz       dataenc/      mixpanel/
00-index.tar.gz.etag  fields-json/  resource-pool/
```

hvr:
I just want to point out that `--index-state` makes only sense for secure type repos and (possibly - I don't rememeber OTTOMH if the current logic allows to use --index-state with 00-index) for non-secure repos which avoid `.cabal` revisions. Using `--index-state` with a non-secure repo + revisions will result in unsound states.

tvestelind:
Sorry, I thought that we wrote a .timestamp file in all cases. I'll have to rethink the solution then.

Do you think that I should look into updating the modification time for XX-index.tar?

23Skidoo:
Re: `--index-state`, it's not enabled by default, while this PR changes the default behaviour. So I'm okay with `--index-state` being secure-only. Though we should fail with an error (if possible) instead of going into an unsound state.

> Do you think that I should look into updating the modification time for XX-index.tar?

I think that'd be a valid solution.

23Skidoo:
Closing in favour of #4950.

PR #4583, opened 2017-07-03
an empty target is an empty string

fgaz:
without this `cabal new-run` without targets errors with an irrefutable pattern

### DIFF
--- a/cabal-install/Distribution/Client/TargetSelector.hs
+++ b/cabal-install/Distribution/Client/TargetSelector.hs
@@ -352,9 +352,10 @@ showTargetString = intercalate ":" . components
 
 showTargetSelector :: Package p => TargetSelector p -> String
 showTargetSelector ts =
-  let (t':_) = [ t | ql <- [QL1 .. QLFull]
-                   , t  <- renderTargetSelector ql ts ]
-   in showTargetString (forgetFileStatus t')
+  case [ t | ql <- [QL1 .. QLFull]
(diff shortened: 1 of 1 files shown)


23Skidoo:
Looks OK to me, by the way.

23Skidoo:
Anyhoozle, I think I'm going to merge this.

ezyang:
I'm generally down on using empty strings as a sigil. Does turning the return type into Maybe cause lots of knock on changes?

23Skidoo:
I don't think using `Maybe` here makes sense, because this function is used during error message formatting, so you'll just end up with a `fromMaybe ""` at every use site. Maybe it should return something like "<empty>" instead of an empty string.

fgaz:
+1, it's called _show_ TargetSelector after all.

`The run command is for running a single executable at once. The target '' refers to the package MultipleExes-1.0 which includes the executables foo and bar.`
vs
`The run command is for running a single executable at once. The target '<empty>' refers to the package MultipleExes-1.0 which includes the executables foo and bar.`
The second does look nicer, but the quotes imply \<empty\> is _literally_ the target.
Of course nobody would interpret it like that. I hope.

(On the other hand with Maybe we could do
`The run command is for running a single executable at once. The empty target refers to the package MultipleExes-1.0 which includes the executables foo and bar.`)

ezyang:
OK, I trust your judgment :)

PR #4586, opened 2017-07-04
new-run

### LINKED ISSUE #4477: new-run not running the executable
I'm testing with the `2.0` branch. My package built successfully and has one executable section. The `new-run` command correctly checks if things are up to date and rebuilds them, but it is not running the executable. I added printouts in `main` that I never see and the program should block on a `getContents` but it doesn't do that either. I've tried adding `exe:mypkg` to the `new-run` command and it doesn't change the behavior.

@dmwit

fgaz:
Fixes #4477.

The ci should be green now. Changelog, docs, and more tests incoming.

Can someone review it in the meantime?


fgaz:
ping @dmwit

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -90,7 +121,8 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     targetSelectors <- either (reportTargetSelectorProblems verbosity) return
-                   =<< readTargetSelectors (localPackages baseCtx) targetStrings
+                   =<< readTargetSelectors (localPackages baseCtx)
+                         (take 1 targetStrings) -- we drop the exe's args
```
ezyang: Urk. Wouldn't it be better to split targetStrings first before doing anything with it? See (A).

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+           of Just x -> " matching " ++ x
+              Nothing -> ""
+        ++ ":\n"
+        ++ unlines (fmap (\(p,_) -> " - in package " ++ display (elabUnitId p)) elabPkgs)
+    let exePath = binDirectoryFor (distDirLayout baseCtx)
+                                  (elaboratedShared buildCtx)
+                                  pkg
+                                  exe
+               </> exe
+    let args = drop 1 targetStrings
```
ezyang: (A) Then it wouldn't be necessary to subsequently drop it here!

cabal-testsuite/PackageTests/NewBuild/T4477/cabal.out
```diff
@@ -0,0 +1,7 @@
+# cabal new-run
+Resolving dependencies...
+In order, the following will be built:
+ - T4477-1.0 (exe:foo) (first run)
+Configuring executable 'foo' for T4477-1.0..
+Preprocessing executable 'foo' for T4477-1.0..
+Building executable 'foo' for T4477-1.0..
```
ezyang: Maybe not for this PR, but we should think carefully about whether or new-run should output this sort of information. In favor: it's good to be able to see when your executable is being built. Against: you won't be able to use new-run to run executables that you want to run as part of a pipeline, since there will be extra Cabal build progress goo in your stderr. Perhaps this is related to the desire for a mode, "please run this as quickly as possible, don't rebuild at all!"

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+-- | The path to the directory that contains a specific executable.
+binDirectoryFor
+  :: DistDirLayout
+  -> ElaboratedSharedConfig
+  -> ElaboratedConfiguredPackage
+  -> FilePath
+  -> FilePath
+binDirectoryFor layout config package exe = case elabBuildStyle package of
+  BuildAndInstall -> installedBinDirectory package
+  BuildInplaceOnly -> inplaceBinRoot layout config package </> exe
```
ezyang: This logic is duplicated with `inplace_bin_dir` in `Distribution/Client/ProjectPlanning.hs`. It would be good to have this in only one place.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+    -- Once we drop support for pre-ftp ghc
+    -- it's safe to remove this.
+    sequenceA' (a, Just b) = Just (a, b)
+    sequenceA' _ = Nothing
+    match :: ElaboratedConfiguredPackage -> Bool
+    match p = matchPackage pkgId p && matchComponent component p
+    matchingExecutable p = atMostOne
+                         $ filter (\x -> Just x == componentString
+                                      || isNothing componentString)
+                         $ executablesOfPackage p
```
ezyang: You don't have to change this, but my preference here would have been to case on p here for the package versus component case, rather than put everything through the "get the list of components it defines, and now check if anything in that list matches". My primary reasoning here is that the component case is more reflective of what the "good" code looks like, but the way things are wired up now it makes it seem like the package case (where multiple components may exist) is more important.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+--  * specified by the user (both Just)
+--  * deduced from an user-specified package (the component is unspecified, Nothing)
+--  * deduced from the cwd (both the package and the component are unspecified)
+extractMatchingElaboratedConfiguredPackages
+  :: Maybe PackageId -- ^ the package to match
+  -> Maybe ComponentName -- ^ the component to match
+  -> ElaboratedInstallPlan -- ^ a plan in with to search for matching exes
+  -> [(ElaboratedConfiguredPackage, UnqualComponentName)] -- ^ the matching package and the exe name
+extractMatchingElaboratedConfiguredPackages
+  pkgId component = nubBy equalPackageIdAndExe
```
ezyang: Why is a nub necessary here? If the same package and component show up multiple times, isn't that a problem?

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -90,7 +121,8 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
     baseCtx <- establishProjectBaseContext verbosity cliConfig
 
     targetSelectors <- either (reportTargetSelectorProblems verbosity) return
-                   =<< readTargetSelectors (localPackages baseCtx) targetStrings
+                   =<< readTargetSelectors (localPackages baseCtx)
```
ezyang: It is too bad there is no singular version of readTargetSelectors, since it seems like that is what you want.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+    -- These are wrapped in Maybes, because the user
+    -- might not specify them
+    (selectedPackage, selectedComponent) <-
+       -- this should always match [x] anyway because
+       -- we already check for a single target in TargetSelector.hs
+       case selectorPackageAndComponent <$> targetSelectors
+         of [x] -> return x
+            [ ] -> die'
+                     verbosity
+                     "No targets given"
```
ezyang: Won't this error message occur when I type `cabal new-run`? I think you can give a more informative message here.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+            _   -> die'
+                     verbosity
+                     "Multiple targets given"
```
ezyang: This cannot happen, by virtue of the `take 1`, correct? In that case, better to mark it as an internal error.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +160,209 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+        ++ unlines (fmap (\(p,_) -> " - in package " ++ display (elabUnitId p)) elabPkgs)
+    let exePath = binDirectoryFor (distDirLayout baseCtx)
+                                  (elaboratedShared buildCtx)
+                                  pkg
+                                  exe
+               </> exe
+    let args = drop 1 targetStrings
+    runProgramInvocation
+      verbosity
+      (simpleProgramInvocation exePath args)
```
ezyang: I don't think you want runProgramInvocation here; it doesn't propagate the exit code. Use `rawSystemExitWithEnv` instead, see `Distribution/Client/Run.hs` for how it is previously done.  You may also want to consider adding `LD_LIBRARY_PATH` if it is necessary.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: See (A).

cabal-install/Distribution/Client/CmdRun.hs
fgaz: (B2) No arguments means "treat the package in the cwd/project as a target, if there is one" (look at the comment above extractMatchingElaboratedConfiguredPackages). But this is also handled before, so same as (B1).
This will work after #4583 is merged.
I'll add a test for it too.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: (B1) Not by virtue of the take 1. For example if we specify a package with 2 exes in the targetStrings there will be two targets. But this is handled by lines 138-149 already, so you are correct. Should iI use an assert?

cabal-install/Distribution/Client/CmdRun.hs
fgaz: I did this because (see B1 and B2) we don't know anything about the targets at the beginning of the function, and an empty targetStrings could have a meaning. If i wrote a match for (command:args) and [] then I'd have to duplicate most of the code or put everything in an helper function.

the take/drop approach handles well the absence of a target component.

But now that i think of it, _what if_ we do
`new-run <no explicit exe here, there is only one in the package> args`? the first arg will be interpreted as a target and the command will fail.
But I don't think there is a way to handle this without ambiguities.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Will do.
Isn't LD_LIBRARY_PATH automated by the os or something?

cabal-install/Distribution/Client/CmdRun.hs
fgaz: If the there are multiple exes, then for each exe there will be an `ElaboratedConfiguredPackage` in the `ElaboratedInstallPlan`.

For each `ElaboratedConfiguredPackage` then, when `elabPkgOrComp` (see `executablesOfPackage`) is `ElabComponent` then there will be one exe, the matching one (it's a `Maybe`), but if it's an `ElabPackage` (which happens in custom builds), then that field is Nothing and we have to look at the `executables` field, which contains every exe in the package, hence the duplication.

I can't dedup before that point because i need both the package and the component name to do so (what if two packages have the seme exe name?).

Maybe I should add a comment about this. Or maybe it's a bug? Or should I add another field to `ElaboratedConfiguredPackage`/`ElaboratedInstallPlan`?

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Will merge.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Wait, did i just duplicate `resolveTargets`' functionality? >.<

23Skidoo:
> Revert "(,) is not Traversable on older base versions"

Note that you can also do `git rebase -i` and just remove that commit from your branch. This is not *super* important, but makes history cleaner and PRs a bit easier to review.

23Skidoo:
And by the way, can you please format your commit messages in the standard way? See this article: https://chris.beams.io/posts/git-commit/ for an explanation of what that means. No need to reword commit messages in this PR, but something to keep in mind for the future.

fgaz:
@23Skidoo
Sorry, I don't know why but I assumed the standard merging procedure here was to squash the commits, so I didn't pay much attention to the messages. Will do next time, thanks for the heads up.

fgaz:
Tests ready!
If they fail it's probably because of #4583

cabal-testsuite/PackageTests/NewBuild/CmdRun/MultipleExes/cabal.out
```diff
@@ -0,0 +1,19 @@
...
+# cabal new-run
+In order, the following will be built:
+ - MultipleExes-1.0 (exe:bar) (first run)
+Configuring executable 'bar' for MultipleExes-1.0..
+Preprocessing executable 'bar' for MultipleExes-1.0..
+Building executable 'bar' for MultipleExes-1.0..
+# cabal new-run
+Up to date
+# cabal new-run
+cabal: The run command is for running a single executable at once. The target '' refers to the package MultipleExes-1.0 which includes the executables foo and bar.
```
fgaz: And here's the need for #4583

cabal-install/Distribution/Client/CmdRun.hs
dmwit: How about `new-run -- args` for "run whatever executable is the default, with arguments `arg`s" and `new-run args` for "search for an executable named `args` and run it"? Separating arguments for a subprocess with `--` is a pretty common idiom.

cabal-testsuite/PackageTests/NewBuild/T4477/cabal.out
dmwit: The question of whether new-run should output its own information is a good one. My stance on "build or not" is that new-run should always rebuild the appropriate files, and new-exec should never rebuild anything (but should ensure that all the project's executables would be on the PATH if they were built). This makes the divide nice and clean, and supports both needs.

cabal-install/Distribution/Client/CmdRun.hs
dmwit: I think "multiple targets" is a bit misleading. It's more like a single, ambiguous target. Perhaps the error could be improved by mentioning longer target strings the user might want to try to disambiguate between the possibilities, e.g. "Ambiguous target my-package; could match my-package:my-exe1 or my-package:my-exe2" or "Ambiguous target my-exe; could match my-package1:my-exe or my-package2:my-exe". (Or am I misunderstanding when we would hit this branch?)

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +187,194 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+    justIfCondition f x = if f x then Just x else Nothing
+    match :: ElaboratedConfiguredPackage -> Bool
+    match p = matchPackage pkgId p && matchComponent component p
+
+matchPackage :: Maybe PackageId
+             -> ElaboratedConfiguredPackage
+             -> Bool
+matchPackage pkgId pkg =
+  pkgId == Just (elabPkgSourceId pkg)
+  || isNothing pkgId --if the package is unspecified (Nothing), all packages match
```
dmwit: Slight bikeshed: `matchPackage pkgId pkg = maybe True (elabPkgSourceId pkg ==) pkgId` seems to say what you want more concisely. Might be worth pulling out

    maybeMatches :: Eq a => a -> Maybe a -> Bool
    maybeMatches x = maybe True (x==)

since `matchComponent` has very similar code.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -314,6 +317,14 @@ extractMatchingElaboratedConfiguredPackages
   where
     match :: ElaboratedConfiguredPackage -> Bool
     match p = matchPackage pkgId p && matchComponent component p
+    matchingExecutable p = atMostOne
+                         $ filter (\x -> Just x == componentString
+                                      || isNothing componentString)
+                         $ executablesOfPackage p
+    componentString = componentNameString =<< component
+    atMostOne [x] = Just x
+    atMostOne _ = Nothing
```
dmwit: exactlyOne?

fgaz: or safeHead, or headMaybe...

yeah, that probably wasn't the best name :-P

cabal-install/Distribution/Client/CmdRun.hs
fgaz: @dmwit: as @ezyang says we shouldn't hit that branch. It's allalready handled with better error reporting before.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Looks ok, but I'd prefer to flip the arguments (so the "thing to search/filter for" is first as usual).

fgaz: I'm also tempted to change Maybe with something like
`data Match a = Anything | Exactly a`
but losing all the functions in Data.Maybe probably isn't worth it.

cabal-testsuite/PackageTests/NewBuild/T4477/cabal.out
fgaz: We also have -v0

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Good idea.

Cabal's flags are filtered from the targetStrings already, so it should simply be a matter of checking the first targetString

fgaz: No it isn't.

The first `--` is stripped from the targetStrings as well, probably because it marks the end of cabal flags and the beginning of the targets.

But if we document it appropriately we could use two of them, the first separating flags and targets, the second targets and targets' args

(actually its only use will be as a "placeholder" for the target when args are also needed, like @dmwit says)

fgaz: Ok, `-- --` is ugly and prevents future modification of the syntax, so I'll leave this as it is for now (args only when the exe/package is specified)

fgaz: btw, #3638 may be the origin of `-- --`

fgaz: Apparently even `new-run target -exearg --exearg ` doesn't work, as all `-`-prefixed args are interpreted as cabal args, while `new-run target -- --exearg` works.

Fixing this requires some modifications to the args parsing, but I don't think this is needed, as this was also the behaviour of old-run (except #4600).

dmwit: Keep in mind: modeling our behavior after old-style `run` is good to a point (because it lets us leverage folks' familiarity with the old UI); but as we're making lots of breaking UI changes here, this is also a good time to fix warts. So I think any behavior we choose should either be the right behavior or have a clear forward-compatibility story with the right behavior; backward-compatibility with the old, wrong behavior is less of a concern.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Apparently it does propagate it, as it uses `rawSystemExit` under the hood.

Cabal/doc/nix-local-build.rst
```diff
@@ -343,6 +349,26 @@ Currently, it is not supported to pass multiple targets to ``new-repl``
...
 each target.)
 
+cabal new-run
+-------------
+
+``cabal new-run [TARGET [ARGS]]`` runs the executable specified by the
+target, which can be a component, a package or can be left blank, as
+long as it can uniquely identify an executable within the project.
+
+See new-build for the target syntax.
```
23Skidoo: s/new-build/`new-build` section/
Can you also add a hyperlink here?

cabal-install/changelog
```diff
@@ -5,6 +5,9 @@
 	roll back the index to an earlier state.
 	* 'cabal new-configure' now backs up the old 'cabal.project.local'
 	file if it exists (#4460).
+	* Completeded the 'new-run' command (#4477). The functionality is the
```
23Skidoo: Typo: completeded.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -128,12 +155,181 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+                     "Multiple targets given"
+
+    let elaboratedPlan = elaboratedPlanOriginal buildCtx
+        matchingElaboratedConfiguredPackages =
+          extractMatchingElaboratedConfiguredPackages
+            selectedPackage
+            selectedComponent
+            elaboratedPlan
+
+    -- the names to match. used only for user feedback, as
```
23Skidoo: Please use proper punctuation in comments (i.e. start sentences with capital letters, end them with full stops, and so on).

cabal-install/Distribution/Client/CmdRun.hs
23Skidoo: Do we have a test case that checks that `new-run` propagates the exit code?

23Skidoo:
Would be nice if @dcoutts could take a look too.

fgaz:
Once I fix those minor problems there are two issues I'd like to address before merging:

**The args issue.**

I thought about it, and i think the old-run command solved the problem well. I often add a cabal flag after the exe name, and the `--` syntax is standard. This conflicts with the `<empty-target> -- args` syntax which I don't think is that useful, so in my opinion we should keep it as it is.

**Returning the target from `runProjectPreBuildPhase`**
There is duplication between the logic in the function passed to `runProjectPreBuildPhase` and around line 170. We check for multiple/no targets two times (the second one assuming everything will go well) when we could simply return the single target from `runProjectPreBuildPhase`.

But in the only function we pass to it we can only alter the `ElaboratedInstallPlan`, so i think we should change its type to one of the following:

*  `.... -> (ElabInstPlan -> (ElabInstPlan, arbitraryData) -> ... -> IO (ProjectBuildContext, arbitraryData)`
  which is probably bad because that type variable only calls for misuse
*  `.... -> (ElaboratedInstallPlan -> (ElaboratedInstallPlan, Targets) -> ... -> IO ProjectBuildContext`
*  `.... ((ElaboratedInstallPlan, ProjectBuildContext )-> (ElaboratedInstallPlan, ProjectBuildContext) -> ... -> IO ProjectBuildContext`
  and put the target in a new field in `ProjectBuildContext`.
  This way future modifications are also easy.

I think @dcoutts has some more [info](https://github.com/haskell/cabal/issues/3638#issuecomment-248969464) about this.

This last issue is only an implementation problem and does not affect the behaviour of new-run, so if you are all ok with this I'll open another issue for this and finally merge this one (once I've fixed those typos and tests >.<) so we can have a working `new-run` and solve that problem when we know more.

cabal-install/Distribution/Client/CmdRun.hs
fgaz: Now we have one

fgaz:
All done here, I'm merging this by tomorrow. I'll also open another pr for the duplication issue.

PR #4592, opened 2017-07-07
WIP: Implement `cabal check` for consistent .cabal naming

hvr:
Hackage requires the .cabal file to be named consistently with the package name,
but `cabal check` didn't detect this.

Example output:

    $ cabal check
    The following errors will cause portability problems on other environments:
    * The filename ./doo.cabal does not match package name (expected: foobar.cabal)

TODO: changelog, testsuite, maybe clarification in user's guide


23Skidoo:
LGTM.

Issue #4550, opened 2017-06-04
Too many invocations of ar


sergv:
tl;dr version: Cabal currently uses slow approach to creating `.a` archives by repeatedly calling `ar` on chunks of object files produced by `ghc`. It should instead pass all the the object files together to one invocation of `ar` via `@file` argument.

Long version: building [`skylighting`](http://hackage.haskell.org/package/skylighting) package on Debian Linux with ghc `8.0.2` and `--enable-split-objs`, produces around 150000 object files for non-profiling build (with profiling enabled, it's double than that). These object files are produced reasonably fast, but then Cabal takes a long time to combine them together into an `.a` archive. In order to do that, Cabal splits all 150000 object files into chunks of around 300 each and then calls `ar qc target.a file1.o file2.o ... file330.o` command for each chunk, which incrementally updates the archive. This process takes time comparable with time needed to produce all the object files (if not more), not in the least because at each invocation the `ar` utility must read the current archive, append 300 new files to it and then write it back, which is pretty much *quadratic complexity*! Looking at process manager it seems to be the case indeed because each successive `ar` invocation uses around 5 Mb more memory that previous one. If you throw in slow spinning hdd that object files and target archive are located on, the whole process takes really long time to complete.

It seems possible to avoid the overhead of chunking and significantly speed up production of the `.a` archive by using only one `ar` invocation and passing all object files in one go. Surely, all 150000 object files would not fit on the command line, but fortunately the `ar` tool supports `@file` options, which Cabal could use to pass all the object files.

Also, I think this way of passing object files should become the default one because `ar` seems to have this option for quite some time already, so it should work well with reasonable amount of previous versions of `ghc` and `binutils`, while still older versions can fall back to current mechanism. I managed to google `ar` man page from 2007 that claims to support `@file` option already: http://www.thelinuxblog.com/linux-man-pages/1/ar, which should cover all ghcs down to 6.12. I'm not sure about older versions though.

So, the question is: is `ar` invocation with `@file` argument is a reasonable thing to add to Cabal-the-library?

23Skidoo:
[...]

PR #4596, opened 2017-07-08
Leave single invocation of ar

sergv:
Fix for #4550.

* [X] Old `DeterministicAr` test split in two to test new and old paths.


Cabal/Distribution/Simple/Compiler.hs
```diff
@@ -339,6 +340,10 @@ libraryDynDirSupported comp = case compilerFlavor comp of
  where
   v = compilerVersion comp
 
+-- | Does this compiler's "ar" command supports @file-style arguments
+arAtFileArgumentSupported :: Compiler -> Bool
```
23Skidoo: I suggest that we standardise on calling them "response files" in our codebase, since there's existing precedent (`haddockSupportsResponseFiles`).

Cabal/Distribution/Simple/Program/Ar.hs
```diff
@@ -83,10 +86,29 @@ createArLibArchive verbosity lbi targetPath files = do
...
+        simpleArgs ++ extraArgs ++ ['@' : atFile]
+
+  if oldVersionManualOverride || atArgumentNotSupported
+    then
+      sequence_
         [ runProgramInvocation verbosity inv
         | inv <- multiStageProgramInvocation
                    simple (initial, middle, final) files ]
+    else
+      withTempFile tmpDir "object-files.txt" $ \argsPath argsHandle -> do
```
23Skidoo: Can we use `withTempFileEx` here, so that `--keep-tmp-files` would work?

23Skidoo:
There's also a minor merge conflict.

Cabal/Distribution/Simple/Program/Ar.hs
sergv: I'm not sure there's a `--keep-tmp-files` flag in cabal-install. There's only `--keep-temp-files` in `cabal haddock`, but that's not relevant I reckon.

Surely I can put the `Ex` version there, but unfortunately I don't see what to pass to it except `defaultTempFileOptions`, which has temp files disabled unconditionally.

Cabal/Distribution/Simple/Compiler.hs
sergv: Good point, I'll rename everything and amend the commits.

sergv:
Actually, there's windows test failure and it's not random. My first version didn't account for response file format, which expects a bunch of whitespace-separated arguments. Thus, without quoting filenames with whitespace in them will not be processed correctly and Windows will be broken because it uses backslashes as path separators which are interpreted as quoting in the response file syntax. Similar issues were discussed in ghc when response files were added. The discussion in https://phabricator.haskell.org/D1158 is relevant, but I've summarized it already - the bottom line is that response files have each file name quoted and surrounded in quotes (double or single).

sergv:
Updated, should be mergeable if tests pass.

Cabal/Distribution/Simple/Program/Ar.hs
23Skidoo: OK. We should create a ticket about adding `--keep-tmp-files` to `GlobalFlags`.

Cabal/Distribution/Simple/Program/Ar.hs
```diff
@@ -83,10 +86,29 @@ createArLibArchive verbosity lbi targetPath files = do
...
+        simpleArgs ++ extraArgs ++ ['@' : atFile]
+
+  if oldVersionManualOverride || responseArgumentsNotSupported
+    then
+      sequence_
         [ runProgramInvocation verbosity inv
         | inv <- multiStageProgramInvocation
                    simple (initial, middle, final) files ]
+    else
+      withTempFile tmpDir ".rsp" $ \responeFilePath responseFileHandle -> do
```
ezyang: s/responeFilePath/responseFilePath/

Cabal/Distribution/Simple/Program/Ar.hs
```diff
@@ -97,9 +119,26 @@ createArLibArchive verbosity lbi targetPath files = do
...
+-- Namely, on Windows standard path separator is backspace but in @ar@
+-- response file format it's escape sequence. Therefore backslashes must
+-- be escaped as well.
+escapeFileName :: FilePath -> FilePath
+escapeFileName = concatMap escapeChar
+  where
+    escapeChar :: Char -> String
+    escapeChar '"'  = "\\\""
+    escapeChar '\\' = "\\\\"
+    escapeChar c    = [c]
```
ezyang: Comparing with https://git.haskell.org/ghc.git/commitdiff/296bc70b5ff6c853f2782e9ec5aa47a52110345e it seems you also need to handle newlines. Also, I am a bit perplexed: it doesn't look like you are ever wrapping this with quotes.

It would be really helpful to have a test case for these edge cases.

sergv: That's good catch, however it seems unplausible that file name will contain newlines. I'm not 100% sure that `ar` will like those newlines either.

Cabal/Distribution/Simple/Program/Ar.hs
sergv: Good catch, thanks!

Cabal/Distribution/Simple/Program/Ar.hs
23Skidoo: We have some code for producing response files in `D.S.Haddock` already, maybe we can consolidate it all under, say, `D.S.Program.ResponseFile`?

ezyang: I'd advise against it: the implementations of the response file parsers are different so it logically makes sense for there to be two implementations to (possibly) need to be different.

23Skidoo: Maybe we can have both impls in a single file at least? I envision something like `withResponseFile (defaultResponseFileOptions { responseFileEscapingStyle = ArEscaping }) args $ \fname -> ...`

ezyang: Well, the code in any case is pretty short...

sergv: I think it would be best to unify the two under single implementation in `Distribution.Simple.Program.ResponseFile`. Haddock's module has following comment

```
      -- Support a gcc-like response file syntax.  Each separate
      -- argument and its possible parameter(s), will be separated in the
      -- response file by an actual newline; all other whitespace,
      -- single quotes, double quotes, and the character used for escaping
      -- (backslash) are escaped.  The called program will need to do a similar
      -- inverse operation to de-escape and re-constitute the argument list.
```

which implies that it tries to match gcc escaping rules and that's precisely what we're trying to achieve for `ar`, and possibly `ld`.

sergv:
A couple of fixes, as discussed previously (can squash them later):
- I factored out response file handling into `Distribution.Simple.Program.ResponseFile`. The handling is taken from current haddock response file handling and should handle even newlines in file names
- I made `ld` invocation that creates library for ghci also use response files via heuristic mentioned in #4550 
- I renamed `--ar-does-not-support-response-files` argument to `configure` and `install` into `--disable-response-files` since it now affects both `ar` and `ld`. It does not affect haddock as it uses haddocks' version to detect whether response files are supported, which is pretty reliable

sergv:
AppVeyor fails with unclear error message when running tests:
```
dist\build\cabal-tests\cabal-tests.exe -j3
threads: 3
tests to run: 186
PackageTests\AllowOlder\cabal.test.hs                                                                             SKIP (1.53s)
PackageTests\AllowNewer\cabal.test.hs                                                                             SKIP (1.73s)
Shutting down GHCi sessions (please be patient)...
Unexpected failure on GHCi exit: fd:3: hClose: resource vanished (Broken pipe)
cabal-tests.exe: fd:3: hClose: resource vanished (Broken pipe)
Command exited with code 1
```

Is it my changes or some intermittent failure?

ezyang:
It shouldn't do that. I restarted the build.

23Skidoo:
> AppVeyor fails with unclear error message when running tests:
>
> [...]

This is #4618.

sergv:
The build is green. Should we merge?

23Skidoo:
Merged, thanks!

sergv:
Great, thanks! :+1:

Issue #4580, opened 2017-07-03
QuasiQuotes should trigger building TH libraries


duog:
I don't fully understand the nuances, but in several places in Distribution.Cabal.Simple.GHC[(JS)] the extensions specified in a cabal file are checked for TemplateHaskell, and if it is present then object files are build for ghci, to enable the template haskell to be typechecked. By the same logic, the presence of QuasiQuotes should trigger the same behaviour.

ezyang:
Do you have an example program where this is causing a problem? I actually *don't* think it is necessary to build TH objects for just QuasiQuotes, because QuasiQuotes by themselves cannot trigger splices to build. For this very reason, QuasiQuotes are enabled on a stage 1 GHC compiler, when TemplateHaskell is not (that is why I added this pragma in the first place!)

duog:
I may be missing something, but I'm pretty sure QuasiQuotes can trigger splices to build, see examples in the docs here
[here](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#th-quasiquotation), are you thinking of TemplateHaskellQuotes?

[This](https://phabricator.haskell.org/D3677) patch contains and fixes a test case that failed due to a QuasiQuote needing an object file.

I have briefly tried to get language-quote-0.12 to fail to build with
```
cabal unpack language-c-quote-0.12
cd language-c-quote-0.12
cabal install --only-dependencies
cabal configure --disable-shared
cabal build
```
However, I can't see how to cajole ghc+cabal into not producing .dyn_o object files. I believe that if one could find such an incantation then this error would surface.

hvr:
@ezyang are you sure you don't mean `TemplateHaskellQuotes`?

ezyang:
Yes, you are right! So this is a legitimate bug, but it also sounds easy to fix: just find where we have `TemplateHaskell` hard coded and generalize it. @duog would you like to take a shot?

duog:
Yes I'd be happy to.

ezyang:
OK, I added you to the collaborators on Cabal. Looking forward to your patch!

PR #4598, opened 2017-07-10
Refine "doingTH" check to include checking for QuasiQuotes

duog:
Addresses #4580. I have not added a test for this. I could easily modify or duplicate the tests in cabal-tests/PackageTests/TemplateHaskell if you think it worthwhile.
Note that ghc-8.2 is smart enough to not need this logic, I'm not sure when those smarts were added.
I verified this works by modifying those tests to have QuasiQuotes in their cabal files, and running them with ghc-7.6.3.

### DIFF
--- a/Cabal/Distribution/PackageDescription.hs
+++ b/Cabal/Distribution/PackageDescription.hs
@@ -77,6 +77,7 @@ module Distribution.PackageDescription (
         allLanguages,
         allExtensions,
         usedExtensions,
+        usesTemplateHaskellOrQQ,
         hcOptions,
         hcProfOptions,
         hcSharedOptions,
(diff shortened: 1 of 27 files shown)


ezyang:
LGTM. Can we get a test? I can assist if it is non-obvious but https://github.com/haskell/cabal/tree/master/cabal-testsuite should give you a start.

Also, minor bikeshedding: shouldn't `usesTemplateHaskellOrQQ` be named based on the invariant it checks, rather than its implementation details? That would make it more robust to future changes. (This is easy enough to fix later so don't feel obligated to do anything.)

duog:
Test won't be a problem, I'll just copy tests/PackageTests/TemplateHaskell. As I noted above, It would only actually fail on older ghcs though.

I agree it should be based on the invariant it checks, but I couldn't come up with anything better. The analogous function in ghc is named needsTemplateHaskellOrQQ. I used "uses" to match the vocabulary of surrounding functions.

23Skidoo:
LGTM as well, merging!

23Skidoo:
Merged, thanks!

PR #4603, opened 2017-07-14
Suggest fix for missing deps in error message

m-renaud:
A common reason for `cabal configure` to fail with the "Encountered missing dependencies: ..." error is forgetting to run `cabal install` beforehand. This change improves the error message by suggesting a fix for the problem.

**Previous error:**

```
$ cabal configure
Resolving dependencies...
<snip>
cabal: Encountered missing dependencies:
turtle -any
```

**New error:**

```
$ cabal configure
Resolving dependencies...
<snip>
cabal: Encountered missing dependencies:
turtle -any
Have you run 'cabal install'?
```

**Testing:**
* cabal-testsuite/PackageTests/ConfigureComponent/SubLib/setup-explicit-fail.out has been updated for the new error message.

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -910,6 +910,7 @@ configureFinalizedPackage verbosity cfg enabled
                      ++ (render . nest 4 . sep . punctuate comma
                                 . map (disp . simplifyDependency)
                                 $ missing)
+                     ++ "\nHave you run 'cabal install'?"
 
     -- add extra include/lib dirs as specified in cfg
     -- we do it here so that those get checked too
--- a/cabal-testsuite/PackageTests/ConfigureComponent/SubLib/setup-explicit-fail.out
+++ b/cabal-testsuite/PackageTests/ConfigureComponent/SubLib/setup-explicit-fail.out
@@ -11,3 +11,4 @@ Registering library 'sublib' for Lib-0.1.0.0..
 Configuring executable 'exe' for Lib-0.1.0.0..
 setup: Encountered missing dependencies:
     sublib -any
+Have you run 'cabal install'?


ivan-m:
I would have something more like `cabal install --dependencies-only` (possibly with testing and benchmark options enabled).

BardurArantsson:
I'm a bit torn on this since "install" will *currently* likely not do what the user *really* wants which is to have a sandbox (or, eventually, use new-build, but at *that* point it'll be fine to just "cabal install", I think?).

Btw, please squash these two commits -- they are really a single logical "unit".

m-renaud:
> cabal install --dependencies-only

Good idea, I'll add `--dependencies-only`. I'm thinking maybe omit testing and benchmark since the prompt to run install will probably be enough to get things working and they can run reinstall with benchmarks and testing later if need be. WDYT?

> since "install" will currently likely not do what the user really wants which is to have a sandbox

Ah, I was operating under the assumption that they were already in a sandbox.

I usually run into this error when I added a new dependency to my .cabal file. If you run `cabal build` (which indirectly runs configuration if your cabal file has changed) then it fails with the missing dependency error, even if you're in a sandbox.

> please squash these two commits

Will do.

BardurArantsson:
If you could make the extra message text dependent on whether the user is in a sandbox, then it's definitely an obvious +1 from me! :)

m-renaud:
> If you could make the extra message text dependent on whether the user is in a sandbox

I dug around in the code a little bit when I was making the initial change but didn't find where that info was stored. Do you have any suggestions/code pointers for where to look? Thanks!

BardurArantsson:
That's a good question. I'm afraid I'm probably so out-of-date with the codebase that I couldn't point you to exactly where, but perhaps grepping around a bit for "whenUsingSandbox", "UseSandbox" might be helpful.

23Skidoo:
> Do you have any suggestions/code pointers for where to look? Thanks!

Your commit touches only the Cabal library code, all sandbox-related stuff lives under `cabal-install/`.

m-renaud:
Closing this since this applies to the old configure command.

PR #4607, opened 2017-07-17
Reduce duplication in CmdRun

fgaz:
followup of #4586

Previously, the extraction of the target from the targetStrings was done
twice: inside the function passed to 'runProjectPreBuildPhase' and right
after it.

Now 'runProjectPreBuildPhase' is able to return information from that
function, and the duplication was removed, along with some assumptions
about the number of targets.

---

This is one of the possible approaches, the other being adding a field in ProjectBuildContext.
The only way to avoid a Maybe or a sum type in said field, is parameterizing it.

```haskell
data ProjectBuildContext arbitraryData = ProjectBuildContext{
  ...
  someFieldName :: arbitraryData
}
```
Which has the advantage of not having to modify all the Cmd*.hs files and passing `()`s everywhere, but after some thinking I like the first solution the best.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -226,8 +226,13 @@ data ProjectBuildContext = ProjectBuildContext {
 runProjectPreBuildPhase
     :: Verbosity
     -> ProjectBaseContext
-    -> (ElaboratedInstallPlan -> IO ElaboratedInstallPlan)
-    -> IO ProjectBuildContext
+    -- ^ A function that selects a subset of the plan and returns it,
+    -- It can also return some arbitrary data, which can be useful in case
+    -- the caller needs it.
+    -> (ElaboratedInstallPlan -> IO (ElaboratedInstallPlan, a))
+    -- ^ The build context, along with the data which was returned by
+    -- the the argument function.
+    -> IO (ProjectBuildContext, a)
 runProjectPreBuildPhase
     verbosity
     ProjectBaseContext {
@@ -253,7 +258,7 @@ runProjectPreBuildPhase
     -- Now given the specific targets the user has asked for, decide
     -- which bits of the plan we will want to execute.
     --
-    elaboratedPlan' <- selectPlanSubset elaboratedPlan
+    (elaboratedPlan', a) <- selectPlanSubset elaboratedPlan
 
     -- Check which packages need rebuilding.
     -- This also gives us more accurate reasons for the --dry-run output.
@@ -267,12 +272,12 @@ runProjectPreBuildPhase
                              pkgsBuildStatus elaboratedPlan'
     debugNoWrap verbosity (InstallPlan.showInstallPlan elaboratedPlan'')
 
-    return ProjectBuildContext {
+    return (ProjectBuildContext {
       elaboratedPlanOriginal = elaboratedPlan,
       elaboratedPlanToExecute = elaboratedPlan'',
       elaboratedShared,
       pkgsBuildStatus
-    }
+    }, a)
 
 
 -- | Build phase: now do it.
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -217,7 +217,11 @@ data ProjectBuildContext = ProjectBuildContext {
 
       -- | The result of the dry-run phase. This tells us about each member of
       -- the 'elaboratedPlanToExecute'.
-      pkgsBuildStatus        :: BuildStatusMap
+      pkgsBuildStatus        :: BuildStatusMap,
+
+      -- | The targets selected by @selectPlanSubset@. This is useful eg. in
+      -- CmdRun, where we need a valid target to execute.
+      targetsMap             :: TargetsMap
     }
 
 
@@ -226,13 +230,8 @@ data ProjectBuildContext = ProjectBuildContext {
 runProjectPreBuildPhase
     :: Verbosity
     -> ProjectBaseContext
-    -- ^ A function that selects a subset of the plan and returns it,
-    -- It can also return some arbitrary data, which can be useful in case
-    -- the caller needs it.
-    -> (ElaboratedInstallPlan -> IO (ElaboratedInstallPlan, a))
-    -- ^ The build context, along with the data which was returned by
-    -- the the argument function.
-    -> IO (ProjectBuildContext, a)
+    -> (ElaboratedInstallPlan -> IO (ElaboratedInstallPlan, TargetsMap))
+    -> IO ProjectBuildContext
 runProjectPreBuildPhase
     verbosity
     ProjectBaseContext {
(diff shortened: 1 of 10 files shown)


cabal-install/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -223,8 +223,13 @@ data ProjectBuildContext = ProjectBuildContext {
 runProjectPreBuildPhase
     :: Verbosity
     -> ProjectBaseContext
-    -> (ElaboratedInstallPlan -> IO ElaboratedInstallPlan)
-    -> IO ProjectBuildContext
+    -- ^ A function that selects a subset of the plan and returns it,
+    -- It can also return some arbitrary data, which can be useful in case
+    -- the caller needs it.
+    -> (ElaboratedInstallPlan -> IO (ElaboratedInstallPlan, a))
```
23Skidoo: Why do you need polymorphism here if you're only ever using this value in monomorphic context -- it's either ignored (`()`) or `(UnitId, ComponentId)`? IMO it makes more sense to stick a `(UnitId, ComponentId)` member into the `ProjectBuildStatus` record.

dmwit: I don't have a strong preference on "return directly" vs. "stuff it in the `ProjectBuildStatus`, but for polymorphic vs. monomorphic, the polymorphic version is nice for forward-compatibility, e.g. if future consumers want some other information out of the pre-build phase target selection process.

If we stick with "return directly", perhaps there should be, like, a `runProjectPreBuildPhase_ f = fst <$> runProjectPreBuildPhase (\x -> (f x, ())` or something like that which has the old type. Then existing uses would be a bit less noisy.

fgaz: Well, if I add another field it has to be `Maybe (UnitId, ComponentId)` or equivalent, because that record is used in other commands too. But then I'd have to unwrap that `Maybe` and rely on the assumption that I set it to  `Just` in preBuild.

fgaz: Oh, right, a wrapper.
Then the only question is whether to add the _ to the wrapper or the wrapped. In `base` it's the wrapper I think

23Skidoo: I'm wary of over-abstracting the code prematurely for the sake of future consumers that may never materialise. If they do, refactoring is easy in Haskell. Existing type is nice because there's only one place you need to look at to see the data that's produced in the pre-build step.

> But then I'd have to unwrap that `Maybe` and rely on the assumption that I set it to  `Just` in preBuild.

That's okay IMO, just make it a fatal error if you get `Nothing`.

fgaz: Note that in that case we still have to modify all call sites to return `Nothing`. 
```haskell
... -> (ElaboratedInstallPlan -> IO (ElaboratedInstallPlan, Maybe (UnitId, ComponentId))) -> BuildCtx
```
Or use a wrapper. In the case of a monomorphic field, a *run-specific* wrapper, which I don't really like.

---

> there's only one place you need to look at to see the data that's produced in the pre-build step.

You mean the return type? I think I can agree with that, I'll add a field.

---

I'd like to see the opinion of others too on this (the polymorphism). Github suggests @dcoutts and @hvr. Even just a +1 here or there.

23Skidoo: > Note that in that case we still have to modify all call sites to return `Nothing`.

Hmm, I wonder if we should stick the whole `TargetMap` inside `ProjectBuildContext`. That would be potentially more useful to non-`new-run` clients (e.g. `new-bench`/`new-test`, where I think we want the ability to run several benchmarks/tests at once), and we could set it to something meaningful other than `Nothing` in those cases.

We'd need to do the check that there's only a single target twice in `new-run` code, though (makes sense to add a helper for this).

> You mean the return type? 

Yep, the definition of `ProjectBuildContext`.

23Skidoo: BTW, it probably makes sense to you to tackle `new-bench` and `new-test` after this.

fgaz: Isn't `new-test` already done, according to that issue (also, it works for me)?

Or we have to redo it, since the current one relies on the "old" Setup, as most of the new-commands do (I think. They seem to do so.)?

`new-ench` builds but does not run the benchmark, same as run before my pr.

23Skidoo: Sorry, you're right, `new-test` is implemented using the setup wrapper because it needs to run hooks. `new-bench` can have custom hooks attached too and should also work this way.

`new-test` seems to be confused by non-trivial project structure, though, so it still makes sense to look at it:

```
~/code/haskell/cabal[master]$ cabal new-test cabal-install:unit-tests               
cabal: Cannot run the test suite 'unit-tests' because it is marked as
'buildable: False' within the 'cabal-install.cabal' file (at least for the
current configuration). If you believe it should be buildable then check the
.cabal file to see if the buildable property is conditional on flags.
Alternatively you may simply have to edit the .cabal file to declare it as
buildable and fix any resulting build problems.

~/code/haskell/cabal[master]$ cabal new-test cabal-install:unit-tests --enable-tests
cabal: Cannot run the test suite 'unit-tests' because it is marked as
'buildable: False' within the 'cabal-install.cabal' file (at least for the
current configuration). If you believe it should be buildable then check the
.cabal file to see if the buildable property is conditional on flags.
Alternatively you may simply have to edit the .cabal file to declare it as
buildable and fix any resulting build problems.

~/code/haskell/cabal[master]$ cabal new-configure --enable-tests   
'cabal.project.local' file already exists. Now overwriting it.
Assertion failed
  assert, called at ./Distribution/Client/ProjectPlanning.hs:280:5 in main:Distribution.Client.ProjectPlanning
```

23Skidoo: Anyhow, I still think that sticking `TargetMap` into `ProjectBuildContext` makes sense.

fgaz: Hmm, so I'll fall back to the one-element assumption, but at least there isn't much duplication now.

I still think the polymorphic solution is cleaner-ish, but I'll proceed with your solution if the others don't chime in in a couple of days

ezyang:
Hey @fgaz, do you need me to look at this commit? It looks like a bunch of people commented already and I don't want to pile on. Let me know.

fgaz:
@ezyang
Just wanted a third opinion on polymorphism (see above), since @dmwit and @23Skidoo have opposite ones and I don't trust myself enough to make the final decision.

(P.S.: I replied by email some hours ago but my comment didn't show up here. Just a heads up in case this happens to any of you)

ezyang:
Let me preface this by saying, it doesn't matter, ship something and be willing to refactor it later if it needs to be done differently.

That being said, let me be philosophical for a moment. The crux of the problem is that the logical interpretation of `runProjectPreBuildPhase` returning a `ProjectBuildContext` (which represents the "result" of pre-build") is in conflict with the fact that, from an implementation perspective, the easiest place to say what targets are selected are during the pruning process.

I think the approach which is most faithful to the original intent of the code is to refine `(ElaboratedInstallPlan -> IO ElaboratedInstallPlan)` to not ONLY return the pruned `ElaboratedInstallPlan`, but also a description of how it was pruned, which, *incidentally*, is going to contain the target information you are interested in here.

23Skidoo:
Needs to be rebased.

fgaz:
I'm going to merge once it's green

23Skidoo:
@fgaz Thanks! Very nice to be able to use `new-run` at last.

PR #4610, opened 2017-07-19
Add datadir env var before running the exe

### LINKED ISSUE #4120: Nix-style builds need to install data files for inplace builds
Currently, `new-build` does not install `data-files` (possibly also `extra-files`). A sufficient condition for this bug is an "inplace" build. It may happen in other cases as well.

fgaz:
This enables the executale to find the datafiles in inplace builds.

Fixes #4120 

* [x] Any changes that could be relevant to users have been recorded in the changelog. (this is part of new-run)
* [x] Added a test for this.


Mikolaj:
FYI: this doesn't work if you have cabal.project and different exes use data files listed in their respective .cabal files. :(

fgaz:
@Mikolaj thanks for reporting the bug. looking into it.

PR #4617, opened 2017-07-21
Add staticlib

angerman:
This adds support for GHCs `-staticlib` flag, which produces a static library that has all dependencies rolled into a massive archive.  This is rather useful when building libraries that are supposed to be deployed and linked. `-staticlib` is only available for macOS/iOS in GHC 8.2 and earlier, due to the reliance on `libtool`. With [D3721](https://phabricator.haskell.org/D3721), this will be available in GHC 8.4+ for all platforms.

### DIFF
--- a/Cabal/Distribution/PackageDescription.hs
+++ b/Cabal/Distribution/PackageDescription.hs
@@ -81,6 +81,7 @@ module Distribution.PackageDescription (
         hcOptions,
         hcProfOptions,
         hcSharedOptions,
+        hcStaticOptions,
 
         -- ** Supplementary build information
         ComponentName(..),
--- a/Cabal/Distribution/PackageDescription/Parse.hs
+++ b/Cabal/Distribution/PackageDescription/Parse.hs
@@ -510,6 +510,8 @@ binfoFieldDescrs =
            sharedOptions      (\val binfo -> binfo{sharedOptions=val})
  , optsField   "ghcjs-shared-options" GHCJS
            sharedOptions      (\val binfo -> binfo{sharedOptions=val})
+ , optsField   "ghc-static-options" GHC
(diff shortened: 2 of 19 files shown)


angerman:
I'm specifically looking for early feedback on the the implementation!

ezyang:
This looks pretty good, looking forward to it merging!

If you are in a refactoring mood, one thing that might be worth looking at is whether or not the logic for flag computation for what we pass to GHC isn't worth rejiggering a little. It seems to accumulate bit by bit every time we add a new mode of invoking GHC. Definitely not required though.

angerman:
I'm a bit confused by the travis failures. It looks like when compiling a custom `setup.hs`, the `--disable-static` flag can not be parsed?!  Is it possible that the `setup.hs` links an older Cabal library?!

angerman:
@ezyang can we merge this, only one of the travis CI fails with:
```
whole job took 2495 seconds so far.
Job taking over 40 minutes. Terminating
```
which I believe is independent of this PR?

PS: I don't feel like refactoring much just yet.

23Skidoo:
@angerman 

There should be a `cabal-version >= 2.1` version check in `Distribution.PackageDescription.Check` for the `ghc-static-options` field. And can you please also update the manual?

angerman:
@23Skidoo I believe the `Distribution.PackageDescription` changes were incorrect. I do not believe it makes sense to put those into the package description. I removed those.

Also added the docs.

angerman:
And again, the CI failures are due to timeout:
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```.

ezyang:
I restarted the Travis builds, maybe it was a cold cache run so it took too long.

23Skidoo:
It's green now, should be good to go.

PR #4605, opened 2017-07-15
Print build profile in new-build command

amir:
Please include the following checklist in your PR:

* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/#conventions).
* [x] Any changes that could be relevant to users have been recorded in the changelog.
* [x] The documentation has been updated, if necessary.

Prints build profile—at the moment only compiler and optimization level. Sample output:
```
Build profile: with-compiler: ghc-7.10.3, optimization: MaximumOptimisation
```

With this `cabal.project.local`:
```
optimization: 2
```

This is a naive attempt to address #3945. I'm not only new to Cabal but also to Haskell. I'd gladly follow any hints by a mentor to get this merged but feel free to discard it if there's absolutely no hope.

PR #4620, opened 2017-07-21
Further compact build profile output in new-build

amir:

This change will further compact the newly added Build Profile to new-build command in #4605 per @hvr's comment: https://github.com/haskell/cabal/pull/4605#issuecomment-315975829

Before this change:
```
Build profile: with-compiler: ghc-7.10.3, optimization: MaximumOptimisation
```

And after this change:
```
Build profile: -w ghc-7.10.3 -O2
```

### DIFF
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -767,10 +767,14 @@ printPlan verbosity
     showMonitorChangedReason  MonitorFirstRun     = "first run"
     showMonitorChangedReason  MonitorCorruptCache = "cannot read state cache"
 
-    showBuildProfile = "Build profile: " ++ (intercalate ", " [
-      "with-compiler: " ++ (showCompilerId . pkgConfigCompiler) elaboratedShared,
-      "optimisation: " ++ (show (fromMaybe NormalOptimisation (Setup.flagToMaybe packageConfigOptimization)))]
-      ) ++ "\n"
+    showBuildProfile = "Build profile: " ++ unwords [
+      "-w " ++ (showCompilerId . pkgConfigCompiler) elaboratedShared,
+      "-O" ++  (case packageConfigOptimization of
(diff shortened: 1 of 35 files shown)


cabal-install/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -767,10 +767,14 @@ printPlan verbosity
...
     showMonitorChangedReason  MonitorCorruptCache = "cannot read state cache"
 
-    showBuildProfile = "Build profile: " ++ (intercalate ", " [
-      "with-compiler: " ++ (showCompilerId . pkgConfigCompiler) elaboratedShared,
-      "optimisation: " ++ (show (fromMaybe NormalOptimisation (Setup.flagToMaybe packageConfigOptimization)))]
-      ) ++ "\n"
+    showBuildProfile = "Build profile: " ++ unwords [
+      "-w " ++ (showCompilerId . pkgConfigCompiler) elaboratedShared,
+      "-O" ++  (case packageConfigOptimization of
+                Setup.Flag NoOptimisation      -> "0"
```
23Skidoo: Don't we already have a utility function for rendering this type?

amir: there probably is one. will start looking for it.

23Skidoo: No, it looks like we don't.

cabal-install/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -767,10 +767,14 @@ printPlan verbosity
...
+                Setup.Flag NormalOptimisation  -> "1"
+                Setup.Flag MaximumOptimisation -> "2"
+                _                              -> "1")]
```
23Skidoo: I'd prefer an exhaustive pattern-match instead of a catch-all clause. That way we'll get a warning about an inexhaustive pattern match when the data type changes.

23Skidoo:
Merged, thanks!

PR #4641, opened 2017-07-28
Use package root as data-file base path, not cwd

### LINKED ISSUE #4639: Data files not working in new-build projects
Here's the problem: https://github.com/haskell/cabal/blob/484bb6c/cabal-install/Distribution/Client/CmdRun.hs#L211

We use the current directory as the base path for data files in `new-run`, but "current directory == package root" was true only before new-build.

Replacing `curDir` with the actual package root should fix this.

fgaz:
Using the current directory worked with single packages, but cabal
projects can use packages in different places.

Fixes #4639

todo:
* [x] ~~A meaningful error message~~ It's a comment in a pure function that returns a Maybe now.
* [x] Modify the cmdrun/datafiles test to take this into account
* [x] Take into account local deps' datafiles


ezyang:
I fixed an upstream CI problem, if you try a rebuild that should help.

fgaz:
https://travis-ci.org/haskell-pushbot/cabal-binaries/builds/258926423
wat.

ezyang:
Looks like QuickCheck uncovered an unrelated bug ;) File an issue, will you?

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -206,23 +206,53 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags)
...
+-- | Construct an environment variable that points
+-- the package's datadir to its correct location.
+-- This might be:
+-- * 'Just' the package's source directory plus the data subdirectory
+--   for inplace packages.
+-- * 'Nothing' for packages installed in the store (the path was
+--   already included in the package at install/build time).
+-- * The other cases are not handled yet. See below.
+dataDirEnvVarForPackage :: ElaboratedConfiguredPackage
+                        -> Maybe (String, Maybe FilePath)
```
hvr: The docstring refers to the outer `Maybe`, but what about the inner `Maybe` in the 2nd element?

fgaz: That's the format that runProgramInvocation accepts. See `progInvokeEnv` in `Cabal/Distribution/Simple/Program/Run.hs`.

The inner Maybe is for valueless vars I think.

Oddly, we don't have an alias for it.

PR #4650, opened 2017-07-31
Use cabal-install 2.0 on Travis for the 2.0 branch as well

23Skidoo:
This is f3ade3fea60a8eadc90c0b0eaba20bcca7aceb1d cherry-picked to the 2.0 branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ before_install:
  - export PATH=$HOME/bin:$PATH
  - export PATH=$HOME/.cabal/bin:$PATH
  - export PATH=$HOME/.local/bin:$PATH
- - export PATH=/opt/cabal/1.24/bin:$PATH
+ - export PATH=/opt/cabal/2.0/bin:$PATH
  - export PATH=/opt/happy/1.19.5/bin:$PATH
  - export PATH=/opt/alex/3.1.7/bin:$PATH
  - ./travis-install.sh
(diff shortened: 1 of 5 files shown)


23Skidoo:
Seems to be failing [because of](https://travis-ci.org/haskell/cabal/jobs/259545699#L2915):

```
$ cabal check
These warnings may cause trouble when distributing the package:
* 'ghc-options: -main-is' is not portable.
```

However, [on `master` we get](https://travis-ci.org/haskell/cabal/jobs/259618292#L2073):

```
$ cabal check
No errors or warnings could be found in the package.
cabal check took 0 seconds.
```

idk why there is a difference.

23Skidoo:
I can reproduce this locally, however. My hypothesis is that `-fmonolithic` business in `master` tricks `cabal check` into accepting the `master` version of `cabal-install.cabal` with no warnings.

23Skidoo:
I wonder whether the "`-main-is` is not portable" check still makes sense. It was originally added in 2008 by @dcoutts (commit 39c03860055116afa365cc5e6610f85b4e4116df).

hvr:
> I wonder whether the "-main-is is not portable" check still makes sense. 

I'd argue that `ghc-options` is not portable ;-)

ezyang:
> I can reproduce this locally, however. My hypothesis is that -fmonolithic business in master tricks cabal check into accepting the master version of cabal-install.cabal with no warnings.

Haha. We should file a proper ticket for this.

Issue #4633, opened 2017-07-27
What's the correct way to build Cabal and alex?


snoyberg:
Sorry for the somewhat vague question, but I'm not sure what the expected behavior is. With Cabal-2.0.0.2 and alex-3.2.1, we have this situation:

* Cabal depends on alex as a build tool
* alex has a custom Cabal setup script

Naively, this is a cyclic dependency. I can see a few ways of getting this both to build:

1. Rely on the Cabal that ships with GHC to build alex, and then build Cabal with that alex. From a reproducibility standpoint, that makes me nervous: if I want to pin my Cabal library version, I'll still end up building one package with a different version of the Cabal library.
2. Require that alex be present before building Cabal, which seems suboptimal.
3. Some kind of weird hack with capturing the generate files from alex?

I may simply be missing out on something obvious here.

phadej:
If you build `Cabal-2.0.0.2` with `-f-parsec`, then you shouldn't need `alex`.

Yet, in next major `Cabal` `alex` will be required. I'm not sure, maybe easiest solution would be to preprocess `alex` files: opinions @hvr, @dcoutts?

snoyberg:
Oh, I didn't realize that it was gated behind the `parsec` flag. That's perfect, thank you! Though I _am_ interested in how this will be addressed once parsec is required.

ezyang:
The proposed solution in https://github.com/haskell/cabal/pull/4654 is to commit a copy of the alex-generated files to the repo, allowing us to remove `build-tools: alex` entirely from the Cabla file.

I am a bit concerned about the effects of checking in automatically generated files on the development process; the obvious issues are (1) depending on what version of alex is used, the output could wobble unnecessarily, and (2) it makes merge conflicts a bit more painful.

I was under the impression that Cabal had some affordance for including generated files in sdists, so it wouldn't be necessary to commit them into the repo. See for example https://github.com/haskell/cabal/issues/2940 where @simonmar makes reference to this feature. However, @hvr is not at all sure how this works at all (maybe Cabal will still be willing to use an installed alex even if it's not build-tool'ed).

hvr:
Fwiw, I consider what we do in #4654 a temporary workaround. And yes, cabal may sometimes pick up an Alex on $PATH, but I consider that to be a bug that I intend to fix before 2.2 gets released; the way it works now is not principled.

ezyang:
[...]

PR #4654, opened 2017-08-01
Use parsec, drop parsec flag

phadej:


23Skidoo:
We should be able to also get rid of the `PARSEC=YES` Travis configuration.

phadej:
yeah, I made PR early to see how CIs break. Will try to iron the issues asap.

appveyor.yml
```diff
@@ -18,13 +18,14 @@ install:
 build_script:
   - cd Cabal
   - alex Distribution\Parsec\Lexer.x
-  - ghc --make -threaded -i -i. Setup.hs -Wall -Werror -XRank2Types -XFlexibleContexts
+  - ghc --make -threaded -i -i. Setup.hs -Wall -XRank2Types -XFlexibleContexts
```
23Skidoo: Why do you need to remove `-Werror`?

phadej: it breaks on unused warning in generated `Lexer.hs`

Cabal/Distribution/Parsec/Lexer.x
```diff
@@ -13,6 +13,7 @@
 #ifdef CABAL_PARSEC_DEBUG
 {-# LANGUAGE PatternGuards #-}
 #endif
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
```
23Skidoo: :+1:

Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY : all lexer lib exe
+
+LEXER_HS:=Cabal/Distribution/Parsec/Lexer.hs
+
+all : exe lib
+
+lexer : $(LEXER_HS)
+
+$(LEXER_HS) : boot/Lexer.x
+	alex -o $@ $^
```
hvr: Please use `alex --ghc` to generate the parser the way cabal would.

phadej:
Seems like green build \o/, please review my "let's make travis green hacks", I'll proceed with this on Monday.

Have good weekend everyone!

boot/Lexer.x
```diff
@@ -13,11 +13,24 @@
 #ifdef CABAL_PARSEC_DEBUG
 {-# LANGUAGE PatternGuards #-}
 #endif
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
 module Distribution.Parsec.Lexer
   (ltest, lexToken, Token(..), LToken(..)
   ,bol_section, in_section, in_field_layout, in_field_braces
   ,mkLexState) where
 
+-- [Note: boostrapping parsec parser]
```
ezyang: I'd really love it if there was a THIS FILE IS AUTOMATICALLY GENERATED warning on checked in bootstrapped copy, ideally done with the tooling. Maybe one way to do this that obviously works is, "If this file's name is FOO, then this is an automatically generated copy."

Cabal/Distribution/Parsec/Lexer.hs
```diff
@@ -0,0 +1,504 @@
...
+#ifdef CABAL_PARSEC_DEBUG
+{-# LANGUAGE PatternGuards #-}
+#endif
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+module Distribution.Parsec.Lexer
+  (ltest, lexToken, Token(..), LToken(..)
+  ,bol_section, in_section, in_field_layout, in_field_braces
+  ,mkLexState) where
+
+-- [Note: boostrapping parsec parser]
```
ezyang: I'm a bit concerned about this. See my comment https://github.com/haskell/cabal/issues/4633#issuecomment-322049015

cabal-testsuite/PackageTests/CustomWithoutCabal/cabal.out
```diff
@@ -1,5 +1,4 @@
 # cabal new-build
-Warning: <ROOT>/custom-setup-without-cabal.cabal: This package requires at least Cabal version 99999
```
ezyang: I guess this is being tracked in https://github.com/haskell/cabal/issues/4681 ?

ezyang:
CI changes look fine. I'm mostly concerned about checking in autogenerated source code, but if you want to go and land this immediately we can figure that out later.

cabal-testsuite/PackageTests/CustomWithoutCabal/cabal.out
phadej: yes

phadej:
@ezyang for now it's way simpler to check in `Lexer.hs`, as it would complicate GHC build system too.

boot/Lexer.x
phadej: Won't do. If something, then `alex` should add notes for files generated with it. I'm heavily :-1: of doing something manual / only for Cabal at this point.

phadej:
This is now green.

The only issue is that parsec can't handle few files (unicode-transforms in particular: https://github.com/haskell/cabal/pull/4687), so I  have to work on `.cabal` legacy patch: preprocess some files, so new parser can accept all of current Hackage.

PR #4657, opened 2017-08-02
Use default flags only for new-* path

### LINKED ISSUE #4581: Documentation not generated with config: Documentation: True
For ages I have this in my config:
~~~~
$ fgrep documentation ~/.cabal/config
documentation: True
~~~~
So far this made Cabal generate and install package documentation.
It seems to have gone with Cabal-2.0.
Now, documentation is only generated if I add the documentation flag to every install invocation:
~~~~
$ cabal install --enable-documentation pkg-name
~~~~

alexbiehl:
Fixes #4581. 
With 1e90ae4380cdcef5662d933e3a34dcfef6080ec1 flags from the global
configuration would be ignored. This patch makes sure the defaults are
mixed in only for the new-build code path.

### DIFF
--- a/cabal-install/Distribution/Client/CmdBench.hs
+++ b/cabal-install/Distribution/Client/CmdBench.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ViewPatterns   #-}
 
 -- | cabal-install CLI command: bench
 --
@@ -17,7 +18,8 @@ import Distribution.Client.ProjectOrchestration
 import Distribution.Client.CmdErrorMessages
 
 import Distribution.Client.Setup
-         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags )
+         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags
+         , applyFlagDefaults )
 import qualified Distribution.Client.Setup as Client
 import Distribution.Simple.Setup
          ( HaddockFlags, fromFlagOrDefault )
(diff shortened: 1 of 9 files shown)


23Skidoo:
Ah yeah, that's because command line flags are mappended on top of the ones loaded from the config, so when they're non-mempty from the beginning, they always override whatever was in the config. I think we should just revert 1e90ae4380cdcef5662d933e3a34dcfef6080ec1.

cabal-install/main/Main.hs
```diff
@@ -313,18 +314,27 @@ mainWorker args = topHandler $
...
+      , regularCmd  CmdFreeze.freezeCommand       (defaultsCmdWrapper CmdFreeze.freezeAction)
+      , regularCmd  CmdHaddock.haddockCommand     (defaultsCmdWrapper CmdHaddock.haddockAction)
+      , regularCmd  CmdRun.runCommand             (defaultsCmdWrapper CmdRun.runAction)
+      , regularCmd  CmdTest.testCommand           (defaultsCmdWrapper CmdTest.testAction)
+      , regularCmd  CmdBench.benchCommand         (defaultsCmdWrapper CmdBench.benchAction)
       ]
 
 type Action = GlobalFlags -> IO ()
 
+defaultsCmdWrapper :: ((ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags) -> [String] -> GlobalFlags -> IO ())
```
23Skidoo: I'd be more happy if instead of doing this we applied the defaults directly inside `Cmd*.*Action`. That'd be a tad more verbose, but also more explicit.

So e.g. for `buildAction` you'd have:

```
buildAction cliFlags@(_configFlags, _configExFlags, _installFlags, _haddockFlags) ... =
let (configFlags, configExFlags, installFlags, haddockFlags) = applyFlagDefaults cliFlags
...
```

hvr:
@23Skidoo if we only revert 1e90ae4 then `new-build` regresses again. There was a reason for 1e90ae4.

cabal-install/main/Main.hs
hvr: @23Skidoo we could use ViewPatterns, then it would look like

```hs
buildAction (cliFlags@(configFlags, configExFlags, installFlags, haddockFlags) <- applyFlagDefaults)
```

hvr:
@23Skidoo it's all green! :-)

PR #3913, opened 2016-09-29
Use 'ShortText' as internal representation of `PackageName`

hvr:

PR #4666, opened 2017-08-05
Refactor/cleanup of `String`/`ByteString` usage

hvr:
This refactoring is intended to clean-up various instances where `[Char]` is used to convey
something other than Unicode text. This picks up unfinished yak-shaving that was put aside when working
on #3913 ... ;-)

Note: this PR has been carefully split into self-contained commits to ease reviewing

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -58,6 +58,7 @@ import Distribution.Text
 import Language.Haskell.Extension
 
 import Control.Monad (mapM)
+import qualified Data.ByteString.Lazy as BS
 import Data.List  (group)
 import qualified System.Directory as System
          ( doesFileExist, doesDirectoryExist )
@@ -67,7 +68,6 @@ import qualified Text.PrettyPrint as Disp
 import Text.PrettyPrint ((<+>))
 
 import qualified System.Directory (getDirectoryContents)
-import System.IO (openBinaryFile, IOMode(ReadMode), hGetContents)
 import System.FilePath
          ( (</>), (<.>), takeExtension, takeFileName, splitDirectories
          , splitPath, splitExtension )
@@ -1726,8 +1726,7 @@ checkPackageFiles pkg root = checkPackageContent checkFilesIO pkg
       doesFileExist        = System.doesFileExist                  . relative,
       doesDirectoryExist   = System.doesDirectoryExist             . relative,
       getDirectoryContents = System.Directory.getDirectoryContents . relative,
-      getFileContents      = \f -> openBinaryFile (relative f) ReadMode
-                                   >>= hGetContents
+      getFileContents      = BS.readFile
     }
     relative path = root </> path
 
@@ -1738,7 +1737,7 @@ data CheckPackageContentOps m = CheckPackageContentOps {
     doesFileExist        :: FilePath -> m Bool,
     doesDirectoryExist   :: FilePath -> m Bool,
     getDirectoryContents :: FilePath -> m [FilePath],
-    getFileContents      :: FilePath -> m String
+    getFileContents      :: FilePath -> m BS.ByteString
   }
 
 -- | Sanity check things that requires looking at files in the package.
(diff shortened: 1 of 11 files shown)


Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1777,12 +1776,16 @@ checkCabalFileBOM ops = do
     -- --cabal-file is specified.  So if you can't find the file,
     -- just don't bother with this check.
     Left _       -> return $ Nothing
-    Right pdfile -> (flip check pc . startsWithBOM . fromUTF8)
```
23Skidoo: Can `startsWithBOM` be removed now?

Cabal/Distribution/Utils/Generic.hs
```diff
@@ -246,8 +246,7 @@ startsWithBOM _            = False
 
 -- | Check whether a file has Unicode byte order mark (BOM).
 fileHasBOM :: FilePath -> NoCallStackIO Bool
-fileHasBOM f = fmap (startsWithBOM . fromUTF8)
-             . hGetContents =<< openBinaryFile f ReadMode
+fileHasBOM f = (startsWithBOM . fromUTF8LBS) <$> BS.readFile f
```
23Skidoo: Can't you use `BS.isPrefixOf` here?

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -810,6 +817,63 @@ rawSystemStdout verbosity path args = withFrozenCallStack $ do
...
+data IODataMode = IODataModeText | IODataModeBinary
+
+-- | 'IOData' Wrapper for 'hGetContents'
+--
+-- __Note__: This operation uses lazy I/O. Use 'NFData' to force all
+-- data to be read and consequently the internal file handle to be
+-- closed.
+--
+-- @since 2.2.0
+ioDataHGetContents :: Handle -> IODataMode -> IO IOData
```
23Skidoo: Maybe move `IOData` & friends to a separate module and import that qualified instead of using a prefix? Like the existing `BS.`/`LBS.` precedent.

Cabal/Distribution/Utils/Generic.hs
```diff
@@ -166,72 +164,22 @@ writeFileAtomic targetPath content = do
...
-fromUTF8 (c:cs)
-  | c <= '\x7F' = c : fromUTF8 cs
-  | c <= '\xBF' = replacementChar : fromUTF8 cs
-  | c <= '\xDF' = twoBytes c cs
-  | c <= '\xEF' = moreBytes 3 0x800     cs (ord c .&. 0xF)
-  | c <= '\xF7' = moreBytes 4 0x10000   cs (ord c .&. 0x7)
-  | c <= '\xFB' = moreBytes 5 0x200000  cs (ord c .&. 0x3)
-  | c <= '\xFD' = moreBytes 6 0x4000000 cs (ord c .&. 0x1)
-  | otherwise   = replacementChar : fromUTF8 cs
+fromUTF8 = decodeStringUtf8 . map c2w
```
23Skidoo: Very nice!

Cabal/Distribution/Utils/Generic.hs
hvr: yes, but we can also just remove `fileHasBOM` as nobody uses it anymore iirc

Cabal/Distribution/PackageDescription/Check.hs
hvr: yes

Cabal/Distribution/Simple/Utils.hs
hvr: @23Skidoo what do you suggest? Something like

```hs
module Cabal.Distribution.Simple.IOData
    ( IOData(..)
    , hGetContents
    , hPutContents
    -- ...
    ) where

-- ...
```

?

Cabal/Distribution/Utils/Generic.hs
hvr: fwiw, we could also remove `to/fromUTF8` w/o any deprecation. Nothing in the code-base needs it anymore

Cabal/Distribution/Simple/Utils.hs
23Skidoo: Right, that's what I had in mind.

Cabal/Distribution/Utils/Generic.hs
23Skidoo: +1.

Cabal/Distribution/Utils/Generic.hs
23Skidoo: Yeah, since this will be in Cabal 2.2, setup scripts shouldn't be affected. May be worth adding a note to the changelog about the stuff that is gone.

Cabal/Distribution/Simple/Utils.hs
hvr: I'll go for `C.D.Utils.IOData`, as that's where e.g. `ShortText` lives too

hvr:
@23Skidoo I've implemented your suggestion; can you look over it?

I'll write the changelog as soon as you confirm this is what you intended :-)

23Skidoo:
@hvr Looks good, though you forgot to `git add` the new module.

hvr:
@23Skidoo Doh! ... fixed...

Cabal/Distribution/Utils/Generic.hs
hvr: for changelog see b1bbf2e

PR #4669, opened 2017-08-06
Augment `--allow-{newer,older}` syntax to support wildcards

hvr:
This continues the work started in  a0d8035 (#4575)

### DIFF
--- a/cabal-install/Distribution/Client/Types.hs
+++ b/cabal-install/Distribution/Client/Types.hs
@@ -42,7 +42,7 @@ import Distribution.Types.AnnotatedId
 import Distribution.Types.UnitId
          ( UnitId )
 import Distribution.Types.PackageName
-         ( PackageName )
+         ( PackageName, mkPackageName )
 import Distribution.Types.ComponentName
          ( ComponentName(..) )
 
@@ -56,6 +56,7 @@ import Distribution.Solver.Types.PackageFixedDeps
 import Distribution.Solver.Types.SourcePackage
 import Distribution.Compat.Graph (IsNode(..))
 import qualified Distribution.Compat.ReadP as Parse
+import Distribution.ParseUtils (parseOptCommaList)
 import Distribution.Simple.Utils (ordNub)
 import Distribution.Text (Text(..))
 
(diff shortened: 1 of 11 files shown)


cabal-install/Distribution/Client/Types.hs
```diff
@@ -454,9 +466,35 @@ instance Text RelaxedDep where
...
+instance Text RelaxDeps where
+  disp rd | not (isRelaxDeps rd) = Disp.text "none"
+  disp (RelaxDepsSome pkgs)      = Disp.fsep .
+                                   Disp.punctuate Disp.comma .
+                                   map disp $ pkgs
+  disp RelaxDepsAll              = Disp.text "all"
+
+  parse =           (const mempty        <$> ((Parse.string "none" Parse.+++
+                                               Parse.string "None") <* Parse.eof))
+          Parse.<++ (const RelaxDepsAll  <$> ((Parse.string "all"  Parse.+++
```
23Skidoo: Shouldn't `*` be accepted here as well?

cabal-install/tests/UnitTests/Distribution/Client/ProjectConfig.hs
```diff
@@ -248,6 +249,12 @@ prop_roundtrip_printparse_RelaxedDep rdep =
       [rdep'] -> rdep' == rdep
       _       -> False
 
+prop_roundtrip_printparse_RelaxDeps :: RelaxDeps -> Bool
+prop_roundtrip_printparse_RelaxDeps rdep =
+    case [ x | (x,"") <- Parse.readP_to_S Text.parse (Text.display rdep) ] of
```
23Skidoo: This never generates the `*`-variants, correct? May be worth introducing a helper function that does that.

cabal-install/Distribution/Client/Types.hs
hvr: We can, but it's not essential, as there's two ways to represent `*` currently. In a future refactoring, I do intend to make `--allow-newer` more like `--preferences` and `--constraints`, at which point the outer `RelaxDep` structure will become merely a newtype, which will avoid these corner-cases which make round-tripping tests so weird.

ezyang:
Tests for the new user-visible functionality?!

hvr:
@ezyang indeed, I need to come up with some high-level tests, besides the parser-centric ones.

PR #4674, opened 2017-08-07
Start adding a benchmark that solves for all packages on Hackage.

grayjay:
This benchmark is based off of a script written by @kosmikus.  It adds
functionality for running multiple trials for each package, but it is currently
missing other features.

___________________________________________________________

____________________________________________________________

I wanted to add a benchmark to make it easier to manually test changes to the solver's algorithm.  I'm not sure how it should fit into the repository and CI, though, since it currently isn't portable, and it doesn't have flexible dependency bounds.

This benchmark is meant to be used for finding packages that are affected by a change to the dependency solver when it is likely that the change will have different effects on different packages.  It tries to find packages where the two versions either give different results or have significantly different run times.  It doesn't try to determine whether the overall change is better or worse, though.

This is still a work in progress, but I wanted to ask for feedback before I got too far.

/cc @kosmikus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
    - env: GHCVER=7.10.3 SCRIPT=script USE_GOLD=YES
      os: linux
      sudo: required
-   - env: GHCVER=8.0.2 SCRIPT=script DEPLOY_DOCS=YES USE_GOLD=YES
+   - env: GHCVER=8.0.2 SCRIPT=script DEPLOY_DOCS=YES USE_GOLD=YES TEST_SOLVER_BENCHMARKS=YES
      sudo: required
      os: linux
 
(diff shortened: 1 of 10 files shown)


23Skidoo:
Re: CI, we can either run it as a separate configuration, or if that's not practical, just use Travis to make sure that it at least compiles.

grayjay:
> Re: CI, we can either run solver-benchmarks as a separate configuration, or if that's not practical, just use Travis to make sure that it at least compiles.

Thanks.  I think that the benchmark is too slow to run on Travis, so I just built it and ran the unit tests in the existing GHC 8.0.2 Linux job.  Is that okay?  I initially tried building it in every job, but the OS X builds timed out.  The tests run in the first stage of the build, since they take less than a second.

The rest of the code is ready for review now.

ezyang:
Great, let's ship it!

23Skidoo:
> I just built it and ran the unit tests in the existing GHC 8.0.2 Linux job. Is that okay? I initially tried building it in every job, but the OS X builds timed out.

Yes, I think it's not important to make this test compile on every supported configuration as long as we know that the latest GHC can build it.

grayjay:
Thanks!  I'll merge it now.

grayjay:
It looks like the build failed on master because of a new version of tasty.  I opened an issue: https://github.com/feuerbach/tasty/issues/181

PR #4679, opened 2017-08-09
Turn `allow-{newer,older}` into an accumulating field

hvr:
This fixes the inconvenience occuring in `cabal.project` files of
separate `allow-*` statements overwriting each other. Currently,
without this patch,

    allow-newer: base
    allow-newer: directory

is *NOT* equivalent to

    allow-newer: base, directory

but rather to the last occurence of `allow-newer`, i.e.

    allow-newer: directory

This behaviour has been present since cabal 1.24

----

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
@@ -829,12 +829,12 @@ legacySharedConfigFieldDescrs =
         disp parse
         configPreferences (\v conf -> conf { configPreferences = v })
 
-      , simpleField "allow-older"
+      , monoidField "allow-older"
         (maybe mempty disp) (fmap Just parse)
         (fmap unAllowOlder . configAllowOlder)
         (\v conf -> conf { configAllowOlder = fmap AllowOlder v })
 
-      , simpleField "allow-newer"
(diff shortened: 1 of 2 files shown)


hvr:
@23Skidoo This one would be a candidate for backporting to 2.0

ezyang:
LGTM, but it would be nice if we had a test...

23Skidoo:
Hmm, I thought I've fixed this before, though maybe not for `cabal.project`.

hvr:
@23Skidoo you're probably referring to #3165

23Skidoo:
Yep, just found that PR myself.

cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -1253,6 +1253,13 @@ listFieldWithSep separator name showF readF get' set =
     set' xs b = set (get' b ++ xs) b
     showF'    = separator . map showF
 
+monoidField :: Monoid a => String -> (a -> Doc) -> ReadP a a
```
23Skidoo: Would be nice to have a Haddock comment here.

hvr:
@ezyang jfyi, I have not forgotten about the tests; I plan to provide a couple lateron

PR #4680, opened 2017-08-09
Add missing separator in solver fallback message

tuncer:
Before, it would print a list of packages and append
`Trying configure anyway.`
without any separator between the list and the new sentence.


23Skidoo:
Looks like the test suite needs to be updated to accept the changed output.

tuncer:
Updated test output strings.

tuncer:
The only CI error left is a Travis job timeout.

23Skidoo:
Merged, thanks!

PR #4687, opened 2017-08-14
cabal check flag names

### LINKED ISSUE #4686: Leading dashes in flag names
After a discussion on the Stack issue tracker, it was revealed that the Cabal flag name requirements are rather lax, and currently seem to allow both sequential separators (- and _), as well as leading and trailing separators. For example, this command line works just fine:

    cabal configure -f -totally--positive__

NOTE: this is using:

```
cabal-install version 1.24.0.2
compiled using version 1.24.2.0 of the Cabal library 
```

However, it's unclear how to set this from the command line to either true or false. If I add this line on its own to my cabal file:

    flag -totally--positive__

I see the following in my shell session:

```
$ cabal configure -f -totally--positive__
Resolving dependencies...
Configuring foo-0.1.0.0...
cabal: '--exact-configuration' was given, but the following flags were not
specified: FlagName "-totally--positive__"
$ cabal configure -f +-totally--positive__
Resolving dependencies...
Configuring foo-0.1.0.0...
cabal: '--exact-configuration' was given, but the following flags were not
specified: FlagName "-totally--positive__"
$ cabal configure -f --totally--positive__
Resolving dependencies...
Configuring foo-0.1.0.0...
```

As you can see, ultimately the double-leading-dash was able to disable the flag, but no combination of + and - I tried enabled the flag.

My recommendation: strengthen to parser to behave like the package name parser, disallowing leading, trailing, or sequential separators. Next best thing: remove leading separators.

phadej:
Related to  #4686 

---

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -58,9 +58,11 @@ import Distribution.Text
 import Distribution.Utils.Generic (isAscii)
 import Language.Haskell.Extension
 
+import Control.Applicative (Const (..))
 import Control.Monad (mapM)
 import qualified Data.ByteString.Lazy as BS
 import Data.List  (group)
+import Data.Monoid (Endo (..))
 import qualified System.Directory as System
          ( doesFileExist, doesDirectoryExist )
 import qualified Data.Map as Map
@@ -75,6 +77,7 @@ import System.FilePath
 import System.FilePath.Windows as FilePath.Windows
          ( isValid )
 
+import qualified Data.Set as Set
 
 -- | Results of some kind of failed package check.
 --
@@ -148,6 +151,7 @@ checkPackage gpkg mpkg =
   ++ checkPackageVersions gpkg
   ++ checkDevelopmentOnlyFlags gpkg
   ++ checkFlagNames gpkg
+  ++ checkUnusedFlags gpkg
   where
     pkg = fromMaybe (flattenPackageDescription gpkg) mpkg
 
@@ -1586,7 +1590,7 @@ checkConditionals pkg =
       COr  c1 c2 -> condfv c1 ++ condfv c2
       CAnd c1 c2 -> condfv c1 ++ condfv c2
 
-checkFlagNames ::GenericPackageDescription -> [PackageCheck]
+checkFlagNames :: GenericPackageDescription -> [PackageCheck]
 checkFlagNames gpd
     | null invalidFlagNames = []
     | otherwise             = [ PackageDistInexcusable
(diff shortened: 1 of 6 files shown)


phadej:
Do we have `cabal check` tests in `cabal test-suite`? @ezyang ?

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1584,6 +1586,27 @@ checkConditionals pkg =
...
+    invalidFlagNames =
+        [ fn
+        | flag <- genPackageFlags gpd
+        , let fn = unFlagName (flagName flag)
+        , invalidFlagName fn
+        ]
+    -- starts with dash
+    invalidFlagName ('-':_) = True
+    -- mon ascii letter
+    invalidFlagName cs = any (\c -> not (isAscii c) && isAlpha c) cs
```
hvr: why not warn about *any* non-ascii flags?

```hs
invalidFlagName cs = any (not . isAscii) cs
```

phadej:
Currently

```
[FL973] ~/mess/lead-dash % cat lead-dash.cabal 
name:                lead-dash
version:             0.1.0.0
synopsis:            Lorem
description:         Lorem Ipsum
license:             BSD3
license-file:        LICENSE
author:              Oleg Grenrus
maintainer:          oleg.grenrus@iki.fi
category:            Testing
build-type:          Simple
extra-source-files:  ChangeLog.md
cabal-version:       >=1.10

flag -foobar
  manual: True
  default: False

library
  exposed-modules:     LeadDash
  build-depends:       base >=4.9 && <4.10
  hs-source-dirs:      src
[FL973] ~/mess/lead-dash % cabal check                                                                                                                          
No errors or warnings could be found in the package.
[FL973] ~/mess/lead-dash % /home/ogre/Documents/other-haskell/cabal/dist-newstyle/build/x86_64-linux/ghc-8.0.2/cabal-install-2.1.0.0/build/cabal/cabal check     
The following errors will cause portability problems on other environments:
* Suspicious flag names: -foobar. To avoid ambiguity in command line
interfaces, flag shouldn't start with a dash. Also for better compatibility,
flag names shouldn't contain non-ascii characters.

[FL973] ~/mess/lead-dash % cabal configure                                                                                                                      
Resolving dependencies...
Configuring lead-dash-0.1.0.0...
[FL973] ~/mess/lead-dash % /home/ogre/Documents/other-haskell/cabal/dist-newstyle/build/x86_64-linux/ghc-8.0.2/cabal-install-2.1.0.0/build/cabal/cabal configure
Resolving dependencies...
Invalid flag assignment: --foobar
  error, called at ./Distribution/ReadE.hs:46:24 in Cabal-2.1.0.0-inplace:Distribution.ReadE
```

I.e. `-foobar` is still allowed in ReadP parser, but you cannot really use it.

ATM it's simpler to disallow leading dash, as you cannot use it with current `--constraint` syntax, alternatively one could make `--constraint +-flag` work. I'm not sure it's worth the trouble.

Let's see what Travis says.

23Skidoo:
> ATM it's simpler to disallow leading dash

+1 on that.

Cabal/doc/developing-packages.rst
```diff
@@ -2162,7 +2162,8 @@ Configuration Flags
 
    Flag section declares a flag which can be used in `conditional blocks`_.
 
-A flag section may contain the following fields:
+Flag names are case-insensitive and must match ``[a-z0-9_][a-z0-9_-]*``
```
hvr: That limitation only applies to Hackage as a point of distribution.

Here's a perfectly valid `.cabal` file including unicode package names and flag names:

```
name:                無
version:             0
synopsis:            The canonical non-package
build-type:          Simple
cabal-version:       >=1.10

source-repository head
  Type:     git
  Location: https://github.com/hvr/-.git

flag 無
  description: 無

library

  if !flag(無)
    buildable:False
```

phadej: I'll change it to `[[:alnum:]_][[:alnum:]_-]*` and add a note about Hackage format.

PR #4689, opened 2017-08-14
Mark FreeBSD as supporting RPaths

arrowd:
FreeBSD has this patch in its GHC port for 2 years now.


23Skidoo:
> FreeBSD has this patch in its GHC port for 2 years now.

Shouldn't this check be dependent on GHC version then?

arrowd:
> Shouldn't this check be dependent on GHC version then?

Don't see how is so. Isn't it about an OS, not compiler?

23Skidoo:
You say that some versions of GHC on FreeBSD have this patch and some don't. I think it makes sense to check **both** the OS and the GHC version then, since our policy is to support all GHC versions released in last 5 years.

arrowd:
Well, the fact that we didn't have this patch for some time doesn't mean it didn't work for us. We might have not yet a chance to stumble upon it.

Another example of  such problem is https://phabricator.haskell.org/rGHC0a891c8c448a0e70abb075702857d743674ad1ad We've been using GHC for a long time until we stumbled upon linking linker scripts issue, and now fixed it.

If you insist, I may try to reach original author of this patch and ask him for a full story, but I think it is a really overkill for such simple one-liner.

23Skidoo:
> Well, the fact that we didn't have this patch for some time doesn't mean it didn't work for us.

I suppose there is a reason that this check currently returns `False`. Ideally we'd use CI to test this patch against each supported GHC version instead of guessing, but unfortunately Travis doesn't support FreeBSD.

arrowd:
Judging from the comment above that case expression, these checks return `False` for all OSes where GHC haven't been extensively checked. And since Cabal team doesn't run FreeBSD on Travis, the burden of testing lies on us. We've been running this patch for 2 years, so I claim that this change is tested.

23Skidoo:
So which versions of GHC have you tested this patch with?

arrowd:
According to SVN logs, 7.10.2 and 8.0.2.

23Skidoo:
Can you please make this check return `True` on FreeBSD only when the GHC version is `>= 7.10.2`?

23Skidoo:
Sorry, I didn't intend to close this, this is some GitHub glitch.

23Skidoo:
Cont. in #4695.

PR #4695, opened 2017-08-15
Mark FreeBSD as supporting RPaths

23Skidoo:
This is an updated version of #4689.

/cc @arrowd


arrowd:
Thanks for doing this!

PR #4698, opened 2017-08-15
Alex latin1

phadej:


Cabal/tests/ParserTests/regressions/nothing-unicode.golden
```diff
@@ -0,0 +1,24 @@
+name: 無
+version: 0
+cabal-version: >=1.10
+build-type: Simple
+license: UnspecifiedLicense
+synopsis: The canonical non-package 無
+x-無: 無
```
phadej: ping @hvr 

It seems that ReadP parser accepts this. I'm 99% sure that all field-names on Hackage are ASCII  though. Should I add a `cabal chech`. Also what would be proper regexp for fieldnames. Currently it's *almost everything*:

```
$namecore        = [$printable] # [$space \: \" \{ \} $paren $symbol']
```

as field names are taken as is, but they can be validated further in the parser stage.

23Skidoo: Surely we don't need to support Unicode field names?

phadej: custom fields can be, but I added a check that we don't want them on  Hackage.

The story is very similar as with flag: French or German shop might have some `x-Straße` there. Let them have it, if they want.

23Skidoo: I doubt that anyone uses this feature, but I can see no harm in supporting it.

PR #4701, opened 2017-08-16
Introduce Distribution.Compat.Lens

phadej:
This patch introduces `Distribution.Compat.Lens`, a small portion of `lens`. The need is two fold:
- to provide lenses to users of `Cabal`
- to use it internally in places where lenses shine (subjectively, see `Check.hs`)

The design is simple:
- `Distribution.Compat.Lens` have some definitions from `lens`.
- type modules have sibling `.Lens` module which exports the type and lenses, with the same name as fields. Users can either

```haskell
import Distribution.Types.PackageDescription
import qualified Distribution.Types.PackageDescription.Lens as L
```

or

```haskell
import Distribution.Types.PackageDescription.Lens
```

Latter variant is not yet viable, as not everything is lensified. But we can make progress there quite easily.

To better compare lens and "old-school" Haskell approaches see e.g. (I realized later, these are essentially same checks):
- lens `checkUnusedFlags` https://github.com/phadej/cabal/blob/2dbd57a4aba07828c9520c7692c63138cfcda658/Cabal/Distribution/PackageDescription/Check.hs#L1620-L1642
- vanilla `checkConditionals` https://github.com/phadej/cabal/blob/2dbd57a4aba07828c9520c7692c63138cfcda658/Cabal/Distribution/PackageDescription/Check.hs#L1561-L1597

--- 

This commit includes some lens defintions to show up, how the Cabal-lens
framework could look like.

As an example checks which used ad-hoc lenses use newly introduced
lens-framework.

### DIFF
--- a/Cabal/Distribution/Compat/Lens.hs
+++ b/Cabal/Distribution/Compat/Lens.hs
@@ -12,12 +12,20 @@ module Distribution.Compat.Lens (
     -- ** rank-1 types
     Getting,
     ASetter,
-    -- * Folds
+    -- * Getter
+    view,
+    -- * Setter
+    set,
+    over,
+    -- * Fold
     toDListOf,
     toListOf,
     toSetOf,
     -- * Common lenses
-    _2,
+    _1, _2,
+    -- * Operators
+    (&),
+    (.~), (%~),
     -- * Cabal developer info
     -- $development
     ) where
@@ -45,7 +53,24 @@ type Getting r s a = (a -> Const r a) -> s -> Const r s
 type ASetter s t a b = (a -> Identity b) -> s -> Identity t
 
 -------------------------------------------------------------------------------
--- Folds
+-- Getter
+-------------------------------------------------------------------------------
+
+view :: s -> Getting a s a -> a
+view s l = getConst (l Const s)
+
+-------------------------------------------------------------------------------
+-- Setter
+-------------------------------------------------------------------------------
+
+set :: ASetter s t a  b -> b -> s -> t
+set l x = over l (const x)
+
+over :: ASetter s t a b -> (a -> b) -> s -> t
+over l f s = runIdentity (l (\x -> Identity (f x)) s)
+
+-------------------------------------------------------------------------------
+-- Fold
 -------------------------------------------------------------------------------
 
 toDListOf :: Getting (DList.DList a) s a -> s -> DList.DList a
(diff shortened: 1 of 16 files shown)


23Skidoo:
I haven't reviewed the code in detail, but in general I'm +1 on this idea.

Cabal/Distribution/Compat/Lens.hs
```diff
@@ -0,0 +1,140 @@
...
+toListOf l = DList.runDList . toDListOf l
+
+toSetOf  :: Getting (Set.Set a) s a -> s -> Set.Set a
+toSetOf l s = getConst (l (\x -> Const (Set.singleton x)) s)
+
+-------------------------------------------------------------------------------
+-- Common
+-------------------------------------------------------------------------------
+
+_2 ::  Functor f => (a -> f b) -> (c, a) -> f (c, b)
```
hvr: You can't say `_2` w/o saying `_1`, IMO ;-)

Cabal/Distribution/Compat/Lens.hs
```diff
@@ -0,0 +1,140 @@
...
+
+-- $development
+--
+-- We cannot depend on @template-haskell@, because Cabal is a boot library.
+-- This fact makes defining optics a manual task. Here is a small recipe to
+-- make the process less tedious.
+--
+-- First start a repl with proper-lens dependency
+--
+-- > cabal new-repl Cabal:lib:Cabal ???
```
hvr: Yes, the documented but not yet implemented `extra-packages:` is supposed to provide just that; but until it is implemented, a cabal flag

```
if flag(deploy-lens-missiles)
   build-depends: lens >= ... && < ...
```

is a reasonable temporary workaround.

Cabal/Distribution/Compat/Lens.hs
phadej: This is btw a good question. Should we introduce `FieldN` classes for these, like `lens` and `microlens` do? (I have `_4` used somewhere already)

phadej:
@23Skidoo seems I cannot restart appveyor builds, can you help?

23Skidoo:
Restarted the build. IIRC @hvr said that he could work around that by logging in to AppVeyor with my user name somehow.

hvr:
Yeah, for some reason I get offered a dropdown allowing me impersonate @23Skidoo or just be myself...

23Skidoo:
It's just AppVeyor being weird and having no good equivalent for Github organizations. I think that if @phadej was added to the cabal team on GitHub he also would be able to log in as me.

23Skidoo:
AppVeyor failure seems legit, BTW.

phadej:
and now it isn't legit:
```
 ghc (exited 1) - ghc not installed. An error occurred during installation:
 The operation has timed out.
```

23Skidoo:
Restarted.

PR #4710, opened 2017-08-18
Fix minor regression in flag assignment marshalling to custom-setup

hvr:
This was introduced by f8dc46a0fe68e35ec0397da92d7e9d4f168aa126 which
resulted in passing `--flags=+foo` to older custom setup scripts which
wouldn't recognize the `+` prefix.

This was reported by @cocreature

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -674,7 +674,7 @@ configureOptions showOrParseArgs =
          configConfigurationsFlags (\v flags -> flags { configConfigurationsFlags = v })
          (reqArg "FLAGS"
               (readP_to_E (\err -> "Invalid flag assignment: " ++ err) parseFlagAssignment)
-              (map showFlagValue))
+              (map showFlagValue'))
 
       ,option "" ["extra-include-dirs"]
          "A list of directories to search for header files"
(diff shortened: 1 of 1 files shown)


hvr:
@ezyang btw, do we have tests that test the communication protocol of `cabal-install` w/ older `lib:Cabal`?

ezyang:
Not comprehensive. The best tests we have are when we run a test that passes through to the packaged Cabal with GHC, but there aren't many of those.

hvr:
@23Skidoo since f8dc46a was backported to 2.0.1, I think we need to backport this one too.

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -778,6 +778,13 @@ configureOptions showOrParseArgs =
       reqArgFlag title _sf _lf d
         (fmap fromPathTemplate . get) (set . fmap toPathTemplate)
 
+    -- We can't use 'showFlagValue' because legacy custom-setups don't
+    -- support the '+' prefix in --flags; so we omit the (redundant) + prefix;
+    -- NB: we assume that we never have to set/enable '-'-prefixed flags here.
```
23Skidoo: > NB: we assume that we never have to set/enable '-'-prefixed flags here.

Those are illegal nowadays, right?

PR #4714, opened 2017-08-20
Demote 'scope' field version check to a warning

DanielG:
In order to migrate from pre `scope` Setup.hs hacks to 2.0 we would like
to specify both `scope` and `x-scope` (as just `scope` can't be accessed
from Setup.hs AFAIK) and handle Cabal<2.0 with custom logic. However
this check makes this undistributable as hackage rejects such packages.


23Skidoo:
Merged, thanks!

PR #4724, opened 2017-08-28
Fix verbosity flags getting removed before being passed to `printPlan`

hvr:
This fixes a case where the log-output "In order, the following would be
installed: ..." would fail to honor verbosity flags.

### DIFF
--- a/cabal-install/Distribution/Client/Install.hs
+++ b/cabal-install/Distribution/Client/Install.hs
@@ -1148,7 +1148,7 @@ performInstallations verbosity
         -- If the user has specified --remote-build-reporting=detailed or
         -- --build-log, use more verbose logging.
         loggingVerbosity :: Verbosity
-        loggingVerbosity | overrideVerbosity = max Verbosity.verbose verbosity
+        loggingVerbosity | overrideVerbosity = modifyVerbosity (max verbose) verbosity
                          | otherwise         = verbosity
 
         useDefaultTemplate :: Bool
--- a/cabal-install/Distribution/Client/Install.hs
+++ b/cabal-install/Distribution/Client/Install.hs
@@ -167,7 +167,7 @@ import Distribution.System
 import Distribution.Text
          ( display )
 import Distribution.Verbosity as Verbosity
-         ( Verbosity, normal, verbose )
+         ( Verbosity, modifyVerbosity, normal, verbose )
 import Distribution.Simple.BuildPaths ( exeExtension )
 
 --TODO:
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Install.hs
```diff
@@ -597,6 +597,13 @@ checkPrintPlan verbosity installed installPlan sourcePkgDb
     dryRun            = fromFlag (installDryRun            installFlags)
     overrideReinstall = fromFlag (installOverrideReinstall installFlags)
 
+    -- Escalate verbosity until level has reached at least 'verbose'
+    -- NB: this makes sure we don't lose any verbosity flags
+    atLeastVerbose :: Verbosity -> Verbosity
```
23Skidoo: I'm confused. Why doesn't ``verbosity `max` verbose`` work?

hvr: because neither

```hs
isVerboseXXX verbosity == isVerboseXXX (verbosity `max` verbose)
```
nor

```hs
isVerboseXXX verbosity == isVerboseXXX (verbose `max` verbosity)
```

(where XXX stands for one of the verbosity flags) holds for all possible values of `verbosity`.

23Skidoo: Ah right, forgot that `Verbosity` has additional state now.

23Skidoo:
BTW, this is not the only place we're using `max` and `Verbosity` together:

https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/ProjectConfig.hs#L329
https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Install.hs#L1150

There are also a couple of places we're using `min`, but they look harmless.

hvr:
@23Skidoo fyi, I just realised that I need a new primitive in `.Verbosity` (because `moreVerbosity` treats `Silent` as a fixpoint... :slightly_frowning_face: ), something like

```hs
modVerbosity :: (Verbosity -> Verbosity) -> Verbosity -> Verbosity
```

and you'd use it like

```hs
   let adaptedVerbosity
        | containsReinstalls && not overrideReinstall = modVerbosity (max verbose) verbosity
        | otherwise                                   = verbosity
```

are you ok w/ that?

23Skidoo:
Yep, I actually considered suggesting adding a helper function for this to `D.Verbosity`.

hvr:
@23Skidoo I've updated the PR

23Skidoo:
Looks good, feel free to merge once it's green!

PR #4730, opened 2017-08-31
Add normaliseVersionRange

phadej:
foldVersionRange' recognises (> x || == v)  and (== x || > x) as
(>= v), and similarly for <=. Therefore `simpleParsec . display` isn't
structurally `id`, but `normalizeVersionRange`.

### DIFF
--- a/Cabal/Distribution/Version.hs
+++ b/Cabal/Distribution/Version.hs
@@ -47,6 +47,7 @@ module Distribution.Version (
   simplifyVersionRange,
   foldVersionRange,
   foldVersionRange',
+  normaliseVersionRange,
   hasUpperBound,
   hasLowerBound,
 
@@ -560,6 +561,24 @@ foldVersionRange' anyv this later earlier orLater orEarlier
     fold (IntersectVersionRanges v1 v2) = intersect (fold v1) (fold v2)
     fold (VersionRangeParens v)         = parens (fold v)
 
+-- | Normalise 'VersionRange'.
+--
+-- @foldVersionRange' anyVersion thisVersion ...@
+--
+-- Note: strips parens constructor.
+normaliseVersionRange :: VersionRange -> VersionRange
+normaliseVersionRange = foldVersionRange'
+    anyVersion
+    thisVersion
+    laterVersion
+    earlierVersion
+    orLaterVersion
+    orEarlierVersion
+    (\v _ -> withinVersion v)
+    (\v _ -> majorBoundVersion v)
+    unionVersionRanges
+    intersectVersionRanges
+    id
 
 -- | Does this version fall within the given range?
 --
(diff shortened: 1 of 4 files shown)


phadej:
Is there a reason not to normalize `VersionRange` on-construction, thus making `foldVersionRange` and `foldVersionRange'` the same.

phadej:
@hvr,  @dcoutts, @ezyang 

I'd actually want to escalate this PR (later) and do following:
- Make `VersionRange` to represent syntax faithfully, i.e. add constructors for `>=` and `<=`.
- Use `VersionIntervals` where we actually compute with ranges (i.e. in solver)

Cabal/Distribution/Version.hs
```diff
@@ -387,15 +388,19 @@ orEarlierVersion v = UnionVersionRanges (ThisVersion v) (EarlierVersion v)
...
 
 -- | The version range @vr1 && vr2@
 --
 -- >   withinRange v' (intersectVersionRanges vr1 vr2)
 -- > = withinRange v' vr1 && withinRange v' vr2
 --
 intersectVersionRanges :: VersionRange -> VersionRange -> VersionRange
-intersectVersionRanges = IntersectVersionRanges
+intersectVersionRanges (IntersectVersionRanges a b) c =
+    intersectVersionRanges a (intersectVersionRanges b c)
```
23Skidoo: Can the result of the second call also be a `IntersectVersionRanges`?

Cabal/Distribution/Version.hs
```diff
@@ -560,6 +565,24 @@ foldVersionRange' anyv this later earlier orLater orEarlier
     fold (IntersectVersionRanges v1 v2) = intersect (fold v1) (fold v2)
     fold (VersionRangeParens v)         = parens (fold v)
 
+-- | Normalize 'VersionRange'.
```
23Skidoo: The explanation here is not very good.

Cabal/tests/UnitTests/Distribution/Version.hs
```diff
@@ -744,6 +782,26 @@ prop_parse_disp1 vr =
       IntersectVersionRanges (stripParens v1) (stripParens v2)
     stripParens v = v
 
+-- | Reassociate union and intersect
```
23Skidoo: Can you expand this comment a bit?

23Skidoo:
> Is there a reason not to normalize `VersionRange` on-construction (e.g. `intersect anyVersion = id` etc.),

I think it's supposed to faithfully represent whatever was entered by the user, and clients are supposed to transform it to `VersionIntervals` for extracting non-trivial info out of it. So both of your intended improvements sound good to me.

Cabal/Distribution/Version.hs
phadej: I removed this change

phadej:
@23Skidoo after tweaking a bit, I got needed roundtrip properties by changing the `disp`.

PR #4732, opened 2017-08-31
Versioned licenses: specifying only-this-version or any-later-version

bucklereed:
These forms will be accepted without complaint when the `cabal-version`
field is less than 2.1, but are mandatory if it's above 2.1. As a
knock-on of that, version numbers for versioned licenses are mandatory
above 2.1 as well.

This is necessary because just saying 'GPL' is hopelessly
underspecified: there is an enormous difference between v2 and v3, and
significant practical differences between v2-only and v2+ (e.g., the
latter can be upgraded to v3 and then can incorporate Apache 2.0
licensed code, but the former cannot). Refusing the temptation to
guess is the only sane option here, and, hence, ambiguous forms must
be rejected.

Old Cabals will complain about the new forms, but not to the point of
rejecting them. New Cabals will happily accept both new and old forms
when `cabal-version` is low, and the strict behaviour is only
triggered by a sufficiently high `cabal-version`. This means that
there will be no surprises for existing projects until they opt in to
having a newer `cabal-version`.

> Please include the following checklist in your PR:

> * [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/#conventions).
> * [x] Any changes that could be relevant to users have been recorded in the changelog.
> * [x] The documentation has been updated, if necessary.

> Please also shortly describe how you tested your change. Bonus points for added tests!

Seven new tests (which are blowing up the diffstat, because that's seven whole copies of the GPLv2...).

### DIFF
--- a/Cabal/Distribution/License.hs
+++ b/Cabal/Distribution/License.hs
@@ -44,7 +44,10 @@
 
 module Distribution.License (
     License(..),
-    knownLicenses,
+    specifiedLicenses,
+    underSpecifiedLicenses,
+    LicenseBound(..),
+    LicenseBoundedVersion(..),
   ) where
 
 import Prelude ()
@@ -55,6 +58,20 @@ import Distribution.Text
 import qualified Distribution.Compat.ReadP as Parse
 import qualified Text.PrettyPrint as Disp
 
+-- | Is the code released under a specific version of the license
+-- only, or can a later version of the license be chosen by the
+-- receiver?
+data LicenseBound
+  = LicenseBoundExactOnly | LicenseBoundOrLater
+  deriving (Generic, Read, Show, Eq, Typeable, Data)
+
+instance Binary LicenseBound
+
+data LicenseBoundedVersion = LicenseBoundedVersion Version (Maybe LicenseBound)
+  deriving (Generic, Read, Show, Eq, Typeable, Data)
+
+instance Binary LicenseBoundedVersion
+
 -- | Indicates the license under which a package's source code is released.
 -- Versions of the licenses not listed here will be rejected by Hackage and
 -- cause @cabal check@ to issue a warning.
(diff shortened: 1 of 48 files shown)


phadej:
The issue is valid, but I'm :-1: on the solution, and rather implement https://github.com/haskell/cabal/issues/2547 then inventing own syntax.

There's http://hackage.haskell.org/package/spdx which we can vendor in.

bucklereed:
My objection to SPDX syntax (and the + convention generally) is that it privileges the exact-match semantics (by being shorter and looking like the 'default'), which is not the recommended way to apply the GPL family of licenses, and which *does* lead to pain down the road if it's thoughtlessly applied.

bucklereed:
I'm not quite sure what's going on with the AppVeyor build. It looks like there are missing reference files, but the tests pass on my machine and `git status` is clean. I generated the reference `setup.out` files with `cabal-tests --accept`; did I need to do something else?

phadej:
... yet, spdx solves other problems like dual licensing  (both conjuctive and disjuctive). Inventing own syntax  for a thing which already has one is :(

BardurArantsson:
I'm with @phadej on this one, I think (FWIW).

@bucklereed What are the chances of "upstreaming" the necessary 'bits' to have spdx fully support the currently unsupported use cases? Is it even possible?

23Skidoo:
I like the idea of using an already-existing standard for this, but the question is who will implement it. #2547 has been open for a while now, and nothing has happened.

phadej:
I can implement spdx stuff as soon as I'm done refactoring parser. AFAIK no green light were given for implementing spdx

peti:
For what it's worth, I too believe that SPDX is the way to go forward. It's a well established standard that's been widely adopted elsewhere and I'd very much like to have its expressive power available in Cabal files. The current solution is very limited and doesn't do much more than "give a hint", which pretty worthless for everyone who actually cares about the legal implications of using certain code.

23Skidoo:
OK, the consensus seems to be that SPDX is the way forward.

@bucklereed
Thanks for the work you've done on this PR and for reviving the discussion about #2547. Would you be interested to work on #2547 together with @phadej?

23Skidoo:
I'm leaving this open for now in case the SPDX PR never materialises and to serve as a reminder.

23Skidoo:
There has been major progress on SPDX front (see #4913), and it looks like @phadej will finish it in time for the 2.2 release, so I'm closing this PR.

PR #4736, opened 2017-09-01
Add make doctest [ci skip]

phadej:
This is poor man doctest.

ping @hvr @23Skidoo 

Prerequisites:
- `doctest` executable is in `$PATH`, compiled with GHC used to
  new-build the Cabal
- `.ghc.environment` file is generated

As GHC picks up dependencies from `.ghc.environment` file, a simple
`doctest MODULES` works.

As you see, I had a bug in doctests. For now, I don't add this to CI,
as it's enough to run doctests occasionally (there are only 4 examples
atm).

### DIFF
--- a/Cabal/Distribution/Compat/Newtype.hs
+++ b/Cabal/Distribution/Compat/Newtype.hs
@@ -55,7 +55,7 @@ ala pa hof = alaf pa hof id
 -- |
 --
 -- >>> alaf Sum foldMap length ["cabal", "install"]
--- 11
+-- 12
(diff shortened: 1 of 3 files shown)


23Skidoo:
Looks OK, though it would be nice to enable it on Travis eventually if it's not too slow.

phadej:
It takes 45 seconds on my machine. 

Yet, eventually, we'd have `new-doctest` so one don't need to rely on hack, than we can add it to CI.

Issue #2832, opened 2015-09-22
Extend `.cabal` format with "common" definition stanzas


hvr:
_( this is a project task description for the upcoming https://github.com/haskell/cabal/wiki/Hackathon2015 )_

One complaint about the `.cabal` format is that there's avoidable duplication when multiple stanzas requiring to repeat some shared properties are involved (most notably `build-depends`)

Here's an example how common stanzas could be used:

```
name: thispkg
version: 1.2.3
cabal-version: >=2.0

-- ...

flag small_base
  description: Choose the new smaller, split-up base package.

-- unnamed global common stanza included implicitly in all library/executable/benchmark/test stanzas
common
  build-depends: base-orphans < 1

-- named common stanza which needs to be explicitly included
common cdef1
  hs-source-dirs: src

  build-depends: text
  if flag(small_base)
    build-depends: base >=2.1 && <5, array, containers, directory
  else
    build-depends: base >= 1.0

library
  include cdef1
  build-depends: parsec
  other-modules:

test-suite tests
  type: exitcode-stdio-1.0
  main-is: test.hs
  include cdef1
  build-depends: thispkg, tasty

benchmark bench1 
  type: exitcode-stdio-1.0
  main-is: bench1.hs
  include cdef1
  build-depends: thispkg, criterion
```

In the example above a single common section named `cdef1` is defined, which is included by all 3 real stanzas. `include cdef` acts as if the respective common stanza's body was inserted in place of `include cdef`.

---

Possible extensions
- reserve a few special common stanza ids `all`, `library`, `test-suites`, and `benchmark` to declare common stanzas that are implicitly `include`d by the respective named real stanza. In the example above we could have used `common all` instead of `common cdef1` and all `include cdef1` simply removed.
- allow recursive (but acyclic) includes. I.e. allow common stanzas to `include` other common stanzas.

/cc @dcoutts

---

@dcoutts suggests to support both,
1. named sections and explicit include at uses sites
2. unnamed global (implicit include into start of all sections) (would subsume #3404)

Moreover, we'd want this mechanism not only for `.cabal` files but also for cabal's config files. For non-`.cabal` files we want to reuse the infrastructure for file-backed include files (`include-file ./some-ast-frament.inc`).

----

refinement: common stanzas ought to  be typed so we can catch mistakes early on during parsing already, see https://github.com/haskell/cabal/issues/2832#issuecomment-236696081

phadej:
[...]

PR #4751, opened 2017-09-08
RFC: Common stanzas

phadej:
See the last commit

---

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -32,15 +32,33 @@ extra-source-files:
   -- Generated with 'misc/gen-extra-source-files.sh'
   -- Do NOT edit this section manually; instead, run the script.
   -- BEGIN gen-extra-source-files
+  tests/ParserTests/errors/common1.cabal
+  tests/ParserTests/errors/common1.errors
+  tests/ParserTests/errors/common2.cabal
+  tests/ParserTests/errors/common2.errors
+  tests/ParserTests/errors/common3.cabal
+  tests/ParserTests/errors/common3.errors
   tests/ParserTests/regressions/Octree-0.5.cabal
+  tests/ParserTests/regressions/Octree-0.5.format
+  tests/ParserTests/regressions/common.cabal
+  tests/ParserTests/regressions/common.format
+  tests/ParserTests/regressions/common2.cabal
+  tests/ParserTests/regressions/common2.format
   tests/ParserTests/regressions/elif.cabal
+  tests/ParserTests/regressions/elif.format
   tests/ParserTests/regressions/elif2.cabal
+  tests/ParserTests/regressions/elif2.format
   tests/ParserTests/regressions/encoding-0.8.cabal
+  tests/ParserTests/regressions/encoding-0.8.format
   tests/ParserTests/regressions/generics-sop.cabal
+  tests/ParserTests/regressions/generics-sop.format
   tests/ParserTests/regressions/haddock-api-2.18.1-check.cabal
   tests/ParserTests/regressions/issue-774.cabal
+  tests/ParserTests/regressions/issue-774.format
   tests/ParserTests/regressions/nothing-unicode.cabal
+  tests/ParserTests/regressions/nothing-unicode.format
   tests/ParserTests/regressions/shake.cabal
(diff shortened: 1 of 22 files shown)


phadej:
Attempt on https://github.com/haskell/cabal/issues/2832

phadej:
I have redone the PR. Ping @23Skidoo @hvr.

There's internal note, I'll write a user guide section about common stanzas tomorrow if that behavior is ok.

Difference to previous is that you **cannot** say

```
library:
  build-depends: foo
  include common-stanza
```

includes should be first in stanza:
```
library:
  include common-stanza
  build-depends: foo
```

23Skidoo:
@phadej That restriction seems reasonable to me, in fact I think that it improves readability.

phadej:
Two questions:

- Should common stanzas be defined before use (currently: no. Done in separate pass)
- Should common stanzas be able to reference other common stanzas? (currently: no)

Easy to implement are:
- No/No: current
- Yes/Yes: we can parse common stanzas as we go, in single pass. Parsing common stanza is not much different than parsing other stanzas
- Yes/No: even simpler than above

Fourth variant is difficult, because with "Define everywhere + reference others (No/Yes):" we can write cycles:
```
common foo
  include bar

common bar
  include foo
```

23Skidoo:
I'd say that requiring common stanzas to be defined before use is fine (provided that the error message is helpful). Regarding inheritance -- do you want to allow only

```
common common-stanza
    include another-common-stanza
    ...
```

or also

```
common common-stanza
    include common-stanza-a
    include common-stanza-b
    ...
```

?

phadej:
You can `include` as many common stanzas as you like. There is no technical reason to limit that.

phadej:
To clarify the previous question, it can be posed also as *Do we want to allow including stanzas from other stanzas?*. If **yes**, than we have to require **define-before-use**.

23Skidoo:
> Do we want to allow including stanzas from other common stanzas?

IMO it can be useful, so I'm +1.

One question: since IIRC list fields have monoid semantics, should there be a way to explicitly set a list field to `[]`? For example, IIUC,

```
common foo
    cpp-options: -foo

common bar
    cpp-options: -bar

common foobar
    include foo
    include bar

common barfoo
    include bar
    include foo
```

will result in `foobar` having `cpp-options: -foo -bar`, and `barfoo` having `cpp-options: -bar -foo`.

Should it be possible to do something like

```
common foo
    cpp-options: -foo

common bar
    include foo
    cpp-options := -bar

common baz
    include bar
```

and have `baz`'s `cpp-options` be set to `-bar` only?

phadej:
@23Skidoo 

> Should it be possible to do something like
```
common bar
    include foo
    cpp-options := -bar
```

That's technically challenging because that alters Cabal's lexical syntax, which is very major change.
For that example one could factor *common* (pun-intended) parts of `foo` into `quux` and have

```
common quux:
    ...

common foo
    include quux
    cpp-options: -foo

common bar
    include quux
    cpp-options := -bar

common baz
    include bar
```

hvr:
> One question: since IIRC list fields have monoid semantics, should there be a way to explicitly set a list field to []?

What about e.g.

```
common bar
    include foo
    !cpp-options: -bar
```

?

However, I feel like we can postpone this, and it at some later spec-version incrementally, when have gained more experienced/feedback with the basic support. (But we can surely spend some thought on how we could do it at some point in the future of course)

23Skidoo:
> However, I feel like we can postpone this, and add it at some later spec-version incrementally, when have gained more experienced/feedback with the basic support.

Yeah, it's totally fine to postpone this. I like the !cpp-options proposal.

phadej:
Last minute changes, but with docs:

screenshot from 2017-12-12 18-09-28

gbaz:
do we need a docs tweak? from reddit: "Unless I'm mistaken, it appears that import has to be the first thing in a stanza, but this isn't mentioned anywhere."

(https://www.reddit.com/r/haskell/comments/85tuly/latest_hackage_deployment/dw4mf1j/)

ekmett:
I was just bitten by precisely the same thing @gbaz mentioned here.

domenkozar:
Doc change: https://github.com/haskell/cabal/pull/5610

PR #4759, opened 2017-09-12
Fix new-build component name docs

W4RH4WK:


23Skidoo:
Merged, thanks! Next time please use `[ci skip]` anywhere in the commit message body to avoid triggering build bots for docs-only changes.

PR #4599, opened 2017-07-10
Fix bug in conflict counting.

grayjay:
Two main commits:

####    Fix bug in conflict counting.
    
In Explore.hs, the solver calculates a conflict set corresponding to the option
to not choose a value for each goal, called `initial`.  That conflict set should
be added to the ConflictMap when the current goal causes a conflict.  However,
there was a bug, and the solver added `initial` to the ConflictMap at every
level.
    
The bug meant that the solver preferred all goals that it had already chosen.
If a new goal started to cause conflicts, it would take a while for the solver
to start preferring the new goal over the ones that it had previously seen.
See #4595 for an example where this problem caused longer backjumps and slowed
down the solver.

####    Refactor handling of conflict set representing option to not choose a goal.
    
The conflict set that represents the option to not choose a value for a goal,
`initial`, has two purposes:
    
1. `initial` contributes to a node's conflict set, because it is unioned with
   the other conflict sets under the same choice node, if the current variable
   is contained in all of them.
2. `initial` contributes to the conflict count.  When it is unioned with the
   other conflict sets, it is also added to the ConflictMap.
    
Previously, `initial` was treated as the first of a node's children for (1) and
the last of the children for (2).  This refactoring treats `initial` as the last
child in both cases, to make the code clearer, and renames it to `lastCS`.
There should be no change in behavior.

____________________________________________________________

/cc @kosmikus 

#### Testing

I ran `cabal install --dry-run` for all packages on Hackage with master and the branch with a 90 second timeout, GHC 8.0.1, and index state `2017-07-08T04:20:18Z`.  Then I filtered out all packages where both runs had the same result (success or failure) and the times were within 10%.  I repeated that process three times to eliminate packages that had different run times due to noise.  Then I ran `cabal install --dry-run --max-backjumps=-1` on the remaining packages and averaged three runs.

Since this branch changes a goal-ordering heuristic, I expected a lot of changes in run time in either direction.  Out of 105 changes, 83 were faster, 20 were slower, and 2 involved a timeout on both branches.  (The two that timed out had different run times before I removed the backjump limit.)

package | master result | master time (seconds) | branch result | branch time (seconds) | speedup
--------|---------------|-----------------------|---------------|-----------------------|--------
AesonBson	|	solution	|	3.06	|	solution	|	2.59	|	1.18
DisTract	|	no solution	|	1.48	|	no solution	|	1.34	|	1.1
GPipe-Examples	|	timeout	|	90.01	|	no solution	|	2.11	|	-
[...]

PR #4760, opened 2017-09-12
Backport a bunch of patches to the 2.0 branch

23Skidoo:
Preparing for the 2.0 point release...


23Skidoo:
Not sure why the test output changed (5c4ec078b2bce25c73789e25e06331dbaa4a6b17), seems to be not the case in `master`.

23Skidoo:
```
GHCi runtime linker: fatal error: I found a duplicate definition for symbol
   fps_reverse
whilst processing object file

   /home/travis/.cabal/store/ghc-7.4.2/bytestring-0.10.8.2-45a5149b4862658365e03d6bcce81ec92ecccb767cd5df0de8882538a40bf3ca/lib/libHSbytestring-0.10.8.2-45a5149b4862658365e03d6bcce81ec92ecccb767cd5df0de8882538a40bf3ca.a
This could be caused by:
   * Loading two different object files which export the same symbol
   * Specifying the same object file twice on the GHCi command line
   * An incorrect `package.conf' entry, causing some object to be
     loaded twice.

GHCi cannot safely continue in this situation.  Exiting now.  Sorry.
GHCi exited with ExitFailure 1 (use -v for more information)
```

I think that this should be fixable by nuking the Travis cache, but @ezyang might be interested to look into coming up with a proper fix.

grayjay:
> Not sure why the test output changed (5c4ec07), seems to be not the case in master.

#4599 changed some of the conflict counts on master.

23Skidoo:
@grayjay Thanks! Do you want that patch backported as well?

23Skidoo:
Nuking the store fixed the error in https://github.com/haskell/cabal/pull/4760#issuecomment-328905882. From looking at the code, to fix that properly one would need to invoke `ghci` with `-hide-all-packages` and package IDs in [`Test.Cabal.Server.startServer`](https://github.com/haskell/cabal/blob/master/cabal-testsuite/Test/Cabal/Server.hs#L213).

PR #4764, opened 2017-09-13
Include foreign libs in allBuildInfo (fixes #4763)

### LINKED ISSUE #4763: allBuildInfo does not include foreign library dependencies
This may be a misunderstanding on my part, I'm not familiar with foreign libraries outside of the type `ForeignLib` itself. But it appears that:

* A `ForeignLib` can introduce dependencies on packages
* The `allBuildInfo` function will not return the `BuildInfo` values within `ForeignLib`s

It seems like this means we will not necessarily track dependencies from `ForeignLib`s. Is this intentional?

snoyberg:

This is the same modification I just made to the Stack codebase, which caused an integration test to pass.


snoyberg:
Awesome, will do once Travis/AppVeyor pass.

PR #4786, opened 2017-09-22
Check MIN_VERSION macros are defined before use

### LINKED ISSUE #4785: Use of MIN_VERSION macros breaks Cabal's Setup.hs with GHC<8.0
When trying to compile Cabal's Setup.hs directly using `ghc --make Setup.hs` with GHC<8.0 or when using `-fno-version-macros` with GHC>=8.0 the compilation errors out because a couple of spots in the code now assume `MIN_VERSION` macros are always defined.

I wonder why the CI isn't catching this?

To reproduce:
```
$ ghc --make -fno-version-macros Setup.hs
```

Summary of problem spots:
```
Distribution/Utils/IOData.hs:36:0: error:
     error: missing binary operator before token "("
     #if MIN_VERSION_bytestring(0,10,0)

Distribution/Parsec/ParseResult.hs:21:0: error:
     error: missing binary operator before token "("
     #if MIN_VERSION_base(4,10,0)

Distribution/Parsec/ParseResult.hs:76:0: error:
     error: missing binary operator before token "("
     #if MIN_VERSION_base(4,10,0)

Distribution/Compat/Map/Strict.hs:15:7: error:
    Ambiguous occurrence ‘fromSet’
    It could refer to either ‘Data.Map.fromSet’,
                             imported from ‘Data.Map’ at Distribution/Compat/Map/Strict.hs:22:1-53
                             (and originally defined in ‘containers-0.5.7.1:Data.Map.Base’)
                          or ‘Distribution.Compat.Map.Strict.fromSet’,
                             defined at Distribution/Compat/Map/Strict.hs:30:1
```

DanielG:
Fix #4785

### DIFF
--- a/Cabal/Distribution/Parsec/ParseResult.hs
+++ b/Cabal/Distribution/Parsec/ParseResult.hs
@@ -1,6 +1,17 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP          #-}
 {-# LANGUAGE RankNTypes   #-}
+
+-- For bootstrapping with older GHCs (#4785)
+#ifdef MIN_VERSION_base
+#if MIN_VERSION_base(4,10,0)
+#define HAVE_base_4_10
+#endif
+
+#elif __GLASGOW_HASKELL__ >= 802
+#define HAVE_base_4_10
+#endif
+
 -- | A parse result type for parsers from AST to Haskell types.
 module Distribution.Parsec.ParseResult (
     ParseResult,
@@ -18,7 +29,7 @@ import Distribution.Parsec.Common
        (PError (..), PWarnType (..), PWarning (..), Position (..), zeroPos)
 import Prelude ()
 
-#if MIN_VERSION_base(4,10,0)
+#ifdef HAVE_base_4_10
 import Control.Applicative (Applicative (..))
 #endif
 
@@ -73,7 +84,7 @@ instance Applicative ParseResult where
         success s2 x'
     {-# INLINE (<*) #-}
 
-#if MIN_VERSION_base(4,10,0)
+#ifdef HAVE_base_4_10
     liftA2 f x y = PR $ \ !s0 failure success ->
         unPR x s0 failure $ \ !s1 x' ->
         unPR y s1 failure $ \ !s2 y' ->
--- a/Cabal/Distribution/Compat/Map/Strict.hs
+++ b/Cabal/Distribution/Compat/Map/Strict.hs
@@ -2,15 +2,18 @@
 
 -- For bootstrapping GHC
 #ifdef MIN_VERSION_containers
+
 #if MIN_VERSION_containers(0,5,0)
 #define HAVE_containers_050
 #endif
+
+#elif __GLASGOW_HASKELL__ >= 706
+#define HAVE_containers_050
 #endif
 
 module Distribution.Compat.Map.Strict
     ( module X
-#ifdef HAVE_containers_050
-#else
+#ifndef HAVE_containers_050
     , insertWith
     , fromSet
 #endif
@@ -20,8 +23,8 @@ module Distribution.Compat.Map.Strict
 import Data.Map.Strict as X
 #else
 import Data.Map as X hiding (insertWith, insertWith')
-import qualified Data.Map
-import qualified Data.Set
+import qualified Data.Map (Map)
+import qualified Data.Set (Set)
 
 insertWith :: Ord k => (a -> a -> a) -> k -> a -> Map k a -> Map k a
 insertWith = Data.Map.insertWith'
--- a/Cabal/Distribution/Utils/IOData.hs
+++ b/Cabal/Distribution/Utils/IOData.hs
@@ -1,5 +1,15 @@
 {-# LANGUAGE CPP #-}
 
+-- For bootstrapping with older GHCs (#4785)
+#ifdef MIN_VERSION_bytestring
+#if MIN_VERSION_bytestring(0,10,0)
+#define HAVE_bytestring_0_10
+#endif
+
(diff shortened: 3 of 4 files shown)


Cabal/Distribution/Parsec/ParseResult.hs
```diff
@@ -1,6 +1,14 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP          #-}
 {-# LANGUAGE RankNTypes   #-}
+
+-- For bootstrapping with older GHCs (#4785)
+#ifdef MIN_VERSION_base
+#if MIN_VERSION_base(4,10,0)
+#define HAVE_base_4_10
+#endif
+#endif
```
hvr: I'd feel better if we had at least a `#warning falling back to heuristic mode due to cabal-install < 1.24 && GHC < 8` in an `#else` branch so it's clear that we're in uncertain territory.

23Skidoo:
LGTM, please merge once it's green. Do you intend to also make patches for old release branches?

23Skidoo:
Seems to break the build on Windows.

Cabal/Distribution/Parsec/ParseResult.hs
hvr: See e.g. https://github.com/ekmett/lens/blob/master/Setup.lhs#L25-L32 for a common boilerplate warning

DanielG: I'm not opposed to that in principle at all but I will not that there are already quite a few places in the code where we do things like this. Should I patch those as well? Maybe it would be better to collect all these `#ifdefs` into a single header file and include that where these are needed?

hvr: I know I've seen `#warning` boilerplate in some Setup.hs but I can't find it rightaway, maybe @phadej knows an example... :-)

23Skidoo: Right, there should be only one warning, in `Setup.hs`.

DanielG: I was thinking of defining the `HAVE_` macros in a central header file and generate the warning there.

hvr: @DanielG For some reason I'm a bit worried about introducing an `#include` file as it would require search paths to set up, potentially breaking `ghc --make Setup.hs`

DanielG: Fair enough. Personally I think all CPP is code smell but what're you gonna do ;)

I've been thinking we need a proper Haskell preprocessor for a while now.. `{-# LANGUAGE HPP #-}` anyone?

hvr: > I've been thinking we need a proper Haskell preprocessor for a while now.. `{-# LANGUAGE HPP #-}` anyone?

Yes! A first step would be https://ghc.haskell.org/trac/ghc/wiki/Proposal/NativeCpp ... if you want to work on that, lemme know... :-)

DanielG:
I don't think old releases are affected by this problem but I'll let you know if I find any that are.

hvr:
one of the travis jobs fails with

```
[  8 of 203] Compiling Distribution.Compat.Map.Strict ( Distribution/Compat/Map/Strict.hs, /home/travis/build/haskell/cabal/dist-newstyle/build/x86_64-linux/ghc-7.4.2/Cabal-2.1.0.0/build/Distribution/Compat/Map/Strict.o )

Distribution/Compat/Map/Strict.hs:22:55:

    Module `Data.Map' does not export `fromSet'

cabal: Failed to build Cabal-2.1.0.0 (which is required by test:unit-tests

from Cabal-2.1.0.0, test:parser-tests from Cabal-2.1.0.0 and others).

```

DanielG:
OK I added fallbacks that check the GHC version instead of MIN_VERSION.

hvr:
> OK I added fallbacks that check the GHC version instead of MIN_VERSION.

*cringe* :-)

DanielG:
It doesn't really re-introduce them, they where never gone, just grep for `__GLASGOW_HASKELL__` and you'll see :)

hvr:
Well, tbh, I was hoping to get rid of those remaining `__GLASGOW_HASKELL__`s and replace it by proper MIN_VERSION_xxx ones... I'm just frustrated this basically thwarts my plans... :-)

phadej:
IMHO for old GHC, bootstrap.sh can define `BOOTSTRAP_containers_050` using little `ghc-pkg` invocations. So we don't need to rely on `__GLASGOW_HASKELL__`

23Skidoo:
I remember that we discussed adding a bootstrap script for Cabal-the-library...

23Skidoo:
There's a CI failure with GHC 7.4.2:

```
[  8 of 203] Compiling Distribution.Compat.Map.Strict ( Distribution/Compat/Map/Strict.hs, /home/travis/build/haskell/cabal/dist-newstyle/build/x86_64-linux/ghc-7.4.2/Cabal-2.1.0.0/build/Distribution/Compat/Map/Strict.o )
Distribution/Compat/Map/Strict.hs:30:14:
    Not in scope: Data.Map.insertWith'
Distribution/Compat/Map/Strict.hs:33:13:
    Not in scope: `Data.Map.fromDistinctAscList'
Distribution/Compat/Map/Strict.hs:33:75:
    Not in scope: `Data.Set.toList'
```

DanielG:
I'm abandoning this PR, I fixed my specific problem by talking to Cabal's Setup.hs directly in `cabal-helper` and I don't think there is a way to get all the CI jobs to work with just heuristics.

PR #4788, opened 2017-09-22
Use tree-diff

phadej:


phadej:
Oh dear.
```
cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the packages are required by a dependency of one
of the other targets
```

That's probably because `tree-diff` has `build-type: Custom`?

23Skidoo:
> That's probably because `tree-diff` has `build-type: Custom`?

...and `tree-diff` ends up depending on Cabal via its custom setup. Probably, though the solver should've picked a different version of Cabal for `tree-diff`'s setup dependency... but `--only-dep` is likely too dumb to deal with that.

hvr:
@phadej ...stating the obvious question: would it hurt if tree-diff became build-type:simple? you could still perform CI in tree-diff's git repo with a modified .cabal file...

phadej:
@hvr, I'd need `extra-packages` to make development process smooth

phadej:
`parsers` is `build-type: Custom`.

PR #4791, opened 2017-09-25
Remove the package instance from D.Solver.Modular.Var (closes #4142).

### LINKED ISSUE #4142: Should we remove the package version from flag/stanza variables in the solver?
Here is an example of the current behavior:

```
$ cabal install --dry-run thorn --max-backjumps -1 -v3
[...]
[__4] fail (backjumping, conflict set: base, comonad, transformers)
[__3] rejecting: comonad-1.1.0 (conflict: base==4.9.0.0/installed-4.9..., comonad => base>=4 && <4.4)
[__3] rejecting: comonad-1.0.3, comonad-1.0.2, comonad-1.0.1, comonad-1.0, comonad-0.9.0.1, comonad-0.9.0, comonad-0.7.0, comonad-0.6.2.1, comonad-0.6.2, comonad-0.6.1.2, comonad-0.6.1.1, comonad-0.6.1, comonad-0.6.0, comonad-0.5.0, comonad-0.4.0, comonad-0.3.0, comonad-0.1.1, comonad-0.1.0 (conflict: profunctors => comonad>=1.1 && <1.2)
[__3] fail (backjumping, conflict set: base, comonad, profunctors, transformers)
[__2] fail (backjumping, conflict set: base, bifunctors, comonad, profunctors, semigroupoids, thorn, transformers, transformers-compat, semigroupoids-4.0.3:flag, semigroupoids-4.0.4:flag, semigroupoids-4.2:flag)
[__1] rejecting: base-4.9.0.0, base-4.8.2.0, base-4.8.1.0, base-4.8.0.0, base-4.7.0.2, base-4.7.0.1, base-4.7.0.0, base-4.6.0.1, base-4.6.0.0, base-4.5.1.0, base-4.5.0.0, base-4.4.1.0, base-4.4.0.0, base-4.3.1.0, base-4.3.0.0, base-4.2.0.2, base-4.2.0.1, base-4.2.0.0, base-4.1.0.0, base-4.0.0.0, base-3.0.3.2, base-3.0.3.1 (constraint from non-upgradeable package requires installed instance)
[__0] fail (backjumping, conflict set: base, bifunctors, comonad, profunctors, semigroupoids, thorn, transformers, transformers-compat, semigroupoids-4.0.3:flag, semigroupoids-4.0.4:flag, semigroupoids-4.2:flag)
[...]
```
The final conflict set contains `semigroupoids-4.0.3:flag, semigroupoids-4.0.4:flag, semigroupoids-4.2:flag`.

As I understand it, the version information has no effect on backjumping, because backjumping never uses variables that were assigned below the current node.  And the only way for a conflict set to have more than one version of a flag is for that flag to have been assigned multiple times, which can only happen on multiple branches below the current node.

Some reasons to remove the version:

* Treating all versions of a flag as the same variable could improve conflict counting. If a flag causes a conflict in one version of a package, then it is more likely to cause a conflict in the next version, and the solver should try to assign the flag sooner.

* It would simplify the final conflict set when we display it for #3570.

* Flag variables would be more consistent with variables for setup dependencies, which contain the name, but not the version, of the package that has the setup script.

Does it make sense to remove the version info?  If we remove it, should we remove it from the `Var` type, or just remove it when we add variables to the conflict set?

/cc @kosmikus @edsko @fmthoma

grayjay:

_____________________________________________________________________________________________________

#### Remove D.Solver.Modular.Var.varPI.

This change is necessary to remove the package instance (I) from Var
(issue #4142).  One of the main uses of the package instance is when varPI is
called by the linking phase in order to get the dependencies introduced by flags
and stanzas.  The validation phase also needs to look up dependencies introduced
by flags and stanzas, but it does so by looking up the dependencies once when it
chooses a package and then storing the dependencies in a map.  I refactored the
linking phase to also store dependencies in a map.

#### Remove the package instance from D.Solver.Modular.Var (closes #4142).

This change has several effects:

- The solver no longer includes the package version in messages that relate to
  a package's flags, stanzas, or dependencies.  However, the solver always
  chooses the package version before choosing any flags, stanzas, or
  dependencies for the package, so it should be easy to find the version
  by looking earlier in the log.
- In conflict counting, the solver treats flags with the same name in different
  versions of a package as the same flag.  This change in the conflict counting
  heuristic can improve the solver's efficiency when the same flag causes the
  same conflicts in different versions of a package.  The same applies to
  enabling tests or benchmarks.
- Each flag or stanza can only appear once in a conflict set.  This has no
  effect on behavior, but it simplifies the message containing the final
  conflict set.

Here is an example of the change in a log message.  It only prints
hackage-server's version once, when it first chooses the package.  The conflict
set also has one fewer variable, but that is probably due to the change in
conflict counting.

```diff
 Resolving dependencies...
 cabal: Could not resolve dependencies:
 trying: hackage-server-0.5.0 (user goal)
-trying: hackage-server-0.5.0:+build-hackage-build
-trying: unix-2.7.2.1/installed-2.7... (dependency of hackage-server-0.5.0
+trying: hackage-server:+build-hackage-build
+trying: unix-2.7.2.1/installed-2.7... (dependency of hackage-server
 +build-hackage-build)
-next goal: aeson (dependency of hackage-server-0.5.0 +build-hackage-build)
+next goal: aeson (dependency of hackage-server +build-hackage-build)
 rejecting: aeson-1.2.2.0, aeson-1.2.1.0, aeson-1.2.0.0, aeson-1.1.2.0,
 aeson-1.1.1.0, aeson-1.1.0.0, aeson-1.0.2.1, aeson-1.0.2.0, aeson-1.0.1.0,
 aeson-1.0.0.0, aeson-0.11.3.0, aeson-0.11.2.1, aeson-0.11.2.0, aeson-0.11.1.4,
 aeson-0.11.1.3, aeson-0.11.1.2, aeson-0.11.1.1, aeson-0.11.1.0,
 aeson-0.11.0.0, aeson-0.9.0.1, aeson-0.9.0.0, aeson-0.8.1.1, aeson-0.8.1.0,
 aeson-0.8.0.2, aeson-0.7.0.6, aeson-0.7.0.4, aeson-0.6.2.1, aeson-0.6.2.0
 (conflict: hackage-server +build-hackage-build => aeson==0.6.1.*)
 rejecting: aeson-0.6.1.0 (conflict: unix => time==1.6.0.1/installed-1.6...,
 aeson => time<1.5)
 rejecting: aeson-0.6.0.2, aeson-0.6.0.1, aeson-0.6.0.0, aeson-0.5.0.0,
 aeson-0.4.0.1, aeson-0.4.0.0, aeson-0.3.2.14, aeson-0.3.2.13, aeson-0.3.2.12,
 aeson-0.3.2.11, aeson-0.3.2.10, aeson-0.3.2.9, aeson-0.3.2.8, aeson-0.3.2.7,
 aeson-0.3.2.6, aeson-0.3.2.5, aeson-0.3.2.4, aeson-0.3.2.3, aeson-0.3.2.2,
 aeson-0.3.2.1, aeson-0.3.2.0, aeson-0.3.1.1, aeson-0.3.1.0, aeson-0.3.0.0,
 aeson-0.2.0.0, aeson-0.1.0.0, aeson-0.10.0.0, aeson-0.8.0.1, aeson-0.8.0.0,
 aeson-0.7.0.5, aeson-0.7.0.3, aeson-0.7.0.2, aeson-0.7.0.1, aeson-0.7.0.0
 (conflict: hackage-server +build-hackage-build => aeson==0.6.1.*)
 After searching the rest of the dependency tree exhaustively, these were the
-goals I've had most trouble fulfilling: aeson, hackage-server,
-hackage-server-0.5.0:build-hackage-build,
-hackage-server-0.4:build-hackage-mirror, template-haskell
+goals I've had most trouble fulfilling: aeson,
+hackage-server:build-hackage-build, hackage-server, template-haskell
```

I ran hackage-benchmark to compare this commit with master (two commits
earlier).  I used --min-run-time-percentage-difference-to-rerun=10 to only rerun
packages if the run times differed by more than 10% in the first trial, and
defaults for the rest of the options (10 trials, p-value of 0.05, 90 second
timeout). The index state was "2017-09-24T03:35:06Z".

1 is master, and 2 is this commit:

package      |      result1  |     result2     |        mean1   |    mean2  |   stddev1  |   stddev2   |  speedup
----------------|----------------|-----------------|------------------|--------------|---------------|---------------|--------------------
CC-delcont-ref   |  Solution   |   Solution     |      1.467s    |  1.505s   |   0.019s   |   0.100s   |   0.975
ascii-cows    |     Solution   |   Solution      |     1.827s   |   1.758s   |   0.159s   |   0.012s   |   1.040
opaleye-classy  |   NoInstallPlan | NoInstallPlan   |   4.588s  |    4.070s   |   0.043s   |   0.032s   |   1.127
range-space    |    NoInstallPlan | NoInstallPlan   |   2.642s   |   2.299s   |   0.016s   |   0.016s   |   1.149
rts         |       PkgNotFound |  PkgNotFound    |    1.323s   |   1.327s  |    0.032s  |    0.033s   |   0.997
servant-auth-docs | Solution   |   Solution       |    1.968s   |   1.998s   |   0.017s   |   0.074s   |   0.985
thorn       |       BackjumpLimit | NoInstallPlan   |   4.793s  |    3.141s   |   0.050s   |   0.034s   |   1.526
unordered-intmap |  Solution   |   Solution      |     1.502s   |   1.511s   |   0.081s   |   0.047s   |   0.994

I looked at the solver logs for the three packages with the largest changes in
run time, opaleye-classy, range-space, and thorn.  Each one showed that the
solver started preferring a flag in an older version of a package after it had
caused conflicts in a newer version of the package.

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Assignment.hs
+++ b/cabal-install/Distribution/Solver/Modular/Assignment.hs
@@ -65,13 +65,13 @@ toCPs (A pa fa sa) rdm =
     -- complete flag assignment by package.
     fapp :: Map QPN FlagAssignment
     fapp = M.fromListWith (++) $
-           L.map (\ ((FN (PI qpn _) fn), b) -> (qpn, [(fn, b)])) $
+           L.map (\ ((FN qpn fn), b) -> (qpn, [(fn, b)])) $
            M.toList $
            fa
     -- Stanzas per package.
     sapp :: Map QPN [OptionalStanza]
     sapp = M.fromListWith (++) $
-           L.map (\ ((SN (PI qpn _) sn), b) -> (qpn, if b then [sn] else [])) $
+           L.map (\ ((SN qpn sn), b) -> (qpn, if b then [sn] else [])) $
            M.toList $
            sa
     -- Dependencies per package.
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -21,7 +21,7 @@ module Distribution.Solver.Modular.Builder (
 
 import Data.List as L
 import Data.Map as M
-import Prelude hiding (pi, sequence, mapM)
+import Prelude hiding (sequence, mapM)
 
 import Distribution.Solver.Modular.Dependency
 import Distribution.Solver.Modular.Flag
@@ -63,13 +63,13 @@ extendOpen qpn' gs s@(BS { rdeps = gs', open = o' }) = go gs' o' gs
   where
     go :: RevDepMap -> [OpenGoal] -> [PotentialGoal] -> BuildState
     go g o []                                             = s { rdeps = g, open = o }
-    go g o ((PotentialGoal (Flagged fn@(FN pi _) fInfo t f)  ) : ngs) =
-        go g (FlagGoal fn fInfo t f (flagGR pi) : o) ngs
+    go g o ((PotentialGoal (Flagged fn@(FN qpn _) fInfo t f)  ) : ngs) =
+        go g (FlagGoal fn fInfo t f (flagGR qpn) : o) ngs
       -- Note: for 'Flagged' goals, we always insert, so later additions win.
       -- This is important, because in general, if a goal is inserted twice,
       -- the later addition will have better dependency information.
-    go g o ((PotentialGoal (Stanza sn@(SN pi _) t)           ) : ngs) =
-        go g (StanzaGoal sn t (flagGR pi) : o) ngs
+    go g o ((PotentialGoal (Stanza sn@(SN qpn _) t)           ) : ngs) =
+        go g (StanzaGoal sn t (flagGR qpn) : o) ngs
     go g o ((PotentialGoal (Simple (LDep dr (Dep _ qpn _)) c)) : ngs)
       | qpn == qpn'       = go                            g               o  ngs
           -- we ignore self-dependencies at this point; TODO: more care may be needed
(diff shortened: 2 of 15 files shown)


23Skidoo:
AFAICT, this is good to go, though would be nice if @kosmikus could take a look.

23Skidoo:
I think I'll just merge this, looks uncontroversial enough.

kosmikus:
Yes, LGTM.

grayjay:
Thanks!

PR #4801, opened 2017-09-28
Add DerivingStrategies and UnboxedSums to Language

RyanGlScott:
These extensions were introduced in GHC 8.2.1.

### DIFF
--- a/Cabal/Language/Haskell/Extension.hs
+++ b/Cabal/Language/Haskell/Extension.hs
@@ -795,6 +795,13 @@ data KnownExtension =
   -- as injective.
   | TypeFamilyDependencies
 
+  -- | Allow multiple @deriving@ clauses, each optionally qualified with a
+  -- /strategy/.
+  | DerivingStrategies
+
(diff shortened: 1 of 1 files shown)


23Skidoo:
Do you want this backported to 2.0

RyanGlScott:
Given that it's an API breaking change, probably not.

hvr:
@bgamari btw, this is something that should be part of the GHC release checklist (if it isn't yet). iirc we have or had a testcase in GHC's testsuite which covers complaining about missing extensions in lib:Cabal

RyanGlScott:
Indeed that very test case ([`driver/T4437.hs`](http://git.haskell.org/ghc.git/blob/e515c7f37be97e1c2ccc497ddd0a730e63ddfa82:/testsuite/tests/driver/T4437.hs)) is what prompted me to add these. Currently, `DerivingStrategies` and `UnboxedSums` are whitelisted in `expectedGhcOnlyExtensions`, so they slipped through the cracks.

PR #4809, opened 2017-10-03
Adds `new-update`

angerman:
new-update uses the new-style logic to update the repositories.  As such it
respects `repository` fields in the `cabal.project(.local)` file and updates
them as well.  This is essential when working with hackage overlays, where
the overlay repositories are specified as `repository` fields in the
`cabal.project(.local)` file.

### DIFF
--- a/cabal-install/Distribution/Client/CmdUpdate.hs
+++ b/cabal-install/Distribution/Client/CmdUpdate.hs
@@ -72,8 +72,8 @@ updateCommand = Client.installCommand {
   }
 
 data UpdateRequest = UpdateRequest
-  { updateRequestRepoName :: String
-  , updateRequestRepoState :: IndexState
+  { _updateRequestRepoName :: String
+  , _updateRequestRepoState :: IndexState
   } deriving (Show)
 
 instance Text UpdateRequest where
@@ -117,7 +117,7 @@ updateAction (applyFlagDefaults -> (configFlags, configExFlags, installFlags, ha
 
     let reposToUpdate = case updateRepoRequests of
           [] -> repos
-          updateRequests -> let repoNames = map updateRequestRepoName updateRequests
+          updateRequests -> let repoNames = map _updateRequestRepoName updateRequests
                             in filter (\r-> repoName r `elem` repoNames) repos
 
     case reposToUpdate of
--- a/cabal-install/Distribution/Client/CmdUpdate.hs
+++ b/cabal-install/Distribution/Client/CmdUpdate.hs
@@ -1,0 +1,143 @@
+{-# LANGUAGE CPP, NamedFieldPuns, RecordWildCards, ViewPatterns #-}
+
+-- | cabal-install CLI command: update
+--
+module Distribution.Client.CmdUpdate (
+    updateCommand,
+    updateAction,
+  ) where
+
+import Distribution.Client.ProjectOrchestration
+import Distribution.Client.ProjectConfig
+         ( ProjectConfig(..)
+         , projectConfigWithSolverRepoContext )
+import Distribution.Client.Types
+         ( Repo(..), RemoteRepo(..), maybeRepoRemote )
+import Distribution.Client.HttpUtils
+         ( DownloadResult(..) )
+import Distribution.Client.FetchUtils
+         ( downloadIndex )
+import Distribution.Client.JobControl
+         ( newParallelJobControl, spawnJob, collectJob )
+import Distribution.Client.Setup
+         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags, UpdateFlags (updateIndexState)
+         , applyFlagDefaults, defaultUpdateFlags, RepoContext(..) )
+import Distribution.Simple.Setup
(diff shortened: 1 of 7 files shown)


angerman:
This is basically a merge of the `CmdFreeze.hs` (new-freeze) and old `Update.hs` (update) files.

The command line options are not correct, so if someone could provide some guidance for those, I'd be happy to fix those properly.

angerman:
@hvr, @23Skidoo can I get a review of this?

angerman:
This still does **not** properly respect the index state, as I have no idea how to thread that through properly.

However it does everything else as needed. Unless the index-state threading is trivial, I'd prefer that to be a separate PR.

hvr:
@angerman I agree; taking into account the `index-state` business will require more thought (and likely a departure from the old-update UI). Let's get this PR wrapped up; and do refinements incrementally in future PRs.

angerman:
@hvr, yea, that's just plain stupid.

I've introduced a new datatype
```
data UpdateRequest = UpdateRequest
   { updateRequestRepoName :: String
   , updateRequestRepoState :: IndexState
   } deriving (Show)
```

and because I have not *yet* used `updateRequestRepoState`, this fails.

What's the proper cabal approved fix for this?

hvr:
@angerman I'd just prefix it by `_` temporarily... it's not like this is a public API anyway

angerman:
Can we merge now?

cabal-install/Distribution/Client/CmdUpdate.hs
```diff
@@ -0,0 +1,179 @@
...
+         ( CommandUI(..), usageAlternatives )
+import qualified Distribution.Client.Setup as Client
+
+import qualified Hackage.Security.Client as Sec
+
+updateCommand :: CommandUI (ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags)
+updateCommand = Client.installCommand {
+  commandName         = "new-update",
+  commandSynopsis     = "Updates list of known packages.",
+  commandUsage        = usageAlternatives "new-update" [ "[FLAGS]" ],
```
hvr: I think `[FLAGS]` should reflect that we may also supply a list of repo labels. C.f.

```
Usage: cabal new-build [TARGETS] [FLAGS]
```

vs.

```
Usage: cabal new-update [FLAGS]
```

which would be better as

```
Usage: cabal new-update [FLAGS] [REPOS]
```

cabal-install/Distribution/Client/CmdUpdate.hs
```diff
@@ -0,0 +1,179 @@
...
+  commandDescription  = Just $ \_ -> wrapText $
+        "For all known remote repositories, download the package list.",
+  commandNotes        = Just $ \pname ->
+        "Examples:\n"
+     ++ "  " ++ pname ++ " new-update\n"
+     ++ "    Download the package list for all known remote repositories.\n\n"
```
hvr: We may want to add an example here showing how to update a single repo

and we may point out that a `TARGET` has the format `<repo-id>[,<index-state>]` where `<index-state>` is the same syntax as supported by the `--index-state` flag which currently is:

```
    --index-state=STATE              Use source package index state as it
                                     existed at a previous time. Accepts
                                     unix-timestamps (e.g. '@1474732068'),
                                     ISO8601 UTC timestamps (e.g.
                                     '2016-09-24T17:47:48Z'), or 'HEAD'
                                     (default: 'HEAD').
```

cabal-install/Distribution/Client/CmdUpdate.hs
```diff
@@ -0,0 +1,179 @@
...
+          writeFileAtomic (dropExtension indexPath) . maybeDecompress
+                                                  =<< BS.readFile indexPath
+          updateRepoIndexCache verbosity (RepoIndex repoCtxt repo)
+    RepoSecure{} -> repoContextWithSecureRepo repoCtxt repo $ \repoSecure -> do
+      let index = RepoIndex repoCtxt repo
+      -- NB: This may be a nullTimestamp if we've never updated before
+      current_ts <- currentIndexTimestamp (lessVerbose verbosity) repoCtxt repo
+      -- NB: always update the timestamp, even if we didn't actually
+      -- download anything
+      writeIndexTimestamp index (fromFlag (updateIndexState updateFlags))
```
hvr: So that's the call where we'd need to propagate the `<index-state>` value to... what's the problem? :-)

hvr: @angerman just to make this more explicit:

As you can see, `writeIndexTimestamp` writes the `01-index.timestamp` file which is stored alongside each `01-index.tar` in the package/repo cache:

```hs
-- | Write the 'IndexState' to the filesystem
writeIndexTimestamp :: Index -> IndexState -> IO () 
writeIndexTimestamp index st
  = writeFile (timestampFile index) (display st)
```

so all you need to do, is pass `writeIndexTimestamp` the optional `[,<index-state>]` argument associated with that specific repository.

angerman:
Let's do this!

PR #4813, opened 2017-10-04
WIP rewording docs

hvr:
[skip ci]

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -1729,9 +1729,8 @@ system-dependent values for these fields.
        than ``1.0``. This is not an issue with the caret-operator
        ``^>=`` described below.
 
-    Starting with Cabal 2.0, there's a new syntactic sugar to express
-    PVP_-style
-    major upper bounds conveniently, and is inspired by similar
+    Starting with Cabal 2.0, there's a new version operator to express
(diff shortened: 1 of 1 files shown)


Cabal/doc/developing-packages.rst
```diff
@@ -1732,15 +1731,46 @@ system-dependent values for these fields.
...
+    contract we can (positively) infer right away that all versions
+    satisfying ``foo >= 1.2.3.4 && < 1.3`` will be compatible as well.
 
-    Ignoring the signaling intent, the equivalences are
+    .. Note::
+
+       More generally, the PVP_ contract implies that we can safely
+       relax the lower bound to ``>= 1.2``, because if we know that
+       ``foo-1.2.3.4`` is semantically compatible, then so is
+       ``foo-1.2``. But we'd need to perform additional static
```
bitonic: that's not true, i think you meant to say `foo-1.2.3` here and below.

hvr: I really intended to say `foo-1.2` here, as this Note is describing the property "if foo-1.2.3.4 compiles & is semantically correct/compatible; then *if* foo-1.2 compiles as well, then foo-1.2 is semantically correct/compatible as well". What you're referring to is the awkwardness, that `^>= 1.2.3.4` doesn't desugar to `>= 1.2.3 && < 1.3` (i.e. dropping the patch-level).

Issue #4558, opened 2017-06-09
`cabal new-install` UI design concept (WIP)


hvr:
## Status

* [ ] Executables
  * [ ] Local
  * [x] Nonlocal #4825 (see also #5000)
* [ ] Libraries
  * [ ] Local
  * [ ] Nonlocal

---

There's at least two axes for use-cases which need to be considered for the UI design: Most importantly is the distinction between

 - making executables available
 - making libraries available in `GHC_ENVIRONMENT`s (for GHC 8.0.2+)

And then sources can be distinguished by being either

 - Non-local packages
 - Local packages/projects

With the new nix-store paradigm, the idiomatic way to "install" things would be to first populate the requested target in the nix-store, and install a reference to the materialised artifact in a "view" (this signficantly differs depending on whether it's an executable or a library).

In this paradigm, the "installation" rather refers to installing a symlink or adding a reference to a package-env, than to build the actual artifacts. In some cases, all a `cabal new-install` operation does degenerates into replacing the target of an existing symlink (e.g. if you first "install" `alex-3.2.0`, and then you install `alex-3.2.1`, and then decide to install `alex-3.2.0` again; the 3rd operation would merely update the symlink, as the store would already contain both versions of `alex`)

## Basic UI

The basic UI would look like

    $ cabal new-install <build-target-spec> [<install-location>]

(NOTE: as of #4825, the syntax is `cabal new-install [--symlink-bindir=<install-location>] <build-target-spec>`)

A build-target would be something like 

Omission of `<install-location>` would denote a default location.

For libraries the default would be the current value of `GHC_ENVIRONMENT`, or otherwise the `default` environment (this matches the ghc-env lookup logic of GHC 8.0.2+).

For executables, the install location denotes a `bin` folder; by default this would be the value configured in `~/.cabal/config` (e.g. `${HOME}/.cabal/bin`) or possibly in the in-scope `cabal.project`.

### Non-local packages

Installing executables from non-local packages is the easiest case; in fact, this can be emulated already now via script like https://gist.github.com/hvr/c77c54d682555b7dd4fe1248732fe978

Which first causes the requested executable to be materialised (if it wasn't already cached) in the nix store via a dummy package description containing `build-tool-depends: ${PKG}:${EXE}`, and then
locates the resulting executable to install a symlink to the desired folder-location.

### Local packages

This case is currently less straightforward when the non-local components have associated package-data (#4120).

[...]

PR #4825, opened 2017-10-14
new-install: nonlocal exes

fgaz:
Add the first part of the new-install command: nonlocal exes.

See #4558 for the design concept.

This part of the command installs executables from outside of a project
(ie from hackage) in the store and then symlinks them in the cabal bin
directory.

This is done by creating a dummy project and adding the targets as extra
packages.

---

* [ ] Tests
* [ ] https://github.com/haskell/cabal/pull/4734#discussion_r137924008

---

Some questions:

* How do I test this? A package test would probably require hackage access. Is there a way to add a local source?
* Is the dummy project thing the right way of doing this? And should I put that code in a more generic module? Or maybe I should substitute the dummy project with some sort of global project?
* How do I get the default `~/.cabal/bin` directory? I couldn't find it in the old install command. Is it hardcoded somewhere? Do I have to hardcode it? What about windows?

---

includes and closes #4734

<small>ping @dcoutts @hvr</small>

### DIFF
--- a/cabal-install/Distribution/Client/CmdBench.hs
+++ b/cabal-install/Distribution/Client/CmdBench.hs
@@ -222,10 +222,6 @@ renderTargetProblem (TargetProblemNoTargets targetSelector) =
            ++ renderTargetSelector targetSelector ++ "."
 
       _ -> renderTargetProblemNoTargets "benchmark" targetSelector
-  where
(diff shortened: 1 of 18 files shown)


fgaz:
Oh well, I'll just merge this. Tests and docs will be in another pr

Issue #4808, opened 2017-10-03
`./Setup configure --flags= ...` fails with ghc-8.2.2-rc1


duog:
The cabal packaged with ghc-8.2.2-rc1 (Cabal commit: c84a3c72196d0a5361f8ab77c6d8cb63b7a5d55d) fails when passed Setup configure is passed "--flags=".

See a log here:
https://gist.github.com/duog/260589ccbf63200addbea00f25436a48

23Skidoo:
@hvr just confirmed that this is a regression relative to 2.0.0.2.

hvr:
@bgamari @23Skidoo  @phadej 

I've noticed this occurs on HEAD too; one way to workaround this is the following change to the parser:

```hs
diff --git a/Cabal/Distribution/Types/GenericPackageDescription.hs b/Cabal/Distribution/Types/GenericPackageDescription.hs
index 1547ade..a278370 100644
--- a/Cabal/Distribution/Types/GenericPackageDescription.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription.hs
@@ -159,7 +159,7 @@ dispFlagAssignment = Disp.hsep . map (Disp.text . showFlagValue)
 
 -- | Parses a flag assignment.
 parsecFlagAssignment :: ParsecParser FlagAssignment
-parsecFlagAssignment = P.sepBy1 (onFlag <|> offFlag) P.skipSpaces1
+parsecFlagAssignment = P.sepBy (onFlag <|> offFlag) P.skipSpaces1
   where
     onFlag = do
         P.optional (P.char '+')
@@ -172,7 +172,7 @@ parsecFlagAssignment = P.sepBy1 (onFlag <|> offFlag) P.skipSpaces1
 
 -- | Parses a flag assignment.
 parseFlagAssignment :: Parse.ReadP r FlagAssignment
-parseFlagAssignment = Parse.sepBy1 parseFlagValue Parse.skipSpaces1
+parseFlagAssignment = Parse.sepBy parseFlagValue Parse.skipSpaces1
   where
     parseFlagValue =
           (do Parse.optional (Parse.char '+')
```

Fwiw, `dispFlagAssignment []` results in an empty output; so this would be consistent.

23Skidoo:
This was fixed by #4830.

PR #4830, opened 2017-10-18
RFC: Potential fix for #4808

hvr:
See https://github.com/haskell/cabal/issues/4808#issuecomment-337048211

### DIFF
--- a/Cabal/Distribution/Types/GenericPackageDescription.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription.hs
@@ -159,7 +159,7 @@ dispFlagAssignment = Disp.hsep . map (Disp.text . showFlagValue)
 
 -- | Parses a flag assignment.
 parsecFlagAssignment :: ParsecParser FlagAssignment
-parsecFlagAssignment = P.sepBy1 (onFlag <|> offFlag) P.skipSpaces1
+parsecFlagAssignment = P.sepBy (onFlag <|> offFlag) P.skipSpaces1
   where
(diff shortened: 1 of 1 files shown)


phadej:
Test would be great.

I can only shake my head around this. If only we could have versioned APIs to call older `Setup.hs`.

hvr:
@23Skidoo I'm not sure yet how such a test would look like; we can probably do easy roundtrip tests (and it would help if FlagAssignment was a newtype, so we could attach parse/display instances to it). 

(relatedly, @ezyang mentioned in https://github.com/haskell/cabal/pull/4710#issuecomment-323524170 we don't have many communication tests with different lib:Cabal versions)

So which kind of tests did you (& @phadej) have in mind specifically?

bgamari:
What is the status of this? For the record, this is something that 8.2.2 is blocking on.

phadej:
@hvr, I'm ok for this going to 2.0 as is, but for `master` could we change `FlagAssignment` to the newtype and do roundtrip stuff?

phadej:
cherry-picked on 2.0. @bgamari could you verify it unblocks GHC-8.2.2?

hvr:
> but for master could we change FlagAssignment to the newtype and do roundtrip stuff?

oh yes; in fact I have already started on refactoring it into a newtype to see how much places I'd have to touch; it's not bad at all :-)

bgamari:
Thanks @phadej. That should do it.

PR #4836, opened 2017-10-21
Remove Nix expressions

ttuegel:
Including Nix expressions in Haskell projects, particularly core infrastructure,
is harmful for the following reasons:

1. The expressions are never maintained. In this case, I should know since I'm
the one who isn't maintaining them.
2. In the unlikely event a Nix expression is up-to-date, it still only works
with a particular version of Nixpkgs. Unless Nixpkgs is included in a submodule,
this is of no use to anybody.
3. Creating Nix expressions for Haskell packages is automated anyway and anyone
who is interested in having them knows how to generate them.
4. We who are working on Cabal need to be "eating our own dog food" or "wearing
the hair shirt" as the case may be.


michalrus:
Somewhat related: https://github.com/NixOS/nixpkgs/issues/28931

peti:
I am not happy about that decision. I, for one, was using (and updating) those Nix files and their absence now is certainly not going to make my life any more convenient than it was before. :-(

23Skidoo:
I don't use/understand this stuff and have nothing against adding it back if @peti agrees to maintain it, but maybe @ttuegel disagrees.

michalrus:
Why not have a very simple `default.nix` that just imports the Nixpkgs definition? Then it’s convenient for Cabal devs that use Nix and there’s a single—and maintained—source of truth of how to build Cabal in the Nix context.

Probably that `import` should use a pinned version of Nixpkgs.

ttuegel:
> if @peti agrees to maintain it, but maybe @ttuegel disagrees.

I have no objection. I do think we should encourage (strongly) developers to use the project-based commands (`new-*`) instead where possible.

> Why not have a very simple default.nix that just imports the Nixpkgs definition?

It is usually behind the development version.

> Probably that import should use a pinned version of Nixpkgs.

Definitely!

PR #4839, opened 2017-10-22
Implement native DNS hackage-security bootstrap via resolv/windns packages

hvr:
This avoids calling out to an external `nslookup` program and having
to screen scrape its output. The `resolv` library supports all
platforms which provide the ubiquitous `libresolv` API.

This is hidden behind the manual cabal flag `resolv` which is
currently disabled by default.

### DIFF
--- a/cabal-install/Distribution/Client/Security/DNS.hs
+++ b/cabal-install/Distribution/Client/Security/DNS.hs
@@ -1,17 +1,23 @@
+{-# LANGUAGE CPP #-}
+
 module Distribution.Client.Security.DNS
     ( queryBootstrapMirrors
     ) where
 
 import Prelude ()
 import Distribution.Client.Compat.Prelude
-
+import Network.URI (URI(..), URIAuth(..), parseURI)
+import Distribution.Verbosity
 import Control.Monad
 import Control.DeepSeq (force)
 import Control.Exception (SomeException, evaluate, try)
-import Network.URI (URI(..), URIAuth(..), parseURI)
-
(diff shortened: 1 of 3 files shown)


23Skidoo:
Why not enable this by default?

hvr:
@23Skidoo because right now, `resolv` needs `binary >= 0.7.5` (which is what GHC 7.10.3 is bundled with). I think that might be in line with our coding convention which says 

> cabal-install must also be buildable on all supported GHCs, although it does not have to be buildable out-of-the-box. Instead, the cabal-install/bootstrap.sh script must be able to download and install all of the dependencies (this is also checked by CI). 

So all I'd need to do to enable this by default is tweaking our bootstrap script.

phadej:
If it's not the default, then we need a CI job to test it, otherwise LGTM

hvr:
@23Skidoo I've toggled to DNS support to default-on and adapted the bootstrap script... if it wasn't for those transient network failures and the spurious failures on appveyor I think this PR would be green

hvr:
rebased; https://travis-ci.org/haskell/cabal/builds/293051072 was the previous run; unfortunately this pushes the GHC 7.8.4 on OSX job over the time limit...

hvr:
@23Skidoo fwiw, I think may just need this to give it enough time to cache the additional cabal deps it had to built. Maybe it's time for that hack we talked about, i.e. check how much time has passed shortly before the time-consuming testsuite starts, and abort the job if it's clearly above the time-budget needed....

Issue #4737, opened 2017-09-02
cli defaults override new-build cabal.project settings


hvr:
It appears the fix in #4657 wasn't enough.

I was looking into why setting

```
allow-boot-library-installs: True
```
in a `cabal.project` file would have no effect, and it turns out it's because in `rebuildProjectConfig` the cli-flags have highest priority, due to

```hs
    return (projectConfig <> cliConfig, localPackages)
```

however, in #4657 we deliberately injected `applyDefaults` so the `cliConfig` constructed via

```hs
    cliConfig = commandLineFlagsToProjectConfig
                  globalFlags configFlags configExFlags
                  installFlags haddockFlags
```

already has those defaults.

And here's a dumped version of what `cliConfig` looks like for a simple invocation of `cabal new-build` w/o any CLI arguments:

```hs
ProjectConfig{projectPackages = [], projectPackagesOptional = [],
              projectPackagesRepo = [], projectPackagesNamed = [],
              projectConfigBuildOnly =
                ProjectConfigBuildOnly{projectConfigVerbosity = NoFlag,
                                       projectConfigDryRun = Flag False, projectConfigOnlyDeps = Flag False,
                                       projectConfigSummaryFile = [], projectConfigLogFile = NoFlag,
                                       projectConfigBuildReports = Flag NoReports,
                                       projectConfigReportPlanningFailure = Flag False,
                                       projectConfigSymlinkBinDir = NoFlag, projectConfigOneShot = Flag False,
                                       projectConfigNumJobs = NoFlag, projectConfigKeepGoing = Flag False,
                                       projectConfigOfflineMode = Flag False, projectConfigKeepTempFiles = Flag False,
                                       projectConfigHttpTransport = NoFlag, projectConfigIgnoreExpiry = NoFlag,
                                       projectConfigCacheDir = NoFlag, projectConfigLogsDir = NoFlag,
                                       projectConfigStoreDir = NoFlag},
              projectConfigShared =
                ProjectConfigShared{projectConfigDistDir = NoFlag,
                                    projectConfigConfigFile = NoFlag, projectConfigProjectFile = NoFlag,
                                    projectConfigHcFlavor = NoFlag, projectConfigHcPath = NoFlag,
                                    projectConfigHcPkg = NoFlag,
                                    projectConfigHaddockIndex = Flag "$datadir/doc/$arch-$os-$compiler/index.html",
                                    projectConfigRemoteRepos = [], projectConfigLocalRepos = [],
[...]

PR #4840, opened 2017-10-23
Make `--enable-documentation` work for non-local packages

alexbiehl:
Before non-local packages would receive all their configuration like this:

```
  perpkg = maybe mempty f (Map.lookup (packageName pkg) perPackageConfig)
```

If there is no explicit configuration for a package the default `mempty` is used.

We add `projectConfigNonLocalPackages :: PackageConfig` field to `ProjectConfig`. `projectConfigNonLocalPackages` reflects cabals global configuration. Instead of falling back back to `mempty` directly we lookup up the desigured configuration in this new field.

- Currently `projectConfigNonLocalPackages` is the same as `projectConfigLocalPackages`. We need to sort out what we want here.
- Building with `documentation: True` does not work yet. But it will once we sort out https://github.com/haskell/cabal/issues/4737
- Find a suitable verb: "Haddocking package-x.y (lib)..." sounds strange

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -936,6 +936,13 @@ buildAndInstallUnpackedPackage verbosity
     annotateFailure mlogFile BuildFailed $
       setup buildCommand buildFlags
 
+    -- Haddock phase
+    whenHaddock $ do
+      when isParallelBuild $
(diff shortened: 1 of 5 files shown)


23Skidoo:
@alexbiehl There are some legit-looking failures, though, can you look into them?

alexbiehl:
This has been addressed in https://github.com/haskell/cabal/commit/c094940abc4266c8d013853b66271540cb2dc7f0.

PR #4841, opened 2017-10-23
Apply 'any' qualifier to new-freeze constraints (fixes #4832).

### LINKED ISSUE #4832: new-freeze does not freeze dependencies of executables
And even though some dependency is frozen, it will be rebuilt because an executable it depends on is updated. In the worst case, all your frozen dependencies will be re-built after a `cabal update` because somethink like "happy" has a new version.

`cabal-install-2.0.0.0`. The observed behaviour is consistent independent of using freeze file or project file `constraints` field.

Reproduction: Init empty cabal lib project, add dependency `gkt3`, try to restrict version of `happy` (an executable dependency) to something non-latest.

grayjay:

new-freeze produces constraints that allow all versions of each package from the
install plan, in order to constrain setup and build tool dependencies without
using qualified constraints. Since the constraints should apply to top-level,
setup, and build tool dependencies, they should use an 'any' qualifier.

I also added a package test for new-freeze.

### DIFF
--- a/cabal-install/Distribution/Client/CmdFreeze.hs
+++ b/cabal-install/Distribution/Client/CmdFreeze.hs
@@ -173,7 +173,7 @@ projectFreezeConstraints plan =
     versionConstraints :: Map PackageName [(UserConstraint, ConstraintSource)]
     versionConstraints =
       Map.mapWithKey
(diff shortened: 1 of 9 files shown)


23Skidoo:
Merged, thanks!

grayjay:
Thanks!

Issue #4845, opened 2017-10-24
cabal-install will not read repository from custom config file


ttuegel:
I copied the repository information from the default config file into my own config file, like so:
```
repository hackage.haskell.org
  url: http://hackage.haskell.org/

extra-lib-dirs: /nix/store/v2pvzmmkar9dsfx4s7gskf2q0zp9wlsn-openblas-0.2.20/lib /nix/store/zcrrq61k354gnjg80v840p232m6xdh3y-zlib-1.2.11-dev/lib /nix/store/1gymfh2rkiii6c5flk8w93xlqbp58g8h-zlib-1.2.11/lib
extra-include-dirs: /nix/store/v2pvzmmkar9dsfx4s7gskf2q0zp9wlsn-openblas-0.2.20/include /nix/store/zcrrq61k354gnjg80v840p232m6xdh3y-zlib-1.2.11-dev/include

program-locations
  ar-location: /nix/store/9i5bj3c48vrdb4m4khbhaipvhy3m0bh0-binutils-2.28.1/bin/ar
  gcc-location: /nix/store/qh1ikg5lsnhiglsrfjz9k69fhzq17v8h-gcc-wrapper-6.4.0/bin/gcc
  ld-location: /nix/store/qh1ikg5lsnhiglsrfjz9k69fhzq17v8h-gcc-wrapper-6.4.0/bin/ld
  strip-location: /nix/store/9i5bj3c48vrdb4m4khbhaipvhy3m0bh0-binutils-2.28.1/bin/strip
  tar-location: /nix/store/q6nivamar2h6havsq73iyhcdnfd5jw7h-gnutar-1.29/bin/tar
```
According to the default configuration file created by cabal-install, this is all the repository information that is required. Now cabal-install thinks there is no remote repository. `cabal update` gives
```
fromFlag NoFlag. Use fromFlagOrDefault
CallStack (from HasCallStack):
  error, called at ./Distribution/Simple/Setup.hs:169:21 in Cabal-2.1.0.0-7Zzlk2k1PajDtaVal7GI34:Distribution.Simple.Setup
  fromFlag, called at ./Distribution/Client/GlobalFlags.hs:142:8 in main:Distribution.Client.Global
```
which points to a missing default remote repository.

ttuegel:
The answer is that one must add the line,
```
remote-repo-cache: /path/to/some/directory
```
[...]

PR #4846, opened 2017-10-24
Initialize globalCacheDir in baseSavedConfig

ttuegel:
globalCacheDir (remote-repo-cache) is now initialized by default because only
cabal-install can determine the correct default location relative to
defaultCabalDir. The globalLogsDir (build-logs) and globalWorldFile (world-file)
already had the same behavior.

Resolves: #4845.

I tested this change by removing the `remote-repo-cache` line from my `$HOME/.cabal/config` and confirming that a simple `cabal update` works. (This command requires `remote-repo-cache` to be set.)

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -421,6 +421,7 @@ instance Semigroup SavedConfig where
 baseSavedConfig :: IO SavedConfig
 baseSavedConfig = do
   userPrefix <- defaultCabalDir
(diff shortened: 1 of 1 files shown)


23Skidoo:
Any idea what's up with the failing Windows test case? Otherwise LGTM.

ttuegel:
> Any idea what's up with the failing Windows test case?

No idea at all. Now it seems to be passing?

23Skidoo:
OK, let's merge then!

PR #4857, opened 2017-11-01
Add s/cmm-sources + extra-bundled-libraries

angerman:

So this is primarily, so that I can give the `rts` package `rts.cabal` file.
But I believe these have uses outside of the rts as well.

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1029,10 +1029,10 @@ checkPaths pkg =
       ++ [ (path, "data-dir")        | path <- [dataDir      pkg]]
       ++ [ (path, "license-file")    | path <- licenseFiles  pkg ]
       ++ concat
-         [    [ (path, "c-sources")        | path <- cSources        bi ]
-           ++ [ (path, "cxx-sources")      | path <- cxxSources      bi ]
-           ++ [ (path, "s-sources")        | path <- sSources        bi ]
+         [    [ (path, "asm-sources")      | path <- asmSources      bi ]
            ++ [ (path, "cmm-sources")      | path <- cmmSources      bi ]
+           ++ [ (path, "c-sources")        | path <- cSources        bi ]
+           ++ [ (path, "cxx-sources")      | path <- cxxSources      bi ]
            ++ [ (path, "js-sources")       | path <- jsSources       bi ]
            ++ [ (path, "install-includes") | path <- installIncludes bi ]
            ++ [ (path, "hs-source-dirs")   | path <- hsSourceDirs    bi ]
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1031,6 +1031,8 @@ checkPaths pkg =
       ++ concat
          [    [ (path, "c-sources")        | path <- cSources        bi ]
            ++ [ (path, "cxx-sources")      | path <- cxxSources      bi ]
+           ++ [ (path, "s-sources")        | path <- sSources        bi ]
+           ++ [ (path, "cmm-sources")      | path <- cmmSources      bi ]
            ++ [ (path, "js-sources")       | path <- jsSources       bi ]
            ++ [ (path, "install-includes") | path <- installIncludes bi ]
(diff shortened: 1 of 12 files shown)


Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1031,6 +1031,8 @@ checkPaths pkg =
       ++ concat
          [    [ (path, "c-sources")        | path <- cSources        bi ]
            ++ [ (path, "cxx-sources")      | path <- cxxSources      bi ]
+           ++ [ (path, "s-sources")        | path <- sSources        bi ]
```
23Skidoo: IMO `asm-sources` would be a better name. One question is whether we should have `asm-x86-sources`, `asm-x86_64-sources`, etc. instead of a single catch-all field... probably not.

Cabal/Distribution/Simple/BuildPaths.hs
```diff
@@ -183,21 +185,27 @@ flibBuildDir lbi flib = buildDir lbi </> nm </> nm ++ "-tmp"
...
-mkLibName lib = "lib" ++ getHSLibraryName lib <.> "a"
+mkLibName lib = mkGenericStaticLibName (getHSLibraryName lib)
 
--- TODO: Should this use staticLibExtension?
 mkProfLibName :: UnitId -> String
-mkProfLibName lib =  "lib" ++ getHSLibraryName lib ++ "_p" <.> "a"
+mkProfLibName lib =  mkGenericStaticLibName (getHSLibraryName lib ++ "_p")
+
+
+mkGenericSharedLibName :: CompilerId -> String -> String
```
23Skidoo: Please add a Haddock comment.

Cabal/Distribution/PackageDescription/Parse.hs
```diff
@@ -493,6 +493,9 @@ binfoFieldDescrs =
  , listFieldWithSep vcat "extra-ghci-libraries"
            showToken          parseTokenQ
            extraGHCiLibs      (\xs    binfo -> binfo{extraGHCiLibs=xs})
+ , listFieldWithSep vcat "extra-bundled-libraries"
```
23Skidoo: Needs docs.

Cabal/Distribution/PackageDescription/Check.hs
angerman: I'd argue for using `if arch(...):` in those cases.

I'm fine with `asm-soures` as well.

Cabal/Distribution/PackageDescription/Parse.hs
angerman: Will see to this and the haddocks tomorrow. Thanks for the review!

Cabal/Distribution/PackageDescription/Check.hs
23Skidoo: Yep, I also think that `if(arch)` is the right approach.

hvr: especially given @phadej's `elif` using `if arch(...)` is quite ok I think

hvr:
This ciritical/urgent, but in general it'd be sensible to have things like `{c,cmm,asm}-sources` imply a constraint on the compiler implementation. E.g. specifying `c-sources` would imply a compiler that supports linking C FFI functions; currently afaik, GHCJS doesn't support C FFI as an example.

angerman:
```
Building test suite 'check-tests' for Cabal-2.1.0.0..
[1 of 1] Compiling Main             ( tests\CheckTests.hs, dist\build\check-tests\check-tests-tmp\Main.o )
Linking dist\build\check-tests\check-tests.exe ...
C:/ProgramData/chocolatey/lib/ghc/tools/ghc-8.0.2/mingw/bin/ld.exe: cannot find -lHSCabal-2.1.0.0-FQui7antlcKGtClAj2Mabp
collect2.exe: error: ld returned 1 exit status
`gcc.exe' failed in phase `Linker'. (Exit code: 1)
```

Hmm... so how did I cause that. let's see.

angerman:
I'm pretty annoyed that travis isn't checking this :(

23Skidoo:
Merged, thanks!

PR #4861, opened 2017-11-03
Allow new-run to run tests and benchmarks too

### LINKED ISSUE #4684: Allow new-run test-suites and benchmarks
E.g.

```
cabal new-run benchmarks -o bench.html
cabal new-run specs SpecificSpec
```

IMHO way simpler then working with `--test-options` or `--bench-options` (which `new-test` doesn't accept. does it)

fgaz:
This is especially useful for passing arguments to a single test suite

* [x] Update the error reporting
* [x] Remove duplication
* [x] Some refactoring on that ugly lambda
* [x] Add some comments about why we allow running tests/benchs
* [ ] Check and update the tests
  * [x] Fix the negative test
  * [ ] Add a positive test
* [x] rebase

ping @merijn this is already usable

Closes #4684

### DIFF
--- a/cabal-install/Distribution/Client/CmdRun.hs
+++ b/cabal-install/Distribution/Client/CmdRun.hs
@@ -29,7 +29,7 @@ import Distribution.Simple.Setup
 import Distribution.Simple.Command
          ( CommandUI(..), usageAlternatives )
 import Distribution.Types.ComponentName
-         ( componentNameString )
+         ( showComponentName )
 import Distribution.Text
          ( display )
 import Distribution.Verbosity
@@ -400,12 +400,13 @@ renderTargetProblem (TargetProblemNoTargets targetSelector) =
 renderTargetProblem (TargetProblemMatchesMultiple targetSelector targets) =
     "The run command is for running a single executable at once. The target '"
  ++ showTargetSelector targetSelector ++ "' refers to "
- ++ renderTargetSelector targetSelector ++ " which includes the executables "
- ++ renderListCommaAnd
-      [ display name
-      | cname@CExeName{} <- map availableTargetComponentName targets
-      , let Just name = componentNameString cname
-      ]
+ ++ renderTargetSelector targetSelector ++ " which includes "
+ ++ renderListCommaAnd ( ("the "++) <$>
+                         showComponentName <$>
+                         availableTargetComponentName <$>
+                         foldMap
+                           (\kind -> filterTargetsKind kind targets)
+                           [ExeKind, TestKind, BenchKind] )
  ++ "."
 
 renderTargetProblem (TargetProblemMultipleTargets selectorMap) =
--- a/cabal-install/Distribution/Client/CmdRun.hs
+++ b/cabal-install/Distribution/Client/CmdRun.hs
@@ -70,10 +70,10 @@ runCommand = Client.installCommand {
   commandDescription  = Just $ \pname -> wrapText $
         "Runs the specified executable, first ensuring it is up to date.\n\n"
 
-     ++ "Any executable in any package in the project can be specified. "
-     ++ "A package can be specified if contains just one executable. "
-     ++ "The default is to use the package in the current directory if it "
-     ++ "contains just one executable.\n\n"
+     ++ "Any executable/test/benchmark in any package in the project can be "
+     ++ "specified. A package can be specified if contains just one "
+     ++ "executable. The default is to use the package in the current "
+     ++ "directory if it contains just one executable.\n\n"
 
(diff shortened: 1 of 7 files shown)


cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -309,11 +311,11 @@ selectPackageTargets targetSelector targets
   where
     (targetsExesBuildable,
      targetsExesBuildable') = selectBuildableTargets'
-                            . filterTargetsKind ExeKind
+                            . (\ts -> concatMap (`filterTargetsKind` ts) [ExeKind, TestKind, BenchKind])
```
phadej: `ExeKind` was first argument, now it's (`ts`) is second. Is `filterTragetsKind

fgaz: What do you mean? The behavior is the same (note the backticks).

Uuh anyway this whole patch is not in a good shape right now. No need to review it until I polish it

phadej: Ah I got confused. Maybe `(\kind -> filterTargetsKind kind ts)` would be less mind-twisting, or  adding `filterTargetKinds :: [ComponentKind] -> ...`

fgaz:
This is now ready for review.

The failing test is obsolete, I'm updating it right now.

23Skidoo:
CI failures look legit.

fgaz:
Yeah, I probably have to update/delete that test too since I changed new-run's behavior

23Skidoo:
CI is still red, there's a failing test.

fgaz:
With this set of changes that test _should_ have that result. I tried to change it to check the new result, but there is little/no documentation about IntegrationTests2. Maybe I should just delete it

phadej:
travis fails because we need for #4962 to get in.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -326,11 +328,13 @@ selectPackageTargets targetSelector targets
 selectComponentTarget :: PackageId -> ComponentName -> SubComponentTarget
                       -> AvailableTarget k -> Either TargetProblem  k
 selectComponentTarget pkgid cname subtarget@WholeComponent t
-  | CExeName _ <- availableTargetComponentName t
-  = either (Left . TargetProblemCommon) return $
-           selectComponentTargetBasic pkgid cname subtarget t
-  | otherwise
-  = Left (TargetProblemComponentNotExe pkgid cname)
+  = case availableTargetComponentName t
```
phadej: amend the haddock of `selectComponentTarget`.

We should be precise here when we mean "executable" and when "executable-like" component. This will cause confusion otherwise.

cabal-install/Distribution/Client/CmdRun.hs
```diff
@@ -70,10 +70,10 @@ runCommand = Client.installCommand {
   commandDescription  = Just $ \pname -> wrapText $
         "Runs the specified executable, first ensuring it is up to date.\n\n"
 
-     ++ "Any executable in any package in the project can be specified. "
-     ++ "A package can be specified if contains just one executable. "
-     ++ "The default is to use the package in the current directory if it "
-     ++ "contains just one executable.\n\n"
+     ++ "Any executable/test/benchmark in any package in the project can be "
+     ++ "specified. A package can be specified if contains just one "
+     ++ "executable. The default is to use the package in the current "
```
phadej: executable-like component

cabal-install/tests/IntegrationTests2.hs
```diff
@@ -869,15 +869,6 @@ testTargetProblemsRun config reportSubCase = do
       [ ( CmdRun.TargetProblemNoTargets, mkTargetPackage "p-0.1" )
       ]
 
-    reportSubCase "test-only"
```
phadej: I'm not super happy of simply removing this test. 

It's invalid with this change, as https://github.com/haskell/cabal/blob/fcabd87406effc0c265312eb38509804682e4623/cabal-install/tests/IntegrationTests2/targets/test-only/p.cabal has only `test-suite` stanza, but now *has an executable*.

Let's change that package to be `lib-only` (you cannot run that).

What is pity, we don't have any positive tests for `CmdRun` in `IntegrationsTests2.hs`. We really should add them, so the things working now continue working in the future.

fgaz:
I'm merging this once it's green, since the 15th is near.

I promise to do that positive test sometime after my next exam (which is also too near)

fgaz:
Oh well, windows and linux is green and one of the four osx builds is green too (the other 3 keep erroring). I'm merging this.

phadej:
:fireworks:

PR #4866, opened 2017-11-07
Look for install includes in buildpref as well

angerman:
When installing a lirbary, we should look for it's headers in it's 'build' preference first, then relative to the current path.

### DIFF
--- a/Cabal/changelog
+++ b/Cabal/changelog
@@ -1,6 +1,8 @@
 -*-change-log-*-
 
 2.2.0.0 (current development version)
+	* `copyCompoment` and `installIncludeFiles` will look for include
+	  headers in the build preference as well.
 	* Added cxx-options and cxx-sources build info fields for seperate
 	  compilation of C++ source files (#3700)
 	* Remove unused '--allow-newer'/'--allow-older' support (#4527)
--- a/Cabal/changelog
+++ b/Cabal/changelog
@@ -1,8 +1,9 @@
 -*-change-log-*-
 
 2.2.0.0 (current development version)
-	* `copyCompoment` and `installIncludeFiles` will look for include
-	  headers in the build preference as well.
+	* `copyCompomenti` and `installIncludeFiles` will look for include
+	  headers in the build preference dir ('dist/build/...' by default)
+	  as well.
 	* Added cxx-options and cxx-sources build info fields for seperate
 	  compilation of C++ source files (#3700)
 	* Remove unused '--allow-newer'/'--allow-older' support (#4527)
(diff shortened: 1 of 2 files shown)


angerman:
Alright, so ideally we'd look into the relative directories from the cabal file. This will work when the CWD is the folder containing the cabal file. However if it is not, and the generic package description in passed in, we have no handle on the location of the cabal file. Or do we?

However, Looking for the include headers in the build directory, prior to trying to locate them relative to the current path seems like a sensible change to me.

Cabal/changelog
```diff
@@ -1,6 +1,8 @@
 -*-change-log-*-
 
 2.2.0.0 (current development version)
+	* `copyCompoment` and `installIncludeFiles` will look for include
+	  headers in the build preference as well.
```
23Skidoo: Should say what the default build preference dir is, i.e.:
```
... will look for include headers in the build preference dir ('dist/build/...' by default) as well.
```

23Skidoo:
> When installing a lirbary, we should look for it's headers in it's 'build' preference first, then relative to the current path.

That's because custom setup scripts can put stuff there? OK, makes sense.

/cc @dcoutts

hvr:
Can you explain the motivation and/or the problem this solving?

angerman:
>>When installing a lirbary, we should look for it's headers in it's 'build' preference first, then relative to the current path.
> That's because custom setup scripts can put stuff there? OK, makes sense.
Yes, because you might be generating those headers.  In which case you might not want to pollute your source tree with generated data. However that is a bit harder, and would likely require #4867 to be solved first.

angerman:
> Can you explain the motivation and/or the problem this solving?

Alright, so you end up "generating" your headers. And they end up in the build directory, however your cabal file does addresses them relative.

angerman:
To clarify why I called this "build preference". This is due wo the variable naming of `buildPref` in the source.

23Skidoo:
@angerman To the users it's either `dist/build/...` or `--builddir`.

hvr:
Oh I see; one thing though, for `*-modules` we have `autogen-modules`, in order to be explicit about what is autogenerated and which things cabal is supposed to warn us about.

I think we should have the corresponding facility also for installed headers, so we can have cabal warn at `sdist` time properly, when you specify a header that doesn't exist yet (because it will be autogenerated)

angerman:
@23Skidoo I'll update the change log to reflect that better.

angerman:
To better understand where I'm coming from. I'm currently trying to roll `ghc-cabal` into hadrian. And all these issues crop up while doing so.

23Skidoo:
Merged, thanks!

PR #4870, opened 2017-11-07
 cabal-install: Remove support for building with GHC < 7.10.

23Skidoo:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,13 +31,14 @@ matrix:
      sudo: required
    # These don't have -dyn/-prof whitelisted yet, so we have to
    # do the old-style installation
+   # NB: TEST_OTHER_VERSIONS doesn't work with USE_GOLD=YES.
    - env: GHCVER=7.4.2 SCRIPT=script CABAL_LIB_ONLY=YES TEST_OTHER_VERSIONS=YES
      os: linux
      sudo: required
    - env: GHCVER=7.6.3 SCRIPT=script CABAL_LIB_ONLY=YES TEST_OTHER_VERSIONS=YES
      os: linux
      sudo: required
-   - env: GHCVER=7.8.4 SCRIPT=script CABAL_LIB_ONLY=YES TEST_OTHER_VERSIONS=YES USE_GOLD=YES
+   - env: GHCVER=7.8.4 SCRIPT=script CABAL_LIB_ONLY=YES TEST_OTHER_VERSIONS=YES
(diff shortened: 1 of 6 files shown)


23Skidoo:
Fun fact: we can also drop support for building with GHC 7.4.2 from `lib:Cabal` as well, since it was released > 5 years ago (Jun 2012).

phadej:
hmm. dropping 7.4.2 sounds cool, as then we can `containers >=0.5`. Is there other good reasons? I'd rather remove the support where there's good reason to do so.

23Skidoo:
Any change that removes code is a good thing in my book, 7.4.2 is positively ancient, and we won't be breaking any promises. Note that we're still supporting `build -w` for old GHC versions back to 7.0 and there are CI tests for that.

Another fun fact: GHC 7.6.3 was released in April 2013, so we can remove support for building lib:Cabal with 7.6.* in the next+1 (i.e. 2.4) release. 7.8 will have to be supported until Dec 2019.

hvr:
> hmm. dropping 7.4.2 sounds cool, as then we can containers >=0.5.

Otoh, you can still support GHC 7.4.2 while requiring containers >= 0.5 ... if you don't insist on `--constraints "containers installed"` :-)

23Skidoo:
Yep, we probably can support building lib:Cabal with old GHC:s as a setup script dependency for a longer period. This needs to be checked by CI, though...

23Skidoo:
There's some weird 7.8.4 failure, will investigate.

23Skidoo:
Yiss, looks like support for `ld.gold` was only added in 7.8: [1](https://ghc.haskell.org/trac/ghc/ticket/4862), [2](https://github.com/ghc/ghc/commit/668151c880f5662e59288a0884d8d3505c0ce7d7).

23Skidoo:
So `PackageTests/TestSuiteTests/ExeV10/setup-with-hpc.multitest.hs` is now failing in the 7.8.4+7.0.4 configuration. No idea how it can pass with 7.6.3+7.0.4 at the same time.

23Skidoo:
Looks like that test doesn't work with `hpc < 0.7`, but determines the version of `hpc` by just looking in `PATH` and not at the version the compiler we're configured with uses. And 7.6.3 just happens to use `hpc` 0.6, so the test passes.

23Skidoo:
Yay, it's green at last.

PR #4874, opened 2017-11-08
basedir

angerman:


Cabal/Distribution/Simple/Setup.hs
```diff
@@ -332,6 +332,12 @@ data ConfigFlags = ConfigFlags {
 
     configDistPref :: Flag FilePath, -- ^"dist" prefix
     configCabalFilePath :: Flag FilePath, -- ^ Cabal file to use
+    configBaseDir       :: Flag FilePath, -- ^ The directory containing the
```
23Skidoo: So now  `takeBaseName configCabalFilePath == configBaseDir` doesn't need to hold? I still don't quite understand why you need a new option and can't use `takeBaseName $ configCabalFilePath`.

angerman: You are not bound to have the `configCabalFilePath`. E.g. given the following invocation:
```
defaultMainWithHooksNoReadArgs hooks gpd ["--basedir", "/path/to/happiness"]
```
you do not have the `--cabalConfigFilePath`.

And I'm not sure we want to force providing the `--cabalConfigFilePath`? If we can agree that we want to require the `--cabalConfigFilePath` in those cases. I'm happy to change to `takeBaseName`.

23Skidoo: > You are not bound to have the `configCabalFilePath`.

But this isn't different from the situation when `configBaseDir` is `NoFlag`. Your example can be written as `defaultMainWithHooksNoReadArgs hooks gpd ["--cabal-file", "/path/to/happiness/happiness.cabal"]` with `--cabal-file`.

angerman: You are right. the more verbose solution, might be the less error prone. Let's do it that way.

Cabal/Distribution/Simple.hs
```diff
@@ -705,8 +743,9 @@ runConfigureScript verbosity backwardsCompatHack flags lbi = do
   shConfiguredProg <- lookupProgram shProg
                       `fmap` configureProgram  verbosity shProg progDb
   case shConfiguredProg of
-      Just sh -> runProgramInvocation verbosity
+      Just sh -> runProgramInvocation verbosity $
                  (programInvocation (sh {programOverrideEnv = overEnv}) args')
+                 { progInvokeCwd = Just (baseDir lbi) }
```
angerman: I believe this change breaks on windows. I'm not sure exactly why yet.

angerman:
@hvr, @23Skidoo why did Jenkins not test this one? How can we *force* a check? Can we have some post to the PR trigger (e.g. `[rerun Travis]`) trigger a rerun of the CI?

23Skidoo:
@angerman idk, it does this sometimes. Maybe connectivity issues between GH and Travis. Try making a small tweak (so that commit hash gets changed) and force-pushing.

23Skidoo:
@angerman Also, if you have permissions, you can go to Settings -> Webhooks and click "Re-deliver" on the offending one. I see quite a lot of failed Travis calls there.

angerman:
Let's see if this does the trick.

angerman:
@23Skidoo, @hvr, while this works for what I'm doing to the GHC build system. I believe there are still some code path's left. Any creative idea how to test this properly and get the sufficient coverage of the cabal codebase?

angerman:
Note to self: need to figure out why test fail.

angerman:
@23Skidoo, @ezyang this also fails with downstream travis.

I'd really like to get this merged. As it's a prerequisite for the relocatable hadrian (building ghc with shake) snowleopard/hadrian#445 PR, and I'm trying to tie up loose ends here.

phadej:
I check locally, if I can reproduce failures

angerman:
That downstream-travis  failure for linux-7.6.3 does not seem legit!
```
+sudo apt-get install --force-yes cabal-install-head cabal-install-2.0 happy-1.19.5 alex-3.1.7 ghc-7.6.3-prof ghc-7.6.3-dyn
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package cabal-install-head
E: Unable to locate package cabal-install-2.0
E: Couldn't find any package by glob 'cabal-install-2.0'
E: Couldn't find any package by regex 'cabal-install-2.0'
E: Unable to locate package happy-1.19.5
E: Couldn't find any package by glob 'happy-1.19.5'
E: Couldn't find any package by regex 'happy-1.19.5'
E: Unable to locate package alex-3.1.7
E: Couldn't find any package by glob 'alex-3.1.7'
E: Couldn't find any package by regex 'alex-3.1.7'
E: Unable to locate package ghc-7.6.3-prof
E: Couldn't find any package by glob 'ghc-7.6.3-prof'
E: Couldn't find any package by regex 'ghc-7.6.3-prof'
E: Unable to locate package ghc-7.6.3-dyn
E: Couldn't find any package by glob 'ghc-7.6.3-dyn'
E: Couldn't find any package by regex 'ghc-7.6.3-dyn'
```

angerman:
@23Skidoo, @phadej can you restart AppVoyer? 

Looks like it’s going to be green ...

angerman:
I'd like to give especially the running of `configure` another though.

angerman:
This produces a rather interesting issue: where do we actually want to generate the `.buildinfo` in?
Shouldn't that end up in the `build` folder? And be specific to the given build?

angerman:
Building GHC with Hadrian (reloc) right now, for verification

Cabal/Distribution/Simple.hs
```diff
@@ -98,7 +98,8 @@ import Distribution.Text
 -- Base
 import System.Environment (getArgs, getProgName)
 import System.Directory   (removeFile, doesFileExist, getCurrentDirectory
-                          ,doesDirectoryExist, removeDirectoryRecursive)
+                          ,doesDirectoryExist, removeDirectoryRecursive
+                          ,makeAbsolute)
```
phadej: cannot use. it's only in `1.2.2.0 (Mar 2015)` https://hackage.haskell.org/package/directory-1.3.2.0/changelog which is GHC-7.10.3...

angerman: Ack

phadej:
```
Distribution\Simple.hs:679:1: warning: [-Wunused-top-binds]
    Defined but not used: `getBaseDir'
``

**Comment by angerman** [comment]
Lots of
```
/Users/travis/build/haskell/cabal/dist-newstyle/src/pretty-show-1.6.16/dist: renamePath:rename: unsatisfied constraints (Directory not empty)Configuring library for pretty-show-1.6.16..
```
failures.

I've no clear idea yet where they come from.

**Comment by angerman** [comment]
What?
```
Registering library for hackage-security-0.5.2.2..
[1 of 1] Compiling Main             ( Setup.hs, Setup.o )
Linking Setup ...
Configuring cabal-install-2.1.0.0...
Setup: Encountered missing dependencies:
network >=2.4 && <2.6, resolv >=0.1.1 && <0.2
```

**Comment by angerman** [comment]
Cleaned up and rebased. Let's get this one green. Will probably still suffer from #5120, but with enough patience and restarts, that should work.

**Comment by angerman** [comment]
massaging travis...

**Comment by 23Skidoo** [comment]
@angerman Thank you for your patience with this one!

**Comment by angerman** [comment]
Turns out there's a legit failure hidden in all the others...
```
Error during cabal-install bootstrap:
Configuring the network package failed.
network-uri-2.5.0.0 will be downloaded and installed.
```

alright, let's see why that one failed...

**Comment by angerman** [comment]
Meh. bootstrap fails, because `networks` ships `include/HsNetworkConfig.h` **and** `include/HsNetworkConfig.h.in`, now that we run `configure` from a different folder (`dist/build`), this results in
```
- dist/build/include/HsNetworkConfig.h
- include/HsNetworkConfig.h
```
and we pick the wrong one.

I've opened haskell/network#307.

**Comment by angerman** [comment]
@phadej, @23Skidoo please review 8b5b2c5

**Comment by angerman** [comment]
@phadej I guess we should, I just don't yet know how to write a proper `cabal check` feature for that, we should probably open a issue for this though.

Looking at haskell/network#307, they currently to figure out how the header file ended up in the uploaded tarball in the first place.

**Comment by angerman** [comment]
I believe something on master started breaking windows :(

**Comment by angerman** [comment]
Rebased onto #5132

**Comment by angerman** [comment]
rebased onto master

**Comment by angerman** [comment]
all green!

**Comment by 23Skidoo** [comment]
I'll merge this later today or tomorrow, there are some small changes I want to make. Sorry for the delay, I've been ill.

**Comment by angerman** [comment]
@23Skidoo thanks!

File: Cabal/Distribution/Simple.hs [hunk 4]
```diff
@@ -648,44 +680,51 @@ autoconfUserHooks
...
-          readHookWithArgs :: (a -> Flag Verbosity) -> Args -> a
+          readHookWithArgs :: (a -> Flag Verbosity)
+                           -> (a -> Flag FilePath)
+                           -> Args -> a
                            -> IO HookedBuildInfo
-          readHookWithArgs get_verbosity _ flags = do
-              getHookedBuildInfo verbosity
+          readHookWithArgs get_verbosity get_dist_pref _ flags = do
+              dist_dir <- findDistPrefOrDefault (get_dist_pref flags)
+              getHookedBuildInfo (dist_dir </> "build") verbosity
```
23Skidoo: I'm not totally keen on the fact that we're hardcoding `"build"` everywhere. This works with `new-build` directory structure, right?

Cabal/Distribution/Simple.hs
```diff
@@ -710,29 +751,30 @@ runConfigureScript verbosity backwardsCompatHack flags lbi = do
...
     notFoundMsg = "The package has a './configure' script. "
                ++ "If you are on Windows, This requires a "
                ++ "Unix compatibility toolchain such as MinGW+MSYS or Cygwin. "
                ++ "If you are not on Windows, ensure that an 'sh' command "
                ++ "is discoverable in your path."
 
-getHookedBuildInfo :: Verbosity -> IO HookedBuildInfo
-getHookedBuildInfo verbosity = do
-  maybe_infoFile <- defaultHookedPackageDesc
+getHookedBuildInfo :: FilePath -> Verbosity -> IO HookedBuildInfo
```
23Skidoo: We should mention in the changelog/migration guide that this function has changed type.

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1536,6 +1536,7 @@ findPackageDesc dir
 tryFindPackageDesc :: FilePath -> IO FilePath
 tryFindPackageDesc dir = either die return =<< findPackageDesc dir
 
+{-# DEPRECATED defaultHookedPackageDesc "Use findHookedPackageDesc with the proper base directory instead" #-}
 -- |Optional auxiliary package information file (/pkgname/@.buildinfo@)
```
23Skidoo: This should also be mentioned in the changelog.

23Skidoo:
Just waiting for it to go green now...

23Skidoo:
Merged, thanks!

Cabal/Distribution/Simple/GHC/Internal.hs
```diff
@@ -276,7 +276,11 @@ componentCcGhcOptions verbosity _implInfo lbi bi clbi odir filename =
       ghcOptCppIncludePath = toNubListR $ [autogenComponentModulesDir lbi clbi
                                           ,autogenPackageModulesDir lbi
                                           ,odir]
-                                          ++ PD.includeDirs bi,
+                                          -- includes relative to the package
+                                          ++ PD.includeDirs bi
+                                          -- potential includes generated by `configure'
+                                          -- in the build directory
+                                          ++ [buildDir lbi </> dir | dir <- PD.includeDirs bi],
```
23Skidoo: Don't you need to filter out absolute paths here as well?

ezyang:
Mac OS X regression caused by this: https://ghc.haskell.org/trac/ghc/ticket/14972

In general, the new files in `build` can cause GHCi to stop working if one of the files/directory unluckily has the same name as a dynamic library (`gmp` in this case), because then `gcc -print-file-name` goes insane (but apparently only on OS X)

angerman:
I've created https://phabricator.haskell.org/D4553, which should ensure that what ever `gcc --print-file-name` returns is verified to be a file prior to passing it on.

PR #4875, opened 2017-11-08
Adds `virtual-modules` to handle cases like `GHC.Prim` from `ghc-prim`.

angerman:
This used to be done in `ghc-cabal`. With the long term goal of getting rid of ghc-cabal eventually, it's functionality needs to end up in cabal.

While this is a custom hack, it is restricted to the GHC module only. A more generic solution would be to add `virtual-modules` I suppose. GHC does custom handling for `GHC.Prim`, and there is no support for `virtual-modules` in the package config file either as far as I can see. As such this seems to be the least invasive solution.

### DIFF
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -375,6 +375,7 @@ buildInfoFieldGrammar = BuildInfo
     <*> monoidalFieldAla "js-sources"           (alaList' VCat FilePathNT)    L.jsSources
     <*> hsSourceDirsGrammar
     <*> monoidalFieldAla "other-modules"        (alaList' VCat MQuoted)       L.otherModules
+    <*> monoidalFieldAla "virtual-modules"      (alaList' VCat MQuoted)       L.virtualModules
     <*> monoidalFieldAla "autogen-modules"      (alaList' VCat MQuoted)       L.autogenModules
     <*> optionalFieldAla "default-language"     MQuoted                       L.defaultLanguage
     <*> monoidalFieldAla "other-languages"      (alaList' FSep MQuoted)       L.otherLanguages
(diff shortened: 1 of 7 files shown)


angerman:
How much documentation would this need?

hvr:
I'm a bit conflicted about this... yet another magic undocumented logic... what if I have a custom package which happens to define a module named `GHC.Prim` for fun...  is "GHC.Prim" now considered to be a reserved module name nobody can use? :-)

23Skidoo:
Hmm, so now I can't make a package that has a module named `GHC.Prim`? Can you perhaps make this hack only apply to the `base` package (and add a "MASSIVE HACK" disclaimer while you're at it)? Though `virtual-modules` wouldn't be that hard to add.

hvr:
Btw, to makes things more interesting, `ghc-prim` has

```
flag include-ghc-prim
    Description: Include GHC.Prim in exposed-modules
    default: False

library
    if flag(include-ghc-prim)
        exposed-modules: GHC.Prim
```

angerman:
I'm not terribly happy with this either. For `virtual-modules`, I'm just not sure how much use that would have outside of this specific case.

I therefore plan to go forward with a guard on the `base` package, and an additional `MASSIVE HACK` annotation.

23Skidoo:
> For `virtual-modules`, I'm just not sure how much use that would have outside of this specific case.

Most likely none, but at least you won't need to patch Cabal again if `base` gets a new virtual module.

angerman:
> Most likely none, but at least you won't need to patch Cabal again if base gets a new virtual module.
I believe `GHC.Prim` to be rather special and unique. Any additional virtual module, would require additional special casing in GHC as well.

angerman:
I'll do some more test, and merge if nothing falls through the cracks.

angerman:
Seems to work for me as expected.

PR #4879, opened 2017-11-09
Disable tests and benchs by default on new-install

### LINKED ISSUE #4855: new-install fails on packages with tests/benchmarks unless --disable-tests/benchmarks is used
For example, `cabal new-install pandoc` fails this assertion:

https://github.com/haskell/cabal/blob/52a7d51e239dce1791b7d1f33590239a6fba0cd5/cabal-install/Distribution/Client/ProjectPlanning.hs#L241-L245

This is easily fixed by setting the tests and benchmarks as disabled in every `ElaboratedConfiguredPackage` (or at an earlier stage) in `CmdInstall.hs`.

fgaz:
And error if they are explicitly enabled, so we never try to install
remote packages with tests and/or benchs.

Fixes #4855

...no tests for new-install yet

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -46,7 +46,7 @@ import Distribution.Client.RebuildMonad
 import Distribution.Client.InstallSymlink
          ( symlinkBinary )
 import Distribution.Simple.Setup
-         ( HaddockFlags, fromFlagOrDefault, flagToMaybe )
+         ( Flag(Flag), HaddockFlags, fromFlagOrDefault, flagToMaybe )
 import Distribution.Simple.Command
          ( CommandUI(..), usageAlternatives )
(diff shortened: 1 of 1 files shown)


fgaz:
It's  the second time that this happens https://travis-ci.org/haskell-pushbot/cabal-binaries/builds/299703001#L655
```
    ShortText Id:                                                                         FAIL

      *** Failed! Falsifiable (after 100 tests and 4 shrinks): 

      "\65534"

      Use --quickcheck-replay=209006 to reproduce.
```

Again the U+FFFE "not a character" (reversed BOM)

23Skidoo:
It's a known issue, I think we had a ticket. You can ignore it. /cc @hvr

23Skidoo:
OK, let's merge. CI errors are spurious.

23Skidoo:
Merged, thanks!

PR #4884, opened 2017-11-12
Solver: Enforce dependencies on executables (fixes #4781).

### LINKED ISSUE #4781: build-tool-depends silently ignored for non-existent executable
Consider the following specification,

```
name:                dummypkg
version:             0
build-type:          Simple
cabal-version:       >=2.0

library
  default-language: Haskell2010
  build-tool-depends: mtl:imaginary-exe
```

when invoking `new-build`, this silently succeeds. And also, `plan.json` doesn't record any `exe-depends` from `dummypkg-0-inplace`  on `mtl`  as it would if `build-tool-depends` referred to an actually existing executable:

```json
    {
      "type": "configured",
      "id": "dummypkg-0-inplace",
      "pkg-name": "dummypkg",
      "pkg-version": "0",
      "flags": {},
      "style": "local",
      "dist-dir": ".../dist-newstyle/build/x86_64-linux/ghc-8.2.1/dummypkg-0",
      "depends": [],
      "exe-depends": [],
      "component-name": "lib"
    },

```

For a `build-tool-depends` declaration, I'd expect the solver to require the existence  of that executable in the respective package as a constraint to solve for. The reason being is that packages may start out as a library, and in later releases start adding an executable. Or sometimes executables may get renamed or removed.

/cc @grayjay @Ericson2314

grayjay:

__________________________________________________________________________________________________

Main commits:

#### Solver: Use more specific failure reasons for dependency conflicts.

This commit splits the `Conflicting` constructor of `FailReason` into five
different constructors, for different types of dependency conflicts.  This
change makes it easier to print different error messages for different
conflicts, though it doesn't change any of the error messages yet.

The refactoring caused a small change to the error message for conflicts caused
by self-dependencies.

#### Solver: Store names of required executables for build-tool-depends dependencies.

This commit changes the field of type `IsExe` in the `Dep` data type to
type `Maybe UnqualComponentName`.  It also adds the executable name to error
messages that previously just contained "(exe)".

#### Solver: Enforce dependencies on executables (fixes #4781).

This commit adds two checks to the validation phase of the solver:

1. It checks that each newly chosen package instance contains all executables
   that are required from that package so far.

2. It checks that each new build tool dependency that refers to a previously
   chosen package can be satisfied by the executables in that package.

This commit also fixes a TODO related to solver log messages. Previously, it was
possible for the log to associate an incorrect executable name with a
dependency.

#### Test that the solver ignores build tool dependencies in legacy mode.

The pre-new-build code path sets the solver parameter `solveExecutables` to
false, which causes the solver to ignore all build-tools and build-tool-depends
dependencies.  This commit adds a test for ignoring each of the two types of
dependencies.

_______________________________________________________________________________________

I compared the performance on this branch (cbf73acd5da16eb4f45289121607881b2c9e0bfa) with master (26426fcc1f78856c9eb95bc778cc16c5d90891d2) to see the effect of the extra validation steps.  I wrote a cabal file with one library build-depends field to test the solver with new-build.  (Only new-build handles build tool dependencies.)  Then I ran `cabal new-build --dry-run --index-state=2017-11-04T18:36:30Z --no-count-conflicts`, with two sets of dependencies in the build-depends field.  I tried to choose one set of dependencies that caused a long run time and another that created a deep search tree.  I turned off conflict counting to prevent the solver from finishing too quickly in the first case.  I also compared the -v3 log from master and the branch to ensure that the change didn't have any effect on conflicts, since I only wanted to measure the overhead of checking for executables.  Each measurement is the average of three trials.

packages              | maximum search tree depth | run time on master | run time on branch | total memory on master, with "+RTS -s" | total memory on branch, with "+RTS -s" 
------------------------|-----------------------------------------|--------------------------|-----------------------------|---------------------------------------------------------|------------------------------------------------------
AesonBson           | 66     | 18.42 s | 18.40 s | 244 MB | 244 MB
gi-glib, leksah, yi | 2349 | 7.10 s   | 7.08 s   | 377 MB | 378 MB

It looks like this change didn't have much effect in either case.

/cc @kosmikus


23Skidoo:
This all sounds absolutely great, but can you please add a short note to the changelog?

grayjay:
Thanks.  I updated the changelog and added a few more tests.

kosmikus:
Looks fantastic. Thanks a lot @grayjay . @23Skidoo Please merge.

23Skidoo:
Merged. Thanks, @grayjay!

grayjay:
Thanks for the review!

PR #4886, opened 2017-11-12
Replace flag-defaulting hack by proper solution

hvr:
This is not finished yet; this is supposed to address issues like #4737 or #5139

/cc @alexbiehl

----

Regressions:

- [x] (note: this is specific to the `all` target)
    ```
    $ cabal new-haddock all
    Resolving dependencies...
    fromFlag NoFlag. Use fromFlagOrDefault
      error, called at ./Distribution/Simple/Setup.hs:172:21 in Cabal-2.1.0.0-inplace:Distribution.Simple.Setup
      fromFlag, called at ./Distribution/Client/CmdHaddock.hs:156:32 in cabal-install-2.1.0.0-inplace:Distribution.Client.CmdHaddock
    ```

### DIFF
--- a/cabal-install/Distribution/Client/CmdHaddock.hs
+++ b/cabal-install/Distribution/Client/CmdHaddock.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE ViewPatterns   #-}
 
 -- | cabal-install CLI command: haddock
 --
@@ -18,8 +17,7 @@ import Distribution.Client.ProjectOrchestration
 import Distribution.Client.CmdErrorMessages
 
 import Distribution.Client.Setup
-         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags
-         , applyFlagDefaults )
+         ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags )
 import qualified Distribution.Client.Setup as Client
 import Distribution.Simple.Setup
          ( HaddockFlags(..), fromFlagOrDefault, fromFlag )
@@ -73,7 +71,7 @@ haddockCommand = Client.installCommand {
 --
 haddockAction :: (ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags)
                  -> [String] -> GlobalFlags -> IO ()
-haddockAction (applyFlagDefaults -> (configFlags, configExFlags, installFlags, haddockFlags))
+haddockAction (configFlags, configExFlags, installFlags, haddockFlags)
                 targetStrings globalFlags = do
 
     baseCtx <- establishProjectBaseContext verbosity cliConfig
--- a/cabal-install/Distribution/Client/CmdHaddock.hs
+++ b/cabal-install/Distribution/Client/CmdHaddock.hs
@@ -20,7 +20,7 @@ import Distribution.Client.Setup
          ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags )
 import qualified Distribution.Client.Setup as Client
 import Distribution.Simple.Setup
-         ( HaddockFlags(..), fromFlagOrDefault, fromFlag )
+         ( HaddockFlags(..), fromFlagOrDefault )
 import Distribution.Simple.Command
          ( CommandUI(..), usageAlternatives )
 import Distribution.Verbosity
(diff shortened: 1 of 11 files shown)


fgaz:
Note to self: this may affect #4861

hvr:
That CI passes is quite deceptive, as with the PR as it stands, `cabal new-haddock` runs into a flag-related bottom rightaway...

fgaz:
Can we remove that LANGUAGE ViewPatterns now or is it needed elsewhere?

hvr:
@fgaz which one? my patch removes already a couple of those...

fgaz:
@hvr I looked at CmdInstall and assumed you leaved it on the other modules too, but apparently I picked the exception, not the rule

fgaz:
Oh you reverted it

23Skidoo:
@hvr What's the status of this?

hvr:
@23Skidoo I've just rebased this, and will try to use it locally and try to observe regressions; but generally I think we want to merge this, and fix `fromFlag`-related exceptions as they get observed/reported.

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -1627,7 +1620,10 @@ installCommand = CommandUI {
      ++ "  " ++ (map (const ' ') pname)
                       ++ "                         "
      ++ "    Change installation destination\n",
-  commandDefaultFlags = (mempty, mempty, mempty, mempty),
+  commandDefaultFlags = (commandDefaultFlags configureCommand,
```
hvr: This is possibly wrong

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -1118,7 +1108,10 @@ upgradeCommand = configureCommand {
     commandSynopsis     = "(command disabled, use install instead)",
     commandDescription  = Nothing,
     commandUsage        = usageFlagsOrPackages "upgrade",
-    commandDefaultFlags = (mempty, mempty, mempty, mempty),
+    commandDefaultFlags = (commandDefaultFlags configureCommand,
```
hvr: This is possibly wrong

hvr: nevermind, the commits were in the wrong order

hvr:
After consulting w/ @alexbiehl it appears like we want (c.f. #4737) something for the haddock flags in the style of

```hs
resolveSolverSettings :: ProjectConfig -> SolverSettings

resolveBuildTimeSettings :: Verbosity -> CabalDirLayout -> ProjectConfig -> BuildTimeSettings
```

specifically,

```hs
resolveSolverSettings :: _ -> ProjectConfig -> HaddockSettings
```

(where `_` is yet to be determined; could be empty) ...and a new type `HaddockSettings` to hold all/most `haddock*` fields from `ProjectConfig`.

cabal-install/Distribution/Client/CmdHaddock.hs
```diff
@@ -151,10 +151,10 @@ selectPackageTargets haddockFlags targetSelector targets
     isRequested (TargetAllPackages (Just _)) _ = True
     isRequested _ LibKind    = True
 --  isRequested _ SubLibKind = True --TODO: what about sublibs?
-    isRequested _ FLibKind   = fromFlag (haddockForeignLibs haddockFlags)
-    isRequested _ ExeKind    = fromFlag (haddockExecutables haddockFlags)
-    isRequested _ TestKind   = fromFlag (haddockTestSuites  haddockFlags)
-    isRequested _ BenchKind  = fromFlag (haddockBenchmarks  haddockFlags)
+    isRequested _ FLibKind   = fromFlagOrDefault False (haddockForeignLibs haddockFlags)
```
hvr: in my local version I had a comment along the lines of

```hs
    -- TODO/HACK, we encode some defaults here as new-haddock's logic;
    -- make sure this matches the defaults applied in
    -- "Distribution.Client.ProjectPlanning"; this may need more work
    -- to be done properly
    --
    -- See also https://github.com/haskell/cabal/pull/4886
```

which may need to be extended even further to point out that this gives CLI flags unduly importance over per-package `cabal.project` settings; which is the thing we'd need to do properly in future work.

alexbiehl: Will add the comment

hvr:
I've merged this as this enables other work to be unblocked (such as #4840)

PR #4893, opened 2017-11-15
Add `checkVersion` for added fields

angerman:
As pointed out by @hvr, #4857, #4875 did not contain the necessary "check" logic. This PR tries to address this shortcoming.

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1186,6 +1186,23 @@ checkCabalVersion pkg =
            [ display (Dependency name (eliminateMajorBoundSyntax versionRange))
            | Dependency name versionRange <- depsUsingMajorBoundSyntax ]
 
+  , checkVersion [2,1] (any (not . null)
+                        (concatMap buildInfoField
+                         [ asmSources
+                         , cmmSources
+                         , extraBundledLibs
+                         , extraLibFlavours ])) $
+      PackageDistInexcusable $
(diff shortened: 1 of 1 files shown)


23Skidoo:
That's what happens when we don't write tests for new features... Thanks to @hvr for spotting this.

23Skidoo:
Merged, thanks!

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1186,6 +1186,23 @@ checkCabalVersion pkg =
...
 
+  , checkVersion [2,1] (any (not . null)
+                        (concatMap buildInfoField
+                         [ asmSources
+                         , cmmSources
+                         , extraBundledLibs
+                         , extraLibFlavours ])) $
+      PackageDistInexcusable $
+           "The use of 'asm-sources', 'cmm-sources', 'extra-bundled-libraries' "
+        ++ " and 'virtual-modules' requires the package "
```
23Skidoo: Oops, looks like you need to `s/virtual-modules/extra-library-flavours/` here. Feel free to commit directly to `master`.

angerman: wow. How the hell did that happen? Ok... I'll just edit it on GH.

angerman:
So this PR supposedly broke Travis. That has two implications:
a) that Travis seems to not react to quite a lot of PRs is pretty bad.
b) It broke on Travis due to: `The job exceeded the maximum time limit for jobs, and has been terminated.`, which isn't that helpful either.

PR #4896, opened 2017-11-17
Add binary distribution download link to Readme

angerman:
I've so often failed to find that link. Let's just add it to the readme.

[skip ci]


23Skidoo:
Merged, thanks!

PR #4897, opened 2017-11-19
Change order of hsLibraries

angerman:
After using this for a while. It occured to me that due to linker ordering, and GHC just forwarding the libraries as specified in the `hsLibraries` field, we should put the `extraBundledLibs` at the end. As such the haskell library can reference the `extraBundledLibs`.

This still leave the question of recursively linked libs. But getting that linked right should rather be deal with in GHC, which optimally would use being and end groups for the libraries to link.


23Skidoo:
Merged, thanks!

Issue #4899, opened 2017-11-20
Forward compat scheme


hvr:
Generally, a given `cabal-install` release is only expected to understand (cabal-)spec(ification)-versions that were known at the time of release.

The Cabal-1.12 release, and more recently the Cabal-2.0 release have made it apparent that we don't have a good forward-compatibility story in place, thereby limiting our ability to evolve the `.cabal` format. Concretely, `cabal-install-1.10` is not expected to understand package meta-data declared using the  `cabal-version:1.12` spec(ification)-version; also, upon encountering such a future `.cabal` file in the package index, `cabal-install-1.10` ought to gracefully recover (instead of failing fatally to parse the package index). A similar situation occurred with the Cabal-2.0 release,  which also introduced new syntax which previous parsers would fail to recognise, again resulting in fatal parsing failures.

To prevent this from happening again, the following forward compatibility has been devised, which is intended to make it possible to evolve the lexical and syntactical structure of the `.cabal` format in future, while not breaking legacy clients.

----

## New-style Cabal-Specification Version Declaration

A new-style spec-version declaration at the beginning of a the `.cabal` file (without any preceding whitespace) and follow the following case-insensitive grammar (expressed in [RFC5234 ABNF](https://tools.ietf.org/html/rfc5234)):

```
newstyle-spec-version-decl = "cabal-version" *WS ":" *WS newstyle-spec-version *WS

newstyle-spec-version      = NUM "." NUM [ "." NUM ]

NUM    = DIGIT0 / DIGITP 1*DIGIT0
DIGIT0 = %x30-39
DIGITP = %x31-39
WS = %20
```

The use of a new-style spec-version is 

 - **invalid** before spec-version 1.12,
 - **optional** starting with spec-version 1.12,
 - **recommended** starting with spec-version 2.0, and
 - **mandatory** starting with spec-version 2.1.

It's also assumed that the following invariant holds:

 - If a new-style spec-version declaration is present, its value must match the spec-version as determined by a full `.cabal` parser.

## Scanning the Cabal-Specification Version

 1. If a `.cabal` file contains a valid new-style spec-version declaration, it is authoritative;
 2. otherwise, the spec-version must be below 2.1 (and either a full `.cabal` parser or a heuristic `cabal-version`-scanner (tbd) must be used to determine the exact spec-version).

The new-style spec-version declaration is designed to be simple to parse by means of common string operations. A simple implementation is shown below

```hs
scanSpecVersion :: ByteString -> Maybe Version
[...]

PR #4900, opened 2017-11-20
Implement preliminary support for new forward-compat scheme

hvr:
This provides a provisional (i.e. hacky) retrofitted implementation of
the forward-compat scheme described in #4899 for the cabal-2.0 branch.

This hack works by constructing a dummy package description in case
the package description fails to be parsed via the standard parser,
and we detect a new-style cabal-spec declaration.


23Skidoo:
@hvr Should I wait for you to finish the write-up in #4900 before merging?

23Skidoo:
Merged, thanks!

Issue #4798, opened 2017-09-27
Program coverage crashes in cabal new-test


jiaochangyang:
```
Preprocessing test suite 'testsuite' for simple-server-0.1.0.0..
Building test suite 'testsuite' for simple-server-0.1.0.0..
Running 1 test suites...
Test suite testsuite: RUNNING...
Tests
  core unit tests
    testing storeInsert empty:                                     OK
    testing storeInsert when already in small:                     OK
    testing storeInsert when already in commit:                    OK
    testing storeInsert when data is in both:                      OK
    testing isSecondaryIndexed when in store:                      OK
    testing isSecondaryIndexed when not in store:                  OK
    testing secondaryLookup when not in secondary:                 OK
    testing secondaryLookup when in small:                         OK
    testing secondaryLookup when in commit:                        OK
    testing secondaryInsert when not in secondary:                 OK
    testing read (ReadAll) when not in store:                      OK
    testing read (lookup) when in small store:                     OK
    testing read (lookup) when in commit store:                    OK
    testing write successfully in small store:                     OK
    testing write unsuccessfully (duplicate) in small store:       OK
    testing write unsuccessfully (duplicate) in commit store:      OK
    testing write failure in store (somehow):                      OK
    testing write failure in secondary (already in store somehow): OK
    testing write failure in store (already in secondary somehow): OK

All 19 tests passed (0.00s)
Test suite testsuite: PASS
Test suite logged to:
/Users/cyj/Documents/simple-server/dist-newstyle/build/x86_64-osx/ghc-8.2.1/simple-server-0.1.0.0/c/testsuite/test/simple-server-0.1.0.0-testsuite.log
hpc: can not find simple-server-0.1.0.0-inplace/Data.API.Read in ./.hpc, /Users/cyj/Documents/simple-server/dist-newstyle/build/x86_64-osx/ghc-8.2.1/simple-server-0.1.0.0/c/testsuite/hpc/vanilla/mix/simple-server-0.1.0.0, /Users/cyj/Documents/simple-server/dist-newstyle/build/x86_64-osx/ghc-8.2.1/simple-server-0.1.0.0/c/testsuite/hpc/vanilla/mix/testsuite
CallStack (from HasCallStack):
  error, called at libraries/hpc/Trace/Hpc/Mix.hs:122:15 in hpc-0.6.0.3:Trace.Hpc.Mix
cabal: Tests failed for test:testsuite from simple-server-0.1.0.0.
```
[...]

Issue #4644, opened 2017-07-29
Unit test Distribution.Utils.ShortText BinaryId fails


fgaz:
https://travis-ci.org/haskell-pushbot/cabal-binaries/builds/258926423
```
$ Cabal/unit-tests
[snip]
  Distribution.Utils.ShortText
    ShortText Id:                                       OK
      +++ OK, passed 100 tests.
    ShortText Ord:                                      OK
      +++ OK, passed 100 tests.
    ShortText Monoid:                                   OK
      +++ OK, passed 100 tests.
    ShortText BinaryId:                                 FAIL
      *** Failed! Falsifiable (after 40 tests and 5 shrinks): 
      "\65534"
      Use --quickcheck-replay=551066 to reproduce.
```

ping @ezyang

ezyang:
CC @hvr who introduced this test in 993d20a2e9b8fb29aefaa2c266f31177a00a5ee6

fgaz:
Wikipedia says `\65534` is [U+FFFE \<noncharacter-FFFE\> not a character.](https://en.wikipedia.org/wiki/Specials_(Unicode_block))
> FFFE and FFFF are not unassigned in the usual sense, but guaranteed not to be a Unicode character at all. They can be used to guess a text's encoding scheme, since any text containing these is by definition not a correctly encoded Unicode text. Unicode's U+FEFF Byte order mark character can be inserted at the beginning of a Unicode text to signal its endianness: a program reading such a text and encountering 0xFFFE would then know that it should switch the byte order for all the following characters.

Whoa.

hvr:
I need to look into why a BOM (which btw makes no sense whatsoever for UTF8 encodings) doesn't round-trip properly. Iirc I specifically tested such corner-cases in the implementation of http://hackage.haskell.org/package/text-short

hvr:
After some investigation, the issue is in fact for the `String`-backed legacy fallback, whose `Binary` instance relies on the roundtrip property of `Distribution.Utils.String.{encode,decode}StringUtf8`, which fails for the BOM:

    > decodeStringUtf8 ( encodeStringUtf8 "\65534")
    "\65533"

because `decodeStringUtf8` (imho rightfully) considers a BOM invalid in an UTF8 stream, and maps it to the replacement-character.

hvr:
I'll take a stab at harmonizing the `decodeStringUtf8` semantics with the more round-tripping friendly ones from `text` and `text-short`.

hvr:
[...]

PR #4560, opened 2017-06-09
Experiment with Travis's build stages.

23Skidoo:
See #4556.

Issue #4556, opened 2017-06-07
Split up Cabal's Travis buildjobs into stages


hvr:
#4529 shows a problem we're starting to run into with cabal's heavy Travis jobs:

If a travis job gets killed due to timeout, it misses the opprotunity to write anything to the Travis cache.

However, there's a recently introduced [staged jobs feature](https://blog.travis-ci.com/2017-05-11-introducing-build-stages) which should allow us to build & cache dependencies in an early stage, so that if a later stage runs into a timeout, we'd still have at least not lost the cached work, which could then be reused by a restarted job.

Moreover, this could allow us to split work into smaller logical workunits which travis could then schedule accordingly (as our current jobs seem rather unevenly distributed in terms of work-length).

phadej:
Seems a bit difficult: https://github.com/travis-ci/beta-features/issues/11#issuecomment-306770378

23Skidoo:
The announcement says that "build stages work as expected when combined with build matrix expansion".

23Skidoo:
> The announcement says that "build stages work as expected when combined with build matrix expansion".

But see https://github.com/travis-ci/beta-features/issues/11#issuecomment-307431139.

23Skidoo:
So I sort of got this to work (with help from @BanzaiMan), but there are following problems:

1. Travis build stages are a bit coarse-grained for our purposes. In principle, each "test" job depends only on one "precache" job, but with build stages *all* "precache" jobs must finish before we get to the "test" stage. Ideally we'd want a more expressive way to describe job graphs that'd allow to expose more parallelism.
1. `new-build --only-dependencies` builds more than is needed by "precache". We don't want to build the Cabal lib and hackage-security, since they're not cached in the store, but `--only-dependencies` can't remove reverse dependencies of a target from the install plan:
```
$ cabal new-build --only-dependencies --enable-tests cabal-install cabal-tests Cabal --dry-run
cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the package Cabal-2.1.0.0 is required by a
dependency of one of the other targets.
```

hvr:
> `new-build --only-dependencies` builds more than is needed by "precache".

Btw, I've wanted to have a way to better control what is built (in this case you'd want only non-local deps built). In the case of matrix builder I would need the granularity of adressing to be able to select the build goal(s) by specifying the nix-hash (which I lookup in the `plan.json` file).

BardurArantsson:
[...]

PR #4902, opened 2017-11-22
Disable per-component builds when coverage is enabled

ttuegel:
#### Disable per-component build when program coverage enabled

Creating the test program coverage report requires the test runner to know the
build directory for the library component. If per-component builds are enabled,
this is not possible because the library is not in-scope at the same time as the
test component. The only apparent solution that does not require rethinking
per-component builds entirely is to disable it when coverage is enabled.

Resolves: #4798

cc @cartazio @cj285


cartazio:
i guess this would be fine (even if its a shame that theres that tension atm)

i'm ok with the coverage checking we (should do) being under a different flag / build way than component stuff .... which isn't terribly documented at the moment as far as I can tell? aside from here?
https://www.haskell.org/cabal/users-guide/installing-packages.html?highlight=component#setup-configure

ttuegel:
> even if its a shame that theres that tension atm

I didn't mean to imply this is a temporary situation ☹️ . It's part of the essential tension of per-component builds. I don't know how to get the information from the library component into the test component without violating the separation of components. Maybe @ezyang knows more about it.

ezyang:
It does seem like you'd have to basically figure out how to do coverage across packages to solve this problem in general.

@ttuegel is another possibility to only disable per-component on the "top" package, not all of them? (Maybe this patch does that already)

ttuegel:
> is another possibility to only disable per-component on the "top" package, not all of them?

This should only disable per-component builds on the package(s) that have coverage enabled. However, because command-line arguments are currently broken, passing `--enable-coverage` will enable coverage for *all* packages.

ttuegel:
> It does seem like you'd have to basically figure out how to do coverage across packages to solve this problem in general.

@ezyang 
The `ElaboratedComponent` knows its own dependencies; is there a way to determine which dependencies are from the same package? If so, we could probably construct the `DistDirParams` for each internal dependency and get the build directories that way.

ezyang:
> The ElaboratedComponent knows its own dependencies; is there a way to determine which dependencies are from the same package? If so, we could probably construct the DistDirParams for each internal dependency and get the build directories that way.

Yes, this should be possible. TBH, I don't know what the idiomatic way to do this is, but reading `elaborateSolverToComponents` should tell you the answer (slash, be the place to add the information you need.)

23Skidoo:
The new test seems to be failing on CI.

ttuegel:
> The new test seems to be failing on CI.

Yeah, in the test I added, I didn't check that GHC ships with shared libraries before testing. I *thought* that was a given; obviously not! 😃

ttuegel:
@cartazio 
Once I fix the test cases, I'm going to merge this and backport to cabal-install-2.0 because I don't have an ETA on the proper fix; hopefully this will not make it into cabal-install-2.2, but if it does then at least program coverage will work!

cartazio:
You have my gratitude

ttuegel:
For reasons currently unbeknownst to me, coverage is still broken with GHC 7.8 and GHC 7.10, but not with more recent versions. I suspect something has changed about how GHC handles HPC search paths, but I'm not sure yet.

cartazio:
Hrmmm.  That no good.

Happily that doesn’t affect me at all though :)
(I have this quaint belief that it’s very hard to get bug fixes into old
ghc releases so I should always use the current major version. )

ttuegel:
> I have this quaint belief that it’s very hard to get bug fixes into old ghc releases so I should always use the current major version.

I agree, nobody should use or support old GHCs, but here we are.

My fear is that the problem actually persists with GHC 8.x but the test case doesn't catch it, so I won't be satisfied until it works with GHC 7.x also.

cartazio:
Ok.  If there’s any ways I can help test on ghc 8 let me known

ttuegel:
The Travis failures are

1. A timeout, and
2. an unrelated test failure,

but I don't have the permissions to restart the failed tests.

23Skidoo:
Hmm, I thought you were already a member of the haskell/cabal group. @hvr,
can you please add @ttuegel?

ttuegel:
> I thought you were already a member of the haskell/cabal group.

Actually I was; I had gotten signed out of Travis and didn't notice. I can only restart the builds for `haskell/cabal` though. I can't restart the `haskell-pushbot/cabal-binaries` build. That was the timeout, though, so as long as the *tests* pass I'm satisfied.

23Skidoo:
@ezyang Can we add everyone in the `haskell/cabal` group to `haskell-pushbot`? Or maybe we can move pushbot to the `haskell` org?

ttuegel:
The timeout problem I thought was transient is not. Between the new tests I added, GHC 7.10 being slow, and the Travis macOS machines being slow, that build consistently fails. Is there a preferred solution for slow tests?

ezyang:
@23Skidoo I'm happy to but we have to decide what to do; haskell-pushbot is a user, not an org.

One thing we could try is to move the `cabal-binaries` repo to an org (we'll have to invite haskell-pushbot to the org before we can do this) and then add other people to the repo.

ezyang:
The "build is too long" situation is a bit of a problem.

So, the reason we are running tests in the "build" (and not the spillover job) is because all of these tests require the Cabal *library* to be used when a test script is run, but all of the compiled libraries are like 5G and are too big to put in the cabal-binaries repo. One way we might be able to work around this is to somehow add a mode to cabal-tests where we *compile in* all of the test scripts into the test binary, so we can go ahead and run it directly. But this requires some nontrivial dev work.

Another possibility is to speed up the build. We've probably gotten most of the low hanging fruit but there might still be some possibilities.

BTW osx-7.8.4 build is a legit failure, not timeout.

23Skidoo:
@ezyang So @hvr tells me that haskell-pushbot can't be part of the `haskell` org because it needs its own Travis build quota. I think it makes sense to do what you suggested -- make it an org (I'm not sure if it's possible to convert a user to an org) and manually add everyone from the `haskell/cabal` team to collaborators.

ttuegel:
The `osx-7.8.4` failure is unrelated, due to #4644.

ttuegel:
> The "build is too long" situation is a bit of a problem.

In this case, I can solve the problem by reducing redundancy. In `PackageTests/TestSuiteTests/ExeV10` we have `cabal-with-hpc.multitest.hs` and `setup-with-hpc.multitest.hs`. The former covers the same code paths as the latter, but runs them through `cabal`. It should be safe to remove the latter set of tests because any failure would be detected in the former.

ezyang:
Yes, we can probably speed things up a bit by running "setup or cabal" tests only in one config. Probably Setup!

ttuegel:
Unfortunately, removing a few redundant tests did not speed up the build enough with GHC 7.8 and 7.10 (the slower versions). I fear this PR is DOA unless we can find a solution to the build time problem.

23Skidoo:
I need to finally sit down and finish #4231, that should help macOS build times a bit. Maybe we can also skip haddock and sdist steps on macOS.

Here is a table of build steps performed by [one of the failed jobs](https://travis-ci.org/haskell/cabal/jobs/311477610) ordered by the time they take:

| Time (s) | Build step |
| -------- | ---------- |
| 976 | `cabal new-build -j2 cabal-install:cabal` |
| 596 | `cabal new-build -j2 Cabal Cabal:unit-tests Cabal:check-tests Cabal:parser-tests Cabal:parser-hackage-tests --enable-tests` |
| 593 | `[...]/cabal-tests/cabal-tests --builddir=[...] -j3 --skip-setup-tests --with-cabal [...]/cabal-install-2.1.0.0/build/cabal/cabal --with-hackage-repo-tool [...]/hackage-repo-tool` |
| 238 | `[...]/cabal-testsuite-2.1.0.0/build/cabal-tests/cabal-tests --builddir=[...] -j3` |
| 107 | `[...]/cabal-install-2.1.0.0/build/cabal/cabal update` |
| 56 | `cabal new-build -j2 cabal-testsuite:cabal-tests` |
| 34 | `[...]/cabal-install-2.1.0.0/setup/setup haddock --builddir=[...]` |
| 27 | `cabal act-as-setup --build-type=Simple -- haddock --builddir=[..]` |
| 24 | `cabal update` |
| 15 | `cabal new-build -j2 hackage-repo-tool` |

A bit weird that `cabal update` is on the fourth place, may be a fluke. The first time we call it takes 1/4 as long. Other than removing haddock from macOS configurations (I think we never had macOS-specific haddock failures), the only thing we can do is move more of the test suite to `cabal-binaries`. @ezyang, how feasible would that be?

ttuegel:
Is it actually necessary to run the package tests with `setup` and using `cabal` as setup? Wouldn't the latter tests exercise the same code paths as the former?

ttuegel:
> Package tests are being run by the cabal-tests executable FWICT.

Yes. Some of those tests are `setupTests`, some are `cabalTests` and some (many) are `setupAndCabalTests`. I'm asking if the `setupAndCabalTest`s could just be `setupTest`s.

ezyang:
You wouldn't have to search replace them all; just change the logic in `setupAndCabalTests` ;)

23Skidoo:
@ttuegel Not sure, but you can follow @ezyang's suggestion and disable setup tests on macOS only, since on Linux this doesn't seem to be a problem right now.

ttuegel:
> You wouldn't have to search replace them all; just change the logic in setupAndCabalTests ;)

It wouldn't really be `setup_And_CabalTests` then, would it? ;-)

I think I will just add a flag to `cabal-tests` to enable/disable `setupAndCabalTests` and we can flip that flag on macOS for one set of tests.

ezyang:
Well, the intent for these tests is that they *could* be run for setup or cabal... not that we actually do ;) Your plan sounds great.

ttuegel:
> I think I will just add a flag to cabal-tests to enable/disable setupAndCabalTests and we can flip that flag on macOS for one set of tests.

My plan was to only run the `setupAndCabalTests` once, but it turns out we already do this! I have no idea how to get the test suite time down or how to make this merge-able. Please advise.

23Skidoo:
@ezyang Do you think it'd be possible to run more of time-consuming tests mentioned in https://github.com/haskell/cabal/pull/4902#issuecomment-349299811 on `haskell-pushbot`?

ttuegel:
> So, the reason we are running tests in the "build" (and not the spillover job) is because all of these tests require the Cabal library to be used when a test script is run, but all of the compiled libraries are like 5G and are too big to put in the cabal-binaries repo.

Since we are already caching `$HOME/.cabal/store`, could we start by splitting the Cabal build from the cabal-install build? For that matter, could we separate the `cabal-tests` runs from the builds, too?

ttuegel:
OK, I think I have a pretty good idea of how to split up the Travis jobs so they don't time-out. I'll have a PR in a day or so.

23Skidoo:
@ttuegel If you want to experiment with build stages, you may want to take a look at my previous attempt (#4560 and https://github.com/haskell/cabal/issues/4556#issuecomment-307554993).

If you won't be able to work around the macOS timeout issue, I can see the following paths forward:

a) Merge your patch into the (soon-to-be-created) 2.2 branch, but without the test.
b) For 2.4/3.0, move macOS builds away from Travis -- not sure where yet, maybe GHC HQ will let us share their CircleCI account. If there's no suitable free service, we'll have to find money/sponsorship. I'm willing to invest time into configuring and administering an alternative macOS build bot.

cartazio:
i think we have at least one dedicated mac mini that @bgamari and @davean
can figure out how to add mac testing for cabal support to

23Skidoo:
This really should be in 2.2. One option (nuclear) is to temporarily mark the slowest macOS configurations `allow-failure` until we figure a more permanent solution.

ezyang:
We clearly don't have the bandwidth at the moment to fix the OS X build situation, so I vote for nuking the slow test.

23Skidoo:
We can add the code for the test, but disable it and add a TODO.

ttuegel:
How do I disable `cabal-testsuite` tests?

23Skidoo:
Good question. We still want to compile it (to prevent bit-rotting), but mark it as disabled for now. I believe that this is not supported right now, so we should add this feature to the test runner first. I may have time to look into it next week.

ezyang:
You can skip tests by calling `skip` in the test main. We'll still fire up the GHC interpreter on the test but then nothing else will happen.

23Skidoo:
Closing in favour of #5004.

ttuegel:
Thank you!

PR #4907, opened 2017-11-24
Have the solver reject packages with a too-new/unsupported spec-version

hvr:
This compares the request spec-version to the lib:Cabal's version in order to
determine whether cabal is able to properly understand the package. If it's
newer than the currently linked lib:Cabal version it's turned into a global
`FailResult` which the solver treats as desired in terms of backtracking and
error reporting.

This is related to the new spec-version forward-compat scheme (see #4899).

This is complements #4900 for the 2.0 branch.

Please include the following checklist in your PR:

* [ ] Patches conform to the [coding conventions](https://github.com/haskell/cabal/#conventions).
* [ ] Any changes that could be relevant to users have been recorded in the changelog.
* [ ] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

PR #4909, opened 2017-11-25
Have the solver reject packages with a too-new/unsupported spec-version

hvr:
This is a forward-port of #4907


23Skidoo:
Merged, thanks!

PR #4910, opened 2017-11-25
Skeleton for SPDX module generation [ci skip]

phadej:

### DIFF
--- a/cabal-spdx-codegen/ChangeLog.md
+++ b/cabal-spdx-codegen/ChangeLog.md
@@ -1,0 +1,5 @@
+# Revision history for cabal-sdpx-codegen
+
+## 0 -- YYYY-mm-dd
+
+* First version. Released on an unsuspecting world.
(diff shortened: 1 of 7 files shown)


23Skidoo:
Do you want `cabal-spdx-codegen` to go into `cabal-install`'s `build-tool-depends`? I'm afraid we can't do that, because we want `cabal install cabal-install` to work on older versions of `cabal-install`. Otherwise looks OK.

23Skidoo:
...so I think we're just going to check in the generated module. So when we do that we should also add a CI check that `cabal-spdx-codegen`'s output is identical to the checked-in module.

phadej:
Is it ok to commit `licenses.json` into Cabal repository.

https://github.com/spdx/license-list-data/issues/6 is related.

23Skidoo:
I think yes, why not?

phadej:
@23Skidoo ok, I guess there isn't problem, https://spdx.org/sites/cpstandard/files/pages/files/spdxversion2.1.pdf is licensed under *CC-BY-3.0* and has the same data, yet in not so nice format.

phadej:
See #4913

PR #4913, opened 2017-11-26
[WIP] Distribution.SPDX.LicenseId generation

phadej:

### DIFF
--- a/Cabal/Distribution/Utils/Generic.hs
+++ b/Cabal/Distribution/Utils/Generic.hs
@@ -353,7 +353,7 @@ isAsciiAlpha c = ('a' <= c && c <= 'z')
 -- False
 --
 isAsciiAlphaNum :: Char -> Bool
-isAsciiAlphaNum c = isAscii c ||  isDigit c
+isAsciiAlphaNum c = isAscii c && isAlphaNum c
 
 unintersperse :: Char -> String -> [String]
 unintersperse mark = unfoldr unintersperse1 where
--- a/cabal-dev-scripts/src/GenSPDXExc.hs
+++ b/cabal-dev-scripts/src/GenSPDXExc.hs
@@ -1,0 +1,127 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Main (main) where
+
+import Control.Lens   hiding ((.=))
+import Data.Aeson     (FromJSON (..), Value, eitherDecode, object, withObject, (.:), (.=))
+import Data.Char      (toUpper, isAlpha)
+import Data.Foldable  (for_)
+import Data.Semigroup ((<>))
+import Data.Text      (Text)
+
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Text            as T
+import qualified Data.Text.IO         as T
+import qualified Data.Text.Lazy       as TL
+import qualified Data.Text.Lazy.IO    as TL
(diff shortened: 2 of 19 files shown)


phadej:
Let's start small and get this merged as it is now.
I'll continue on SPDX stuff (next week the latest).

Cabal/Distribution/Utils/Generic.hs
```diff
@@ -353,7 +353,7 @@ isAsciiAlpha c = ('a' <= c && c <= 'z')
 -- False
 --
 isAsciiAlphaNum :: Char -> Bool
-isAsciiAlphaNum c = isAscii c ||  isDigit c
+isAsciiAlphaNum c = isAscii c && isAlphaNum c
```
phadej: this weren't used anywhere before :S

phadej:
This PR introduces SPDX modules. They aren't yet used in parsing `.cabal` files (need to get #4953 merged first)

travis-meta.sh
```diff
@@ -12,8 +12,13 @@ export PATH=/opt/cabal/head/bin:$PATH
 # Currently doesn't work because Travis uses --depth=50 when cloning.
 #./Cabal/misc/gen-authors.sh > AUTHORS
 
+timed cabal update
+
 # Regenerate files
 timed make gen-extra-source-files
+timed make spdx
```
23Skidoo: :+1: 

Would be nice to fix `make lexer` as well.

cabal-dev-scripts/src/GenSPDXExc.hs
```diff
@@ -0,0 +1,126 @@
...
+
+    special :: Text -> Text
+    special "389_exception" = "DS389_exception"
+    special t      = t
+
+-------------------------------------------------------------------------------
+-- Licenses
+-------------------------------------------------------------------------------
+
+data License = License
```
23Skidoo: Shouldn't this be `LicenseException`? A bit confusing that this one and the one in `GenSPDX.hs` have the same name, but are slightly different. Maybe move these types to a common module as well?

travis-meta.sh
phadej: I tried, it failed on travis (there were whitespace difference, no idea why)

cabal.project.meta
```diff
@@ -1 +1,2 @@
 packages: cabal-dev-scripts
+other-packages:
```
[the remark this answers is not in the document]
phadej: to not include `*/*.cabal` which causes it resolve to weird things.

cabal-dev-scripts/src/GenSPDXExc.hs
phadej: It should. I'll add a TODO note. I'm not keen on refactoring that too much now, as it might need bigger changes (hopefully not). It can be done after the release, as it's not part of delivery

cabal.project.meta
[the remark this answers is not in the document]
phadej: Ah, I meant it should be `optional-packages`, will fix that.

travis-meta.sh
23Skidoo: Different version of `alex` or some of its dependencies maybe? Though I have 3.1.7 from @hvr's PPA installed and the diff after `make lexer` is empty.

PR #4915, opened 2017-11-28
D.S.Program.Hpc: Use relative .mix search paths

ttuegel:
Prior to GHC 8.0, hpc requires relative search paths. We now pass relative paths
in all cases; we do not want to separately condition on GHC and HPC versions.

Frankly, it is surprising that this ever worked.

* [ ] ~Any changes that could be relevant to users have been recorded in the changelog.~
* [ ] ~The documentation has been updated, if necessary.~

I tested this on #4902, but I am making a separate pull request because this bugfix needs to be backported as far as possible.


ttuegel:
The one CI failure is due to a transient network problem, so I will merge this as-is.

Issue #4784, opened 2017-09-21
add support for generating BOTH library and app stanzas in cabal init 


cartazio:
currently the choices are 

```
What does the package build:
   1) Library
   2) Executable
```
there should be an option 3: BOTH! 

this is something i've been wanting for a while :) 

@cj285 might be interested in trying this out as a newcommer task if cabal folks would support adding this change in maybe even 2.2/2.4? :)

23Skidoo:
> adding this change in maybe even 2.2/2.4? :)

Sure, why not?

cartazio:
@cj285 do you still wanna  do this ?

hvr:
this has been accomplished via #4916

PR #4916, opened 2017-11-28
Initial feature add: create both Library and Executable on cabal init

cj285:

@cartazio @hvr This is the initial proposed change for allowing users to initialize their .cabal to have both a library and executables. I will be adding the changelog in a following commit.

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -734,25 +734,15 @@ createSourceDirectories flags = case sourceDirs flags of
 -- | Create Main.hs, but only if we are init'ing an executable and
 --   the mainIs flag has been provided.
 createMainHs :: InitFlags -> IO ()
-createMainHs flags@InitFlags{ sourceDirs = Just (srcPath:_)
-                            , packageType = Flag Executable
-                            , mainIs = Flag mainFile } =
-  writeMainHs flags (srcPath </> mainFile)
-createMainHs flags@InitFlags{ sourceDirs = _
-                            , packageType = Flag Executable
-                            , mainIs = Flag mainFile } =
-  writeMainHs flags mainFile
-createMainHs flags@InitFlags{ sourceDirs = Just (srcPath:_)
-                            , packageType = Flag LibraryAndExecutable
-                            , mainIs = Flag mainFile } =
-  writeMainHs flags (srcPath </> mainFile)
-createMainHs flags@InitFlags{ sourceDirs = _
-                            , packageType = Flag LibraryAndExecutable
-                            , mainIs = Flag mainFile } =
-  writeMainHs flags mainFile
-createMainHs _ = return ()
+createMainHs flags =
+  if hasMainHs flags then
+    case sourceDirs flags of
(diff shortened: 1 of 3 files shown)


cartazio:
this would add the feature proposed in https://github.com/haskell/cabal/issues/4784

cartazio:
@23Skidoo tada! can your or whomever owns cabal init give him some code review / feedback? as a naive onlooker, it looks fine to me, but i'm out of the loop on this stuff

cabal-install/Distribution/Client/Init/Types.hs
```diff
@@ -78,12 +78,14 @@ data InitFlags =
...
   -- not Flag [foo].
 
-data PackageType = Library | Executable
+data BuildType = LibBuild | ExecBuild
+
+data PackageType = Library | Executable | LibraryAndExecutable
   deriving (Show, Read, Eq)
 
 instance Text PackageType where
   disp = Disp.text . show
```
23Skidoo: Maybe make the `Text` instance display `LibraryAndExecutable` as "Library and Executable" so that it'd be shown like that in the menu? Though you'd also need to fix the `parse` function (unless we don't actually use it, then you can just get rid of the `Text` instance and make `disp` a standalone function).

cabal-install/Distribution/Client/Init.hs
```diff
@@ -742,6 +742,14 @@ createMainHs flags@InitFlags{ sourceDirs = _
                             , packageType = Flag Executable
                             , mainIs = Flag mainFile } =
   writeMainHs flags mainFile
+createMainHs flags@InitFlags{ sourceDirs = Just (srcPath:_)
```
23Skidoo: This function has become a bit boilerplatey, you can simplify it by getting rid of pattern matching and adding some helpers.

byorgey:
LGTM too.

cabal-install/Distribution/Client/Init.hs
cj285: I am not positive if the way I made the change will make things simpler to understand/less boilerplatey or not. Could I get a critique? Thanks!

cabal-install/Distribution/Client/Init.hs
```diff
@@ -870,30 +878,18 @@ generateCabalFile fileName c =
...
+           Flag Executable -> executableStanza
+           Flag Library    -> libraryStanza
+           Flag LibraryAndExecutable -> libraryStanza $+$ executableStanza
            _               -> empty
        ]
  where
-   generateBuildInfo :: PackageType -> InitFlags -> Doc
-   generateBuildInfo pkgtype c' = vcat
+   generateBuildInfo :: BuildType -> InitFlags -> Doc
+   generateBuildInfo artType c' = vcat
```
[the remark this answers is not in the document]
cj285: I forgot to change that part when I changed up the type name. Done!

23Skidoo:
Merged, thanks!

cabal-install/Distribution/Client/Init/Types.hs
23Skidoo: Done in 2b835e662cbc9cfb7dd6dea2ed5399409a566f6f.

PR #4917, opened 2017-11-29
D.S.Program.Hpc: Use relative .mix search paths

ttuegel:
Prior to GHC 8.0, hpc requires relative search paths. We now pass relative paths
in all cases; we do not want to separately condition on GHC and HPC versions.

* [ ] ~Any changes that could be relevant to users have been recorded in the changelog.~
* [ ] ~The documentation has been updated, if necessary.~

If necessary, I can backport this further, but I don't know that we make bugfix releases for older branches.

### DIFF
--- a/Cabal/Distribution/Simple/Program/Hpc.hs
+++ b/Cabal/Distribution/Simple/Program/Hpc.hs
@@ -19,6 +19,9 @@ module Distribution.Simple.Program.Hpc
 import Prelude ()
 import Distribution.Compat.Prelude
 
+import Control.Monad (mapM)
+import System.Directory (makeRelativeToCurrentDirectory)
+
 import Distribution.ModuleName
 import Distribution.Simple.Program.Run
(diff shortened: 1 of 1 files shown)


23Skidoo:
A bit of a pity this just missed 8.2.2, but it can get included into 8.2.3, assuming there will be one.

ttuegel:
> A bit of a pity this just missed 8.2.2, but it can get included into 8.2.3, assuming there will be one.

The hpc that ships with GHC 8.2 understands absolute paths. This fix is only necessary if you're using cabal-install-2.0 with GHC 7.10 or earlier, which would require building one's own Cabal library anyway. This would require doing an out-of-sync Cabal release, though.

23Skidoo:
> This would require doing an out-of-sync Cabal release, though.

OK, I can do that.

23Skidoo:
Merged, thanks!

Issue #4918, opened 2017-11-29
vty's test suite doesn't build with Cabal 2


felixonmars:
ref: https://github.com/jtdaugherty/vty/issues/132

I am getting the same errors as that report. The author of vty thinks it's a Cabal issue, but I didn't find it reported here.

```
Preprocessing test suite 'verify-parse-terminfo-caps' for vty-5.18.1..
Building test suite 'verify-parse-terminfo-caps' for vty-5.18.1..
[1 of 4] Compiling Verify           ( test/Verify.hs, dist/build/verify-parse-terminfo-caps/Verify.dyn_o )
[2 of 4] Compiling Verify.Data.Terminfo.Parse ( test/Verify/Data/Terminfo/Parse.hs, dist/build/verify-parse-terminfo-caps/Verify/Data/Terminfo/Parse.dyn_o )
[3 of 4] Compiling Verify.Graphics.Vty.Output ( test/Verify/Graphics/Vty/Output.hs, dist/build/verify-parse-terminfo-caps/Verify/Graphics/Vty/Output.dyn_o )
[4 of 4] Compiling VerifyParseTerminfoCaps ( test/VerifyParseTerminfoCaps.hs, dist/build/verify-parse-terminfo-caps/VerifyParseTerminfoCaps.dyn_o )

test/VerifyParseTerminfoCaps.hs:39:61: warning: [-Wdeprecations]
    In the use of ‘tiGetStr’
    (imported from System.Console.Terminfo, but defined in System.Console.Terminfo.Base):
    Deprecated: "use tiGetOutput instead."
   |
39 | fromCapname ti name = fromJust $ Terminfo.getCapability ti (Terminfo.tiGetStr name)
   |                                                             ^^^^^^^^^^^^^^^^^

test/VerifyParseTerminfoCaps.hs:49:66: warning: [-Wdeprecations]
    In the use of ‘tiGetStr’
    (imported from System.Console.Terminfo, but defined in System.Console.Terminfo.Base):
    Deprecated: "use tiGetOutput instead."
   |
49 |                         return $ case Terminfo.getCapability ti (Terminfo.tiGetStr capName) of
   |                                                                  ^^^^^^^^^^^^^^^^^

<no location info>: error: module ‘Verify’ cannot be found locally

<no location info>: error:
    module ‘Verify.Data.Terminfo.Parse’ cannot be found locally

<no location info>: error:
    module ‘Verify.Graphics.Vty.Output’ cannot be found locally
```

[...]

PR #4921, opened 2017-12-01
Don't pass `other-modules` to stub executable for detailed-0.9

hvr:
The stub executable is linked against the test-library
which already contains all `other-modules`, so we need
to remove those from the stub-exe's build-info to avoid
build errors due to duplication.

This addresses #4918

* [ ] Regression tests


hvr:
@ttuegel @23Skidoo fwiw, I wonder if there's more things that ought to be masked from the BuildInfo to avoid duplication (given the stub executable is a very thin shim calling into the actual test-library; a minimal BuildInfo should probably suffice); I just addressed the immediate issue from #4918

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.0.

PR #4922, opened 2017-12-02
Some minor fixes for the 2.0 branch

23Skidoo:
Updated to the latest release of `tasty` and added missing `other-modules`. The former patch will also be forward-ported to `master`.


23Skidoo:
Travis is performing macOS maintenance right now, so I think I'll just go ahead and merge this, changes are minor and CI is green on all Linux configurations.

PR #4926, opened 2017-12-03
Preliminary base-4.11 compatibility fixes

hvr:

### DIFF
--- a/cabal-install/Distribution/Client/List.hs
+++ b/cabal-install/Distribution/Client/List.hs
@@ -327,9 +327,9 @@ showPackageDetailedInfo :: PackageDisplayInfo -> String
 showPackageDetailedInfo pkginfo =
   renderStyle (style {lineLength = 80, ribbonsPerLine = 1}) $
    char '*' <+> disp (pkgName pkginfo)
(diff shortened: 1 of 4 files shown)


cabal-install/Distribution/Client/List.hs
```diff
@@ -327,9 +327,9 @@ showPackageDetailedInfo :: PackageDisplayInfo -> String
 showPackageDetailedInfo pkginfo =
   renderStyle (style {lineLength = 80, ribbonsPerLine = 1}) $
    char '*' <+> disp (pkgName pkginfo)
-            <>  maybe empty (\v -> char '-' <> disp v) (selectedVersion pkginfo)
+            Disp.<> maybe empty (\v -> char '-' Disp.<> disp v) (selectedVersion pkginfo)
```
23Skidoo: This kinda makes me long for OCaml's local module open feature.

hvr: :-) yeah...

fwiw, I used the module-qualifier strategy as it's more robust than having to use imports w/ `hiding`...

snoyberg:
Test

PR #4929, opened 2017-12-04
Solver: Deduplicate flags and stanzas in DependencyReason.

grayjay:
This commit changes DependencyReason's list fields to maps and sets. Duplicate
flags were possible when a flag appeared multiple times in nested conditionals
or a flag controlled a "Buildable: False" field. The duplicate flag could show
up in log messages:

Before:
```
[__5] trying: json-rpc-client:+demo
[__6] trying: process-1.6.1.0/installed-1.6... (dependency of json-rpc-client +demo +demo)
```

After:
```
[__5] trying: json-rpc-client:+demo
[__6] trying: process-1.6.1.0/installed-1.6... (dependency of json-rpc-client +demo)
```

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -21,6 +21,7 @@ module Distribution.Solver.Modular.Builder (
 
 import Data.List as L
 import Data.Map as M
+import Data.Set as S
 import Prelude hiding (sequence, mapM)
 
 import Distribution.Solver.Modular.Dependency
@@ -86,7 +87,7 @@ extendOpen qpn' gs s@(BS { rdeps = gs', open = o' }) = go gs' o' gs
     -- GoalReason for a flag or stanza. Each flag/stanza is introduced only by
     -- its containing package.
     flagGR :: qpn -> GoalReason qpn
-    flagGR qpn = DependencyGoal (DependencyReason qpn [] [])
+    flagGR qpn = DependencyGoal (DependencyReason qpn M.empty S.empty)
 
 -- | Given the current scope, qualify all the package names in the given set of
 -- dependencies and then extend the set of open goals accordingly.
--- a/cabal-install/Distribution/Solver/Modular/Dependency.hs
+++ b/cabal-install/Distribution/Solver/Modular/Dependency.hs
@@ -35,6 +35,8 @@ module Distribution.Solver.Modular.Dependency (
   ) where
 
 import Prelude ()
+import qualified Data.Map as M
+import qualified Data.Set as S
 import Distribution.Client.Compat.Prelude hiding (pi)
 
 import Language.Haskell.Extension (Extension(..), Language(..))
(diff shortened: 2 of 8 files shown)


grayjay:
I need to debug the test failures.

grayjay:
Thanks!

I fixed the bug in a second commit:

#### Solver: Remove redundant flag choices from FlaggedDeps.

Before this change, it was possible for the solver to create DependencyReasons
with conflicting flag values. The conflicting values could cause a problem when
one was FlagBoth, the value used when the same dependency appears on both sides
of a conditional.  Here is an example (an edge case):

```
if flag(A)
  if flag(A)
    build-depends: unknown-package
  else
    build-depends: unknown-package
```

Previously, the solver inserted both A=FlagTrue and A=FlagBoth into the
DependencyReason for the dependency on unknown-package. When the previous
commit changed the association list of flag values into a map, FlagBoth replaced
FlagTrue and prevented the solver from trying -A.

This commit simplifies the conditionals to:

```
if flag(A)
  build-depends: unknown-package
```

It also adds some unit tests for dependencies occurring on both sides of
conditionals. "testBackjumpingWithCommonDependency" tests the example above.

grayjay:
I think I fixed the issue that caused the quickcheck tests to frequently time out or run out of memory after I made the second commit.  I added two more commits, with a regression test in the fourth commit:

```haskell
-- | Test for an issue related to lifting dependencies out of conditionals when
-- converting a PackageDescription to the solver's internal representation.
--
-- Issue:
-- For each conditional and each package B, the solver combined each dependency
-- on B in the true branch with each dependency on B in the false branch. It
-- added the combined dependencies to the build-depends outside of the
-- conditional. Since dependencies could be lifted out of multiple levels of
-- conditionals, the number of new dependencies could grow exponentially in the
-- number of levels. For example, the following package generated 4 copies of B
-- under flag-2=False, 8 copies under flag-1=False, and 16 copies at the top
-- level:
--
-- if flag(flag-1)
--   build-depends: B, B
-- else
--   if flag(flag-2)
--     build-depends: B, B
--   else
--     if flag(flag-3)
--       build-depends: B, B
--     else
--       build-depends: B, B
--
-- This issue caused the quickcheck tests to start frequently running out of
-- memory after an optimization that pruned unreachable branches (See PR #4929).
-- Each problematic test case contained at least one build-depends field with
-- duplicate dependencies, which was then duplicated under multiple levels of
-- conditionals by the solver's "buildable: False" transformation, when
-- "buildable: False" was under multiple flags. Finally, the branch pruning
-- feature put all build-depends fields in consecutive levels of the condition
-- tree, causing the solver's representation of the package to follow the
-- pattern in the example above.
--
-- Now the solver avoids this issue by combining all dependencies on the same
-- package within a build-depends field before lifting them out of conditionals.
--
-- This test case is an expanded version of the example above, with library and
-- build-tool dependencies.
duplicateDependencies :: String -> SolverTest
duplicateDependencies name =
    mkTest pkgs name ["A"] $ solverSuccess [("A", 1), ("B", 1)]
  where
    copies, depth :: Int
    copies = 50
    depth = 50

    pkgs :: ExampleDb
    pkgs = [
        Right $ exAv "A" 1 (flaggedDependencies 1)
      , Right $ exAv "B" 1 [] `withExe` ExExe "exe" []
      ]

    flaggedDependencies :: Int -> [ExampleDependency]
    flaggedDependencies n
        | n > depth = buildDepends
        | otherwise = [exFlagged (flagName n) buildDepends
                                              (flaggedDependencies (n + 1))]
      where
        buildDepends = replicate copies (ExFix "B" 1)
                    ++ replicate copies (ExBuildToolFix "B" "exe" 1)

    flagName :: Int -> ExampleFlagName
    flagName x = "flag-" ++ show x
```

23Skidoo:
Merged, thanks!

PR #4934, opened 2017-12-05
Skip the Haddock step on macOS.

23Skidoo:
I don't think it ever uncovered any bugs that the Linux bot missed. This should speed up the build a little.


23Skidoo:
Travis is green, AppVeyor is stuck for some reason, but it doesn't matter. Merging.

23Skidoo:
See discussion in #4902 for context.

Issue #4452, opened 2017-04-17
Cannot override --flags= on command line


kolmodin:
**WARNING** There are updates to this ticket, please read the discussion.

**Summary.** It's common that the last given value of a flag "wins".

```
$ cabal configure -fdev -v
...
Flags chosen: dev=True
...

$ cabal configure -f-dev -v
...
Flags chosen: dev=False
....

$ cabal configure -fdev -f-dev -v
...
Flags chosen: dev=True
...
```

I expected the last line to have set `dev=False`. According to the [documentation](https://www.haskell.org/cabal/users-guide/installing-packages.html#controlling-flag-assignments) it should be `dev=False`.

**Explanation (by @ezyang).**

Cabal has quite flexible handling of duplicate instances of arguments when doing command line handling. It does so using *monoids*, which are types which have a binary "combining" operation defined for them. ConfigFlags, defined in `Cabal/Distribution/Simple/Setup.hs`, defines the monoid instance for ALL configuration flags, by applying the monoid for each individual field.

Flag assignment is recorded in `configConfigurationsFlags`, which is defined to be a list. The monoid instance for lists is append. This means that when a flag shows up twice, the lists get appended together; with the example above, you end up with `[("dev", True), ("dev", False)]`.

The bug is in this code `Distribution/PackageDescription/Configuration.hs`:

```
    d2c manual n b = case lookup n userflags of
                     Just val -> [val]
                     Nothing
                      | manual -> [b]
                      | otherwise -> [b, not b]
```

Here, `userflags` is the list of flag assignments, possibly with duplicates, in the order they occurred on the command line. The problem is that `lookup` will pick up the FIRST entry, rather than the LAST entry. A better and more efficient strategy is to build a map, preferring the last entry, and then do a lookup on this map. If you do that, the bug will be fixed!

**What you will learn.** How to build Cabal. How to convert a key-value list of items into a map, and then do a lookup on it. How to do command line arguments with Monoids. How to look through the verbose output of cabal configure to find the info you need.

**Going further.** This bug doesn't really require a test, but if you want to add one, it should be fairly easy. Follow the instructions in https://github.com/haskell/cabal/tree/master/cabal-testsuite

kolmodin:
```
$ cabal --version
cabal-install version 1.25.0.0
compiled using version 1.25.0.0 of the Cabal library 
```

23Skidoo:
Makes sense.

colinwahl:
[...]

PR #4940, opened 2017-12-08
Makes flags overridable on the command line.

dalaing:
This is to fix issue #4452, but updating `FlagAssignment` to be a `Map` with the appropriate `Semigroup` and `Monoid` instance that favours later entries.

I ran variations of `cabal act-as-setup -- configure -fdev -f-dev -v ` and `cabal configure -fdev -f-dev -v ` on a dummy project to test this out.  I'm happy to have a go at adding some tests on Monday if that helps - those bonus points are tempting me...

### DIFF
--- a/Cabal/Distribution/Types/GenericPackageDescription.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription.hs
@@ -16,6 +16,7 @@ module Distribution.Types.GenericPackageDescription (
     lookupFlagAssignment,
     insertFlagAssignment,
     diffFlagAssignment,
+    findDuplicateFlagAssignments,
     nullFlagAssignment,
     showFlagValue,
     dispFlagAssignment,
@@ -25,11 +26,11 @@ module Distribution.Types.GenericPackageDescription (
 ) where
 
 import Prelude ()
-import Data.List ((\\))
 import Distribution.Compat.Prelude
 import Distribution.Utils.ShortText
 import Distribution.Utils.Generic (lowercase)
 import qualified Text.PrettyPrint as Disp
+import qualified Data.Map as Map
 import qualified Distribution.Compat.ReadP as Parse
 import qualified Distribution.Compat.Parsec as P
 import Distribution.Compat.ReadP ((+++))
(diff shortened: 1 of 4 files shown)


23Skidoo:
Nice, thanks!

dalaing:
I just added some things to tidy up some things based on comments in the surrounding code - mostly to do with the duplicates check that is no longer necessary in `configuredPackageProblems`.

hvr:
@23Skidoo  The reason I originally didn't turn the newtype into a `Map` was because this would throw the duplicate-flag validation out. Do we really want to fail silently here?

dalaing:
I could change it to a map to `(Int, Bool)` where the `Int` counts the number of times that flag was observed and restore the duplicate flags warning via those means.

23Skidoo:
So that you'd get a warning about `-fflag -f-flag`? Makes sense.

dalaing:
I'm not exactly sure if/when you get a warning, but it should at least be compatible with the existing code that checks for duplicate flags.

Cabal/Distribution/Types/GenericPackageDescription.hs
```diff
@@ -154,55 +155,74 @@ instance Text FlagName where
...
 lookupFlagAssignment :: FlagName -> FlagAssignment -> Maybe Bool
-lookupFlagAssignment fn = lookup fn . unFlagAssignment
+lookupFlagAssignment fn = fmap snd . Map.lookup fn . getFlagAssignment
 
 -- | Insert or update the boolean value of a flag.
 --
+-- If the flag is already present in the 'FlagAssigment', the
+-- value will be updated and the fact that multiple values have
+-- been provided for that flag will be recorded so that a
+-- a warning can be generated later on.
```
gwils: There is an 'a' both before and after the line break.

23Skidoo: Good catch!

dalaing:
I'm going to try bumping the hash of the custom-setup component for the failing test to see if that helps / if the other tests pass after that.  I figured I should ask while that is running: is that a valid thing to do for this kind of change?

23Skidoo:
> bumping the hash of the custom-setup component

I'm not quite sure what you have in mind here. Anyway, I think the failing tests were mostly due to #4962.

dalaing:
In the last run, there was a difference in the custom-setup hash in `cabal-testsuite/PackageTests/Regression/T3436/sandbox.out` that was breaking things.

23Skidoo:
I see. It's most likely okay to do this in this case.

dalaing:
This is now failing in the print / parse round-trip property based tests.

I can see two ways to deal with this: I can write something to reset the count of duplicates in the various data structures that contain `FlagAssignments`, or I can write an `instance Eq FlagAssignment` that ignores the count of duplicates.

I'm inclined to go for the latter, because with have `findDuplicateAssignments` for when the counts matter, and other than where that code is used I don't think the counts will come up that often.

dalaing:
I've put up custom instances of `Eq` and `Ord` as a first pass for this.  I can write a version of the `Binary` instance as well if that seems helpful / needed.

23Skidoo:
The sole remaining is a timeout. Restarted.

hvr:
@dalaing fwiw, there was also the idea to make make `FlagAssignment` take a type-parameter; so we could have 

- `FlagAssignment Bool`
- `FlagAssignment BoolSet`

```hs
data BoolSet = BoolSet !Bool !Int
```

or some other variant

that way we'd be able to have two types, one where the uniqueness invariant is guaranteed and it has a heap-friendly representation (since `Bool`'s values are singletons on the heap; and also because we don't lose anything by using a type-var, since the underlying `Map` also has a type-var for the value)

while the `BoolSet` encodes our input;

that way we also don't have to introduce `Eq` instances which violate leibniz equality

dalaing:
I think something is actually a bit wrong with this at the moment.

We have the `Int`s in there so that we can use them to detect duplicates in `cabal-install/Distribution/Client/Dependency.hs:configuredPackageProblems`, which is used in `resolveDependencies`.  

That seems to have been not picking up on duplicates since before this change.  I've been trying a few things and I'm unable to trigger a duplicate flag error, and I'm having trouble tracking down why.  Maybe the `FlagAssignment`s are round tripping through a list or getting otherwise partly de-duplicated already?

If that check is no longer needed (rather than just being temporarily broken), then maybe we don't need the `Int`s in the `FlagAssignment`s anymore.

23Skidoo:
I'd prefer to fix the duplicate check if possible.

23Skidoo:
Looks like the duplicates check in `D.C.Dependency` is just a sanity check for the solver's output. I'm actually starting to think that with this change it could be omitted, since the result should be now correct by construction. @grayjay, do you agree?

But we can add checks for duplicates at the locations where `FlagAssignment` is constructed, so that `-ffoo -ffoo` would generate a warning.

23Skidoo:
Merged manually (see 4e4180b9c86779a86e489159f250289303e2da70). Thanks!

23Skidoo:
> But we can add checks for duplicates at the locations where `FlagAssignment` is constructed, so that `-ffoo -ffoo` would generate a warning.

Looks like doing this is somewhat harder than I expected, since it's normally constructed directly via `parsecFlagAssignment`. I tried emitting a `PWarning`, but it looks like it just gets ignored by `parsecToReadE`.

grayjay:
> Looks like the duplicates check in D.C.Dependency is just a sanity check for the solver's output. I'm actually starting to think that with this change it could be omitted, since the result should be now correct by construction. @grayjay, do you agree?

Yes, I think that configuredPackageProblems just double checks the solver's output.

Issue #1375, opened 2013-06-17
Installs in sandbox 'forgotten' if final build fails


hesselink:
If you `add-source` some packages in a sandbox, and they get installed as dependencies of a package whose build ultimately fails, they are reinstalled afresh the next time you issue an install command. For example:
- Have packages `a` and `b`, where `b` depends on `a`.
- `cabal sandbox add-source a/`
- `cabal install b/`

This installs `a` and then `b`. If `b` has e.g. a type error, you fix it and issue another `cabal install b/`, `a` also gets installed again, even though it was already installed and nothing about it changed.

acfoltzer:
Is this now on track for 1.24? I have to do a bunch of workarounds currently to avoid lost time due to this, like scripting smaller install plans together to avoid packages that might be temporarily broken during development.

ygale:
Recording here the simple workaround that a lot of people already know:

Unless you really do expect the external source to change often and require rebuilds, use `add-source --snapshot`. If at some point later you do need it rebuilt, you cad do `delete-source` and then `add-source` without `--snapshot`. If you just need a new version, you can do `add-source --snapshot` also for the newer version, and the solver will then consider both of them.

23Skidoo:
Fixed in fcabd87406effc0c265312eb38509804682e4623.

PR #4946, opened 2017-12-11
Fix how the timestamps of add-source packages are updated.

dalaing:
This is to address #1375.

Previously they were updated in the post install phase, which
would mean that the sources did not have their timestamps
updated unless the package being installed completed successfully.

This change moves the updating of the timestamps to the end of
the install phase, so that the add-source packages that are
successfully installed get their timestamps updated regardless of
the overall outcome of the installation.

This was tested locally with a setup mirroring that in the issue.

### DIFF
--- a/cabal-install/Distribution/Client/Install.hs
+++ b/cabal-install/Distribution/Client/Install.hs
@@ -817,7 +817,7 @@ postInstallActions :: Verbosity
                    -> BuildOutcomes
                    -> IO ()
 postInstallActions verbosity
-  (packageDBs, _, comp, platform, progdb, useSandbox, mSandboxPkgInfo
+  (packageDBs, _, comp, platform, progdb, useSandbox, _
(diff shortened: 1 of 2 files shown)


dalaing:
I'll look at adding a test for this today - I just thought I'd put what I've got up here for comment in the meantime.

dalaing:
I thought the same thing when looking at the code, but the bug still remains without the change and goes away with the change. I should probably take a look and work out what doesn't get run and why when an install fails, just for my own understanding.

23Skidoo:
Weird! Maybe an exception gets raised before we get to that step?

23Skidoo:
Can confirm that this fixes the issue.

23Skidoo:
Committed a slightly tweaked version of this in fcabd87406effc0c265312eb38509804682e4623. Verified that that version also fixes the issue. Thanks!

Issue #4723, opened 2017-08-28
cabal fetch should support the --enable-tests option


asr:
Is there some reason why `cabal fetch` is not supporting the `--enable-tests` option?

23Skidoo:
No deep reason, patches welcome.

dalaing:
Is this closed by the merged PR, or do we want to keep it around as a reminder for when cabal new-fetch lands?

23Skidoo:
Fixed by #4948.

PR #4948, opened 2017-12-11
Adds --enable-tests and --enable-benchmarks to 'cabal fetch'.

dalaing:
This is a work in progress to address #4723, that I'm putting up here to get some feedback because I'm not sure what other wiring needs to be done for new options to commands.

I've tested this with the various combinations of `--enable-tests` / `--enable-benchmarks` with `cabal-fetch --dry-run (other options here) ad` and it seems to do what it should.

### DIFF
--- a/cabal-install/Distribution/Client/Fetch.hs
+++ b/cabal-install/Distribution/Client/Fetch.hs
@@ -1,5 +1,4 @@
------------------------------------------------------------------------------
--- |
+------------------------------------------------------------------------------- |
 -- Module      :  Distribution.Client.Fetch
 -- Copyright   :  (c) David Himmelstrup 2005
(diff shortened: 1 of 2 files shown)


23Skidoo:
This looks good to me. @grayjay, can you also please take a look?

grayjay:
It looks good to me, though I'm not sure what else needs to be done.

hvr:
We'll, we'll need the `cabal new-fetch` equivalent of this long-term, don't we? `cabal fetch` computes a different install-plan from `cabal new-build`, doesn't it?

23Skidoo:
Merged, thanks! I think `new-fetch` can be dealt with separately.

Issue #4939, opened 2017-12-08
cabal `new-build` is utterly unusable for cross-compilation


angerman:
With `new-build` we apply the `--with-PROG` and `--PROG-option(s)` only to the current project local package. As such, they are not pushed down to dependencies. To illustrate this:

```
- a
  - b
    - d
  - c
```
so we have `a -> b -> d` and `a -> c -> d`

If we try to build with new-build:

```
$ cabal new-build --with-hsc2hs=myhsc2hs --hsc2hs-options=--cross-compile
```

we compile `a` with `myshc2hs --cross-compile`, but for `b`, `c` and `d` we get no `hsc2hs` in the package database, and as such do the canonical lookup, which gives us `hsc2hs`. Of course we are missing `--cross-compile` now as well.

Let us assume without restriction `c` has some `.hsc` files. `c` now fails to compile, and subsequently the complete build fails. Worse, I have no option to provide those via the `cabal.project` file, as the `program-options` and `program-locations` stanzas are seemingly not picked up as nodes of the `package` stanza.

---

Example:
```
packages: .

repository hackage.mobilehaskell
  url: http://hackage.mobilehaskell.org/
  secure: True
  root-keys: 8184c1f23ce05ab836e5ebac3c3a56eecb486df503cc28110e699e24792582da
             81ff2b6c5707d9af651fdceded5702b9a6950117a1c39461f4e2c8fc07d2e36a
             8468c561cd02cc7dfe27c56de0da1a5c1a2b1b264fff21f4784f02b8c5a63edd
  key-threshold: 3

repository head.hackage
   url: http://head.hackage.haskell.org/
   secure: True
   root-keys: 07c59cb65787dedfaef5bd5f987ceb5f7e5ebf88b904bbd4c5cbdeb2ff71b740
              2e8555dde16ebd8df076f1a8ef13b8f14c66bad8eafefd7d9e37d0ed711821fb
              8f79fd2389ab2967354407ec852cbe73f2e8635793ac446d09461ffb99527f6e
   key-threshold: 3

package zlib
  program-options
    ghc-options: -fllvmng
    hsc2hs-options: --cross-compile

  program-locations
    gcc-location: armv7-linux-androideabi-clang
    ghc-location: armv7-linux-androideabi-ghc
    ghc-pkg-location: armv7-linux-androideabi-ghc-pkg
    hsc2hs-location: armv7-linux-androideabi-hsc2hs
    ld-location: armv7-linux-androideabi-ld
```

[...]

PR #4949, opened 2017-12-11
Mark 'ghc-boot-th' as non-installable.

23Skidoo:

### DIFF
--- a/cabal-install/Distribution/Client/Dependency.hs
+++ b/cabal-install/Distribution/Client/Dependency.hs
@@ -374,6 +374,7 @@ dontUpgradeNonUpgradeablePackages params =
       -- If you change this enumeration, make sure to update the list in
       -- "Distribution.Solver.Modular.Solver" as well
       , pkgname <- [ mkPackageName "base"
+                   , mkPackageName "ghc-boot-th"
                    , mkPackageName "ghc-prim"
                    , mkPackageName "integer-gmp"
                    , mkPackageName "integer-simple"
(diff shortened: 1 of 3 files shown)


angerman:
Ah! I didn't even know we had this! Would this mean, that if I pushed flags through all dependencies, cabal would never try to rebuild those packages? (E.g. related to #4939).

Coincidentally there was a discussion on #hackage today with @sclv and @hvr where the question cam up if `ghc --info` should just list the non-reinstallable packages.

GHC basically knows about
```
wiredInUnitIds :: [UnitId]
wiredInUnitIds = [ primUnitId,
                       integerUnitId,
                       baseUnitId,
                       rtsUnitId,
                       thUnitId,
                       thisGhcUnitId,
                       dphSeqUnitId,
                       dphParUnitId ]
```
as being wired in packages.

Anyway, this is only tangential to this PR.

23Skidoo:
@angerman There's `--allow-boot-library-installs` that allows you to ignore the blacklist.

23Skidoo:
I discussed this with @hvr and we decided that it's better to keep the blacklist as small as possible. `ghc-boot-th` is only non-reinstallable because `template-haskell` depends on it, otherwise it could very well be.

23Skidoo:
I wouldn't mind if the blacklist was populated automatically. 

> ```
> dphSeqUnitId,
> dphParUnitId ]
> ```

Looks like those are not built by default, e.g. `ghc-pkg list` on my system doesn't list them.

angerman:
I think dph is effectively dead

PR #4950, opened 2017-12-11
Update mtime on repo index file even if up to date

psiska:
Resolves: #4444

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -19,6 +19,7 @@
 module Distribution.Client.IndexUtils (
   getIndexFileAge,
   getInstalledPackages,
+  indexBaseName,
   Configure.getInstalledPackagesMonitorFiles,
   getSourcePackages,
   getSourcePackagesMonitorFiles,
--- a/cabal-install/Distribution/Client/Update.hs
+++ b/cabal-install/Distribution/Client/Update.hs
@@ -26,7 +26,7 @@ import Distribution.Client.FetchUtils
 import Distribution.Client.IndexUtils.Timestamp
 import Distribution.Client.IndexUtils
          ( updateRepoIndexCache, Index(..), writeIndexTimestamp
-         , currentIndexTimestamp )
+         , currentIndexTimestamp, indexBaseName )
 import Distribution.Client.JobControl
          ( newParallelJobControl, spawnJob, collectJob )
 import Distribution.Client.Setup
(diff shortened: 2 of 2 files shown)


23Skidoo:
Haven't reviewed in detail, but this looks similar to #4578. Have you looked at that PR?

psiska:
Yes I did have a look on #4578. Approach in that issue was to use mtime of timestamp files instead of index files. But actually you have pointed out it works only for secure remote repos. 

Approach taken in this issue is like follows.
Warnings is emitted only when mtime on index is more than 15 days. So let's just modify mtime even if the local index is same as remote.
Covers both secure and unsecure repos.

23Skidoo:
Awesome, I'll try to test this today or tomorrow.

23Skidoo:
I tested this and it seems to work.

23Skidoo:
Merged, thanks!

23Skidoo:
Merged manually, see f05615c0f7972d4aa64284a10e1be1c7dc3a34ce.

phadej:
GHC-7.10.3 (`directory-1.2.2.0`):
```
Distribution/Client/Update.hs:43:26:
    Module ‘System.Directory’ does not export ‘setModificationTime’
```

`setModificationTime` is only in `directory-1.2.3.0`.  http://hackage.haskell.org/package/directory-1.3.1.5/changelog

Please, be **very careful** with manual merges. Also https://travis-ci.org/haskell/cabal/jobs/315016658 (build/job of this PR) shows this failure.

Ping @23Skidoo

23Skidoo:
Ugh, I suspected that this would be the case. Will fix.

23Skidoo:
BTW, it looks like we should apply the same fix to `new-update` too (`D.C.CmdUpdate`). @psiska, would you be interested in looking into that?

psiska:
Yes, I can have a look on Thursday / Friday. 
Regarding the breakage - it looks like there is difference in what version of 'directory' is installed. The older version (< 1.2.3.0) is installed only with 7.10.3 - all other builds are choosing newer directory. 

Looking into travis-install.sh does not reveal anything strange to me as the cabal-2.0.0.0 is installed in all environments. The fact that same cabal chooses different version of directory, looks to me like some kind of problem. Is this normal behavior?

23Skidoo:
@psiska GHC 7.10 [comes with `directory-1.2.2.0`](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory), and since that version is not excluded by the version bounds in `cabal-install.cabal`, it gets picked up by the solver. #4962 fixes the build error.

PR #4958, opened 2017-12-17
Improve `build-type` defaulting

hvr:
A personal pet-peeve of mine when writing `.cabal` files from scratch is forgetting about declaring the `build-type`, which historically will then be inferred to be `Custom` which is almost always *not* what is desired. Instead, a better default is `Simple` as that represents the most basic and recommended `build-type`. Morever, starting with `cabal-version:1.24`, we have a redundant signal in the `.cabal` file to statically imply a custom `build-type`, namely the `custom-setup` stanza.

To this end, this PR finally removes this papercut by improving the defaulting to use the following simple rules:

 - When `cabal-version` is set to `2.1` or higher,   the default is `Simple` unless a `custom-setup` stanza exists, in which case the inferred default is `Custom`.
- For lower `cabal-version`s below `2.1`, the default is `Custom` unconditionally (legacy defaulting).

This allows us to bring down the minimal (modern) trivial cabal package definition down to a single file with 4 lines:

```    
cabal-version: 2.1
name: mu
version: 0
library
```
    
NB: We don't need any `Setup.hs` file, as `cabal new-build` doesn't need any, and `cabal sdist` will magically generate one on the fly and include it in the source distribution.

----

* [x] Test-suite has been extended/updated

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -448,7 +448,7 @@ checkFields pkg =
            "Package names with the prefix 'z-' are reserved by Cabal and "
         ++ "cannot be used."
 
-  , check (isNothing (buildTypeRaw pkg)) $
+  , check (isNothing (buildTypeRaw pkg) && specVersion pkg < mkVersion [2,1]) $
       PackageBuildWarning $
            "No 'build-type' specified. If you do not need a custom Setup.hs or "
         ++ "./configure script then use 'build-type: Simple'."
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -448,20 +448,20 @@ checkFields pkg =
            "Package names with the prefix 'z-' are reserved by Cabal and "
         ++ "cannot be used."
 
-  , check (isNothing (buildType pkg)) $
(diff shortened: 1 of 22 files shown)


hvriedel:
In a future refactoring, `UnknownBuildType` should IMO be removed

23Skidoo:
Yes, makes sense.

23Skidoo:
Test failure is spurious, some `apt-get` failure.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -448,20 +448,20 @@ checkFields pkg =
            "Package names with the prefix 'z-' are reserved by Cabal and "
         ++ "cannot be used."
 
-  , check (isNothing (buildType pkg)) $
+  , check (isNothing (buildTypeRaw pkg) && specVersion pkg < mkVersion [2,1]) $
```
phadej: This is great change, which one should mention in the changelog, It's updated in the manual (great!), but let's advertise this change in Changelog too.

Optionally, we can add a section in changelog gathering all cabal format related changes.

hvr: Right!

Fwiw, I've got more related changes to for `cabal check` & build-type in the pipeline; the idea about structuring the changelog is a good one too, although I've been wanting to have an explicit cabal spec changelog in the user's guide as well... :-)

PR #4962, opened 2017-12-19
Unbreak the build with directory < 1.2.3.0.

23Skidoo:
See https://github.com/haskell/cabal/pull/4950#issuecomment-352753882 for context.

### DIFF
--- a/cabal-install/Distribution/Client/Compat/Directory.hs
+++ b/cabal-install/Distribution/Client/Compat/Directory.hs
@@ -1,0 +1,13 @@
+{-# LANGUAGE CPP #-}
+module Distribution.Client.Compat.Directory (setModificationTime) where
+
+#if MIN_VERSION_directory(1,2,3)
+import System.Directory (setModificationTime)
+#else
+
+import Data.Time.Clock (UTCTime)
+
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Compat/Directory.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE CPP #-}
+module Distribution.Client.Compat.Directory (setModificationTime) where
+
+#if MIN_VERSION_directory(1,2,3)
+import System.Directory (setModificationTime)
+#else
+
+import Data.Time.Clock (UTCTime)
+import System.IO (FilePath)
```
phadej: IIRC `FIlePath` is in `Prelude`: http://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#t:FilePath

23Skidoo: Yep. Fixed.

23Skidoo:
This passed on Linux, I'll just go ahead and merge because a bunch of other stufff is blocked on this.

Issue #1966, opened 2014-06-26
cabal format adds empty lines with trailing whitespace


tibbe:
If you run `cabal format` on e.g. in the `cabal-install/` directory, you'll see that `cabal format` adds empty lines with trailing whitespace (i.e. 4 spaces). It shouldn't do that.

osa1:
Cabal file generators(`cabal format`, `cabal init`, probably some other commands?) always do this. As also pointed out by @23Skidoo this is related with `pretty` library, which generates trailing space if we have a code like `nest 4 (text "" $+$ ...)`.

I can think several solutions:

1. We can modify `pretty` library and add a new `EmptyLine` constructor to `Doc`, then handle that in vertical composition functions specially to not generate any whitespace, only generate a "\n". Then we can use `emptyLine` instead of `text ""`.

2. We can make significant refactoring in Cabal pretty printer to not have any `text ""` inside `nest`. (this is probably very hard to do, and resulting code may not be very good)

3. We can wrap `writeCabalFile` and `writeGenericPackageDescription` with a function that removes trailing white space with the cost of an extra copy.

(3) seems to be the easiest solution, however I think (1) is the best, because as far as I can understand from the API docs of `pretty`, it doesn't provide a way to insert newlines, without having them indented and this seems to be a limitation of the library.

What do you think? Should we go with (3) for now?

23Skidoo:
I like option (1) best, (3) could be a temporary workaround.

osa1:
I implemented (1), but I need some reviewers. I'm trying to get in contact with `pretty` maintainers, hopefully we can fix this in a principled way.

23Skidoo:
@osa1 Great, thank you for working on this.

byorgey:
What's the status of this?  @osa1 , can you post here a link to your implementation of (1)?

osa1:
I feel like it's trivial to implement for someone who knows internals of `pretty`, and I asked for help here https://github.com/haskell/pretty/issues/26 because I don't know the internals, and lack of documentation is a problem. Main problem is that even though it's possible to do without adding any new `Doc` constructor, some internals, documentation and maybe even invariants need to be changed, and it's not trivial for me to figure out what to change. I think I can spare some time for this after tomorrow.

gbaz:
Is this resolved?

phadej:
Not yet. #5737 resolve this though.

jneira:
[...]

PR #4968, opened 2017-12-22
Kill trailing whitespaces in `.cabal` file emitted by `cabal init`

hvr:
`cabal init` currently generates trailing whitespaces; this patch
simply adds a final post-processing step filtering out any such trailing
whitespace.

Below is an example (the red markers represent the annoying end-of-line trailing whitespace); with this patch, all those red areas are gone.

whitespace_cabal_init

----

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -76,6 +76,8 @@ import Distribution.ReadE
   ( runReadE, readP_to_E )
 import Distribution.Simple.Setup
   ( Flag(..), flagToMaybe )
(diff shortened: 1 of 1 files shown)


hvr:
btw, I'm not saying this is the only way to fix this; but this was the quickest one ;-)

23Skidoo:
OK. Someone tried to fix this in `pretty`, but that patch was never finished. See #1966.

23Skidoo:
Also `cabal format` apparently suffers from the same issue.

23Skidoo:
Merged, thanks!

PR #4970, opened 2017-12-23
Enable 'make lexer' in travis-meta.

23Skidoo:

### DIFF
--- a/Cabal/Distribution/Parsec/Lexer.hs
+++ b/Cabal/Distribution/Parsec/Lexer.hs
@@ -53,7 +53,6 @@ import qualified Data.Text.Encoding as T
 import qualified Data.Text.Encoding.Error as T
 #endif
 
-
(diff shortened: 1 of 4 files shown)


23Skidoo:
The part where it differs apparently comes from `/usr/include/stdc-predef.h`. I can imagine that file being slightly different on my machine and on Travis for some reason. One solution is to squash repeated empty lines in the generated file.

23Skidoo:
Meta seems to pass now.

23Skidoo:
Pushbot failures are OOM:s, will need to look at these separately.

phadej:
Btw, why @23Skidoo you removed generated `Lexer.hs`?

phadej:
ah, sorry it's not removed, it just shrink a lot `-94 Lexer.hs` looked suspicious when I pulled `master`!

PR #4974, opened 2017-12-25
Issue 989 range op wild

### LINKED ISSUE #989: Context needed for ParseFailed: >= 1.2.*
I had a `build-depends` section with a trailing comma, a syntax error which is arguably hard to spot.

Cabal gave me this error:

```
cabal: <name>.cabal:<lineno>: Parse of field 'build-depends' failed.
```
1. The `<lineno>` was always the first line of the `build-depends` section, not the last (which had the syntax error).
2. There is no mention of the character that failed to parse.

Can we improve this? The logic is defined in `ParseUtils.hs`. I find the code there to be a little dense and wasn't sure how to get started.

phadej:
- No more `try`
- It recognises wild-card version after all operators,
  but fails with more descriptive warning after non-==
  Fixes #989

---

### DIFF
--- a/Cabal/Distribution/Types/VersionRange.hs
+++ b/Cabal/Distribution/Types/VersionRange.hs
@@ -445,13 +445,14 @@ instance Parsec VersionRange where
 
         -- trailing: wildcard (.y.*) or normal version (optional tags) (.y.z-tag)
         verLoop :: CabalParsing m => DList.DList Int -> m (Bool, Version)
-        verLoop acc = verLoop' acc <|> (tags >> pure (False, mkVersion (DList.toList acc)))
+        verLoop acc = verLoop' acc <|> (tags *> pure (False, mkVersion (DList.toList acc)))
 
         verLoop' :: CabalParsing m => DList.DList Int -> m (Bool, Version)
         verLoop' acc = do
             _ <- P.char '.'
-            ((\x -> verLoop (DList.snoc acc x)) =<< P.integral)
-                <|> (\_ -> (True, mkVersion (DList.toList acc))) <$> P.char '*'
+            let digit = P.integral >>= verLoop . DList.snoc acc
+            let wild  = (True, mkVersion (DList.toList acc)) <$ P.char '*'
+            digit <|> wild
 
         parens p = P.between
             ((P.char '(' P.<?> "opening paren") >> P.spaces)
--- a/Cabal/Distribution/Types/VersionRange.hs
+++ b/Cabal/Distribution/Types/VersionRange.hs
@@ -411,37 +411,61 @@ instance Parsec VersionRange where
                         return (intersectVersionRanges f t)
                      <|>
                      return f)
-        factor = P.choice
-            $ parens expr
-            : parseAnyVersion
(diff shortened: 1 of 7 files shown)


Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -411,37 +411,61 @@ instance Parsec VersionRange where
...
+                        _ -> fail $ "Unknown version operator " ++ show op
+
+        -- either wildcard or normal version
+        verOrWild :: CabalParsing m => m (Bool, Version)
+        verOrWild = do
+            x <- P.integral
+            verLoop (x :)
+
+        -- trailing: wildcard (.y.*) or normal version (optional tags) (.y.z-tag)
+        verLoop :: CabalParsing m => ([Int] -> [Int]) -> m (Bool, Version)
```
23Skidoo: May be a good idea to vendor `dlist` to improve readability in cases like this one.

phadej: Will do.

phadej: In fact, we have DList in `Compat`

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -440,17 +441,17 @@ instance Parsec VersionRange where
...
         verOrWild = do
             x <- P.integral
-            verLoop (x :)
+            verLoop (DList.singleton x)
 
         -- trailing: wildcard (.y.*) or normal version (optional tags) (.y.z-tag)
-        verLoop :: CabalParsing m => ([Int] -> [Int]) -> m (Bool, Version)
-        verLoop acc = verLoop' acc <|> (tags >> pure (False, mkVersion (acc [])))
+        verLoop :: CabalParsing m => DList.DList Int -> m (Bool, Version)
+        verLoop acc = verLoop' acc <|> (tags >> pure (False, mkVersion (DList.toList acc)))
```
23Skidoo: You can also cut down the number of parentheses by rewriting this bit as `mkVersion . DList.toList $ acc`

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -440,17 +441,17 @@ instance Parsec VersionRange where
...
 
-        verLoop' :: CabalParsing m => ([Int] -> [Int]) -> m (Bool, Version)
+        verLoop' :: CabalParsing m => DList.DList Int -> m (Bool, Version)
         verLoop' acc = do
             _ <- P.char '.'
-            ((\x -> verLoop (acc . (x :))) =<< P.integral)
-                <|> (\_ -> (True, mkVersion (acc []))) <$> P.char '*'
+            ((\x -> verLoop (DList.snoc acc x)) =<< P.integral)
```
23Skidoo: and here you can drop the lambda on the first line and use `const` on the second line.

phadej: done, used `<$`

23Skidoo:
Merged, thanks!

Issue #4914, opened 2017-11-27
Backpack: Too long paths for Windows


alexbiehl:
A project of mine which uses Backpack fails to compile on Windows: 

```
CreateDirectory "C:\\Users\\abiehl\\git\\ignite\\dist-newstyle\\build\\x86_64-windows\\ghc-8.2.1\\ignite-0.1.0.0\\l\\ignite-prelude-indef\\ignite-0.1.0.0-inplace-ignite-prelude-indef+8Vkgfyobi1D5mXxZcObkdr\\build\\ignite-0.1.0.0-inplace-ignite-prelude-indef+8Vkgfyobi1D5mXxZcObkdr": does not exist (Der Dateiname oder die Erweiterung ist zu lang.)
```
_"Der Dateiname oder die Erweiterung ist zu lang"_ is a complaint about a too long path. 

Would it be possible to compress the path a bit?

hvr:
Windows' `MAX_PATH` is the truly the gift that keeps on giving... ;-)

ezyang:
Blegh. The filepath shenanigans are not that easy to fix, but I'll try to resolve it again.

Mistuke:
If the linked PR does not fix this, can I have a small repro for this? I don't know how to make a backpack example :)

ezyang:
I don't see any reason why Backpack wouldn't have been fixed by your patch too :) I'm going to close this unless it pops up again.

hvr:
@Mistuke fwiw, see https://github.com/alexbiehl/ignite

Issue #4515, opened 2017-05-15
Can't install package: `copyFile: does not exist`


alexbiehl:
I recently switched to cabal HEAD (was prior https://github.com/haskell/cabal/commit/3b8ee8324029396298bfe588f1d56426ead216c4) and for installations of new libs I get:

```
Installing library in C:\Users\abiehl\AppData\Roaming\cabal\store\ghc-8.0.2\incoming\new-16045\Users\abiehl\AppData\Roaming\cabal\store\ghc-8.0.2\regex-compat-_-0.95.1.4-a13ab0d5ccd9fb2c4a580fc0674bee71629b38f1\lib
copyFile: does not exist (...)
```

I suspected a too-long path but the actual path is only ~190 characters long. So maybe it actually doesn't create parent directories? I am not sure what the culprit here is exactly.

The respective copy command is issued here: https://github.com/haskell/cabal/commit/3b8ee8324029396298bfe588f1d56426ead216c4#diff-d998ae67b007b4623e9c789feb2c47abR929.

Also the path seems quite redundant! Why not put the library under `C:\Users\abiehl\AppData\Roaming\cabal\store\ghc-8.0.2\incoming\new-16045` instead of `C:\Users\abiehl\AppData\Roaming\cabal\store\ghc-8.0.2\incoming\new-16045\Users\abiehl\AppData\Roaming\cabal\store\ghc-8.0.2\regex-compat-_-0.95.1.4-a13ab0d5ccd9fb2c4a580fc0674bee71629b38f1\lib`. The `new-*` directory is unique already! By dropping the redundant part we can save some precious bytes for paths.

Mistuke:
I don't see anything particularly wrong (aside from the redundant path). If you can reliably reproduce this, could you use https://technet.microsoft.com/en-us/sysinternals/bb896645 (set the filter to cabal.exe to reduce the noise) and you should be able to see the actual error from the file system.

Can you copy and paste the log for that entry here?  or a screenshot.

ezyang:
The problem is probably

```
withTempIncomingDir :: StoreDirLayout -> CompilerId
                    -> (FilePath -> IO a) -> IO a
withTempIncomingDir StoreDirLayout{storeIncomingDirectory} compid action = do
    createDirectoryIfMissing True incomingDir
    withTempDirectory silent incomingDir "new" action
  where
    incomingDir = storeIncomingDirectory compid
```

[...]

PR #4977, opened 2017-12-26
Add support on Windows for copying to long paths.

Mistuke:

Tested by trying to compile cabal itself on a long path.

Should fix partially #3972, #4914, #4515. The reason why partial is because any operation that hits the GHC I/O manager will destroy device path support. Because the GHC I/O manager is using non-native posix like APIs which are mapped to the old Win32 APIs that are bound by MAX_PATH.

This patch works because it avoids touching the I/O manager for doing the file copy and instead does a direct copy using the Win32 API. The Windows APIs do not have the same limitations as the posix ones and so do not require the file to be on the same drive before issuing a copy/move. So this should be much faster too.

### DIFF
--- a/Cabal/Distribution/Compat/CopyFile.hs
+++ b/Cabal/Distribution/Compat/CopyFile.hs
@@ -27,7 +27,7 @@ import System.IO.Error
 import System.Directory
          ( doesFileExist, renameFile, removeFile )
 import System.FilePath
-         ( isRelative, normalise, takeDirectory )
+         ( takeDirectory )
 import System.IO
          ( IOMode(ReadMode), hClose, hGetBuf, hPutBuf, hFileSize
          , withBinaryFile )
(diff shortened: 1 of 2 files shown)


Mistuke:
cc @alexbiehl , @hvr

Cabal/Distribution/Compat/CopyFile.hs
```diff
@@ -25,20 +27,36 @@ import System.IO.Error
 import System.Directory
          ( doesFileExist, renameFile, removeFile )
 import System.FilePath
-         ( takeDirectory )
```
23Skidoo: This breaks the build on non-Windows, where `takeDirectory` is still used, see e.g. https://travis-ci.org/haskell/cabal/jobs/321706530#L778.

23Skidoo:
Do you reckon it'd make sense to implement a compat filepath layer that'd use extended length paths and native operations on Windows, and standard operations (`System.Directory` and friends) otherwise?

Mistuke:
@23Skidoo Well, `System.Directory` mostly does the right thing, it does convert to device paths and calls out to `Win32` which calls the proper APIs. It's specifically `copyFile` that's problematic because of the limitations of the posix calls.

I only quickly glanced but I don't believe the other functions use the I/O manager and so should be fine. A `"proper"` fix would be a new I/O manager in GHC/base. I'm working towards an implementation, but it's a big task so not sure I'll finish by `8.6`. But somewhere in the next two releases it should be done and then Cabal and cabal-install should just work without any changes.

23Skidoo:
Build still fails on Linux, now due to `-Werror`:

```
Distribution/Compat/CopyFile.hs:145:1:
    Warning: Defined but not used: `toExtendedLengthPath'
```

23Skidoo:
Merged, thanks!

BardurArantsson:
Great work! Especially the bit about trying to upstream into base... :)

(Not that I'm using Windows, personally, but this has been a pain for Cabal users and package developers who want "cross-platform-by-default" for quite a while...)

Mistuke:
For future reference, this workaround is no longer need when `cabal` is compiled with GHC 8.6. See https://github.com/ghc/ghc/commit/4de585a5c1ac3edc2914cebcac1753b514051a89

PR #4978, opened 2017-12-26
Fix path redundancy new-build output folders.

### LINKED ISSUE #4515: Can't install package: `copyFile: does not exist`

Mistuke:

Tested it by compiling Cabal itself using the Makefile on a long path.

Fixes #4515, #3972.

The `setup` call with `copyCommand` doesn't do what the comment in `ProjectBuilding.hs` suggested. It literally only assigns the path you specified. So you can choose any format you want as long as you pass it to `copyCommand`. So instead of using `@$tmpDir/$prefix@` we now use `@$tmpDir/$packageName-$version@`. Because `prefix` is a prefix to `tmpDir` this means that before this patch cabal was halving the Windows MAX_PATH. Effectively meaning `new-build` would only work on paths shorter than 125 characters.

The consistency is maintained with `absoluteInstallDirs` by also taking the `base name` instead of dropping the `drive` when combing the paths. Proof that this is all consistent is that the build succeeds. Cabal only issues `createDirectory` calls in one place. So if the paths did not end up being the same the build would fail because it can't find the specified path.

### DIFF
--- a/Cabal/Distribution/Simple/InstallDirs.hs
+++ b/Cabal/Distribution/Simple/InstallDirs.hs
@@ -59,7 +59,7 @@ import Distribution.Text
 import System.Directory (getAppUserDataDirectory)
 import System.FilePath
   ( (</>), isPathSeparator
-  , pathSeparator, dropDrive )
+  , pathSeparator, takeBaseName )
 
 #ifdef mingw32_HOST_OS
 import qualified Prelude
@@ -286,7 +286,7 @@ absoluteInstallDirs :: PackageIdentifier
                     -> InstallDirs FilePath
 absoluteInstallDirs pkgId libname compilerId copydest platform dirs =
     (case copydest of
-       CopyTo destdir -> fmap ((destdir </>) . dropDrive)
+       CopyTo destdir -> fmap ((destdir </>) . takeBaseName)
        _              -> id)
   . appendSubdirs (</>)
   . fmap fromPathTemplate
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -940,12 +940,10 @@ buildAndInstallUnpackedPackage verbosity
     annotateFailure mlogFile InstallFailed $ do
 
       let copyPkgFiles tmpDir = do
-            setup Cabal.copyCommand (copyFlags tmpDir)
-            -- Note that the copy command has put the files into
-            -- @$tmpDir/$prefix@ so we need to return this dir so
-            -- the store knows which dir will be the final store entry.
-            let prefix   = dropDrive (InstallDirs.prefix (elabInstallDirs pkg))
+            let prefix   = let path = InstallDirs.prefix (elabInstallDirs pkg)
+                           in takeBaseName path
                 entryDir = tmpDir </> prefix
+            setup Cabal.copyCommand (copyFlags entryDir)
             LBS.writeFile
               (entryDir </> "cabal-hash.txt")
               (renderPackageHashInputs (packageHashInputs pkgshared pkg))
@@ -954,7 +952,7 @@ buildAndInstallUnpackedPackage verbosity
             -- While this breaks the prefix-relocatable property of the lirbaries
             -- it is necessary on macOS to stay under the load command limit of the
             -- macOS mach-o linker. See also @PackageHash.hashedInstalledPackageIdVeryShort@.
-            otherFiles <- filter (not . isPrefixOf entryDir) <$> listFilesRecursive tmpDir 
+            otherFiles <- filter (not . isPrefixOf entryDir) <$> listFilesRecursive tmpDir
             -- here's where we could keep track of the installed files ourselves
             -- if we wanted to by making a manifest of the files in the tmp dir
             return (entryDir, otherFiles)
--- a/cabal-install/Distribution/Client/Store.hs
+++ b/cabal-install/Distribution/Client/Store.hs
@@ -209,7 +209,10 @@ newStoreEntry verbosity storeDirLayout@StoreDirLayout{..}
             -- Atomically rename the temp dir to the final store entry location.
             renameDirectory incomingEntryDir finalEntryDir
             forM_ otherFiles $ \file -> do
-              let finalStoreFile = storeDirectory compid </> makeRelative (incomingTmpDir </> (dropDrive (storeDirectory compid))) file
+              let finalStoreFile = storeDirectory compid
+                                </> makeRelative (incomingTmpDir
+                                     </> (takeBaseName (storeDirectory compid)))
+                                     file
               createDirectoryIfMissing True (takeDirectory finalStoreFile)
               renameFile file finalStoreFile
 
--- a/cabal-install/changelog
+++ b/cabal-install/changelog
@@ -46,6 +46,8 @@
 	* Added support for '--enable-tests' and '--enable-benchmarks' to
 	'cabal fetch' (#4948).
 	* Removed support for building cabal-install with GHC < 7.10.
+	* Redundancy in new-build incoming store paths removed. This makes incoming
+	tmp paths much shorter. (#4515, #3972)
 
 2.0.0.1 Mikhail Glushenkov <mikhail.glushenkov@gmail.com> December 2017
 	* Support for GHC's numeric -g debug levels (#4673).


Mistuke:
cc @alexbiehl , @hvr

23Skidoo:
Awesome, I was just bitten by this the other day when trying to build `cabal-install` binaries on Windows.

23Skidoo:
Looks like there's a test failure on Linux.

/cc @dcoutts

Mistuke:
Hmm indeed, it seems to be skipped on a lot of configuration. But the error is a bit odd..

```
Main.hs:1:8:
    Could not find module `RemotePkg'
    There are files missing in the `remote-pkg-2.0@remot_0O9PJ1ccB6bKshvGETwlge' package,
    try running 'ghc-pkg check'.
```
But the log shows it registering the package.. Is there a way for me to run this test locally on Windows?

23Skidoo:
@Mistuke See `cabal-testsuite/README.md`. Though it looks like that test is skipped on Windows, which is why the AppVeyor build succeeded.

Mistuke:
@23Skidoo Thanks, I'll boot to Linux and give it a try.

23Skidoo:
Some clues to why that test is failing:

```
$ ghc-pkg --package-db [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/package.db list 
[...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/package.db
    remote-pkg-2.0
    remote-setup-dep-3.0

$ ghc-pkg --package-db [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/package.db check
[...]
There are problems in package remote-pkg-2.0:
  Warning: library-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib doesn't exist or isn't a directory
  Warning: dynamic-library-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib doesn't exist or isn't a directory
  Warning: haddock-interfaces: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/share/doc/html/remote-pkg.haddock doesn't exist or isn't a file
  Warning: haddock-html: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/share/doc/html doesn't exist or isn't a directory
  import-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib doesn't exist or isn't a directory
  cannot find any of ["RemotePkg.hi","RemotePkg.p_hi","RemotePkg.dyn_hi"]
  cannot find any of ["libHSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced.a","libHSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced.p_a","libHSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced-ghc8.2.2.so","libHSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced-ghc8.2.2.dylib","HSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced-ghc8.2.2.dll"] on library path
[...]

The following packages are broken, either because they have a problem
listed above, or because they depend on a broken package.
remote-pkg-2.0

$ ghc-pkg --package-db [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/package.db describe remote-pkg-2.0           
name: remote-pkg
version: 2.0
id: remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced
key: remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced
license: UnspecifiedLicense
exposed: True
indefinite: False
exposed-modules:
    RemotePkg
abi: aec19496483f5a6f8e3d43d06c33eb16
trusted: False
import-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib
library-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib
dynamic-library-dirs: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/lib
data-dir: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/share
hs-libraries: HSremote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced
depends:
    base-4.10.1.0
abi-depends: base-4.10.1.0=35a7f6be752ee4f7385cb5bf28677879
haddock-interfaces: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/share/doc/html/remote-pkg.haddock
haddock-html: [...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/share/doc/html
pkgroot: "[...]/cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/store/ghc-8.2.2"

$ cd cabal-testsuite/PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.dist/home/.cabal/        

$ tree
.
├── config
├── packages
│   └── test-local-repo
│       ├── 01-index.cache
│       ├── 01-index.tar
│       ├── 01-index.tar.gz
│       ├── 01-index.tar.idx
│       ├── 01-index.timestamp
│       ├── mirrors.json
│       ├── remote-pkg
│       │   └── 2.0
│       │       └── remote-pkg-2.0.tar.gz
│       ├── remote-setup-dep
│       │   └── 3.0
│       │       └── remote-setup-dep-3.0.tar.gz
│       ├── root.json
│       ├── snapshot.json
│       └── timestamp.json
└── store
    └── ghc-8.2.2
        ├── incoming
        │   ├── remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced.lock
        │   └── remote-setup-dep-3.0-95e21c2632e40b32212bf1add9fcfe5272e685432265d9d7cc77221afb0d4286.lock
        ├── package.db
        │   ├── package.cache
        │   ├── package.cache.lock
        │   ├── remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced.conf
        │   └── remote-setup-dep-3.0-95e21c2632e40b32212bf1add9fcfe5272e685432265d9d7cc77221afb0d4286.conf
        ├── remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced
        │   ├── cabal-hash.txt
        │   └── home
        │       └── refold
        │           └── code
        │               └── haskell
        │                   └── cabal
        │                       └── cabal-testsuite
        │                           └── PackageTests
        │                               └── NewBuild
        │                                   └── CustomSetup
        │                                       └── RemotePackageWithCustomSetup
        │                                           └── build-package-from-repo-with-custom-setup.dist
        │                                               └── home
        └── remote-setup-dep-3.0-95e21c2632e40b32212bf1add9fcfe5272e685432265d9d7cc77221afb0d4286
            ├── cabal-hash.txt
            └── lib
                ├── libHSremote-setup-dep-3.0-95e21c2632e40b32212bf1add9fcfe5272e685432265d9d7cc77221afb0d4286.a
                ├── libHSremote-setup-dep-3.0-95e21c2632e40b32212bf1add9fcfe5272e685432265d9d7cc77221afb0d4286-ghc8.2.2.so
                ├── RemoteSetupDep.dyn_hi
                └── RemoteSetupDep.hi

25 directories, 24 files

```

So `remote-pkg-2.0` gets registered, but instead of `lib` and `share` under `store/remote-pkg-2.0-49b4ea173d672c8399dff7218dd2512ab818030cc9fab93eca633ac6d5885ced/` we get some crazy thing.

Cabal/Distribution/Simple/InstallDirs.hs
```diff
@@ -286,7 +286,7 @@ absoluteInstallDirs :: PackageIdentifier
                     -> InstallDirs FilePath
 absoluteInstallDirs pkgId libname compilerId copydest platform dirs =
     (case copydest of
-       CopyTo destdir -> fmap ((destdir </>) . dropDrive)
+       CopyTo destdir -> fmap ((destdir </>) . takeBaseName)
```
23Skidoo: I don't think this is right. Here you take `InstallDirs` and substitute all absolute paths by `destdir </> basename absPath`. E.g. if `libdir` was `$prefix/lib/$abi/$libname`, you make it `$destdir/$libname`.

23Skidoo: Though it seems to me that it's underspecified what `copy --destdir` should mean. I'd expect the value of `--destdir` to replace `$prefix`, but apparently it just gets prepended to the full absolute path including `$prefix` minus the drive letter.

23Skidoo: Looks like there was a `--copy-prefix` feature that actually worked that way, but it was removed ages ago (981473b890912d1dbed58082190723ad6db97501).

Mistuke: Hmm thanks! that's a great hint. I've also managed to reproduce it on Linux so hopefully will have a fix today.

Mistuke:
Haven't  forgotten this, just taking some time to really understand how cabal constructs all it's paths.

Mistuke:
Hmm so I figured out what's wrong and I fixed the paths, but now cabal seems to think the package has already been registered and skips the registration. It does so because the folder already exists, so It assumes it already registered it. Have to take a closer look at that logic now.

In the mean time, I've also written patches for GHC 8.6 which will fix long path support in the I/O manager.

hvr:
> It does so because the folder already exists, so It assumes it already registered it.

did you weaken the transactional assumption that the folder is staged in a different place, and moved atomically via a single syscall into its target space at once?

Mistuke:
> did you weaken the transactional assumption that the folder is staged/constructed in a different place, and moved atomically via a single syscall into its target location at once?

Hmm maybe, I'll take a look tomorrow, but it shouldn't have though. I'm still honouring any `CopyTo` call, I'm just doing it `before` it flattens the structure by applying `prefix` to the existing path. This is where the duplication comes from.

gbaz:
Is this still something worth pursuing given the other longpath support?

Mistuke:
I believe it is, since the long paths support in ghc and toolchain doesn't extend to any random program that may be called by cabal. For instance in a custom preprocessor

Mikolaj:
In that case, what can we do to help rescue this PR?

Mistuke:
I'm not sure how easily this still applies to current master. At the time I was having a hard time tracking down why after the change the folder would be created earlier. So it created a race condition. But I have limited understanding of Cabal's overal inner working for new-build so it lingered..

I guess the first question is if the code can be rebased or if it needs to be rewritten

Mikolaj:
@mergifyIO rebase

Mikolaj:
Well, at least it didn't rebase automatically. :(

Volunteers kindly welcome.

Kleidukos:

ffaf1:

Issue #4957, opened 2017-12-17
Distribution.Version QuickCheck test failure


grayjay:
I noticed a failure on AppVeyor:

https://ci.appveyor.com/project/23Skidoo/cabal/build/%234373%20(master)

```
Test suite unit-tests: RUNNING...
File modification time resolution calibration completed, maximum delay
observed: 4.0 ms. Will be using delay of 20.0 for test runs.
Unit Tests
  Distribution.Version
    Range Property 19 (VersionRange -> Property):                                         FAIL
      *** Failed! Falsifiable (after 6 tests and 2 shrinks): 
      UnionVersionRanges (VersionRangeParens (ThisVersion (mkVersion [1]))) (EarlierVersion (mkVersion [1]))
      OrEarlierVersion (mkVersion [1]) /= UnionVersionRanges (ThisVersion (mkVersion [1])) (EarlierVersion (mkVersion [1]))
      Use --quickcheck-replay=833531 to reproduce.
1 out of 98 tests failed (3.11s)
Test suite unit-tests: FAIL
```

phadej:
I'll check, property might be bogus

PR #4979, opened 2017-12-27
Add showFullInstalledPackageInfo

phadej:
@hvr I chose to go more BC way, I'm afraid we will out of run to test ghc-pkg changes properly.
`showFullInstalledPackageInfo` could be removed for next Cabal / GHC-8.6 when we have better time.

---

showInstalledPackageInfo uses full variant, but with pkgroot set to
Nothing. This way we field descriptions have roundtrip property (wo it's
simpler to proceed with reworking the parsing using parsec framework).

We introduce new name, so no (coordinated) changes to ghc-pkg are needed.

### DIFF
--- a/Cabal/Distribution/InstalledPackageInfo.hs
+++ b/Cabal/Distribution/InstalledPackageInfo.hs
@@ -41,6 +41,7 @@ module Distribution.InstalledPackageInfo (
         emptyInstalledPackageInfo,
         parseInstalledPackageInfo,
         showInstalledPackageInfo,
+        showFullInstalledPackageInfo,
         showInstalledPackageInfoField,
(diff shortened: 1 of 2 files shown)


Cabal/Distribution/InstalledPackageInfo.hs
```diff
@@ -359,8 +360,16 @@ parseInstalledPackageInfo =
 -- -----------------------------------------------------------------------------
 -- Pretty-printing
 
+-- | Pretty print 'InstalledPackageInfo'.
+--
```
23Skidoo: s/it' s/it/

23Skidoo:
Test failure is #4957.

23Skidoo:
Merged, thanks!

PR #4991, opened 2018-01-01
Fix #4986, and extra refactoring

### LINKED ISSUE #4986: Duplicate build-tools-depends/build-depends chokes new-build
This happened in practice with pretty-show (https://hackage.haskell.org/package/pretty-show-1.6.15/pretty-show.cabal)

Manifested itself as #4621

I am actually not sure if `getAllToolDependencies` should merge tool dependencies or not. Maybe not, in which case this patch is sufficient:

```
diff --git a/Cabal/Distribution/Backpack/ConfiguredComponent.hs b/Cabal/Distribution/Backpack/ConfiguredComponent.hs
index 1d929636f..9e095a39f 100644
--- a/Cabal/Distribution/Backpack/ConfiguredComponent.hs
+++ b/Cabal/Distribution/Backpack/ConfiguredComponent.hs
@@ -40,6 +40,7 @@ import Distribution.Simple.LocalBuildInfo
 import Distribution.Version
 import Distribution.Utils.LogProgress
 import Distribution.Utils.MapAccum
+import Distribution.Utils.Generic
 
 import Control.Monad
 import qualified Data.Set as Set
@@ -191,7 +192,7 @@ toConfiguredComponent pkg_descr this_cid dep_map component = do
                          , pn /= packageName pkg_descr
                          , (cn, e) <- Map.toList comp_map
                          , cn == CLibName ]
-    exe_deps =
+    exe_deps = ordNub $
         [ exe
         | ExeDependency pn cn _ <- getAllToolDependencies pkg_descr bi
         -- The error suppression here is important, because in general
diff --git a/Cabal/Distribution/Types/AnnotatedId.hs b/Cabal/Distribution/Types/AnnotatedId.hs
index cc5991a3c..a424335fd 100644
--- a/Cabal/Distribution/Types/AnnotatedId.hs
+++ b/Cabal/Distribution/Types/AnnotatedId.hs
@@ -16,7 +16,7 @@ data AnnotatedId id = AnnotatedId {
         ann_cname :: ComponentName,
         ann_id    :: id
     }
-    deriving (Show)
+    deriving (Show, Eq, Ord)
 
 instance Package (AnnotatedId id) where
     packageId = ann_pid
```

This might help unblock someone. I've got a bigger patch coming with (1) lots of new debugging Pretty instances, (2) more asserts and (3) a regression test.

ezyang:

Added test for the bugfix

### DIFF
--- a/Cabal/Distribution/Backpack/ConfiguredComponent.hs
+++ b/Cabal/Distribution/Backpack/ConfiguredComponent.hs
@@ -40,6 +40,7 @@ import Distribution.Simple.LocalBuildInfo
 import Distribution.Version
 import Distribution.Utils.LogProgress
 import Distribution.Utils.MapAccum
+import Distribution.Utils.Generic
 
 import Control.Monad
 import qualified Data.Set as Set
@@ -191,7 +192,11 @@ toConfiguredComponent pkg_descr this_cid dep_map component = do
                          , pn /= packageName pkg_descr
                          , (cn, e) <- Map.toList comp_map
                          , cn == CLibName ]
-    exe_deps =
+    -- We have to nub here, because 'getAllToolDependencies' may return
+    -- duplicates (see #4986).  (NB: This is not needed for lib_deps,
+    -- since those elaborate into includes, for which there explicitly
+    -- may be multiple instances of a package)
+    exe_deps = ordNub $
         [ exe
         | ExeDependency pn cn _ <- getAllToolDependencies pkg_descr bi
         -- The error suppression here is important, because in general
--- a/Cabal/Distribution/Types/AnnotatedId.hs
+++ b/Cabal/Distribution/Types/AnnotatedId.hs
@@ -11,12 +11,21 @@ import Distribution.Types.ComponentName
 -- | An 'AnnotatedId' is a 'ComponentId', 'UnitId', etc.
 -- which is annotated with some other useful information
 -- that is useful for printing to users, etc.
+--
+-- Invariant: if ann_id x == ann_id y, then ann_pid x == ann_pid y
+-- and ann_cname x == ann_cname y
 data AnnotatedId id = AnnotatedId {
         ann_pid   :: PackageId,
         ann_cname :: ComponentName,
         ann_id    :: id
     }
     deriving (Show)
+
+instance Eq id => Eq (AnnotatedId id) where
+    x == y = ann_id x == ann_id y
+
+instance Ord id => Ord (AnnotatedId id) where
+    compare x y = compare (ann_id x) (ann_id y)
 
 instance Package (AnnotatedId id) where
     packageId = ann_pid
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -126,6 +126,7 @@ import           Distribution.Simple.LocalBuildInfo
                    ( ComponentName(..), pkgComponents )
 import qualified Distribution.Simple.Setup as Setup
 import           Distribution.Simple.Command (commandShowOptions)
+import           Distribution.Simple.Configure (computeEffectiveProfiling)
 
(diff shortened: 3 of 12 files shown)


phadej:
I have uneasy feeling about re-using `Pretty` for this. The idea is to use `Pretty` to display data to user. For debugging we could use `Show` but it's not *pretty*. I propose adding `PShow` class, and define `show = pshow`. I can make a PR soon. This can go in as is, but I **will** revert `Pretty` related commits

ezyang:
@phadej Show is not all that appropriate either, because conventionally the Show instance of a datatype is intended to be valid Haskell, and I don't necessarily want to force Pretty-printed variants of types to be valid Haskell; a very simple case where this would be extremely annoying are newtypes of Text, where barfing out the constructor name would be a big problem for readability. So I'm :-1: on PShow, but :+1: for Outputable. I guess I'll add that class...

phadej:
Given `IsString`, PShow could be a little "reality twister". My doubt is more about inconsistencies, how `Maybe InstalledPackageInfo` or `[PackageDescription]` would be debug output? Cabal-like syntax is compact, but it doesn't *compose* well. So I'm :+1: for Haskell-like record syntax for those.

I'd recommend splitting this PR into bug fix, which can go in (with old style asserts) and then we can bikeshed on Outputtable/PShow/Pretty without hurry (that's not critical to get into 2.2, is it?)

phadej:
FWIW PShow can have Generic implementation, taking into account newtypes and records!

A note about composability: parens!

BardurArantsson:
Seems like outputting somethink S-expression-like would be pretty trivial and sufficient for debugging?

ezyang:
I force-pushed the reduced patch set, without updated asserts or pretty-printing.

> Given `IsString`, PShow could be a little "reality twister".

I don't know what you mean by this. Are you suggesting that a PackageName can acceptably render as "pkgname", because -XOverloadedStrings can interpret this as a PackageName?

But there are still situations where Haskell syntax is unacceptably verbose, and there is no way to "hack" it with OverloadedStrings. One example that came up in the patchset:

```
-- | A ConfiguredId is a package ID for a configured package.
--
-- Once we configure a source package we know its UnitId. It is still
-- however useful in lots of places to also know the source ID for the package.
-- We therefore bundle the two.
--
-- An already installed package of course is also "configured" (all its
-- configuration parameters and dependencies have been specified).
data ConfiguredId = ConfiguredId {
    confSrcId  :: PackageId
  , confCompName :: Maybe ComponentName
  , confInstId :: ComponentId
  }
  deriving (Eq, Ord, Generic)
```

By default, I'd like ConfiguredId to render as just its ComponentId, since that canonically identifies a ConfiguredId. However, if I am debugging corrupt ConfiguredId metadata, I might also like to display the confSrcId, confCompName. In GHC, the way we do this is we render `x` by default, but `x{some metadata}` if you turn on `-dppr-debug`. I think this behavior is quite useful and I'd like it to apply to Cabal too.

> Cabal-like syntax is compact, but it doesn't *compose* well.

With pretty-printing, this is not true; indentation is all you need. To take your example of `[PackageDescription]`, with pretty-printing it with the default list renderer will cause it to look something like this:

```
[name: foo
 field: 1
 field: 2,
 name: bar
 field: 3
 field: 4]
```

Which is basically as good as you're going to get with Haskell-style list syntax with multi-line elements. Or you be a bit more clever and have a different format:

```
- name: foo
  field: 1
  field: 2
- name: bar
  field: 3
  field: 4
```

Indentation is great! And it is much easier to parse than parentheses.

P.S. Also, why do I care so much about conciseness? Because if I'm debugging I'm going to have to read all of this information, and it is far easier to understand what is going on if a single ElaboratedConfiguredPackage can fit in a few lines, rather than multiple pages.

23Skidoo:
Merged, thanks!

23Skidoo:
Cherry-picked the first four commits into 2.0.

PR #4993, opened 2018-01-03
Outputable typeclass

ezyang:
This PR adds a GHC-style Outputable type class for debug/user-visible pretty printing, and rewrites everything to use it. It also adds `pprAssert` and rewrites/adds some asserts for it.

Stacked on top of #4991

### DIFF
--- a/Cabal/Distribution/Outputable.hs
+++ b/Cabal/Distribution/Outputable.hs
@@ -1,0 +1,313 @@
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+module Distribution.Outputable (
+    -- * Type class
+    Outputable(..),
+
+    -- * Pretty printing combinators
+    SDoc,
+    showSDoc,
+    showPpr,
+    docToSDoc,
+    sdocWithVerbosity,
+
+    -- No empty to avoid conflict with Applicative
+    nest,
+    char,
+    text,
+    int, integer, float, double, rational,
+    parens, brackets, braces, quotes,
+    doubleQuotes, angleBrackets, paBrackets,
+    semi, comma, colon, space, equals, dot, vbar,
+    lparen, rparen, lbrack, rbrack, lbrace, rbrace, underscore,
+    blankLine,
+    (<>), (<+>), hcat, hsep,
+    ($$), ($+$), vcat,
+    sep, cat,
+    fsep, fcat,
+    hang, punctuate, ppWhen, ppUnless,
+) where
+
+import Distribution.Verbosity
+import Distribution.Pretty
+import Distribution.Compat.Semigroup
+
+import Data.Int
+import Data.Word
+import Data.Set (Set)
+import Data.Map (Map)
+import Data.IntMap (IntMap)
+import qualified Data.Map as Map
+import qualified Data.IntMap as IntMap
+import qualified Data.Set as Set
+
+import Text.PrettyPrint (Doc)
+import qualified Text.PrettyPrint as Pretty
+
+newtype SDoc = SDoc { runSDoc :: SDocContext -> Doc }
+
+instance Semigroup SDoc where
+    (<>) = (<<>>)
+
+instance Monoid SDoc where
+    mempty = empty
+    mappend = (<>)
+
+showSDoc :: Verbosity -> SDoc -> String
+showSDoc verbosity sdoc = Pretty.render $ runSDoc sdoc (SDC verbosity)
+
+showPpr :: Outputable a => Verbosity -> a -> String
+showPpr verbosity = showSDoc verbosity . ppr
+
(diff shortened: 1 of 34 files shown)


phadej:
I don't like Outputtable. We have Pretty, Outputtable and Show. Two last ones seems to be similar in their use case: development & debugging aid.

Note: I personally dislike there aren't Show instances in GHC, that makes hacking on GHC very different from hacking on everything else. I hope we don't get anything circular in Cabal so Show won't work anymore

Cabal/Distribution/Outputable.hs
```diff
@@ -0,0 +1,313 @@
...
+                     go d (e:es) = (d <> p) : go e es
+
+ppWhen, ppUnless :: Bool -> SDoc -> SDoc
+ppWhen True  doc = doc
+ppWhen False _   = empty
+
+ppUnless True  _   = empty
+ppUnless False doc = doc
+
+-- | Debug-printable entity.
```
23Skidoo: The comment here should mention how this class relates to `Pretty` and `Show` and when it's right to use it.

23Skidoo:
CI failures seem to be due to `-Werror` and the like.

ezyang:
> Haddock. What S stands for?

Name is borrowed from GHC. I did some sleuthing, and in GHC the S stands for "style", in the sense that it represents `Style -> Doc` (where `Style` is an abstract type that represents parameters about how to style the output.) Which is appropriate here too, though the only Style we have at the moment is Verbosity.

ezyang:
> Please name the member output. Note: IMHO GHC is not a good codebade to take examples from, especially on how to name things!

output is not a very good name, it is too likely to conflict with local variable definitions. I really like `ppr` and would rather rename the type class to `Ppr` (ew?) rather than rename `ppr` to `output`.

ezyang:
(Also some commentary at https://github.com/haskell/cabal/pull/4991#issuecomment-354925595 )

So, if I understand correctly, the competing proposal to "Make a new type class, and hand implement the instances" is "Use Generic to mechanically derive the bulk of debugging instances."  So, there is one more reason I don't want to use Generic in a 'PShow' style, is that  pretty-show already exists, so there isn't really a need for another PShow type class, you can just feed your (GHC automatically derived) Show output to it, if that's your kettle of tea. The whole point of Outputable is to let humans make aesthetic decisions about what information to show by default or not, and you can't do that with an automatically derived instance.

phadej:
> output is not a very good name, it is too likely to conflict with local variable definitions. I really like ppr and would rather rename the type class to Ppr (ew?) rather than rename ppr to output.

`Ppr` is IMO even worse. Why not `DebugDoc + debugDoc`?

phadej:
> So, if I understand correctly, the competing proposal to "Make a new type class, and hand implement the instances" is "Use Generic to mechanically derive the bulk of debugging instances." So, there is one more reason I don't want to use Generic in a 'PShow' style, is that pretty-show already exists, so there isn't really a need for another PShow type class, you can just feed your (GHC automatically derived) Show output to it, if that's your kettle of tea. The whole point of Outputable is to let humans make aesthetic decisions about what information to show by default or not, and you can't do that with an automatically derived instance.

---

Yes: *"Use Generic to mechanically derive the bulk of debugging instances.*.

The problem with `pretty-show` is that it cannot parse all of `Show` instances! (E.g. `UTCTime`!). FWIW, that's how `tree-diff` works, it outputs Haskell-like expr, which it can parse back. See e.g. https://github.com/haskell/cabal/pull/4788/files#diff-42ad2ddebf1a7e337c44d4bb3ad01cfb

I really disagree that *aesthetic decisions* should affect **debug** output. If the assert fails (which happens rarely outside the development), then there should be all information available dumped to the users screen, so it can be pasted to the bug report (not a bad idea if it can be pasted as is!)

And yes, I'd use *a tool* to make sense of that output. E.g. making things like http://hackage.haskell.org/package/dump-core would be easier (no need to do a plugin), if GHC could dump Core (also) in a machine readable format.

23Skidoo:
> I really disagree that aesthetic decisions should affect debug output. If the assert fails (which happens rarely outside the development), then there should be all information available dumped to the users screen, so it can be pasted to the bug report (not a bad idea if it can be pasted as is!)

IIUC, `Outputable`/`DebugPpr`/whatever is for making debug logs more readable. For example, this patch adds install plan logging during various stages of `ProjectConfiguration`. Being able to change the level of install plan logging detail during debugging sounds useful. We may later want to add a `--debug-verbosity` option to make this easier.

Asserts enabled in production can either use `Show` or higher verbosity by default.

23Skidoo:
This PR seems to be bitrotting. What can be done to move it forward?

ezyang:
Resolve philosophical difference between me and @phadej :)

TBH it's probably not a big deal if this isn't merged, because I really only needed it for debugging. I (or maybe someone else) will probably dust it off the next time such debugging is needed...

PR #4997, opened 2018-01-04
Add cabal-doctest and idris Setup.hs as smoke tests for BC

phadej:

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -496,6 +496,7 @@ test-suite custom-setup-tests
     directory,
     filepath,
     process
+  default-language: Haskell2010
 
 test-suite parser-hackage-tests
   type: exitcode-stdio-1.0
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -483,6 +483,20 @@ test-suite check-tests
   ghc-options: -Wall
   default-language: Haskell2010
 
+test-suite custom-setup-tests
+  type: exitcode-stdio-1.0
+  hs-source-dirs: tests/custom-setup
+  main-is: CustomSetupTests.hs
+  other-modules:
+    CabalDoctestSetup
+    IdrisSetup
+  build-depends:
+    Cabal,
+    base,
+    directory,
+    filepath,
+    process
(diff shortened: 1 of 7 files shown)


23Skidoo:
Good idea!

phadej:
Ah, idris script breaks on windows too.

Note: I'd  add Agda's Setup.hs too, seems that its `Setup.hs` is under MIT license. Is my interpretation of https://hackage.haskell.org/package/Agda-2.5.3/src/LICENSE correct, @hvr 

What other non-trivial custom setups we should be aware off?

I answer above:  `ghc-paths` - I'll add that one a bit later today.

Cabal/Cabal.cabal
```diff
@@ -483,6 +483,20 @@ test-suite check-tests
   ghc-options: -Wall
   default-language: Haskell2010
 
+test-suite custom-setup-tests
```
23Skidoo: You also need `default-language: Haskell2010` here, that's why CI is red.

23Skidoo:
Feel free to merge once it's green.

Cabal/Cabal.cabal
23Skidoo: Oh kewl, looks like I can push to this branch myself.

23Skidoo:
> What other non-trivial custom setups we should be aware off?

Maybe something from Kmett-verse?

phadej:
cabal-doctest is used in a far share of Kmett-verse. I'll check if `lens` copies doc image manually still

23Skidoo:
There's a failure on 7.4.2, because `IdrisSetup` doesn't support versions that old.

23Skidoo:
Also `gen-extra-source-files` apparently doesn't take `hs-source-dirs` into account.

23Skidoo:
See also https://github.com/23Skidoo/all-custom-setups.

23Skidoo:
@phadej I pushed some patches, it's green now (except one macOS timeout).

phadej:
Great! let's merge and add more later.

23Skidoo:
Merged, thanks!

Issue #4995, opened 2018-01-03
Latest Cabal fails to update or install


melston:
I just updated my Haskell Platform to 8.2.2 on my Windows 7 laptop.  But now I can't make use of cabal.

```
D:\>cabal install time
dieVerbatim: user error (cabal: 'C:\cygwin64\bin\curl.exe' exited with an error:
)

D:\>cabal update
dieVerbatim: user error (cabal: 'C:\cygwin64\bin\curl.exe' exited with an error:
)

D:\>cabal --version
cabal-install version 2.0.0.1
compiled using version 2.0.1.1 of the Cabal library
```

This was working in a previous version of the Haskell Platform (I think it was 8.0.1)

Is there anything I can do to correct this?

melston:
I just found a 'workaround' though I am not sure if it is the correct solution or not.

I added <Haskell Platform>\msys\usr\bin to the path ahead of my cygwin installation and now cabal seems to work.

If cabal needs specific applications (and provides them) then I would have thought it would have been able to find those applications on its own without mucking in the user's environment.

gbaz:
Cabal neither needs or provides `curl`. It just tries to use the installed one if it finds it lying around. For some reason calling the cygwin one seems to fail from it? 

You can find other choices for HttpTransport that can be configured by looking for the documentation with `cabal help`. 

One thing we could perhaps do is give a better error when shelling out to `curl` or `wget` fails...

23Skidoo:
Seems like trying to use anything provided by Cygwin results in fail. Cygwin's `sh` is another major offender. Would be nice to have some kind of detection of these cases.

gbaz:
This is why we have the three lines we have in the install instructions for the platform to add certain paths to cabal's lookup directory. (now that I think about that, that would fix this too).

    extra-prog-path: C:\Program Files\Haskell Platform\8.2.2\msys\usr\bin
    extra-lib-dirs: C:\Program Files\Haskell Platform\8.2.2\mingw\lib
    extra-include-dirs: C:\Program Files\Haskell Platform\8.2.2\mingw\include

It would be really nice (since this comes up a lot, in that people miss adding these) if there could be some more-automatic way of handling adding those lines or the equivalent to the config file, or otherwise providing the equivalent functionality.

With them in place, people can have cygwin or whatever on their path and it doesn't matter to cabal.

23Skidoo:
Yep, we should do that, no-one reads the release notes.

ezyang:
[...]

PR #4999, opened 2018-01-06
Pass extra dirs to http transport

gbaz:

This should resolve half of #4995 -- as usual, two lines of straightforward changes, and 20x as many lines to make sure the data gets threaded to where it needs to go. In this case I've added an extra prog path field to the GlobalFlags object so it can thread through to be used in looking up binaries for the http transport. It is automatically derived from the setting in the config file.

### DIFF
--- a/cabal-install/Distribution/Client/GlobalFlags.hs
+++ b/cabal-install/Distribution/Client/GlobalFlags.hs
@@ -70,7 +70,7 @@ data GlobalFlags = GlobalFlags {
     globalHttpTransport     :: Flag String,
     globalNix               :: Flag Bool,  -- ^ Integrate with Nix
     globalStoreDir          :: Flag FilePath,
-    globalProgPathExtra     :: NubList FilePath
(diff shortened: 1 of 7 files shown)


23Skidoo:
Looks like this breaks some tests.

cabal-install/Distribution/Client/GlobalFlags.hs
```diff
@@ -69,7 +69,8 @@ data GlobalFlags = GlobalFlags {
     globalIgnoreExpiry      :: Flag Bool,    -- ^ Ignore security expiry dates
     globalHttpTransport     :: Flag String,
     globalNix               :: Flag Bool,  -- ^ Integrate with Nix
-    globalStoreDir          :: Flag FilePath
+    globalStoreDir          :: Flag FilePath,
+    globalProgPathExtra     :: NubList FilePath
```
23Skidoo: Please add a `-- ^ Haddock comment` here.

PR #5003, opened 2018-01-06
Remove `UnknownBuildType` from `BuildType`

hvr:
This is a follow-up to #4958 which opened up the opportunity to do
this make-illegal-states-unrepresentable refactoring as well.


23Skidoo:
Why did we have it in the first place?

23Skidoo:
Merged, thanks!

hvr:
@23Skidoo I think the original reasons can be inferred from the original comment attached to the removed constructor: basically defer error handling to `cabal check`. This likely has to do with the previous paradigm of parsing future .cabal versions (which #4899 gives up on) which could in theory (but given history imho rather unlikely) add new build-types.

PR #5004, opened 2018-01-07
Disable per-component build when program coverage enabled

23Skidoo:
Rebased version of #4902 that skips the slow test.


23Skidoo:
Actually, I have a better idea: skip the slow test only on macOS.

PR #5006, opened 2018-01-08
Test suite: cache the result of hasProfiledLibraries.

23Skidoo:
What it says on the tin. No need to run this every time if the compiler stays fixed.

---


ezyang:
I don't think this does what you want it to: IIUC, now, we run the test at the beginning of executing *every* test script (which are running as separate processes); i.e., this is not the level of cache you are looking for. If you want the orchestrating process to run the test once, you then have to pass the result of the test as a flag to the subprocesses.

23Skidoo:
OK, thanks, will look into it.

23Skidoo:
Closing this for now.

Issue #4901, opened 2017-11-22
Add '--with-msys-location=/path/to/msys' to 'cabal user-config init'


pold500:
Cmd: cabal install hindent

Cabal output:
$ cabal install hindent
Resolving dependencies...
Downloading distributive-0.5.3...
Downloading hashable-1.2.6.1...
Configuring old-time-1.1.0.3...
Downloading lifted-base-0.2.3.11...
Downloading temporary-1.2.1.1...
Configuring distributive-0.5.3...
Configuring hashable-1.2.6.1...
Configuring temporary-1.2.1.1...
Configuring lifted-base-0.2.3.11...
Building hashable-1.2.6.1...
Building temporary-1.2.1.1...
Building lifted-base-0.2.3.11...
Failed to install old-time-1.1.0.3
Build log ( C:\Users\dmytriie\AppData\Roaming\cabal\logs\ghc-8.2.1\old-time-1.1.0.3-B8XpIJbhq9uLX1PMmzKlcp.log ):
Configuring old-time-1.1.0.3...
configure: WARNING: unrecognized options: --with-compiler
checking for gcc... C:\PROGRA~1\HASKEL~1\826561~1.1\mingw\bin\gcc.exe
checking whether the C compiler works... no
configure: error: in `/cygdrive/c/Users/dmytriie/AppData/Local/Temp/cabal-tmp-21116/old-time-1.1.0.3':
configure: error: C compiler cannot create executables
See `config.log' for more details
cabal: Leaving directory 'C:\Users\dmytriie\AppData\Local\Temp\cabal-tmp-21116\old-time-1.1.0.3'
Installed temporary-1.2.1.1
Installed hashable-1.2.6.1
Installed lifted-base-0.2.3.11
Building distributive-0.5.3...
Installed distributive-0.5.3
cabal: Error: some packages failed to install:
cpphs-1.20.8-5cEQ4JG4zEO48rflCalI7n depends on cpphs-1.20.8 which failed to
install.
haskell-src-exts-1.19.1-GkIKqZuv8CjDj4xYyAfokK depends on
haskell-src-exts-1.19.1 which failed to install.
hindent-5.2.4.1-6Jn2HtqpNbYCTNcSVqr91c depends on hindent-5.2.4.1 which failed
to install.
old-time-1.1.0.3-B8XpIJbhq9uLX1PMmzKlcp failed during the configure step. The
exception was:
ExitFailure 77

OS type: Windows 10 Professional
Cabal version: 
[...]

PR #5008, opened 2018-01-08
append option for user-config command

gbaz:
* [x] Patches conform to the [coding conventions (https://github.com/haskell/cabal/#conventions).

Should resolve https://github.com/haskell/cabal/issues/4901

This adds an additional option for `user-config`, as `--append=` or `-a`. This can be used multiple times, each time adding a new line to the delta on `user-config` files for `init` `update` or `diff`. This lets tools like the platform installer pass in, e.g. msys directories for `extra-prog-path` and `extra-lib-dirs` etc.

Tested on some example files, and new-test passes.

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -2403,14 +2403,16 @@ instance Semigroup ExecFlags where
 -- ------------------------------------------------------------
 
 data UserConfigFlags = UserConfigFlags {
-  userConfigVerbosity :: Flag Verbosity,
-  userConfigForce     :: Flag Bool
-} deriving Generic
+  userConfigVerbosity   :: Flag Verbosity,
+  userConfigForce       :: Flag Bool,
+  userConfigAppendLines :: Flag [String]
+  } deriving Generic
 
 instance Monoid UserConfigFlags where
   mempty = UserConfigFlags {
-    userConfigVerbosity = toFlag normal,
-    userConfigForce     = toFlag False
+    userConfigVerbosity   = toFlag normal,
+    userConfigForce       = toFlag False,
+    userConfigAppendLines = toFlag []
     }
   mappend = (<>)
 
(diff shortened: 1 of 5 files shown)


gbaz:
@randen once this is  merged we can update the windows installer to call either `cabal init` or `cabal update` with the relevant msys lines (I guess since we don't have an init-or-update it'll need to be a two step process to check for the existence of the file first...)

gbaz:
(travis error seems spurious)

23Skidoo:
So, if we're updating an already existing config that has an `extra-prog-path` entry, it will have two such entries. Does `cabal-install` handle this case correctly?

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -2446,6 +2448,12 @@ userConfigCommand = CommandUI {
      "Overwrite the config file if it already exists."
      userConfigForce (\v flags -> flags { userConfigForce = v })
      trueArg
+ , option ['a'] ["append"]
```
23Skidoo: I'd call this `--append-lines`. Note that you'll still be able to say `--append=foo`, since `--append` is an unique prefix.

gbaz:
> So, if we're updating an already existing config that has an extra-prog-path entry, it will have two such entries. Does cabal-install handle this case correctly?

Actually, it won't have two entries. The semantics of `update` in general will replace the field, not augment it. That said, cabal-install does handle two such lines correctly.

It may be the "right thing" to augment some lines instead of replace them with update semantics, but we'd have to handle this case by case for all config lines, and that seems too broad for this PR.

gbaz:
As such it might be clearer to name the setting "augment" and change the doc to read  "Additional setting to augment the config file (replacing a previous setting if it existed)." It isn't hidden from the help output -- no reason not to expose this to end users, imho.

23Skidoo:
Sorry for not reacting earlier, was travelling for work. I haven't managed to come up with a better name, so let's use "augment".

PR #5018, opened 2018-01-12
Put command line --with-PROG into AllPackages

angerman:
This should help with #4939. `--with-PROG` flags should be applied to every
package.  If we change hsc2hs via `--with-hsc2hs=` or any other program and
similarly `--hsc2hs-options` should also be provided to that program.  Same
holds for `--PROG-options`.  Because we do no inherit the properties from
AllPackages for LocalPackages, we need to have the set of programs and options
in both.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
@@ -166,10 +166,23 @@ commandLineFlagsToProjectConfig globalFlags configFlags configExFlags
       projectConfigShared        = convertLegacyAllPackageFlags
                                      globalFlags configFlags
                                      configExFlags installFlags,
-      projectConfigLocalPackages = convertLegacyPerPackageFlags
-                                     configFlags installFlags haddockFlags
+      projectConfigLocalPackages = localConfig,
+      projectConfigAllPackages   = allConfig
     }
-
+  where (localConfig, allConfig) = splitConfig
+                                 (convertLegacyPerPackageFlags
+                                    configFlags installFlags haddockFlags)
+        -- split the package config (from command line arguments) into
+        -- those applied to all packages and those to local only.
+        --
+        -- for now we will just copy over the ProgramPaths/Args/Extra into
+        -- the AllPackages.  The LocalPackages do not inherit them from
+        -- AllPackages, and as such need to retain them.
+        splitConfig :: PackageConfig -> (PackageConfig, PackageConfig)
+        splitConfig pc = (pc
+                         , mempty { packageConfigProgramPaths = packageConfigProgramPaths pc
+                                  , packageConfigProgramArgs  = packageConfigProgramArgs  pc
+                                  , packageConfigProgramPathExtra = packageConfigProgramPathExtra pc })
 
 -- | Convert from the types currently used for the user-wide @~/.cabal/config@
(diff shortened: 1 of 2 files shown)


angerman:
What's up with travis? Jobs take for ever and die with random (e.g. apt-get) errors...

ezyang:
I dunno what's up with Travis, but if it's a build system failure keep refreshing the failing build until it passes.

I know this is orthogonal to the problem you are trying to solve, but could you take a look over `PackageConfig` and see if there are other fields we want to apply globally? Is there a simple rule which we can apply to make this decision one way or another?

ezyang:
> every time I type that, I wonder if we should rather go with package * for consistency; all-packages is yet another key word I need to remember while package * feels much more natural to me...

I don't have any problem with `package *` :)

Your decision criteria seems reasonable to me, let's put it in the code!

23Skidoo:
`package *` was also suggested by @hvr during the discussion on the origingal all-packages PR, and everyone seemed to agree that it's a good idea. Patches welcome.

angerman:
@ezyang, @23Skidoo see #5053 for `package *`.

angerman:
I hereby declare Travis dead.

angerman:
@ezyang the following should be resolved now:
> Your decision criteria seems reasonable to me, let's put it in the code!

ezyang:
Alright, LGTM.

angerman:
Looks like Travis macOS builder are hopelessly overlaoded.

Also there is one run in which travis fails...
```
$ ./Cabal/misc/travis-diff-files.sh
+git status
+git diff-files -p --exit-code
diff --git a/cabal-install/cabal-install.cabal b/cabal-install/cabal-install.cabal
index 5cd1014..18a9f9d 100644
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -65,6 +65,7 @@ Extra-Source-Files:
   tests/IntegrationTests2/targets/exes-disabled/cabal.project
   tests/IntegrationTests2/targets/exes-disabled/p/p.cabal
   tests/IntegrationTests2/targets/exes-disabled/q/q.cabal
+  tests/IntegrationTests2/targets/lib-only/p.cabal
   tests/IntegrationTests2/targets/libs-disabled/cabal.project
   tests/IntegrationTests2/targets/libs-disabled/p/p.cabal
   tests/IntegrationTests2/targets/libs-disabled/q/q.cabal
```

ezyang:
That error says that you forgot to regenerated some autogenerated parts of cabal-install.cabal. Run update-cabal-files.sh

phadej:
@ezyang @angerman rather run `make gen-extra-source-files`

phadej:
... though I already pushed the fix to master, so no need. `rebase` would be enough if you really want to rerun all the Travis.

angerman:
If we agree this is good to go, I’m fine with merging without stressig poor Travis once again. macOS builds seem dead for the time being anyway, due to some capacity issues at travisci.

angerman:
Whee! It's finally green, can we merge it?

angerman:
@23Skidoo can we merge this?

23Skidoo:
Merged, thanks!

dalaing:
Gah, misclicked while looking for the names of milestones, ignore me.

PR #5035, opened 2018-01-14
Merge dependencies more effectively in D.Solver.Modular.IndexConversion.

grayjay:

______________________________________________________________________

This commit is a more thorough fix for the problem that
4f7ac10 attempted to fix.  4f7ac10 combined
all dependencies on the same package within the same build-depends or
build-tool-depends field, in order to avoid performance problems in the solver
QuickCheck tests.  That wasn't enough to prevent the solver from creating many
duplicate dependencies when converting the PackageDescription to the
solver-specific format, though.  When a package contained many conditionals that
each contained the same dependency in both branches, the solver lifted those
dependencies out of the conditionals without deduplicating them.

This commit moves the step that merges dependencies after the step that lifts
common dependencies out of conditionals.

_____________________________________________________________________

This PR should fix the CI failures that appeared in #5023.  The issue was introduced by #4929.


grayjay:
I'll merge this, since all but two of the build jobs finished.

PR #5041, opened 2018-01-15
Fix #5039: getIPID picks first id

### LINKED ISSUE #5039: getIPID fails when there are multiple packages
When running `PackageTests/Backpack/Includes3/setup-external-ok.test.hs`, it fails because I have non-global `containers` in `.store`.

but that's often the case as `new-build` gets used more and more!

```
id: containers-0.5.10.2
id: containers-0.5.7.0-9028308afe4ede14e53eba8de7c4a08422b3cef30ab40c590d8611f3715008a4

stderr:
*** Exception: could not determine id of containers
```

cc @ezyang

phadej:
If there are many package versions, any should work.


phadej:
merged manually

PR #5042, opened 2018-01-15
Derive NFData for FlagName and FlagAssignment.

peti:
This is useful for code that uses "parallel" in combination with Cabal data
structures (like we do in cabal2nix and hackage2nix).


23Skidoo:
OK.

PR #5047, opened 2018-01-16
Derive NFData instance for GenericPackageDescription.

peti:
Define NFData instance for GenericPackageDescription. GenericPackageDescription contains a whole lot of other types, each of which requires NFData instances, too. All the definitions rely on the respective Generic instance via `genericRnf`.

A unit test to verify that the NFData instance can find bottoms in GPD has been added.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -445,6 +445,7 @@ test-suite unit-tests
     UnitTests.Distribution.Simple.Utils
     UnitTests.Distribution.SPDX
     UnitTests.Distribution.System
+    UnitTests.Distribution.Types.GenericPackageDescription
     UnitTests.Distribution.Utils.Generic
     UnitTests.Distribution.Utils.NubList
     UnitTests.Distribution.Utils.ShortText
--- a/Cabal/Distribution/Compiler.hs
+++ b/Cabal/Distribution/Compiler.hs
@@ -66,6 +66,8 @@ data CompilerFlavor =
 
 instance Binary CompilerFlavor
 
+instance NFData CompilerFlavor where rnf = genericRnf
+
 knownCompilerFlavors :: [CompilerFlavor]
 knownCompilerFlavors = [GHC, GHCJS, NHC, YHC, Hugs, HBC, Helium, JHC, LHC, UHC]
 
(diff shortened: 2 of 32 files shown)


hvr:
> but "Generic" gives them all to us for free.

You have to be very careful here though... that's only true for `build-depends: deepseq ^>= 1.4.0.0`! (if you want to use `Generic` with older `deepseq` versions, see the compat notes in its haddocks)

(I know how you meant "for free", but I'd nitpick there's a minor cost involved: compile time + object size)

peti:
@hvr, I see what you mean. I edited all definitions to use the `genericRnf` function provided by Cabal's internal compat code to make sure the code works properly with deepseq versions prior to 1.4.

@phadej, I added the bottom-chasing unit test you requested.

phadej:
Fails with GHC-7.4.2:

```
tests/UnitTests/Distribution/Types/GenericPackageDescription.hs:36:3:
    Ambiguous occurrence `catch'
    It could refer to either `Distribution.Compat.Prelude.Internal.catch',
                             imported from `Distribution.Compat.Prelude.Internal' at tests/UnitTests/Distribution/Types/GenericPackageDescription.hs:6:1-43
                             (and originally defined in `System.IO.Error')
                          or `Control.Exception.catch',
                             imported from `Control.Exception' at tests/UnitTests/Distribution/Types/GenericPackageDescription.hs:11:1-24
                             (and originally defined in `Control.Exception.Base')
```

Docs of `Control.Exception.catch`, https://hackage.haskell.org/package/base-4.5.0.0/docs/Control-Exception.html#v:catch  in `base-4.5` say

> Note that the Prelude also exports a function called catch with a similar type to catch, except that the Prelude version only catches the IO and user families of exceptions (as required by Haskell 98).

Later `Prelude` don't export `catch`. I'd `import qualified Control.Exception as C`  and use `C.catch` to avoid ambiguity.

peti:
`catch` name clash is fixed by 29cf7b9f5bcd05b5ce84588f2a6529e15addd0bb. Or so I hope. :-)

phadej:
@peti: 
```
Not in scope: data constructor `ErrorCall'
    Perhaps you meant `C.ErrorCall' (imported from Control.Exception)
```

peti:
Yeah, I force-pushed a fix.

phadej:
Seems meta check fails. I'll merge this manually later today.

PR #5050, opened 2018-01-17
Change license :: License to Either SPDX.License License

### LINKED ISSUE #2547: Consider adopting/support SPDX licence identifiers

phadej:
~WIP: [ci skip] Have to fix cabal-tests and check that we don't break
older (i.e all) ghc-pkgs~

Resolves #2547

I introduce SimpleLicenseExpression to make "OrAnyLater LicenseRef"
unrepresentable. That also simplifies types.

license field is parsed as old License when cabal-version <2.2,
and as SPDX expression otherwise. NONE is recognised.

There are also IPI changes, there both old and SPDX license expressions
are accepted, as both can occur in package database.

### DIFF
--- a/Cabal/Distribution/Simple/Register.hs
+++ b/Cabal/Distribution/Simple/Register.hs
@@ -74,6 +74,7 @@ import qualified Distribution.Simple.Program.HcPkg as HcPkg
 import Distribution.Simple.Setup
 import Distribution.PackageDescription
 import Distribution.Package
+import Distribution.License (licenseToSPDX, licenseFromSPDX)
 import qualified Distribution.InstalledPackageInfo as IPI
 import Distribution.InstalledPackageInfo (InstalledPackageInfo)
 import Distribution.Simple.Utils
@@ -407,7 +408,11 @@ generalInstalledPackageInfo adjustRelIncDirs pkg abi_hash lib lbi clbi installDi
     IPI.instantiatedWith   = componentInstantiatedWith clbi,
     IPI.sourceLibName      = libName lib,
     IPI.compatPackageKey   = componentCompatPackageKey clbi,
-    IPI.license            = license     pkg,
+    -- If GHC >= 8.4 we register with SDPX, otherwise with legacy license
+    IPI.license            =
+        if ghc84
+        then Left $ either id licenseToSPDX $ licenseRaw pkg
+        else Right $ either licenseFromSPDX id $ licenseRaw pkg,
     IPI.copyright          = copyright   pkg,
     IPI.maintainer         = maintainer  pkg,
     IPI.author             = author      pkg,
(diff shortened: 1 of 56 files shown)


phadej:
I probably end up introducing `spdx` field into IPI, so `license` field can be left intact, and new field used for SPDX stuff. I have to run `cabal-tests` to see how badly things are broken right now.

phadej:
I made `Register` write SPDX info with GHC >= 8.4 only. That will do it. Simple.

Cabal/Distribution/Simple/Register.hs
```diff
@@ -407,7 +408,11 @@ generalInstalledPackageInfo adjustRelIncDirs pkg abi_hash lib lbi clbi installDi
     IPI.instantiatedWith   = componentInstantiatedWith clbi,
     IPI.sourceLibName      = libName lib,
     IPI.compatPackageKey   = componentCompatPackageKey clbi,
-    IPI.license            = license     pkg,
+    -- If GHC >= 8.4 we register with SDPX, otherwise with legacy license
+    IPI.license            =
+        if ghc84
+        then Left $ either id licenseToSPDX $ licenseRaw pkg
```
phadej: This choice converts everything to SPDX, so in GHC-8.4 package databases will only be SPDX expressions.
Alternatively we can let `licenseRaw` to go though unmodified, or convert only "sure" things (e.g. MIT, BSD3, GPL-3, not `PublicDomain` or `OtherLicense`)

phadej:
How it's possible that in GHC-7.8.4 output https://travis-ci.org/haskell/cabal/jobs/330116951#L2091 I see

```
Linked component graph:
    unit my-0
        include base-4.11.0.0
```

Cabal/Distribution/Simple/Register.hs
hvr: Sounds good to me :-)

phadej:
ping @ezyang , any ideas why tests fail on older GHC?

phadej:
Ah I see, it actually tests with `ghc-head`! I have to skip that tests before GHC-head is updated :)

23Skidoo:
You can change the GHC version check to `8.4.0.2018MMDD`: make this a month or two in the future now and update later once you know the correct snapshot date.

phadej:
@23Skidoo head is 8.5 so that won't work, I'll just temporarily skip the test for 8.4+ and unskip it when Cabal is updated in GHC-tree.

phadej:
@23Skidoo do you want to take a closer look, or should I merge?

23Skidoo:
I'll try to later today, feel free to merge if I won't.

phadej:
I merged this now, any comments are still welcome

Issue #4691, opened 2017-08-15
Building with Stack is broken


tfausak:
After a fresh clone (at f3a5767f95e575d8c1a59a18a351b89472432451), I am unable to build this project with Stack:

```
cabal> stack test
Invalid flag specification:
- Package 'Cabal' does not define the following flags (specified in stack.yaml):
  parsec
- Flags defined by package 'Cabal':
  Cabal:bundled-binary-generic
  Cabal:old-directory
  Cabal:parsec-struct-diff
- Package 'cabal-install' does not define the following flags (specified in stack.yaml):
  parsec
- Flags defined by package 'cabal-install':
  cabal-install:debug-conflict-sets
  cabal-install:debug-expensive-assertions
  cabal-install:debug-tracetree
  cabal-install:lib
  cabal-install:monolithic
  cabal-install:network-uri
  cabal-install:old-bytestring
  cabal-install:old-directory
```

I think #4654 caused this.

23Skidoo:
Yep, it has been broken for a while because most active contributors don't use Stack.

tfausak:
Sure, but Travis CI is supposed to be testing with Stack: https://github.com/haskell/cabal/blob/f3a5767f95e575d8c1a59a18a351b89472432451/.travis.yml#L84

Two problems with that though:

- It manually sets a resolver, which is weird because the `stack.yaml` already does that.
- It is allowed to fail, and has indeed been failing: https://travis-ci.org/haskell/cabal/jobs/264667352#L548

23Skidoo:
> It manually sets a resolver,

If you think that this is wrong, feel free to fix!

> It is allowed to fail,

This is by design, we want to be aware whether it works or not (plus it's useful for validating PRs such as this one), but we don't want to impose the burden of keeping it working on all our contributors. See e.g. [here](https://github.com/haskell/cabal/pull/4667#issuecomment-321036564).

tfausak:
> This is by design, we want to be aware whether it works or not (plus it's useful for validating PRs such as this one), but we don't want to impose the burden of keeping it working on all our contributors.

Why even bother having it? It's been broken for at least 5 months: https://travis-ci.org/haskell/cabal/jobs/208471402

23Skidoo:
> Why even bother having it?

Um, I think I explained it? Because we find it occasionally useful when someone who uses Stack for working on Cabal decides to fix it.

tfausak:
[...]

PR #5052, opened 2018-01-18
Fix stack configuration and improve travis stack setup

mgsloan:

Tested by running `stack build --test` in the repo.  Travis changes not yet tested, we'll see once it checks this PR.  Previously it would immediately yield the following error:

```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for Cabal-2.1.0.0:
    QuickCheck-2.10.1 from stack configuration does not match >=2.11.3 && <2.12 (latest matching
                      version is 2.11.3)
    tree-diff must match >=0.0.1 && <0.1, but the stack configuration has no specified version
              (latest matching version is 0.0.1)
needed since Cabal is a build target.

Some different approaches to resolving this:

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some
    working build configuration. This can be convenient when dealing with many complicated
    constraint errors, but results may be unpredictable.

  * Recommended action: try adding the following to your extra-deps
    in /home/mgsloan/fpco/stack/cabal/stack.yaml:

- QuickCheck-2.11.3
- tree-diff-0.0.1

Plan construction failed.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,15 +85,13 @@ matrix:
    - env: GHCVER=8.0.2 SCRIPT=bootstrap
      os: osx
 
-   - env: GHCVER=via-stack SCRIPT=stack STACKAGE_RESOLVER=lts
+   - env: GHCVER=via-stack SCRIPT=stack STACK_CONFIG=stack.yaml
      os: linux
-     addons:
-        apt:
-            packages:
-                - libgmp-dev
 
+   # See https://github.com/haskell/cabal/pull/4667#issuecomment-321036564
+   # for why failures are allowed.
   allow_failures:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -85,15 +85,8 @@ matrix:
...
 
-   - env: GHCVER=via-stack SCRIPT=stack STACKAGE_RESOLVER=lts
+   - env: GHCVER=via-stack SCRIPT=stack STACK_CONFIG=stack.yaml
      os: linux
-     addons:
-        apt:
-            packages:
-                - libgmp-dev
-
-  allow_failures:
```
23Skidoo: Please leave `allow_failures` on, see [previous discussion](https://github.com/haskell/cabal/issues/4691#issuecomment-322471915) for rationale.

PR #5054, opened 2018-01-18
Make Paths_ autogen module work with default-extensions (stack 3789)

mgsloan:
Fixes https://github.com/commercialhaskell/stack/issues/3789

I couldn't get the tests working for me, so instead I interacted with `cabal-testsuite/PacakgeTests/AutogenModules/SrcDist` manually.  Running `cabal install` there yielded

```
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.
Resolving dependencies...
AutogenModules-0.1 (reinstall)
Warning: Note that reinstalls are always dangerous. Continuing anyway...
Configuring AutogenModules-0.1...
Building AutogenModules-0.1...
Failed to install AutogenModules-0.1
Build log ( /home/mgsloan/.cabal/logs/ghc-8.2.2/AutogenModules-0.1-7iTLfhgxOYB4FmH7GTg4PZ.log ):
cabal: Entering directory '.'
Configuring AutogenModules-0.1...
Preprocessing executable 'Exe' for AutogenModules-0.1..
cabal: can't find source for MyExeHelperModule in ., dist/build/Exe/autogen,
dist/build/global-autogen

cabal: Leaving directory '.'
AutogenModules-0.1-7iTLfhgxOYB4FmH7GTg4PZ failed during the building phase.
The exception was:
ExitFailure 1
```

So, instead I commented out all the stanzas but the library, and commented out the `*HelperModule` stuff.  Then I tested that this new test cabal file reproduced https://github.com/commercialhaskell/stack/issues/3789 .  Then, I made the change to fix the Paths_ file generation.  Then I did `stack build` and `stack exec bash` to enter the stack environment with this modified version of `cabal`.  Then I ran `unset GHC_PACKAGE_PATH` to prevent it from crashing. 
 Then I used `cabal install` in the dir and observed that it now builds after the modifications.

### DIFF
--- a/Cabal/Distribution/Simple/Build/PathsModule.hs
+++ b/Cabal/Distribution/Simple/Build/PathsModule.hs
@@ -39,10 +39,28 @@ import System.FilePath ( pathSeparator )
 
 generate :: PackageDescription -> LocalBuildInfo -> ComponentLocalBuildInfo -> String
 generate pkg_descr lbi clbi =
-   let pragmas = cpp_pragma ++ ffi_pragmas ++ warning_pragmas
+   let pragmas =
+            cpp_pragma
+         ++ no_overloaded_strings_pragma
+         ++ no_rebindable_syntax_pragma
+         ++ ffi_pragmas
+         ++ warning_pragmas
 
-       cpp_pragma | supports_cpp = "{-# LANGUAGE CPP #-}\n"
-                  | otherwise    = ""
+       cpp_pragma
+         | supports_cpp = "{-# LANGUAGE CPP #-}\n"
+         | otherwise    = ""
+
+       -- -XOverloadedStrings is problematic because 'fromString' is not
+       -- in scope, so disable it.
+       no_overloaded_strings_pragma
+         | supports_overloaded_strings = "{-# LANGUAGE NoOverloadedStrings #-}\n"
+         | otherwise                   = ""
+
+       -- -XRebindableSyntax is problematic because when paired with
+       -- -XOverloadedLists, 'fromListN' is not in scope, so disable it.
+       no_rebindable_syntax_pragma
+         | supports_rebindable_syntax = "{-# LANGUAGE NoRebindableSyntax #-}\n"
+         | otherwise                  = ""
 
        ffi_pragmas
         | absolute = ""
@@ -53,8 +71,7 @@ generate pkg_descr lbi clbi =
           "{-# OPTIONS_JHC -fffi #-}\n"
 
        warning_pragmas =
-        "{-# OPTIONS_GHC -fno-warn-missing-import-lists #-}\n"++
-        "{-# OPTIONS_GHC -fno-warn-implicit-prelude #-}\n"
+        "{-# OPTIONS_GHC -fno-warn-missing-import-lists #-}\n"
 
        foreign_imports
         | absolute = ""
@@ -235,13 +252,15 @@ generate pkg_descr lbi clbi =
 
         path_sep = show [pathSeparator]
 
-        supports_cpp = compilerFlavor (compiler lbi) == GHC
+        supports_cpp = supports_language_pragma
+        supports_overloaded_strings = supports_language_pragma
+        supports_rebindable_syntax= ghc_newer_than (mkVersion [7,0,1])
+        supports_language_pragma = ghc_newer_than (mkVersion [6,6,1])
 
-        supports_language_pragma =
-          (compilerFlavor (compiler lbi) == GHC &&
-            (compilerVersion (compiler lbi)
-              `withinRange` orLaterVersion (mkVersion [6,6,1]))) ||
-           compilerFlavor (compiler lbi) == GHCJS
+        ghc_newer_than minVersion =
+          case compilerCompatVersion GHC (compiler lbi) of
+            Nothing -> False
+            Just version -> version `withinRange` orLaterVersion minVersion
 
 -- | Generates the name of the environment variable controlling the path
 -- component of interest.
--- a/cabal-testsuite/PackageTests/PathsModule/Library/my.cabal
+++ b/cabal-testsuite/PackageTests/PathsModule/Library/my.cabal
@@ -5,7 +5,7 @@ author: Johan Tibell
 stability: stable
 category: PackageTests
 build-type: Simple
-Cabal-version: >= 1.2
+Cabal-version: >= 1.10
 
 description:
     Check that the generated paths module compiles.
@@ -13,3 +13,22 @@ description:
 Library
     exposed-modules: Paths_PathsModule
     build-depends: base
+    default-language: Haskell2010
+    default-extensions:
+        -- This is a non-exhaustive list of extensions that can cause code to
+        -- not compile when it would if the extension was disabled. This ensures
+        -- that autogen modules are compatible with default extensions.
+        NoImplicitPrelude
+        CPP
+        TemplateHaskell
+        QuasiQuotes
+        Arrows
+        OverloadedStrings
+    if impl(ghc >= 6.12)
+       default-extensions: MonoLocalBinds
+    if impl(ghc >= 7.0.1)
+       default-extensions: RebindableSyntax
+    if impl(ghc >= 7.4.1)
+       default-extensions: NoTraditionalRecordSyntax
+    if impl(ghc >= 7.8.1)
+       default-extensions: OverloadedLists
--- a/cabal-install/changelog
+++ b/cabal-install/changelog
@@ -59,6 +59,9 @@
 	* Removed support for building cabal-install with GHC < 7.10 (#4870).
 	* New 'all-packages' section in 'cabal.project' files that applies
 	options to all packages, not just those local to the project.
+	* Paths_ autogen modules now compile when `RebindableSyntax` or
+	`OverloadedStrings` is used in `default-extensions`.
+	[stack#3789](https://github.com/commercialhaskell/stack/issues/3789)
 
 2.0.0.1 Mikhail Glushenkov <mikhail.glushenkov@gmail.com> December 2017
 	* Support for GHC's numeric -g debug levels (#4673).


mgsloan:
Appveyor error is questionable:
> solver-quickcheck.exe: getMBlocks: VirtualAlloc MEM_COMMIT failed: The paging file is too small for this operation to complete.

grayjay:
Sorry about that!  The failure should be fixed by #5035.

Cabal/Distribution/Simple/Build/PathsModule.hs
```diff
@@ -67,6 +68,16 @@ generate pkg_descr lbi clbi =
           "import System.Environment (getExecutablePath)\n"
         | otherwise = ""
 
+       overloaded_strings_imports
+         | supports_overloaded_strings =
+           "import Data.String (fromString)\n"
```
hvr: Tbh, I'd prefer if we went for different way here. I can imagine a couple of ways to break this fragile workaround. But there's a simpler way to accomplish this w/ less assumptions about the compilation scope/environment: Just use `{-# LANGUAGE NoOverloadedStrings #-}` to neutralise `default-languages: OverloadedStrings` locally.

rohit507: I'd also like to point out that overloaded lists has the same effect, so you'd have to add both suppressors.

rohit507: Alternately stopping rebindable syntax seems like the best bet.

rohit507: Alternately stopping rebindable syntax seems like the best bet.

mgsloan: @hvr Ah, thanks for pointing that out!  I'm sure you could break this if you wanted to. I had indeed forgotten about `-XNo*`, since I never use it other than for `NoImplicitPrelude`.  For example:

```haskell
{-# LANGUAGE NoWeirdStuff #-}
```

causes

```
hmm.hs:1:14: error: Unsupported extension: NoWeirdStuff
```

So, this will need to be conditional on compilation scope/environment regardless.  I think there should be a pragma to reset language extensions.  I have opened https://ghc.haskell.org/trac/ghc/ticket/14685

mgsloan: There is also precedent for my original approach.  Why do you think the generated code has `import Prelude`??  I will also switch that over to `{-# ImplicitPrelude #-}`.

@rohit507 `OverloadedLists` is not a problem on its own, because it only needs `fromListN` to be in scope if `RebindableSyntax` is enabled.

hvr: Indeed, I'm not happy with the current approach either; The generated `Paths_` module should try to make as little assumptions about the module namespace as possible and this has been a pet peeve of mine anyway; but there's no reason to increase the technical debt now; and I think there's better ways to accomplish the goal without making things worse.

hvr: fwiw, there's a general pattern here: Starting with GHC 7.0, Cabal can emulate a resetting `-X` flag: it just needs to neutralise (almost) all  `-X` flags (and Cabal knows enough about them to know which have a `No` counterpart). Then we could easily also avoid the redundant `import Prelude`.

23Skidoo:
@phadej 

FTR, we made changes to `Paths_` autogeneration previously without making it dependent on `cabal-version`, and no-one has complained yet. IMO doing what you suggest adds too much complexity for little benefit.

Cabal/Distribution/Simple/Build/PathsModule.hs
hvr: We only need this for `Paths_*.hs`

phadej:
@23Skidoo yes, probably the risk is low (`build-type: Custom` and `default-extensions` combination is rare, /now/), but we still need a `cabal check` for authors, because there is possibility things won't work.

I'd still require a `cabal check` (no need to do in this PR) before this go into release, because so far there weren't packages with problematic `default-extensions` and `Paths_pkg`, because they simply didn't work

```haskell
       -- | An issue that might not be a problem for the package author but
       -- might be annoying or detrimental when the package is distributed to
       -- users. We should encourage distributed packages to be free from these
       -- issues, but occasionally there are justifiable reasons so we cannot
       -- ban them entirely.
     | PackageDistSuspicious { explanation :: String }
```

mgsloan:
Making this conditional on cabal-version trades one kind of surprising behavior for another.

Users like @rohit507 would encounter the same error and need to google around until they found that adding a constraint on cabal version magically fixes the problem.   Do you guys make all fixes conditional on cabal version??

Pros of making it conditional on cabal-check:

phadej:
@mgsloan to comment out of other discussion: what's about `NoImplicitPrelude`

phadej:
Seems we have
```
         "import Prelude\n"++
```
there.  Maybe still worth adding to the test?

phadej:
Screw me, it's already there. I'm sorry, I'm blind today.

mgsloan:
Yup, NoImplicitPrelude is in the test, along with all other extensions I've ever found to cause Haskell98 code to no longer compile.

cabal-testsuite/PackageTests/PathsModule/Library/my.cabal
```diff
@@ -13,3 +13,22 @@ description:
...
+        -- This is a non-exhaustive list of extensions that can cause code to
+        -- not compile when it would if the extension was disabled. This ensures
+        -- that autogen modules are compatible with default extensions.
+        NoImplicitPrelude
+        CPP
+        TemplateHaskell
+        QuasiQuotes
+        Arrows
+        OverloadedStrings
+        MonoLocalBinds
```
phadej: this is twice

mgsloan: Fixed

hvr:
> Do you guys make all fixes conditional on cabal version??

Why the double `?`s? The irony isn't lost on me that given you usually seem to claim to care lot about reproducibility, you'd advocate sneaking in semantic changes without version guarding them properly...

23Skidoo:
> we can easily end up with packages which despite specifying `cabal-version:1.10` would only work with `cabal-install-2.2`

Fair enough, but I'm also worried that `Build.PathsModule` will become an unmaintainable mess. It already has all this auto-generated CPP and compat stuff. Adding a `cabal check` warning sounds like a good compromise, but it should only check for extensions that this PR/Cabal 2.2 adds workarounds for.

hvr:
> Adding a `cabal check` warning sounds like a good compromise, but it should only check for extensions that this PR/Cabal 2.2 adds workarounds for.

...and any other incompatiblities we may *know* about (even if we don't yet have workarounds).

23Skidoo:
@hvr
Sounds good, though in the latter case bumping `cabal-version` won't help.

mgsloan:
> Why the double ?s? The irony isn't lost on me that given you usually seem to claim to care lot about reproducibility, you'd advocate sneaking in semantic changes without version guarding them properly...

It is clear that we have very different attitudes about software. Your tone here is fairly offensive.  You just took my mild snark of double `??` and brought it to a whole 'nother level of ad hominem. The prospect of interacting with you in discussions on the ghc and Cabal projects greatly dampens my enthusiasm for contributing.  Please consider the net effect of your attitudes and behaviors...

I am for making as many builds successful as possible.  You and @phadej are suggesting that we should withhold fixes from users unless they know that they need to specify a `cabal-version` value.

That is like saying if ghc has a panic, then the panic should only be fixed if the enables an extension.  Otherwise users of the older ghc might be confused.

Lets make a table!

|   | Require cabal-version for fix | Don't |
| ---- | ---- | ---- |
| older cabal + no cabal-version: specified | confusing error | confusing error |
| older cabal + `cabal-version:` specified | says I need a new version of cabal without explanation | confusing error |
| newer cabal + no `cabal-version:` specified | confusing error | works |
| newer cabal + `cabal-version:` specified | works | works |

Lets please have our tools move towards "just works".

mgsloan:
Ah I see the point now about `cabal check`.  The cassava thing still pisses me off, so I didn't take the time to consider @hvr 's suggestion properly.  I think my table and reasoning still holds, though, since `cabal check` is not run on every build.  How about instead:

* Require `cabal-version`
* Have `cabal configure` error out if it detects the following conditions:

  - Build includes `Paths_` module

  - `cabal-version:` constraint is too old

  - `default-extensions:` contains either `OverloadedStrings` or `RebindableSyntax`.

The updated table is

|   | Require cabal-version for fix + detect need for fix | Don't |
| ---- | ---- | ---- |
| older cabal + no `cabal-version:` specified | confusing error | confusing error |
| older cabal + `cabal-version:` specified | error saying I need a newer cabal version.  Still confusing, no explanation given | confusing error |
| newer cabal + no `cabal-version:` specified | clear error | works |
| newer cabal + `cabal-version:` specified | works | works |

23Skidoo:
IMO the "newer cabal + no `cabal-version`" cell should be "fails with a good explanation" in the first case, since we'd have enough info to make it fail better if we wanted it to fail. @mgsloan is right that it is more about backwards compat than reproducibility, however.

Anyway, the plan proposed by @phadej (merge this PR as-is and add a `cabal check` warning later) sounds good to me, let's do that and move on.

phadej:
Wait a minute.

> That is like saying if ghc has a panic, then the panic should only be fixed if the enables an extension. Otherwise users of the older ghc might be confused.

**No**, I say that panic will be fixed in *the next version*, and you have to use *newer version* of GHC. To communicate that to users of your package you specify that in some way:
usually a lower bound on `base` as a proxy to GHC version.

In this case, with generation of `Paths_*` module, it means package author must specify the lower bound of `Cabal` to be used. `cabal-version` for that.

As I don't think this will break any existing package on Hackage, the fix to generation code can be unconditional. Old broken configurations will magically start working, but nothing *should* break. Yet, `cabal check` is a warning & linter check, it **should** notify package authors that something might not work with tools compiled against older `Cabal`, `cabal-install` and `stack` and...

To rephrase to avoid confusion:
I ask to add a **warning** check only. I was to hasty to default to
*do not change behaviour of old `cabal-version* (that's the safest default).
I apologize for that.

@mgsloan, to reformulate your table:

## old cabal-install

Old `cabal-install`s cannot warn about this.

| | Warn about old cabal-version | Don't |
|---|---|---|
| old-cabal version | *confusing error* &dagger; | *confusing error* &dagger; |
| new-cabal version | *says you need to update cabal* | *says you need to update cabal* |

## new cabal-install

| | Warn about old cabal-version | Don't |
|---|---|---|
| old-cabal version | *warns* | *works silently* |
| new-cabal version | *works* | *works* |

I want to concentrate on fields marked with &dagger;. We can assume
that these don't exist on Hackage now. THere's only "old `cabal-install`
for now users are unluckily to upload versions of packages they cannot compile themselves.

If we pick *right* column, then there will be *old-cabal-version + incompatible configurations* packages on Hackage,
and users will get confusing errors.

If we pick *left* column, then as *new cabal-install* gives a warning, Hackage (it uses newest `Cabal`)
will reject such packages, and the repository will stay clear of *old cabal-version + incompatible configuration* packages.

---

TL;DR this patch is ok from me. `Cabal/Distribution/Simple/Build/PathsModule.hs` might need cleanup, or resetting of configuration maybe done in more principled way, I don't know.
But, I will add a check about possible problems with *old cabal-version*, `Paths_pkg` module, and `default-extensions`. 

That's what @23Skidoo said, and I agree.

mgsloan:
> I ask to add a warning check only. I was to hasty to default to
> do not change behaviour of old `cabal-version (that's the safest default).
> I apologize for that.

No worries, a warning in `cabal check` seems good to me!

I hadn't considered unconditionally having this fix but having `cabal check` warn about it.  Retrospectively that's an obvious option, but I was distracted by hvr's snark which triggered me to some degree because I'm still fairly annoyed about the cassava thing.  Though, I haven't seen a user report of running into that for a few days now.. I guess everyone upgraded.

23Skidoo:
AppVeyor failure seems to be legit.

mgsloan:
@23Skidoo Fixed the appveyor error, current CI errors seems to be something else.

Not surprising that the test didn't pass.  I hadn't run the test, just verified that `cabal install` with the modified file works.

23Skidoo:
Closed & reopened to restart CI.

23Skidoo:
Downstream Travis sez:

```
$ ./hackage-tests parsec
hackage-tests: /home/travis/.cabal/config: openBinaryFile: does not exist (No such file or directory)
The command "./travis-test.sh" exited with 1.
```

@phadej, do you know why this is failing?

phadej:
I fixed that in master

23Skidoo:
Closed & reopened to restart Travis once again.

23Skidoo:
Merged, thanks!

PR #5061, opened 2018-01-20
Improve file globbing logic (fixes #5057)

### LINKED ISSUE #5057: `cabal sdist` does not pick up files with multiple dots
I've distilled this example into https://github.com/RyanGlScott/cabal-sdist-bug/tree/910078306b4e127996bd7c9208a7e741b9f74351. This has a `.cabal` file with the following `extra-source-files`:

```
extra-source-files: check-hs/*.check.hs
```

And one can verify that there are indeed [several files in that location](https://github.com/RyanGlScott/cabal-sdist-bug/tree/910078306b4e127996bd7c9208a7e741b9f74351/check-hs) with a `.check.hs` extension. But when one attempts to run `cabal sdist` on that project, `cabal` becomes confused:

```
$ cabal sdist
Warning: Cannot run preprocessors. Run 'configure' command first.
Building source dist for cabal-sdist-bug-0.1.0...
cabal: filepath wildcard 'check-hs/*.check.hs' does not match any files.
```

For some reason, `cabal` seems to get confused when the wildcard asterisk is used in conjunction with file extensions with multiple dots like `.check.hs` (at least, that's my current theory). In order to work around this bug, I have to manually enumerate every single `.check.hs`, like what I [currently must do in `base-compat`](https://github.com/haskell-compat/base-compat/commit/c990dfb9a774b73d43e1b2421d8dd2ed4e72ab1d).

```
$ cabal --version
cabal-install version 2.0.0.1
compiled using version 2.0.1.1 of the Cabal library
```

mitchellwrosen:
I tested this change manually by running `cabal sdist` on https://github.com/RyanGlScott/cabal-sdist-bug/tree/910078306b4e127996bd7c9208a7e741b9f74351:

```
$ cabal sdist
Building source dist for cabal-sdist-bug-0.1.0...
Source tarball created: dist/cabal-sdist-bug-0.1.0.tar.gz
```

### DIFF
--- a/Cabal/Distribution/Simple/Utils.hs
+++ b/Cabal/Distribution/Simple/Utils.hs
@@ -1148,7 +1148,7 @@ matchDirFileGlob dir filepath = case parseFileGlob filepath of
     case   [ dir' </> file
            | file <- files
            , let (name, ext') = splitExtensions file
-           , not (null name) && ext' == ext ] of
+           , not (null name) && ext `isSuffixOf` ext' ] of
       []      -> die $ "filepath wildcard '" ++ filepath
                     ++ "' does not match any files."
       matches -> return matches


hvr:
This likely also needs a `cabal check`, as old cabals will choke.

mitchellwrosen:
Sorry, I'm unclear on which came first, these TODOs or your comment about how the current behavior was actually intended. Should I close this PR?

23Skidoo:
That behaviour was intended originally, but there seems to be a consensus that we should relax it a bit.

mitchellwrosen:
I've added a test and updated `changelog`. Could I have some guidance on these remaining TODOs?

> (& make sure that the documented limitations for * in filenames remain in place).

What are the documented limitations for `*` in filenames?

> possibly a cabal check compat-warning

What does this mean?

> possibly user guide update (to document that starting with version x.y, fileglobbing supports multi-part extensions)

Where is the user guide located?

Thanks!

phadej:
I'm really sorry that we (me and @hvr) are demanding. I try to explain:

---

> possibly a cabal check compat-warning`

run `cabal check`. The implementation probably have to be in `Distribution.PackageDescription.Check`.
There is e.g. `check use of "data-files: data/*.txt" syntax`.

The issue is that:
- the change is good for ergonomics
- but if someone else developing the package would use older `cabal-install`, their `cabal sdist` won't pick all the files, resulting in broken package.
    - unfortunately there isn't `cabal check` which checks that globs actually pick at least *some* files. There should be.
    - new version should warn, that 'extended globbing' is needed, and `cabal-version` is too low.
        - that's tricky, but doable

The scenario is not that different to #5073:
if some CI setup (for any reason) uses older `cabal-install` for some job, than `cabal sdist` will produce broken tarballs, resulting in confusingly failing builds.

> Where is the user guide located?

`Cabal/doc`

PR #5070, opened 2018-01-24
Enable hackage-tests on downstream travis

phadej:

### DIFF
--- a/travis/binaries/travis-test.sh
+++ b/travis/binaries/travis-test.sh
@@ -21,7 +21,10 @@ ln -s $TRAVIS_BUILD_DIR $UPSTREAM_BUILD_DIR
 
 # Test we can parse Hackage
 # Note: no $TEST_OPTIONS as this isn't tasty suite
-cabal update # fetch 01-index.tar
+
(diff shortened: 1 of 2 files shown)


travis/binaries/travis-test.sh
```diff
@@ -21,9 +21,8 @@ ln -s $TRAVIS_BUILD_DIR $UPSTREAM_BUILD_DIR
 
 # Test we can parse Hackage
 # Note: no $TEST_OPTIONS as this isn't tasty suite
-if [ -f ~/.cabal/config ]; then
+cabal update # fetch 01-index.tar
```
23Skidoo: I see, that was a temporary band-aid. Can we do `user-config init` here instead? That should be faster.

phadej: We need `update`, `hackage-tests` need index file.

phadej: ... should I add that as a comment?

23Skidoo: No harm in that.

PR #5077, opened 2018-01-30
Pull 4874 basedir

phadej:
#4874, I tested locally with GHC-7.10.3 and GHC-8.2.2 - it was green, I hope Travis agrees

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -96,10 +96,10 @@ import Distribution.Text
 
 -- Base
 import System.Environment (getArgs, getProgName)
-import System.Directory   (removeFile, doesFileExist
+import System.Directory   (removeFile, doesFileExist, getCurrentDirectory
                           ,doesDirectoryExist, removeDirectoryRecursive)
 import System.Exit                          (exitWith,ExitCode(..))
-import System.FilePath                      (searchPathSeparator)
+import System.FilePath                      (searchPathSeparator, takeDirectory, (</>))
 import Distribution.Compat.Environment      (getEnvironment)
 import Distribution.Compat.GetShortPathName (getShortPathName)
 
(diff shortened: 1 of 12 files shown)


phadej:
Ok, bootstrap seems to be broken. Otherwise this is hopefully fine to merge.

ping @23Skidoo @angerman

angerman:
I can’t make sense of the version mismatch during `cabal clean` yet. I’ll try to look into it end of week/weekend as I’m bed-bound right now. If someone else could take a look earlier, that would be great!

angerman:
What seems to happen is the following: when building we produce a `setup-config` in the `Cabal` folder. When running `cabal clean` we stumble over that `setup-config` and fail.

Now I'm trying to figure out, if this stumbling is *new* or if the `setup-config` is *new*.

In any case, I do not think that `clean` should care whether or not there is an existing `setup-config`.

Cabal/Distribution/Simple.hs
```diff
@@ -328,7 +334,10 @@ haddockAction hooks flags args = do
 cleanAction :: UserHooks -> CleanFlags -> Args -> IO ()
 cleanAction hooks flags args = do
     distPref <- findDistPrefOrDefault (cleanDistPref flags)
-    let flags' = flags { cleanDistPref = toFlag distPref }
+
+    lbi <- getBuildConfig hooks verbosity distPref
+    let flags' = flags { cleanDistPref = toFlag distPref
+                       , cleanCabalFilePath = maybeToFlag (cabalFilePath lbi)}
```
angerman: This is essentially the offending part. We try to `getBuildConfig` in 338, and then try to use the local build info here. But this might just fail, if the build Config is too old. (In most cases we should be able to read it, as we produce it with the same cabal, that we try to clean it with, unless we bootstrap...).

angerman:
Just pushed to #4874. Let's see what Travis has to say.

angerman:
#4874 is finally green (except for AppVoyer, which seems to have gotten confused :-/)

I’ll close this in Favor of #4874

PR #5078, opened 2018-01-30
CompilerId: add NFData instance (based on genericRnf)

peti:

I have code that depends on the NFData instance (previously defined locally as an orphan), and I verified that code still works properly.


23Skidoo:
Merged, thanks!

Issue #5043, opened 2018-01-15
Why does mkFlagName not convert the string to lower-case?


peti:
My understanding is that flag names are supposed to be case-less, i.e. they are normalized to all-lowercase in the `Parsec` instance of `FlagName`. This feels like an awkward place to do this, however. Why doesn't that conversion take place in `mkFlagName`? As it is now, a construct like

    mkFlagAssignment [(mkFlagName "foo", True), (mkFlagName "FOO", True), (mkFlagName "Foo", False)]

gives

    fromList [(FlagName "FOO",(1,True)),(FlagName "Foo",(1,False)),(FlagName "foo",(1,True))]

instead of:

    fromList [(FlagName "foo",(3,False))]

That doesn't feel right?

phadej:
cc @hvr

23Skidoo:
@hvr Ping.

phadej:
Caseness is tricky (it's dependent on Unicode standard), the differences are very subtle. I'd rather make flags case-insensitive in `cabal-version: 3.0`

(Hackage doesn't accept non-ascii flags, but that's its choice)

peti:
I've extended the test suite in https://github.com/haskell/cabal/pull/5082 to verify that flag names are case-less.

Personally, I believe flag names *should* be case-less (and they should be constrained to plain ASCII, i.e. no Unicode). Change that behavior now or in the future would be a rather nasty surprise to everyone who's relied on the current behavior in their scripts and Cabal config files.

hvr:
@phadej I'd limit case-insensitivity to the ASCII subset; but everything outside the ASCII subset ought to be case-sensitive; DNS case-folding does something like that as well.  However, I'd rather make flags completely case-sensitive in `cabal-version:3.0` to make things more consistent with the case-sensitivity of package names, and eschew the subtle issues of case-folding altogether.

BardurArantsson:
Why are we not doing the obvious here and limiting the flag names to ASCII? (Plus, case folding)

Is there *any* plausible use case for non-ASCII flags, especially given that Hackage doesn't accept them per @phadej ?

BardurArantsson:
I don't see much of an argument there, I'm afraid.

It's just "I want to be able to do X". There doesn't seem to be any reasoning as to *why* the restriction would be a problem. Yes, I understand that some programming communities may want to *prefer* being able to use whatever characters they want, but I can't see it would actually be a problem to just say "sorry, not supported".

[...]

PR #5082, opened 2018-01-31
Ensure proper case-less handling of FlagName

peti:
This patch moves the conversion to all-lower-case out of the `FlagName` parser instances and into the `mkFlagName` constructor function. 

Fixes https://github.com/haskell/cabal/issues/5043.

### DIFF
--- a/Cabal/tests/UnitTests/Distribution/Types/GenericPackageDescription.hs
+++ b/Cabal/tests/UnitTests/Distribution/Types/GenericPackageDescription.hs
@@ -6,14 +6,18 @@ import Prelude ()
 import Distribution.Compat.Prelude.Internal
 import Distribution.Types.GenericPackageDescription
 
+import qualified Control.Exception as C
+import Control.Monad
 import Test.Tasty
 import Test.Tasty.HUnit
-import qualified Control.Exception as C
 
 tests :: [TestTree]
 tests =
   [ testCase "GenericPackageDescription deepseq" gpdDeepseq
+  , testCase "findDuplicateFlagAssignments" testFindDuplicateFlagAssignments
   ]
+
+----- Verify "NFData GenericPackageDescription" instance ----------------------
 
 gpdFields :: [(String, GenericPackageDescription -> GenericPackageDescription)]
 gpdFields =
@@ -35,3 +39,13 @@ throwsUndefined :: NFData a => String -> a -> Assertion
 throwsUndefined field a =
   C.catch (C.evaluate (rnf a) >> assertFailure ("Deepseq failed to evaluate " ++ show field))
           (\(C.ErrorCall _) -> return ())
+
+----- Verify duplicate flag recognition in "FlagAssignment" -------------------
+
+testFindDuplicateFlagAssignments :: Assertion
+testFindDuplicateFlagAssignments = do
+  let flag n v = (mkFlagName n, v)
+      findDuplicates = findDuplicateFlagAssignments . mkFlagAssignment
+  forM_ [ (b1,b2) | b1 <- [True,False], b2 <- [True,False] ] $ \(v1,v2) -> do
+    assertEqual "" [] (findDuplicates [flag "foo" v1, flag "bar" v2])
+    assertEqual "" [mkFlagName "foo"] (findDuplicates [flag "foo" v1, flag "FOO" v2])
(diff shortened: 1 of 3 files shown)


23Skidoo:
There are some legit CI failures: it fails to build on older GHCs and AppVeyor found this:

```
Unit Tests
  Distribution.Types.GenericPackageDescription
    findDuplicateFlagAssignments:                                                         FAIL
      tests\UnitTests\Distribution\Types\GenericPackageDescription.hs:53:
      unexpected: []
    FlagName is case-insensitive:                                                         FAIL
      *** Failed! Falsifiable (after 26 tests):
      [(FlagName "B",True),(FlagName "Vp",True),(FlagName "T",False),(FlagName "Y",False),(FlagName "u",True),(FlagName "We",False),(FlagName "I",True),(FlagName "Y",True),(FlagName "m",False),(FlagName "o",False),(FlagName "D",False),(FlagName "C",False),(FlagName "f",False),(FlagName "d",True),(FlagName "Tv",False),(FlagName "l",True),(FlagName "dM",False),(FlagName "g",False),(FlagName "sc",False),(FlagName "M",False),(FlagName "DS",False),(FlagName "aE",True),(FlagName "ea",True),(FlagName "W",False),(FlagName "df",False)]
      Use --quickcheck-replay=474123 to reproduce.
```

Cabal/tests/UnitTests/Distribution/Types/GenericPackageDescription.hs
```diff
@@ -35,3 +41,33 @@ throwsUndefined :: NFData a => String -> a -> Assertion
...
+genFlagValue :: Gen (FlagName, Bool)
+genFlagValue = (,) <$> genFlagName <*> arbitrary
+
+propertyFlagNamesAreCaseInsensitive :: Property
+propertyFlagNamesAreCaseInsensitive =
+  forAll (listOf1 (resize 2 genFlagValue)) $ \fa ->
+    let fa'        = nub [ map toLower (unFlagName n) | (n,_) <- fa ]
+        duplicates = findDuplicateFlagAssignments (mkFlagAssignment fa)
+    in
+       length duplicates == length fa - length fa'
```
phadej: use `===` (triple equals) to print LHS and RHS

peti:
@23Skidoo, yes, of course! That failure documents the fact that our current API does not ensure that flag name are treated in a case-insensitive way. We apply `lowercase` when parsing a FlagAssignment from String, but any other way of creating that type bypasses this conversion and then `foo` and `FOO` are, in fact, two distinct flag names. These tests should succeed, but they don't.

23Skidoo:
Do you want to mark it as expected failure (to be fixed later) and merge as is, or do you plan to fix the underlying issue as part of the same PR?

Cabal/tests/UnitTests/Distribution/Types/GenericPackageDescription.hs
peti: @phadej, I force-pushed a new version (72e4f5575243305b3524d5666f5660c3f965b9c3) that uses `===`.

peti:
@23Skidoo, I would like to fix the issue in this PR. The obvious solution would be to move the conversion to all lower-case into the `mkFlagName` function. I suggested that in https://github.com/haskell/cabal/issues/5043. So far, nobody has spoken out against that change, but curiously enough no-one has spoken out in favor of it either, so I haven't made change yet. I probably should.

23Skidoo:
Noted.

peti:
OK, I fixed the underlying issue and moved the `map toLower` conversion out of the `FlagName` parser code and into the `mkFlagName` constructor function. I also fixed the build errors with older compilers and re-based my patch to the current `master`.

There are still CI failures in Travis et al, but these failures seem to be unrelated to my changes.

peti:
All CI tests on travis are green. I think this PR is ready to be merged.

23Skidoo:
There are some failing tests in downstream CI builds.

peti:
I don't what is causing these CI test failures in appveyor. I have run the cabal test suite locally and it succeeds just fine on my machine.

hvr:
Well, there's still the issue that this PR introduces a technical debt we'll have to revert when cleaning up the semantics as e.g. suggested in https://github.com/haskell/cabal/issues/5043#issuecomment-361914463, as then we can't just normalise in `mkFlagName`; instead we'll have to move it out again, effectively reverting large parts of this PR

peti:
Well, there is also the issue that `mkFlagName` allows users to create a `FlagName` that violates the semantics of that type, and that problem exists *today*, whereas those new-and-improved semantics are going to exist at some point in the future of Cabal 3.0, maybe.

peti:
So ... I trust this all means that this PR will be lying here forever and that I might just as well close it?

PR #5085, opened 2018-01-31
Remove 'ExecutableScopeUnknown'.

### LINKED ISSUE #5083: fix pretty-printing of `executable.scope`
It always printed as `scope: unknown`.

This is not biggie but easy to fix.

There is no label for pretty-printer, so I tag it as parser issue :P

phadej:
Fixes #5083
Related to #5003

Note: `buildable: ` field has `All` semantics. `scope: ` has
`AnyPrivate` semantics. Both might be surprising, but are "logical".

https://hackage.haskell.org/package/Cabal-2.0.1.1/docs/src/Distribution.PackageDescription.Parse.html#line-248

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -53,7 +53,6 @@ import Distribution.System
 import Distribution.Text
 import Distribution.Types.ComponentRequestedSpec
 import Distribution.Types.CondTree
-import Distribution.Types.ExecutableScope
 import Distribution.Types.ExeDependency
 import Distribution.Types.UnqualComponentName
 import Distribution.Utils.Generic                    (isAscii)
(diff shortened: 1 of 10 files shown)


23Skidoo:
We didn't allow users to write `scope: unknown`, so this is backwards compatible, right? Otherwise LGTM.

phadej:
`ReadP` parser is

```
instance Text ExecutableScope where
      parse = Parse.choice
         [ Parse.string "public"  >> return ExecutablePublic
         , Parse.string "private" >> return ExecutablePrivate
         ]
```

so cabal-2.20 didn't accept anything else than `public` and `private`. I think I made Parsec more lenient to allow roundtrip, so it was forgotten to do item.

23Skidoo:
OK, let's merge.

phadej:
Failure is legit, added test in master "conflicts"

23Skidoo:
Merged, thanks!

Issue #5092, opened 2018-02-01
`cabal check` doesn't detect using invalid version syntax for spec versions prior to 1.24


hvr:
This already caused 4 incidents in the package index (see https://github.com/haskell-infra/hackage-trustees/issues/128)

@phadej found the culprit at 6b3457de66772d958fd5fe96066b08e93d0fb0c7   which relaxed the `Version` parser to accommodate parsing `pkg-config` output (which imho deserves its own version parsing anyway), and as an overlooked side-effect also relaxed the parsing of the cabal format.

23Skidoo:
+1 on having a `PkgConfigVersion` type.

PR #5093, opened 2018-02-01
Warn on leading zeros in version digits for old cabal spec

phadej:
Fixes https://github.com/haskell/cabal/issues/5092

### DIFF
--- a/Cabal/Distribution/Types/Version.hs
+++ b/Cabal/Distribution/Types/Version.hs
@@ -102,13 +102,13 @@ instance Parsec Version where
         mkVersion <$> P.sepBy1 digit (P.char '.') <* tags
       where
         digitParser v
-            | v >= CabalSpecV1_24 = P.integral
-            | otherwise           = (some d >>= toNumber) P.<?> "non-leading-zero integral"
+            | v >= CabalSpecV2_0 = P.integral
+            | otherwise          = (some d >>= toNumber) P.<?> "non-leading-zero integral"
           where
             toNumber :: CabalParsing m => [Int] -> m Int
             toNumber [0] = return 0
             toNumber xs@(0:_) = do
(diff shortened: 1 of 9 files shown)


Cabal/Distribution/Types/Version.hs
```diff
@@ -96,8 +97,25 @@ instance Pretty Version where
...
+        mkVersion <$> P.sepBy1 digit (P.char '.') <* tags
       where
+        digitParser v
+            | v >= CabalSpecV1_24 = P.integral
+            | otherwise           = (some d >>= toNumber) P.<?> "non-leading-zero integral"
+          where
+            toNumber :: CabalParsing m => [Int] -> m Int
+            toNumber [0] = return 0
+            toNumber xs@(0:_) = do
+                parsecWarning PWTVersionLeadingZeros "Version digit with leading zero. Use cabal-version: 1.24 or later to write such versions. For more information see https://github.com/haskell/cabal/issues/5092"
```
hvr: It's a bit more tricky; while technically cabal 1.24 happily copes with this, cabal 1.22 would not, as cabal 1.22 would still try to parse future spec versions, and then choke on unexpected syntax.

So as far as `cabal check`'s category of "inacceptable things for distribution via hackage" errors is concerned, we should probably require at least `cabal-version:2.0`; *or* we may consider leading zeroes a mistake, and prevent them from being for *any* .cabal file uploaded to hackage.

phadej: I'll change the cutoff to 2.0.

If we want to completely prevent them from Hackage is something I don't feel confident to decide right now (even it will affect only few packages)

23Skidoo:
The new test fails on AppVeyor.

23Skidoo:
The only remaining failure here was meta: `.check` files were not included in `extra-source-files`. Fixed that and merged manually in b36cf3c9a91e4e270a190e23fe9f65fb8b593f4e.

phadej:
@23Skidoo thanks!

PR #5098, opened 2018-02-02
Quick proof of concept refactoring of `VersionRange`

hvr:
This is a preparatory refactoring for exploring/evaluation how to handle introducing a separate `Version` type for `pkg-config` version numbers (which have a richer syntax/grammar, such as `openssl-1.0.2j` and according to the `pkg-config` docs uses RPM-style version comparision semantics, see http://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/ for the details).

The current refactoring is effectively limited to modifications in a single file, by introducing `type` syns for compatiblity. I've only tried compiling `lib:Cabal`, which did build fine. I'd expect the heap-representation and the `Binary` instances to be unaffected by this refactoring.

The ugliest part is for me is the current ad-hoc `IsVersion` class hack that factors out the `Version`-specific logic.

The reason we need this, is because of

```hs
-- | Describes a dependency on a pkg-config library
--
-- @since 2.0.0.2
data PkgconfigDependency = PkgconfigDependency
                           PkgconfigName
                           VersionRange
```

which currently uses a `VersionRange`; but if we want to introduce a `PkgconfigVersion` type, we'd have to 

 1. either duplicate `VersionRange` (into a `PkgconfigVersionRange`) or 
 2. abstract the current one over the "version field".

And this refactoring explores the 2nd option, which tries to reduce code duplication (think also of e.g. how the solver code would be affected).

Any suggestions/ideas?

### DIFF
--- a/Cabal/Distribution/Types/VersionRange.hs
+++ b/Cabal/Distribution/Types/VersionRange.hs
@@ -4,9 +4,12 @@
 {-# LANGUAGE DeriveGeneric      #-}
 {-# LANGUAGE DeriveTraversable  #-}
 {-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
 module Distribution.Types.VersionRange (
     -- * Version ranges
-    VersionRange(..),
+    VersionRange,
+    VersionRangeOver (..),
 
     -- ** Constructing
     anyVersion, noVersion,
@@ -28,7 +31,8 @@ module Distribution.Types.VersionRange (
     hasLowerBound,
 
     -- ** Cata & ana
-    VersionRangeF (..),
+    VersionRangeF,
+    VersionRangeOverF (..),
     cataVersionRange,
     anaVersionRange,
     hyloVersionRange,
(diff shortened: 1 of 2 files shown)


Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -185,28 +193,37 @@ withinVersion = WildcardVersion
...
 majorBoundVersion = MajorBoundVersion
 
+-- We could unify VersionRangeOver and VersionRangeOverF
+--
+-- > type VersionRangeOver v = VersionRangeOverF v (VersionRangeOver v)
+--
+-- But that results in a cycle in the type-syn definition...
+--
+
+type VersionRangeF a = VersionRangeOverF Version a
```
phadej: i'd omit eta-reduce, will make type-synonym more useful.

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -340,34 +358,94 @@ withinRange v = foldVersionRange
...
-
 -------------------------------------------------------------------------------
 -- Parsec & Pretty
 -------------------------------------------------------------------------------
 
-instance Pretty VersionRange where
+class Ord v => IsVersion v where
+  dispWild :: v -> Disp.Doc
+  noVersion :: VersionRangeOver v
+  majorUpperBound :: v -> v
```
phadej: need to preserve haddocks.

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -340,34 +358,94 @@ withinRange v = foldVersionRange
...
+    where v = mkVersion [1]
+
+
+  -- | Compute next greater major version to be used as upper bound
+  --
+  -- Example: @0.4.1@ produces the version @0.5@ which then can be used
+  -- to construct a range @>= 0.4.1 && < 0.5@
+  --
+  -- @since 2.2
+  majorUpperBound = alterVersion $ \numbers -> case numbers of
```
phadej: because instance member haddocks are IMHO hidden quite well.

phadej: also what will `^>=` mean for `PkgConfigVersion`? Should that be disallowed? structurally or in `cabal check`?

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -340,34 +358,94 @@ withinRange v = foldVersionRange
...
 -- Parsec & Pretty
 -------------------------------------------------------------------------------
 
-instance Pretty VersionRange where
+class Ord v => IsVersion v where
+  dispWild :: v -> Disp.Doc
+  noVersion :: VersionRangeOver v
+  majorUpperBound :: v -> v
+  wildcardUpperBound :: v -> v
+  verOrWild :: CabalParsing m => m (Bool, v)
```
phadej: this is ugly.

phadej: ... I mean, we probably need an extension point to allow `version` specific `VersionRange` syntax constructs. And this isn't the prettiest one. Have to think what would work best (given `Version` and `PkgConfigVersion` differences).

IMO `openssl ==1.0.1*` isn't something we want to allow. I don't know what `pkg-configs` own syntax looks like.

phadej: One idea is to bundle `versionRangeReadP` as a whole into `IsVersion`

hvr: This is indeed the part I'm most unsure/unhappy about.

Cabal/Distribution/Types/VersionRange.hs
hvr: true; I didn't bother yet, as this was just a quick proof of concept; but haddocks definitely need polish if this PR shall proceed.

Cabal/Distribution/Types/VersionRange.hs
hvr: Well,  `^>=` assumes a well established semantic versioning policy/convention. This doesn't  exist for  `pkg-config(1)` versions. The meaning of `^>=` for pkgconfig is thus IMO not well-defined, and we should discourage its use until we come up with a good one. So at the very least a `cabal check` warning/error is in order.

idontgetoutmuch:
FYI - I was installing HEAD of the gnu scientific library (gsl) and it had a version of 

> 4.5+

Kleidukos:

ffaf1:

PR #5117, opened 2018-02-05
Add more lenses for PackageDescription and GenericPackageDescription

Ericson2314:
Just enough to simplify 'Distribution.PackageDescription.Configuration' for now.

#5100 will rely on this, both for brevity and package-description polymorphism.

CC @kmicklas

### DIFF
--- a/Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Rank2Types #-}
 module Distribution.Types.GenericPackageDescription.Lens (
     GenericPackageDescription,
     Flag,
@@ -10,11 +11,6 @@ import Prelude()
 import Distribution.Compat.Prelude
 import Distribution.Compat.Lens
 
-import Distribution.Types.GenericPackageDescription (GenericPackageDescription(GenericPackageDescription), Flag(MkFlag), FlagName, ConfVar (..))
-
--- lens
-import Distribution.Types.BuildInfo.Lens
-
 -- We import types from their packages, so we can remove unused imports
 -- and have wider inter-module dependency graph
 import Distribution.Types.CondTree (CondTree)
@@ -23,6 +19,9 @@ import Distribution.Types.Executable (Executable)
 import Distribution.Types.PackageDescription (PackageDescription)
 import Distribution.Types.Benchmark (Benchmark)
 import Distribution.Types.ForeignLib (ForeignLib)
+import Distribution.Types.GenericPackageDescription
+  ( GenericPackageDescription(GenericPackageDescription)
+  , Flag(MkFlag), FlagName, ConfVar (..))
 import Distribution.Types.Library (Library)
 import Distribution.Types.TestSuite (TestSuite)
 import Distribution.Types.UnqualComponentName (UnqualComponentName)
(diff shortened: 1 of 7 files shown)


Ericson2314:
Hmm AppVeryor failures involves allow older and newer, which makes sense, but I cannot even figure out what's going on with Travis.

Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
```diff
@@ -35,51 +35,37 @@ import Distribution.Version (VersionRange)
 -------------------------------------------------------------------------------
 
 condBenchmarks :: Lens' GenericPackageDescription [(UnqualComponentName, CondTree ConfVar [Dependency] Benchmark)]
-condBenchmarks f (GenericPackageDescription x1 x2 x3 x4 x5 x6 x7 x8) = fmap (\y1 -> GenericPackageDescription x1 x2 x3 x4 x5 x6 x7 y1) (f x8)
+condBenchmarks f s = fmap (\x -> s { T.condBenchmarks = x }) (f (T.condBenchmarks s))
```
phadej: these are auto-generated, why change?

Cabal/Distribution/Types/Benchmark/Lens.hs
```diff
@@ -25,3 +25,6 @@ benchmarkInterface f s = fmap (\x -> s { T.benchmarkInterface = x }) (f (T.bench
 benchmarkBuildInfo :: Lens' Benchmark BuildInfo
 benchmarkBuildInfo f s = fmap (\x -> s { T.benchmarkBuildInfo = x }) (f (T.benchmarkBuildInfo s))
 {-# INLINE benchmarkBuildInfo #-}
+
+class HasBenchmarks a where
+  traverseBenchmarks :: Traversal' a Benchmark
```
phadej: I'd name this and others just `benchmarks`.

Also add

```haskell
instance HasBenchmarks Benchmark where
    benchmarks = id
```

Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
Ericson2314: Oh I didn't realize that. Can they be easily regenerated or was it a one-off copy-the-th-splice?

Cabal/Distribution/Types/Benchmark/Lens.hs
Ericson2314: I tired that first, but that conflicted with the lenses for PacksheDescription.

Ericson2314:
@phadej the the big, yet to be realized, polymorphism is for Check.hs. Now maybe we shouldn't be checking both PD and GPD long term, but that's not clearly write or wrong to me, and I don't want to prematurely commit. Even in that case, I'd argue it would be easier to land this, land #5100, and then revert some or all of this.

Ericson2314:
The allow * errors would seem to be because I am missing soem sort of dependency somewhere, but I don't see anything obviously missing in the traversals, hmm.

Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
Ericson2314: If it was a one-off generation, I'm inclined to keep this as more readable and immune to argument order changes.

Ericson2314:
I think these remaining failures are unrelated? https://github.com/haskell/cabal/pull/5100 has been rebased on top of this showing the benefits: Check.hs is now entirely `GenericPackageDesccription`-specific, or polymorphic.

phadej:
@Ericson2314 to keep myself sane I added `after-2.2-release` label to this.

Ericson2314:
OK 2.2 is branched, so this can be looked at again.

Ericson2314:
@phadej I slimmed this down to the bare minimum to clean up `PackageDescription.Configuration`

Ericson2314:
Looks like I might just need to rebase.

Ericson2314:
Spurious failure?

Ericson2314:
@phadej are we good now?

Ericson2314:
No sure where @phadej is, and I'm quite confident I addressed the review.

Issue #4815, opened 2017-10-06
foreign-library: Which RTS to link against?


treblacy:
Thank you for the foreign-library addition, it is pretty sweet, it makes my life easier as the author of [Calling Haskell Shared Libraries from C (Linux)](http://www.vex.net/~trebla/haskell/so.xhtml) :laughing:

Currently it links against the unthreaded GHC RTS. So there is potential that one day someone has a multi-threaded C program that uses a Haskell library built this way, calls Haskell functions from multiple threads concurrently, and there will be a problem. (The unthreaded RTS detects reentrancy and simply aborts the whole program with a "schedule: re-entered unsafely" message.)

I am not sure what I request instead:

1.  Add moar settings so a dev of a foreign-library can choose. But I'm not sure who is in the best position to choose, the Haskell foreign-library dev or the C program dev.

2.  Link against the threaded RTS across the board. But I'm not sure this harms no one; I don't even know the statistics of how many people have to avoid the threaded RTS.

TravisWhitaker:
I'd like this feature too, and would lean towards option one. I think the Haskell library author should choose which RTS to link against, and if it's truly necessary to make this switchable a flag could be used.

I'm interested in how the plumbing works for the `foreign-import` stanza; I was a bit surprised when `ghc-options: ... -threaded ...` didn't work.

treblacy:
I have a cunning plan! Produce both the unthreaded version and the threaded version.

sergv:
Just my 2 cents: even if C part has no concurrency whatsoever, it may be nice to still link Haskell shared library with the threaded runtime so that it will do its job faster. In a way, such a C program could delegate all the parallelism into Haskell.

zlonast:
The problem was solved long ago in https://github.com/haskell/cabal/pull/5431

PR #5118, opened 2018-02-06
WIP: foreign-library RTS Selection

TravisWhitaker:
This PR attempts to address https://github.com/haskell/cabal/issues/4815. GHC seems to ignore the "-threaded" flag when "-shared" is used, and one needs to decide which linker flag to use to get the desired RTS flavor. With this change, cabal checks if the "-threaded" flag is provided via ghc-options when building a foreign-library. If it is, we link against the threaded RTS, otherwise we link against the vanilla RTS. The current behavior when "-threaded" is provided is to decide to pass "-lHSrts-ghc***" _and_ "-threaded" to GHC, which confusingly results in a foreign library that's linked against the vanilla RTS even if the user explicitly asks for the threaded one.

It'd be nice hook up the plumbing to link against the other RTS flavors too, but I haven't done enough spelunking to work that out yet.

This is my first time peeking at the Cabal source, so if I've broken some rules please do let me know. This seems to work locally on my machine, but it should be fairly straightforward to test if the tests already include mock builds.


23Skidoo:
Code looks fine, though the approach of making `ghc-options: -threaded` do special magic feels hacky to me. Perhaps it's better to add a `rts-flavour` field to the `foreign-library` stanza instead?

Would be nice if @dcoutts could provide some input.

> It'd be nice hook up the plumbing to link against the other RTS flavors too, but I haven't done enough spelunking to work that out yet.

Do you think it'll be possible to implement using the same "magic `ghc-options`" approach?

TravisWhitaker:
I agree that it feels hacky, but from the user's perspective it's consistent with with how one would choose the threaded runtime when building other targets, and that outweighs the hackiness in my estimation. An `rts-flavor` field would be good; I think it'd be nice to use that field for library and executable targets too, and always warn the user if they put `-threaded` in `ghc-options`. This is what happens right now if the `ghc-options` field has `-lfoo` in it, cabal-install asks the the user to do `extra-libraries: foo` instead.

Some of the RTS flavors have existing cabal flags, e.g. `--enable-profiling` and `--enable-debug-info`, while `-eventlog` needs to put in `ghc-options` just like `-threaded`.

TravisWhitaker:
>  I think it'd be nice to use that field for library and executable targets too, 

Naturally `rts-flavor` doesn't matter for libraries because you don't choose an RTS at all when building those.

angerman:
I’ve been thinking about this a bit as well. We used to hack the rts package together as the idea of multiple libraries in a single package just wasn’t representable with cabal until very recently. And even then cabal can’t build the rts, because we do not have a concept of flavours...

At this point I wonder if we really need and want this setup, or if we could just have multiple rts packages? We might need backpack though? @ezyang might be able to tell us if we could use backpack here.

What we definitely want is to be able to just relink with with a different rts and get profiling/threading,...

ezyang:
I don't think Backpack will work here since you need a module to instantiate a signature and the rts package has no modules. I agree that this is a very strange special case that doesn't fit in the current package model; I'd rather special-case rts (which, btw, is binary compatible between different variants; not true for most packages) than try to work out some general mechanism for selecting one or another.

angerman:
Thanks @ezyang; yes the rts is a funny library in that respect.

This PR has finally brought (and made me understand foreign-library). As I've been adding and using the `--enable-static` with cabal to pass this into ghc (8.4+) to use the `-staticlib` flag. In that case ghc takes care of linking the library and all it's dependencies together.

This then makes me wonder if we really want to replicate any rts-selection logic in cabal, or if we should just not pass the `rts`, and expect GHC to take care of that?

@TravisCardwell did you check what `ghc` does if it is not given the `-lHSrts...` flag? Or even what the `-v3` output of `ghc` is in that case? Does it pass `-lHSrts...` and `-lHSrts-threaded...` to the linker?

TravisWhitaker:
If no RTS is linked against explicitly then the resulting shared object won't depend on any RTS at load time. This is perfectly reasonable behavior; perhaps downstream I'm building a binary that depends on multiple Haskell shared libraries and I'd like to choose which RTS to link against at that point.

A.hs:
```
module A where

f :: Int -> Int
f = (+1)

foreign export ccall f :: Int -> Int
```

Yields:
```
$ ghc -v3 A.hs -dynamic -shared -o liba.so
Glasgow Haskell Compiler, Version 8.2.2, stage 2 booted by GHC version 8.0.2
Using binary package database: /nix/store/ilc2xqyw1xhabdc4adk4gbkacbasvr1d-ghc-8.2.2-with-packages/lib/ghc-8.2.2/package.conf.d/package.cache
package flags []
loading package database /nix/store/ilc2xqyw1xhabdc4adk4gbkacbasvr1d-ghc-8.2.2-with-packages/lib/ghc-8.2.2/package.conf.d
wired-in package ghc-prim mapped to ghc-prim-0.5.1.1
wired-in package integer-gmp mapped to integer-gmp-1.0.1.0
wired-in package base mapped to base-4.10.1.0
wired-in package rts mapped to rts
wired-in package template-haskell mapped to template-haskell-2.12.0.0
wired-in package ghc mapped to ghc-8.2.2
wired-in package dph-seq not found.
wired-in package dph-par not found.
package flags []
loading package database /nix/store/ilc2xqyw1xhabdc4adk4gbkacbasvr1d-ghc-8.2.2-with-packages/lib/ghc-8.2.2/package.conf.d
wired-in package ghc-prim mapped to ghc-prim-0.5.1.1
wired-in package integer-gmp mapped to integer-gmp-1.0.1.0
wired-in package base mapped to base-4.10.1.0
wired-in package rts mapped to rts-1.0
wired-in package template-haskell mapped to template-haskell-2.12.0.0
wired-in package ghc mapped to ghc-8.2.2
wired-in package dph-seq not found.
wired-in package dph-par not found.
*** Chasing dependencies:
Chasing modules from: *A.hs
!!! Chasing dependencies: finished in 0.31 milliseconds, allocated 0.233 megabytes
Stable obj: []
Stable BCO: []
Ready for upsweep
  [NONREC
      ModSummary {
         ms_hs_date = 2018-02-13 16:40:53.226251103 UTC
         ms_mod = A,
         ms_textual_imps = [(Nothing, Prelude)]
         ms_srcimps = []
      }]
*** Deleting temp files:
Deleting:
compile: input file A.hs
*** Checking old interface for A (use -ddump-hi-diffs for more details):
[1 of 1] Compiling A                ( A.hs, A.o )
*** Parser [A]:
!!! Parser [A]: finished in 0.11 milliseconds, allocated 0.125 megabytes
*** Renamer/typechecker [A]:
!!! Renamer/typechecker [A]: finished in 24.65 milliseconds, allocated 14.173 megabytes
*** Desugar [A]:
Result size of Desugar (after optimization)
  = {terms: 17, types: 8, coercions: 0, joins: 0/1}
!!! Desugar [A]: finished in 0.31 milliseconds, allocated 0.152 megabytes
*** Simplifier [A]:
Result size of Simplifier iteration=1
  = {terms: 25, types: 12, coercions: 0, joins: 0/0}
Result size of Simplifier
  = {terms: 25, types: 12, coercions: 0, joins: 0/0}
!!! Simplifier [A]: finished in 0.33 milliseconds, allocated 0.262 megabytes
*** CoreTidy [A]:
Result size of Tidy Core
  = {terms: 25, types: 12, coercions: 0, joins: 0/0}
!!! CoreTidy [A]: finished in 0.10 milliseconds, allocated 0.096 megabytes
writeBinIface: 3 Names
writeBinIface: 22 dict entries
Created temporary directory: /tmp/ghc22103_0
*** CorePrep [A]:
Result size of CorePrep
  = {terms: 27, types: 13, coercions: 0, joins: 0/0}
!!! CorePrep [A]: finished in 0.16 milliseconds, allocated 0.084 megabytes
*** Stg2Stg:
*** CodeGen [A]:
!!! CodeGen [A]: finished in 0.90 milliseconds, allocated 0.801 megabytes
*** C Compiler:
/nix/store/5l6b97jscljkvsq9y9amg71v5jfzddm4-gcc-wrapper-6.4.0/bin/cc -fno-stack-protector -DTABLES_NEXT_TO_CODE -x c /tmp/ghc22103_0/ghc_3.c -o /tmp/ghc22103_0/ghc_4.s -no-pie -fPIC -U__PIC__ -D__PIC__ -Wimplicit -S -include /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/include/ghcversion.h -I. -I/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/base-4.10.1.0/include -I/nix/store/biv9a4szj648s7d77a1r607ndvnwv8k6-gmp-6.1.2-dev/include -I/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/integer-gmp-1.0.1.0/include -I/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/include
*** Assembler:
/nix/store/5l6b97jscljkvsq9y9amg71v5jfzddm4-gcc-wrapper-6.4.0/bin/cc -fno-stack-protector -DTABLES_NEXT_TO_CODE -I. -no-pie -fPIC -U__PIC__ -D__PIC__ -x assembler -c /tmp/ghc22103_0/ghc_4.s -o /tmp/ghc22103_0/ghc_6.o
*** Assembler:
/nix/store/5l6b97jscljkvsq9y9amg71v5jfzddm4-gcc-wrapper-6.4.0/bin/cc -fno-stack-protector -DTABLES_NEXT_TO_CODE -I. -no-pie -fPIC -U__PIC__ -D__PIC__ -x assembler -c /tmp/ghc22103_0/ghc_2.s -o /tmp/ghc22103_0/ghc_7.o
*** Linker:
/nix/store/5l6b97jscljkvsq9y9amg71v5jfzddm4-gcc-wrapper-6.4.0/bin/cc -fno-stack-protector -DTABLES_NEXT_TO_CODE '-Wl,--hash-size=31' -Wl,--reduce-memory-overheads -Wl,--no-as-needed -nostdlib -Wl,-r -no-pie -nodefaultlibs '-Wl,--build-id=none' -o A.o /tmp/ghc22103_0/ghc_8.ldscript
Upsweep completely successful.
*** Deleting temp files:
Deleting: /tmp/ghc22103_0/ghc_9.rsp /tmp/ghc22103_0/ghc_8.ldscript /tmp/ghc22103_0/ghc_7.o /tmp/ghc22103_0/ghc_6.o /tmp/ghc22103_0/ghc_5.rsp /tmp/ghc22103_0/ghc_4.s /tmp/ghc22103_0/ghc_3.c /tmp/ghc22103_0/ghc_2.s /tmp/ghc22103_0/ghc_1.s
Warning: deleting non-existent /tmp/ghc22103_0/ghc_1.s
link: linkables are ...
LinkableM (2018-02-13 16:48:51.892386367 UTC) A
   [DotO A.o]
Linking liba.so ...
*** Linker:
/nix/store/5l6b97jscljkvsq9y9amg71v5jfzddm4-gcc-wrapper-6.4.0/bin/cc -fno-stack-protector -DTABLES_NEXT_TO_CODE '-Wl,--hash-size=31' -Wl,--reduce-memory-overheads -Wl,--no-as-needed -lm -o liba.so A.o -shared -Wl,-Bsymbolic -Wl,-h,liba.so -L/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/base-4.10.1.0 -Xlinker -rpath -Xlinker /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/base-4.10.1.0 -L/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/integer-gmp-1.0.1.0 -Xlinker -rpath -Xlinker /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/integer-gmp-1.0.1.0 -L/nix/store/9q004jgdiqv9pziik3sxffbzwz49fgjn-gmp-6.1.2/lib -Xlinker -rpath -Xlinker /nix/store/9q004jgdiqv9pziik3sxffbzwz49fgjn-gmp-6.1.2/lib -L/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/ghc-prim-0.5.1.1 -Xlinker -rpath -Xlinker /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/ghc-prim-0.5.1.1 -L/nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/rts -Xlinker -rpath -Xlinker /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/rts -lHSbase-4.10.1.0-ghc8.2.2 -lHSinteger-gmp-1.0.1.0-ghc8.2.2 -lHSghc-prim-0.5.1.1-ghc8.2.2 -lm -lgmp -lm
link: done
*** Deleting temp files:
Deleting: /tmp/ghc22103_0/ghc_10.rsp
*** Deleting temp dirs:
Deleting: /tmp/ghc22103_0
```

With no load-time dependency on any RTS:
```
$ ldd liba.so
        linux-vdso.so.1 (0x00007ffce89db000)
        libm.so.6 => /nix/store/d54amiggq6bw23jw6mdsgamvs6v1g3bh-glibc-2.25-123/lib/libm.so.6 (0x00007f2e4ddf9000)
        libHSbase-4.10.1.0-ghc8.2.2.so => /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/base-4.10.1.0/libHSbase-4.10.1.0-ghc8.2.2.so (0x00007f2e4d232000)
        libHSinteger-gmp-1.0.1.0-ghc8.2.2.so => /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/integer-gmp-1.0.1.0/libHSinteger-gmp-1.0.1.0-ghc8.2.2.so (0x00007f2e4cff7000)
        libHSghc-prim-0.5.1.1-ghc8.2.2.so => /nix/store/xjgn21ai9mg1ghfb1sb34wgarnpbcrzm-ghc-8.2.2/lib/ghc-8.2.2/ghc-prim-0.5.1.1/libHSghc-prim-0.5.1.1-ghc8.2.2.so (0x00007f2e4c95b000)
        libgmp.so.10 => /nix/store/9q004jgdiqv9pziik3sxffbzwz49fgjn-gmp-6.1.2/lib/libgmp.so.10 (0x00007f2e4c6c8000)
        libc.so.6 => /nix/store/d54amiggq6bw23jw6mdsgamvs6v1g3bh-glibc-2.25-123/lib/libc.so.6 (0x00007f2e4c328000)
        /nix/store/d54amiggq6bw23jw6mdsgamvs6v1g3bh-glibc-2.25-123/lib64/ld-linux-x86-64.so.2 (0x00007f2e4e30f000)
```

The argument that GHC should handle RTS selection seems reasonable. It's a little like GCC's `libstdc++.so` or `libgcc.so`. I don't have to ask for those explicitly, GCC just assumes that I want them even if I'm building a shared object. However, one could also argue that it's much more flexible to defer linking the RTS if none is specified (I actually have a build system that depends on the current behavior). I'd vote for GHC's current behavior and leaving RTS selection to Cabal. Library authors (in the `foreign-library` sense) are best suited to decide whether or not the threaded RTS should be used, and if I need to link against the RTS downstream there should be some mechanism to do so.

angerman:
I see. In that case I still believe teaching ghc a `-link-rts` flag is useful, as that would contain the special naming for the rts flavours in ghc; rather than spreading it across tools. That however would only work with 8.6+ I believe, unless we sneak that into 8.4; and even then there would be no support for GHC prior to 8.4 for this.

sergv:
What's the status of this PR, please? Maybe it needs help with something?

I am just another user that would like to be able to produce shared libraries with threaded runtime. Regarding points raised before, namely whether `cabal` should use use `-threaded` flag or some kind of `rts-flavour`, I belive that `rts-flavour` stanza is a better long-term solution. But the mismatch in treatment of `-threaded` within `ghc-options:` for executables and foreign libraries pains me and it looks like it could be fixed with small amount of changes. So I'd like to argue that `-threaded` flag should be allowed for foreign libraries just like it is currently allowed for executables. Moreover, going with `-threaded` will work with all GHC versions released so far and will not require a major release of `Cabal`.

TravisWhitaker:
I've been using this PR branch internally for a while without issue. I'd be interested in assembling some sort of consensus around abuse of the `-threaded` flag as implemented in this PR, versus adding an `rts-flavor:` field. I think that either choice could be made to work with previous versions of GHC, and I too am leaning towards a dedicated `rts-flavor` field. My concern with this option is introducing inconsistency with how other flags operate; if we add `rts-flavor`, I think it should also absorb the `--enable-profiling`, `--enable-debug-info`, and `-eventlog` flags. Ideally, one would be able to switch the `threaded`, `evenglog`, `debug-info`, and `profiling` bits with an `rts-flavor` cabal file field, and in an ad-hoc fashion with `--enable-*`/`--disable-*` command line arguments.

Thoughts?

23Skidoo:
I suggest we do the following:

* Merge this PR now.
* Later: implement `rts-flavour` and add a warning for `ghc-options: -threaded` and the like. The warning should be an error if `cabal-version` is sufficiently high.

quasicomputational:
Is the only blocker here the need for a clean rebase?

23Skidoo:
Yes, looks like it.

TravisWhitaker:
I'm a little hesitant to destroy this branch, since at least a few people outside of my organization are using it. Perhaps I'll open a fresh PR.

23Skidoo:
That's fine.

23Skidoo:
This has been merged, see #5431.

Issue #1847, opened 2014-05-05
Allow executables to have Main modules with other names


mcandre:
Originally #172. This ticket edited by @ttuegel on 2015-09-18 to include original problem description.

Query on [haskell-cafe](http://haskell.org/pipermail/haskell-cafe/2007-November/034686.html):

```
It seems the meaning of the -main-is switch for GHC and the Main-Is
build option for Cabal executables differ. With GHC, I can point to
any function "main" in any module, but in Cabal I must point to a
filename with precisely the module name "Main". This is tying my hands
with regard to organizing a default executable and exposing some of
its functionality as a library. Is there a way to get around this
restriction?
Concretely, I want to point Cabal's Main-Is to Program/Main.hs
which starts with
  module Program.Main where
instead of just
  module Main where
```

So in GHC it has [this meaning](http://haskell.org/ghc/docs/latest/html/users_guide/options-phases.html#options-linker):

```
-main-is thing
The normal rule in Haskell is that your program must supply a
main function in module Main. When testing, it is often
convenient to change which function is the "main" one, and
the -main-is flag allows you to do so. The thing can be one of:
A lower-case identifier foo. GHC assumes that the main
function is Main.foo.
    An module name A. GHC assumes that the main function
    is A.main.
    An qualified name A.foo. GHC assumes that the main
    function is A.foo.
```

In Cabal the <tt>main-is:</tt> field specifies the <strong>file name</strong> of the Main module.

GHC's <tt>-main-is</tt> flag is an extension to Haskell that is not supported by the other Haskell implementations.

ttuegel:
Cabal *does* allow any filename for an executable.

gbaz:
The problem is more subtle, and just bit someone again. The original ticket characterized it correctly, and this does not. cabal does not allow for arbitrary _module_ names for executables -- i.e. whatever the _file_ is named, the _module_ must be named main.

(this was the underlying cause of the confusion here: https://www.reddit.com/r/haskell/comments/3lax6y/discussion_thread_about_stack/cv5uv0p)

ttuegel:
@gbaz Try `ghc-options: -main-is Foo`. Can `ghc --make` alone even infer which module is `Main`?

23Skidoo:
@ttuegel +1.

gbaz:
[...]

PR #5122, opened 2018-02-07
indicate explicitly name req for main-is module

gbaz:
This is a small doc change that addresses #1847 by just indicating limitations very clearly.


23Skidoo:
Merged, thanks!

Issue #5108, opened 2018-02-04
cabal check requires absolute cabal-version numbers


ndmitchell:
Cabal HEAD recently started warning on `cabal check` for `cabal-version: >= 1.18`. Why? It seems requiring people to break the version checking code on old Cabal versions is a bad idea...

ndmitchell:
Concretely:

```
neil: Cabal check gave bad warnings:
"* Packages relying on Cabal 1.12 or later should specify a version range of"
"the form 'cabal-version: x.y'. Use 'cabal-version: 1.18'."
```

ndmitchell:
But if my user has Cabal version 1.10, and I write `cabal-version: 1.18`, won't that have nasty consequences? Isn't using `>= 1.18` forevermore preferable? It seems odd to demand breaking old users for the sake of omitting 2 characters, when those two characters arguably make the intent clearer.

phadej:
@ndmitchell the warning is only when `cabal-version: 1.12` or up. You **have** to write `cabal-version: >= 1.10:

```
$ cabal check
The following warnings are likely to affect your build negatively:
* With Cabal 1.10 or earlier, the 'cabal-version' field must use range syntax
rather than a simple version number. Use 'cabal-version: >= 1.10'.

Hackage would reject this package.
```
---

To my understanding, no, `>= 1.18` isn't preferable, as it's misleading. It is interpreted as `1.18`, the field value is not a range. This is why I added a check.

Also note: that that for `cabal-version`s before 2.0 the check is not fatal:

```
$ cabal check
These warnings may cause trouble when distributing the package:
* Packages relying on Cabal 1.12 or later should specify a version range of
the form 'cabal-version: x.y'. Use 'cabal-version: 1.12'.

$ echo $?    
0
```

Only with `2.0` `cabal check` will fail:

```
$ cabal check
These warnings may cause trouble when distributing the package:
* Packages relying on Cabal 1.12 or later should specify a version range of
the form 'cabal-version: x.y'. Use 'cabal-version: 2.0'.

$ echo $?
1
```

and with 2.2 already the parser rejects the file:

```
$ cabal check
Warning: base64-bytestring-type.cabal:15:23:
unexpected cabal-version higher than 2.2 cannot be specified as a range. See
https://github.com/haskell/cabal/issues/4899
expecting ".", "-", white space, "&&" or "||"

>= 2.2
cabal: Failed parsing "./base64-bytestring-type.cabal".
```

---

[...]

PR #5131, opened 2018-02-09
Drop `>=` for `cabal-version` going forward.

angerman:
See #5108. We want `>=` to not be used anymore.

ci skip

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -785,7 +785,7 @@ describe the package as a whole:
 
         package-version = 1*DIGIT *("." 1*DIGIT)
 
-.. pkg-field:: cabal-version: >= x.y
+.. pkg-field:: cabal-version: x.y
 
     The version of the Cabal specification that this package description
     uses. The Cabal specification does slowly evolve, introducing new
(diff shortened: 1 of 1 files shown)


angerman:
Question: should we just drop the `>=` in `cabal-version: >=` from the whole documentation going forward?

phadej:
@angerman thanks for picking this up!

I'm :+1:, but I think we need to amend this with `>= 1.10` case (as then it's still required) and points from #5108

hvr:
@angerman Great that you started on this; I intended to update the docs for `cabal-version` at some point anyway to document some other aspects that have changed. But as far as your PR is concerned, I'd just add a historical comment about the historical syntax for pre-1.12 as it's still part of the (historic) cabal spec. One thing that should be also noted is that `cabal-install` clients prioer to 1.12 will likely not be able to consume today's 00-index.tar, as they'd choke on all the `cabal-version:` files which omit the `^>=`; there's quite a few of those by now.

ntc2:
If you were brought here by a Cabal error message and just want to fix your Cabal file, replace

    cabal-version: >= VERSION

with

    cabal-version: VERSION

m-renaud:
It appears the docs were updated in some other commit along the way (see https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-cabal-version), so I'm going to close out this PR. Sorry we didn't get yours merged sooner.

Issue #5120, opened 2018-02-06
Strange Travis failures


angerman:
Travis seems to fail with:
```
cabal/dist-newstyle/src/pretty-show-1.6.16/dist: renamePath:rename: unsatisfied constraints (Directory not empty)Configuring library for pretty-show-1.6.16..
```
lately a lot.

So I've basically run `git bisect` with `git clean -xfd && cabal new-build cabal-install` on the tree.

That ended up with:

```
84ab693b0162c372c5127e0a27a52bf902beea55 is the first bad commit
commit 84ab693b0162c372c5127e0a27a52bf902beea55
Author: Mikhail Glushenkov <mikhail.glushenkov@gmail.com>
Date:   Thu Jan 4 10:35:30 2018 +0000

    Unbreak Haddock doc generation.
    
    + a whitespace fix.

:040000 040000 375f64d216a741dba303b827e5da2513bd47f490 93da2a12bae5c288aecccff300d636296751937f M	cabal-install
bisect run success
```

log is

```
git bisect start
# good: [abc18aa7a2677a549f2ff57729a0dee026a7f69c] Capitalisation.
git bisect good abc18aa7a2677a549f2ff57729a0dee026a7f69c
# bad: [18317ce96afbdc08e317b41d10802132266b2ec3] Add buildDir includes as needed.
git bisect bad 18317ce96afbdc08e317b41d10802132266b2ec3
# good: [ab585e864929a8d19ed844d82ff2e9d002ae6313] Merge remote-tracking branch 'upstream/dcoutts/project-config-explicit-includes'
git bisect good ab585e864929a8d19ed844d82ff2e9d002ae6313
# bad: [67276f870a55e6a913b98ef7b2ca22a8def21a6e] Merge pull request #4861 from fgaz/new-run/allow-running-tests-benchs
git bisect bad 67276f870a55e6a913b98ef7b2ca22a8def21a6e
# bad: [ce775dbbfe7d04fb9376e2562002a3bbcb0337ed] Add duplicate field negative tests.
git bisect bad ce775dbbfe7d04fb9376e2562002a3bbcb0337ed
# bad: [ec9ce1c82dcff66608a7b1875e11c220828ec2ee] cabal-testsuite: test cabal new-test with coverage
git bisect bad ec9ce1c82dcff66608a7b1875e11c220828ec2ee
# bad: [552ae22bdc7a265bd57b131f3749e239476bd7ca] Rename rewriteFileV to rewriteFileEx.
[...]

PR #5133, opened 2018-02-09
Don't ignore goal order parameter when generating summarized solver log.

grayjay:

_____________________________________________________________________

After #5012, the solver generated the summarized log by using a goal order that
preferred goals from the final conflict set.  It completely ignored the original
goal order parameter.  This caused a failure in one of the unit tests that set a
goal order and then checked the summarized log, when the solver's goal order
changed for other reasons (See #5012).

This commit combines the two goal orders, with the goal order that prefers goals
from the conflict set taking precedence over the original goal order.

_____________________________________________________________________

This should fix the solver unit test failure on master.  I think it was caused by the combination of #5012 and #5035.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular.hs
+++ b/cabal-install/Distribution/Solver/Modular.hs
@@ -9,7 +9,9 @@ module Distribution.Solver.Modular
 -- and finally, we have to convert back the resulting install
 -- plan.
 
-import Data.Map (Map)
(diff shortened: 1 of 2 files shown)


23Skidoo:
Thank you, this is great.

grayjay:
Thanks!  There is only one failed build job (issue #5120), so I'll merge it now.

PR #5137, opened 2018-02-12
Make Cabal-1.24 compile with ghc-8.2.2

bfrk:
No semantic changes, just added the necessary `import Prelude hiding ((<>))` lines where needed. Compiles now with ghc-8.2.2 and ghc-7.10.3.

### DIFF
--- a/Cabal/Distribution/Compiler.hs
+++ b/Cabal/Distribution/Compiler.hs
@@ -60,6 +60,8 @@ import Text.PrettyPrint ((<>))
 import qualified Data.Char as Char (toLower, isDigit, isAlphaNum)
 import Control.Monad (when)
 
+import Prelude hiding ((<>))
+
 data CompilerFlavor = GHC | GHCJS | NHC | YHC | Hugs | HBC | Helium | JHC | LHC | UHC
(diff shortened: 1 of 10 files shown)


bfrk:
OMG, I meant to write ghc-8.4.1 instead of ghc-8.2.2.

phadej:
I'm very against this. `Cabal-1.24` won't be able to speak with GHC-8.4, even you manage to compile it.

There are packages on Hackage without `custom-setup` stanza, they should add it. Or "upgrade' to `build-type: Simple`.

I think https://github.com/simonmar/alex/pull/120 is relevant.

hvr:
I don't think this is a good idea.

bfrk:
I noticed AppVeyor CI fails because you have the -Werror flag on. Do you really want this flag on release branches?

bfrk:
So in order to use ghc-8.4 I must use Cabal > 2 (Cabal the library)? And thus (most probably) make changes to Setup.hs? Where can I read about what I need to change?

23Skidoo:
See https://github.com/haskell/cabal/wiki/2.0-migration-guide.

bfrk:
Thanks.

PR #5140, opened 2018-02-12
Support Haddock's hyperlinker

### LINKED ISSUE #4613: 'cabal haddock --hyperlink-source' should use haddock '--hyperlinked-source' when available
Saw this on Reddit today: https://www.reddit.com/r/haskell/comments/6o4rnb/psa_regarding_cabal_haddock/

> PSA regarding `cabal haddock`
`cabal haddock --haddock-option="--hyperlinked-source"` is different from `cabal haddock --hyperlink-source`.
The former is how you create the nice haddocks that have links on every identifier in the highlighted source. The latter is how you get the old haddocks that just have hscoloured source pages.

I think we should update `cabal haddock --hyperlink-source` to make it use `haddock --hyperlinked-source` instead of `hscolour` when the version of Haddock is sufficiently new.

harpocrates:
Fixes #4613. If a new enough version of Haddock is being used, `--hyperlink-source` will run via Haddock's `--hyperlinked-source` instead of going through HsColour.

I've tested this out with a recent version of GHC/Haddock (8.0.2) and an old version of GHC/Haddock (7.10.2), just to confirm that the fallback behaviour was working properly.

I'm doubly motivated to see this merge since there has been substantial progress on the Haddock hyperlinker backend (we use GHC's tokenizer, so tokenization is accurate as of https://github.com/haskell/haddock/pull/714), and since I was recently bitten by the difference between `cabal haddock --haddock-option="--hyperlinked-sources` and `cabal haddock --hyperlinked-source`.

---


23Skidoo:
Nice, thanks a lot!

harpocrates:
Looks like in some cases CI is unreliable? Anything I should do to make things green?

23Skidoo:
I've restarted those builds. I think you should now also be able to do that if you log in to Travis with your GitHub ID.

Cabal/changelog
```diff
@@ -54,6 +54,8 @@
 	  into per-component condition trees anyway. Now it's finally been put
 	  out of its misery.
 	* Add `BlockArguments` to `KnownExtension`.
+	* `--hyperlinked-source` now uses Haddock's hyperlinker backend when
```
23Skidoo: The `cabal haddock` option name is `--hyperlink-source`. We may want to add a `--hyperlinked-source` synonym.

23Skidoo:
I guess we should also deprecate the `cabal hscolour` command in favour of `cabal haddock --hyperlink-source`.

harpocrates:
@23Skidoo What does deprecation involve (beyond changing the options description)?

Cabal/changelog
harpocrates: @23Skidoo If you don't mind, I would very much like this - I've hit my head on it several times. In fact, could we change the primary option name to `--hyperlinked-source`?

23Skidoo:
Merged, thanks!

Cabal/changelog
23Skidoo: Adding a synonym is fine, but I don't think we should change the config setting name, since people have old config files lying around and I don't want to break those without sufficient reason.

23Skidoo:
@harpocrates 
> What does deprecation involve (beyond changing the options description)?

A changelog note + I think we should make `cabal hscolour` print a warning that it's going to be removed in the next release (3.0).

PR #5142, opened 2018-02-13
Apply global config to all packages

alexbiehl:
I think this was an oversight from https://github.com/haskell/cabal/commit/32ff1cac28206ae320c2cb39dce2d07d74991799 when we introduced `projectConfigAllPackages`. 

Without this patch `convertLegacyGlobalConfig` considers global configuration (e.g. `~/.cabal/config`) only for packages that are local to the project. This patch puts global configuration into `projectConfigAllPackages` to make sure it is applied to all packages whether they are project local or not.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
@@ -217,9 +217,9 @@ convertLegacyGlobalConfig
       savedHaddockFlags      = haddockFlags
     } =
     mempty {
-      projectConfigShared        = configAllPackages,
-      projectConfigLocalPackages = configLocalPackages,
-      projectConfigBuildOnly     = configBuildOnly
+      projectConfigBuildOnly   = configBuildOnly,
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -217,9 +217,9 @@ convertLegacyGlobalConfig
       savedHaddockFlags      = haddockFlags
     } =
     mempty {
-      projectConfigShared        = configAllPackages,
-      projectConfigLocalPackages = configLocalPackages,
-      projectConfigBuildOnly     = configBuildOnly
+      projectConfigBuildOnly   = configBuildOnly,
```
alexbiehl: (I reordered the fields to match declaration order in `Distribution.Client.ProjectConfig.Types`)

23Skidoo:
Makes sense.

23Skidoo:
Downstream Travis is flaky today...

23Skidoo:
Merged, thanks!

PR #5143, opened 2018-02-13
Make --enable-documentation work

alexbiehl:
Adds the ability to generate haddock documentation for non-inplace
package builds.

In other words this allows

  - `$ cabal new-build --enable-documentation`
  - `$ echo "documentation: true" >> cabal.project && cabal new-build`
  - `$ echo "documentation: true" >> ~/.cabal/config && cabal new-build`

And it will generate (and install) haddock documentation for any
package.

(This works best with https://github.com/haskell/cabal/pull/5142 as otherwise global configuration cannot be applied to non-local packages)

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -936,6 +936,13 @@ buildAndInstallUnpackedPackage verbosity
     annotateFailure mlogFile BuildFailed $
       setup buildCommand buildFlags
 
+    -- Haddock phase
+    whenHaddock $ do
+      when isParallelBuild $
+        notice verbosity $ "Generating " ++ dispname ++ " documentation..."
+      annotateFailureNoLog HaddocksFailed $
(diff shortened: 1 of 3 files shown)


hvr:
@alexbiehl does `--enable-documentation` result in a different nix-hash? i.e. does it trigger recompilations of existing packages in the store, or does this result in packages missing docs in the store *not* being recompiled w/ haddocks?

alexbiehl:
These are the inputs for the nix package hash originating from configuration values:

https://github.com/haskell/cabal/blob/4a6fb38ddd6b6987216b8ee9ce778e5c5396021c/cabal-install/Distribution/Client/PackageHash.hs#L189-L220

`documentation` is not part of that input and thus `documentation: true` doesn't affect the hash.

23Skidoo:
Would be cool to later add some kind of automatic doc rebuilding like we have for `--enable-profiling`, but I agree that this shouldn't be part of the package hash.

23Skidoo:
Merged, thanks!

Issue #5145, opened 2018-02-14
SPDX 3.0


phadej:
As @peti pointed to me at https://github.com/phadej/spdx/issues/10

> Just a quick heads-up: The SPDX license list https://spdx.org/licenses/ was updated to version 3.0 and I understand that there were a couple of surprising changes with regard to GPL etc. It would be nice to see those modifications reflected in this library.

PR #5146, opened 2018-02-14
Update SPDX license list to 3.0 (28 December 2017)

phadej:
Resolves https://github.com/haskell/cabal/issues/5145

(Note: the spec is 2.1 but license list version is 3.0 - they have different version numbers).

---

### DIFF
--- a/Cabal/Distribution/License.hs
+++ b/Cabal/Distribution/License.hs
@@ -152,11 +152,11 @@ knownLicenses = [ GPL  unversioned, GPL  (version [2]),    GPL  (version [3])
 -- @since 2.2.0.0
 licenseToSPDX :: License -> SPDX.License
 licenseToSPDX l = case l of
-    GPL v | v == version [2]      -> spdx SPDX.GPL_2_0
-    GPL v | v == version [3]      -> spdx SPDX.GPL_3_0
-    LGPL v | v == version [2,1]   -> spdx SPDX.LGPL_2_1
-    LGPL v | v == version [3]     -> spdx SPDX.LGPL_3_0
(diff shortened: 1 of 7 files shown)


phadej:
Question: Should we include deprecated identifiers (e.g. now also `GPL-2.0`) in the list and what to do with them, warn in *cabal check*?

23Skidoo:
Do you want this included in 2.2?

phadej:
@23Skdoo, yes that would be great. If we *want* to include deprecated identifiers too (say leaving out the very old one with trailing `+`), adding a cabal check would be a small thing to do. (matter of hours).

phadej:
To clarify for now we can pretened older SPDX identifier lists never existed. At some point accepting also deprecated identifiers will be a must. The question is
- do we want to accept them all today
- accept only deprecated in 2.5&ndash;3.0 period (i.e. GPL-2.0), given list v3.0 is quite new
- or don't accept any deprecated identifiers

23Skidoo:
IMO accepting only 3.0+ identifiers makes sense, but we must make sure that entering `GPL-2.0` produces a good error message recommending the use of `GPL-2.0-only` or `GPL-2.0-or-later` instead. Perhaps we can use Levenshtein distance for that? Also, we should document somewhere which SPDX version corresponds to which Cabal spec version.

23Skidoo:
I'm going to create the 2.2 branch now, but will cherry-pick this to the release branch once it's ready.

phadej:
@23Skidoo pushed an update

phadej:
ping @23Skidoo travis is green

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.2.

PR #5157, opened 2018-02-21
bootstrap: Pass SCOPE_OF_INSTALL to ghc when building Setup.hs

DanielG:
Now that Cabal depends on parsec bootstrap.sh should compile its `Setup.hs` with the bootstrapping libraries' package-db in scope, otherwise users will be forced to manually install the correct version of parsec etc. to even get the bootstrap script to work!


23Skidoo:
Merged, thanks!

PR #5163, opened 2018-02-24
Convert Cabal/changelog to markdown [skip ci]

hvr:
Notational conventions used:

 - `----` separators to group major version series
 - `#` headings to highlight major version bumps,
 - `##` headings for minor version bumps, and
 - `###` headings for patch-levels


hvr:
I haven't fixed up places which may use single quotes instead of backquotes or other places which use sub-optimal markup; but I can do a second pass if this is what we want to go for.

Cabal/ChangeLog.md
```diff
@@ -0,0 +1,784 @@
...
+
+### 1.6.0.1 Duncan Coutts <duncan@haskell.org> October 2008
+  * Export a compat function to help alex and happy
+
+# 1.6.0.0 Duncan Coutts <duncan@haskell.org> October 2008
+  * Support for ghc-6.10
+  * Source control repositories can now be specified in .cabal files
+  * Bug report URLs can be now specified in .cabal files
+  * Wildcards now allowed in data-files and extra-source-files fields
+  * New syntactic sugar for dependencies "build-depends: foo ==1.2.*"
```
23Skidoo: Does the star here need to be escaped so that the rest of the file is not parsed as italics?

23Skidoo:
[Rendered](https://github.com/hvr/cabal/blob/8bf5a2aa1b589380b52aa88fe02e1dc3c603ef62/Cabal/ChangeLog.md).

Cabal/ChangeLog.md
```diff
@@ -0,0 +1,784 @@
...
+  * `--hyperlink-source` now uses Haddock's hyperlinker backend when
+    Haddock is new enough, falling back to HsColour otherwise.
+  * `D.S.defaultHookedPackageDesc` has been deprecated in favour of
+    `D.S.findHookedPackageDesc` (#4874).
+  * `D.S.getHookedBuildInfo` now takes an additional parameter
+    specifying the build directory path (#4874).
+  * Emit warning when encountering unknown GHC versions (#415).
+  * TODO
+
+### 2.0.1.1 Mikhail Glushenkov <mikhail.glushenkov@gmail.com> December 2017
```
23Skidoo: GitHub's renderer removes angle brackets here for some reason, would be nice to fix this.

Cabal/ChangeLog.md
23Skidoo: Looks like it's only the highlighter that is confused by this.

hvr:
re

> add hyperlinks to ticket references

for that I'd write a simple script which goes over the changelog and expands all unexpanded ticket-references into hyperlinks... this is something I've wanted for other projects as well... :-)

Cabal/ChangeLog.md
hvr: That's a feature of markdown called "autolink"; something like `<url>` or `<email-addr>` is turned into a link w/ the address as label

hvr:
@23Skidoo I've done a 2nd pass w/ the backtick markup

(but I'm not motivated enough to do the ticket-link expansion now... this is boring work :-) )

Cabal/ChangeLog.md
ivan-m: Maybe then it's worth doing more "normal" Markdown links `[Name](email)`? Slightly less readable as text, more readable when rendered?

hvr:
Also cherry-picked to 2.2 branch via c66a118dab462423c0195f8c9cc92d7945cd3fe2

Issue #4976, opened 2017-12-26
Solver needs around 20000 backjumps to solve for servant-mock-0.8.3


phadej:
With GHC-7.8.4 or GHC-7.10.3 and
```
--index-state='2017-12-25T17:31:19Z'
```

This is probably due recently released `ansi-terminal-0.8`, which is quite early in the dependency-graph:

[dependency-graph.pdf](https://github.com/haskell/cabal/files/1586303/deps.pdf) with `ansi-terminal` revdeps highlighted

cc @grayjay you might like to solve this puzzle!

grayjay:
Does this issue require `--enable-tests`?  I was able to produce a failure by checking out version 0.8.3 of servant-mock and running this command:

`cabal new-build --dry-run --index-state='2017-12-25T17:31:19Z' --enable-tests --max-backjumps=20000`

I think that the conflict between ansi-wl-pprint and the new version of ansi-terminal is the problem.  The solver chooses ansi-terminal first and then chooses ansi-wl-pprint after following a chain of dependencies (servant-server, wai-app-static, optparse-applicative, ansi-wl-pprint).  Here is a common pattern from the log:

```
[181] trying: servant-server-0.12 (dependency of servant-mock)
[182] trying: servant-server:setup.base~>base-4.8.2.0/installed-0d6... (dependency of servant-server)
[183] trying: wai-app-static-3.1.6.1 (dependency of servant-server)
[184] trying: optparse-applicative-0.14.0.0 (dependency of wai-app-static)
[185] next goal: ansi-wl-pprint (dependency of optparse-applicative)
[185] rejecting: ansi-wl-pprint-0.6.8.1 (conflict: ansi-terminal==0.8, ansi-wl-pprint => ansi-terminal>=0.4.0 && <0.8)
[185] rejecting: ansi-wl-pprint-0.6.8, ansi-wl-pprint-0.6.7.3, ansi-wl-pprint-0.6.7.2, ansi-wl-pprint-0.6.7.1, ansi-wl-pprint-0.6.7, ansi-wl-pprint-0.6.6 (conflict: ansi-terminal==0.8, ansi-wl-pprint => ansi-terminal>=0.4.0 && <0.7)
[...]

Issue #4594, opened 2017-07-08
Print stats about solver performance costs that don't vary by run, such as backjump count.


grayjay:
See https://github.com/haskell/cabal/pull/4562#issuecomment-310889879.

grayjay:
One option is to print the total number of nodes that the solver explores.  I think that would correlate more closely with run time than the backjump count, because the solver can traverse any number of levels during one backjump.

hvr:
The backjump count is still interesting though, as that value corresponds to a parameter the user can configure. Currently, we have no good way to figure out which backjump limit to set manually (without having to use an infinite upper limit), when the default one doesn't suffice to find a solution.

PR #5166, opened 2018-02-25
Enforce the backjump limit before creating the solver log. 

grayjay:

________________________________________________________________________

This PR enforces the backjump limit when the solver explores the search tree, which could simplify implementing features such as #2860 or #4594.  It also skips processing the log when the log doesn't need to be printed, since the log is no longer needed for enforcing the backjump limit.  The main commits are:

#### Enforce the backjump limit before creating the solver log.

Previously, the solver enforced the backjump limit by counting `Backjump`
messages during log processing.  This commit instead enforces the backjump limit
when the solver explores the search tree.  This change simplifies the log
processing step and allows us to skip creating the log when we don't need to
print it, though that's not implemented in this commit.

This commit changes the meaning of the backjump count slightly.  Previously the
solver counted a backjump every time it backtracked and added at least one
variable to the conflict set.  Now it counts a backjump every time it unions all
conflict sets at the current level, even if the union is the same as the
conflict set from the level below.  That means that more steps count as
backjumps, and the solver can reach the backjump limit slightly earlier.

This commit also changes the log messages.  Previously, the solver sometimes
omitted the last `Backjump` message, but this commit always prints a `Backjump`
message when the solver fails.  Additionally, the level numbers for some
`Backjump` messages changed to match the levels where the conflict sets were
unioned.

#### Skip processing the solver log when the log isn't needed.

I tested this change by comparing performance with both
c01d92f (before any refactoring needed to
implement this change) and 4d28102 (the
previous commit), using hackage-benchmark.  Since the refactoring changed the
meaning of the backjump count, I only timed "install" commands that didn't reach
the backjump limit.  I chose several commands that ran for different amounts of
time, including the long running example from issue #4976.

Comparing c01d92f (cabal1) and this commit (cabal2):

compiler: GHC 8.2.1
index state: `2018-02-16T02:47:32Z`
extra hackage-benchmark flags:
`--packages="aeson yesod wolf" --pvalue=0.01 --trials=50 --print-skipped-packages`

```
package result1       result2             mean1       mean2     stddev1     stddev2     speedup
aeson   (not significant)
yesod   Solution      Solution           2.270s      2.258s      0.027s      0.031s      1.005
wolf    Solution      Solution           7.337s      7.234s      0.056s      0.047s      1.014
```

Comparing 4d28102 (cabal1) and this commit (cabal2) with the same environment
and flags as above:

```
package result1       result2             mean1       mean2     stddev1     stddev2     speedup
aeson   (not significant)
yesod   (not significant)
wolf    Solution      Solution           7.297s      7.245s      0.045s      0.048s      1.007
```

hackage-benchmark currently doesn't print the results when they aren't
significant, so I reran "`cabal install --dry-run aeson`", and it ran for about
1.6 seconds and found a solution.

Comparing c01d92f (cabal1) and this commit (cabal2) on issue #4976:

compiler: GHC 7.10.3
extra hackage-benchmark flags:
`--cabal1-flags="--index-state='2017-12-25T17:31:19Z' --enable-tests --max-backjumps=-1"
--cabal2-flags="--index-state='2017-12-25T17:31:19Z' --enable-tests --max-backjumps=-1"
--packages=servant-mock --pvalue=0.01 --trials=50 --print-skipped-packages --print-trials`

```
trial/summary    package      result1       result2             mean1       mean2     stddev1     stddev2     speedup
...
summary          servant-mock Solution      Solution          39.693s     38.863s      0.181s      0.174s      1.021
```

Comparing 4d28102 (cabal1) and this commit (cabal2) on issue #4976 with the
same environment and flags as above:

```
trial/summary    package      result1       result2             mean1       mean2     stddev1     stddev2     speedup
...
summary          servant-mock Solution      Solution          39.659s     38.960s      0.195s      0.162s      1.018
```

Overall, this seems like a very small performance improvement.

### DIFF
--- a/cabal-install/Distribution/Solver/Modular.hs
+++ b/cabal-install/Distribution/Solver/Modular.hs
@@ -115,28 +115,28 @@ solve' :: SolverConfig
        -> Set PN
        -> Progress String String (Assignment, RevDepMap)
 solve' sc cinfo idx pkgConfigDB pprefs gcs pns =
-    foldProgress Step createErrorMsg Done (runSolver sc)
+    foldProgress Step (uncurry createErrorMsg) Done (runSolver sc)
   where
     runSolver :: SolverConfig
-              -> Progress String SolverFailure (Assignment, RevDepMap)
+              -> Progress String (SolverFailure, String) (Assignment, RevDepMap)
     runSolver sc' =
         logToProgress (solverVerbosity sc') (maxBackjumps sc') $ -- convert log format into progress format
         solve sc' cinfo idx pkgConfigDB pprefs gcs pns
 
-    createErrorMsg :: SolverFailure
+    createErrorMsg :: SolverFailure -> String
                    -> Progress String String (Assignment, RevDepMap)
-    createErrorMsg (NoSolution cs        msg) =
+    createErrorMsg (ExhaustiveSearch cs _) msg =
         Fail $ rerunSolverForErrorMsg cs ++ msg
-    createErrorMsg (BackjumpLimitReached msg) =
+    createErrorMsg BackjumpLimitReached    msg =
         Step ("Backjump limit reached. Rerunning dependency solver to generate "
               ++ "a final conflict set for the search tree containing the "
               ++ "first backjump.") $
-        foldProgress Step f Done $
+        foldProgress Step (f . fst) Done $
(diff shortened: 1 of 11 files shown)


kosmikus:
Sorry, hadn't seen this any earlier. All sounds reasonable to me. Thanks!

grayjay:
Thanks!

PR #5170, opened 2018-02-26
Look for transitive setup dependency on Cabal when choosing Cabal spec version.

### LINKED ISSUE #4288: new-build failure
Reproducible with ghc-8.0.1 and ghc-8.0.2

```
distributive-0.5.2-e10f1d2339ae8fc2ff95360c5e91f0ace4f1e27add106f7939389ad5ec9ca0aa.
The failure occurred during the final install step. The exception was:
user error ('/opt/ghc/8.0.1/bin/ghc-pkg' exited with an error:
distributive-0.5.2: Warning: Unrecognized field indefinite on line 16
distributive-0.5.2: Warning: haddock-interfaces:
/home/ogre/.cabal/store/ghc-8.0.1/distributive-0.5.2-e10f1d2339ae8fc2ff95360c5e91f0ace4f1e27add106f7939389ad5ec9ca0aa/share/doc/html/distributive.haddock
doesn't exist or isn't a file
distributive-0.5.2: Warning: haddock-html:
/home/ogre/.cabal/store/ghc-8.0.1/distributive-0.5.2-e10f1d2339ae8fc2ff95360c5e91f0ace4f1e27add106f7939389ad5ec9ca0aa/share/doc/html
doesn't exist or isn't a directory
distributive-0.5.2: installed package info from too old version of Cabal (key
field does not match id field)
```

### Steps to reproduce

```
cabal get comonod-5
cd comonad-5
cabal new-build --disable-tests -j1
```

/note/ `-j1` isn't necessary

#### part of new-build log building the distributive

```
% /home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/setup/setup register --verbose=2 --builddir=/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2 --inplace --gen-pkg-config=/home/ogre/mess/distributive-0.5.2/dist-newstyle/tmp/package-registration--12792/pkgConf
/opt/ghc/8.0.1/bin/ghc --abi-hash -fbuilding-cabal-package -O -outputdir /home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -odir /home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -hidir /home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -stubdir /home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -i -i/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -isrc -i/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build/autogen -I/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build/autogen -I/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build -optP-include -optP/home/ogre/mess/distributive-0.5.2/dist-newstyle/build/x86_64-linux/ghc-8.0.1/distributive-0.5.2/build/autogen/cabal_macros.h -this-unit-id distributive-0.5.2-3Qb1gotTB50JIJFSZJyPVw -hide-all-packages -no-user-package-db -package-db /home/ogre/.cabal/store/ghc-8.0.1/package.db -package-db /home/ogre/mess/distributive-0.5.2/dist-newstyle/packagedb/ghc-8.0.1 -package-id base-4.9.0.0 -package-id base-orphans-0.5.4-221596738ac95aba5b36b0b0362492907162ea1af94cf8f335da4c27a8bef6e9 -package-id tagged-0.8.5-6a76d83406e0466a505c9569c11ade18b2bb910a6a09df4ba0d355bf7922badf -package-id transformers-0.5.2.0 -package-id transformers-compat-0.5.1.4-a6f22a40e821e7b9a582fac258c4122bb958675f217a786f78176ce395f17184 -XHaskell98 Data.Distributive Data.Distributive.Generic -Wall
Creating package registration file:
/home/ogre/mess/distributive-0.5.2/dist-newstyle/tmp/package-registration--12792/pkgConf
/home/ogre/mess/distributive-0.5.2/dist-newstyle/tmp/package-registration--12792/:
openBinaryTempFileWithDefaultPermissions: does not exist (No such file or
directory)
```

That directory doesn't seem to exist.

grayjay:

______________________________________________________________________

Fixes #4288.

Previously, new-build only looked for a direct dependency on Cabal when choosing
the spec version to use for running the setup script.  When the setup script
only had a transitive dependency on Cabal, cabal used the package's
cabal-version field, which could easily differ from the actual Cabal
dependency's version.  Then cabal could pass flags to the setup script that
weren't recognized by the Cabal dependency.

This commit considers any setup dependency on Cabal when choosing the Cabal spec
version.

/cc @ezyang

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning.hs
@@ -1210,7 +1210,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
             if null not_per_component_reasons
                 then return comps
                 else do checkPerPackageOk comps not_per_component_reasons
-                        return [elaborateSolverToPackage mapDep spkg g $
+                        return [elaborateSolverToPackage spkg g $
                                 comps ++ maybeToList setupComponent]
            Left cns ->
             dieProgress $
@@ -1276,7 +1276,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
                         fsep (punctuate comma reasons)
             -- TODO: Maybe exclude Backpack too
 
-        elab0 = elaborateSolverToCommon mapDep spkg
+        elab0 = elaborateSolverToCommon spkg
         pkgid = elabPkgSourceId    elab0
         pd    = elabPkgDescription elab0
 
@@ -1501,13 +1501,11 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
                                 Just (Just n) -> display n
                                 _ -> ""
 
-    elaborateSolverToPackage :: (SolverId -> [ElaboratedPlanPackage])
-                             -> SolverPackage UnresolvedPkgLoc
+    elaborateSolverToPackage :: SolverPackage UnresolvedPkgLoc
                              -> ComponentsGraph
                              -> [ElaboratedConfiguredPackage]
                              -> ElaboratedConfiguredPackage
     elaborateSolverToPackage
-        mapDep
         pkg@(SolverPackage (SourcePackage pkgid _gdesc _srcloc _descOverride)
                            _flags _stanzas _deps0 _exe_deps0)
         compGraph comps =
@@ -1516,7 +1514,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
         -- of the other fields of the elaboratedPackage.
         elab
       where
-        elab0@ElaboratedConfiguredPackage{..} = elaborateSolverToCommon mapDep pkg
+        elab0@ElaboratedConfiguredPackage{..} = elaborateSolverToCommon pkg
         elab = elab0 {
                 elabUnitId = newSimpleUnitId pkgInstalledId,
                 elabComponentId = pkgInstalledId,
@@ -1613,10 +1611,9 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
               (compilerId compiler)
               pkgInstalledId
 
-    elaborateSolverToCommon :: (SolverId -> [ElaboratedPlanPackage])
-                            -> SolverPackage UnresolvedPkgLoc
+    elaborateSolverToCommon :: SolverPackage UnresolvedPkgLoc
                             -> ElaboratedConfiguredPackage
-    elaborateSolverToCommon mapDep
+    elaborateSolverToCommon
         pkg@(SolverPackage (SourcePackage pkgid gdesc srcloc descOverride)
                            flags stanzas deps0 _exe_deps0) =
         elaboratedPackage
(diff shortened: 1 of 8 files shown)


hvr:
This is a bit of a coincidence... I'll use this opportunity to ask something... :-)

Yesterday I was looking into how I could inject a qualified per-package constraint on the lower bound matching the declared `cabal-version:` of a package for its custom-setup component (i.e.  (PkgConstraint (ScopeQual (QualSetup "foo") "Cabal") ... )) ... and I was going to ask you about what'd be the best way do this... (I was considering extending the `PInfo` field with a version-field for recording the specVersion during index conversion, and go from there... not sure if that's the best approach). Any suggestion? :-)

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1687,10 +1684,9 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
         elabRegisterPackageDBStack = buildAndRegisterDbs
 
         elabSetupScriptStyle       = packageSetupScriptStyle elabPkgDescription
-        -- Computing the deps here is a little awful
-        deps = fmap (concatMap (elaborateLibSolverId mapDep)) deps0
```
grayjay: I'm not completely sure I calculated the closure of the setup dependencies correctly, since I skipped calling `elaborateLibSolverId mapDep`.  I called `CD.setupDeps` on `deps0` directly, because all setup dependencies are library dependencies.

Ericson2314:
sounds good, but the name through me off haha. We are looking at `setup-depends . build-depends*` not `setup-depends*`, right?

grayjay:
@hvr I think that extending PInfo with the cabal-version makes sense.  I'm not sure of the best way to enforce the constraint, though.  I think it would need to somehow hook into the validation phase, in https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Solver/Modular/Validate.hs.  In that case, it might be better to add the constraint to the FlaggedDeps in the PInfo (using the Dep type), similar to a setup dependency.  Dep is currently defined as:

```haskell
-- | A dependency (constraint) associates a package name with a constrained
-- instance. It can also represent other types of dependencies, such as
-- dependencies on language extensions.
data Dep qpn = Dep (Maybe UnqualComponentName) qpn CI  -- ^ dependency on a package (possibly for executable)
             | Ext  Extension                          -- ^ dependency on a language extension
             | Lang Language                           -- ^ dependency on a language version
             | Pkg  PkgconfigName VR                   -- ^ dependency on a pkg-config package
deriving Functor
```

I think that the Cabal constraint would need to be handled slightly differently than the existing Dep data constructor, since it would constrain but not require Cabal (and it would need a different error message when it can't be satisfied).  One way to handle the constraint would be to add a Constraint data constructor to Dep that is ignored in the tree building phase (https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Solver/Modular/Builder.hs) but is treated like a Dep in the validation phase.

@Ericson2314 I'm not sure I understand.  What is `setup-depends . build-depends*`?  This PR only affects the choice of the elabSetupScriptCliVersion field in ElaboratedConfiguredPackage for new-build: https://github.com/haskell/cabal/blob/d504388c1c1183a64671aff106d1ce3a649d51c0/cabal-install/Distribution/Client/ProjectPlanning/Types.hs#L282-L285

Ericson2314:
That is relation notation. We are looking at the reflexive-transitive closure of build-inputs joined with setup depends, not reflexive-transitive closure of setup depends.

grayjay:
@Ericson2314 Yes, it isn't following multiple levels of setup-depends.

PR #5171, opened 2018-02-26
Fix split-sections flag

### LINKED ISSUE #5011: --enable-split-sections doesn't do anything
I was reading this commit https://github.com/bgamari/cabal/commit/450d6bc4d8dda8e4c56390833cec34f4b34149df , because someone contacted me to see if I was favorable about adding more direct support for split-sections in stack.

I noticed that [this line](https://github.com/bgamari/cabal/commit/450d6bc4d8dda8e4c56390833cec34f4b34149df#diff-6c1f0428958c136d788ac9fa6a9452b1R356) looks like it contains a copy paste error.  In that change, and still on master, it is

```haskell
  , [ "-split-sections" | flagBool ghcOptSplitObjs ]
  , [ "-split-objs" | flagBool ghcOptSplitObjs ]
```

and not

```haskell
  , [ "-split-sections" | flagBool ghcOptSplitSections ]
  , [ "-split-objs" | flagBool ghcOptSplitObjs ]
```

It looks like that change also intends to make split sections a default, but that also will not happen with the current code.

Also, if I'm reading the change correctly, it seems like with >= ghc 8 and no `--disable-split-sections`, specifying split-objs will cause it to always warn about mixing split-sections with split-objs even though the user did not specify split-sections.

bgamari:
The patch which introduced support for split-sections, 450d6bc, contained a
cut-and-paste error such that it was controlled by the split-objs flag. Fix
this.

Fixes #5011.

### DIFF
--- a/Cabal/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/Distribution/Simple/Program/GHC.hs
@@ -354,7 +354,7 @@ renderGhcOptions comp _platform@(Platform _arch os) opts
         | flagProfAuto implInfo -> ["-fprof-auto-exported"]
         | otherwise             -> ["-auto"]
 
-  , [ "-split-sections" | flagBool ghcOptSplitObjs ]
+  , [ "-split-sections" | flagBool ghcOptSplitSections ]
   , [ "-split-objs" | flagBool ghcOptSplitObjs ]
 
   , case flagToMaybe (ghcOptHPCDir opts) of


mgsloan:
@23Skidoo Hmm, if I understand correctly, you are suggesting preserving compatibility with the buggy behavior?  I don't see the point of that for the following reasons:

1.  People who specified `split-objs: True` probably assumed it was working and wanted it.  They may not have realized

2. AFAICT, ideally `split-objs: True` should not affect buildability and be essentially transparent for runtime behavior.  So I don't see any good reason to retain compatibility with broken behavior.  Instead, the behavior should be fixed to correspond to what the configuration was asking for.

23Skidoo:
@mgsloan I'm thinking only about the `cabal-install` side of things. I think that when someone puts `split-objs: True` in `~/.cabal/config`, it's normally because they want smaller binaries, and don't really care about the concrete mechanism used to get this result. If they really want `-split-objs`, I think that they can use `ghc-options: -split-objs`.

The only problem is that it gets a bit confusing when you have `split-sections: True` in `~/.cabal/config` and use a GHC that only supports `split-objs`, in which case I think it should default to `split-objs` (and maybe also print a warning).

>  AFAICT, ideally `split-objs: True` should not affect buildability and be essentially transparent for runtime behavior.

Agreed, but I only suggested printing a warning. Something along the following lines:

```
$ cat ~/.cabal/config
[...]
split-objs: True
$ cabal new-build
Warning: The 'split-objs' config file option is deprecated in favour of 'split-sections'.
Defaulting to 'split-sections: True'.

$ cat ~/.cabal/config
[...]
split-objs: False
$ cabal new-build
Warning: The 'split-objs' config file option is deprecated in favour of 'split-sections'.

$ cat ~/.cabal/config
[...]
-- split-objs: False
$ cabal new-build
$ # No warning
```

mgsloan:
Hmm, sure, that seems reasonable!

23Skidoo:
Travis failure seems unrelated, merging.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.2.

PR #5173, opened 2018-02-26
Fail parsing -none or ^>= if old cabal-version

phadej:


phadej:
`check-tests` are "broken". Now already parser fails, we don't wait until `cabal check`. Out of two options

- Downgrade error to warning
- Make error text similar to what's check now, and update check tests

I prefer the latter.

hvr:
I prefer

> Make error text similar to what's check now, and update check tests

as well.

IMO, `cabal check` should be only used for things we can't easily catch as errors early during parsing or which make little sense to catch during parsing.

phadej:
@hvr, @23Skidoo is this good to merge?

hvr:
@phadej looks good to me

I've got one question though: is it *easily* possible to detect the use of `-none` at `cabal check` time?

phadej:
@hvr currently no. I'll make it for *next-after-2.2* to represent `-any` and -none` in the AST too. (as we do with wildcard and major bounds and even parens).

PR #5180, opened 2018-03-03
Last part of addressing #415

hvr:
See commit comments; this completes the work for addressing #415 
(NB: This PR only touches `cabal-install`)

Issue #415, opened 2012-05-24
Cabal should warn if using a version of ghc that is newer than Cabal knows about


bos:
(Imported from [Trac #422](http://hackage.haskell.org/trac/hackage/ticket/422), reported by @dcoutts on 2008-12-01)

It seems to often happen that newer versions of ghc make changes that break old versions of Cabal. This is particularly a problem for <tt>cabal-install</tt> since that need not necessarily be upgraded when ghc is upgraded.

Probably it should warn if the major version is bigger than the known version at the time.

phadej:
Let's reopen this, people use `CABALVER=1.24 GHCVER=8.2.1` in CI, and it works by accident. I don't think this is hard issue to fix, if someone (maybe new contributor) picks this up!

ezyang:
Hey @phadej, I'd like the newcomer tag to refer to tickets which have instructions for how to get started (e.g., where to look in the codebase, etc.) Maybe you could fill those out and then retag this one?

jneira:
hi, from the commits and prs linking this issue I would say it can be closed (specially for #5194 which added just what was requested originally in the issue)

PR #5183, opened 2018-03-03
Last part of addressing #415 (fwd-ported)

hvr:
This is a forward-port of #5180 / 7ce486e6553d24877ae3b23868b95d20bf83760a

### DIFF
--- a/cabal-testsuite/PackageTests/Regression/T4154/install-time-with-constraint.test.hs
+++ b/cabal-testsuite/PackageTests/Regression/T4154/install-time-with-constraint.test.hs
@@ -9,7 +9,9 @@ main = cabalTest $ withRepo "repo" $ do
 
   -- Temporarily disabled recording here because output is not stable
   recordMode DoNotRecord $ do
-      -- Constraining all uses of 'time' results in a cyclic dependency
-      -- between 'Cabal' and the new 'time'.
-      r <- fails $ cabal' "new-build" ["time", "--constraint=any.time==99999", "--dry-run"]
-      assertOutputContains "cyclic dependencies; conflict set: time:setup.Cabal, time:setup.time" r
+      r <- fails $ cabal' "new-build" ["time", "--constraint=any.time==99999", "--constraint=setup.Cabal installed", "--dry-run"]
+      -- Constraining all uses of 'time' originally resulted in a cyclic dependency
+      -- between 'Cabal' and the new 'time':
+      -- assertOutputContains "cyclic dependencies; conflict set: time:setup.Cabal, time:setup.time" r
+      -- However, this doesn't work anymore, so instead we more directly look for:
+      assertOutputContains "time:setup.time~>time-99999 (conflict: time:setup.Cabal" r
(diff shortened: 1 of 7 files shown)


cabal-testsuite/PackageTests/Regression/T4154/install-time-with-constraint.test.hs
```diff
@@ -9,7 +9,9 @@ main = cabalTest $ withRepo "repo" $ do
 
   -- Temporarily disabled recording here because output is not stable
   recordMode DoNotRecord $ do
-      -- Constraining all uses of 'time' results in a cyclic dependency
-      -- between 'Cabal' and the new 'time'.
-      r <- fails $ cabal' "new-build" ["time", "--constraint=any.time==99999", "--dry-run"]
-      assertOutputContains "cyclic dependencies; conflict set: time:setup.Cabal, time:setup.time" r
+      r <- fails $ cabal' "new-build" ["time", "--constraint=any.time==99999", "--constraint=setup.Cabal installed", "--dry-run"]
```
grayjay: Why does this command need the `setup.Cabal installed` constraint now?  It seems like cabal can only choose the installed Cabal library now that it enforces a max setup Cabal version.  I thought that the two new-build commands should only differ by the constraint scope, to show that the constraint on `time` succeeds, but the constraint on `any.time` fails.

cabal-testsuite/PackageTests/Regression/T4154/install-time-with-constraint.test.hs
```diff
@@ -9,7 +9,9 @@ main = cabalTest $ withRepo "repo" $ do
...
+      -- Constraining all uses of 'time' originally resulted in a cyclic dependency
+      -- between 'Cabal' and the new 'time':
+      -- assertOutputContains "cyclic dependencies; conflict set: time:setup.Cabal, time:setup.time" r
+      -- However, this doesn't work anymore, so instead we more directly look for:
+      assertOutputContains "time:setup.time~>time-99999 (conflict: time:setup.Cabal" r
```
grayjay: I'm not sure I understand this assertion.  Shouldn't it check the line that rejects the installed time due to the command line flag, now that the cycle isn't the cause of the failure?

cabal-testsuite/PackageTests/Regression/T4154/time.cabal
```diff
@@ -4,7 +4,7 @@ cabal-version:   >=1.8
 build-type:      Custom
 
 custom-setup
-  setup-depends: base, Cabal == 99999
```
grayjay: I think that Cabal-99999 can be deleted now that it doesn't fit the implicit setup constraints.

cabal-testsuite/PackageTests/Regression/T4154/install-time-with-constraint.test.hs
hvr: @grayjay 

I was mostly paranoid and wanted the resulting output to be as deterministic as possible but I realize this wasn't probably the most principled way to retain the spirit of this test... 

...and I gladly take your offer on changing the tests after the PR is merged, as you clearly have a better understanding on how to write this regression test in an effective way. :-)

grayjay: I made a PR: #5186

Issue #680, opened 2012-05-24
adopt XDG basedir spec


bos:
(Imported from [Trac #688](http://hackage.haskell.org/trac/hackage/ticket/688), reported by guest on 2010-05-12)

I want to be able to tell <tt>cabal</tt> to read the configuration from a different place than <tt>$HOME/.cabal/config</tt>.

The ultimate goal is to get rid of the non-standard <tt>.cabal</tt> directory altogether, so far careful <tt>--package-db</tt>/<tt>--prefix</tt> etc. juggling served me well; <tt>config</tt> is the last remaining file there.

<pre class="wiki">&gt; cabal --version
cabal-install version 0.8.2
using version 1.8.0.2 of the Cabal library
</pre>

Earnestly:
The only difference is by needing `CABAL_CONFIG` we add to our already bloated
`environ`s.  For reference here are a few:

```
export SLRNHOME=$XDG_CONFIG_HOME/slrn
export _RR_TRACE_DIR=$XDG_DATA_HOME/rr
export CARGO_HOME=$XDG_DATA_HOME/cargo
export GNUPGHOME=$XDG_CONFIG_HOME/gnupg
export INPUTRC=$XDG_CONFIG_HOME/inputrc
export RLWRAP_HOME=$XDG_DATA_HOME/rlwrap
export LYNX_CFG_PATH=$XDG_CONFIG_HOME/lynx
export DVDCSS_CACHE =$XDG_CACHE_HOME/dvdcss
export TIGRC_USER=$XDG_CONFIG_HOME/tig/tigrc
export WEECHAT_HOME=$XDG_CONFIG_HOME/weechat
export PERL_CPANM_HOME=$XDG_CACHE_HOME/cpanm
export GIMP2_DIRECTORY=$XDG_CONFIG_HOME/gimp
export ELINKS_CONFDIR=$XDG_CONFIG_HOME/elinks
export CABAL_CONFIG=$XDG_DATA_HOME/cabal/config
export HTTPIE_CONFIG_DIR=$XDG_CONFIG_HOME/httpie
export LESSHISTFILE=$XDG_CACHE_HOME/less/history
export XAUTHORITY=$XDG_RUNTIME_DIR/x11/xauthority
export XCOMPOSEFILE=$XDG_CONFIG_HOME/x11/xcompose
export GUILE_HISTORY=$XDG_CONFIG_HOME/guile/history
[...]

PR #5188, opened 2018-03-05
Fix new-install symlink location

### LINKED ISSUE #5000: new-install symlink location
Currently new-install errors out without additional flags because it uses `symlink-bindir` as symlink location, which is unset by default. I actually am not sure what that flag was supposed to do originally (can someone clarify? Is the current behavior the intended one?).

After some digging I found that `bindir` is the only flag that is automatically set to `~/.cabal/bin` in the config file.

Now there are two courses of action:
1. set `symlink-bindir` to `~/.cabal/bin` and equivalents on config creation
2. use `bindir` instead

The downside of (1) is that existing configurations will not have that flag set and will give an error. edit: well, I just discovered `cabal user-config update`
The downside of (2) is that the `bindir` flag serves an obvious purpose for packagers, and using it for the symlink and not the executable violates that.

I think (1) is the correct one, maybe with a default to address the empty flag problem, but I'd like to hear some opinions first.

We should also document `symlink-bindir` [here](https://cabal.readthedocs.io/en/latest/installing-packages.html#installation-paths)

fgaz:
Now new-install(-nonlocal-exes) should work without issues (if the ~/.cabal/config was generated with this)

Closes #5000 (yay!)

@23Skidoo can we cherry-pick it on 2.2 too?

  * [x] the symlink-bindir default
  * [x] https://cabal.readthedocs.io/en/latest/installing-packages.html#installation-paths
    * Not necessary since those are Cabal/Setup.hs paths

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -58,7 +58,8 @@ import Distribution.Types.UnqualComponentName
 import Distribution.Verbosity
          ( Verbosity, normal )
 import Distribution.Simple.Utils
-         ( wrapText, die', withTempDirectory, createDirectoryIfMissingVerbose )
+         ( wrapText, die', notice
+         , withTempDirectory, createDirectoryIfMissingVerbose )
 
 import qualified Data.Map as Map
 import System.Directory ( getTemporaryDirectory, makeAbsolute )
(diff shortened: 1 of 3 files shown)


23Skidoo:
Sure, if it's new-build only, then it's low-risk.

fgaz:
Yup. new-build _and_ I set `symlink-bindir` (which is only used by new-install despite being there for literally 10 years) to `~/.cabal/bin` by default in `~/.cabal/config`

hvr:
I don't give up hope on #680 ... maybe for 3.0? :-)

fgaz:
Merging this now for the same reason as #4825 
(travis was green. I stopped it because i forgot [ci skip] when force pushing the changelog)

fgaz:
@23Skidoo fyi I just pushed it on 2.2 too

PR #5193, opened 2018-03-08
Look for transitive setup dependency on Cabal when choosing Cabal spec version. (2.2)

grayjay:
#5170 cherry-picked onto 2.2.

@23Skidoo Do you think we should apply this to 2.2?  It only affects new-build.


hvr:
I'm generally for including this in 2.2 since it fixes an issue, i.e. #4288

23Skidoo:
Sure, why not.

grayjay:
Thanks, I'll merge it.

PR #5196, opened 2018-03-08
Make GenericPackageDescription lenses more maintainable

Ericson2314:
Record syntax is easier to read and more robust against refactors.

@phadej you were saying the old one was autogenerated, but is there a way to automatically regenerate it? If not, I think this is better.

### DIFF
--- a/Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
+++ b/Cabal/Distribution/Types/GenericPackageDescription/Lens.hs
@@ -11,6 +11,8 @@ import Prelude()
 import Distribution.Compat.Prelude
 import Distribution.Compat.Lens
 
+import qualified Distribution.Types.GenericPackageDescription as T
+
 -- We import types from their packages, so we can remove unused imports
 -- and have wider inter-module dependency graph
 import Distribution.Types.CondTree (CondTree)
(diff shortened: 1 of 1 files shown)


phadej:
No need to tweak autogenerated code. Easier just to regenerate everything.

phadej:
In fact there is way to generate it: http://hackage.haskell.org/package/Cabal-2.2.0.0/docs/Distribution-Compat-Lens.html#g:11

(This PR looks like it's generated, but I close it out of principle, if it wasn't)

Ericson2314:
@phadej ah OK. I had no idea that existed, and was confused when you commented on this commit before.

Ericson2314:
I'm reopening this with a blocked label to remind me to a) fix lens generation b) regenerate automatically.

Ericson2314:
Ok @phadej. Was able to redo this now that the instructions are updated. The function spit out the lenses in a different order so I kept that order: it's a bigger diff now, but smaller diffs going forward.

Ericson2314:
One of the OSX builds timed out, but it's completely absurd that this would exhibit os+ghc-version-specific behavior, so I'm going to merge now before the restarted job completes.

PR #5197, opened 2018-03-09
Fix powershell transport

Mistuke:

This fixes the powershell transport protocol. The reason for the breakage is because certain headers like
"Range" are protected in the .NET framework https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest(v=vs.110).aspx which means you can't set them purely as metadata and need to go through the proper getter/setters.

### DIFF
--- a/cabal-install/Distribution/Client/HttpUtils.hs
+++ b/cabal-install/Distribution/Client/HttpUtils.hs
@@ -642,7 +642,7 @@ powershellTransport prog =
                                              then break (== '-') value'
                                              else error $ "Could not decode range: " ++ value
                                          value' = drop 6 value
-                                     in "AddRange(" ++ escape start ++ ", " ++ escape end ++ ");"
+                                     in "AddRange(\"bytes\", " ++ escape start ++ ", " ++ escape end ++ ");"
               name                -> "Headers.Add(" ++ escape (show name) ++ "," ++ escape value ++ ");"
 
     setupAuth auth =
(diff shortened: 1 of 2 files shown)


23Skidoo:
We should add some AppVeyor tests so that we get notified if this breaks again. Note that AppVeyor has `curl` in `PATH` and so chooses the `curl` transport by default. You can force the PowerShell transport with the `--http-transport` option.

hvr:
@23Skidoo is it enough to change the `cabal update` line in the `appveyor.yml` script to say `cabal --http-transport=powershell update -v`, or did you have something more extensive in mind?

23Skidoo:
If @Mistuke can come up with more extensive tests, that'd be appreciated, but I won't ask for anything more.

cabal-install/Distribution/Client/HttpUtils.hs
```diff
@@ -545,16 +545,37 @@ powershellTransport prog =
...
+          [ "$response = $request.GetResponse()"
+          , "$responseStream = $response.GetResponseStream()"
+          , "$buffer = new-object byte[] 10KB"
+          , "$count = $responseStream.Read($buffer, 0, $buffer.length)"
+          , "while ($count -gt 0)"
+          , "{"
+          , "    $targetStream.Write($buffer, 0, $count)"
+          , "    $count = $responseStream.Read($buffer, 0, $buffer.length)"
+          , "}"
           , "Write-Host \"200\";"
```
hvr: The unconditional "200" seems wrong

23Skidoo:
Merged, thanks!

PR #5199, opened 2018-03-09
Add OS alias for Android (linux-android)

kmicklas:

### DIFF
--- a/Cabal/Distribution/System.hs
+++ b/Cabal/Distribution/System.hs
@@ -124,6 +124,7 @@ osAliases Permissive FreeBSD = ["kfreebsdgnu"]
 osAliases Compat     FreeBSD = ["kfreebsdgnu"]
 osAliases Permissive Solaris = ["solaris2"]
 osAliases Compat     Solaris = ["solaris2"]
+osAliases _          Android = ["linux-android"]
 osAliases _          _       = []
 
 instance Pretty OS where


phadej:
cc @hvr, We'll have to teach `cabal-version` to `OS` `parsec`.

kmicklas:
@23Skidoo that would be much appreciated. I've started writing `if os(linux-android)` and I'd like to avoid that spreading too far into the wild when we could have `if os(android)` instead.

phadej:
I'm confused, why add `linux-android` alias if we have `android` (which is preferred)?

hvr:
What's the motivation/rationale for introducing this alias btw? (note, we also need to make sure to version gate this, as older cabal versions wouldn't understand it, which can easily result in not well defined behaviour when compiling such packages).

23Skidoo:
@phadej 
> I'm confused, why add `linux-android` alias if we have `android` (which is preferred)?

It's used to implement `classifyOS`, without this change `linux-android` is parsed as `OtherOS "linux-android"` instead of `Android`.

kmicklas:
@phadej @hvr  Yeah to be honest, I'm slightly confused myself about the intention of this existing alias mechanism. It appears to be used both for parsing conditions in the cabal file (e.g. `if os(linux)`) and for parsing the platform string passed in from the external environment (e.g. `i686-unknown-linux`).

My motivation for this change is really the second part - it took me longer than it should have to figure out that I needed `if os(linux-android)`, and `if os(android)` was the obvious thing to try first. Personally I wish we could ban people from writing the obscure platform strings passed from above in their cabal files, but if `if os(linux-android)` is out there in the wild it seems reasonable that we could continue to support that as well.

23Skidoo:
I believe that with this change `if os(linux-android)` will continue to work as before - both the user-specified string and the one that comes from the environment (be that `"android"` or `"linux-android"`) will be parsed as `Android :: OS`. The change simply means that packages using `if os(android)` will no longer fail to build in Android environments that describe themselves as `linux-android`, so can be considered a bugfix.

phadej:
Shouldn't we use Compat/Strict/Permissive modifiers then?

Ericson2314:
In wild LLVM-style target triple land, android is an ABI for linux, and it would be nice to do that in Cabal and GHC eventually. This seems like its moving in that direction.

23Skidoo:
> Shouldn't we use Compat/Strict/Permissive modifiers then?

You mean that `if os(linux-android)` should trigger a warning?

Ericson2314:
@23Skidoo But if anything I'd think `if os(android)` should be the warning...

angerman:
Does this add support for `os(ios)` as well? What about `os(tvos)`, should we maybe just try to overhaul them all in one go?

>  @angerman was going to remove ios from GHC IIRC, and so I'm curious what happened to that.

As far as GHC is concerned, Android is Linux, and iOS is darwin. There should be no conditional logic be left that discriminates specifically for iOS or Android in GHC. (See https://phabricator.haskell.org/D3579)

23Skidoo:
@angerman We already had `os(ios)`, it may make sense to add a `tvos` alias for `ios`.

kmicklas:
After digging around a bit more in GHC I realized the situation is pretty complicated. GHC's `aclocal.m4` canonicalizes the platform triples passed from outside to an internal standard which is what Cabal is parsing here. Up until GHC 8.2 the specific iOS triples like `aarch64-apple-darwin14` were canonicalized to `ios` so `if os(ios)` would work as expected in Cabal.

But in GHC 8.4 @angerman I see you made these iOS triples canonicalize to just `darwin`, which means Cabal files expecting `if os(ios)` to work will break. I'm a bit confused about the intention of that change though since if configure is passed `ios`, `tvos` or `watchos` these will still canonicalize to `ios`. LLVM now supports `ios`, `tvos`, and `watchos` in platform triples so it seems possible that Apple is moving toward considering these as more distinct than just some variant of darwin. (@Ericson2314 and I are working on iOS cross support in nixpkgs and would like to use these clearer strings if possible and not against any kind of normalization scheme. See http://llvm.org/doxygen/Triple_8cpp_source.html for LLVM's handling/normalization which notably appears to leave all Apple variants as they are passed in.)

I think this is a good opportunity to think about exactly what is the point of these conditional types in Cabal and GHC. There is an argument to be made that Android is Linux and iOS is Darwin so `if os(linux)` and `if os(darwin)` should work for these mobile platforms. But that means OS types can no longer be mutually exclusive if we still want to be able to single out more specific variants like mobile OSes. Should we accept that? Should we add some new notion of OS 'families'? I don't know.

Ericson2314:
Some links
 - https://github.com/ghc/ghc/blob/master/aclocal.m4#L1953-L1955 here was have GHC's normalization distinguishing the non-macos darwin oses. as for https://github.com/ghc/ghc/commit/6ef6e7c6a74729a6a6bae4f9ba39e216ad13ac70

 - https://github.com/ghc/ghc/blob/master/aclocal.m4#L245 here `darwin` and `ios`, but not the newer ones, are all funneled down into `OSDawin`, is the only haskell variant for that after https://github.com/ghc/ghc/commit/cb4878ffd18a3c70f98bdbb413cd3c4d1f054e1f 

So at the very least, https://github.com/ghc/ghc/commit/6ef6e7c6a74729a6a6bae4f9ba39e216ad13ac70 should have updated https://github.com/ghc/ghc/blame/master/aclocal.m4#L245  to be `darwin|ios|watchos|tvos)`.

More broadly, I wonder why https://github.com/ghc/ghc/blame/master/aclocal.m4#L1955 doesn't just make those `darwin` too if we only have `OSDarwin` (though I could see good reasons to keep as is, e.g the very Cabal use-case we're talking about here), and I need to look into the `llvm-target` file @angerman mentioned to me on IRC.

angerman:
Am I understanding this correctly? `cabal` relies on the `System.Info` module from `base`, to compute the `buildOS` from `System.Info.os`. Which in `base` is the `HOST_OS` define.

This is now even more confusing. We build `cabal-install` on the `build` machine. With a `ghc` that targets the `build` machine, so that we actually obtain a valid `cabal` executable for the build machine. How could `cabal-install` ever properly link against a `base` that contains anything other than the `build` machines OS?

I fear I'm missing something important here.

In general I think cabal should at least interrogate the `ghc` it is using to `build` for the `Target` (from the `--info` data) or even better, allow us to properly provide `--host`, with an appropriate triple, otherwise we end up with issues when ghc becomes multi-target.

I'll see if I can find enough time to hack some `--host` into `cabal` and if it only provides the proper `os` selector for now, and no prefixing of tools.

For `nix` it's probably quite important that `Setup.hs` supports the `--host` configure flag.

Ericson2314:
OK so basically, I'm on board with the general idea @angerman, but for backwards comparability, I think we need `os(darwin)` to still be true for all Darwin variants. The best way to do that on the GHC side I think is to also introduce a `macos` for macOS in the normalized triple, and make either `{macos,ios}_HOST_OS` imply `darwin_HOST_OS` (and same for build and target). `OSDarwin` however would continue to encompass all variants as today. I stared working on that I started working out some commits on https://github.com/obsidiansystems/ghc/tree/darwin-configure. Not sure how to code up the Cabal part off-hand but you or @kmicklas would. I'd hope that ideally we'd get this on the `8.4` and `2.2` branches if there's ever going to be a patch release too.

(For the record, I made an analogous PR at https://github.com/NixOS/nixpkgs/pull/37359 for nixpkgs. I think this OS granularity is a good idea all around.)

CC @bgamari

angerman:
I’m confused. What’s the exact issue we are tying to solve here?

angerman:
After looking at the GHC patch, which looks fine to me. (@Ericson2314 do you open a diff, or do you want me to open one against ghc?)

I still don’t understand the issue at hand? I would expect that any OS identifier picked up from the `Target` properly.

If we are now discussing of making `os(darwin)` a shorthand for `os(ios) || os(macos)`, then this should be purely in cabal. @phadje, @23Skidoo what do you think?

Overall my primary source of confusion probably comes from not understanding how GHCs configure / aclocal should interact with Cabal. To me those two should not be interlinked at all.

Ericson2314:
So at the very least there is slight interaction in that Cabal gets it from `ghc --info`. Additionally I've seen non-ghc code with `xxxx_HOST_OS`, which I assumed GHC set, so it's not just an internal change to GHC.

angerman:
> So at the very least there is slight interaction in that Cabal gets it from `ghc --info`.

Right. But that should be independent of what ever autoconf/aclocal happens in GHC.

> Additionally I've seen non-ghc code with xxxx_HOST_OS, which I assumed GHC set, so it's not just an internal change to GHC.

`CPP` is eventually going to kill us all... alright. The offer to open the Differential from your PR (if you want me to) still stands.

angerman:
@kmicklas while we are at it, should we also add `Aarch64` as an architecture to

```haskell
 data Arch = I386  | X86_64 | PPC | PPC64 | Sparc
           | Arm   | Mips   | SH
           | IA64  | S390
           | Alpha | Hppa   | Rs6000
           | M68k  | Vax
           | JavaScript
           | OtherArch String
```
?

Ericson2314:
@angerman 

> Right. But that should be independent of what ever autoconf/aclocal happens in GHC.

Unfortunately not, because it's the *normalized* triple that `ghc --info` returns, not the one the user actually passed.

> `CPP` is eventually going to kill us all... alright. The offer to open the Differential from your PR (if you want me to) still stands.

Thanks, but anyone can do that now, and I've suffered through learning what `arc` does already :), so I'll just do it myself when ready. OTOH If you have any ideas what should be squashed vs sent separately / what is OK for `8.4.x`, that would be great.

peti:
@angerman, https://github.com/haskell/cabal/pull/5221 adds arm64 a.k.a. aarch64.

23Skidoo:
Adding a notion of OS families sounds like a good idea, let's open a separate ticket for it and continue the discussion there.

23Skidoo:
Merged manually.

PR #5202, opened 2018-03-11
Add isAcceptedLicense to the test-suite

phadej:

### DIFF
--- a/Cabal/tests/UnitTests/Distribution/SPDX.hs
+++ b/Cabal/tests/UnitTests/Distribution/SPDX.hs
@@ -18,6 +18,8 @@ spdxTests =
     , testProperty "LicenseRef roundtrip" licenseRefRoundtrip
     , testProperty "SimpleLicenseExpression roundtrip" simpleLicenseExpressionRoundtrip
     , testProperty "LicenseExpression roundtrip" licenseExpressionRoundtrip
+    , testProperty "isAcceptableLicense l = True"  shouldAcceptProp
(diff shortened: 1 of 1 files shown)


Cabal/tests/UnitTests/Distribution/SPDX.hs
```diff
@@ -55,6 +57,63 @@ reassoc (EAnd a b) = case reassoc a of
...
+    , simpleLicenseExpression BSD_4_Clause `EAnd` simpleLicenseExpression MIT
+    ]
+
+-- | A sketch of what Hackage could accept
+--
+-- * NONE is rejected
+--
+-- * "or later" syntax (+ postfix) is rejected
+--
+-- * "WITH exc" exceptions are rejected
```
ivan-m: Won't this reject [HaXml](http://hackage.haskell.org/package/HaXml-1.25.4/src/COPYRIGHT)?

hvr: @ivan-m what would be the SPDX expression for accurately describing HaXml's licence?

gbaz: this thread ends up  listing more standard licenses that seem to take the place of LGPL+linking exception: https://softwareengineering.stackexchange.com/questions/179084/is-there-a-modified-lgpl-license-that-allows-static-linking

23Skidoo:
Only touches the test suite and is green, merging.

23Skidoo:
Merged, thanks!

PR #5205, opened 2018-03-13
Add verifyGenericPackageDescription

phadej:

### DIFF
--- a/Cabal/Distribution/PackageDescription/Parsec.hs
+++ b/Cabal/Distribution/PackageDescription/Parsec.hs
@@ -221,8 +221,8 @@ parseGenericPackageDescription' cabalVerM lexWarnings utf8WarnPos fs = do
 
     -- Sections
     let gpd = emptyGenericPackageDescription & L.packageDescription .~ pd
-
-    view stateGpd <$> execStateT (goSections specVer sectionFields) (SectionS gpd Map.empty)
+    gpd' <- view stateGpd <$> execStateT (goSections specVer sectionFields) (SectionS gpd Map.empty)
+    verifyGenericPackageDescription gpd'
   where
     safeLast :: [a] -> Maybe a
     safeLast = listToMaybe . reverse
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -522,6 +522,21 @@ When/if we re-implement the parser to support formatting preservging roundtrip
...
+
+-- | This is a post processing step for GPD
+--
+-- Note that if you are unsure if something have to be checked there
+-- or in "Distribution.PackageDescription.Check", it most likely should be there.
+verifyGenericPackageDescription :: GenericPackageDescription -> ParseResult GenericPackageDescription
+verifyGenericPackageDecsription gpd = do
+    -- TODO: currently there aren't any checks, but this is where they can be added.
+    -- Note: if we will check that used flags are the ones declared,
+    -- we can do that already in parseCondTree: then errors will have a correct position.
```
hvr: so... how does one fail here? e.g. let's assume an undeclared flag use-site is detected; how do I throw an error pointing to the use-site of the unknown flag?

23Skidoo:
Looks like CI failures are due to `-Wall`.

Cabal/Distribution/PackageDescription/Parsec.hs
phadej: @hvr with `parseFailure`. As the comment implies, at this point we don't have positions attached, so we can only `parseFailure zeroPos`.

Let's do better flag handling as a follow-up or an alternative.

phadej: Note to previous: Unfortunately we might need to do some trickery-pockery, as flag may be declared after it's usage, if someone puts `flag` stanza after `library`. This is slight :/

phadej:
Cabal-3.0 has `checkForUndefinedFlags`

Issue #5206, opened 2018-03-13
checkPackageFiles does not work for subdirs in Cabal 2.2


snoyberg:
I believe this is a regression between Cabal 2.0 and 2.2. I discovered this when using a `stack` built against Cabal 2.2. When running the command `stack sdist ./some-subdir`, I got an error message about the cabal file not existing in the local directory. You can see the original report at https://github.com/commercialhaskell/stack/issues/3919.

__Repro__ I'm including Stack script headers to make it easier to test against both Cabal 2.0 (shipped with GHC 8.2.2) and Cabal 2.2 (shipped with GHC 8.4.1), but testing directly with GHC will produce the same result. Working code for GHC 8.2.2:

```haskell
#!/usr/bin/env stack
-- stack --resolver ghc-8.2.2 script
import Distribution.PackageDescription.Check
import Distribution.PackageDescription.Parse
import System.Environment (getArgs)
import System.FilePath (takeDirectory)
import Distribution.PackageDescription.Configuration (flattenPackageDescription)

main :: IO ()
main = getArgs >>= mapM_ go

go :: FilePath -> IO ()
go fp = do
  putStrLn $ "Checking: " ++ fp
  gpd <- readGenericPackageDescription minBound fp
  let dir = takeDirectory fp
      pd = flattenPackageDescription gpd
  checkPackageFiles pd dir >>= print
  putStrLn "\n\n"
```

Runs with:

```
$ ./check-ghc-822.hs http-client/http-client/http-client.cabal 
Checking: http-client/http-client/http-client.cabal
[]
```

Tweaked slightly to use GHC 8.4.1 instead:

```
#!/usr/bin/env stack
-- stack --resolver ghc-8.4.1 script
import Distribution.PackageDescription.Check
import Distribution.PackageDescription.Parsec
import System.Environment (getArgs)
import System.FilePath (takeDirectory)
import Distribution.PackageDescription.Configuration (flattenPackageDescription)

main :: IO ()
main = getArgs >>= mapM_ go

go :: FilePath -> IO ()
go fp = do
  putStrLn $ "Checking: " ++ fp
  gpd <- readGenericPackageDescription minBound fp
  let dir = takeDirectory fp
      pd = flattenPackageDescription gpd
  checkPackageFiles pd dir >>= print
  putStrLn "\n\n"
```

Results in:

```
$ ./check-ghc-841.hs http-client/http-client/http-client.cabal 
Checking: http-client/http-client/http-client.cabal
[...]

PR #5208, opened 2018-03-13
Add missing relative for getFileContents #5206

snoyberg:

I confirmed with the repro in #5206

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -1830,7 +1830,7 @@ checkPackageFiles pkg root = checkPackageContent checkFilesIO pkg
       doesFileExist        = System.doesFileExist                  . relative,
       doesDirectoryExist   = System.doesDirectoryExist             . relative,
       getDirectoryContents = System.Directory.getDirectoryContents . relative,
-      getFileContents      = BS.readFile
+      getFileContents      = BS.readFile                           . relative
     }
     relative path = root </> path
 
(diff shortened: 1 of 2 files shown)


Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1830,7 +1830,7 @@ checkPackageFiles pkg root = checkPackageContent checkFilesIO pkg
       doesFileExist        = System.doesFileExist                  . relative,
       doesDirectoryExist   = System.doesDirectoryExist             . relative,
       getDirectoryContents = System.Directory.getDirectoryContents . relative,
-      getFileContents      = BS.readFile
+      getFileContents      = BS.readFile . relative
```
hvr: obviously: indentation

23Skidoo:
Right, would be nice to have a test case.

snoyberg:
I've updated with the formatting change requested. Unfortunately, I don't know enough about the test suite to know the right way to write a test case for this. I could do it by refactoring the set of functions and exposing `checkFilesIO`, but that would be an API change that I'm not sure would be accepted.

hvr:
Why can't the repro case from #5206 turned into a regression test?

snoyberg:
That may be the case. The missing bit is knowing how to add the appropriate package files into the repo. I see a bunch of parser tests, but I don't see any complete packages in the repo that could be hijacked for this. I don't mind adding such a minimal package, I'm just not sure where to put it.

23Skidoo:
Merged, thanks! A regression test would still be nice to have, but can be added later.

hvr:
@23Skidoo note this was confusingly filed against 2.2 rather than `master`, so may need forward-porting

23Skidoo:
Yep, I cherry-picked it to `master`.

PR #5209, opened 2018-03-13
Refactoring removing `data ReadPackageIndexMode`

hvr:
This is anticipated refactoring (since `readCacheStrict` will
be removed anyway once we remove the `cabal sandbox` code) is
motivated by workarounding a GHC 8.4.1 caused regression
(see https://ghc.haskell.org/trac/ghc/ticket/13930 for details)

Addresses #5201

PR #5215, opened 2018-03-15
bootstrap: prepare for ghc 8.4.1

tuncer:


tuncer:
Only item left is hackage-security which should be released any day now (https://github.com/haskell/hackage-security/issues/211).

hvr:
@tuncer we also need #5209 merged & backported, as otherwise the resulting `cabal` executable will be unusable

tuncer:
Right, but the bootstrap changes alone are safe with 8.2.2. Thus, we could merge this now and later bump only hackage-security.

tuncer:
ping?

23Skidoo:
Merged, thanks!

PR #5216, opened 2018-03-18
Bumped network-uri versions

LightAndLight:
`Distribution/Client/Security/DNS.hs` requires `instance NFData URI`, which was only introduced in `network-uri-2.6.0.2`.

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -323,7 +323,7 @@ library
     -- is pulled in, otherwise the constraint solver doesn't have enough
     -- information
     if flag(network-uri)
-      build-depends: network-uri >= 2.6 && < 2.7, network >= 2.6 && < 2.7
+      build-depends: network-uri >= 2.6.0.2 && < 2.7, network >= 2.6 && < 2.7
     else
       build-depends: network     >= 2.4 && < 2.6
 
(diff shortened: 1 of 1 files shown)


LightAndLight:
Hm. It seems like there's no way to make this build if the `network-uri` flag is disabled, because `network <2.6` doesn't have `instance NFData URI`.

Should we consider adding an orphan instance that is only available when `network-uri` is disabled?

hvr:
> which was only introduced in network-uri-2.6.0.2.

That's unfortunate... no Changelog and not even a minor version bump to signal the addition... :-/

> Should we consider adding an orphan instance that is only available when network-uri is disabled?

This affects only `cabal-install` for which we have a 3-year compat window policy; and I've looked at https://www.haskell.org/platform/contents.html which tells us that the first version of `network-uri` to ever be included in a Haskell Platform release was `network-uri-2.6.0.3` with GHC 7.10.2

Consequently, I think it's time to just drop support for `network(-uri) < 2.6`; and given you're the first one to report this (how did you notice this btw?) I don't think this will cause much problems for users.

LightAndLight:
No worries. I was trying to build HEAD with GHC 8.2.1 package set on nix, which only has `network-uri-2.6.0.0`. It was a bit of a head scratcher.

hvr:
@LightAndLight Ok, but why would have Nix such a rather old release of `network-uri`?

23Skidoo:
Dropping old `network` sounds good.

23Skidoo:
Closing in favour of #5235.

PR #5221, opened 2018-03-19
add support for the aarch64 architecture

peti:
We recognize arm64 as an alias.

### DIFF
--- a/Cabal/Distribution/Simple/PreProcess.hs
+++ b/Cabal/Distribution/Simple/PreProcess.hs
@@ -598,6 +598,7 @@ platformDefines lbi =
       PPC64       -> ["powerpc64"]
       Sparc       -> ["sparc"]
       Arm         -> ["arm"]
+      AArch64     -> ["aarch64"]
       Mips        -> ["mips"]
       SH          -> []
       IA64        -> ["ia64"]
--- a/Cabal/Distribution/System.hs
+++ b/Cabal/Distribution/System.hs
@@ -151,9 +151,9 @@ buildOS = classifyOS Permissive System.Info.os
 -- * Machine Architecture
 -- ------------------------------------------------------------
 
--- | These are the known Arches: I386, X86_64, PPC, PPC64, Sparc
--- ,Arm, Mips, SH, IA64, S39, Alpha, Hppa, Rs6000, M68k, Vax
--- and JavaScript.
+-- | These are the known Arches: I386, X86_64, PPC, PPC64, Sparc,
+-- Arm, AArch64, Mips, SH, IA64, S39, Alpha, Hppa, Rs6000, M68k,
+-- Vax, and JavaScript.
 --
 -- The following aliases can also be used:
 --    * PPC alias: powerpc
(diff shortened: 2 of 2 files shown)


Ericson2314:
CC @kmicklas

Cabal/Distribution/Simple/PreProcess.hs
```diff
@@ -598,6 +598,7 @@ platformDefines lbi =
       PPC64       -> ["powerpc64"]
       Sparc       -> ["sparc"]
       Arm         -> ["arm"]
+      Arm64       -> ["arm64"]
```
Ericson2314: Can we normalize to `aarch64`? This seems to be the dominant name, unfortunately...

Ericson2314: *reads PR desccription* even on LLVM, I think?

Ericson2314: http://lists.llvm.org/pipermail/llvm-dev/2014-April/071903.html so there was two implementations of the backend in LLVM, sigh: `arm64` by Apple and `aarch64` by everyone else. They merged them shortly after the open sourcing, thank goodness.

peti: Note that the Linux kernel also uses `arm64`. It's a bike-shedding issue, no doubt, but personally I find that the name `arm64` conveys the information "I am an 64 bit arm cpu" nicely, whereas `aarch64` doesn't convey that kind of information and feels like an arbitrary choice that could have been "fooglbrmphf64" just as well.

angerman: Ghc went with aarch64, because that’s the name that ARM gave the architecture.  Arm64 is something only apple came up with.

I’m aware of Linus rant regarding arm64/aarch64.

As such I would argue this should be aarch64 in cabal as well to be consistent with GHC.

I am against having aarch64 AND arm64 in Haskell.

/cc @Mistuke

Ericson2314: Oh haha I hadn't seen https://lkml.org/lkml/2012/7/15/133 . I agree with everything @angerman just said.

Ironically the Haskell variant name in GHC is still `ArchARM64` though, so if we just change the printing logic here can keep the variant name the same, we'll be *very* consistent :D.

Mistuke: Yeah, I think we should stick with what the vendor called it, which is `AArch64` which is what the [execution state](https://developer.arm.com/products/architecture/a-profile) is called.
`AArch64` actually conveys a lot, it's an acronym for `Arm 64-bit Architecture`.  Whereas `Arm64` is ambiguous, does it refer to the `execution state` or `ISA`?

Ericson2314: @Mistuke when the new iteration of the naming scheme adds more indirection, stuff's going to be confusing no matter what. And then to make matters even more confusing "The AArch64 Execution state supports the A64 instruction set"

Mistuke: @Ericson2314 But that's my point. Even with new additions to the naming scheme `AArch64` and `A64` will mean the same things. Just like `AArch32` and `T32`, `A32`. In this whole picture, `Arm64` has no meaning because it's not part of the specification.

Ericson2314: @Mistuke true. `Arm64` is definitely not needed.

peti:
Okay. I've changed the constructor name to `AArch64` and that name is pretty-printed as `aarch64`, too. The name `arm64` is recognized as an alias.

peti:
The CI failures on travis look quite unrelated to these changes. Two builds timed out because they ran too long and one build job failed with an "ghc is out of memory" error. These issues are most likely due to system-level problems at travis-ci.org, not because of these changes.

angerman:
LGTM. @23Skidoo if this was green, I would have just merged it. Are you ok with the CI failiures?

23Skidoo:
Merged, thanks!

Issue #4506, opened 2017-05-07
`new-haddock`'s file monitoring broken


hvr:
Consider the following simple reproduction instructions:

```
$ cabal get sort-1.0.0.1 && cd sort-0.0.0.1/
Unpacking to sort-0.0.0.1/
```

The first invocation of `new-haddock` works as expected:

```
$ cabal new-haddock
Resolving dependencies...
In order, the following will be built (use -v for more details):
 - sort-0.0.0.1 (lib) (first run)
Configuring library for sort-0.0.0.1..
Preprocessing library for sort-0.0.0.1..
Running Haddock on library for sort-0.0.0.1..
Haddock coverage:
  67% (  2 /  3) in 'Data.Sort'
  Missing documentation for:
    Module header
Documentation created:
/tmp/sort-0.0.0.1/dist-newstyle/build/x86_64-linux/ghc-8.0.2/sort-0.0.0.1/doc/html/sort/index.html
```

However, the 2nd invocation of `new-haddock` rebuilds the documentation (NB: the install-plan wasn't resolved, and the action plan says "(first run)"):

```
$ cabal new-haddock
In order, the following will be built (use -v for more details):
 - sort-0.0.0.1 (lib) (first run)
Preprocessing library for sort-0.0.0.1..
Running Haddock on library for sort-0.0.0.1..
Haddock coverage:
  67% (  2 /  3) in 'Data.Sort'
  Missing documentation for:
    Module header
Documentation created:
/tmp/sort-0.0.0.1/dist-newstyle/build/x86_64-linux/ghc-8.0.2/sort-0.0.0.1/doc/html/sort/index.html
```

What's really bad though: if we remove `dist-newstyle` again, and start with `cabal new-build`, then `new-haddock` doesn't generate any documentation at all:

```
$ rm -rf dist-newstyle/
$ cabal new-build
Resolving dependencies...
In order, the following will be built (use -v for more details):
 - sort-0.0.0.1 (lib) (first run)
Configuring library for sort-0.0.0.1..
Preprocessing library for sort-0.0.0.1..
Building library for sort-0.0.0.1..
[1 of 1] Compiling Data.Sort        ( src/Data/Sort.hs, /tmp/sort-0.0.0.1/dist-newstyle/build/x86_64-linux/ghc-8.0.2/sort-0.0.0.1/build/Data/Sort.o )

$ cabal new-haddock
Up to date
```

chris-martin:
[...]

Issue #5178, opened 2018-03-02
haddock after new-repl


llelf:
Recently cabal started to basically do a haddock build after new-repl.
Is it intentional (or my setup is broken)?  It's *definitely* surprising.

This is what I mean:
```
$ cat cabal.project
packages: .

package foo
  documentation: True

$ cabal new-repl lib:foo                        --  new-repl

⟨…⟩

Ok, six modules loaded.
Collecting type info for 6 module(s) ... 	--  ghci repl
→ :quit 					--  exit from ghci
Leaving GHCi.
Preprocessing library ⟨…⟩			--  starting haddock build!!!
Running Haddock on library ⟨…⟩
⟨…⟩

```

23Skidoo:
Likely because `repl` and `build` share the same code path. Agreed that `repl` shouldn't do this. Is this with `2.2` or only with `master`?

alexbiehl:
This might be the result of the recent https://github.com/haskell/cabal/commit/c094940abc4266c8d013853b66271540cb2dc7f0. Maybe we should disable haddock'ing for repl builds.

llelf:
@23Skidoo master. But I tried  2.2, the same result.

@alexbiehl new-test too likely. It shouldn't fail because of haddock error, … or  is it? Not sure.

quasicomputational:
I verified that I can reproduce this with 2.2, and that I can't reproduce now on master. Presumably #5226 has fixed it, so closing.

PR #5226, opened 2018-03-21
Haddock: Generate haddock for components

alexbiehl:
Currently settings `documentation: true` enables documentation
generation via haddock for your whole package, including tests and benchmarks.
However, there are additional flags to control generation of
documentation for these "second class" documentation targets, which are
currently not honored at the cabal-install side of things. Namely,
`tests`, `benchmarks`, `executables`, etc. provided under the
`haddock` section in your `$CABAL_HOME/config`.

This patch adds a more sensible approach to documentation generation
via haddock. The new behaviour works like this:

- Setting `documentation: true` or passing
   `--enable-documentation` to cabal-install enables documentation
   for any component in the build plan honoring the respective
   flags for "second class" doc targets from configuration
   or CLI.

- Invoking new-haddock with a target selector will make sure
   the respective flags for "second class" doc targets are set
   correctly. E.g.

         $ new-haddock tests

  Will generate documentation for the test suite of your package
   even if you have `tests: false` in your haddock configuration.

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -7,6 +7,7 @@
     into per-component condition trees anyway. Now it's finally been put
     out of its misery (#4383).
   * Added `Eta` to `CompilerFlavor` and to known compilers.
+  * `cabal haddock` now generates per-component documentation (#5226).
 
 ----
 
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -329,7 +329,7 @@ haddockAction hooks flags args = do
 
   hookedAction preHaddock haddockHook postHaddock
                (return lbi { withPrograms = progs })
-               hooks flags' args
+               hooks flags' { haddockArgs = args } args
 
 cleanAction :: UserHooks -> CleanFlags -> Args -> IO ()
 cleanAction hooks flags args = do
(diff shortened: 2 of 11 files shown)


alexbiehl:
@merijn this should fix your problem.

alexbiehl:
(This resolves https://github.com/haskell/cabal/issues/4506)

23Skidoo:
Needs a changelog note and maybe some tweaks to docs.

hvr:
As this extends the `Setup.hs haddock` CLI, this may need version gating in e.g. `Distribution.Client.Setup` for `custom-setup` situations when we talk to a `Setup.hs`

merijn:
I've been using this patch the past few days and it seems to work fine, but introduces one silly issue. With `documentation: True` set it will *also* rebuild docs every time when you run `cabal new-repl` which is probably the wrong thing to do.

alexbiehl:
@merijn Generating documentation with new-repl was present before this patch (c.f. 
 https://github.com/haskell/cabal/issues/5178).

merijn:
Ok, my bad. I just didn't wanna rebuild all of cabal to test that.

alexbiehl:
@merijn I just corrected the faulty behaviour.

merijn:
works now \o/

alexbiehl:
@23Skidoo I added change log entry and documentation.

alexbiehl:
Nice! A comment from the cabal gurus would be much appreciated to move this
patch forward!

Merijn Verstraaten <notifications@github.com> schrieb am Do., 12. Apr.
2018, 14:23:

alexbiehl:
@23Skidoo ping

hvr:
@ezyang did you take into account https://github.com/haskell/cabal/pull/5226#issuecomment-375624613 ?

PR #5228, opened 2018-03-22
Fix the SPDX spec version number

peti:
We implement version 3.0, not 2.1.


phadej:
Good catch, except that spec and license list versions are different numbers, we have to mention both

peti:
@phadej, oh, you are right! I updated the PR accordingly.

PR #5235, opened 2018-03-27
Drop support for network < 2.6

23Skidoo:
Also drop `bytestring-builder` from the bootstrap script, cabal-install no longer supports `bytestring < 0.10.2` anyway.

---


23Skidoo:
OK, bootstrap seems to still work.

PR #5236, opened 2018-03-27
Haddock deprecate hscolour, tweak options

harpocrates:

--- 

Small paper-cuts for `cabal haddock`:

  * add a `--hyperlinked-source` synonym for the `--hyperlink-source` option and deprecate `hscolour` (see comments on https://github.com/haskell/cabal/pull/5140)
  * pass in `--since-qual=external` to new enough Haddock versions (see https://github.com/haskell/haddock/pull/749)
  * add a `--quickjump` option and enable it when `--for-hackage` is specified (see https://github.com/haskell/haddock/issues/683).

I'm still in the process of testing this.


alexbiehl:
Alec, could you also add the '--quickjump' flag when generating docs for Hackage?

harpocrates:
@alexbiehl Only for Hackage? Is there any disadvantage to always enabling it?

alexbiehl:
Not really a good reason but we don't enable hyperlinked source by default
either.

Alec Theriault <notifications@github.com> schrieb am Di., 27. März 2018,
19:11:

gbaz:
Quickjump doesn't work with local docs  at the moment, due to ajax restrictions :-/

An alternate local mode that embedded the quickjump data thus bypassing the xhr request, just for local docs, might be quite nice. (in fact, why can't that be done in general?)

harpocrates:
I'm sold on enabling it only in `--for-hackage`.

I'll add `--quickjump` as a separate flag, and `--for-hackage` will imply it.

alexbiehl:
The reason to have it separately from the html is that we want Hackage to
read the index on the package page. Also we pay n times the index if we
store them in each html file.

gbaz <notifications@github.com> schrieb am Di., 27. März 2018, 19:21:

gbaz:
right -- i'm not suggesting to embed directly. I'm suggesting to use a `<link>` tag or something to a js file for those cases. So you can still share it. And hackage can still read it (either by embedding the js file itself, or by xhr to a separate json file). just move to a load mechanism that isn't dynamic, but is static with the page, to avoid xhr restrictions.

23Skidoo:
I'm +1 on always enabling hyperlink-source, btw.

harpocrates:
I am _strongly_ against always enabling `hyperlink-source` - it can be quite slow for large generated source files. Once https://github.com/haskell/haddock/issues/758 gets fixed, I'll go from "strongly against" to "don't care".

23Skidoo:
Good point, let's wait for that issue to get fixed.

alexbiehl:
Thanks Alec!

harpocrates:
I'll merge as soon as all CI is green.

@23Skidoo Is the deprecation message around `hscolour` good enough? I looked for other similar instances of deprecation but didn't really find any.

harpocrates:
@alexbiehl Which version of Haddock should I expect the since annotation stuff to be in? 2.19.1 or 2.20.0?

alexbiehl:
@harpocrates Make it 2.20

Alec Theriault <notifications@github.com> schrieb am Mi., 28. März 2018,
20:14:

harpocrates:
I let the tests run for `22c11f19039ca86be2dc5c25f25b1dbf3cf1492e` and Travis had only a handful of failures, all with error messages that cannot be related to that commit. Since then, all I've done is bump up the required Haddock for `--since-qual` (as per Alex's comment).

@23Skidoo Does this look OK to merge now?

alexbiehl:
This looks good. @23Skidoo can we merge this?

harpocrates:
I've rebased on master. Now that https://github.com/haskell/cabal/pull/5261 is merged, I think this should be OK again.

23Skidoo:
Merged, thanks!

Issue #4247, opened 2017-01-19
Some ghc-options should not be considered part of build hash


ezyang:
This is related to #3883 but it is a distinct problem so I wanted to make a ticket for it by itself.

Some ghc-options don't have any affect on build outputs, e.g., `-v`, but if someone specifies it via `ghc-options`, e.g., in their cabal.project file, it will cause us to rebuild all deps because `elabProgramArgs` has changed, which changes the hash, which means we need to rebuild. This is not good.

It's unclear how to go about fixing this. We could build in knowledge to cabal-install what arguments should and should not be hashed. But we'd be constantly chasing the most up-to-date set, and there would always be some time where we wouldn't have gotten it exactly right. Alternately, we could let the user specify, "use this option when building, but don't include it in the hash." But here it seems easy for a user to shoot themselves in the foot, ESP when it's a package being installed to the global store; worst case scenario you'd have to blow away the entire store because you corrupted it.

CC @bgamari who reported this most recently.

DanielG:
How about adding a command to GHC that tells cabal which flags should influence the hash? For old versions this knowledge would still have to be included but at least then we don't need to chase after GHC.

ezyang:
That's an interesting idea. The relevant code in GHC is `fingerprintDynFlags` in `FlagChecker`, we'd have to reverse engineer the flags that affect fields in DynFlags here.

It's worth noting GHC's recompilation avoidance doesn't always get this right. See #3891. So it' still challenging to compute the flags, even on GHC's side ;)

23Skidoo:
Yep, just bumped into this when I tried to enable `-Werror` for my project.

mightybyte:
I also hit this when I did `cabal new-build --ghc-option=-fno-code`

hvr:
I think that `-fno-code` is a bit special here, as it *does* affect the `.o` output... by not producing any! This could have the ability to confuse cabal quite a lot. Maybe this rather demands a special "dry" mode to invoke, i.e. `cabal new-build --no-code` which takes of passing this to ghc and not get confused about the `.o` files not being generated.

mightybyte:
No-code is great for getting something to compile.  It's much faster than anything else I know of, and significantly speeds up the compile-fix-errors loop.  The ideal situation in my mind would be to have a `cabal tc` (for type check) command that does this.  Making it a separate command makes it more convenient and discoverable (as well as being a little quicker to type).

merijn:
[...]

PR #5237, opened 2018-03-28
Infrastructure to filter commandline arguments

merijn:
This piggybacks any filters on Cabal's support for known programs, it allows us to specify a version specific argument filtering function for each program Cabal knows about.

See #4247.

### DIFF
--- a/Cabal/Distribution/Simple/Program/Types.hs
+++ b/Cabal/Distribution/Simple/Program/Types.hs
@@ -38,6 +38,7 @@ module Distribution.Simple.Program.Types (
 import Prelude ()
 import Distribution.Compat.Prelude
 
+import Distribution.PackageDescription
 import Distribution.Simple.Program.Find
 import Distribution.Version
 import Distribution.Verbosity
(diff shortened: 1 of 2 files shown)


merijn:
Some example usage code: https://github.com/merijn/cabal/commit/b3e55d956226bf60b523986830989ce442ae31f2

Probably the filters should be written in program specific modules, but moving them into the `Program.GHC` module leads to a recursive import I couldn't be arsed working around right now.

Cabal/Distribution/Simple/Program/Types.hs
```diff
@@ -74,10 +74,14 @@ data Program = Program {
        -- | A function to do any additional configuration after we have
        -- located the program (and perhaps identified its version). For example
        -- it could add args, or environment vars.
-       programPostConf :: Verbosity -> ConfiguredProgram -> IO ConfiguredProgram
+       programPostConf :: Verbosity -> ConfiguredProgram -> IO ConfiguredProgram,
+       -- | A function that filters any arguments that don't impact the output
+       -- from a commandline. Used to limit the volatility of dependency hashes
+       -- when using new-build.
+       programFilterArgs :: Maybe Version -> [String] -> [String]
```
hvr: name bikeshedding: I'd rather call this something like `programNormalizeArgs` to emphasize this is about normalizing

merijn: I don't have a strong preference for the name, but I hate trying to come up with them, so I'm fine with whatever the majority wants.

alexbiehl: I am with Herbert regarding the name.

merijn:
Updated to deal with `-Werror` issues and changed the name.

merijn:
Updated this to match the naming issue and deal with `-Werror`

merijn:
Now updated so that is actually looks at the list of builtin programs Cabal knows about...

merijn:
Last call for comments or I'm merging this!

Issue #5225, opened 2018-03-21
new-build does not rebuild when project-file is changed


andrewthad:
When I use `cabal new-build --project-file=...` to manually select a project file and then run the same command with a different project file, `cabal` finishes immidiately the second time, claiming that the project is up to date. Cabal should take the project file into account when deciding whether or not to rebuild things. I can work around this by just changing an arbitrary module in the project.

andrewthad:
That makes sense. I'll start doing that. It would be nice if `builddir` could be specified in the `cabal.project` file.

lspitzner:
Mildly certain that this is a duplicate of #4206.

23Skidoo:
@andrewthad Can you please confirm that #5239 fixed this?

quasicomputational:
I'm happy to confirm that this is fixed in HEAD and, considering when the commit landed, it should also be fixed in 2.4.

PR #5239, opened 2018-03-29
Fix cabal project caching bug, Add verbose printing

### LINKED ISSUE #4662: switching project-file does not properly mark stuff as dirty
Setup: Two project files:

- cabal.project.opt
  ~~~~
  packages: .
  package mypackage
    optimization: 1
  ~~~~
- cabal.project.noopt
  ~~~~
  packages: .
  package mypackage
    optimization: 0
  ~~~~

Now, doing

~~~~
[manually remove dist-newstyle]
> cabal new-build --project-file=cabal.project.noopt lib:mypackage
[builds lib]
> cabal new-build --project-file=cabal.project.opt exe:myexecutable
[builds exe]
~~~~

In the second step, I expected that `lib` is rebuilt, since the only existing build products for it are without optimizations. Instead, it seems to build a `-O1` exe against the `-O0` lib.

Note that doing the same without `--project-file` works as expected, i.e.

~~~~
[manually remove dist-newstyle]
> cabal new-build -O0 lib:mypackage
[builds lib]
> cabal new-build -O1 exe:myexecutable
[rebuilds lib, builds exe]
~~~~

lspitzner:
- Use (projectConfigConfigFile, distProjectFile "") as key
  for config FileMonitor, instead of ().
- Add verbose printing of relevant config file paths.

Fixes #4662 and probably #5225.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig.hs
@@ -415,41 +415,35 @@ renderBadProjectRoot (BadProjectRootExplicitFile projectFile) =
 -- | Read all the config relevant for a project. This includes the project
 -- file if any, plus other global config.
 --
-readProjectConfig :: Verbosity -> Flag FilePath -> DistDirLayout -> Rebuild ProjectConfig
+readProjectConfig :: Verbosity
+                  -> Flag FilePath
+                  -> DistDirLayout
+                  -> Rebuild ProjectConfig
 readProjectConfig verbosity configFileFlag distDirLayout = do
-    global <- readGlobalConfig             verbosity configFileFlag
-    local  <- readProjectLocalConfig       verbosity distDirLayout
-    freeze <- readProjectLocalFreezeConfig verbosity distDirLayout
-    extra  <- readProjectLocalExtraConfig  verbosity distDirLayout
+    global <- readGlobalConfig                verbosity configFileFlag
+    local  <- readProjectLocalConfigOrDefault verbosity distDirLayout
+    freeze <- readProjectLocalFreezeConfig    verbosity distDirLayout
+    extra  <- readProjectLocalExtraConfig     verbosity distDirLayout
     return (global <> local <> freeze <> extra)
 
 
 -- | Reads an explicit @cabal.project@ file in the given project root dir,
 -- or returns the default project config for an implicitly defined project.
 --
-readProjectLocalConfig :: Verbosity -> DistDirLayout -> Rebuild ProjectConfig
-readProjectLocalConfig verbosity DistDirLayout{distProjectFile} = do
+readProjectLocalConfigOrDefault :: Verbosity
+                                -> DistDirLayout
+                                -> Rebuild ProjectConfig
+readProjectLocalConfigOrDefault verbosity distDirLayout = do
   usesExplicitProjectRoot <- liftIO $ doesFileExist projectFile
   if usesExplicitProjectRoot
     then do
-      monitorFiles [monitorFileHashed projectFile]
-      addProjectFileProvenance <$> liftIO readProjectFile
+      readProjectFile verbosity distDirLayout "" "project file"
     else do
       monitorFiles [monitorNonExistentFile projectFile]
       return defaultImplicitProjectConfig
 
   where
-    projectFile = distProjectFile ""
-    readProjectFile =
-          reportParseResult verbosity "project file" projectFile
-        . parseProjectConfig
-      =<< readFile projectFile
-
-    addProjectFileProvenance config =
-      config {
-        projectConfigProvenance =
-          Set.insert (Explicit projectFile) (projectConfigProvenance config)
-      }
+    projectFile = distProjectFile distDirLayout ""
 
     defaultImplicitProjectConfig :: ProjectConfig
     defaultImplicitProjectConfig =
@@ -470,7 +464,7 @@ readProjectLocalConfig verbosity DistDirLayout{distProjectFile} = do
 readProjectLocalExtraConfig :: Verbosity -> DistDirLayout
                             -> Rebuild ProjectConfig
 readProjectLocalExtraConfig verbosity distDirLayout =
-    readProjectExtensionFile verbosity distDirLayout "local"
+    readProjectFile verbosity distDirLayout "local"
                              "project local configuration file"
 
 -- | Reads a @cabal.project.freeze@ file in the given project root dir,
@@ -480,19 +474,22 @@ readProjectLocalExtraConfig verbosity distDirLayout =
 readProjectLocalFreezeConfig :: Verbosity -> DistDirLayout
                              -> Rebuild ProjectConfig
 readProjectLocalFreezeConfig verbosity distDirLayout =
-    readProjectExtensionFile verbosity distDirLayout "freeze"
+    readProjectFile verbosity distDirLayout "freeze"
                              "project freeze file"
 
 -- | Reads a named config file in the given project root dir, or returns empty.
 --
-readProjectExtensionFile :: Verbosity -> DistDirLayout -> String -> FilePath
-                         -> Rebuild ProjectConfig
-readProjectExtensionFile verbosity DistDirLayout{distProjectFile}
+readProjectFile :: Verbosity
+                -> DistDirLayout
+                -> String
+                -> String
+                -> Rebuild ProjectConfig
+readProjectFile verbosity DistDirLayout{distProjectFile}
                          extensionName extensionDescription = do
     exists <- liftIO $ doesFileExist extensionFile
     if exists
       then do monitorFiles [monitorFileHashed extensionFile]
-              liftIO readExtensionFile
+              addProjectFileProvenance <$> liftIO readExtensionFile
       else do monitorFiles [monitorNonExistentFile extensionFile]
               return mempty
   where
@@ -502,6 +499,12 @@ readProjectExtensionFile verbosity DistDirLayout{distProjectFile}
           reportParseResult verbosity extensionDescription extensionFile
         . parseProjectConfig
       =<< readFile extensionFile
+
+    addProjectFileProvenance config =
+      config {
+        projectConfigProvenance =
+          Set.insert (Explicit extensionFile) (projectConfigProvenance config)
+      }
 
 
 -- | Parse the 'ProjectConfig' format.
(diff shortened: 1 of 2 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -294,48 +295,62 @@ sanityCheckElaboratedPackage ElaboratedConfiguredPackage{..}
...
+          projectConfig <- phaseReadProjectConfig
+          localPackages <- phaseReadLocalPackages projectConfig
+          return (projectConfig, localPackages)
+
+    when (verbosity >= verbose) $ do
+      monitorFileList <- getFileMonitorSimpleFileList fileMonitorProjectConfig
+      info verbosity
+        $ unlines
+        $ ("this build was affected by the following (project) config files:" :)
+        $ filter (\x -> not (".cabal" `isSuffixOf` x) && x /= ".")
```
alexbiehl: Filtering out only cabal files might lead to a lot of false-posives in this list. Consider the support for local tarballs, which adds a file monitor to a `.tar.gz` file which will count as "config file" here.

lspitzner: raah that's a good point, but I don't see any elegant to solution for this without creating some kind of duplication or implication what config files we have. Perhaps this should just not do any filtering and not pretend it only contains config files, but "configuration input stuff"?

alexbiehl: Well, we have `distProjectFile` in scope, we could do something like this:

filterM doesFileExist $ map distProjectFile ["", "local", "freeze"]

alexbiehl: Or better: we have `projectConfigProvenance` which should hold all the paths from the involved configuration files.

lspitzner:
@alexbiehl Like that?

cabal-install/Distribution/Client/ProjectConfig.hs
```diff
@@ -418,38 +418,19 @@ renderBadProjectRoot (BadProjectRootExplicitFile projectFile) =
 readProjectConfig :: Verbosity -> Flag FilePath -> DistDirLayout -> Rebuild ProjectConfig
 readProjectConfig verbosity configFileFlag distDirLayout = do
     global <- readGlobalConfig             verbosity configFileFlag
-    local  <- readProjectLocalConfig       verbosity distDirLayout
+    hasCabalProjectFile <- liftIO $ doesFileExist projectFile
```
alexbiehl: I liked it better as it was before with 

```
thing1 <- read this
thing2 <- read that
thing3 <- read that 
... 
```

The logic here looks a bit out of place.

alexbiehl: you could lift `addProjectFileProvenance` to top level for example to reduce the duplication that way.

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -294,48 +294,61 @@ sanityCheckElaboratedPackage ElaboratedConfiguredPackage{..}
...
-rebuildProjectConfig verbosity
-                     distDirLayout@DistDirLayout {
-                       distProjectRootDirectory,
-                       distDirectory,
-                       distProjectCacheFile,
-                       distProjectCacheDirectory
-                     }
-                     cliConfig = do
+                     -> IO
+                          ( ProjectConfig
```
alexbiehl: IMHO these stylistic changes distract the diff from the meat somewhat, there are some others further down

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -294,48 +294,61 @@ sanityCheckElaboratedPackage ElaboratedConfiguredPackage{..}
...
-      runRebuild distProjectRootDirectory $
-      rerunIfChanged verbosity fileMonitorProjectConfig () $ do
-
-        projectConfig <- phaseReadProjectConfig
-        localPackages <- phaseReadLocalPackages projectConfig
-        return (projectConfig, localPackages)
+      runRebuild distProjectRootDirectory
+      $ rerunIfChanged verbosity
+                       fileMonitorProjectConfig
+                       (projectConfigConfigFile, distProjectFile "")
```
alexbiehl: Don't pass the tuple directly. Instead make a new `fileMonitorProjectConfigValue` or `fileMonitorProjectConfigInput` or something.

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -294,48 +294,61 @@ sanityCheckElaboratedPackage ElaboratedConfiguredPackage{..}
...
+      $ do
+          liftIO $ info verbosity "Project settings changed, reconfiguring..."
+          projectConfig <- phaseReadProjectConfig
+          localPackages <- phaseReadLocalPackages projectConfig
+          return (projectConfig, localPackages)
+
+    info verbosity
+      $ unlines
+      $ ("this build was affected by the following (project) config files:" :)
+      $ [ path
```
alexbiehl: Maybe make this a list of filepaths by prepending '- ' to the paths.

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -294,48 +294,61 @@ sanityCheckElaboratedPackage ElaboratedConfiguredPackage{..}
...
+                  , distProjectCacheDirectory
+                  , distProjectFile
+                  }
+      = distDirLayout
+    ProjectConfigShared { projectConfigConfigFile } =
+      projectConfigShared cliConfig
+
+    fileMonitorProjectConfig =
+      newFileMonitor (distProjectCacheFile "config") :: FileMonitor
+          (Flag FilePath, FilePath)
```
alexbiehl: we should not pass a `Flag FilePth` into the monitor.  I suggest you use `getConfigFilePath` to resolve to properly resolve the config file path.

cabal-install/Distribution/Client/ProjectConfig.hs
lspitzner: The downside to that is that the reader might very well expect the same kind of logic behind the three `read` functions, especially when they have so similar names (`readProjectLocalConfig`, `readProjectLocalFreezeConfig`, `readProjectLocalExtraConfig`).

The `addProjectFileProvenance` change is not necessary afaict.

cabal-install/Distribution/Client/ProjectPlanning.hs
lspitzner: Right, sorry about that. I blanket-applied `brittany` to some stuff and did not revert these unnecessary changes. I will fix that.

cabal-install/Distribution/Client/ProjectPlanning.hs
lspitzner: You mean a new datatype, right?

cabal-install/Distribution/Client/ProjectPlanning.hs
lspitzner: Ah, good I added that type signature to counteract the `NoMonoLocalBinds` then :D

cabal-install/Distribution/Client/ProjectPlanning.hs
alexbiehl: I think tuple is ok here. 

I meant something like 
  ```
  let 
    fileMonitorProjectConfigInput = 
      (..., ...)
```

File: cabal-install/Distribution/Client/ProjectConfig.hs (hunk 2 above)
Comment (alexbiehl) [reply cabal-install/Distribution/Client/ProjectConfig.hs]: Right, would it make sense to pass the implicit config as a parameter in case the file doesn't exist? Then you could handle the "file does not exist case"  in `readProjectFile`. Does that make sense?

**Comment by lspitzner** [comment]
@alexbiehl Squashed the changes, hope that is alright.

This probably still contains some stylistic changes, but have a really hard time determining what the proper style is, as the codebase is not very consistent. I might also argue that the prevalent use of hanging indentation is also at fault for creating larger diffs than necessary, for example on the function renames in this commit.

Unless I have missed something, all comments should be addressed, apart from "pass the implicit config as a parameter". With the rename, the code seems to be sufficiently clear and free of redundancies already.

File: cabal-install/Distribution/Client/ProjectConfig.hs [hunk 7]
```diff
@@ -480,19 +474,22 @@ readProjectLocalExtraConfig verbosity distDirLayout =
...
 
 -- | Reads a named config file in the given project root dir, or returns empty.
 --
-readProjectExtensionFile :: Verbosity -> DistDirLayout -> String -> FilePath
-                         -> Rebuild ProjectConfig
-readProjectExtensionFile verbosity DistDirLayout{distProjectFile}
+readProjectFile :: Verbosity
+                -> DistDirLayout
+                -> String
+                -> String
```
alexbiehl: I think this used to be FilePath

lspitzner:
Oh, there was one other thing I wanted to ask: Would it make sense to output this at lower verbosity? It seems sufficiently important, almost (and i don't like to have to wade through the `-v2` output just to see this).

cabal-install/Distribution/Client/ProjectConfig.hs
lspitzner: Which is clearly wrong, is it not? It is the `extensionDescription` after all, a string meant for displaying purposes.

alexbiehl: yeah, sorry

lspitzner:
Thanks for the very prompt reviewing!

alexbiehl:
@23Skidoo can we merge this?

cabal-install/Distribution/Client/ProjectConfig.hs
```diff
@@ -480,19 +474,22 @@ readProjectLocalExtraConfig verbosity distDirLayout =
...
+                -> Rebuild ProjectConfig
+readProjectFile verbosity DistDirLayout{distProjectFile}
                          extensionName extensionDescription = do
     exists <- liftIO $ doesFileExist extensionFile
     if exists
       then do monitorFiles [monitorFileHashed extensionFile]
-              liftIO readExtensionFile
+              addProjectFileProvenance <$> liftIO readExtensionFile
       else do monitorFiles [monitorNonExistentFile extensionFile]
```
23Skidoo: Perhaps we shouldn't mention "extension file" here now that the function has been renamed.

23Skidoo:
Merged, thanks!

23Skidoo:
@lspitzner Congrats on commit number 10000!

lspitzner:
@23Skidoo 
Heh, thanks. If I could I would pass this on to the much more frequent contributors. Perhaps we can just view it more as a collective achievement.

PR #5250, opened 2018-04-02
HaskellSuite supports multiple package instances.

Lemmih:

I'm not 100% sure what the 'registerMultiInstance' flag means but since it's always enabled when building packages, HaskellSuite should definitely support it.


Lemmih:
Did HaskellSuite at some point /not/ support multiple package instances? I'm more interested in why 'registerMultiInstance' is always True. That this flag is always true means that UHC and JHC are broken at the moment. Not sure if either is being actively developed anymore, though.

23Skidoo:
I don't know whether it didn't. We probably don't care.

PR #5251, opened 2018-04-02
Fix doc link to --constraint option

Lysxia:
The correct link:

https://www.haskell.org/cabal/users-guide/installing-packages.html#cmdoption-setup-configure-constraint

previously was `...configure--constraint` instead of `...configure-constraint`.

---


23Skidoo:
Merged, thanks!

Issue #4969, opened 2017-12-22
cabal haddock fails with internal libraries


shlevy:
When using internal libraries, cabal haddock seems to fail unconditionally:

```shell
$ cabal haddock
Preprocessing library 'internal' for internal-library-test-0.1.0.0..
Running Haddock on library 'internal' for internal-library-test-0.1.0.0..
Haddock coverage:
   0% (  0 /  2) in 'Internal'
  Missing documentation for:
    Module header
    foo (internal/Internal.hs:3)
Documentation created: dist/doc/html/internal-library-test/index.html
Preprocessing library for internal-library-test-0.1.0.0..
Running Haddock on library for internal-library-test-0.1.0.0..
cabal: internal error when calculating transitive package dependencies.
Debug info: []
```

23Skidoo:
See also #4215, #1085 and #1919. Would be nice to fix all these in one go...

hvr:
fwiw, I managed to generate haddocks back then in July for http://hackage.haskell.org/package/haddock-library

23Skidoo:
Hmm, so broken only with old-build? Interesting.

shlevy:
Is the plan to not fix this and wait for new-haddock?

ezyang:
Yeah, I'm pretty OK with just saying "use new-build" for this one.

peti:
~~~
$ runhaskell Setup.hs new-haddock
unrecognised command: new-haddock (try --help)
~~~

Now what? I think this bug is quite nasty and it effectively prevents all kinds of distributors like Nix and openSUSE from generating Haddock documentation for any package that uses an internal library. :disappointed:

hvr:
@peti There's not going to be a `Setup.hs new-haddock` just as there isn't going to be a `Setup.hs new-build` command, as that's a low-level interface (c.f. "plumbing" in the Git model)  which  `cabal new-haddock` (the "porcelaine") builds upon.

TLDR, you need to invoke `Setup.hs haddock` for each component individually.

peti:
> You need to invoke Setup.hs haddock for each component individually, similiar to how cabal new-haddock does.

[...]

PR #5253, opened 2018-04-03
Haddock: add internal libraries to package index

alexbiehl:
`setup haddock` now updates the package index with each internal library it haddocks. This fixes https://github.com/haskell/cabal/issues/4969 

(don't mind the commit from https://github.com/haskell/cabal/pull/5226 here)

/cc @shlevy

---

### DIFF
--- a/Cabal/Distribution/Simple/Build.hs
+++ b/Cabal/Distribution/Simple/Build.hs
@@ -23,6 +23,7 @@ module Distribution.Simple.Build (
     startInterpreter,
 
     initialBuildSteps,
+    createInternalPackageDB,
     componentInitialBuildSteps,
     writeAutogenFiles,
   ) where
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -53,6 +53,8 @@ import Distribution.Simple.BuildTarget
 import Distribution.Simple.InstallDirs
 import Distribution.Simple.LocalBuildInfo hiding (substPathTemplate)
 import Distribution.Simple.BuildPaths
+import Distribution.Simple.Register
+import qualified Distribution.Simple.Program.HcPkg as HcPkg
 import qualified Distribution.Simple.PackageIndex as PackageIndex
 import qualified Distribution.InstalledPackageInfo as InstalledPackageInfo
 import Distribution.InstalledPackageInfo ( InstalledPackageInfo )
@@ -66,9 +68,10 @@ import Language.Haskell.Extension
 
 import Distribution.Compat.Semigroup (All (..), Any (..))
 
+import Control.Monad
 import Data.Either      ( rights )
 
-import System.Directory (doesDirectoryExist, doesFileExist)
+import System.Directory (getCurrentDirectory, doesDirectoryExist, doesFileExist)
 import System.FilePath  ( (</>), (<.>), normalise, isAbsolute )
 import System.IO        (hClose, hPutStrLn, hSetEncoding, utf8)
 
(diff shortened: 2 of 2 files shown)


alexbiehl:
Sorry about the diff noise for now, I will rebase later.

alexbiehl:
Alright, rebased! Off we go! 

@shlevy I tried compiling `haddock-library's `Setup.hs` locally. And ran `setup haddock attoparsec haddock-library`

shlevy:
@alexbiehl Cool! So the recommended workflow for packaging from here on in will be to list all of the libraries in the `setup haddock` invocation? Or is there a way to make `setup haddock` standalone do the right thing?

alexbiehl:
I just tried `haddock-library` and the following works:

```
$ ./Setup configure
$ ./Setup build
$ ./Setup haddock
``` 

And it does the right thing (tm).

shlevy:
Awesome! cc @peti how do you want to integrate this?

peti:
I'll have to think about this ... Off the cuff, I see the following options:

- Always run builds with the latest available Cabal release (plus patches that we'd like to cherry-pick), regardless of which compiler is driving the build. The generic Nix builder hasn't done that in the past because we had super old compilers like GHC 6.12.3 that could not build recent Cabal versions. So we ended up with the current environment, where the compiler's native Cabal version drives the build. If we'd go for using the latest Cabal for all builds with all compilers, then we could simplify our builder a bit (and easily introduce patches like this one). The downside is that we need to build that Cabal version with all compilers before we can use it, which may or may not be a problem.

- Apply that patch to our `ghc` so that the bundled core library can run `./Setup haddock` successfully.

- Manually patch the handful of packages that use internal libraries by way of override to include a (patched) Cabal version into their `setupHaskellDepends`.

- Do nothing and live with missing Haddock documentation until a new compiler comes out. I understand that GHC 8.4.2 is just around the corner, so if this patch makes it into that release, then we'll get it reasonably soon anyway.

shlevy:
I'm leaning toward the "do nothing" option, *if* this will make it in soon.

gbaz:
Given that we're near-final on 8.4.2, I don't see this making it in for that release.

23Skidoo:
Rebased, let's see what Travis says.

alexbiehl:
Travis likes it!

23Skidoo:
Merged, thanks!

Infinisil:
Assuming this fix is in the version I use (GHC 8.6.5, Cabal 2.4.0.1). This doesn't seem to work with an internal library depending on the non-internal one. I made a reproducible example here: https://github.com/Infinisil/haddock-internal-repro. The error is

```
$ ghc Setup.hs -o Setup
$ ./Setup haddock
Preprocessing library for repro-0.1.0.0..
Running Haddock on library for repro-0.1.0.0..
Haddock coverage:
  50% (  1 /  2) in 'Lib'
  Missing documentation for:
    Module header
Documentation created: dist/doc/html/repro/index.html
Preprocessing library 'internal' for repro-0.1.0.0..
Running Haddock on library 'internal' for repro-0.1.0.0..
Setup: internal error when calculating transitive package dependencies.
Debug info: []
```

My versions are
- ghc-8.6.5
- Cabal-2.4.0.1

It works no problem if I make the internal library not depend on the non-internal one. This might have been an oversight in this PR, because in the `haddock-library` example this [didn't occur](https://github.com/haskell/haddock/blob/979c7338cfcdc59f0b0dda562a53558c416cc362/haddock-library/haddock-library.cabal#L51-L82) (`attoparsec` doesn't depend on `haddock-library`).

Infinisil:
This is also the reason for @peti reverting the generation of packages with internal libraries in nixpkgs with https://github.com/NixOS/cabal2nix/pull/406, which mentions the [`stylish-cabal` build](https://hydra.nixos.org/build/88843163) failing, whose internal library also depends on the `stylish-haskell` library: https://github.com/pikajude/stylish-cabal/blob/d0b32857a70025f8bed9e42ed629bbc82896d5da/stylish-cabal.cabal#L69-L90.

PR #5259, opened 2018-04-13
Fix architecture mismatch of the Paths_ modules on Windows

### LINKED ISSUE #5254: Linker error on Windows with chocolate
Here is the log exhibiting the linker failure with cabal 2.2.0.0: https://ci.appveyor.com/project/Mikolaj/allure/build/1.0.148

```
Linking dist\build\Allure\Allure.exe ...
Warning: resolving _GetModuleFileNameW by linking to _GetModuleFileNameW@12
Use --enable-stdcall-fixup to disable these warnings
Use --disable-stdcall-fixup to disable these fixups
    `_GetModuleFileNameW' referenced in section `.text' of dist\build\Allure\Allure-tmp\Paths_Allure.o: defined in discarded section `.text' of C:/ProgramData/chocolatey/lib/ghc/tools/ghc-8.2.2/mingw/bin/../lib/gcc/i686-w64-mingw32/6.2.0/../../../../i686-w64-mingw32/lib/../lib/libkernel32.a(dghgds00535.o)
collect2.exe: error: ld returned 1 exit status
```

And the same project succeeds after downgrading cabal to 2.0.0.1: https://ci.appveyor.com/project/Mikolaj/allure/build/1.0.149

I both cases, initial 64bit build succeed fine, only the subsequent 32bit build fails (with the new cabal version).

I tried `--enable-stdcall-fixup`, which made the warnings disappear, but the error remain. Then I tried `--disable-stdcall-fixup` and I got a different error (as before, only in 32bit build, https://ci.appveyor.com/project/Mikolaj/allure/build/1.0.150):

```
Linking dist\build\Allure\Allure.exe ...
dist\build\Allure\Allure-tmp\Paths_Allure.o:fake:(.text+0xe65): undefined reference to `GetModuleFileNameW'
collect2.exe: error: ld returned 1 exit status
```

When comparing the logs, nothing strikes me, except that only the 64bit compilation (the one that doesn't fail) shows the warning (but only with the old cabal):

```
dist\build\Allure\autogen\Paths_Allure.hs:73:1: warning: [-Wunsupported-calling-conventions]
    * the 'stdcall' calling convention is unsupported on this platform,
      treating as ccall
    * When checking declaration:
        foreign import stdcall unsafe "windows.h GetModuleFileNameW" c_GetModuleFileName
          :: Ptr () -> CWString -> Int32 -> IO Int32
   |
73 | foreign import stdcall unsafe "windows.h GetModuleFileNameW"
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

(So it seems as if old cabal had the stdcall problem only on 64bit, but the fixes changed also how 32bit builds work and broke them.)

Also, the log shows some new chocolatey Windows shims are present in the new chocolatey cabal package. But they probably only add the exes to path, so should be benign. 

The error does not appear if the exe is in the same cabal package as the lib: https://ci.appveyor.com/project/Mikolaj/lambdahack/build/1.0.428

Some vaguely related links:

https://mcuoneclipse.com/2017/11/12/solving-problem-with-gnu-linker-and-referenced-in-section-defined-in-discarded-section-error-message
https://stackoverflow.com/questions/31085538/function-referenced-in-section-text-of-defined-in-discarded-section-te
https://ghc.haskell.org/trac/ghc/ticket/9524
https://ghc.haskell.org/trac/ghc/ticket/14300

Mistuke:

Cabal is causing an abi issue by forcing it's own architecture's calling convention on each module it creates in the `Path_` modules on Windows.
Compiling a 64 bit executable with a 32 bit cabal happens to work because the linker can perform fixups, adding an indirection to to correct the calling convention. However compiling a 32 bit executable from a 64 bit cabal will fail since the linker can't change the callee.

This corrects that by if cpp is enabled to use predefines from the compiler to determine what to do. Unfortunately due to a bug in GHC I have to rely on macro GHC puts out instead of the standard platform macros. So this only works for GHC code.

It has the old method as fallback for if cpp is not supported. Fixes #5254

Tested it by building a cabal-install, then using it to build cabal-install again.

### DIFF
--- a/Cabal/Distribution/Simple/Build/PathsModule.hs
+++ b/Cabal/Distribution/Simple/Build/PathsModule.hs
@@ -243,13 +243,13 @@ generate pkg_descr lbi clbi =
 
         paths_modulename = autogenPathsModuleName pkg_descr
 
-        get_prefix_stuff = get_prefix_win32 buildArch
+        get_prefix_stuff = get_prefix_win32 supports_cpp buildArch
 
         path_sep = show [pathSeparator]
 
         supports_cpp = supports_language_pragma
-        supports_rebindable_syntax= ghc_newer_than (mkVersion [7,0,1])
         supports_language_pragma = ghc_newer_than (mkVersion [6,6,1])
+        supports_rebindable_syntax= ghc_newer_than (mkVersion [7,0,1])
 
         ghc_newer_than minVersion =
           case compilerCompatVersion GHC (compiler lbi) of
@@ -280,8 +280,8 @@ get_prefix_reloc_stuff =
   "  let (bindir,_) = splitFileName exePath\n"++
   "  return ((bindir `minusFileName` bindirrel) `joinFileName` dirRel)\n"
 
-get_prefix_win32 :: Arch -> String
-get_prefix_win32 arch =
+get_prefix_win32 :: Bool -> Arch -> String
+get_prefix_win32 supports_cpp arch =
   "getPrefixDirRel :: FilePath -> IO FilePath\n"++
   "getPrefixDirRel dirRel = try_size 2048 -- plenty, PATH_MAX is 512 under Win32.\n"++
   "  where\n"++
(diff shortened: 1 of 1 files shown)


hvr:
So does this mean, that with this fix, you will get the same result regardless of whether `cabal` was compiled for 32bit or 64bit on Windows?

Mistuke:
@hvr yup, now you should be able to target both architectures regardless of which cabal you use.

23Skidoo:
Merged, thanks!

23Skidoo:
Should this be backported?

23Skidoo:
Cherry-picked into 2.2.

Mistuke:
Thanks!

PR #5263, opened 2018-04-16
fix infinite loop in D.Client.TargetSelector

### LINKED ISSUE #5081: Internal error/infinite loop in D.Client.TargetSelector
I reproduced the issue on master (378153142745135c203f8301a2af16fe55528ea2) by running the following command in the cabal repository:

```
$ cabal new-build test:unknown-test-suite
cabal: Internal error in target matching. It should always be possible to find
a syntax that's sufficiently qualified to give an unambiguous match. However
when matching 'test:unknown-test-suite' we found test:unknown-test-suite
(unknown-component) which does not have an unambiguous syntax. The possible
syntax and the targets they match are as follows:
cabal: <<loop>>
```

It also occurs when I run the same command on a simple project with one directory containing a simple .cabal file:

```
name: example
version: 1.0
build-type: Simple
cabal-version: >= 1.2

library
```

The error message is from D.Client.TargetSelector: https://github.com/haskell/cabal/blob/378153142745135c203f8301a2af16fe55528ea2/cabal-install/Distribution/Client/TargetSelector.hs#L653

gelisam:
The pattern guard was clearly meant as a list-comprehension filter,
but on a let-bound pattern, it instead caused a circular loop in which
whether the LHS variables are bound the the RHS values depends on the
values of the LHS variables.

fixes #5081

---

I tested this by running the problematic command from `#5081`. Result before the fix:

    $ cabal new-build test:unknown-test-suite
    cabal: Internal error in target matching. It should always be possible to find
    a syntax that's sufficiently qualified to give an unambiguous match. However
    when matching 'test:unknown-test-suite' we found test:unknown-test-suite
    (unknown-component) which does not have an unambiguous syntax. The possible
    syntax and the targets they match are as follows:
    cabal: <<loop>>

Result after the fix:

    $ cabal new-build test:unknown-test-suite
    cabal: Internal error in target matching. It should always be possible
    to find a syntax that's sufficiently qualified to give an unambiguous match.
    However when matching 'test:unknown-test-suite' we found
    test:unknown-test-suite (unknown-component) which does not have an unambiguous
    syntax. The possible syntax and the targets they match are as follows:
    'test:unknown-test-suite' which matches test:unknown-test-suite
    (unknown-component), :pkg:test:lib:test:file:unknown-test-suite (unknown-file)

I also tested with `all:test` (typo for `all:tests`), which has similar results before and after the fix.

### DIFF
--- a/cabal-install/Distribution/Client/TargetSelector.hs
+++ b/cabal-install/Distribution/Client/TargetSelector.hs
@@ -611,8 +611,9 @@ disambiguateTargetSelectors matcher matchInput exactMatch matchResults =
             Left  ( originalMatch
                   , [ (forgetFileStatus rendering, matches)
                     | rendering <- matchRenderings
-                    , let (Match m _ matches) | m /= Inexact =
+                    , let Match m _ matches =
                             memoisedMatches Map.! rendering
+                    , m /= Inexact
                     ] )
 
       | (originalMatch, matchRenderings) <- matchResultsRenderings ]


alexbiehl:
This looks good to me! Let's merge this!

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.2.

PR #5264, opened 2018-04-16
Bump upper bound on unix to < 2.9

RyanGlScott:
See https://ghc.haskell.org/trac/ghc/ticket/15042.

I ran the `Cabal` test suites against `unix-2.8`, and they all passed.


RyanGlScott:
Those test suite failures don't appear to be my doing (and should be resolved by #5267), so I'm going to be bold and just land this.

PR #5266, opened 2018-04-17
Prototype flag normalisation for GHC 8.0-8.4

merijn:
Uses the just merged infrastructure to normalise commandline arguments to filter out any GHC flags that don't affect the build products. Currently only filters commandline arguments for GHC >=8.0 && <8.5 as those are the only versions I checked the flags for.

I went through the GHC DynFlags source for all those version to figure out which flags can be filtered and to follow the right format. Anything I wasn't sure about is left alone, some flags I'm still unsure about are documented in #4247.

Also filters out all haddock arguments, as those should never affect GHC build products.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -254,6 +254,7 @@ library
     Distribution.Simple.Command
     Distribution.Simple.Compiler
     Distribution.Simple.Configure
+    Distribution.Simple.Flag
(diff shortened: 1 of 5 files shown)


merijn:
Ok, I've been using this code for about a week now with no issues, none of the people in the original issue or that I've asked have objected, so unless someone has a complaint before tomorrow I'm merging this.

Issue #5232, opened 2018-03-25
`cabal new-build` treats haddock failures fatally


hvr:
originally reported at https://github.com/haskell-infra/hackage-trustees/issues/146

`cabal new-build` currently fails when `haddock` fails (e.g. due to `documentation: True` set globally; note that Haddock currently fails for packages w/o any local modules);
instead, `cabal new-build` ought to merely record/log/warn about them, but otherwise carry on

cartazio:
this seems like a high priority bug :) 

where in the cabal code base is this issue? i can poke around but if theres some know how, i'd like to help get this into a point fix post haste!

alexbiehl:
Carter, look here https://github.com/haskell/cabal/blob/457ebb8aff13e86e8928f301132f14ccaa23c28f/cabal-install/Distribution/Client/ProjectBuilding.hs#L941-L945

`annotateFailureNoLog` catches the error from cabal and rethrows it. You probably don't want that behaviour in case of haddock.

cartazio:
@alexbiehl ok, that helps!
better question, should we fix just that, or also not generate haddocks when theres no modules? (later seems more complicated perhaps but more correct?)

alexbiehl:
We should do both I think:

1) Cabal shouldn't treat haddock errors as fatal errors and 2) Cabal should avoid invoking haddock if there are no inputs.

cartazio:
@alexbiehl is your haddock work on master a fix to for master, or unrealted clean up?

should i do this patch for 2.2 or master?

cartazio:
i'm trying to test build cabal and i'm getting this fun error 

Resolving dependencies...
Configuring cabal-install-2.2.0.0...
Building cabal-install-2.2.0.0...
Installed cabal-install-2.2.0.0
Updating documentation index
/Users/carter/.cabal/share/doc/x86_64-osx-ghc-8.2.2/index.html
<command line>: cannot satisfy -package-id cabal-testsuite-2.2.0.0-inplace
    (use -v for more information)

cartazio:
this hacky first cut of a patch seems to work now that i got the build working :)

cartazio:
the current patch just demotes haddock failures unconditionally to an adhoc warning, the right approach might be to add support for "warnings" along with full build failures.  

additionally: it'd probably be good to have / retain the option to treat haddock failures as errors either per package build or globally?

additionally additionally: cabaly shouldn't even invoke haddock when theres no docs to build probably?!

[...]

PR #5269, opened 2018-04-17
Fixed new haddock 2.2 - resolves partly https://github.com/haskell/cabal/issues/5232

cartazio:
attempt at fixing https://github.com/haskell/cabal/issues/5232 

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -941,8 +941,8 @@ buildAndInstallUnpackedPackage verbosity
     whenHaddock $ do
       when isParallelBuild $
         notice verbosity $ "Generating " ++ dispname ++ " documentation..."
-      annotateFailureNoLog HaddocksFailed $
-        setup haddockCommand haddockFlags
+      -- handle documentation build failures correctly ish
+      annotateIgnoredFailureNoLog "a new documentation build failure: " $ setup haddockCommand haddockFlags
 
     -- Install phase
     annotateFailure mlogFile InstallFailed $ do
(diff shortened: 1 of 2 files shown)


cartazio:
woops, initially based it on master rather than 2.2 for the PR, fixed that

cartazio:
@hvr  @alexbiehl  @gbaz  the more precise fix is also needed, but I also dont know how to wire up the warning logging to the usual output....

even this PR alone should fix up the main main issue in 2.2 cabal-install ... though the more polished one is needed to... 

how do we get this IN and released?

cartazio:
@23Skidoo  thoughts?

alexbiehl:
Sorry for the late reply. I think this is an OK way to go. Cabal doesn't give us too many opportunities to hook into error handling. But: I think we should really reuse error reporting from the hard failing one to have consistent look of error messages.

cartazio:
Makes sense. I wasn’t sure how to do that.

It still needs to do the no docs bit

What would be a better than OK approach?

cabal-install/Distribution/Client/ProjectBuilding.hs
```diff
@@ -1371,6 +1371,25 @@ withTempInstalledPackageInfoFile verbosity tempdir action =
...
+#if MIN_VERSION_base(4,7,0)
+      Handler $ \async -> throwIO (async :: SomeAsyncException)
+#else
+      Handler $ \async -> throwIO (async :: AsyncException)
+#endif
+    , Handler $ \other -> handler (other :: SomeException)
+    ]
+  where
+    handler :: Exception e => e -> IO ()
+    handler except= do putStr system ; putStrLn (show except)
```
23Skidoo: Should use `warn` instead of `putStr[Ln]`.

cartazio:
Thx! I’ll see if I have time after Tuesday.  But if you wanna donate this
to a new contributor at zurihac please feel welcome to

cartazio:
so action items i'm gonna update this for today

1) dont run haddock  when a package is empty

2) move to using warn


@23Skidoo hrmm, beyond adding warn, how do you propose i use/adapt the dieOnBuildFailures to this case?

cartazio:
ok, i pushed some partially done changes,
I added to the haddock exports a "compute input files " function,
though i'm not sure how i get the arguments i need for it in the `ProjectBuilding.hs` file

```
haddockInputFilesFromLibrary :: Verbosity
            -> LocalBuildInfo
            -> ComponentLocalBuildInfo
            -> Library
            -> IO [FilePath]
haddockInputFilesFromLibrary  verbosity lbi clbi lib =
  map snd `fmap` getLibSourceFiles verbosity lbi lib clbi

```

typedrat:
Superseded by #5459.

PR #5275, opened 2018-04-19
BUG/RFC: Fix `scope` field backwards compatibility

DanielG:
This reverts commit 442dcf88d809bcfa9c68818947683f02e7cde47d.

[cabal-helper](https://github.com/DanielG/cabal-helper/blob/master/cabal-helper.cabal#L34) declares a lower cabal spec version than would ordinarily be required for support for the `scope` field on purpose and then handles the fallout in [`Setup.hs`](https://github.com/DanielG/cabal-helper/blob/master/Setup.hs#L34) in order to support much older versions of Cabal.

The aforementioned commit broke the behavior of this package by introducing an `availableSince` declaration which forces the value of the `exeScope` field to `ExecutablePublic` in case the spec version is below the required version.

Now my initial instinct was to just remove the `availableSince` and leave it at that but since the associated `checkSpecVersion` was removed in favor of `availableSince` this won't do.

Hence I'm here to ask how this situation should be handled? Would it be acceptable to just add back the `checkSpecVersion` for the scope field and remove the `availableSince` declaration or is there some other way to fix this I'm not seeing?

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -53,6 +53,7 @@ import Distribution.System
 import Distribution.Text
 import Distribution.Types.ComponentRequestedSpec
 import Distribution.Types.CondTree
+import Distribution.Types.ExecutableScope
 import Distribution.Types.ExeDependency
 import Distribution.Types.UnqualComponentName
 import Distribution.Utils.Generic                    (isAscii)
@@ -315,6 +316,12 @@ checkExecutable pkg exe =
       PackageBuildImpossible $
            "On executable '" ++ display (exeName exe) ++ "' an 'autogen-module' is not "
         ++ "on 'other-modules'"
+
+  , checkSpecVersion pkg [2,0] (exeScope exe /= ExecutableScopeUnknown) $
+      PackageDistSuspiciousWarn $
+           "To use the 'scope' field the package needs to specify "
+        ++ "at least 'cabal-version: >= 2.0'."
+
   ]
   where
     moduleDuplicates = dups (exeModules exe)
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -52,7 +52,6 @@ import Distribution.Parsec.Common
 import Distribution.Parsec.Newtypes
 import Distribution.Parsec.ParseResult
 import Distribution.Text                      (display)
-import Distribution.Types.ExecutableScope
 import Distribution.Types.ForeignLib
 import Distribution.Types.ForeignLibType
 import Distribution.Types.UnqualComponentName
@@ -159,8 +158,7 @@ executableFieldGrammar
 executableFieldGrammar n = Executable n
     -- main-is is optional as conditional blocks don't have it
     <$> optionalFieldDefAla "main-is" FilePathNT L.modulePath ""
-    <*> optionalFieldDef    "scope"              L.exeScope ExecutablePublic
-        ^^^ availableSince [2,0] ExecutablePublic
+    <*> monoidalField       "scope"              L.exeScope
     <*> blurFieldGrammar L.buildInfo buildInfoFieldGrammar
 {-# SPECIALIZE executableFieldGrammar :: UnqualComponentName -> ParsecFieldGrammar' Executable #-}
 {-# SPECIALIZE executableFieldGrammar :: UnqualComponentName -> PrettyFieldGrammar' Executable #-}
--- a/Cabal/Distribution/Types/ExecutableScope.hs
+++ b/Cabal/Distribution/Types/ExecutableScope.hs
@@ -16,18 +16,23 @@ import qualified Distribution.Compat.CharParsing as P
 import qualified Distribution.Compat.ReadP as Parse
 import qualified Text.PrettyPrint as Disp
 
-data ExecutableScope = ExecutablePublic
+data ExecutableScope = ExecutableScopeUnknown
+                     | ExecutablePublic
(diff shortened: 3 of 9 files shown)


phadej:
There is only approve or request changes, but I simply disapprove. I see the problem, but this is not the right fix.

Note: `setup-depends` (in custom-setup) and `build-depends` (in library) can be different. It's totally fine for former to be stricter:
- You need new Cabal to compile the `./Setup`
- but it's ok to use old Cabal to compile the library itself

**More correct solution:** In fact if I read `Setup.hs` right, `cabal-helper` can be `build-type: Simple` with tight enough `cabal-version: `. No custom setup, no problems.

Then you'll only need recent enough `cabal-install` or `stack` (which would use the `Cabal` lib they are compiled with to setup `cabal-helper`). If you insist on supporting old `cabal-install`, I don't see a solution.

DanielG:
@phadej the problem is I support cabal-install/Cabal version that don't even implement setup-depends yet! If I drop support for older cabal versions you're right I could switch to `build-type:Simple` and all is good with the world.

Either way this is a change in behavior and I think the warning should at least say the offending field will be ignored or something.

DanielG:
Actually I think its completely unacceptable for a newer Cabal version to break existing pacakges, I don't think I can guard my library against newer Cabal versions, right? So how do I unbreak the existing packages on Hackage with the new Cabal version now?

DanielG:
@phadej ping. Given that packages already on Hackage were broken by this change in Cabal with no apparent way to do fix them by doing a revision or somesuch what would be an acceptable solution to fixing these packages?

The way I see it Cabal should never break existing packages, right? So this has to be considered a bug at least for packages that don't declare a spec version >=2.2 or something.

DanielG:
@23Skidoo The whole reason I contributed the `scope: private` stuff was to get rid of Setup.hs in cabal-helper down the road :cry: and I already do use `x-scope` in Setup.hs.

I intended this sort of transition to be possible from the beginning and no-one told me it's wrong to do this so I just went with it, remember https://github.com/haskell/cabal/pull/4714? I still maintain that the current (2.2) behavior is a bug relative to my original design and (I think I haven't actually mentioned this) the behavior that 2.0 had and has to be fixed one way or another.

phadej:
I'm strict on this. We should not "patch" old versions of Cabal spec. Two options:
- `x-scope`, Custom setup, lower `cabal-version`,
- `scope`, possibly Simple build-type, `cabal-version: 2.0` - only newer cabal-install

Bumping `cabal-version` and (therefore) requiring newer `cabal-install` is IMO smooth enough transition.

Something in between *"newer and older cabal-install will do different things with same (old cabal-version) .cabal"* is only going to cause confusion.

`build-tool-depends` is unfortunate exception, which already causes confusion! And that exception we had to make. Not really proud of that hack.

DanielG:
@gbaz I am moving to that for the next release, but that still leaves the older releases dead in the water. Any user who upgrades to cabal-install-2.2 right now will not be able to install a working cabal-helper/ghc-mod until I release a new version. Even after I do the solver could be forced to fall back to older versions which will just fail to work because the private executable cannot be found in the right spot.

@phadej Again this used to work with Cabal-2.0 and with Cabal-2.2 it doesn't. So we're already in the   »Something in between "newer and older cabal-install will do different things with same (old cabal-version) .cabal" is only going to cause confusion. « situation!! Which is precisely what I'm trying to fix.

I feel like we're having some trouble communicating here.

gbaz:
`Even after I do the solver could be forced to fall back to older versions which will just fail to work because the private executable cannot be found in the right spot.`

Could revisions help with that?

DanielG:
@gbaz  I don't think it's possible to even say: "This package only works when built using Cabal lib <=2.2" let alone is it possible to do that in a Hackage revision. If I'm wrong and that is in fact possible then that would be a perfectly acceptable solution. But as far as I understand that's not the case.

gbaz:
What I mean is there are three versions in the 8.0 series that use both `x-scope` and `scope`. What if the cabal files were modified in them to remove `scope` or something similar?

DanielG:
@gbaz Well that would have the same effect as 2.2 already has. It will break the packages because they expect the scope:private executable to end up in `$libexecdir` and then they wont.

gbaz:
Oh, I see -- the problem is you explicitly case on cabal version in the Setup.hs, which is of course not subject to revisions. I have two ideas, but you probably won't like either.

The first is that the "bad" versions could just be deprecated, and you could upload two new versions --  one that works with a prior cabal spec version and uses `x-scope` always (to replace the deprecated ones) -- and one that works with 2.0 and uses `scope` always. The second is similar, except instead of deprecating the bad versions, you amend their cabal files to bump their cabal versions to 2.0. (But again, you could upload another version with `x-scope` and the old workaround).

DanielG:
I've been playing with my reproducer some more and realized I don't have a `setup-depends` section in it. If I add one with `Cabal < 2.1` which cabal-helper also has everything just works as it should. It seems my earlier assumption that I can't gate against Cabal-2.2 was false after all, I've already been gating against it :)

I can't seem to reproduce the failure that prompted me to report this in the first place anymore. I still have logs that show cabal-install picking Cabal-2.2 to build cabal-helper, [here](https://gist.github.com/DanielG/9bdd012fc01e2024691963f77826cabb) so I am pretty sure I didn't imagine it. Weird..

@gbaz Thanks for your suggestions, somehow they made me thing of setup-depends which I'd forgotten about before. Glad I won't have to put them to use though ;)

23Skidoo:
Looks like @DanielG found an acceptable alternative workaround for his problem, and @phadej is still strongly opposed to merging this PR. Closing, thanks to everyone for their contributions to the discussion.

PR #5277, opened 2018-04-20
Add check for "ghc-options: -j"

lspitzner:


quasicomputational:
Good idea. Can I ask for a changelog note and a new test for this behaviour? It looks simple but I know I've written simple and wrong code, and it'd keep the feature from accidentally breaking later.

PR #2522, opened 2015-04-04
Implement bash (with globstar) style globbing

hdgarrood:
Fixes #784, #2030, resolves #713, subsumes #1343, #1344. Picks up where #1975 left off.

Implement a reduced form of GNU bash style globbing. The supported
features should account for 99% of use cases.

Still to do:

- [ ] minimum cabal-version constraint (etc). This seems to be
necessary because, for example, this commit changes the meaning of *.js,
which previously would not have matched jquery.cookie.js, and now does.
- [ ] should we add a limitation to disallow `*` and `**` from appearing as the very last item of a glob pattern, to avoid including unwanted items? The documentation in this PR currently describes this limitation but I haven't implemented it yet.
- [ ] test correct behaviour re cabal version constraints in all four cases
- [x] ensure that the new globs are actually being used in all of the cabal file fields that we expect them to be, that is, `data-files`, `extra-source-files`, `extra-doc-files` (I did this manually).

Issue #2030, opened 2014-08-05
Incorrect error message about globbing


bos:
In the criterion package, I tried to have a `data-files` section like this:

```
data-files:
  templates/js/jquery-*.min.js
```

When I try to run `cabal`, I get this error:

```
cabal: invalid file glob 'templates/js/jquery-*.min.js'. Wildcards '*' are
only allowed in place of the file name, not in the directory name or file
extension. If a wildcard is used it must be with an file extension.
```

The actual name of the file I'm trying to include is as follows:

```
templates/js/jquery-1.6.4.min.js
```

It looks like `cabal` is misidentifying where the "`*`" is showing up in the pattern here.

quasicomputational:
The error message has been improved.

Issue #4935, opened 2017-12-05
Manual does not describe how to escape metacharacters


quasicomputational:
The [section of Cabal's manual describing file globbing](http://cabal.readthedocs.io/en/latest/developing-packages.html#pkg-field-data-files) does not cover how to escape metacharacters. It ought to say how to include a file with a literal `*` in its name.

23Skidoo:
I'm actually not sure whether that is at all supported.

mitchellwrosen:
I'm not sure either but I can find out for you in a couple hours.

mitchellwrosen:
It's not possible to include a file with a literal `*` in its name: https://github.com/haskell/cabal/blob/b36cf3c9a91e4e270a190e23fe9f65fb8b593f4e/Cabal/Distribution/Simple/Utils.hs#L1133

mitchellwrosen:
Actually, I thing I'm wrong. The code looks like it would accept a filename such as
```
foo.*.txt
```
(but the user probably intended for `*` to be treated like a glob in this case).

mitchellwrosen:
Whoops, forget that last comment, I read the code wrong. I confused `filepath` for `filepath'`. The code is confusing, heh.

quasicomputational:
In which case I am going to turn this into a feature request, since I'd think that someone, somewhere will have a good reason to include a file with an asterisk in its path; it'll also free preprocessors like hpack of the duty to check for that special case.

PR #5284, opened 2018-04-25
Allow ** wildcards in globs.

### LINKED ISSUE #3178: Revive the globstar globbing patch
#2522 added support for globstar globbing (`**/*`) to `extra-source-files` and related fields; however the patch had some issues and wasn't merged. We should revive and finally merge it for the next Cabal release.

quasicomputational:
These are inspired by a plan described in a comment in #2522, and only
implement a quite limited form of recursive matching: only a single **
wildcard is accepted, it must be the final directory, and, if a **
wildcard is present, the file name must include a wildcard.

Or-patterns are not implemented, for simplicity.

Closes #3178, #2030.

Includes tests!


Cabal/doc/developing-packages.rst
```diff
@@ -981,17 +981,28 @@ describe the package as a whole:
...
-    The limitation is that ``*`` wildcards are only allowed in place of
-    the file name, not in the directory name or file extension. In
-    particular, wildcards do not include directories contents
-    recursively. Furthermore, if a wildcard is used it must be used with
-    an extension, so ``data-files: data/*`` is not allowed. When
-    matching a wildcard plus extension, a file's full extension must
-    match exactly, so ``*.gz`` matches ``foo.gz`` but not
-    ``foo.tar.gz``. A wildcard that does not match any files is an
-    error.
+    ``images`` directory. ``data-files: audio/**/*.mp3`` matches all
```
hvr: the documentation should mention that `**` is available starting with `cabal-version:3.0` (if we assume that 3.0 will be the next major release)

quasicomputational: Done. Also went ahead and changed all the '2.4' mentions to '3.0'.

BardurArantsson:
Looks like a very well crafted PR, kudos.

What's the reasoning for the "must be the final directory bit"?

(I'm stating all of the following as my personal opinion on the matter. I have no real decision-making power on this.)

Generally, I'd object to making cabal's glob behavior *even more* distinct from the usual shell glob behavior. Yes, even if it *is* well documented as in this case. Most people don't actually read documentation and instead just copy examples and extrapolate from what the already know -- and then either give up or file issues when things don't work as they expect. Now, that *is* on them, but I predict that non-shell-like behavior (which even mimics the syntax of the shell!) will lead to more "support burden" for the cabal development team when issues/questions arise.

I'd like to propose an alternative solution which I don't believe was discussed before: How about just using regex matches against the 'normalized' project-relative path? I.e. the whole string must match) against the path (using '/' as separator for cross-platformness).

23Skidoo:
Looks nice on first glance, do you plan to implement or-patterns for file extensions as well (`*.{html,jpg,css}`)? We might also want to change wildcards in 3.0 to make `*.js` catch `*.min.js` (but we must preserve the old semantics for `cabal-version < 3.0` files).

hvr:
@23Skidoo btw, you mentioning `{`/`}` reminds that we also need to provide a way to escape characters like `*` or `{`, `}`, or `,`; do we have support for that already?

quasicomputational:
> What's the reasoning for the "must be the final directory bit"?

It's forced by the data structure in the comment on #2522 by @23Skidoo, and it does slightly simplify implementation. This PR is targetting the absolute minimum possible version of this feature, to try and head off the bikeshedding that killed the previous one; after all, it's easier to lift restrictions in later versions than to add more. I'd rather see a limited patch merged and then iterated on rather than a more ambitious patch not get merged.

> or-patterns

Not in this PR? I think globstar is worth having on its own merits, and that or-patterns can be separately implemented. They'd need new syntax and stealing `{` is a different beast to repurposing `*`: `docs/**/*.html` is currently a glob syntax error and will make Cabal 2.2 and earlier error out, but `docs/{a,b,c}.html` is a cromulent pattern today. I also have an efficiency concern about or-patterns if they're going to lead us to always walk the whole directory tree from the project root, including `.git`, `dist-newstyle`, etc: this isn't unfixable but it's more complication. (Also I need to fix that in this PR; more pushes incoming.)

> We might also want to change wildcards in 3.0 to make *.js catch *.min.js (but we must preserve the old semantics for cabal-version < 3.0 files).

Agreed, but I think that's also separable work. A behaviour change to extension matching has different considerations (e.g., would `cabal check` need to walk the directory tree and look for things that might have been missed?). Globstar alone isn't actually a breaking change, again because `**` is already rejected, so I don't think it makes sense to try to cram in more features on the basis that one big breaking change is better than many small ones (which might be valid thinking if there was a breaking change here).

> escaping

Nope, no way today; I asked a few months ago in #4935. At least this PR doesn't make things worse, since `*` was already a magic character. We could go for backslash escaping and that also wouldn't be a breaking change, but, again, separable work.

23Skidoo:
Sure, implementing those additional features I mentioned is not required for this PR to be accepted.

quasicomputational:
Oops, this is buggy now. I know what I broke!

quasicomputational:
OK, I fixed the bug I introduced, and also another target of opportunity I spotted while I was staring at SrcDist.hs. If Travis and the reviewers are happy, I think this is good to go (but I thought that before, too).

quasicomputational:
Note that we definitely need the fix from #5289 (otherwise we will do the wrong thing with `data-dir: **`, which currently makes sdist fail). The included tests are the awkward part there.

quasicomputational:
Fingers crossed I've fixed the Windows problem. I squashed and rebased, and this ought to be ready for review (again) if Travis and AppVeyor like it.

quasicomputational:
I expanded the changelog note a bit, squashed and rebased again. CI ought to come back totally green since the persistent failures have been taken care of in master (hurrah!). If I don't get any objections, I'll merge this next week?

Cabal/doc/developing-packages.rst
```diff
@@ -981,23 +981,42 @@ describe the package as a whole:
...
+      allowed).
+
+    - A wildcard that does not match any files is an error.
 
     The reason for providing only a very limited form of wildcard is to
     concisely express the common case of a large number of related files
     of the same file type without making it too easy to accidentally
     include unwanted files.
 
+    On efficiency: the directory tree will be walked starting with the
```
23Skidoo: We might even want to add a warning about this.

Cabal/Distribution/Simple/Glob.hs
```diff
@@ -0,0 +1,177 @@
...
+fileGlobMatches :: GlobPat -> FilePath -> Bool
+fileGlobMatches pat = fileGlobMatchesSegments pat . splitDirectories
+
+fileGlobMatchesSegments :: GlobPat -> [FilePath] -> Bool
+fileGlobMatchesSegments _ [] = False
+fileGlobMatchesSegments pat (seg : segs) = case pat of
+  PatStem dir pat' ->
+    dir == seg && fileGlobMatchesSegments pat' segs
+  PatMatch Recursive ext ->
+    ext == takeExtensions (foldl' (flip const) seg segs)
```
23Skidoo: > `(foldl' (flip const) seg segs)`

IMO `last (seg : segs)` is more readable. Or we can vendor `lastDef` from the `safe` package.

23Skidoo:
Merged, thanks!

jtojnar:
What are good defaults for `Verbosity` and `Version` arguments if we want to replace the removed [`Distribution.Simple.Utils.matchFileGlob`](https://www.stackage.org/haddock/lts-10.9/Cabal-2.0.1.1/Distribution-Simple-Utils.html#v:matchFileGlob) with [`Distribution.Simple.Glob.matchDirFileGlob`](https://www.stackage.org/haddock/lts-13.9/Cabal-2.4.1.0/Distribution-Simple-Glob.html#v:matchDirFileGlob)?

Some packages depend on this:

* https://github.com/haskell-hvr/hgettext/issues/15
* https://gitlab.freedesktop.org/bustle/bustle/issues/13

quasicomputational:
The `Version` argument needs to come from the package description being processed, since the globbing behaviour is spec-version-dependent. The `proto-lens` diff you linked to is the right approach and hard-coding in a version as in haskell-hvr/hgettext#16 isn't. Looking at the code for that package, you'll have to do a little bit of plumbing through to get it to where it needs to be.

You can find ready-made `Verbosity` values in `Distribution.Verbosity`; `normal` is a reasonable default.

PR #5286, opened 2018-04-26
travis: Move the 8.4.1 builders to 8.4.2.

quasicomputational:
This ought to fix the build matrix. I believe the issue is GHC trac
5129, where evaluate is being miscompiled: in cabal-testsuite's
Test.Cabal.Server, in the function readUntilEnd, the 'read' function
is protected by an 'if' to ensure that it doesn't result in the
dreaded 'no parse'. However, some sophisticated putStrLn
debugging has revealed that it is being called in *both* branches, not
just the safe one; this is the worst sort of heisenbug because it
vanished when I put a putStrLn before the evaluate!

Since the evaluate function is fixed in 8.4.2 and bumping to 8.4.2
has a patch for trac 5129, this is as fixed as it can get.

Tested locally, found it failed on 8.4.1 and succeeded on 8.4.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,20 +55,20 @@ matrix:
    - env: GHCVER=8.2.2 SCRIPT=script USE_GOLD=YES
      os: linux
      sudo: required
-   - env: GHCVER=8.4.1 SCRIPT=script USE_GOLD=YES DEPLOY_DOCS=YES
+   - env: GHCVER=8.4.2 SCRIPT=script USE_GOLD=YES DEPLOY_DOCS=YES
      os: linux
      sudo: required
 
-   - env: GHCVER=8.4.1 SCRIPT=solver-debug-flags USE_GOLD=YES
+   - env: GHCVER=8.4.2 SCRIPT=solver-debug-flags USE_GOLD=YES
      sudo: required
      os: linux
-   - env: GHCVER=8.4.1 SCRIPT=script DEBUG_EXPENSIVE_ASSERTIONS=YES TAGSUFFIX="-fdebug-expensive-assertions" USE_GOLD=YES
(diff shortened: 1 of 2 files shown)


23Skidoo:
Feel free to merge once it's green.

23Skidoo:
The remaining error is because of `-Werror`.

quasicomputational:
Just testing locally to make sure pre-8.4 versions are happy with the imports now. It'll still be red because of T3436, which is being poked at in #5288. If that's the only issue I can see then I'll go ahead and merge, but it'll be a little while before Travis runs, I think.

23Skidoo:
You can try cherry-picking the patch from #5288 as well, then we can merge it all in one go.

23Skidoo:
Actually, it looks like that's how it was implemented initially (https://github.com/grayjay/cabal/commit/9ec402877af62e5657e6ec7c277950bd9ea0dff6), but then the test was changed for some reason.

quasicomputational:
And now a file I didn't even touch (and which doesn't import anything I did touch) has decided to start failing. I am quite confused about this one.

23Skidoo:
Sorry, that was my fault. Fixed on `master`.

quasicomputational:
Huh. I had no idea Travis would auto-rebase like that. No wonder I couldn't figure out what had changed in the PR to cause it!

quasicomputational:
Anyone got any idea what's up with the OSX downstream builder? Is that a known flaky test, or did something really subtle break?

23Skidoo:
> Is that a known flaky test, or did something really subtle break?

No, I don't remember it failing before, may be a new regression.

quasicomputational:
I re-triggered it and it went away, so I am going to blame some Travis weirdness and go ahead and merge. The notorious T3436 is still failing, but that's being worked on and this PR is making the situation strictly less broken.

PR #5288, opened 2018-04-26
Attempt at fixing T3436 regression test.

merijn:
The T3436 regression test is responsible for literally every PR being marked as failing, because it fails on GHC 8.4. I don't really understand the test, so someone has to check whether this defeats it's entire purpose (@grayjay ?).

The main problem is that GHC 8.4 isn't officially supported by older Cabal versions, so the test's use of "Cabal >= 2.0" leads it to pick the Cabal 2.2 that comes with GHC and everything breaks. The new version simply ups the Cabal version to 2.4 so that the test will properly compile with GHC 8.4 and hopefully fix everyone's failing tests. Rejoice!


23Skidoo:
Can we set the version to e.g. 3000, so that it wouldn't break with the next GHC release?

23Skidoo:
Right, because of #4909. Perhaps we should add a switch for turning that off and run the test suite with it enabled.

grayjay:
@merijn Is `PackageTests/Regression/T3436/cabal.test.hs` the only failing test?  I think that the main regression test for #3436 is the one that uses `cabal sandbox` (`cabal-testsuite/PackageTests/Regression/T3436/sandbox.test.hs`), because the test needs to install a package into the user or global package db, and that is easier to do with a test sandbox.  cabal.test.hs is a new-build version of the regression test, but there is a comment saying that it doesn't work:

https://github.com/haskell/cabal/blob/a3a027fce3dd1250c0e5c4c91a11247c7761acb4/cabal-testsuite/PackageTests/Regression/T3436/cabal.test.hs#L3-L7

It might be better to just delete cabal.test.hs, and we can migrate it again if necessary when we remove `cabal sandbox`.  It's possible that the regression test won't be relevant then, because new-build can completely ignore the installed packages when choosing the setup Cabal version, once it doesn't need to interact with the old build code path.

merijn:
I think it's only cabal.test.hs that's failing, yeah. I'm fine with just nuking that one if everyone's okay with it.

merijn:
Aside from the AppVeyor build (which seems to fail in a completely unrelated test?) this seems to turn everything green. Any objections to just nuking the new-build version of T3436, considering it wasn't really doing anything anyway?

merijn:
Actually, looks like it's just AppVeyor being difficult, on my branch it's marked as succeeding, so should be fine to merge if people agree with ditching the new-build test.

merijn:
Ping @ezyang 

I was told you knew the test-suite best, any comments?

merijn:
Any objections? Else I'm merging this.

PR #5289, opened 2018-04-26
sdist: don't fall over when data-dir has * in it.

quasicomputational:
Also includes some tests for sdist.

Spun off from #5284 because it was breaking AppVeyor. Dunno if it's fixable without dropping the test.


23Skidoo:
Looks like this breaks `PackageTests/SDist/ListSources/list-sources`.

quasicomputational:
Yeah, that's a new test that I'm adding; it's evidently not right yet. It works on my machine but I didn't test with old GHCs before pushing, so I'll need to figure out what's making those fail on Travis, but I don't think those are fundamental problems.

One actual problem is determinism: I want to stuff the `--list-sources` output into a temporary file somewhere, but the functions from the temporary package pick random names and those are leaking into cabal's output. What's the solution for this sort of thing that doesn't lead to `git status` winding up full of untracked files?

23Skidoo:
Or you can add a `--print-sources` option to `sdist`.

quasicomputational:
Will do. Probably the latter since it's a useful thing to have anyway.

RE the AppVeyor failure: looks like it's legit and fundamental. Paths on Windows apparently [cannot have asterisks in them](https://msdn.microsoft.com/en-us/library/aa365247(VS.85\).aspx). This makes me think that Cabal ought to disallow asterisks in paths for portability's sake. Currently having them in `data-dir` is semi-broken: sdist will fail, but other commands (e.g., `register`) will succeed. I'll see if something sane can be done here...

23Skidoo:
> This makes me think that Cabal ought to disallow asterisks in paths for portability's sake.

Makes sense.

quasicomputational:
Spun this out into #5295 for some tests for sdist and #5294 for further thinking on portability-guided restrictions on paths. I'll implement the `data-dir` fix for #5284 directly in that branch; I've given up on having a reasonable test for the case where it would have gone wrong, but it's a tiny, obviously-correct change.

hvr:
> This makes me think that Cabal ought to disallow asterisks in paths for portability's sake. 

I strongly disagree. This is a distribution concern, as such it's rather Hackage that should disallow it. But I certainly want to be able to use `*` in my paths, since they're perfectly valid characters. Add a `cabal check` warning if you like, but certainly don't disallow them!

Issue #5292, opened 2018-04-27
Possible incorrect handling of whitespace in .cabal file parser


piyush-kurur:
In cabal file's `mixin` stanza the `requires` clause has a problem with white spaces.
In particular the code

```
mixin: foo requires ( A as B) -- This fails (note the space after the parenthesis)
mixin: foo requires (A as B) -- This works.
```

Here is an example of a package that fails <https://github.com/piyush-kurur/trying-backpack/blob/master/whitespace/whitespace.cabal#L40>

phadej:
That's known bug. Try with `cabal-version: 2.2`

hvr:
We still haven't documented the `mixin` field in the user's guide, have we?

piyush-kurur:
I think I managed to fix this particular bug. Shall I send a pull request?

phadej:
Ah, seems I haven't fixed that because backward compatibility:

https://github.com/haskell/cabal/issues/5150

phadej:
I.e. the fix isn't *simply* adding `spaces` in right place, if we don't want to break `cabal-install-2.0` and `cabal-install-2.2` users.

phadej:
@piyush-kurur you can check how `VersionRange` is cased by cabal version. It's not tricky, just something to be aware of. (And if one changes the parser, it's good idea to write a lot of test for *should be accepted and rejected* cases, so we don't end up with layers of versions :))

piyush-kurur:
@phadej I guess what you mean is that the parser skipping spaces should depend on the version declared  in the cabal file ? I just pushed the parser changes to my repository https://github.com/piyush-kurur/cabal/commit/cb0d813277b5fd1b509425beef21169fd470f217. As far as I see these are instace declarations of  certain types. I do not think I have access to the version ranges here.

phadej:
Resolved by https://github.com/haskell/cabal/pull/5514

PR #5293, opened 2018-04-27
more liberal white space parsing in the the mixin field

piyush-kurur:
This is not yet complete. It is sent for review.

I will add the test cases when I get a go ahead on the commits so far.

### DIFF
--- a/Cabal/Distribution/Types/ModuleRenaming.hs
+++ b/Cabal/Distribution/Types/ModuleRenaming.hs
@@ -8,6 +8,7 @@ module Distribution.Types.ModuleRenaming (
     isDefaultRenaming,
 ) where
 
+import Distribution.CabalSpecVersion
 import Distribution.Compat.Prelude hiding (empty)
 import Prelude ()
 
@@ -67,6 +68,19 @@ isDefaultRenaming :: ModuleRenaming -> Bool
 isDefaultRenaming DefaultRenaming = True
 isDefaultRenaming _ = False
 
+-- | For versions < 3.0 white spaces were not skipped after the '('
+-- and ')' tokens in the mixin field. This parser checks the cabal file version
+-- and does the correct skipping of spaces.
+betweenParens :: CabalParsing m => m a -> m a
+betweenParens p = do
+  csv <- askCabalSpecVersion
+  if csv >= CabalSpecV3_0
+    then P.between (P.char '(' >> P.spaces) (P.char ')' >> P.spaces) p
+    else P.between (P.char '(') (P.char ')') p
+
+-- TODO: Give a sensible error on input "( A as B) instead of (A as B)
+-- for cabal files that are declared to be older than 3.0.
+
 instance Binary ModuleRenaming where
 
 instance NFData ModuleRenaming where rnf = genericRnf
@@ -87,18 +101,18 @@ instance Parsec ModuleRenaming where
     -- NB: try not necessary as the first token is obvious
     parsec = P.choice [ parseRename, parseHiding, return DefaultRenaming ]
       where
+        cma = P.char ',' >> P.spaces
         parseRename = do
-            rns <- P.between (P.char '(') (P.char ')') parseList
+            rns <- betweenParens parseList
             P.spaces
             return (ModuleRenaming rns)
         parseHiding = do
             _ <- P.string "hiding"
             P.spaces
-            hides <- P.between (P.char '(') (P.char ')')
-                        (P.sepBy parsec (P.char ',' >> P.spaces))
+            hides <- betweenParens (P.sepBy parsec cma)
             return (HidingRenaming hides)
         parseList =
-            P.sepBy parseEntry (P.char ',' >> P.spaces)
+            P.sepBy parseEntry cma
         parseEntry = do
             orig <- parsec
             P.spaces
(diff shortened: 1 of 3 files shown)


piyush-kurur:
The travis builds are failing left right and center. However, I do not think it is this commit that is responsible for it. Noticed that there are some further commits on the master branch which is supposed to fix some of the travis builds.  Please feel free to give feedback on what is the best course of action.

piyush-kurur:
The missing piece is to make cabal-install executable realise that it can now handle CabalSpec = 3.0. After chat with @phadej  on IRC, I came to understand that this version is taken from the version of Cabal package. However, a mere bumping up of Cabal version and cabal-install versions would not work as there is the package hackage-security which I guess has an upperbound on Cabal version. That too is needed to be updated.

23Skidoo:
Hmm, `master` itself looks fine on Travis, except for the T3436/8.4.2 issue. Try rebasing?

ezyang:
@piyush-kurur, thanks a lot for taking this on. I also had the "easy" version of this patch languishing on my local copy, but it's a bit more work getting all of the CabalSpec in order. Very happy to see it being done.

Cabal/Distribution/Types/ModuleRenaming.hs
```diff
@@ -115,18 +135,21 @@ instance Text ModuleRenaming where
   parse = do fmap ModuleRenaming parseRns
              <++ parseHidingRenaming
              <++ return DefaultRenaming
-    where parseRns = do
-             rns <- Parse.between (Parse.char '(') (Parse.char ')') parseList
+    where lparen = Parse.char '(' >> Parse.skipSpaces
```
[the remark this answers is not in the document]
piyush-kurur: Okey. I will remove it in further commits.

piyush-kurur:
@23Skidoo Do you recommend rebasing? I have found github pull requests that are rebased to be quite confusing.

Cabal/Distribution/Types/ModuleRenaming.hs
```diff
@@ -87,18 +104,21 @@ instance Parsec ModuleRenaming where
     -- NB: try not necessary as the first token is obvious
     parsec = P.choice [ parseRename, parseHiding, return DefaultRenaming ]
       where
+        lparen = P.char '(' >> liberalSpaces
+        rparen = P.char ')' >> liberalSpaces
```
piyush-kurur: @phadej The liberalSpaces parser just checks for appropriate cabal version and then skips spaces. For lower versions of the cabal spec it says gives an error if there is space. So far so good but I noticed that the spaces stuff was not needed in the version 2.2 case for rparen as there is a P.spaces after its call. Further down the code in the hiding clause there is no P.spaces used which means that the rparen there does not skip spaces (in the version 2.2 behaviour) 

If we want to maintain complete bug compatibility, then this will lead to a rather convoluted rparen behaviour. I feel this will eventually lead to more confusing whitespace behaviour in the end and the code will be pepperd with
various flavors of paren-parsers (some that skips spaces and some that do not).

piyush-kurur: The question therefore is shall I replace the rparen definition with

```
rparen = P.char ')' >> P.spaces
```
This is not really compatible with V2.2 I think due to the use of rparen in the hiding clause.

ezyang:
@piyush-kurur If you haven't yet decided on a way to structure the old and new parsers, I'd recommend just doing a straight up copy paste of the entire grammar and having two copies.

piyush-kurur:
@ezyang What you say makes sense. I will make the necessary changes and get back.

piyush-kurur:
@ezyang I have kind of done what you have asked me. There are two questions that I want to make

1. Is it okey for me to merge the last three commits (ebb6a67, 9df8823, and 95fd2a9) into one commit via a rebase -i. This might make the commit more readable and it would be clear what changed ?

2. There is one problem still which I have added as a TODO in the file. If a user declares cabal-version:2.2 and uses a more liberal spaces, then this version will just throw an error like the old version. However, we might want to give a helpful suggestion here. I cannot see a way to have this incorporated cleanly.

ezyang:
LGTM, though @phadej should probably sign off on the BC bits (I don't quite understand the correctness criterion here.) Happy to merge it in a few days if radio silence.

ezyang:
(and yes, I would squash)

piyush-kurur:
@ezyang Okey I will do the squash. 

Also @phadej please hold on merging this change as there is one piece still missing. The cabal-install executable should be made to realise that it can now handle V3 specs. As I remarked in the previous comment, this requires changing the version in the .cabal files of Cabal, cabal-install and I think also hackage-security. Is that all that is required ?

piyush-kurur:
@ezyang done the squash

hvr:
> As I remarked in the previous comment, this requires changing the version in the .cabal files of Cabal

not necessarily; we just need to infer (or hard-code) a "rounded up"  value as the current spec version; something we should do anyway as it would simplify our workflow.

piyush-kurur:
@hvr. Do let me know then how to proceed from here. Is there any plans to sync Cabal (the file format) specs with Cabal (the library version) ?

piyush-kurur:
Also test case. I have a failing case that I have mentioned in #5292. Let me know how this can be added to the testsuite.

23Skidoo:
@phadej, can you please take a look at this?

piyush-kurur:
any updates on this?

23Skidoo:
@phadej, can you please take a look?

quasicomputational:
There's a specific mention of this in the docs that will need to be removed once this is merged.

phadej:
merged as https://github.com/haskell/cabal/pull/5514

PR #5297, opened 2018-04-28
Avoid leaving leftover files in the source tree in tests.

quasicomputational:
This test was leaving a 'Main' executable behind. Since we don't ever
actually run it, just check that the compilation succeeds, we can send
it straight to /dev/null.


quasicomputational:
OK, so how do I determine if I'm on Windows or not? Googling suggested that testing `System.Info.os` for equality against `"mingw"` would work, but evidently not.

grayjay:
The test Prelude has an `isWindows` function: https://github.com/haskell/cabal/blob/a3a027fce3dd1250c0e5c4c91a11247c7761acb4/cabal-testsuite/Test/Cabal/Prelude.hs#L745

quasicomputational:
Thanks! You're a hero.

PR #5304, opened 2018-05-06
Solver: Enforce dependencies on libraries (fixes #779).

### LINKED ISSUE #779: build-depends on packages that provide no library
(Imported from [Trac #789](http://hackage.haskell.org/trac/hackage/ticket/789), reported by @dcoutts on 2011-01-13)

It is possible to list build-depends on packages that have just an executable and no library. Currently the behaviour of cabal-install does not consider this and it will allow such packages to be pulled in as dependencies. It leads to problems however because the non-lib package never gets registered with ghc-pkg and so it leads to re-install loops.

The solution is for cabal-install to not consider exe-only packages to be valid for satisfying a build-depends requirement. It should produce a suitable error message.

Note that it must take into account the posibility that a newer or older version of a package may switch from exe only to exe+lib or just lib, or indeed vica versa.

We should mark exe-only versions of the package as excluded with a reason. Then if the dep is not satisfiable then we can explain why all the excluded ones were excluded (ie due to not exposing a lib).

grayjay:

___________________________________________________________________________

The first two commits only modify the tests:

#### Solver DSL: Support benchmarks, internal libraries, and foreign libraries.

This change allows the solver quickcheck tests to test all types of components.
Previously, the tests generated the components in the solver DSL, but the code
that converted the packages to GenericPackageDescriptions removed some of the
components, i.e., it filtered out benchmarks and foreign libraries and merged
internal libraries with the main library.

#### Solver DSL: Support packages without libraries.

The solver DSL previously always added a default library component.

#### Solver: Enforce dependencies on libraries (fixes #779).

This commit generalizes the fix for issue #4781
(e86f838) by tracking dependencies on
components instead of dependencies on executables.  That means that the solver
always checks whether a package contains a library before using it to satisfy a
build-depends dependency.  If a version of a package doesn't contain a library,
the solver can try other versions.  Associating each dependency with a component
also moves towards the design for component-based dependency solving described
in issue #4087.

/cc @kosmikus

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -72,7 +72,7 @@ extendOpen qpn' gs s@(BS { rdeps = gs', open = o' }) = go gs' o' gs
       -- the later addition will have better dependency information.
     go g o ((Stanza sn@(SN qpn _) t)           : ngs) =
         go g (StanzaGoal sn t (flagGR qpn) : o) ngs
-    go g o ((Simple (LDep dr (Dep _ qpn _)) c) : ngs)
+    go g o ((Simple (LDep dr (Dep (PkgComponent qpn _) _)) c) : ngs)
       | qpn == qpn'       =
             -- We currently only add a self-dependency to the graph if it is
             -- between a package and its setup script. The edge creates a cycle
--- a/cabal-install/Distribution/Solver/Modular/Dependency.hs
+++ b/cabal-install/Distribution/Solver/Modular/Dependency.hs
@@ -16,6 +16,8 @@ module Distribution.Solver.Modular.Dependency (
   , FlaggedDep(..)
   , LDep(..)
   , Dep(..)
+  , PkgComponent(..)
+  , ExposedComponent(..)
   , DependencyReason(..)
   , showDependencyReason
   , flattenFlaggedDeps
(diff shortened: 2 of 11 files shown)


23Skidoo:
Thanks for doing this! @fgaz, I think this is relevant to your GSoC project.

23Skidoo:
Looks good from what I can tell, please feel free to merge unless @kosmikus objects.

kosmikus:
Superficially looks good to me as well, but I haven't looked at the changes in very much detail. Thanks.

grayjay:
Thanks!  I updated the changelog in #5316.

hvr:
@grayjay is this expected to fix issues like with the following situation where `build-depends` refers to a package whose library component is unavailable, e.g.:

```
cabal-version:2.2
name: zzz
version: 0

library
  build-depends: http-client == 0.5.12.1
```


```
$ cabal new-build -w ghc-7.8.4 --index-state='2018-05-09T06:20:23Z'
Error:
    Dependency on unbuildable library from http-client
    In the stanza 'library'
    In the inplace package 'zzz-0'
```

grayjay:
No, this change only checks whether the library stanza exists.  Checking that the library is buildable will probably require implementation of the "Component on/off" variable from https://github.com/haskell/cabal/issues/4087#issuecomment-287665861.

hvr:
@23Skidoo I had to cherry-pick this one to 2.2 as well as #5337 needed this to cherry-pick cleanly to 2.2

PR #5305, opened 2018-05-06
Print all test output when running solver-quickcheck on AppVeyor.

grayjay:
Previously, --hide-successes caused solver-quickcheck to only print the seed
that was added in 4d83212fbec3f990f3d90801077da05e5133bd99.  This commit removes
--hide-successes so that the test suite prints the test names with the seeds.


23Skidoo:
Merged, thanks!

PR #5310, opened 2018-05-09
globbing: improve test coverage.

quasicomputational:
For efficiency reasons, matchDirFileGlob isn't a simple call to
getDirectoryContentsRecursive and then a filter with
fileGlobMatches. So test both that naive approach and the actual
approach to make sure they are both correct.

I only realised this was a problem right after #5284 was merged!

Since this is tests-only, I'll merge once Travis and AppVeyor are green.


23Skidoo:
Looks good, thanks!

PR #5320, opened 2018-05-13
Be more efficient about globbing

quasicomputational:
Before, we were walking the directory tree recursively
unconditionally. Now we check if the pattern is recursive before doing
that, and also only check the directory's full listing if a glob
pattern is involved at all, so literal patterns are more efficient.


quasicomputational:
I added another change here, to exclude files with leaving dots, which is a restoration of the pre-globstar behaviour except now it's tested and documented.

quasicomputational:
Also Travis is doubtless going to come back red because of the SSH key thing :(

quasicomputational:
I'll merge this in a few days if no-one objects.

PR #5336, opened 2018-05-20
Rotate haskell-pushbot keys

ezyang:


phadej:
Why can't we use https://docs.travis-ci.com/user/encryption-keys/ and not play games?

phadej:
FWIW, my `id_ed25519` is 444 bytes long. That one can live in env var directly.

phadej:
Ah I see why we cannot use secrets. Is there a comment about that somewhere in the code?

23Skidoo:
Looks like there's a bug in the bit where we inoke `rot13`:

```
+rot13
./travis/upload.sh: line 42: id_rsa: No such file or directory
```

travis/upload.sh
```diff
@@ -39,7 +39,7 @@ cd travis
 # Setup SSH key we will use to push to binaries repository.
 # umask to get the permissions to be 600 (not 400, because the deploy
 # script in .travis.yml is going to clobber this private key)
-(umask 177 && cp id_rsa $HOME/.ssh/id_rsa)
+(umask 177 && rot13 < id_rsa > $HOME/.ssh/id_rsa)
```
23Skidoo: Shouldn't the input file be `id_rsa.rot13`?

travis/upload.sh
```diff
@@ -65,7 +65,7 @@ mkdir cabal-install
 cp -R $TRAVIS_BUILD_DIR/Cabal/tests                                  Cabal
 cp -R $TRAVIS_BUILD_DIR/cabal-install/tests                          cabal-install
 # Copy in credentials so we can delete branch when done
-cp $TRAVIS_BUILD_DIR/travis/id_rsa .
+rot13 < $TRAVIS_BUILD_DIR/travis/id_rsa > id_rsa.rot13
```
23Skidoo: Here as well.

23Skidoo:
Pushed a fix, let's see if it unbreaks things.

ezyang:
I am a bit perplexed by the JMU errors. What's going on here?

23Skidoo:
Looks like the usual intermittent Travis flakiness, I've restarted the build.

quasicomputational:
I've tried to fix it. Let's see what Travis thinks...

23Skidoo:
Uhh, maybe that's because we're pushing the un-ROT13ed key to a branch in the `cabal-binaries` repo, and GitHub is smart enough to notice that.

quasicomputational:
That sounds very plausible. Does the cabal-binaries repo actually need the key? If it does, we can just avoid putting it in the clear (since we deliberately do that now), and then whatever it is that needs the key will need to learn to 'decrypt' it. Otherwise, we can just delete that line and go on happily.

Either way I think @ezyang needs to generate a fresh key and install it on the haskell-pushbot account, right?

ezyang:
LOL. I thought I was pushing the rot13'ed key; apparently not. We do need to push the key with the commit because the downstream jobs are responsible for deleting their binaries when they fail. Issuing new key...

23Skidoo:
Yes, I think we can push a ROT13-ed key to cabal-binaries and then decrypt it during the build step.

ezyang:
Looking good.

23Skidoo:
The 7.10.3 failure is an unrelated issue, otherwise everything is green.

23Skidoo:
Merged, thanks!

Issue #5325, opened 2018-05-15
Only choose buildable components for build-depends and build-tool-depends dependencies.


sebright:
See https://github.com/haskell/cabal/pull/5304#issuecomment-388626764 and https://github.com/haskell/cabal/issues/3978#issuecomment-388504699 for an example involving `http-client-0.5.12.1`.  The library stanza contains

```
  -- See build failure at https://travis-ci.org/snoyberg/http-client/jobs/359573631
  if impl(ghc < 7.10)
    buildable: False
```

The solver doesn't backtrack to find an older version with a buildable library, and cabal fails with (from https://github.com/haskell/cabal/pull/5304#issuecomment-388626764):

```
$ cabal new-build -w ghc-7.8.4 --index-state='2018-05-09T06:20:23Z'
Error:
    Dependency on unbuildable library from http-client
    In the stanza 'library'
    In the inplace package 'zzz-0'
```

This issue is a continuation of #3978, which described a bad error message for an unbuildable component.  Its regression test tests the current error message:

https://github.com/haskell/cabal/blob/898fcc230fb537064f0992cfffb5955b6bc36d34/cabal-testsuite/PackageTests/NewBuild/T3978/cabal.out#L1-L6

This issue should probably be fixed as part of #4087.

hvr:
@grayjay I'm promoting this issue to high prio as this is effectively causing a lot of install-plans to regress on Hackage currently, and adding a `build-depends: base>=4.8` constraint (as @phadej did) to help the solver avoid running into the `impl(ghc >= 7.10) buildable:False` case unfortunately doesn't work because of a property that @ezyang already diagnosed 2 years ago in the referenced #3978:

> when `buildable: False` we ignore the `build-depends`. This is desirable for executables but not for libraries.

Would something like https://github.com/ezyang/cabal/commit/272b50433aeed6eb369ca80b05ad17b5a8fcb49f improve the situation for this `http-client` incident?

phadej:
I made a PR #5332, and it seems to solve my `http-client` related issue.

phadej:
That works even too well: https://github.com/haskell-servant/servant/blob/f88cfd0740be8b2e3085584ee59f20e034168e58/doc/cookbook/jwt-and-basic-auth/jwt-and-basic-auth.cabal#L17
errors with

```
cabal: Could not resolve dependencies:
[__0] next goal: cookbook-jwt-and-basic-auth (user goal)
[...]

PR #5337, opened 2018-05-21
Solver: Check whether components are buildable in the current environment.

grayjay:
This commit handles the most common case of issue #5325 by checking that each
component that is required as a dependency is buildable in the current
environment, where environment refers to the compiler, os, arch, and global flag
constraints.  The solver records whether each component is buildable in the
package's PInfo during index conversion.  Then it checks that each required
component is buildable in the validation phase, similar to the check for missing
components.

The buildable check can give false-positives, because it only considers flags
that are set by unqualified flag constraints, and it doesn't check whether the
intra-package dependencies of a component are buildable.  The check is also
incomplete because it is performed before any automatic flags are assigned.  It
is possible for the solver to later choose a value for a flag that makes the
package unbuildable.

/cc @hvr @phadej


phadej:
I used this patch in `servant`'s CI, and it fixes the issue there. So :+1: from me, solves my problem :)

phadej:
I have dogfooded this for over a week, so I guess @hvr.

Herbert: what's your experiences, let's merge this!

hvr:
@grayjay hrm... so `yesod-0.0.0.2` pkg spec appears ill-specified anyway, as toggling the flag results in the lib-component's `build-depends` being disabled as well, and since this is a cabal-version:1.6 description, it means that the exe component becomes effectively unusable as well as it e.g. loses its implicit `build-depends` on `base` as well.

grayjay:
The build-depends field isn't under the conditional, so it might work with older versions of cabal that required dependencies for unbuildable components.

PR #5350, opened 2018-05-28
GHCJS on GHC 8.2+ supports -Wmissing-home-modules

luite:

If the underlying GHC version is 8.2 or higher, GHCJS supports `-Wmissing-home-modules`. Cabal not knowing this results in some incorrect warnings: ghcjs/ghcjs#658


23Skidoo:
Merged, thanks!

PR #5358, opened 2018-06-02
[WIP] Musical chairs to switch `new-` and `old-`

typedrat:
This is the start of work to make the current `new-` commands the standard versions and to make the current commands have the `old-` prefix. Of course, to have this, we have to decide what we are willing to give up commands-wise and what we want to wait on being ready. `clean` and `sdist` are the two most important I think, of the ones that are still out.


quasicomputational:
Is it worth splitting into two commits, do you think, with one creating the `old-*` aliases, and another dropping the `new-` prefix? We probably want to apply the first of those anyway, regardless of when the second is ready to happen, and it makes falling back to the old codepaths less painful if you're testing on multiple Cabal versions than having to check `cabal --version`.

typedrat:
That’s a good idea.

phadej:
As there are people using cabal-install-head in production like environments, i'd ask to do this in steps with a clear schedule:

- duplicate build,haddock,... commands with old-prefix(can be done immediately)
- change prefixless versions to new-variants
(must be communicated in advance)

I guess that's what @quasicomputational talks about.

Bikeshedding: i'd use `one-` prefix, or something else than `old` (absolute vs relative: store birthday, not age).

I'm also realistic: sandboxes are still useful, before there is per-component solving (e.g. benchmarks of unordered-containers, which loop through criterion-aeson).

typedrat:
Okay, that's a fair point. I went with old because I couldn't think of another prefix that couldn't almost be taken as a positive (if not now, once the new prefix goes away). `one` almost seems like it's a unified feature, not a semi-deprecated one.

I can kill this and replace it with just adding old prefixes, since yeah we do need to accommodate people using HEAD in production. Not sure how long to wait for the switch-over, though.

BardurArantsson:
> As there are people using cabal-install-head in production like environments, 

Anything that happens as a consequence of that is *on them*.

If they're not using fixed snapshots then they're *begging* for instability *and* for their infrastructure and/or builds to be "owned". A *lot* of people have the ability to push directly to Cabal master at this point without any oversight except someone noticing after the fact. (I'm sure most of them are very trustworthy, but if you're running directly from HEAD in production -- rather than stable hand-picked snapshots -- you need *ALL* of them to be 100% trustworthy and 100% conscientious about how they handle their GitHub SSH keys, etc. etc.)

That said, I think I agree that's its worth starting by first introducing old-* aliases so that people who are worried about new-* can start transitioning to those *before* The World Changes. (However, that's only the case if the intent is to keep those old-* aliases around for at least one release... which I presume it is?)

typedrat:
My own personal inclination would be to make it *exactly* one release. Not taking v1-/old- away that soon but at least making it not the default soon.

typedrat:
Unless anyone has any concerns or desired changes (and assuming that my test fix worked), I think this is ready to be merged.

typedrat:
The Appveyor failure seems to be unrelated (unless adding the aliases somehow is making it not get pkg-config)

cabal-install/main/Main.hs
```diff
@@ -275,44 +276,33 @@ mainWorker args = topHandler $
...
-      , regularCmd testCommand testAction
-      , regularCmd benchmarkCommand benchmarkAction
       , hiddenCmd  uninstallCommand uninstallAction
       , hiddenCmd  formatCommand formatAction
       , hiddenCmd  upgradeCommand upgradeAction
       , hiddenCmd  win32SelfUpgradeCommand win32SelfUpgradeAction
       , hiddenCmd  actAsSetupCommand actAsSetupAction
       , hiddenCmd  manpageCommand (manpageAction commandSpecs)
-
+
```
quasicomputational: Trailing whitespace.

cabal-install/Distribution/Client/CmdLegacy.hs
```diff
@@ -0,0 +1,67 @@
...
+
+import qualified Data.Text as T
+
+-- Duplicated code, but makes things cleaner and lets me keep how this happens a dirty little secret.
+makeCmd :: CommandUI flags -> (flags -> [String] -> action) -> CommandSpec action
+makeCmd ui action = CommandSpec ui ((flip commandAddAction) action) NormalCommand
+
+deprecationNote :: String
+deprecationNote = 
+    "This command is a part of the legacy v1 style of cabal usage.\n\n" ++
```
quasicomputational: Rather than saying "This command", better to name the command being used, to save digging in complicated scripts.

cabal-install/Distribution/Client/CmdLegacy.hs
```diff
@@ -0,0 +1,67 @@
...
+                Just notes -> notes pname ++ "\n" ++ deprecationNote 
+                Nothing -> deprecationNote
+            }
+        
+        aliasedUi = origUi
+            { commandName = updateMsg commandName
+            , commandUsage = updateMsg . commandUsage
+            , commandDescription = (updateMsg .) <$> commandDescription
+            , commandNotes = Just $ \pname -> case commandNotes of
+                Just notes -> updateMsg (notes pname) ++ "\n" ++ deprecationNote
```
quasicomputational: This is achieving too much:

```
cabal new-run cabal -- install --help

...

Examples:
  cabal v1-install                     Package in the current directory
  cabal v1-install foo                 Package from the hackage server
  cabal v1-install foo-1.0             Specific version of a package
  cabal v1-install 'foo < 2'           Constrained package version
  cabal v1-install v1-haddock --bindir=$HOME/hask-bin/ --datadir=$HOME/hask-data/
                                    Change v1-installation destination
```

I think you're going to have to put the `v1` in each legacy command's `commandNotes`, and then strip it out in the `legacyUI` (rather than adding it in the `aliasedUI` and keeping the `commandNotes` as they are).

23Skidoo:
AppVeyor seems to be flaky lately.

typedrat:
If these lights turn green (or at least, don't turn red in a way that looks like my fault) I'll squash and merge this tonight.

typedrat:
I don't understand what's causing the failures.

It's like it's running against an old revision of the tests, because the expected output files match what it's claiming is wrong.

ezyang:
@typedrat, I'm not sure how you determined the expected output files, but the `cabal run` commands in this test are specified with `-v0`, so you shouldn't be seeing any of the warning output in the expected output transcript. Can you maybe run this test with verbose output on Linux, and paste it here?

typedrat:
I looked at the diffs the test suite prints when there's a mismatch and applied them by hand, just about.

-v0 doesn't turn off the messages right now, since verbosity is for some reason not a global flag so there's no way to uniformly get it as it stands.

ezyang:
My guess is that the verbosity flag is being non-uniformly applied depending on if you're in Linux or Windows. I guess, if you are not in the mood for sussing out what platform specific difference is causing the problem, you can suppress it by removing `recordMode RecordAll` and instead grepping the stdout for the expected text.

quasicomputational:
Over-replacement is still happening:

```
$ cabal new-run cabal -- v1-build --help
Up to date
Compile all/specific components.

Usage: cabal v1-build [FLAGS]
   or: cabal v1-build COMPONENTS [FLAGS]

Components encompass v1-executables, v1-tests, and v1-benchmarks.
```

```
$ cabal new-run cabal -- v1-test --help
Up to date
Run all/specific tests in the test suite.

Usage: cabal v1-test [FLAGS]
   or: cabal v1-test TESTCOMPONENTS [FLAGS]

If necessary (re)v1-configures with `--enable-v1-tests` flag and v1-builds the v1-test
suite.

Remember that the v1-tests' dependencies must be v1-installed if there are
additional ones; e.g. with `cabal v1-install --only-dependencies --enable-v1-tests`.

By defining UserHooks in a custom Setup.hs, the package can define actions to
be v1-executed before and after v1-running v1-tests.
```

23Skidoo:
> ```
> If necessary (re)v1-configures with `--enable-v1-tests` flag and v1-builds the v1-test
> suite.
> ```

That's a funny one.

typedrat:
Huh, I'm not sure how to proceed. 

I guess if I gotta just bite the bullet and do it all by hand, I can. That way, I can have it _strip_ the `v1-`s, and that's a lot more fool-proof.

quasicomputational:
I think putting `v1-` in the canonical description and stripping it out has to be the way to go; the other approach sounds too complicated and will probably still fail in weird ways unless you do something like mark up the description with 'this is a command name' and 'this isn't' information, and that's not less work than the first option.

typedrat:
I don't know how the message has started showing up on Setup programs, too. Shouldn't that be separate?

quasicomputational:
Another (minor) UI problem: the same warning is printed for both the `v1-` and unprefixed versions:

```
The (v1-)build command is a part of the legacy v1 style of cabal usage.

Please switch to using either the new project style or the legacy v1- aliases,
as new-style projects will become the default in the next version of
cabal-install. Please file a bug if you cannot replicate a working v1- use
case with the new-style commands.
```

But, if the user's invoked `v1-build`, we don't need to tell them to switch to explicit prefices, and we can probably reduce it down to just a nag that it'll be removed eventually and they should move over to new-style commands. The full warning should stay for `build`, of course.

CI failures look legit, but it looks like a technicality and not something fundamental, hopefully?

These issues aside I think that this looks good and, with a little more work, should be ready to merge soon.

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -657,12 +696,26 @@ buildExOptions _showOrParseArgs =
 
 buildCommand :: CommandUI (BuildFlags, BuildExFlags)
 buildCommand = parent {
+    commandName = "build",
+    commandDescription  = Just $ \_ -> wrapText $
+      "Components encompass executables, tests, and benchmarks.\n"
+        ++ "\n"
+        ++ "Affected by configuration options, see `v1-configure`.\n",
```
quasicomputational: I think the convention is to use straight quotes? This occurs a bunch so you'll have to search-and-replace.

typedrat:
Fixed the CI failures. Forgot about one of the places that things get special cased based on what command is used.

cabal-install/Distribution/Client/Setup.hs
typedrat: That's just copy-pasted from `Distribution.Simple.Setup` so if it's inconsistent that's not on me. 😛

quasicomputational: Oh, fair point. I do see we're inconsistent today. I've opened #5377 as a TODO item about it.

typedrat:
Appveyor has the demons again and Travis ran out of memory once, but otherwise it looks good.

lspitzner:
Does this not unnecessarily break any existing scripts that expect (old) semantics, if "cabal" switches? How about creating a new executable with a somewhat close name (canoe?) that exposes just the new-style semantics, without the prefix? Or is this the plan all along? (What is the plan? I see no issues or other motivation linked for this PR..)

I get that a new executable name creates friction, but the alternative seems to force users and scripts to check "cabal --version" way too much in the foreseeable future, which certainly is friction too, and much worse.

23Skidoo:
I don't think that we should have eternal backwards compatibility in the command-line UI. We want everyone to migrate to `new-build` after all.

> I get that a new executable name creates friction, but the alternative seems to force users and scripts to check "cabal --version" way too much in the foreseeable future, which certainly is friction too, and much worse.

The alternative for users who want legacy semantics is to either use an old version of `cabal-install` or replace `cabal CMD` with `cabal v1-CMD` in their scripts. There will be time (at least one release cycle) for everyone to update.

gbaz:
Just for reference, I know we have had a number of prior discussions, though I don't remember where they are all located, but the idea that `new-build` will eventually become `build` goes back quite some time (to the beginning of this project even). Here is an announcement making reference to it in 2016: http://blog.ezyang.com/2016/05/announcing-cabal-new-build-nix-style-local-builds/

This work is being done as part of a gsoc project for the purpose of making that switch possible, as described on the ideas page here: https://summer.haskell.org/ideas.html#cabal-new-build and in the accepted proposals page here: https://summer.haskell.org/news/2018-04-23-accepted-projects.html#helping-cabal-new-build-become-just-cabal-build

lspitzner:
> I don't think that we should have eternal backwards compatibility in the command-line UI.

The old executable can easily be phased out after a while. I don't see what would be particularly eternal about this.

> We want everyone to migrate to new-build after all.

And if the old executable is phased out after a while, this forces them to switch, only without also breaking scripts and UI expectation in a surprising fashion, and creating confusion during a lengthy transition period.

If my script stops working because there is no "cabal" any longer, I can easily tell what is up. If it breaks in some fashion because commands switched semantics this is much more confusing.

The same for any haskell tutorials, project installation instructions, and other sorts of user-facing docs.

> The alternative for users [..] is to either use an old version of cabal-install or replace cabal CMD with cabal v1-CMD in their scripts

This simply does not address the question of which approach creates less friction.

I am sure I miss some of the disadvantage that come with a new executable name. Feel free to address those.

lspitzner:
a bit more explicitly, an explanation what i have in mind:

~~~~            
command            semantics
------------------------------------------------------------
cabal build        old
cabal new-build    new
cabal old-build    [does not exist]
cabal v1-build     [does not (need to) exist]
canoe build        new
canoe old-build    [does not exist]
canoe new-build    [does not exist]
canoe v1-build     [does not (need to) exist]
~~~~

and no need to drastically change existing semantics. The whole `cabal` executable can be phased out in a few releases, leaving only `canoe`.

This approach almost comes naturally when you consider the meaning of the parts of this invocation:

~~~~
cabal v1-build
|     |   `which function exactly?
|     `which general mode of operation?
`which build tool?
~~~~

where the only of these three that changes regularly when working on any given project is the third, so it makes much more sense to write this as

~~~~
cabal-v1 build
~~~~

and the trivial conclusion is to just rename the executable, and encode build-tool+operation-mode in its name.

lspitzner:
and sorry if this PR is the wrong place to discuss this, but the links @gbaz provided don't exactly feel more appropriate for such feedback.

23Skidoo:
> The old executable can easily be phased out after a while. I don't see what would be particularly eternal about this.

You suggestion is that we should never be allowed to reuse the `cabal` name, i.e. the semantics of `cabal build` should be eternally fixed to `v1-build` because of backwards compat considerations.

lspitzner:
> You suggestion is that we should never be allowed to reuse the cabal name, i.e. the semantics of cabal build should be eternally fixed to v1-build because of backwards compat considerations.

If, as I propose, we phase out "cabal", its semantics change from "old" to "does not exist". I maintain that this is not eternal.

lspitzner:
But yes, I think it would be rather useful if I was able to tell with certainty the intended meaning of a "cabal build" invocation I will find in some old build script at any point in the future.

andreabedini:
I agree with @lspitzner and I made a similar comment in https://github.com/haskell/cabal/pull/5399#issuecomment-404180262

PR #5369, opened 2018-06-09
Avoid deduplication of GHC options that can break builds (rebase of #5368)

23Skidoo:
This is #5368, but rebased against `master` to make history cleaner.


qnikst:
It seems that `ghcOptGHCiScripts` are affected by this problem as well, should I prepare PR for that?

23Skidoo:
@qnikst Sure, please do. I think you should be able to push directly to this branch.

qnikst:
I've added a commit with the fix for `-ghci-script`

23Skidoo:
So CI failed because GHCs < 7.10 don't support `-trust`... fixing.

23Skidoo:
@roman-kashitsyn Merged, thanks!

Issue #4683, opened 2017-08-12
'cabal check' should warn when there are no upper bounds in setup-depends


23Skidoo:
`wxc` [has the following `custom-setup` section](http://hackage.haskell.org/package/wxc-0.92.3.0/wxc.cabal):

```
custom-setup
  setup-depends: 
    base, 
    Cabal,
    bytestring,
    split,
    process,
    directory,
    filepath
```

That is, it says that its setup.hs works with any version of Cabal. However, it is lying - it actually fails to build against Cabal 2. We should warn about such obviously incorrect `setup-depends` and maybe even auto-rewrite the constraint on Cabal to < 2.

23Skidoo:
Related: https://sourceforge.net/p/wxhaskell/bugs/112/

hvr:
As discussed on IRC, it'd also make sense to validate that the lower bound on lib:Cabal is consistent with the spec-version specified in the .cabal file.

grayjay:
> As discussed on IRC, it'd also make sense to validate that the lower bound on lib:Cabal is consistent with the spec-version specified in the .cabal file.

#3751 is related to this.

phadej:
If you review&accept my #4701, then I'd like to tackle this as a follow-up ;)

phadej:
Note: 

```
      +++ custom-setup
      +++     setup-depends: base -any
      +++                    Cabal -any
      +++                    cabal-doctest >=1.0.2 && <1.1
```

here I know that omitting upper bounds on Cabal is safe. Cabal dependency is needed because of https://github.com/haskell/cabal/issues/4288

qnikst:
I'm going to work on this

ulysses4ever:
Since #7409 has been megen, this probably can be closed. Cc @Mikolaj

Mikolaj:
Right. High time. :)

Thank you again, @ulysses4ever, @qnikst and everybody who took part in the struggle.

ulysses4ever:
Thanks a lot @Mikolaj for helping out with it!

PR #5370, opened 2018-06-09
Introduce checks for upper dependencies in setup.

qnikst:
setup depends are now checked for having upper dependencies.
Upper bounds are mandatory for base and Cabal libraries, and
are optional (but emit warning) for other libraries.
Implicit dependencies are not being checked.

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -152,6 +152,7 @@ checkPackage gpkg mpkg =
   ++ checkUnusedFlags gpkg
   ++ checkUnicodeXFields gpkg
   ++ checkPathsModuleExtensions pkg
+  ++ checkSetupVersions gpkg
   where
     pkg = fromMaybe (flattenPackageDescription gpkg) mpkg
 
@@ -1525,21 +1526,7 @@ checkPackageVersions pkg =
 
   ]
   where
-    -- TODO: What we really want to do is test if there exists any
-    -- configuration in which the base version is unbounded above.
-    -- However that's a bit tricky because there are many possible
-    -- configurations. As a cheap easy and safe approximation we will
-    -- pick a single "typical" configuration and check if that has an
-    -- open upper bound. To get a typical configuration we finalise
-    -- using no package index and the current platform.
-    finalised = finalizePD
-                              mempty defaultComponentRequestedSpec (const True)
-                              buildPlatform
-                              (unknownCompilerInfo
-                                (CompilerId buildCompilerFlavor nullVersion)
-                                NoAbiTag)
-                              [] pkg
-    baseDependency = case finalised of
+    baseDependency = case typicalPkg pkg of
       Right (pkg', _) | not (null baseDeps) ->
           foldr intersectVersionRanges anyVersion baseDeps
         where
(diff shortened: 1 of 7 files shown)


23Skidoo:
Re: tests, something like `PackageTests/AutogenModules/Package/setup.test.hs` will do (basically just run `configure` or `check` and then use `assertOutputContains` to check that the messages are present).

23Skidoo:
> Implicit dependencies are not being checked.

You mean for pre-`setup-depends` `build-type: Custom` packages?

qnikst:
I don't cover cases like: https://github.com/haskell/cabal/issues/4683#issuecomment-323101824 or if dependency on a Cabal exists due to the dependency on the other package.

23Skidoo:
That's fine.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -2136,6 +2116,41 @@ checkTarPath path
...
+          | Just spi <- pure $ setupBuildInfo pkgs'
+          , Dependency pname vr <- setupDepends spi
+          ]
+      _ -> Map.empty
+    emitMessage nm
+      | nm `elem` criticalPkgs = emitError nm
+      | otherwise = emitWarning nm
+    emitError nm =
+      PackageDistInexcusable $
+           "The dependency 'build-depends: '"++nm++"' does not specify an "
```
23Skidoo: s/build-depends/setup-depends/.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -2136,6 +2116,41 @@ checkTarPath path
...
+    | (name, vr) <- Map.toList deps
+    , not (boundedAbove vr)
+    ]
+  where
+    criticalPkgs = ["Cabal", "base"]
+    deps = case typicalPkg pkg of
+      Right (pkgs', _) ->
+        Map.fromListWith intersectVersionRanges
+          [ (pname, vr)
+          | Just spi <- pure $ setupBuildInfo pkgs'
```
23Skidoo: I think you can use `maybeToList` here. Also, "spi" should be "sbi" ("bi" by convention means "build info").

quasicomputational:
Coincidentally, I've just implemented making `check` play nicely with `cabal-testsuite` in #5372. I could split that commit off into its own PR and merge that if you think it'd make this PR's test any nicer.

23Skidoo:
A bunch of solver tests now fail...

cabal-testsuite/PackageTests/CheckSetup/setup.test.hs
```diff
@@ -0,0 +1,24 @@
+import Test.Cabal.Prelude
+
+-- Test that setup shows all the 'autogen-modules' warnings.
```
23Skidoo: Copy-pasto.

23Skidoo:
>  I could split that commit off into its own PR and merge that if you think it'd make this PR's test any nicer.

Thanks, but I think we can update the tests that run `cabal check` after your PR has been merged.

qnikst:
I'll need to make this change more complex, currently cabal does not distinguish between `-any` and no dependency set at all. As a result it's not possible to silence the warning by passing `-any`.

23Skidoo:
The change we want to do is replace `data Dependency = Dependency PackageName VersionRange` with `data Dependency = Dependency PackageName (Maybe VersionRange)`.

ip1981:
For a mandatory upper bound (at Hackage if I recall) I set `<50`, how does it help?

gbaz:
closing in favor of #7409

PR #5371, opened 2018-06-09
Try to make unbuildable error message more obvious

hvr:
@hsenag pointed out that "dependency on unbuildable library" isn't necessarily obviously related to `buildable: False`, and may be misinterpreted as a more general cause (like e.g. being unbuildable due to version constraints). This PR is a first suggestion to start the bikeshedding round...

----

Previously,

    Dependency on unbuildable library from darcs
    In the stanza 'test-suite darcs-test'
    In the inplace package 'darcs-2.11.0'

With this patch,

    Dependency on unbuildable (i.e. 'buildable: False') library from darcs
    In the stanza 'test-suite darcs-test'
    In the inplace package 'darcs-2.11.0'


23Skidoo:
OK.

23Skidoo:
Merged, thanks!

PR #5372, opened 2018-06-10
Multi-dot globbing

quasicomputational:

New test added for the `cabal check` and specific tests for this behaviour have been added to the pre-existing glob tests.

The output of `cabal check` has changed a bit, but not for the worse, I don't think; it's now more like the other commands.


Cabal/Distribution/Simple/Glob.hs
```diff
@@ -86,35 +106,59 @@ explainGlobSyntaxError filepath VersionDoesNotSupportGlob =
...
+  GlobStem dir pat' -> do
+    guard (dir == seg)
+    fileGlobMatchesSegments pat' segs
   GlobFinal final -> case final of
-    FinalMatch Recursive ext ->
+    FinalMatch Recursive multidot ext -> do
       let (candidateBase, candidateExts) = splitExtensions (last $ seg:segs)
-      in ext == candidateExts && not (null candidateBase)
-    FinalMatch NonRecursive ext ->
+      guard (not (null candidateBase))
```
23Skidoo: So `*.html` won't match `.foo.html`. Makes sense, since that's how it works in shell, but we should document it.

quasicomputational:
Thanks! Rebased and squashed for tidiness and should be ready to merge, but I'll let it sit for a few days to see if anyone else spots anything that needs changing before the big green button gets pressed.

PR #5375, opened 2018-06-11
travis: Build hackage-tests with -O0 on 7.10.

quasicomputational:
Building hackage-tests has been sporadically OOMing on the 7.10
builder only, so try building them with -O0 to see if that helps
matters.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -168,6 +168,12 @@ flag old-directory
   description:  Use directory < 1.2 and old-time
   default:      False
 
+flag less-memory
+  description:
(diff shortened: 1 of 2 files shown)


quasicomputational:
This Travis build seems to be succeeding, which is a good sign in that it's not a bad one. Hard to say definitively if this fixes the problem since it's a sporadic failure...

23Skidoo:
Let's wait and see if this goes away by itself somehow, I'd prefer if we had less workarounds.

23Skidoo:
Seems no longer to be happening. I'll close this PR for now, we can always reopen.

Issue #4831, opened 2017-10-19
Improve  `outdated`  support for nix-local-build


hvr:
Here's some ideas for how to improve `cabal outdated` with the nix-local-build UI in mind:

 -  `cabal outdated` should support target-selectors like `new-build` et al:

     `cabal outdated all` would then iterate over all local packages configured in `cabal.project` and check all components for over-constricted bounds.

     And it also makes sense to limit `outdated` to some components, i.e. if only care about a few components (and e.g. not the testsuite): `cabal outdated lib:foo exe:doo` would only check the two named components in whichever local package they're located.

    Conversely, we have special names to e.g. refer to all testsuites in the target-selector; so you could also say `cabal outdated tests` to iterate over all test-suite components in your `cabal.project`.

 - `cabal outdated` already supports `cabal.project.freeze`; however, it would be convenient to have a variant which consults the last computed install-plan (and possibly compute one if it's needs (re)computation); IOW, a variant that combines `cabal new-freeze` with `cabal outdated --new-freeze-file` but without generating/overwriting an existing `cabal.project.freeze` file.

/cc @phadej @23Skidoo

andreabedini:
Closing in favour of https://github.com/haskell/cabal/issues/8283

PR #5378, opened 2018-06-13
Backport deprecation warnings to 2.2

typedrat:
I was already working on this, but seeing @lspitzner's concerns in #5358 made me decide to wrap it up and make a PR.

To help ease the transition to 3.0 and the default new-build future, this backports the aliases and most importantly the warnings to ensure that people have time to update any scripts to use the aliases before the semantics would change.

Before this gets merged/released, if it does, I do want to come up with a list of all commands that need `new-` versions before 3.0 so that there will be all relevant warnings included.

--


Cabal/Cabal.cabal
```diff
@@ -1,5 +1,5 @@
 name:          Cabal
-version:       2.2.0.1
+version:       2.2.0.2
```
phadej: why bump `Cabal`?

phadej: Oh, I see, added symbol. That should require third digit bump, patch (4th) level isn't enough.

typedrat:
Seems to have an error in the travis script, but I'm not sure how that works.

typedrat:
Commands that should also get warnings:
- `sdist`
- `outdated`
- `doctest`
- `copy`, `register` are likely not being replaced (based on my understanding, at least), but still
- `reconfigure` 
- `hscolour`?

23Skidoo:
`cabal hscolour` will be killed in favour of Haddock's built-in colourising.

typedrat:
I know, but not sure if that should come with a warning or not.

23Skidoo:
I've just merged a PR adding a warning to `hscolour`. Only to `master` though.

typedrat:
Any that I'm not thinking of?

23Skidoo:
`sandbox`? It's also will be eventually removed in favour of `new-build`.

typedrat:
Okay, adding those. Should I also backport 6922cfc34071f735ad6d9dad0bab695822235ccc in this PR?

23Skidoo:
Please do.

quasicomputational:
If `sdist` and `outdated` are printing warnings, what should the warning message tell users to do instead? AFAIK we don't have new-style equivalents to those yet; until we have a concrete 'here's what you can do to future-proof your code and make this warning go away' suggestion, I don't believe they should warn.

typedrat:
Without sdist install won't work, so it'll be there by the time it changes. outdated? IThere's #4831. And anyway, their suggested option is to switch to `v1-sdist` and `v1-outdated`, because the warning is only on the unprefixed versions.

quasicomputational:
There's still a nag for the `v1-` versions. On HEAD:

```
$ cabal new-run -v0 cabal -- v1-build --help
Compile all/specific components.

[...]

The v1-build command is a part of the legacy v1 style of cabal usage.

It is a legacy feature and will be removed in a future release of cabal-install.
Please file a bug if you cannot replicate a working v1- use case with the new-style
commands.
```

I could get behind the suggestion that `sdist` should warn but `v1-sdist` should be totally nag-free until `new-sdist` is in. I don't understand the interaction with `install`; if `new-sdist` is going to be implemented on the way then this works out, but I do think having a nag for things that don't have replacements will be the best way to get users to ignore the nagging and not upgrade in good time.

typedrat:
Okay, I'll not give them to `outdated` since that's just a nebulously planned thing. `new-sdist` is required to implement `new-install`, so it's coming soon.

I mean, that warning is extremely not wrong and it only shows up in `--help`... I wonder if just wording it differently would work.

PR #5384, opened 2018-06-17
Recognize `powerpc64le` as `PPC64`

trommler:
Add powerpc64le to the list of aliases and document in haddock docs.

Tested by building store-core on a PowerPC 64-bit little endian and verified that `arch(PPC64)` in a conditional evaluates to true.


23Skidoo:
LGTM.

23Skidoo:
Please try to use `rebase` instead of creating merge commits when possible. I fixed the issue myself in this case.

23Skidoo:
This was green before, so merging.

23Skidoo:
Merged, thanks!

PR #5387, opened 2018-06-18
Bump upper bound on containers to <0.7

bgamari:

Just as it says on the tin. Note that I had to do a bit of import-wrangling to get this to compile as `Data.Map.insertWith'` is now gone.


bgamari:
It doesn't look like the one failure is due to this patch.

Cabal/Distribution/PackageDescription/Configuration.hs
```diff
@@ -63,7 +63,8 @@ import Distribution.Types.CondTree
 import Distribution.Types.Condition
 import Distribution.Types.DependencyMap
 
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map.Strict
```
phadej: this module doesn't exist in `containers-0.4`. We have `Cabal.Distribution.Compat.Map` to deal with that.

bgamari: Actually, it looks like `Data.Map.Lazy` also doesn't exist yet there is not a compatibility wrapper for this. I suppose one should be introduced.

phadej: Yes. I think there we need the lazy variant too.

phadej:
Note: we'll need to update bootstrap script too.

bgamari:
> Note: we'll need to update bootstrap script too.

I checked the bootstrap script and saw no particular reference to `containers`. I suspect it takes advantage of the fact that `containers` is GHC core library.

Issue #5386, opened 2018-06-18
Build-type Configure on Windows with all-numeric path segments causes sed failure


snoyberg:
If that repro sounds far-fetched, let me justify it a bit: the temporary paths on AppVeyor look something like `c:\foo\bar\1\baz`, meaning that whenever using a temporary directory to unpack and build, say, `network`, you get a build failure. Original issue: https://github.com/commercialhaskell/stack/issues/3944.

To repro:

1. Create the directory `c:\1`
2. `cd \1`
3. `cabal unpack network-2.7.0.0`
4. `cd network-2.7.0.0`
5. `runghc Setup.hs configure --user`

When using Cabal 2.2, I get the output at the end:

```
sed: -e expression #1, char 213: invalid reference \1 on `s' command's RHS
```

Note that the `\1` changes based on the directory name chosen. I do not get a sed error for Cabal 2.0. I confirmed further that GHC 8.2.2 with its default-shipped Cabal (2.0) works, but upgrading to Cabal 2.2 does not.

snoyberg:
I've reproduced this on AppVeyor using the script:

https://github.com/snoyberg/bug-network-ghc84/blob/44fb43df669afc08415af83a96fbdfdd6d9ec6b3/appveyor.yml

With AppVeyor results available at:

https://ci.appveyor.com/project/snoyberg/bug-network-ghc84

Note that, at the time of writing this, it failed for GHC 8.4.3, but succeeded for GHC 8.2.2.

snoyberg:
Note that, while this is a terrible approach, the following change does in fact fix the bug:

```diff
diff --git a/Cabal/Distribution/Simple.hs b/Cabal/Distribution/Simple.hs
index e45c8312a..06da8d264 100644
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -734,7 +734,9 @@ runConfigureScript verbosity backwardsCompatHack flags lbi = do
                 ((intercalate spSep extraPath ++ spSep)++) $ lookup "PATH" env
       overEnv = ("CFLAGS", Just cflagsEnv) :
                 [("PATH", Just pathEnv) | not (null extraPath)]
-      args' = configureFile:args ++ ["CC=" ++ ccProgShort]
+      args' = map (map replace) $ configureFile:args ++ ["CC=" ++ ccProgShort]
+      replace '\\' = '/'
+      replace c = c
       shProg = simpleProgram "sh"
[...]

PR #5388, opened 2018-06-18
Take more care with special characters in paths to configure scripts

quasicomputational:

Includes two new tests, one making sure that the warnings are emitted and one that tickles the original problem of #5386 on Windows (backslash path delimiters followed by numbers being interpreted by sed).


quasicomputational:
Ugh, it's not triggering on AppVeyor as I'd hoped. I might need to get my hands on a Windows machine and figure out why it's not breaking.

angerman:
Here's the shortest receipt I came up with to reproduce it. Doesn't even use haskell at all:
```
mkdir \\1 && cd \\1
cabal unpack network && cd network-*
mkdir build && cd build
../configure
```

snoyberg:
I'm not following why my original approach of using a relative filepath and setting the working directory wouldn't work instead (I didn't follow the comment on the original issue either). That approach restores something close to the Cabal 2.0 behavior, and works for filepaths with backslashes in them.

quasicomputational:
We need warnings anyway even if we use relative paths; paths with backslashes are actually broken. Quoting the autoconf docs:

> These restrictions apply not only to the files that you distribute, but also to the absolute file names of your source, build, and destination directories.

This patch doesn't implement all of that, but it gets us started.

That's mostly independent of whether we want to go back to relative paths or not to invoke the configure script, but AFAIK there's no other known breakage due to that.

angerman:
@snoyberg, your patch from https://github.com/haskell/cabal/issues/5386#issuecomment-398044552 would run `configure` in the projects root directory (where the cabal file is located).  Preventing this was the original motivation behind the change.  Stipping configure from dirtying the source tree and instead to generating its files in the build folder.

If configure generates file `a`
```
pkg $ ./configure
```
Will produce `pkg/a`

However
```
pkg/build-ghcX.Y $ ../configure
```
Will produce `pkg/build-ghcX.Y/a`

snoyberg:
I didn't glean that from the patches I was reviewing, my mistake. OK, I guess this regression in behavior is simply inherent then, and we'll have to work around Cabal downstream.

quasicomputational:
@snoyberg What's left to work around after we make backslashes not be a problem? If it's not too complicated we can probably do something in Cabal.

snoyberg:
> after we make backslashes not be a problem

If that's the case, then everything's fine. I guess I simply don't understand how this solution works. If this makes it that `runghc Setup.hs configure` works inside `c:\1\foo`, then it does in fact solve the problem and no downstream workaround will be necessary (excluding workarounds for already-shipped Cabal versions).

quasicomputational:
The added test `cabal-testsuite/PackageTests/T5386` exercises that exact case and it works now. The fix is nothing more sophisticated than turning backslashes separating components into slashes on Windows, which is a legitimate operation and one we need to do anyway if we're not going to issue five million false positive warnings about bogus characters in paths.

quasicomputational:
Oh, oops, I changed something in the docs and got it wrong. One second!

quasicomputational:
OK, squashed and pushed it to master. Thank you to @snoyberg for chasing the bug down from such an obscure situation, and for suggesting the fix!

snoyberg:
Thanks @quasicomputational for taking the ball on this and fixing it so promptly, and better than I would have 👍

Issue #4675, opened 2017-08-07
Add support for "setup" components to target selector & new-build


hvr:
Currently, we cannot refer to `custom-setup` components

See [`Distribution.Client.TargetSelector`](https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/TargetSelector.hs) for currently supported syntax.

In order to reify custom-setup components in the target selector syntax, we'd have to introduce a new )(singleton) namespace "`setup`", which has only a single inhabitant (hence a singleton), also named "`setup`". Consequently, this gives us the following ways to refer to the custom setup script for a package foo

 - `foo:setup` (package-name : component-name)
 - `setup` (component-name)
 - `setup:setup` (namespace-name : component-name)
 - `foo:setup:setup` (package-name : namespace-name : component-name )
 - `:pkg:foo:setup:setup` (explicitly qualified package-name : namespace-name : component-name )
 
(Optional) Moreover, we could add the "filter" `setups`, which then gives us the ability to refer to custom-setups components of multiple packages via

 - `setups`
 - `all:setups`
 - `:all:setups`
 - `foo:setups` (trivial: all `setup` components in the package `foo`; IOW, same as `foo:setup:setup`)

This may be a pre-requisite for #4293 which needs the ability to `cabal new-build setup(s)` w/o building all remaining targets.

ezyang:
In fact, we already have a concept for this, in Distribution.Solver.Types.ComponentDep:

```
-- | Component of a package.
data Component =
    ComponentLib
  | ComponentSubLib UnqualComponentName
  | ComponentFLib   UnqualComponentName
  | ComponentExe    UnqualComponentName
  | ComponentTest   UnqualComponentName
  | ComponentBench  UnqualComponentName
  | ComponentSetup
  deriving (Show, Eq, Ord, Generic)
```

So extending this idea further seems profitable.

ezyang:
A prereq for doing this profitability is out-of-line custom setup build, i.e., a stripped down version of https://github.com/haskell/cabal/issues/4047

Issue #5380, opened 2018-06-14
cabal new-build --help says default build-dir is dist/


quasicomputational:
That's a lie: it should be `dist-newstyle/`. I started chasing this up but I got lost in the twisty maze of option parsing.

hvr:
fwiw, `dist-newstyle` is supposed to be renamed to `dist` when `new-build` becomes `build`; so I'm not sure this is of much priority.

typedrat:
I'm gonna say that makes this a WONTFIX.

PR #5389, opened 2018-06-19
Add new-sdist command

### LINKED ISSUE #4293: sdist fails with non-installed custom-depends:
```
$ cabal sdist
[1 of 1] Compiling Main             ( dist/setup/setup.hs, dist/setup/Main.o )
dist/setup/setup.hs:1:1: error:
    Failed to load interface for ‘Distribution.Extra.Doctest’
```

Do we need `cabal new-sdist`?

typedrat:
`new-sdist` is done, and so are the tests (thanks @quasicomputational). There are some unfortunate implications I don't see a way around in terms of how new-style builds work, meaning that the sdist hooks are dead. Fortunately, I don't see many cases this will actually effect. I only found three packages in a representative sample that would be "broken" (at sdist time, has no effects on end users of these packages) and honestly it'd not be much sweat off my back to fix them myself.

Status:
* [x] `.tar.gz` sdists
* [x] `.zip` sdists
* [x] Listing source files
* [x] Works with custom setup (in a sense)

Closes #4293, but does *not* touch #4047 or #4675, previously thought to be required. It solves the issue by simply not running preprocessors and not running `Setup configure` under the logic that whoever is downloading this sdist will be able to do it themselves in all likelihood. Doing otherwise would make it impossible to insure that the same source files will generate an identical sdist, which is required for `new-install` to work.


typedrat:
It is stable at least on the same machine:
```❯ shasum -a 256 dist-newstyle/sdist/Cabal-2.3.0.0.tar.gz ../test/cabal/dist-newstyle/sdist/Cabal-2.3.0.0.tar.gz
1e70454db3e883ef791fc851c63933d0017c790b0f592d7b9e7efeaf8a144042  dist-newstyle/sdist/Cabal-2.3.0.0.tar.gz
1e70454db3e883ef791fc851c63933d0017c790b0f592d7b9e7efeaf8a144042  ../test/cabal/dist-newstyle/sdist/Cabal-2.3.0.0.tar.gz```

**Comment by hvr** [comment]
@23Skidoo this is kinda orthogonal, but do you know who actually benefits from `.zip` support? and why we're keeping it around?

**Comment by typedrat** [comment]
Either way, it's there now at the cost of four (soon to be two, once jgm/zip-archive#47 goes through) dependencies for `cabal-install`.

**Comment by typedrat** [comment]
Formally note down:
- Invariant that custom Setups can't change files in such a way that the file generated by `cabal new-sdist` changes
- `Distribution.Simple.UserHooks.UserHooks{preSDist, sDistHook, postSDist}` are/will be deprecated now, because they interfere with the Glorious Deterministic Future. A replacement is needed, probably some sort of script hooks.

**Comment by typedrat** [comment]
I need to add tests for `new-sdist`, to confirm the determinicity.

**Comment by typedrat** [comment]
To a first approximation, if the sdist hooks completely disppeared tomorrow (and obviously it's more gradual than that), the following (actively maintained and used enough to get into certain curated systems) packages would break:
- gl-0.8.0
- hlibsass-0.1.6.1
- ShellCheck-0.4.7

I've basically managed to fix `gl`, I'd try the other two but I do need some sleep. That's 0.12% of the packages in my sample (latest Stackage LTS).

**Comment by typedrat** [comment]
While I understand there are some weighty elements to this, at the same time it is blocking new-install.

File: cabal-install/Distribution/Client/CmdSdist.hs [hunk 1]
```diff
@@ -0,0 +1,289 @@
...
+        LocalUnpackedPackage path -> return path
+        _ -> die' verbosity "The impossible happened: a local package isn't local"
+    setCurrentDirectory dir
+
+    let norm flag = fmap ((flag, ) . normalise)
+    (norm NoExec -> nonexec, norm Exec -> exec) <- 
+        listPackageSources verbosity (flattenPackageDescription $ packageDescription pkg) knownSuffixHandlers
+
+    let write = if outputFile == "-" then BSL.putStrLn else BSL.writeFile outputFile
+        files =  nub . sortBy (\(_, a) (_, b) -> compare a b) $ nonexec ++ exec
```
quasicomputational: `sortBy (\(_, a) (_, b) -> compare a b)` === `sortOn snd`

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,289 @@
...
+
+    if 
+        | listSources -> do
+            notice verbosity $ "File manifest for package " ++ prettyShow (packageId pkg) ++ ":\n"
+            write (BSL.pack . unlines . fmap snd $ files)
```
quasicomputational: `--list-only` is producing an extra newline at the end of its output, e.g.:

```
~/src/cabal$ cabal new-run cabal -- new-sdist Cabal --list-only
[...]
tests/custom-setup/IdrisSetup.hs
tests/hackage/check.sh
tests/hackage/download.sh
tests/hackage/unpack.sh
tests/misc/ghc-supported-languages.hs

~/src/cabal$ 
```

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,289 @@
...
+        | archiveFormat == TargzFormat -> do
+            let entriesM :: StateT [FilePath] (WriterT [Tar.Entry] IO) ()
+                entriesM = forM_ files $ \(perm, file) -> do
+                    let fileDir = takeDirectory file
+                        perm' = case perm of
+                            Exec -> Tar.executableFilePermissions
+                            NoExec -> Tar.ordinaryFilePermissions
+                    needsEntry <- gets (notElem fileDir)
```
quasicomputational: Quadratic behaviour here in the number of files. Better to use a `Set`.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,289 @@
...
+        "Usage: " ++ pname ++ " new-sdist [FLAGS]\n"
+    , commandDescription  = Just $ \_ ->
+        "Generates a tarball of a package suitable for upload to Hackage."
+    , commandNotes = Nothing
+    , commandDefaultFlags = defaultSdistFlags
+    , commandOptions = \showOrParseArgs ->
+        [ optionVerbosity
+            sdistVerbosity (\v flags -> flags { sdistVerbosity = v })
+        , optionDistPref
+            sdistDistDir (\dd flags -> flags { sdistDistDir = dd })
```
quasicomputational: This is claiming its default is `dist`, but it actually plops the tarball in `dist-newstyle`.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,289 @@
...
+    , createDirectoryIfMissing )
+import System.FilePath
+    ( (</>), (<.>), normalise, takeDirectory )
+
+sdistCommand :: CommandUI SdistFlags
+sdistCommand = CommandUI
+    { commandName = "new-sdist"
+    , commandSynopsis = "Generate a source distribution file (.tar.gz)."
+    , commandUsage = \pname ->
+        "Usage: " ++ pname ++ " new-sdist [FLAGS]\n"
```
quasicomputational: Missing the component name argument.

cabal-install/bootstrap.sh
```diff
@@ -261,6 +261,15 @@ HACKAGE_SECURITY_VER="0.5.3.0"; HACKAGE_SECURITY_VER_REGEXP="0\.5\.((2\.[2-9]|[3
                        # >= 0.5.2 && < 0.6
 TAR_VER="0.5.1.0";     TAR_VER_REGEXP="0\.5\.([1-9]|1[0-9]|0\.[3-9]|0\.1[0-9])\.?"
                        # >= 0.5.0.3  && < 0.6
+# These two are temporary: https://github.com/jgm/zip-archive/pull/47
```
quasicomputational: Not actually a blocker for merging, right? I'd be OK with merging as-is and then fixing up when the PR is merged upstream.

cabal-testsuite/Test/Cabal/Prelude.hs
```diff
@@ -282,6 +282,7 @@ cabalG' global_args cmd args = do
           | cmd `elem` ["v1-update", "outdated", "user-config", "manpage", "v1-freeze", "check"]
           = [ ]
           -- new-build commands are affected by testCabalProjectFile
+          | cmd == "new-sdist" = [ "--project-file", testCabalProjectFile env ]
```
quasicomputational: Isn't this covered by the next case, catching all the `new-` commands?

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,289 @@
...
+    , commandDescription  = Just $ \_ ->
+        "Generates a tarball of a package suitable for upload to Hackage."
```
quasicomputational: s/a package/one or more packages/?

Cabal/doc/nix-local-build.rst
```diff
@@ -478,6 +478,31 @@ and caches if the ``--save-config`` option is given, in which case it only remov
...
+---------------
+
+``cabal new-sdist [FLAGS] [TARGETS]`` takes the crucial files needed to build ``TARGETS``
+and puts them into an archive format ready for upload to Hackage. These archives are stable
+and two archives of the same format built from the same source will hash to the same value.
+
+``cabal new-sdist`` takes the following flags:
+
+- ``--list-only``: Rather than creating an archive, lists files that would be included.
+  Output is to ``stdout`` by default.
```
quasicomputational: Should mention it takes an optional output file argument and that making its argument `-` sends it to stdout.

Cabal/doc/nix-local-build.rst
```diff
@@ -478,6 +478,31 @@ and caches if the ``--save-config`` option is given, in which case it only remov
...
+
+- ``--targz``: Output an archive in ``.tar.gz`` format.
+
+- ``--zip``: Output an archive in ``.zip`` format.
+
+- ``--output-dir``: Sets the output dir, if a non-default one is desired. The default is
+  ``dist-newstyle/sdist/``.
```
quasicomputational: It's only `dist-newstyle` by default, and that's controlled by a separate flag; if `--builddir=foo`, then it's plopped in `foo/sdist`.

quasicomputational:
Oh, one more thing: can `--list-only` grow a `-z` / `--null` option, that will separate file names with a NUL byte rather than a newline, so that anything processing the file list downstream doesn't break when files have newlines in their name?

quasicomputational:
Ah, found a bug: `cabal new-sdist Cabal --output-dir .` drops the tarball in `Cabal/`. That should be relative to the CWD of the invocation. Same goes for `cabal new-sdist Cabal --output-dir foo`, where it'll try to put it in `Cabal/foo`, rather than `./foo`

Cabal/doc/nix-local-build.rst
quasicomputational: Oops, sorry, I got confused. This is nonsense.

quasicomputational:
Sorry, I spotted another thing on second look: `cabal new-sdist Cabal --list-only -o -` does not work as advertised in the docs: it tries to write to `-/Cabal-2.3.0.0.list`, rather than stdout. I'm also not sure that `-o -` is sensible for generating multiple tarballs; maybe it should error out in that case?

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: Whole thing needs to be pluralized, since there's one package per tarball but it can do multiple tarballs in one run.

cabal-install/bootstrap.sh
typedrat: Nope, not a blocker. That's to remind us when it does get merged (hopefully, I don't see a reason it wouldn't be).

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: That's #5380 and it's its own issue. `optionDistPref` needs fixing, but that's neither here nor there for this patchset.

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: 👍

typedrat: I generally don't feel like being too aggressively optimizing before we know it's an issue, but adding four characters to go from `O(n^2)` to `O(n log n)` is worth it.

typedrat:
I forgot to make the base path absolute before we start hopping around, that's trivial to fix and... Oh wow I have no idea how I managed to not catch that.

Cabal/doc/nix-local-build.rst
typedrat: I mean, it is the default, and I feel like you'd know if it wan't. `BUILDDIR/sdist/` seems harder.

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: Whoops, why is it `putStrLn`?

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: Whoops, know why. Not an instance of `Ord`.

typedrat:
I will write more tests later today.

quasicomputational:
I can chip in with a bunch as well. Thanks for the changes so far; I think that tests and a little bit more documenting (which I'll prepare a PR into your branch for) and then it's in great shape in terms of nuts and bolts.

typedrat:
I gotta say, huge *huge* props to @quasicomputational for how much help they've been today.

typedrat:
The Appveyor failure is specious, it's ready to merge unless there is a substantial yet heretofore unvoiced objection.

quasicomputational:
The AppVeyor failure is actually half-legit: it's testing for a path with slashes in the `.list` file, but of course it's backslashes on Windows. I managed to make that exact same mistake a few weeks ago and apparently didn't learn from it the first time. typedrat/cabal#5 fixes that and makes a few other tweaks to the tests; with that merged I'm happy with the test coverage and the code itself.

Cabal/doc/nix-local-build.rst
```diff
@@ -478,6 +478,36 @@ and caches if the ``--save-config`` option is given, in which case it only remov
 the build artefacts (``.hi``, ``.o`` along with any other temporary files generated
 by the compiler, along with the build output).
 
+cabal new-sdist
+---------------
+
+``cabal new-sdist [FLAGS] [TARGETS]`` takes the crucial files needed to build ``TARGETS``
+and puts them into an archive format ready for upload to Hackage. These archives are stable
+and two archives of the same format built from the same source will hash to the same value.
```
23Skidoo: For `autogen-modules` we can't guarantee this 100% since we have no control over non-determinism in preprocessors, especially custom ones. For example, a preprocessor can add a comment with the date when the file was generated, or have output that depends on system locale settings. I guess that we can settle on saying that the behaviour is undefined when a preprocessor is non-deterministic in a way that can be observed from compiled code.

Cabal/doc/nix-local-build.rst
```diff
@@ -478,6 +478,36 @@ and caches if the ``--save-config`` option is given, in which case it only remov
...
+- ``--zip``: Output an archive in ``.zip`` format.
+
+- ``-o``, ``--output-dir``: Sets the output dir, if a non-default one is desired. The default is
+  ``dist-newstyle/sdist/``. ``--output-dir -`` will send output to ``stdout``
+  unless multiple archives are being created.
+
+- ``-z``, ``--null``: Only used with ``--list-only``. Separates filenames with a NUL
+  byte instead of newlines.
+
+``new-sdist`` is inherently incompatible with sdist hooks, not due to implementation but due
```
23Skidoo: OK from me, let's deprecate them.

cabal-install/changelog
```diff
@@ -1,6 +1,8 @@
 -*-change-log-*-
 
 2.4.0.0 (current development version)
+	* Add 'new-sdist' command (#5389). Creates stable archives based on
```
23Skidoo: Please also add a note to the Cabal changelog saying that sdist hooks are now deprecated and `autogen-modules` and `build-tool-depends` should be used instead.

Cabal/Distribution/Simple/UserHooks.hs
```diff
@@ -165,6 +165,9 @@ data UserHooks = UserHooks {
   }
 
 {-# DEPRECATED runTests "Please use the new testing interface instead!" #-}
+{-# DEPRECATED preSDist  "SDist hooks violate the invariants of new-sdist." #-}
```
23Skidoo: Should point the user towards what they should use instead.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE ViewPatterns #-}
+module Distribution.Client.CmdSdist ( sdistCommand, sdistAction, packageToSdist, OutputFormat(..), ArchiveFormat(..) ) where
```
23Skidoo: Nitpick: please try to keep all lines <= 80 chars long. I know that this rule is not followed 100%, but we should try to follow it in new code.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
...
+    , sdistNulSeparated  = toFlag False
+    , sdistArchiveFormat = toFlag TargzFormat
+    , sdistOutputPath    = mempty
+    }
+
+--
+
+sdistAction :: SdistFlags -> [String] -> GlobalFlags -> IO ()
+sdistAction SdistFlags{..} targetStrings globalFlags = do
+    let verbosity = fromFlagOrDefault normal sdistVerbosity
```
23Skidoo: If you use Emacs, you can make formatting prettier here (and elsewhere) with `M-x align-regexp =`. If you don't, don't bother.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
...
+    
+        outputPath pkg = case mOutputPath' of
+            Just path
+                | path == "-" -> "-"
+                | otherwise   -> path </> prettyShow (packageId pkg) <.> ext
+            Nothing
+                | listSources -> "-"
+                | otherwise   -> distSdistFile distLayout (packageId pkg) archiveFormat
+
+    createDirectoryIfMissing True (distSdistDirectory distLayout)
```
23Skidoo: Strictly speaking, this is not needed when outputting to stdout, but no-one probably cares.

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
...
+    
+    case reifyTargetSelectors localPkgs targetSelectors of
+        Left errs -> die' verbosity . unlines . fmap renderTargetProblem $ errs
+        Right pkgs 
+            | length pkgs > 1, not listSources, Just "-" <- mOutputPath' -> 
+                die' verbosity "Can't write multiple tarballs to standard output!"
+            | otherwise ->
+                mapM_ (\pkg -> packageToSdist verbosity (distProjectRootDirectory distLayout) format (outputPath pkg) pkg) pkgs
+
+data IsExec = Exec | NoExec
```
23Skidoo: No boolean blindness, nice!

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
...
+        ([], sels') -> Right sels'
+        (errs, _)   -> Left errs
+    where
+        flatten (SpecificSourcePackage pkg@SourcePackage{}) = pkg
+        flatten _ = error "The impossible happened: how do we not know about a local package?"
+        pkgs' = fmap flatten pkgs
+
+        getPkg pid = case find ((== pid) . packageId) pkgs' of
+            Just pkg -> Right pkg
+            Nothing -> error "The impossible happened: we have a reference to a local package that isn't in localPackages."
```
23Skidoo: Perhaps add a `TargetProblemImpossible` and use it instead of `error` here?

cabal-install/Distribution/Client/CmdSdist.hs
```diff
@@ -0,0 +1,321 @@
...
+        LocalUnpackedPackage path -> return path
+        _ -> die' verbosity "The impossible happened: a local package isn't local"
+    setCurrentDirectory dir
+
+    let norm flag = fmap ((flag, ) . normalise)
+    (norm NoExec -> nonexec, norm Exec -> exec) <- 
+        listPackageSources verbosity (flattenPackageDescription $ packageDescription pkg) knownSuffixHandlers
+
+    let write = if outputFile == "-"
+          then putStr . withOutputMarker verbosity . BSL.unpack
```
23Skidoo: Maybe add a helper function for this to `D.S.Utils`?

Cabal/doc/nix-local-build.rst
typedrat: `autogen-modules` *shouldn't* (and it's a bug if they are, that  probably needs a test to be sure) be included for this very reason.

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: I typically don't care much about that, but that is getting a bit silly.

cabal-install/Distribution/Client/CmdSdist.hs
typedrat: Didn't do it since it's harder to get rid of the one in flatten and I don't feel like it's worth the work when there's another five lines above.

Cabal/doc/nix-local-build.rst
23Skidoo: You're correct! Sorry, it's getting a bit late here.

Issue #5341, opened 2018-05-23
new-haddock can't produce documentation suitable for Hackage?


RyanGlScott:
I'm not sure if this is expected behavior or not, but it's certainly throwing me for a loop. I occasionally need to upload Haddocks to Hackage manually. With old-style `cabal`, my usual workflow would be:

1. Create a sandbox.
2. Install all dependencies with `--enable-documentation` on.
3. Run `cabal haddock --for-hackage`.
4. Run `cabal upload -d <doc-tarball> --publish`

I'm trying to figure out how to emulate this workflow with `new-haddock`, but it doesn't seem to be quite up to the task. My first inclination was to try this:

```
$ cabal new-build --enable-documentation
```

However, this did not build my dependencies with documentation enabled. It simply built the package itself (using previously built versions of the dependencies, without documentation) and ran Haddock on the package, and unsurprisingly, that resulted in unbound link destinations:

```
Warning: Data.Eliminator.TH: could not find link destinations for:
    ~> Sing @@
Warning: Data.Eliminator: could not find link destinations for:
    ~> Sing @@ Nat Z S
Warning: Data.Eliminator.TypeNats: could not find link destinations for:
    ~> Sing @@
```

Passing `--haddock-for-hackage` didn't seem to help either.

I also tried directly invoking `cabal new-haddock --haddock-for-hackage`, but that too did not build the dependencies with documentation enabled.

23Skidoo:
I think `new-haddock` should support this. Most likely just unfinished.

merijn:
I think this works in HEAD or one of Alex's branches, since I've successfully build and uploaded Hackage docs using new-haddock 1 or 2 months ago when he was working on that.

@RyanGlScott are you using cabal-install HEAD or older?

RyanGlScott:
@merijn, I used `cabal-install-2.2` in my original attempt. I just tried to use the latest `cabal-install` HEAD from @hvr's Ubuntu PPA, but I experience the same issues.

RyanGlScott:
I'd say that this is fixed after #5395, since I can now run:

```
$ cabal new-build --enable-documentation --haddock-for-hackage
```

And it will build suitable-for-Hackage Haddocks, building dependencies with documentation enabled if need be.

PR #5395, opened 2018-06-24
Add Haddock flags to package hash

alexbiehl:
This should help with https://github.com/haskell/cabal/issues/5341

This adds haddock flags and arguments to the package hash. Changing haddock flags for non-local packages will retrigger a build of these.

I view this as a quick fix. I think we really want documentation to be treated as a separate component.

### DIFF
--- a/Cabal/Distribution/Simple/Program/Builtin.hs
+++ b/Cabal/Distribution/Simple/Program/Builtin.hs
@@ -316,7 +316,7 @@ haddockProgram = (simpleProgram "haddock") {
         (_:_:ver:_) -> takeWhile (`elem` ('.':['0'..'9'])) ver
         _           -> "",
 
-    programNormaliseArgs = \_ _ _ -> []
+    programNormaliseArgs = \_ _ args -> args
   }
 
 greencardProgram :: Program
--- a/cabal-install/Distribution/Client/PackageHash.hs
+++ b/cabal-install/Distribution/Client/PackageHash.hs
@@ -211,11 +211,25 @@ data PackageHashConfigInputs = PackageHashConfigInputs {
        pkgHashExtraFrameworkDirs  :: [FilePath],
        pkgHashExtraIncludeDirs    :: [FilePath],
        pkgHashProgPrefix          :: Maybe PathTemplate,
-       pkgHashProgSuffix          :: Maybe PathTemplate
+       pkgHashProgSuffix          :: Maybe PathTemplate,
+
+       -- Haddock options
+       pkgHashDocumentation       :: Bool,
+       pkgHashHaddockHoogle       :: Bool,
+       pkgHashHaddockHtml         :: Bool,
+       pkgHashHaddockHtmlLocation :: Maybe String,
+       pkgHashHaddockForeignLibs  :: Bool,
+       pkgHashHaddockExecutables  :: Bool,
+       pkgHashHaddockTestSuites   :: Bool,
+       pkgHashHaddockBenchmarks   :: Bool,
+       pkgHashHaddockInternal     :: Bool,
+       pkgHashHaddockCss          :: Maybe FilePath,
+       pkgHashHaddockLinkedSource :: Bool,
+       pkgHashHaddockQuickJump    :: Bool,
+       pkgHashHaddockContents     :: Maybe PathTemplate
 
 --     TODO: [required eventually] pkgHashToolsVersions     ?
 --     TODO: [required eventually] pkgHashToolsExtraOptions ?
---     TODO: [research required] and what about docs?
      }
   deriving Show
 
@@ -248,7 +262,7 @@ renderPackageHashInputs PackageHashInputs{
     -- the default value for that feature. So if we avoid adding entries with
     -- the default value then most of the time adding new features will not
     -- change the hashes of existing packages and so fewer packages will need
-    -- to be rebuilt. 
+    -- to be rebuilt.
 
     --TODO: [nice to have] ultimately we probably want to put this config info
     -- into the ghc-pkg db. At that point this should probably be changed to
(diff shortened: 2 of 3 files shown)


RyanGlScott:
Very cool, thanks @alexbiehl! I tested this out on the package in https://github.com/haskell/cabal/issues/5341#issue-325697577 and it works quite well for my use case.

I have one question: once this change is in place, will it still be possible to build a package's Haddocks _without_ having to build the Haddocks for all of its libraries? The use case for this feature would be something like `haskell-ci`, where it would be convenient to quickly test that a package's Haddocks work without needing to rebuild all of the dependencies anew. (If the answer is "no", I don't think that should prevent this patch from being landed—I'm just curious.)

alexbiehl:
I ran out of time to test unfortunately but my understanding is that passing --enable-documentation should only haddock local packages. Could you verify?

alexbiehl:
Ok. I will see what I can do. Meanwhile on my local machine: 

Trying to build cabal with this patch enbled:
```
$ ./cabal-dev new-build exe:cabal
Build profile: -w ghc-8.4.2 -O1
 - Cabal-2.3.0.0 (lib) (first run)
 - base16-bytestring-0.1.1.6 (lib:base16-bytestring) (requires build)
 - base64-bytestring-1.0.0.1 (lib) (requires build)
 - echo-0.1.3 (lib) (requires build)
 - clock-0.7.2 (lib) (requires build)
 - call-stack-0.1.0 (lib) (requires build)
 - colour-2.3.4 (lib) (requires build)
 - cryptohash-sha256-0.11.101.0 (lib) (requires build)
 - ed25519-0.0.5.0 (lib) (requires build)
Assertion failed
  assert, called at ./Distribution/Client/ProjectPlanning.hs:236:5 in cabal-install-2.3.0.0-inplace:Distribution.Client.ProjectPlanning
```

RyanGlScott:
Out of curiosity, if `--enable-documentation` alone is intended to just be for local packages, how do you envision someone specifying that the documentation should be built for _all_ packages?

alexbiehl:
I have a small time frame left, I will write something up in a minute
hopefully.

Ryan Scott <notifications@github.com> schrieb am So., 24. Juni 2018, 19:48:

alexbiehl:
@hvr @RyanGlScott @23Skidoo Here is a (hopefully complete) list on how to pass `documentation: true` to cabal and what packages it should consider for documentation generation:

- Global configuration `$HOME/.cabal/config`

    ```
    documentation: true
    ```

    - [x] local 
    - [x] non local

- CLI argument`--enable-documentation`

    - [x] local
    - [ ] non local

- Project local configuration `cabal.project` / `cabal.project.local`

    ```
     documentation: true
    ```

    - [x] local
    - [ ] non local

- Project local configuration `cabal.project` / `cabal.project.local`

    ```
    package *
      documentation: true 
    ````

    - [x] local
    - [x] non local

Does it make sense/is it what we want?

RyanGlScott:
> @RyanGlScott btw, for the purpose of haskell-ci, why do you think that having --enable-documentation apply to non-local packages is a bad thing

For the same reason that I wouldn't want `haskell-ci` to build all the test suite and benchmark components of all downstream dependencies: they're irrelevant to ensuring that the package you actually care about builds correctly, and building them simply wastes cycles.

> (I specifically am confused about the "rebuild everything" part... wouldn't those be cached just like the no-documentation artifacts)?

Caching exists, but it doesn't change the fact that on a from-scratch build, you're now going to have to build every single dependency twice. For large projects like `lens`, this can add up to the order of 10–15 extra minutes.

hvr:
@RyanGlScott I'm still confused about the need to compile things twice during CI -- just build everything always (& once) w/ enable-docs?

However, what's the purpose of building haddocks during haskell-ci to begin with, if you're only building partial docs? isn't the intent of buidling haddocks as part of CI in order to validate that all haddock-links are valid (which requires having transitive deps being haddocked as well)?

RyanGlScott:
> @RyanGlScott I'm still confused about the need to compile things twice during CI -- just build everything always (& once) w/ enable-docs?

Ah, never mind. For some reason, my mind was stuck in the old paradigm of building the package, then building documentation separately. I forgot that you could just enable `documentation: true` from the get-go and have that be on when you're building dependencies, eliminating the need to build everything twice.

In that case, please ignore my ramblings above.

merijn:
The downside of this PR seems to be that if I build a package, then rebuild it with ``--enable-documentation`` then my global store will include all my dependencies *twice*.

merijn:
@alexbiehl also double check how things interact with https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Program/Builtin.hs#L319

When I was asking which flags to filter I was initially told haddock flags shouldn't affect the hash, so that filters them out. It's called from https://github.com/haskell/cabal/blob/2ce4858881924fbd1ed22e133b6e7a999737b965/cabal-install/Distribution/Client/ProjectPlanning/Types.hs#L317 which in turns is called by https://github.com/haskell/cabal/blob/2ce4858881924fbd1ed22e133b6e7a999737b965/cabal-install/Distribution/Client/ProjectBuilding.hs#L334-L359 (for determining whether to rebuild local dependencies) and https://github.com/haskell/cabal/blob/2ce4858881924fbd1ed22e133b6e7a999737b965/cabal-install/Distribution/Client/ProjectPlanning.hs#L3564-L3593 (to compute the package hash for non-local dependencies).

hvr:
@merijn 

> When I was asking which flags to filter I was initially told haddock flags shouldn't affect the hash, so that filters them out.

Indeed, the rationale was according to the rule-of-thumb, that they wouldn't affect the compiled `.o` artefacts (and thus affect the primary goal of compiled programs). So consider including them now more of a pragmatic choice until we have a different way to address the problem that haddocks can't be  generated lazily *yet*, and so we need to consider them (for now) part of the nix-style hash in order to improve the user experience until we have a different more granular way to represent haddocks as artefacts.

phadej:
@bgamari is asking for a release branch, @23Skidoo said this should go in. What's the state of this PR?

alexbiehl:
This is fine to my knowledge. @merijn, @hvr and @sjakobi dog fooded. I haven't negative things (apart from the obvious store trashing).

alexbiehl:
Ugh. I just saw https://github.com/haskell/cabal/pull/5395#issuecomment-400637196. @merijn is this still a thing? Should I update my patch to remove haddock flags from your filter?

merijn:
I didn't actually touch the filtering I added for Haddock flags, on account of being to busy. It looks like @alexbiehl added custom fields to the package hash, but I'm not sure if anyone looked at the interaction? For safety's sake my change in Cabal should probably be reverted? This should be simple enough to do, the relevant line is: https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Program/Builtin.hs#L319

This function just has to be changed to return it's 3rd argument (the list of args) unchanged, so `\_ _ xs -> xs`. So should be trivial for @alexbiehl to include in his patches.

merijn:
(Yay for proper design resulting in only a single trivial change to avoid any problems! ;))

23Skidoo:
Merged, thanks! Sorry for taking so long.

PR #5396, opened 2018-06-24
Correct formatting of the legacy command name warning

quasicomputational:
We don't need to hard-wrap the message as we originate it, since
that's handled elsewhere.

### DIFF
--- a/cabal-install/Distribution/Client/CmdLegacy.hs
+++ b/cabal-install/Distribution/Client/CmdLegacy.hs
@@ -113,9 +113,9 @@ deprecationNote cmd =
     "The " ++ cmd ++ " command is a part of the legacy v1 style of cabal usage.\n\n" ++
 
     "Please switch to using either the new project style and the new-" ++ cmd ++ 
-    " command or the legacy v1-" ++ cmd ++ "alias as new-style projects will" ++
-    "become the default in the next version of cabal-install. Please file a" ++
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/CmdLegacy.hs
```diff
@@ -112,9 +112,9 @@ deprecationNote :: String -> String
 deprecationNote cmd =
     "The " ++ cmd ++ " command is a part of the legacy v1 style of cabal usage.\n\n" ++
 
-    "Please switch to using either the new project style and the ++ new-" ++ cmd ++ 
-    "command or the legacy v1-" ++ cmd ++ "\nalias as new-style projects will\n" ++
-    "become the default in the next version of cabal-install. Please file a\n" ++
+    "Please switch to using either the new project style and the new-" ++ cmd ++ 
+    " command or the legacy v1-" ++ cmd ++ "alias as new-style projects will" ++
```
typedrat: That needs a space, or else we get `v1-buildalias`.

quasicomputational: Good spot, thanks.

quasicomputational:
Upon further investigation, it is wrapped properly for, e.g., `cabal update` but the `cabal update --help` note is not wrapped at all. Kind of annoying that the `--help` text is going to have to be hard-wrapped and won't adjust to terminal size, but unless I'm missing something it looks unavoidable.

PR #5401, opened 2018-06-27
Allow 2.4-spec files to be parsed.

quasicomputational:
Lexically, they're identical to 2.2, but the semantics of globs have
changed slightly.

cc @phadej. I grepped for `CabalSpecV2_2` and put in an analogous `CabalSpecV2_4` wherever I found it. Is there anything else that needs doing?

### DIFF
--- a/Cabal/Distribution/CabalSpecVersion.hs
+++ b/Cabal/Distribution/CabalSpecVersion.hs
@@ -34,6 +34,7 @@ cabalSpecFeatures CabalSpecV2_2  = Set.fromList
 cabalSpecFeatures CabalSpecV2_4  = Set.fromList
     [ Elif
     , CommonStanzas
+    , Globstar
     ]
 
 cabalSpecSupports :: CabalSpecVersion -> [Int] -> Bool
@@ -61,6 +62,9 @@ specHasElif _             = NoElif
 data CabalFeature
     = Elif
     | CommonStanzas
+    | Globstar
+      -- ^ Implemented in #5284. Not actually a change to the parser,
+      -- as filename patterns are opaque to it currently.
   deriving (Eq, Ord, Show, Read, Enum, Bounded, Typeable, Data, Generic)
 
 -------------------------------------------------------------------------------
@@ -72,3 +76,5 @@ data HasElif = HasElif | NoElif
 
 data HasCommonStanzas = HasCommonStanzas | NoCommonStanzas
   deriving (Eq, Show)
+
+data HasGlobstar = HasGlobstar | NoGlobstar
--- a/Cabal/Distribution/CabalSpecVersion.hs
+++ b/Cabal/Distribution/CabalSpecVersion.hs
@@ -20,7 +20,7 @@ data CabalSpecVersion
   deriving (Eq, Ord, Show, Read, Enum, Bounded, Typeable, Data, Generic)
 
 cabalSpecLatest :: CabalSpecVersion
-cabalSpecLatest = CabalSpecV2_2
+cabalSpecLatest = CabalSpecV2_4
 
 cabalSpecFeatures :: CabalSpecVersion -> Set.Set CabalFeature
 cabalSpecFeatures CabalSpecOld   = Set.empty
--- a/Cabal/Distribution/CabalSpecVersion.hs
+++ b/Cabal/Distribution/CabalSpecVersion.hs
@@ -16,6 +16,7 @@ data CabalSpecVersion
     | CabalSpecV1_24
     | CabalSpecV2_0
     | CabalSpecV2_2
+    | CabalSpecV2_4
   deriving (Eq, Ord, Show, Read, Enum, Bounded, Typeable, Data, Generic)
 
 cabalSpecLatest :: CabalSpecVersion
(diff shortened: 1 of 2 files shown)


quasicomputational:
Hmm. There's one more wrinkle: `D.S.Check` has its own, independent `cabal-version` check. I wonder if it's worth dropping that, considering that the shiny new Parsec parser will outright refuse to parse anything it doesn't know about.

phadej:
Do we have tests (`parser-tests`) which test that with 2.4 feature-full cabal files are rejected if they have `cabal-version: 2.2` or less.

e.g. like 
https://github.com/haskell/cabal/blob/4d3febdfd5848f312f579905bcfc89ca6e8be9c9/Cabal/tests/ParserTests/errors/leading-comma.cabal and 
https://github.com/haskell/cabal/blob/4d3febdfd5848f312f579905bcfc89ca6e8be9c9/Cabal/tests/ParserTests/errors/leading-comma.errors

or if they are rejected by `cabal check` then in `check-tests`?

phadej:
Generally, we need a separate changelog for cabal spec. @hvr probably have thought about that already.

phadej:
:+1: for additional chapter. I'd really like to have it. Having another source (i.e. non defined by the implementation) would be great, Than we can say whether implementation is buggy and can be simply fixed.

quasicomputational:
AFAICT there have been no changes to the parser itself between 2.2 and 2.4.

And AFAIK the only thing gated on 2.4 is my glob stuff, which is a `cabal check` thing, yeah. That's tested in a few places:

* `Cabal/tests/ParserTests/pre-2.4-globstar.cabal` and most things under `cabal-testsuite/PackageTests/Check` are for ensuring `cabal check` issues warnings
* `Cabal/tests/UnitTests/Distribution/Simple/Glob.hs` tests the actual behaviour switching

So I'll need to add a spec changelog chapter to the user's guide? That's doable.

phadej:
@quasicomputational :+1: for tests.

If you could start the chapter (as a part of separate PR) that would be super great!

quasicomputational:
OK, I've documented the changes in 2.4 in a new changelog in #5406.

Are there any concerns about the code changes in this PR, or other things I need to do?

Cabal/Distribution/CabalSpecVersion.hs
```diff
@@ -30,20 +31,27 @@ cabalSpecFeatures CabalSpecV2_2  = Set.fromList
     [ Elif
     , CommonStanzas
     ]
+cabalSpecFeatures CabalSpecV2_4  = Set.fromList
+    [ Elif
+    , CommonStanzas
```
23Skidoo: Would it make sense to add `Globstar` here?

Cabal/Distribution/PackageDescription/Parsec.hs
```diff
@@ -125,7 +125,7 @@ parseGenericPackageDescription bs = do
     setCabalSpecVersion ver
     -- if we get too new version, fail right away
     case ver of
-        Just v | v > mkVersion [2,2] -> parseFailure zeroPos
+        Just v | v > mkVersion [2,4] -> parseFailure zeroPos
```
23Skidoo: Hmm, why did this work before? We don't have `.cabal` files with `cabal-version: 2.4` in the test suite?

Cabal/Distribution/CabalSpecVersion.hs
quasicomputational: Hmm. We could. I didn't think to because the parser is wholly agnostic to the pattern syntax. AFAICT, this 'feature' concept is only used to toggle bits of `D.PD.Parsec` on and off, depending on the version.

Cabal/Distribution/PackageDescription/Parsec.hs
quasicomputational: Apparently not! I only found I needed to do this when I tried to add one in the course of #5403.

23Skidoo: Nice, that means that with #5403 we will have one.

Cabal/Distribution/CabalSpecVersion.hs
23Skidoo: Could you please implement this then if it's not too much trouble?

quasicomputational:
OK. This has been up for a week, @23Skidoo has looked at it and I've put in the `Globstar` feature as suggested, so I'm going to go ahead and merge.

23Skidoo:
@quasicomputational 

Great, thanks! BTW, was it you who did this:

```
* [new tag]             1283           -> 1283
* [new tag]             1285           -> 1285
* [new tag]             1311           -> 1311
* [new tag]             1312           -> 1312
```
?

`git show 1285` says

```
tag 1285
Tagger: EXIA$ <EXIA$@EXIA>
Date:   Sun Jul 1 21:35:43 2018 +0100
commit c718a22fa6aa93916269537b27cde72dbe32d613 (HEAD -> master, tag: show, tag: 13ag: 1285, tag: 1283)
Author: quasicomputational <quasicomputational@gmail.com>
```

quasicomputational:
Huh. I have no idea what that is or how I might have done that. `1311` and `1312` could plausibly have been me fat-fingering workspace switching somehow, but I can't work out how I'd manage to accidentally tag anything and then push tags (which isn't the default).

23Skidoo:
@Mistuke Apparently it was you who is responsible for pushing these tags:

```
21:06 < hexagoxel> refold: have you looked at https://api.github.com/repos/haskell/cabal/events ?
21:07 < hexagoxel> refold: if i am reading this stuff correctly, the pusher is account "Mistuke"
21:16 < hexagoxel> {"type": "CreateEvent", "payload": { "ref": "1348", .. }, "actor": { .. "login": "Mistuke", .. }}
21:19 < hexagoxel> (their email would be "tamar@zhox.com", according to some more api digging
```

Can you please stop doing that?

Mistuke:
@23Skidoo well it certainly wasn't intentional, I think I got rid of the flag that was causing this. If not let me know.

PR #5402, opened 2018-06-27
Remove matchFileGlob.

quasicomputational:
This was only a convenience function, but its use could obscure how it
is introducing a dependency on the CWD. By removing it, the "."
argument to `matchDirFileGlob` is explicit.

Any external code using `matchFileGlob` would have needed to be
changed as #5284 changed its signature and the module it lives in; it
is not much more of a burden to switch to `matchDirFileGlob` at the
same time.


23Skidoo:
Looks good, let's merge.

PR #5403, opened 2018-06-27
cabal check improvements for globs

quasicomputational:

Tests added!

This PR is based on top of #5401 and #5402 (which explains today's PR frenzy: I didn't expect to need to do #5401 when I started!). I'll rebase on master once those two are merged. As noted in #5401, the `cabal-version` check in `Check.hs` is kind of weird; I have left it in the expected output of the relevant test on the basis that it's easy to remove later.

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -65,6 +65,8 @@
     ([#5386](https://github.com/haskell/cabal/issues/5386)).
   * `Distribution.PackageDescription.Check.checkPackageFiles` now
     accepts a `Verbosity` argument.
+  * `cabal check` now warns about globs that refer to missing
+    directories.
 
 ----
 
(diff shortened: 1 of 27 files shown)


quasicomputational:
This PR actually needs correction on two counts:

* one, it's a lie to use `PackageDistInexcusable` here because Hackage *won't* reject these. The no-empty-glob check could actually be done Hackage-side, though. What's the policy on introducing new ways for Hackage to reject code?

* two, now that globs are being expanded by `cabal check`, `cabal check` will bork if the globs have bad syntax. This isn't new with this PR but I should fix it anyway.

quasicomputational:
I've fixed the two problems above. I think it'd still make sense to bump up the 'glob matches nothing' and the 'glob failed to parse' warnings to `PackageDistInexcusable` and having Hackage enforce those, but I don't know the backwards-compatibility story here.

quasicomputational:
Oh, and this will also conflict with #5397, but that's an easy merge; if I'd thought, I would have based this on top of that one too. I'd definitely suggest putting #5397 in first because that's causing some grief for Hadrian.

typedrat:
Tested this a bit and read the code. LGTM.

quasicomputational:
Thanks!

PR #5405, opened 2018-06-27
Make --dependency accept a component too [WIP]

fgaz:
About time I merge something. This change should be fairly independent from the rest.

---

The old `--dependency` could only do `--dependency=pkg=cid`,
but with public sublibraries this will become insufficient.

Now there is the option to also specify a component name using
`--dependency=pkg:component=cid`

---

/cc @ezyang @23Skidoo

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -434,7 +434,7 @@ configure (pkg_descr0, pbi) cfg = do
     -- that is not possible to configure a test-suite to use one
     -- version of a dependency, and the executable to use another.
     (allConstraints  :: [Dependency],
-     requiredDepsMap :: Map PackageName InstalledPackageInfo)
+     requiredDepsMap :: Map (PackageName, ComponentName) InstalledPackageInfo)
         <- either (die' verbosity) return $
               combinedConstraints (configConstraints cfg)
                                   (configDependencies cfg)
@@ -864,7 +864,7 @@ dependencySatisfiable
     -> PackageName
     -> InstalledPackageIndex -- ^ installed set
     -> Map PackageName (Maybe UnqualComponentName) -- ^ internal set
-    -> Map PackageName InstalledPackageInfo -- ^ required dependencies
+    -> Map (PackageName, ComponentName) InstalledPackageInfo -- ^ required dependencies
     -> (Dependency -> Bool)
 dependencySatisfiable
   use_external_internal_deps
@@ -884,7 +884,7 @@ dependencySatisfiable
         -- Except for internal deps, when we're NOT per-component mode;
         -- those are just True.
         then True
-        else depName `Map.member` requiredDepsMap
+        else (depName, CLibName) `Map.member` requiredDepsMap
 
     | isInternalDep
     = if use_external_internal_deps
@@ -1012,7 +1012,7 @@ configureDependencies
     -> UseExternalInternalDeps
     -> Map PackageName (Maybe UnqualComponentName) -- ^ internal packages
     -> InstalledPackageIndex -- ^ installed packages
-    -> Map PackageName InstalledPackageInfo -- ^ required deps
+    -> Map (PackageName, ComponentName) InstalledPackageInfo -- ^ required deps
     -> PackageDescription
     -> ComponentRequestedSpec
     -> IO [PreExistingComponent]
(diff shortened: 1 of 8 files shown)


Cabal/Distribution/Simple/Configure.hs
```diff
@@ -1397,13 +1397,18 @@ combinedConstraints constraints dependencies installedPackages = do
...
+      | (pkgname, cname, cid, Nothing) <- dependenciesPkgInfo ]
 
     dispDependencies deps =
       hsep [    text "--dependency="
-             <<>> quotes (disp pkgname <<>> char '=' <<>> disp cid)
-           | (pkgname, cid) <- deps ]
+             <<>> quotes
+                    (disp pkgname
+                    <<>> case cname of CSubLibName sublib -> ":" <<>> disp sublib
+                                       _ -> "" --XXX
```
fgaz: (A) I'm not sure about this. Maybe I should put error "impossible non-library dependecy" if I get something different than CLibName or CSubLibName. Or maybe i should add an entirely different datatype (probably not worth it).

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -645,9 +647,13 @@ configureOptions showOrParseArgs =
...
          "A list of exact dependencies. E.g., --dependency=\"void=void-0.5.8-177d5cdf20962d0581fe2e4932a6c309\""
          configDependencies (\v flags -> flags { configDependencies = v})
-         (reqArg "NAME=CID"
+         (reqArg "NAME[:COMPONENT_NAME]=CID"
                  (parsecToReadE (const "dependency expected") ((\x -> [x]) `fmap` parsecDependency))
-                 (map (\x -> display (fst x) ++ "=" ++ display (snd x))))
+                 (map (\(pn, cn, cid) ->
+                     display pn ++ "="
+                     ++ case cn of CSubLibName sublib -> ":" ++ display sublib
+                                   _ -> "" --XXX
```
fgaz: See (A)

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -256,7 +258,7 @@ data ConfigFlags = ConfigFlags {
     configStripLibs :: Flag Bool,      -- ^Enable library stripping
     configConstraints :: [Dependency], -- ^Additional constraints for
                                        -- dependencies.
-    configDependencies :: [(PackageName, ComponentId)],
+    configDependencies :: [(PackageName, ComponentName, ComponentId)],
```
fgaz: This is now a 3-tuple. Should it be a proper datatype? something like `data GivenDependency` or `GivenComponent`

23Skidoo: I'd say yes, if it's not too much of a trouble.

Cabal/Distribution/Simple/Configure.hs
ezyang: In this case, since this is "display" code, you should still print out the non-library component in the "obvious way". If you just return an empty "" or put an error it will make it more difficult to debug if you end up with something invalid here.

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -645,9 +647,13 @@ configureOptions showOrParseArgs =
...
          configDependencies (\v flags -> flags { configDependencies = v})
-         (reqArg "NAME=CID"
-                 (parsecToReadE (const "dependency expected") ((\x -> [x]) `fmap` parsecDependency))
-                 (map (\x -> display (fst x) ++ "=" ++ display (snd x))))
+         (reqArg "NAME[:COMPONENT_NAME]=CID"
+                 (parsecToReadE (const "dependency expected") ((\x -> [x]) `fmap` parsecGivenComponent))
+                 (map (\(GivenComponent pn cn cid) ->
+                     display pn ++ "="
+                     ++ case cn of CSubLibName sublib -> ":" ++ display sublib
+                                   _ -> "" --XXX
```
ezyang: Don't forget to update here too

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -3256,13 +3258,15 @@ setupHsConfigureFlags (ReadyPackage elab@ElaboratedConfiguredPackage{..})
...
-    configDependencies        = [ (case mb_cn of
-                                    -- Special case for internal libraries
-                                    Just (CSubLibName uqn)
-                                        | packageId elab == srcid
-                                        -> mkPackageName (unUnqualComponentName uqn)
-                                    _ -> packageName srcid,
-                                   cid)
+    configDependencies        = [ GivenComponent
+                                    (case mb_cn of
+                                       -- Special case for internal libraries
```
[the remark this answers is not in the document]
fgaz: Is it even possible without breaking some compatibility?

fgaz:
7.10+gold keeps OOMing and I don't have a windows machine to test the appveyor failure on :-/

fgaz:
Closing in favor of #5526. I included this change there.

PR #5421, opened 2018-07-09
Remove some old (and really, really old) compatibility code.

### LINKED ISSUE #2819: Drop IPI642 and IPI641 compatibility support
We're not running continuous integration tests on these ancient version of GHC, there's no reason to expect this code to actually work. We should get rid of it.

quasicomputational:
The GHC versions this code was supporting varied from 7.4, which has
only recently been dropped, and some code all the way back to 6.4.2!

None of this is being tested on a regular basis and it's highly likely
that our support for these old GHC versions is thoroughly broken
anyway.

This does not attempt to refactor HcPkgInfo and GhcImplInfo, which
both now contain information that we already know if we are in a
supported configuration.

Closes #2819.


typedrat:
Looking at [the release page](https://www.haskell.org/ghc/), we could be more aggressive if there's benefit. I dropped 7.4 because it was breaking, but 7.6 is also outside the window if we want to be aggressive about it.

quasicomputational:
I didn't think we could drop 7.6 until next April, which is 7.8's fifth birthday.

typedrat:
I was reading it as "five years since release" not "the most recent GHC release at any point in the last five years".

quasicomputational:
Also, I don't think there's a huge benefit to dropping 7.6, from what I saw looking through the code. It's got its share of random bodges, but so do most versions.

typedrat:
7.8 going is gonna be a big payoff, but we've got a while to wait for that.

quasicomputational:
Also - it's worth explicitly noting that I am reading the compatibility policy as limiting not just which GHCs Cabal will build with, but also which GHCs Cabal knows how to use as a compiler, which may be reading a bit too much into it. This PR is a mix of removing both kinds of support. If that's not appropriate, then (a) I'll change it, but (b) we need some kind of policy on dropping support for old GHCs as user code compilers.

quasicomputational:
Oh, and Travis has just informed me that we *are* testing back to 7.0, which resolves (c). That's nifty.

Cabal/Cabal.cabal
```diff
@@ -165,29 +165,18 @@ source-repository head
...
+    array      >= 0.4      && < 0.6,
+    base       >= 4.6      && < 5,
+    bytestring >= 0.10.0.0 && < 0.11,
+    containers >= 0.5.0.0  && < 0.7,
+    deepseq    >= 1.3      && < 1.5,
+    directory  >= 1.2      && < 1.4,
+    filepath   >= 1.3      && < 1.5,
+    pretty     >= 1.1.1    && < 1.2,
+    process    >= 1.1.0.2  && < 1.7,
+    time       >= 1.4      && < 1.10
```
phadej: 1.4.0.1

Cabal/Cabal.cabal
```diff
@@ -165,29 +165,18 @@ source-repository head
...
-    build-depends: directory >= 1.2      && < 1.4,
-                   process   >= 1.1.0.2  && < 1.7
+    array      >= 0.4      && < 0.6,
+    base       >= 4.6      && < 5,
+    bytestring >= 0.10.0.0 && < 0.11,
+    containers >= 0.5.0.0  && < 0.7,
+    deepseq    >= 1.3      && < 1.5,
+    directory  >= 1.2      && < 1.4,
+    filepath   >= 1.3      && < 1.5,
+    pretty     >= 1.1.1    && < 1.2,
```
phadej: 1.1.0.2?

Cabal/Cabal.cabal
```diff
@@ -165,29 +165,18 @@ source-repository head
...
-  else
-    build-depends: directory >= 1.2      && < 1.4,
-                   process   >= 1.1.0.2  && < 1.7
+    array      >= 0.4      && < 0.6,
+    base       >= 4.6      && < 5,
+    bytestring >= 0.10.0.0 && < 0.11,
+    containers >= 0.5.0.0  && < 0.7,
+    deepseq    >= 1.3      && < 1.5,
+    directory  >= 1.2      && < 1.4,
+    filepath   >= 1.3      && < 1.5,
```
phadej: 1.3.0.1... and so on.

Let's have stricttly what's with GHC-7.6: https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory

Cabal/Distribution/Simple/Build/PathsModule.hs
```diff
@@ -242,19 +224,10 @@ generate pkg_descr lbi clbi =
 
         paths_modulename = autogenPathsModuleName pkg_descr
 
-        get_prefix_stuff = get_prefix_win32 supports_cpp buildArch
+        get_prefix_stuff = get_prefix_win32
 
         path_sep = show [pathSeparator]
 
-        supports_cpp = supports_language_pragma
```
phadej: tnose have to stay. We can build `Cabal` only with GHC-7.6+, but we should support building with GHC-7.0+. (i.e. via `cabal-install`).

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -127,7 +126,7 @@ configure verbosity hcPath hcPkgPath conf0 = do
 
   (ghcProg, ghcVersion, progdb1) <-
     requireProgramVersion verbosity ghcProgram
-      (orLaterVersion (mkVersion [6,11]))
+      (orLaterVersion (mkVersion [7,6]))
```
phadej: Revert.

phadej:
@quasicomputational we test that `cabal-install` can build stuff with `GHC-7.0`, we should not drop that support.

Cabal/Cabal.cabal
quasicomputational: I only touched the bounds where I was specifically removing a CPP hack. AFAIK the rest don't strictly need bumping, but I guess we might as well.

Cabal/Distribution/Simple/GHC.hs
quasicomputational: I bumped it up to 7.0.1 (we test on Travis with 7.0.4, I have discovered). Doing that lets us simplify `PathsModule.hs` above. I can revert these if we really want to avoid changing this, but there should be no-one running an ancient development snapshot (right?).

phadej: ping @hvr.

quasicomputational: Github keeps hiding this section after I push. Let's see if replying will keep it open...

Cabal/Distribution/Simple/Build/PathsModule.hs
23Skidoo: OK.

23Skidoo: Fixed.

23Skidoo:
This was green before, and the only failure is the 7.10.3 OOM. Merging.

23Skidoo:
Merged, thanks!

PR #5432, opened 2018-07-13
Ignore case when matching SPDX identifiers

sol:
(as per the SPDX License List Matching Guidelines, v2.0)

https://spdx.org/spdx-license-list/matching-guidelines

### DIFF
--- a/Cabal/Distribution/SPDX/LicenseExceptionId.hs
+++ b/Cabal/Distribution/SPDX/LicenseExceptionId.hs
@@ -137,7 +137,7 @@ licenseExceptionName WxWindows_exception_3_1 = "WxWindows Library Exception 3.1"
 
 -- | Create a 'LicenseExceptionId' from a 'String'.
 mkLicenseExceptionId :: String -> Maybe LicenseExceptionId
-mkLicenseExceptionId s = Map.lookup s stringLookup
+mkLicenseExceptionId s = Map.lookup (map toLower s) stringLookup
 
 stringLookup :: Map String LicenseExceptionId
-stringLookup = Map.fromList $ map (\i -> (licenseExceptionId i, i)) $ [minBound .. maxBound]
+stringLookup = Map.fromList $ map (\i -> (map toLower (licenseExceptionId i), i)) $ [minBound .. maxBound]
--- a/Cabal/Distribution/SPDX/LicenseId.hs
+++ b/Cabal/Distribution/SPDX/LicenseId.hs
@@ -1479,7 +1479,7 @@ licenseIsOsiApproved ZPL_2_1 = False
 
 -- | Create a 'LicenseId' from a 'String'.
 mkLicenseId :: String -> Maybe LicenseId
-mkLicenseId s = Map.lookup s stringLookup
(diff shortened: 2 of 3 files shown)


sol:
This may actually also apply to `LicenRef` and license expressions in general, but I didn't see it explicitly mentioned in the matching guidelines.

quasicomputational:
Isn't this going to change the semantics of 2.2-spec files between Cabal 2.2 and Cabal 2.4? Then again, the worst that can happen is that 2.4 accepts a file that 2.2 doesn't; I guess we could add a `DistInexcusable` check for a 2.2-spec file with a lowercase license and that would be OK.

sol:
@quasicomputational hmm, I guess you have a point.  I'm looking into the check.

phadej:
I.e. I cannot write

```haskell
cabal-version: 2.2
name: pkg
version: 0

license: bsd-3-clause
```

That will cause missparse. At the very least it should be `cabal check`, that licenseid's for old `cabal-version` match case-exactly.

sol:
> This should be `cabal-version` feature guarded.
@phadej I think @quasicomputational already pointed that out, but thanks for the noise.

phadej:
In fact, I'm not sure we are reading right guidelines.

The https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60 spec appendix doesn't say anythign about case-sensitivity or insensitivity. Matching guidelines are about matching the *license texts* to the ones in SPDX license list. I'm very unsure it applies to the identifiers. I'll ask SPDX technical team for clarification

sol:
@phadej hmm, the matching guidelines explicitly mention "SPDX License List short identifiers":

> may match on the license or exception text itself, the official license header, or the SPDX License List short identifier

I'm not sure about license expressions in general (specifically `AND` and `OR`) and `LicenseRef`.

sol:
Related issue for Composer: https://github.com/composer/composer/issues/7039

phadej:
I found the issue: https://github.com/spdx/spdx-spec/issues/63 It's unsolved, and to me it's not clear which way it will turn out. So I'd rather keep `Cabal` as it is now.

Last comment says:

> Discussed on tech call on 6/5/2018: Need to respect the case for the license ID's since they translate to URI's in RDF. There are also other use cases that may break other parsers.

sol:
After I read the whole discussion at spdx/spdx-spec#63 I'm less inclined to move this forward.

Assuming https://github.com/spdx/spdx-spec/pull/37 gets merged in its current form, this is what we may end up with:

> Tools are encouraged, but not required, to take advantage of those commitments and support parsing license expressions which alter the canonical casing of any expression components.

23Skidoo:
@sol Feel free to close this PR if you've changed your mind.

pombredanne:
FWIW, this is Python so likely not your cup of tea.... but the right way when parsing or detecting SPDX expressions is to ignore case alright (and use the canonical case when producing them).
See https://github.com/nexB/license-expression/ and https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/match_spdx_lid.py

PR #5439, opened 2018-07-17
Preserve comments in lexer

pikajude:

For my tool `stylish-cabal`, I'd like to be able to preserve existing comments in a .cabal file.

This change passes all of the parser-related tests. Comments are retained up until the stuff in `Distribution.FieldGrammar`, where they get filtered out by `takeFields`.


pikajude:
Latest commit has some pretty major changes in the structure of the cabal file parser, namely that comments are only parsed before elements now - otherwise in something like this:

```
flag foo
  manual: True

-- a test-suite
test-suite bar
  main-is: bar.hs
```

the "a test-suite" comment will be parsed as part of the `manual` field value (since comments ignore indentation).

phadej:
Quick question: have you run Hackage-tests (accepting and roundtrip with pretty-printing)? What are before and after run times?

I don't like the use of `error`.

How GHC does comment handling for exactprint? I won't be surprised if they splice comments out of original source code I.e. don't have them in AST (and that we can do too, if we start to use sourcespans for ann)

phadej:
Another observation: comments in cabal file are always whole line comments. So you can extract them by pre-pass on source text.

This is refutable hypothesis, but if it holds then I don't think complicating lexer is worth for an optional feature (in fact it can be used to simplify the lexer)

pikajude:
@phadej you're absolutely right here, they could just be extracted alongside the other lexer steps without complicating everything. That's an approach I didn't think of since I've never tried to make a big change to a lexer before. Will push a commit later today using your suggestion.

PR #5440, opened 2018-07-17
Respect ghc-options and with-gcc when compiling C sources

syntheorem:
This fix allows users to override the C/C++ compiler, either by passing
--with-gcc to configure or adding -pgmc to ghc-options in their Cabal
file.

Tested manually with a project using C++ sources that require a more
recent version of clang than the default install (macOS). Inspected the
output of `Setup.hs build -v` to confirm that the correct arguments are
being passed to GHC when compiling the C++ sources.

Resolves: #4439 

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -70,6 +70,8 @@
     ([#5386](https://github.com/haskell/cabal/issues/5386)).
   * `Distribution.PackageDescription.Check.checkPackageFiles` now
     accepts a `Verbosity` argument.
+  * `ghc-options` and `--with-gcc` are now passed to GHC when compiling
+    C and C++ sources ([#4439](https://github.com/haskell/cabal/issues/4439)).
 
 ----
 
--- a/Cabal/Distribution/Simple/GHC/Internal.hs
+++ b/Cabal/Distribution/Simple/GHC/Internal.hs
@@ -293,7 +293,10 @@ componentCcGhcOptions verbosity _implInfo lbi bi clbi odir filename =
                                   NormalDebugInfo  -> ["-g"]
                                   MaximalDebugInfo -> ["-g3"]) ++
                                   PD.ccOptions bi,
-      ghcOptObjDir         = toFlag odir
+      ghcOptCcProgram      = maybeToFlag $ programPath <$>
+                                 lookupProgram gccProgram (withPrograms lbi),
+      ghcOptObjDir         = toFlag odir,
+      ghcOptExtra          = hcOptions GHC bi
     }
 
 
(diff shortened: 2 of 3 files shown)


angerman:
Shouldn’t this be `—with-cc`? Didn’t we already respect `CC` env vars?

lloydac:
> Shouldn’t this be `—with-cc`?

No, this is using the existing program replacement infrastructure, in which the C compiler is named gcc. So `--with-gcc` is the already existing option (stack uses it as well).

> Didn’t we already respect `CC` env vars?

Possibly, if GHC itself respects those env vars. The problem is that we were calling GHC to compile C files without explicitly passing `-pgmc`, so GHC didn't know about any `--with-gcc` arguments that were passed to `Setup configure`.

lloydac:
Since `--ghc-options` are always passed last to GHC, they should override any previous options. So the precedence would be:
1. `--ghc-options` on the command line
2. `ghc-options` in the `.cabal` file
3. `--with-gcc`

lloydac:
Sure, but that applies to any argument that Cabal passes to GHC. I think that `ghc-options` overriding everything is expected behavior.

Although maybe we want to warn upon ever seeing `-pgmc` in a `.cabal` file, since it seems questionable to configure it at all from there (because which program to use will depend on the system you're compiling on). Seems like a separate issue should be filed for that though.

lloydac:
FYI, this is the Travis CI failure, only in the `DEBUG_EXPENSIVE_ASSERTIONS` build:

>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

And this is the AppVeyor failure:

> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Not sure that these are actionable failures.

nh2:
What's holding this back?

It looks like a simple change.

ezyang:
This looks fine. Needs a rebase and merge conflict resolution but o/w ok to go in./

phadej:
... and test

gbaz:
@syntheorem sorry this languished. do you care to resolve the conflicts against master so we can proceed with it?

Mikolaj:
Hi! Heads up about https://github.com/haskell/cabal/pull/7873, which is this PR just rebased by @jberryman and we are desperate to merge it before it bitrots.

We need to add a small test for that or at least test the branch on the examples you have. Please help!

Mikolaj:
Please kindly also have a look (and review) at #7900, which provides a test and discusses its result. Last chance to provide feedback before we merge. Thank you!

Mikolaj:
@jberryman: thank you for #7900, which is now merged, which ends the saga! Please, everybody, test and reopen if anything doesn't work as expected.

PR #5443, opened 2018-07-18
Do not pass -js-lib-src to ghcjs

### LINKED ISSUE #5442: ghcjs with multiple js-sources broken
The fix to #4449 in https://github.com/haskell/cabal/commit/6148cea8749aad714db97f821b8141727a04bdab uncovered a bug in ghcjs, which I reported at
https://github.com/ghcjs/ghcjs/issues/678

I wonder if Cabal wants to create a command line that works around this GHCJS bug in affected versions of GHCJS.

nomeata:
because ghcjs does not like if the flag appears more than once, and it
seems it was happy before with the `jsSrc` files to be passed without a
flag.
This works-round https://github.com/ghcjs/ghcjs/issues/678 and
fixes #5442

### DIFF
--- a/Cabal/Distribution/Simple/GHCJS.hs
+++ b/Cabal/Distribution/Simple/GHCJS.hs
@@ -320,7 +320,7 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
                         ghcOptExtra =
                           [ "-link-js-lib"     , getHSLibraryName uid
                           , "-js-lib-outputdir", libTargetDir ] ++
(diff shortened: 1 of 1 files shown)


hvr:
I'm merging this now as it appears to strictly improve the situation , and we *have* to do *something*, as otherwise it GHCJS (e.g. from https://launchpad.net/~hvr/+archive/ubuntu/ghcjs ) chokes early on for packages like `ghcjs-base` and thus is *broken*; I hope @luite may provide us some insight at ghcjs/ghcjs#678 why the `-js-lib-src` flag was introduced in the first place.

PR #5454, opened 2018-07-24
Add --build-depends flag, associated support to new-repl

### LINKED ISSUE #5425: cabal new-repl without a project
Sometimes I just need to check a couple of things in GHCi, not specific to any project. I could run `ghci`, but then I don't get access to any packages except the ones that are installed globally.

With `stack` I can run `stack ghci --package lens`, for instance.

With `cabal` I have to create something like this:

```
cabal-version: 2.2
name: global
version: 0

library
  build-depends: base, lens
  default-language: Haskell2010
```

and then run `cabal new-repl` in a directory with this `.cabal`-file.

I'd prefer to write `cabal new-repl --package lens`:

* it would find a suitable build plan for the specified packages and launch GHCi in this environment
* it would not create `dist-newstyle` in the current directory, instead it could use `~./cabal/dist-global` or similar

typedrat:
Closes #5425. Adds `--package` flag, allows `new-repl` use outside of project folders.

### DIFF
--- a/cabal-install/Distribution/Client/CmdRepl.hs
+++ b/cabal-install/Distribution/Client/CmdRepl.hs
@@ -119,7 +119,7 @@ defaultEnvFlags = EnvFlags
 
 envOptions :: ShowOrParseArgs -> [OptionField EnvFlags]
 envOptions _ =
-  [ option ['p'] ["package"]
+  [ option ['b'] ["build-depends"]
     "Include an additional package in the environment presented to GHCi."
     envPackages (\p flags -> flags { envPackages = p ++ envPackages flags })
     (reqArg "DEPENDENCY" dependencyReadE (fmap prettyShow :: [Dependency] -> [String]))
@@ -127,7 +127,7 @@ envOptions _ =
     "Don't automatically include transitive dependencies of requested packages."
     envIncludeTransitive (\p flags -> flags { envIncludeTransitive = p })
     falseArg
-  , option ['z'] ["only-specified"]
+  , option ['z'] ["ignore-project"]
     "Only include explicitly specified packages (and 'base'). This implies '--no-transitive-deps'."
     envOnlySpecified (\p flags -> flags { envOnlySpecified = p, envIncludeTransitive = not <$> p})
     trueArg
--- a/cabal-install/Distribution/Client/CmdRepl.hs
+++ b/cabal-install/Distribution/Client/CmdRepl.hs
@@ -30,8 +30,7 @@ import qualified Distribution.Client.InstallPlan as InstallPlan
 import Distribution.Client.ProjectBuilding
          ( rebuildTargetsDryRun, improveInstallPlanWithUpToDatePackages )
 import Distribution.Client.ProjectConfig
-         ( ProjectConfig(..), BadPackageLocations(..), BadPackageLocation(..)
-         , ProjectConfigProvenance(..)
+         ( ProjectConfig(..), withProjectOrGlobalConfig
          , projectConfigConfigFile, readGlobalConfig )
 import Distribution.Client.ProjectOrchestration
 import Distribution.Client.ProjectPlanning 
@@ -41,8 +40,6 @@ import Distribution.Client.RebuildMonad
 import Distribution.Client.Setup
          ( GlobalFlags, ConfigFlags(..), ConfigExFlags, InstallFlags )
 import qualified Distribution.Client.Setup as Client
-import Distribution.Client.TargetSelector
-         ( TargetSelector(..), TargetImplicitCwd(..), ComponentKind(..) )
 import Distribution.Client.Types
          ( PackageLocation(..), PackageSpecifier(..), UnresolvedSourcePackage )
 import Distribution.Simple.Setup
@@ -78,7 +75,7 @@ import Distribution.Types.PackageDescription
 import Distribution.Types.Library
          ( Library(..), emptyLibrary )
 import Distribution.Types.PackageId
-         ( PackageIdentifier(..), PackageId )
+         ( PackageIdentifier(..) )
 import Distribution.Types.Version
          ( mkVersion, version0 )
 import Distribution.Types.VersionRange
@@ -92,10 +89,6 @@ import Distribution.Simple.Utils
 import Language.Haskell.Extension
          ( Language(..) )
 
-import Control.Exception
-         ( catch, throwIO )
-import Control.Monad 
-         ( when, unless )
 import Data.List
          ( (\\) )
 import qualified Data.Map as Map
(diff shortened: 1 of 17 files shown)


cabal-install/Distribution/Client/Types.hs
```diff
@@ -236,7 +236,7 @@ data PackageSpecifier pkg =
      -- | A fully specified source package.
      --
    | SpecificSourcePackage pkg
-  deriving (Eq, Show, Generic)
+  deriving (Eq, Show, Functor, Generic)
```
23Skidoo: I'd rather prefer if we used an explicit lambda with case analysis or a `mapSpecificSourcePackage` helper here. Code like `fmap fmap` is not super readable.

cabal-install/Distribution/Client/CmdRepl.hs
```diff
@@ -13,34 +17,134 @@ module Distribution.Client.CmdRepl (
...
+  [ option ['p'] ["package"]
+    "Include an additional package in the environment presented to GHCi."
+    envPackages (\p flags -> flags { envPackages = p ++ envPackages flags })
+    (reqArg "PACKAGE" packageIdReadE (fmap prettyShow :: [PackageId] -> [String]))
+  , option [] ["no-transitive-deps"]
+    "Don't automatically include transitive dependencies of requested packages."
+    envIncludeTransitive (\p flags -> flags { envIncludeTransitive = p })
+    falseArg
+  , option ['z'] ["only-specified"]
+    "Only include explicitly specified packages (and 'base'). This implies '--no-transitive-deps'."
```
23Skidoo: Description doesn't make it clear how `--only-specified` differs from `--no-transitive-deps`. I guess we should mention environment files somewhere.

cabal-install/Distribution/Client/CmdRepl.hs
```diff
@@ -100,62 +212,223 @@ replCommand = Client.installCommand {
...
+          without
+        else
+          catch with
+            $ \case
+              (BadPackageLocations prov locs) 
+                | prov == Set.singleton Implicit
+                , let 
+                  isGlobErr (BadLocGlobEmptyMatch _) = True
+                  isGlobErr _ = False
+                , any isGlobErr locs ->
```
23Skidoo: Code golf: can also be written as `notNull [ () | BadLocGlobEmptyMatch _ <- locs ]`.

cabal-install/Distribution/Client/CmdRepl.hs
```diff
@@ -100,62 +212,223 @@ replCommand = Client.installCommand {
...
+
+  baseCtx' <- establishDummyProjectBaseContext
+            verbosity
+            config
+            tempDir
+            [SpecificSourcePackage sourcePackage]
+
+  let
+    targetSelectors = [TargetPackage TargetExplicitNamed [pkgId] Nothing]
+    finalizer = return () --handleDoesNotExist () (removeDirectoryRecursive tempDir)
```
23Skidoo: Hmm, so `finalizer` is `return ()` in both cases?

cabal-install/Distribution/Client/CmdRepl.hs
```diff
@@ -100,62 +212,223 @@ replCommand = Client.installCommand {
...
+          ++ "use 'repl'."
+
+    baseCtx' <- if null (envPackages envFlags)
+      then return baseCtx
+      else
+        -- Unfortunately, the best way to do this is to let the normal solver
+        -- help us resolve the targets, but that isn't ideal for performance,
+        -- especially in the no-project case.
+        withInstallPlan (lessVerbose verbosity) baseCtx $ \elaboratedPlan -> do
+          -- Interpret the targets on the command line as repl targets
```
23Skidoo: The `targets <- ...` and the `Reject multiple targets...` bits look like they could be moved to a helper function to avoid copy-paste.

cabal-install/Distribution/Client/CmdRepl.hs
```diff
@@ -13,34 +17,134 @@ module Distribution.Client.CmdRepl (
...
+
+envOptions :: ShowOrParseArgs -> [OptionField EnvFlags]
+envOptions _ =
+  [ option ['p'] ["package"]
+    "Include an additional package in the environment presented to GHCi."
+    envPackages (\p flags -> flags { envPackages = p ++ envPackages flags })
+    (reqArg "PACKAGE" packageIdReadE (fmap prettyShow :: [PackageId] -> [String]))
+  , option [] ["no-transitive-deps"]
+    "Don't automatically include transitive dependencies of requested packages."
+    envIncludeTransitive (\p flags -> flags { envIncludeTransitive = p })
```
23Skidoo: I don't see how `envIncludeTransitive` is used.

cabal-install/Distribution/Client/CmdRepl.hs
typedrat: They differ if you're in a project/package context, `--only-specified` ignores that context.

cabal-install/Distribution/Client/CmdRepl.hs
typedrat: Nope, that's just for debugging. This is a WIP.

cabal-install/Distribution/Client/CmdRepl.hs
typedrat: Because I haven't added code to include the transitive dependencies yet. :)

cabal-install/Distribution/Client/CmdRepl.hs
23Skidoo: Yep, I figured, but it's not obvious from the description. Maybe we should call it `--ignore-project`?

hvr: I'm against introducing a new flag for something we can already express with current flags (especially because it makes things awkward when such competing/conflciting flags get turned into cabal.config/cabal.project flags):

We already have `--project-file=`, and I've been arguing that we can just have `--project-file=''` denote to ignore any project context; i.e. force a project-less context.

hvr:
I briefly discussed the naming choice `--package` with @typedrat, and I'm convinced that it's a misnomer; the better naming would be something akin to `--build-depends`, as a) we care only about the library components, and b) we want consistency w/ with pre-existing terms, in this case from the `.cabal` file; c) when we extend the syntax of `build-depends` to specify public multi-libs packages, we'd increase the level of misnaming even more, as some multi-lib (I have at least one in mind which will be) that will be designed in a way to deliberate clash with each other, and so you *must* be able to select the scope at the component level. d) reusing the `build-depends` syntax is more convenient as you can specify the version constraints syntax directly w/o being forced to resort to `--constraints`.

typedrat:
For some reason, outside of packages, this is failing with an error about being unable to find `base`.

PR #5461, opened 2018-07-27
Disambiguate selectors by the command being used

### LINKED ISSUE #4676: cabal new-run foo should default to exe:foo (same for test and bench)
If I have a project `foo.cabal` that defines both a library and an executable named `foo`, which is a common situation, then currently
```
$ cabal new-run foo -- some args
```
complains that it does not know whether I want to run `foo` the library or foo the executable, and makes me write
```
$ cabal new-run exe:foo -- some args
```
I see how this behavior makes sense for `cabal new-build` etc, but I do not even know what it means to `cabal new-run` a library. And even if that makes sense, I believe the ambiguity should just be resolved to the executable by default.

typedrat:
Closes #4676.

### DIFF
--- a/cabal-install/Distribution/Client/TargetSelector.hs
+++ b/cabal-install/Distribution/Client/TargetSelector.hs
@@ -84,7 +84,8 @@ import qualified Data.Map.Lazy   as Map.Lazy
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import Control.Arrow ((&&&))
-import Control.Monad
+import Control.Monad 
+  hiding ( mfilter )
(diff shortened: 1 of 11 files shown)


cabal-install/Distribution/Client/TargetSelector.hs
```diff
@@ -201,20 +202,22 @@ instance Binary SubComponentTarget
 -- the available packages (and their locations).
 --
 readTargetSelectors :: [PackageSpecifier (SourcePackage (PackageLocation a))]
+                    -> Maybe ComponentKindFilter
```
23Skidoo: Please add a comment here explaining what this parameter does.

PR #5471, opened 2018-07-30
Bump dependencies

phadej:


phadej:
7.10.3 also fails in master with no memory to compile TreeDiff instances :(

PR #5474, opened 2018-07-30
outdated: accept --project-file

quasicomputational:
This doesn't get anywhere near the improvements suggested in #4831,
but it's a very respectable improvement over the status-quo for not
much effort.

New test added for this.

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -1336,10 +1336,12 @@ The following flags are supported by the ``outdated`` command:
 ``--new-freeze-file``
     Read dependency version bounds from the new-style freeze file
     (``cabal.project.freeze``) instead of the package description file.
+    This is equivalent to ``--project-file cabal.project``.
 ``--project-file`` *PROJECTFILE*
     Read dependendency version bounds from the new-style freeze file
     related to the named project file (i.e., ``$PROJECTFILE.freeze``)
-    instead of the package desctription file.
+    instead of the package desctription file. If multiple ``--project-file``
+    flags are provided, only the final one is considered.
 ``--simple-output``
     Print only the names of outdated dependencies, one per line.
 ``--exit-code``
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -1338,11 +1338,13 @@ The following flags are supported by the ``outdated`` command:
     (by default, ``cabal.project.freeze``) instead of the package
     description file.
 ``--project-file`` *PROJECTFILE*
+    :since: 2.4
+
     Read dependendency version bounds from the new-style freeze file
     related to the named project file (i.e., ``$PROJECTFILE.freeze``)
     instead of the package desctription file. If multiple ``--project-file``
     flags are provided, only the final one is considered. This flag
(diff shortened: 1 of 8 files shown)


Cabal/doc/developing-packages.rst
```diff
@@ -1336,6 +1336,10 @@ The following flags are supported by the ``outdated`` command:
 ``--new-freeze-file``
     Read dependency version bounds from the new-style freeze file
     (``cabal.project.freeze``) instead of the package description file.
+``--project-file`` *PROJECTFILE*
```
hvr: How does this interact w/ the `--new-freeze-file` flag?

quasicomputational: Last one wins (or it should be). I'll add a test and make sure it's explicitly mentioned in the docs.

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -1140,9 +1140,14 @@ outdatedCommand = CommandUI {
...
 
     ,option [] ["new-freeze-file"]
-     "Act on the new-style freeze file"
-     outdatedNewFreezeFile (\v flags -> flags { outdatedNewFreezeFile = v })
-     trueArg
+     "Act on the new-style freeze file named cabal.project.freeze"
+     outdatedProjectFile (\_ flags -> flags { outdatedProjectFile = pure "cabal.project" })
+     (noArg mempty)
+
+    ,option [] ["project-file"]
```
hvr: Something doesn't smell right to me about duplicating the global `--project-file=` flag as a local `outdated` sub-command flag. What does this make possible that wasn't possible before?

quasicomputational: I'm sorry, I don't understand this question. Where is the global `--project-file` flag being defined? `outdated` doesn't currently accept it:

```
$ cabal outdated --project-file=foo.project
cabal: unrecognized 'outdated' option `--project-file=foo.project'
$ cabal --version
cabal-install version 2.2.0.0
compiled using version 2.2.0.1 of the Cabal library 
```

Bear in mind that `outdated` isn't really a new-style command: it's more like a v1 command that happens to know a little bit about new-style freeze files. `new-outdated` as outlined in #4831 might be a worthy project, but that's a much bigger task. I think that this patch has a good power-to-weight ratio.

hvr: Oh you're right, I stand corrected. I confused `--project-file=` to be a global one (like `--store-dir=`)

Now this is starting to make more sense. But if we're going to migrate to make the `new-*`-style commands default, then it's also kinda strange to have `--project-file=` imply a side-effect; rather than have it act like merely setting a property, and requiring an explicit `--new-freeze-file`; that would feel more consistent to me w/ how `--project-file=` usually acts.

quasicomputational: I've gone ahead and implemented that. In the future we'll probably want to remove `--new-freeze-file` outright; I've made sure that `--project-file` without it errors out, so that we've got freedom to make that change.

23Skidoo:
Bikeshedding: should `--new-freeze-file` be called `--v2-freeze-file` in light of #5429?

PR #5476, opened 2018-07-31
[RFC] Expose cabal-install as a library.

### LINKED ISSUE #1597: Cabal-install as a library

23Skidoo:
This has been requested multiple times in the past, and now that the releases of Cabal and cabal-install are more frequent, the downsides are IMO not as big as before.

To make the maintenance easier, I propose that we don't require `@since` annotations in lib:cabal-install and don't go out of our way to support clients that want to be compatible with multiple major releases of lib:cabal-install.

Additionally, this simplifies `cabal-install.cabal` a bit by removing the `lib` flag.

Fixes #1597.

---

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -121,11 +121,6 @@ Flag debug-tracetree
   default:      False
   manual:       True
 
-flag lib
-  description:  Build cabal-install as a library. Please only use this if you are a cabal-install developer.
-  Default:      False
-  manual:       True
-
 -- Build everything (including the test binaries) as a single static binary
 -- instead of 5 discrete binaries.
 -- This is useful for CI where we build our binaries on one machine, and then
@@ -352,9 +347,6 @@ library
       cpp-options: -DDEBUG_TRACETREE
       build-depends: tracetree >= 0.1 && < 0.2
 
-    if !flag(lib)
-      buildable: False
-
     default-language: Haskell2010
 
 executable cabal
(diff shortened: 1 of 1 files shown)


23Skidoo:
> we do not maintain any API-level changelog;

Yes, this shouldn't be required. 

> and are free to make backward compatible changes w/o major ver bumps)?

I guess you meant "incompatible". No, we will support clients that set e.g. `build-depends: cabal-install == 2.4.*`. The idea is that if the major releases are more frequent, we will need to do less minor ones. If we desperately need to backport a API-breaking change for `lib:cabal-install`, we can do a major release of `cabal-install` only. We can advance `lib:Cabal` version numbers by bigger increments to keep them in sync with `cabal-install`.

phadej:
`cabal-install` would be bad library breaking all the clients. I don't want to consider backwards compatibility at all.

Maybe expose parts though (public) sub-libraries

quasicomputational:
I think this is a good idea. Having access to the solver and to non-ad-hoc rendering & parsing of configuration files are both fairly important to the wider tooling ecosystem. If compatibility breaks, it's better to find out at compile time than at runtime since users are having to parse `--dry-run` output to get at the solver (true story).

`lib:ghc` changes something every major version (`doctest` has the CPP to prove it), but having access to the guts of it is so valuable that a little bit of adaptation is fine; `cabal-install` isn't a boot library, so users might not even need to break the CPP out.

Also - how often do `cabal-install` releases from old stable branches happen, anyway? The only scenario I can imagine where we'd want to is when we've got a terrible bug affecting a recent release that was the last one before a significant behavioural change (so we can't expect users to upgrade major versions). That seems very much like the exception, not the rule.

hvr:
> You'll only need to think about it when backporting a patch to a stable branch.

in other words, we won't bother backporting anymore ;-)

23Skidoo:
@quasicomputational 
> Also - how often do cabal-install releases from old stable branches happen, anyway?

As you can see from http://hackage.haskell.org/package/cabal-install, 1.18, 1.20, and 1.22 had quite a few minor releases. But then the gaps between major releases were much longer.

@hvr 

> in other words, we won't bother backporting anymore ;-)

Yeah, the idea is basically to not make minor releases unless we absolutely have to, but instead cut major releases more frequently.

> which means we should reenable the warning in cabal which reminds users to upgrade as soon as a newer cabal version is available in the index!

Well, we want to do this anyway. Can't find the ticket for some reason.

> It'd take an intentionally designed library design for me to consider depending on it, rather than some arbtirary kitchen sink API which wasn't designed for public use by anyone else other than cabal-install.

See this as a (tiny) step towards that goal. Next steps would be to split the pure core out of `lib:Cabal` and the solver out of `lib:cabal-install`. Maybe we can achieve all of this in 3.0. Need to make a decision about the version scheme, more frequent release schedule should make that easier as well.

BardurArantsson:
Beware Hyrum's law.

IMO, it would be much better to start with e.g. the solver (which is *moderately* stable, interface-wise I guess?) to try to gain a little experience in what expectations users will come to have.

It's one thing to say "we guarantee nothing", but it's quite another to actually stick to that when users are screaming at you...

23Skidoo:
> It's one thing to say "we guarantee nothing", but it's quite another to actually stick to that when users are screaming at you...

Seems to work okay for GHC API. But we should clearly communicate our API stability guarantees to the users.

emilypi:
Done as of the 3.6 release.

PR #5478, opened 2018-07-31
Set abiDepends field in IPI to []

### LINKED ISSUE #5465: Silence GHC 8.4's "ignoring (possibly broken) abi-depends field for packages" warning
GHC 8.4 started emitting the following noisy warning whenever a package registration (IPI) includes a `abi-depends:` field (see https://ghc.haskell.org/trac/ghc/ticket/14381 for more details) cluttering the output:

```
Building cabal-doctest-1.0.6 (lib)...
ignoring (possibly broken) abi-depends field for packages
Configuring reflection-2.1.3 (lib)...
ignoring (possibly broken) abi-depends field for packages
Configuring semigroups-0.18.4 (lib)...
Building reflection-2.1.3 (lib)...
ignoring (possibly broken) abi-depends field for packages
```

This affects GHC 8.4.3 (but it probably won't affect GHC 8.4.4); since `abi-depends` is ignored loudly  we ought to simply omit populating this field for GHCs which noisy warn about it being ignored anyway.

More generally, as it's effectively redundant (it would mostly provide an additional check in case cabal screws up the nix-style hash computation) and some distributions started to patch GHC 8.2.2; so we might as well never populate `abi-depends`.

vrom911:
Resolves #5465

I have tried to `new-build` one personal package, and the output was without any `abi-depends` related warnings. However, when I was using `new-install` on the same package, the warnings were still printed out. So this is not the complete fix of the task, but I don't know what can be done about that.

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -80,6 +80,8 @@
   * Added a parameter to
     `Distribution.Backpack.ConfiguredComponent.toConfiguredComponent` in order to fix
     [#5409](https://github.com/haskell/cabal/issues/5409).
+  * Partially silence `abi-depends` warnings
+    ([#5465](https://github.com/haskell/cabal/issues/5465)).
 
 ----
 
--- a/Cabal/Distribution/Simple/Register.hs
+++ b/Cabal/Distribution/Simple/Register.hs
@@ -437,7 +437,7 @@ generalInstalledPackageInfo adjustRelIncDirs pkg abi_hash lib lbi clbi installDi
     IPI.includeDirs        = absinc ++ adjustRelIncDirs relinc,
     IPI.includes           = includes bi,
     IPI.depends            = depends,
-    IPI.abiDepends         = abi_depends,
+    IPI.abiDepends         = [], -- due to #5465
     IPI.ccOptions          = [], -- Note. NOT ccOptions bi!
                                  -- We don't want cc-options to be propagated
                                  -- to C compilations in other packages.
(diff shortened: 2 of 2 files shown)


vrom911:
@23Skidoo , I rebased on master and now I have troubles with building the project after the version bump. I've noticed bunch of `Cabal .. < 2.4` conditions in `Cabal/Cabal.cabal` file. As I see it causes issues on master as well.

hvr:
@vrom911 I've just pushed 46d486f196606bac0e51ff7bfa304f5396072385 which should mitigate the over-constraining

vrom911:
Thanks! Rebased.

23Skidoo:
@ezyang, can you please approve this fix?

hvr:
@23Skidoo note that we should backport this one to the next lib:Cabal-2.2 point rls, if we ever do one

ezyang:
As long as this has been tested against versions of GHC which respected the field, I'm OK with it.

Cabal/ChangeLog.md
```diff
@@ -80,6 +80,8 @@
   * Added a parameter to
     `Distribution.Backpack.ConfiguredComponent.toConfiguredComponent` in order to fix
     [#5409](https://github.com/haskell/cabal/issues/5409).
+  * Partially silence `api-depends` warnings
```
23Skidoo: s/api-depends/abi-depends/

ezyang:
I don't expect problems, but if there was a problem, GHC 8.2 is where I would expect it to manifest. So if it's been tested, all good by me.

hvr:
alright then; then I'll merge this and thus extend the testing user-base to our early adopters... :-)

23Skidoo:
Cherry-picked to `2.4`.

PR #5480, opened 2018-07-31
outdated: accept v1/v2 aliases for freeze-file flags

quasicomputational:
As suggested in #5474.

---

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -1166,12 +1166,12 @@ outdatedCommand = CommandUI {
     optionVerbosity outdatedVerbosity
       (\v flags -> flags { outdatedVerbosity = v })
 
-    ,option [] ["freeze-file"]
+    ,option [] ["freeze-file", "v1-freeze-file"]
      "Act on the freeze file"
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Setup.hs
```diff
@@ -1166,12 +1166,12 @@ outdatedCommand = CommandUI {
...
       (\v flags -> flags { outdatedVerbosity = v })
 
-    ,option [] ["freeze-file"]
+    ,option [] ["freeze-file", "v1-freeze-file"]
      "Act on the freeze file"
      outdatedFreezeFile (\v flags -> flags { outdatedFreezeFile = v })
      trueArg
 
-    ,option [] ["new-freeze-file"]
+    ,option [] ["new-freeze-file", "v2-freeze-file"]
```
phadej: I'd rather make freeze files state their format than rely on user to remember it right.

23Skidoo:
LGTM.

cabal-install/Distribution/Client/Setup.hs
23Skidoo: Freeze file format is actually the same for v1 and v2, but the file names are different.

23Skidoo:
Merged, thanks!

PR #5482, opened 2018-07-31
Split out cabal-solver as a copy of Distribution.Solver from cabal-in…

phadej:
…stall

---

### DIFF
--- a/cabal-solver/LICENSE
+++ b/cabal-solver/LICENSE
@@ -1,0 +1,31 @@
+Copyright (c) 2003-2017, Cabal Development Team.
+See the AUTHORS file for the full list of copyright holders.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+    * Neither the name of Isaac Jones nor the names of other
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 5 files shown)


phadej:
To clarify this PoC, it took me 15 minutes to assemble. I'd really like to see a single downstream which would use this, or some other part of `cabal-install` (they can do it too, add `cabal-lib-solver` as an internal lib). We have bad experience where people asked for the feature, but then didn't use it - yet there's a maintenance burden for us.

So I'd really like to see realistic experiments where this - `cabal-lib-solver`, or refactored out `Types` (for config reading/writing) or something else is used for great benefit outside `cabal-install`.

I want to see existential motivation, not vague "it would be nice".

To be clear: I don't think this is good idea either. Public multi-libs is better idea, when `cabal-install` will be able to use them for itself.

23Skidoo:
It'll be a while until we can use multilibs, the cabal-lib- split can be done in 3.0.

phadej:
Isn't splitting Cabal something we can only reasonably do when there are no need to use Custom setups?

quasicomputational:
On concrete use-cases, I have a couple of day-one candidates:

* Implementing something like Stack's `pvp-bounds` on top of freeze files: adding constraints to the `.cabal` files based on the freeze files' constraints, on the basis that then you're only declaring compatibility with what's been actually tested. This needs to be able to read the freeze files. I actually have a prototype of this tool (it's not complicated), but I don't know if that'll meet the "great benefit" threshold.

* Stack would like to use the solver. commercialhaskell/stack#1615 is the bug over there about it.

* There are [at least two efforts](https://github.com/sboosali/project2nix) to [convert `cabal.project` to Nix derivations](https://github.com/NixOS/cabal2nix/issues/286). Note that that's first one is already using `cabal-install` as a library.

Less concretely or with less great benefit:

* In a multi-package project, if some of the packages don't work with some GHC versions but you'd like to test the rest, you need multiple `.project` files. I'm currently generating them with ad-hoc code but I'd like to use the proper renderer.

* Any static analysis tool would benefit from being able to read `cabal.project` and discover all the packages in the project. Imagine pointing hlint at your `cabal.project` instead of having to direct it to each package. Along the same lines, IDE integration: discovering local packages so that jump-to-definition can work without manual set-up.

23Skidoo:
Re: packages with custom setups, they must update setup-depends anyway for the new release, and we can also add a compatibility shim lib:Cabal that re-exports cabal-lib and cabal-lib-build with a deprecation warning.

phadej:
> if some of the packages don't work with some GHC versions but you'd like to test the rest, you need multiple .project files.

https://github.com/phadej/darkmatter/blob/master/src/DarkMatter.hs renders 
https://github.com/phadej/acme-kmett/blob/master/cabal.darkmatter into
- https://github.com/phadej/acme-kmett/blob/master/cabal.7.6.3.project
- https://github.com/phadej/acme-kmett/blob/master/cabal.7.8.4.project
- https://github.com/phadej/acme-kmett/blob/master/cabal.8.2.2.project
- ...

It's "ad-hoc" approach, but very robust one. `cabal-install` would use something very similar.

PR #5487, opened 2018-08-01
Extend packages in plan.json with pkg-src field

alexbiehl:
This PR extends packages in the `plan.json` file with a `pkg-src` field. This is really helpful for third-party tooling to find the paths to all the project local packages including source repositories and such.

/cc @hvr @shlevy

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanOutput.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanOutput.hs
@@ -211,8 +211,7 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
 
       sourceRepoToJ :: PD.SourceRepo -> J.Value
       sourceRepoToJ PD.SourceRepo{..} =
-        J.object [ "kind"     J..= jdisplay repoKind
-                 , "type"     J..= fmap jdisplay repoType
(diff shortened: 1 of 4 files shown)


cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -168,6 +169,59 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
...
+                     , "path" J..= J.String repoLocalDir
+                     ]
+          RepoRemote{..} ->
+            J.object [ "type"   J..= J.String "remote-repo"
+                     , "path"   J..= J.String repoLocalDir
+                     , "remote" J..= J.String (remoteRepoName repoRemote)
+                     ]
+          RepoSecure{..} ->
+            J.object [ "type"   J..= J.String "secure-repo"
+                     , "path"   J..= J.String repoLocalDir
```
hvr: do we really need the `"path"`? I don't feel like that's a relevant information; we already know the "remote" label; also the local repo cache path is to be considered an implementation detail; its on-disk format may easily change at some point to account for situations we can't currently represent properly yet.

alexbiehl: yeah, let's remove that one.

PR #5489, opened 2018-08-01
Try to fix AppVeyor (2.4).

23Skidoo:
Like #5486, but for 2.4. See https://github.com/haskell/cabal/commit/123291c0939789aa7c3e4970c4a22adf50c0a9ee#comments.

---

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,9 @@
 install:
   # Using '-y' and 'refreshenv' as a workaround to:
   # https://github.com/haskell/cabal/issues/3687
-  - choco install -y cabal
-  - choco install -y ghc --version 8.0.2
+  - choco source add -n mistuke -s https://www.myget.org/F/mistuke/api/v2
+  - choco install -y ghc --version 8.0.2 --ignore-dependencies
+  - choco install -y cabal-head -pre
   - refreshenv
   # See http://help.appveyor.com/discussions/problems/6312-curl-command-not-found#comment_42195491
   # NB: Do this after refreshenv, otherwise it will be clobbered!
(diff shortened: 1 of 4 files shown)


23Skidoo:
This is what seems to be the problem:

```
[__1] rejecting: cabal-testsuite:setup.Cabal-2.4.0.0 (constraint from maximum
version of Cabal used by Setup.hs requires <2.4)
```

This constraint comes from `setupMaxCabalVersionConstraint` in [`D.C.ProjectPlanning`](https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/ProjectPlanning.hs):

```
    -- As we can't predict the future, we also place a global upper
    -- bound on the lib:Cabal version we know how to interact with:
    --
    -- The upper bound is computed by incrementing the current major
    -- version twice in order to allow for the current version, as
    -- well as the next adjacent major version (one of which will not
    -- be released, as only "even major" versions of Cabal are
    -- released to Hackage or bundled with proper GHC releases).
    --
    -- For instance, if the current version of cabal-install is an odd
    -- development version, e.g.  Cabal-2.1.0.0, then we impose an
    -- upper bound `setup.Cabal < 2.3`; if `cabal-install` is on a
    -- stable/release even version, e.g. Cabal-2.2.1.0, the upper
    -- bound is `setup.Cabal < 2.4`. This gives us enough flexibility
    -- when dealing with development snapshots of Cabal and cabal-install.
    --
    setupMaxCabalVersionConstraint =
      alterVersion (take 2) $ incVersion 1 $ incVersion 1 cabalVersion
```

On Travis, we are using `cabal-install-head` and apparently don't bump into this constraint (though on HEAD/2.5, I actually think that we should -- not sure what's going here). So we should upgrade `cabal-install` on AppVeyor to a fresher snapshot, and maybe add an option for `new-build` to ignore the `maxCabalVersionConstraint`.

hvr:
@23Skidoo btw, @Mistuke's https://hub.zhox.com/posts/chocolatey-introduction/ describes how to get access to `cabal-head` which correspond to `cabal-install-head`

23Skidoo:
@Mistuke How do I install a combination of a stable GHC and a pre-release `cabal` with `choco`? No matter what I try, stable `ghc` pulls in stable `cabal`, which overrides the pre-release one.

23Skidoo:
Figured it out: just install `cabal-head` after `ghc`. Still, the fact that `ghc` pulls in unneeded stable `cabal` is mildly annoying.

23Skidoo:
Now there's a test that needs to be fixed...

Mistuke:
@23Skidoo `choco install ghc -y --ignore-dependencies` will skip the stable `cabal` install.

23Skidoo:
Yay, looks like it worked. BTW, we seem to have a test that fails intermittently -- `CustomPlain`.

PR #5493, opened 2018-08-02
Fix x-package references with --haddock-for-hackage

alexbiehl:
Herbert noticed that with `--haddock-for-hackage` we were passing strange `file://..` urls to the html directories. It soon showed that I had screwed up in an earlier PR.


hvr:
I've merged this to `master` and `2.4` now to increase the likelihood @bgamari will pick it up for the upcoming GHC 8.6.1beta...

PR #5496, opened 2018-08-03
`--haddock-for-hackage` works with older Haddock

### LINKED ISSUE #5494: `--haddock-for-hackage` fails hard with
Currently, when I run `new-haddock` with e.g. GHC 8.2 or older it fails:

```
$ cabal new-haddock --haddock-for-hackage
Resolving dependencies...
 - HsYAML-0.1.1.2 (lib) (first run)
Configuring library for HsYAML-0.1.1.2..
cabal: haddock prior to 2.19 does not support the --quickjump flag.

Warning: Failed to build documentation for HsYAML-0.1.1.2.
```

I assume this is because (haven't looked at the code), `--haddock-for-hackage` *unconditionally* turns on `--quickjump`, which fails with an error for too old haddocks, and rightly so!

However, in the case of `--haddock-for-hackage` I'd rather expect the following:

 1. Have `--haddock-for-hackage` imply `--quickjump` **only** when `version(haddock) >= 1.19`
 2. Have `--haddock-for-hackage` warn when called with `version(haddock) < 1.19` that `--quickjump` had to be disabled and that your locally generated docs on Hackage won't have the QuickJump feature

cc @harpocrates @alexbiehl

harpocrates:
We error-out when `--quickjump` is specified on a version of Haddock that
doesn't support it. However, if `--quickjump` was implied by
`--haddock-for-hackage`, we should just warn.

Fixes #5494.

---

### DIFF
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -153,6 +153,7 @@ haddock pkg_descr lbi suffixes flags' = do
         comp          = compiler lbi
         platform      = hostPlatform lbi
 
+        quickJmpFlag  = haddockQuickJump flags'
         flags = case haddockTarget of
           ForDevelopment -> flags'
           ForHackage -> flags'
@@ -185,9 +186,9 @@ haddock pkg_descr lbi suffixes flags' = do
     when (flag haddockQuickJump && version < mkVersion [2,19]) $ do
       let msg = "Haddock prior to 2.19 does not support the --quickjump flag."
           alt = "The generated documentation won't have the QuickJump feature."
-      case haddockTarget of
-        ForDevelopment -> die' verbosity msg
-        ForHackage -> warn verbosity (msg ++ "\n" ++ alt) 
+      if Flag True == quickJmpFlag
+        then die' verbosity msg
+        else warn verbosity (msg ++ "\n" ++ alt)
 
     haddockGhcVersionStr <- getProgramOutput verbosity haddockProg
                               ["--ghc-version"]
--- a/Cabal/Distribution/Simple/Haddock.hs
+++ b/Cabal/Distribution/Simple/Haddock.hs
@@ -178,13 +178,16 @@ haddock pkg_descr lbi suffixes flags' = do
         (orLaterVersion (mkVersion [2,0])) (withPrograms lbi)
 
     -- various sanity checks
-    when ( flag haddockHoogle
-           && version < mkVersion [2,2]) $
-         die' verbosity "haddock 2.0 and 2.1 do not support the --hoogle flag."
+    when (flag haddockHoogle && version < mkVersion [2,2]) $
+      die' verbosity "Haddock 2.0 and 2.1 do not support the --hoogle flag."
 
-    when ( flag haddockQuickJump
-           && version < mkVersion [2,19]) $
-         die' verbosity "haddock prior to 2.19 does not support the --quickjump flag."
+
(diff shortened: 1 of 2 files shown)


Cabal/Distribution/Simple/Haddock.hs
```diff
@@ -178,13 +178,16 @@ haddock pkg_descr lbi suffixes flags' = do
...
-    when ( flag haddockQuickJump
-           && version < mkVersion [2,19]) $
-         die' verbosity "haddock prior to 2.19 does not support the --quickjump flag."
+
+    when (flag haddockQuickJump && version < mkVersion [2,19]) $ do
+      let msg = "Haddock prior to 2.19 does not support the --quickjump flag."
+          alt = "The generated documentation won't have the QuickJump feature."
+      case haddockTarget of
+        ForDevelopment -> die' verbosity msg
+        ForHackage -> warn verbosity (msg ++ "\n" ++ alt)
```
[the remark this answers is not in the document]
harpocrates: I didn't want a visual break - just a new line. Should there be more of a visual break?

hvr:
@harpocrates do you have time to try to implement @23Skidoo suggestion?

i.e. so that

- `haddock --for-hackage` merely warns and gracefully degrades to disabling quickjump, whereas
- `haddock --for-hackage --quickjump` fails hard if `haddock` doesn't support quickjump

harpocrates:
@hvr Done. Now, any mention of `--quickjump`, with or without `--for-hackage`, will cause Cabal run with an older Haddock to fail hard.

Also noticed that I'd forgotten to let through a `--haddock-quickjump` flag for `new-haddock`... 😄

Cabal/Distribution/Simple/Haddock.hs
```diff
@@ -185,9 +186,9 @@ haddock pkg_descr lbi suffixes flags' = do
     when (flag haddockQuickJump && version < mkVersion [2,19]) $ do
       let msg = "Haddock prior to 2.19 does not support the --quickjump flag."
           alt = "The generated documentation won't have the QuickJump feature."
-      case haddockTarget of
-        ForDevelopment -> die' verbosity msg
-        ForHackage -> warn verbosity (msg ++ "\n" ++ alt) 
+      if Flag True == quickJmpFlag
```
23Skidoo: This is not quite what @hvr suggested IIUC. Since `--for-hackage` implies `--quickjump`, all of `--for-hackage`, `--for-hackage --quickjump`, and `--quickjump` will now fail hard when using an older haddock.

harpocrates: It is though (I've even tried it out). `quickJmpFlag` records what the quickjump flag was _prior_ to the `--for-hackage` implication. That way, only `--for-hackage --quickjump`, and `--quickjump` will fail hard when using an older haddock - `--for-hackage` by itself is just a warning.

23Skidoo: Great! Don't we still get a warning when neither `--for-hackage` nor `--quickjump` were specified, but `haddock` is old?

hvr: @23Skidoo I don't see any warning w/ e.g. GHC 8.2.2 (i.e. old haddock) and a plain `cabal new-haddock -w ghc-8.2.2` w/o any additional flags.

23Skidoo: OK. Let's merge!

hvr:
also cherry-picked into 2.4 via 76933f4

hvr:
@harpocrates thanks!

PR #5510, opened 2018-08-10
Solver: Improve handling of failures involving whole packages, such as unknown packages.

### LINKED ISSUE #5501: Better errors for reject-unconstrained-dependencies
In #5475, @grayjay pointed out that the current error reporting from the solver assumes that it is individual versions of a package being rejected. However, `reject-unconstrained-dependencies` will reject packages wholesale rather than individual versions. There may be a way to get better error messages if we can talk about packages rather than versions.

### LINKED ISSUE #5502: Rejecting whole packages not counted as a backjump
In #5475, @grayjay suggests that rejecting a package isn't counting as a backjump for the purposes of `--max-backjumps`. This should presumably be fixed, since (IIUC) it will potentially matter more now that `reject-unconstrained-packages` exists and rejects packages wholesale.

grayjay:
Two main commits:

#### Solver: Represent an unknown package with a failure node.

Previously, the solver represented an unknown package with an empty PChoice
node, because an unknown package is like a package with no versions.  Now that
--reject-unconstrained-dependencies also needs to reject whole packages, it is
simpler to handle both types of failures similarly, and use a failure node for
both.  Another advantage of using a failure node now is that the node can store
more information about why the package was rejected.  The explicit node also
simplifies generating the solver log.

This commit treats a failure node under a goal choice as a backjump in
D.S.Modular.Explore.exploreLog, which fixes #5502.

#### Solver: Print package rejection message on a single line.

The one-line error message is more similar to the message for rejecting a
package version.

Closes #5501.

Before:

```
[__0] trying: A-3.0.0 (user goal)
[__1] next goal: C (dependency of A)
[__1] fail (not a user-provided goal nor mentioned as a constraint, but reject-unconstrained-dependencies was set)
[__1] fail (backjumping, conflict set: A, C)
```

After:

```
[__0] trying: A-3.0.0 (user goal)
[__1] rejecting: C (not a user-provided goal nor mentioned as a constraint, but reject-unconstrained-dependencies was set) (dependency of A)
[__1] fail (backjumping, conflict set: A, C)
```

This commit also expands the tests for the error message for
--reject-unconstrained-dependencies.

_______________________________________________________

I'm not sure that putting the rejected package message on one line is an improvement, now that I've seen it, since that message contains more information than the rejected version message.  Let me know if you have any suggestions for improving it or prefer the original format.

/cc @kosmikus @quasicomputational 

---

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/Builder.hs
+++ b/cabal-install/Distribution/Solver/Modular/Builder.hs
@@ -24,6 +24,7 @@ import Data.Map as M
 import Data.Set as S
 import Prelude hiding (sequence, mapM)
 
+import qualified Distribution.Solver.Modular.ConflictSet as CS
 import Distribution.Solver.Modular.Dependency
 import Distribution.Solver.Modular.Flag
 import Distribution.Solver.Modular.Index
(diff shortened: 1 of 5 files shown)


quasicomputational:
Yeah, I'm also not sure about the one-liner. The `(dependency of A)` information is important, and I think we should put it at the front of the line, or on a line itself. Having it dangling at the end like that is sort of the worst of all worlds.

Here's a sketch of what it could look like with it at the front of the line:

```
[__0] trying: A-3.0.0 (user goal)
[__1] rejecting: C (dependency of A; not a user-provided goal nor mentioned as a constraint, but reject-unconstrained-dependencies was set)
[__1] fail (backjumping, conflict set: A, C)
```

Or, with longer package names:

```
[__0] trying: some-package-3.0.0 (user goal)
[__1] rejecting: some-other-package (dependency of some-package; not a user-provided goal nor mentioned as a constraint, but reject-unconstrained-dependencies was set)
[__1] fail (backjumping, conflict set: some-package, some-other-package)
```

That bikeshedding aside, the code looks sensible.

23Skidoo:
Code looks good from what I can tell. Re: the error message change, I think I prefer the original format.

grayjay:
Thanks for the feedback!  I removed the error message change and updated the unit test to test the original message.  I was worried that putting "dependency of A" between "rejecting: C" and the reason for the failure could imply that "dependency of A" was part of the reason that C was rejected.

PR #5515, opened 2018-08-11
Treat parse errors as errors instead of warnings

sol:
This produces error messages of the form:

    cabal: Failed parsing "./foo.cabal".
    foo.cabal:0:0: "version" field missing

### DIFF
--- a/Cabal/Distribution/Parsec/ParseResult.hs
+++ b/Cabal/Distribution/Parsec/ParseResult.hs
@@ -180,5 +180,4 @@ parseString parser verbosity name bs = do
     case result of
         Right x -> return x
         Left (_, errors) -> do
-            traverse_ (warn verbosity . showPError name) errors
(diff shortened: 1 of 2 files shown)


23Skidoo:
Looks reasonable. @phadej, What do you think?

hvr:
IMO, "Failed parsing" is the final conclusion (after having encountered a non-zero amount of hard parser failures), and as such should be the last line of output, after having enumerated encountered errors (and warnings, if we wanted to show those as well).

hvr:
@sol well, the problem is that now each error message is concatenated into a single `die`-message; thereby losing the individual error messaging structure... I'm starting to see why originally `warn` was picked... cause we didn't have a non-exiting variant of `die` at your disposal.

gbaz:
I think that the cost/benefit to figuring this out is sufficiently low and the patch is sufficiently old that I will close this.

PR #5521, opened 2018-08-14
Add flags to --version output

phadej:
```
% .../cabal --version
cabal-install version 2.4.0.0
compiled using version 2.4.0.0 of the Cabal library 
flags enabled: native-dns, lib, monolithic
```

### DIFF
--- a/cabal-install/main/Main.hs
+++ b/cabal-install/main/Main.hs
@@ -288,6 +288,30 @@ mainWorker args = do
                                   ++ "\ncompiled using version "
                                   ++ display cabalVersion
                                   ++ " of the Cabal library "
+                                  ++ flags_enabled
+      where
+        flags_enabled | null enabledFlags   = "\nall flags disabled"
+                      | otherwise = "\nflags enabled: " ++ intercalate ", " enabledFlags
+        enabledFlags =
+#ifdef CABAL_FLAG_NATIVEDNS
+            "native-dns" :
+#endif
+#ifdef CABAL_FLAG_LIB
+            "lib" :
+#endif
+#ifdef CABAL_FLAG_MONOLITHIC
+            "monolithic" :
+#endif
+#ifdef DEBUG_TRACETREE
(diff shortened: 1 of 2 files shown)


cabal-install/main/Main.hs
```diff
@@ -288,6 +288,30 @@ mainWorker args = do
                                   ++ "\ncompiled using version "
                                   ++ display cabalVersion
                                   ++ " of the Cabal library "
+                                  ++ flags_enabled
+      where
+        flags_enabled | null enabledFlags   = "\nall flags disabled"
```
23Skidoo: I'd rather have it say `flags enabled: none`.

cabal-install/main/Main.hs
```diff
@@ -288,6 +288,30 @@ mainWorker args = do
...
+                      | otherwise = "\nflags enabled: " ++ intercalate ", " enabledFlags
+        enabledFlags =
+#ifdef CABAL_FLAG_NATIVEDNS
+            "native-dns" :
```
23Skidoo: Maybe even render them as `-fnative-dns`, `-flib`, etc.

hvr: @23Skidoo then you could also enumerate all known flags explicitly everytime using the more explicit `--flags=` syntax:

`--flags="-native-dns +lib -monolithic"`

23Skidoo: That's even better!

23Skidoo:
Also there should be a comment somewhere in `cabal-install.cabal` reminding you to update the `--help` output when adding a new flag.

m-renaud:
Should I revive this PR?

fgaz:
@m-renaud +1 (on master though)

emilypi:
@m-renaud and @fgaz: +1 for reviving. If you give the word, I can get this fixed up.

phadej:
Cleaning up old PRs.

Issue #5119, opened 2018-02-06
Spurious warning: The package has an extraneous version range for a dependency on an internal library


phadej:
```
% cabal new-build cabal-plan:exe:cabal-plan
Resolving dependencies...
Build profile: -w ghc-8.2.2 -O1
In order, the following will be built (use -v for more details):
 - cabal-plan-0.3.0.0 (lib) (first run)
 - cabal-plan-0.3.0.0 (lib:topograph) (first run)
 - cabal-plan-0.3.0.0 (exe:cabal-plan) (first run)
Configuring library for cabal-plan-0.3.0.0..
Configuring library 'topograph' for cabal-plan-0.3.0.0..
Preprocessing library for cabal-plan-0.3.0.0..
Building library for cabal-plan-0.3.0.0..
Preprocessing library 'topograph' for cabal-plan-0.3.0.0..
Building library 'topograph' for cabal-plan-0.3.0.0..
[1 of 1] Compiling Cabal.Plan       ( src/Cabal/Plan.hs, /home/ogre/Documents/shared-haskell/cabal-plan/dist-newstyle/build/x86_64-linux/ghc-8.2.2/cabal-plan-0.3.0.0/build/Cabal/Plan.o )
[1 of 1] Compiling Topograph        ( src-topograph/Topograph.hs, /home/ogre/Documents/shared-haskell/cabal-plan/dist-newstyle/build/x86_64-linux/ghc-8.2.2/cabal-plan-0.3.0.0/l/topograph/build/topograph/Topograph.o )
Configuring executable 'cabal-plan' for cabal-plan-0.3.0.0..
Warning: The package has an extraneous version range for a dependency on an
internal library: cabal-plan -any && ==0.3.0.0 && ==0.3.0.0. This version
range includes the current package but isn't needed as the current package's
library will always be used.
```

There aren't such additional version restriction:

```
executable cabal-plan
  default-language:    Haskell2010
  other-extensions:    RecordWildCards

  main-is: src-exe/cabal-plan.hs
  other-modules: Paths_cabal_plan

  if flag(exe)
    -- dependencies w/ inherited version ranges via 'cabal-plan' library
    build-depends: cabal-plan
                 , topograph
....
```

To help bisecting: 
```
ii  cabal-install-head                              2.1+git20180202.0.2773999-5~16.04             amd64        Command-line interface for Cabal and Hackage
```

previous `cabal-install-head` (from around 2017-01-15) didn't had this issue, IIRC.

[...]

PR #5524, opened 2018-08-14
Hack around #5119 by disabling the warning.

quasicomputational:
Issue #5119 is tracking spurious generation of these warnings. There
are a lot of them! We can't release with so many false positives, so
just squelch the warning outright.

This is a hack and this commit should be reverted if we get a proper
fix that can be backported to the 2.4 branch.

---

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -1,4 +1,7 @@
 # 2.4.0.0 (current development version)
+  * Due to [#5119](https://github.com/haskell/cabal/issues/5119), the
+    `cabal check` warning for bounds on internal libraries has been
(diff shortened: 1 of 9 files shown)


quasicomputational:
Please note that this is targetting 2.4 only and not master. When/if this gets merged, it'll mean #5119 is no longer critical to fix before the release, though we should definitely try to get it done before 3.0. cc @Ericson2314.

23Skidoo:
OK, let's apply a workaround in 2.4 and later try to fix the issue properly in `master`.

quasicomputational:
OK, the only failure is because of a timeout on OSX. Let's do this.

PR #5526, opened 2018-08-14
[GSoC2018] Add support for multiple public libraries in a single package

### LINKED ISSUE #4206: Multiple public libraries in a package

fgaz:
This is my work for Google Summer of Code 2018.

It adds support for multiple public libraries, as described in #4206.

See https://github.com/fgaz/gsoc/blob/master/2018/final-report.md for a more detailed report. I'll expand a bit this description shortly.

**~~There are still a few nonessential things to do and some things to clean, so don't review yet! (see the link above or the checklist below). It should all work though.~~ Go ahead!**

Closes #4206 

/cc @ezyang @23Skidoo 

Things to fix/clean/update before merging:
* [x] Add a few comments in crucial places
* [x] Unbreak the few tests that still fail
  * [x] Adapt tests to new output
  * [x] Legit breakages
    * [x] Duplicate output on some tests
* [x] Clean the history
* [x] Make sure roundtripping works
  * [x] Generate only LMainLibName libraries in ProjectConfig's extra-packages and show them as pkg-only syntax, either in Dependency itself or only in ProjectConfig
  * [x] and/or remove Dependency from there
* [x] Clean up in a few places (memptys and todos)
* [x] Version guard the new functionality

After merging (better to merge soon and avoid to let it rot)
* [ ] Add a visibility field to sublibraries
* [ ] Maybe: remove all the bits of code that handle the old syntax (there's lots of them) and do it all in a single spot (hint: grep for `packageNameToUnqualComponentName`)
* [ ] Add tests
* [ ] Update changelog
* [ ] Update docs
* [ ] Maybe more source comments
* [ ] Deprecate old syntax
* [ ] Update `cabal init` to take into account sublibraries when populating build-depends

### DIFF
--- a/Cabal/Distribution/PackageDescription/Check.hs
+++ b/Cabal/Distribution/PackageDescription/Check.hs
@@ -585,7 +585,7 @@ checkFields pkg =
              , name `elem` map display knownLanguages ]
 
     testedWithImpossibleRanges =
-      [ Dependency (mkPackageName (display compiler)) vr
+      [ Dependency (mkPackageName (display compiler)) vr Set.empty
       | (compiler, vr) <- testedWith pkg
       , isNoVersion vr ]
 
@@ -598,7 +598,7 @@ checkFields pkg =
     internalLibDeps =
       [ dep
       | bi <- allBuildInfo pkg
-      , dep@(Dependency name _) <- targetBuildDepends bi
+      , dep@(Dependency name _ _) <- targetBuildDepends bi
       , name `elem` internalLibraries
       ]
 
@@ -611,14 +611,14 @@ checkFields pkg =
 
     depInternalLibraryWithExtraVersion =
       [ dep
-      | dep@(Dependency _ versionRange) <- internalLibDeps
+      | dep@(Dependency _ versionRange _) <- internalLibDeps
       , not $ isAnyVersion versionRange
       , packageVersion pkg `withinRange` versionRange
       ]
 
     depInternalLibraryWithImpossibleVersion =
       [ dep
-      | dep@(Dependency _ versionRange) <- internalLibDeps
+      | dep@(Dependency _ versionRange _) <- internalLibDeps
       , not $ packageVersion pkg `withinRange` versionRange
       ]
 
@@ -1243,8 +1243,8 @@ checkCabalVersion pkg =
         ++ ". To use this new syntax the package need to specify at least "
         ++ "'cabal-version: >= 1.6'. Alternatively, if broader compatibility "
         ++ "is important then use: " ++ commaSep
-           [ display (Dependency name (eliminateWildcardSyntax versionRange))
-           | Dependency name versionRange <- depsUsingWildcardSyntax ]
+           [ display (Dependency name (eliminateWildcardSyntax versionRange) Set.empty)
+           | Dependency name versionRange _ <- depsUsingWildcardSyntax ]
 
     -- check use of "build-depends: foo ^>= 1.2.3" syntax
   , checkVersion [2,0] (not (null depsUsingMajorBoundSyntax)) $
@@ -1255,8 +1255,8 @@ checkCabalVersion pkg =
         ++ ". To use this new syntax the package need to specify at least "
         ++ "'cabal-version: 2.0'. Alternatively, if broader compatibility "
         ++ "is important then use: " ++ commaSep
-           [ display (Dependency name (eliminateMajorBoundSyntax versionRange))
-           | Dependency name versionRange <- depsUsingMajorBoundSyntax ]
+           [ display (Dependency name (eliminateMajorBoundSyntax versionRange) Set.empty)
+           | Dependency name versionRange _ <- depsUsingMajorBoundSyntax ]
 
   , checkVersion [2,1] (any (not . null)
                         (concatMap buildInfoField
@@ -1292,8 +1292,8 @@ checkCabalVersion pkg =
         ++ ". To use this new syntax the package need to specify at least "
         ++ "'cabal-version: >= 1.6'. Alternatively, if broader compatibility "
         ++ "is important then use: " ++ commaSep
-           [ display (Dependency name (eliminateWildcardSyntax versionRange))
-           | Dependency name versionRange <- testedWithUsingWildcardSyntax ]
+           [ display (Dependency name (eliminateWildcardSyntax versionRange) Set.empty)
+           | Dependency name versionRange _ <- testedWithUsingWildcardSyntax ]
 
     -- check use of "source-repository" section
   , checkVersion [1,6] (not (null (sourceRepos pkg))) $
@@ -1367,11 +1367,11 @@ checkCabalVersion pkg =
     buildInfoField field         = map field (allBuildInfo pkg)
 
     versionRangeExpressions =
-        [ dep | dep@(Dependency _ vr) <- allBuildDepends pkg
+        [ dep | dep@(Dependency _ vr _) <- allBuildDepends pkg
               , usesNewVersionRangeSyntax vr ]
 
     testedWithVersionRangeExpressions =
-        [ Dependency (mkPackageName (display compiler)) vr
+        [ Dependency (mkPackageName (display compiler)) vr Set.empty
         | (compiler, vr) <- testedWith pkg
         , usesNewVersionRangeSyntax vr ]
 
@@ -1395,16 +1395,16 @@ checkCabalVersion pkg =
         alg (VersionRangeParensF _) = 3
         alg _ = 1 :: Int
 
-    depsUsingWildcardSyntax = [ dep | dep@(Dependency _ vr) <- allBuildDepends pkg
+    depsUsingWildcardSyntax = [ dep | dep@(Dependency _ vr _) <- allBuildDepends pkg
                                     , usesWildcardSyntax vr ]
 
-    depsUsingMajorBoundSyntax = [ dep | dep@(Dependency _ vr) <- allBuildDepends pkg
+    depsUsingMajorBoundSyntax = [ dep | dep@(Dependency _ vr _) <- allBuildDepends pkg
                                       , usesMajorBoundSyntax vr ]
 
     usesBackpackIncludes = any (not . null . mixins) (allBuildInfo pkg)
 
     testedWithUsingWildcardSyntax =
-      [ Dependency (mkPackageName (display compiler)) vr
+      [ Dependency (mkPackageName (display compiler)) vr Set.empty
       | (compiler, vr) <- testedWith pkg
       , usesWildcardSyntax vr ]
 
@@ -1493,7 +1493,7 @@ checkCabalVersion pkg =
     allModuleNamesAutogen = concatMap autogenModules (allBuildInfo pkg)
 
 displayRawDependency :: Dependency -> String
-displayRawDependency (Dependency pkg vr) =
+displayRawDependency (Dependency pkg vr _sublibs) =
   display pkg ++ " " ++ display vr
 
 
@@ -1545,7 +1545,7 @@ checkPackageVersions pkg =
           foldr intersectVersionRanges anyVersion baseDeps
         where
           baseDeps =
-            [ vr | Dependency pname vr <- allBuildDepends pkg'
+            [ vr | Dependency pname vr _ <- allBuildDepends pkg'
                  , pname == mkPackageName "base" ]
 
       -- Just in case finalizePD fails for any reason,
--- a/Cabal/Distribution/Types/Dependency.hs
+++ b/Cabal/Distribution/Types/Dependency.hs
@@ -19,6 +19,7 @@ import Distribution.Version ( VersionRange, thisVersion
 
 import qualified Distribution.Compat.ReadP as Parse
 
+import Distribution.CabalSpecVersion
 import Distribution.Text
 import Distribution.Pretty
 import qualified Text.PrettyPrint as PP
@@ -70,11 +71,26 @@ instance Pretty Dependency where
         prettySublib LMainLibName = PP.text $ unPackageName name
         prettySublib (LSubLibName un) = PP.text $ unUnqualComponentName un
 
+versionGuardMultilibs :: (Monad m, CabalParsing m) => m a -> m a
+versionGuardMultilibs expr = do
+  csv <- askCabalSpecVersion
+  if csv < CabalSpecV3_0
+  then fail $ unwords
+    [ "Sublibrary dependency syntax used."
+    , "To use this syntax the package needs to specify at least 'cabal-version: 3.0'."
+    , "Alternatively, if you are depending on an internal library, you can write"
+    , "directly the library name as it were a package."
+    ]
+  else
+    expr
+
 instance Parsec Dependency where
     parsec = do
         name <- lexemeParsec
+
         libs <- option [LMainLibName]
               $ (char ':' *> spaces *>)
+              $ versionGuardMultilibs
               $ between (char '{' *> spaces) (spaces <* char '}')
               $ parsecCommaList (makeLib name <$> parsecUnqualComponentName)
         ver  <- parsec <|> pure anyVersion
@@ -87,6 +103,7 @@ instance Text Dependency where
              Parse.skipSpaces
              libs <- option [LMainLibName]
                    $ (char ':' *>)
+                   $ versionGuardMultilibs
                    $ between (char '{') (char '}')
                    $ parsecCommaList (makeLib name <$> parsecUnqualComponentName)
              Parse.skipSpaces
--- a/Cabal/Distribution/Types/Dependency.hs
+++ b/Cabal/Distribution/Types/Dependency.hs
@@ -91,12 +91,15 @@ instance Parsec Dependency where
         libs <- option [LMainLibName]
               $ (char ':' *> spaces *>)
               $ versionGuardMultilibs
-              $ between (char '{' *> spaces) (spaces <* char '}')
-              $ parsecCommaList (makeLib name <$> parsecUnqualComponentName)
+              $ pure <$> parseLib name <|> parseMultipleLibs name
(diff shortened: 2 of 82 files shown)


Ericson2314:
Oh man this overlaps a lot with my old https://github.com/haskell/cabal/pull/4265 . I wish I could have gotten you that earlier.

Cabal/Distribution/Types/Dependency.hs
```diff
@@ -20,53 +21,82 @@ import qualified Distribution.Compat.ReadP as Parse
...
+        return $ Dependency name ver $ Set.fromList libs
+      where makeLib pn ln | unPackageName pn == ln = LMainLibName
+                          | otherwise = LSubLibName $ mkUnqualComponentName ln
 
 instance Text Dependency where
   parse = do name <- parse
              Parse.skipSpaces
              ver <- parse Parse.<++ return anyVersion
              Parse.skipSpaces
-             return (Dependency name ver)
```
phadej: ... but editing this `Text` instance shouldn't affect anything.

Cabal/Distribution/Types/Dependency.hs
```diff
@@ -20,53 +21,82 @@ import qualified Distribution.Compat.ReadP as Parse
...
+      where
+        prettySublibs = PP.text $ concat $ intersperse "," $ showSublib <$> Set.toList sublibs
+        showSublib LMainLibName = unPackageName name
+        showSublib (LSubLibName un) = unUnqualComponentName un
 
 instance Parsec Dependency where
     parsec = do
         name <- lexemeParsec
         ver  <- parsec <|> pure anyVersion
-        return (Dependency name ver)
```
phadej: *Note*: we need to case on `askCabalSpecVersion`.

phadej:
Also a dump question, are all internal sublibraries public with this change?

fgaz:
@Ericson2314 Yeah, I discovered that pr halfway through GSoC (and I actually [tried to rebase it](https://github.com/fgaz/cabal/tree/library-dependency)), but at that point my work diverged too much from that :-/ (I use `Set LibraryName` instead of a single library per dep).

I did not manage to separate the dependency-as-constraint from the dependency-dependency as you did, but I really think it should be done: https://github.com/fgaz/gsoc/blob/master/2018/final-report.md#the-dependency-as-constraint-problem

Ericson2314:
Glad you found it and it was of some use :)

phadej:
@fgaz is there any part which can be separated into own PR and merged already (even it's little or even no use by itself alone)? `ComponentName` refactor looks like that at a first glance.

fgaz:
@phadej yup, that and the GivenComponent one probably.

Also note that the LibraryName refactor could be [much more complete](https://github.com/fgaz/gsoc/blob/master/2018/final-report.md#the-pervasive-maybe-unqualcomponentname), but I ran into some problems (I probably just forgot something. I'll have to recheck why the tests broke)

fgaz:
Is it ok if I add the visibility field in another pr after the merge, so we merge soon and avoid conflicts?

23Skidoo:
@fgaz Yes, sure.

Cabal/Distribution/Types/Dependency.hs
fgaz: Why not? Isn't it used for `display`?

Cabal/Distribution/Types/Dependency.hs
```diff
@@ -20,53 +21,82 @@ import qualified Distribution.Compat.ReadP as Parse
...
+                                       <+> pretty ver
+      where
+        prettySublibs = PP.text $ concat $ intersperse "," $ showSublib <$> Set.toList sublibs
+        showSublib LMainLibName = unPackageName name
+        showSublib (LSubLibName un) = unUnqualComponentName un
 
 instance Parsec Dependency where
     parsec = do
         name <- lexemeParsec
+        libs <- option [LMainLibName]
```
phadej: Note: this should be guarded by `askCabalSpecVersion`

fgaz:
@phadej Just tested it and only direct deps are passed

phadej:
@fgaz that's good news :+1:

ezyang:
@fgaz Do you want us to review now? :)

fgaz:
@ezyang Sorry, not yet. Just wait a literal couple of days

fgaz:
@ezyang 
Got a cold and had to postpone this another couple of days. Now it's finally reviewable though.

Just ignore (for now):
  * the conflict: it's trivial
  * the messy history: working on it
  * the PackageVersionConstraint docs: working on it

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -523,7 +524,16 @@ filterConfigureFlags flags cabalLibVersion
...
 
-    flags_2_1_0 = flags_latest {
+    flags_2_5_0 = flags_latest {
+      -- Cabal < 2.5.0 does not understand --dependency=pkg:COMPONENT=cid
+      --                                   (public sublibraries)
+      configDependencies =
+        let isMainLib CLibName = True
+            isMainLib _        = False
+        in filter (\(_, c, _) -> isMainLib c) $ configDependencies flags
+      }
```
ezyang: I've always thought this filtering was somewhat questionable :) This code shouldn't ever get exercised, but what you have here seems reasonable enough (though, unlikely to lead to a successful build, if Cabal ever thinks it actually needs to pass it through :) (no action needed here)

ezyang:
@fgaz, I've started reviewing, and I wanted to say, this commit stack is beautifully structured, thank you so much!

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -528,9 +533,15 @@ filterConfigureFlags flags cabalLibVersion
       -- Cabal < 2.5.0 does not understand --dependency=pkg:COMPONENT=cid
       --                                   (public sublibraries)
       configDependencies =
-        let isMainLib CLibName = True
-            isMainLib _        = False
-        in filter (\(_, c, _) -> isMainLib c) $ configDependencies flags
+        let convertToLegacyInternalDep (GivenComponent _ (CSubLibName cn) cid) =
```
ezyang: oho... now I see what is going on here... this is for old style, per-component internal library deps :)

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1493,8 +1493,8 @@ checkCabalVersion pkg =
     allModuleNamesAutogen = concatMap autogenModules (allBuildInfo pkg)
 
 displayRawDependency :: Dependency -> String
-displayRawDependency (Dependency pkg vr) =
-  display pkg ++ " " ++ display vr
+displayRawDependency (Dependency pkg vr cs) =
+  display pkg ++ " " ++ display vr ++ " " ++ (unwords . fmap display . Set.toList $ cs) --MAYBE maybe drop cs entirely
```
ezyang: I guess it depends on what context `displayRawDependency` is used in. It looks like it is a faithful rendering, so I would drop it (in fact, I might suggest putting it in braces to demarcate it more clearly).

cabal-install/Distribution/Solver/Types/PackageConstraint.hs
```diff
@@ -97,6 +99,8 @@ data PackageProperty
    | PackagePropertySource
    | PackagePropertyFlags     FlagAssignment
    | PackagePropertyStanzas   [OptionalStanza]
+  -- MAYBE if we want to filter out packages without the specified sublibs,
+  -- add a PackagePropertySublibraries
```
ezyang: Nah, that seems like a silly thing to filter for :)

cabal-install/Distribution/Solver/Types/PackageIndex.hs
```diff
@@ -210,10 +210,10 @@ deletePackageName name =
   delete name (\pkg -> packageName pkg == name)
 
 -- | Removes all packages satisfying this dependency from the index.
---
+-- Does not take into account sublibraries, only version ranges.
```
ezyang: Err... so does it *not* delete sublibraries or does it delete all sublibraries in the package? I guess the answer to this question depends on if the `name` index here is munged or not.

ezyang:
I'm slightly concerned about merging without visibility checks, though this depends on what the default visibility should be. If default visibility is private (consistent with how convenience libraries were treated before), then if we merge this, we are doing a one way trip into everything is public land, and the only way to go back is a BC-breaking change. (Of course, if visibility gets landed very soon after, before we get any users, no harm done.) If default visibility is public there are no objections here.

Otherwise, this PR is looking really sharp. Thank you!

cabal-install/Distribution/Client/Dependency.hs
```diff
@@ -225,6 +225,7 @@ data PackagePreference =
      -- | If we would prefer to enable these optional stanzas
      -- (i.e. test suites and/or benchmarks)
    | PackageStanzasPreference   PackageName [OptionalStanza]
+   --TODO sublibrary preference
```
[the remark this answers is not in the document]
fgaz: Oops, I missed it during the rebase

cabal-install/Distribution/Client/Install.hs
```diff
@@ -407,7 +407,8 @@ planPackages verbosity comp platform mSandboxPkgInfo solver
       . addPreferences
           -- preferences from the config file or command line
           [ PackageVersionPreference name ver
-          | Dependency name ver <- configPreferences configExFlags ]
+          | Dependency name ver _ <- configPreferences configExFlags ]
+          --TODO sublib preference
```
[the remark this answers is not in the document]
fgaz: Same here

cabal-install/Distribution/Solver/Types/PackageConstraint.hs
fgaz: Removing it

cabal-install/Distribution/Solver/Types/PackageIndex.hs
fgaz: Yeah, I didn't think of that. This function isn't even used anywhere though, and cabal-install's api isn't consiedered stable

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -2829,12 +2829,12 @@ parsePackageArgs = parsePkgArgs []
                   ++ " package dependency."
 
 parseDependencyOrPackageId :: Parse.ReadP r Dependency
-parseDependencyOrPackageId = parse Parse.+++ liftM pkgidToDependency parse
+parseDependencyOrPackageId = undefined {-parse Parse.+++ liftM pkgidToDependency parse
```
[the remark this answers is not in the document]
fgaz: Nope, good catch

fgaz:
@ezyang

Thanks for the review! Good catches!

I'll remove those todos and that `undefined` and merge then

> If default visibility is public there are no objections here.

I think you mean private?

Sublibraries do become public by default with this pr, but only because there's no visibility flag yet, which will default to private

ezyang:
> Sublibraries do become public by default with this pr, but only because there's no visibility flag yet, which will default to private

I'm primarily pointing out that changing the default from public to private is a BC-breaking change. If you do it quickly enough no one will notice, but if you don't...

cabal-testsuite/PackageTests/Backpack/Includes2/setup-internal-fail.cabal.out
```diff
@@ -12,5 +12,6 @@ Error:
                            brought into scope by mixins: fail:mysql (Database.MySQL as Database)
                         or 'fail-0.1.0.0-postgresql:Database.PostgreSQL'
                            brought into scope by mixins: fail:postgresql (Database.PostgreSQL as Database)
-      While filling requirements of build-depends: fail:mylib
+      While filling requirements of     build-depends: fail:mylib
+                                    and build-depends: fail:mylib
```
fgaz: @ezyang

Any pointers on fixing the duplicate output? I think it may be some only-main-lib code I didn't delete when adding multilib handling. It does not affect functionality

fgaz:
I'm going to merge this later today unless someone objects.

@ezyang I'll make sure to add the visibility field before the 3.0 release

23Skidoo:
What about the duplicate output on some tests?

fgaz:
@23Skidoo I wasn't able to pinpoint the cause, but it seems not to affect functionality. Should I delay the merge until I fix it?

23Skidoo:
Well, it's just that you have that in your "to be fixed before merge" column.

fgaz:
@23Skidoo please don't merge yet. I think I found the issue. I'll merge it myself when it's fixed

fgaz:
It's reasy, but appveyor continues to fail with this error:

```
ghc-pkg.exe: cannot create: C:\Users\appveyor\AppData\Roaming\cabal\store\ghc-8.0.2\package.db already exists
```

Is it normal?

Ericson2314:
Congrats!!!

Issue #5484, opened 2018-08-01
Request: Add alias storedir for store-dir


AndreasPK:
Given that options specifying build output paths usually use --<X>dir (builddir, bindir, ...) that seems reasonable.

hvr:
maybe do it the other way around; rename `store-dir` to `storedir`, and add a (deprecated) `store-dir` aliasing `storedir`

cc'ing @alexbiehl who implemented `--store-dir`

23Skidoo:
Don't think we need to deprecate the alias, it's easier to keep supporting it.

ChShersh:
@hvr @23Skidoo Could you, please, provide some guidance regarding this issue?

I found this place in `cabal-install` where `store-dir` command-line option is mentioned:

```
cabal-install/Distribution/Client/Setup.hs
449:      ,option [] ["store-dir"]
```

Am I right that I just need to add the following to `argsNotShown`:

```haskell
      ,option [] ["storedir"]
         "The location of the nix-local-build store"
         globalStoreDir (\v flags -> flags { globalStoreDir = v })
         (reqArgFlag "DIR")
```

Or am I'm missing some mechanism for aliasing?

23Skidoo:
I think you can just add `"storedir"` to the `["store-dir"]` list.

ChShersh:
Fixed in #5527

Issue #5516, opened 2018-08-11
new-install fails with empty store:  ghc-pkg: .../package.db: getDirectoryContents:openDirStream: does not exist (No such file or directory)


hvr:
Invoking `cabal new-install` fails to initialize the respective store pkg-db if it didn't exist yet; this manifests as e.g.

```
$ cabal new-install -w ghc-8.6.1 hlint -vnormal+nowrap
cabal: ghc-pkg dump failed: dieVerbatim: user error (cabal: '/opt/ghc/bin/ghc-pkg-8.6.1' exited with an error:
ghc-pkg: /home/hvr/.cabal/store/ghc-8.6.0.20180810/package.db: getDirectoryContents:openDirStream: does not exist (No such file or directory)
)
```

cc @typedrat 

----

## WORKAROUND

For the respective `-w` argument, invoke

```
cabal new-repl -w ghc-8.6.1 --build-dep fail
```

for non-interactive environments, below seems to work:

```
echo "" | cabal new-repl -w ghc-8.6.1 --build-dep fail
```

this will build the trivial package `fail` and initialise the store pkg-db; after this `new-install` will work.

Yet another workaround suggested in https://github.com/haskell/cabal/issues/5516#issuecomment-424184066 is to create the folder `package.db` at the location mentioned in the error message.

hvr:
Raising priority as this was reported to cause problems for users bootstrapping/upgrading to newer versions of `cabal-install` via `new-install`

phadej:
This hit me in CI, where I needed to `new-install` some stuff into builder image. I workaround it by

```bash
    && echo 'cabal-version: 2.2'      > z.cabal \
    && echo 'name: z'                >> z.cabal \
    && echo 'version: 0'             >> z.cabal \
    && echo 'library'                >> z.cabal \
    && echo '  build-depends: aeson' >> z.cabal \
    && cat z.cabal \
    && (cabal new-build || true) \
    && rm -f z.cabal \
```

hvr:
The likely place where this is caused is at 

https://github.com/haskell/cabal/blob/b07e39fa414b89b18861b060b159625d40ebbf27/cabal-install/Distribution/Client/CmdInstall.hs#L445-L449

Now the question is how to tweak the code to not choke; a missing store pkg-db index folder in this codepath ought to be handled the same as if it was empty...

psftw:
[...]

Issue #5485, opened 2018-08-01
store-dir is broken with relative paths


AndreasPK:
The store dir is created, but ghc-pkg fails.

The issue persists if I invoke cabal multiple times.

```
find foobar-store/
foobar-store/
foobar-store/ghc-8.4.3
foobar-store/ghc-8.4.3/package.db
foobar-store/ghc-8.4.3/package.db/package.cache
```

```
cabal --store-dir=foobar-store new-build --enable-tests
Resolving dependencies...
Build profile: -w ghc-8.4.3 -O1
In order, the following will be built (use -v for more details):
 - primitive-0.6.4.0 (lib) (requires build)
 - random-1.1 (lib) (requires build)
 - safe-0.3.17 (lib) (first run)
 - tf-random-0.5 (lib) (requires build)
 - QuickCheck-2.11.3 (lib) (requires build)
 - safe-0.3.17 (test:safe-test) (first run)
Starting     random-1.1 (lib)
Configuring library for safe-0.3.17..
Starting     primitive-0.6.4.0 (lib)
Preprocessing library for safe-0.3.17..
Building library for safe-0.3.17..
[1 of 5] Compiling Safe.Partial     ( Safe\Partial.hs, C:\temp\safe\dist-newstyle\build\x86_64-windows\ghc-8.4.3\safe-0.3.17\build\Safe\Partial.o )
[2 of 5] Compiling Safe.Util        ( Safe\Util.hs, C:\temp\safe\dist-newstyle\build\x86_64-windows\ghc-8.4.3\safe-0.3.17\build\Safe\Util.o )
[3 of 5] Compiling Safe.Foldable    ( Safe\Foldable.hs, C:\temp\safe\dist-newstyle\build\x86_64-windows\ghc-8.4.3\safe-0.3.17\build\Safe\Foldable.o )
[4 of 5] Compiling Safe.Exact       ( Safe\Exact.hs, C:\temp\safe\dist-newstyle\build\x86_64-windows\ghc-8.4.3\safe-0.3.17\build\Safe\Exact.o )
[5 of 5] Compiling Safe             ( Safe.hs, C:\temp\safe\dist-newstyle\build\x86_64-windows\ghc-8.4.3\safe-0.3.17\build\Safe.o )

Failed to build primitive-0.6.4.0. The failure occurred during the configure
step.
Build log (
[...]

PR #5527, opened 2018-08-15
[#5484] Add 'storedir' alias for 'store-dir'

ChShersh:
---

I've executed the following command:

```
$ cabal --storedir foo/ new-build cabal-install
```

And got this error:

```
ghc-pkg: foo/ghc-8.4.3/package.db: getDirectoryContents:openDirStream: does
not exist (No such file or directory)
```

Without change in this PR it produces some error regarding parsing CLI arguments.

Maybe I should change some existing test to make sure that `storedir` option won't disappear? But I don't know which one and whether I should do this or not...

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -436,7 +436,7 @@ globalCommand commands = CommandUI {
          globalLocalRepos (\v flags -> flags { globalLocalRepos = v })
          (reqArg' "DIR" (\x -> toNubList [x]) fromNubList)
 
-      ,option [] ["logs-dir"]
+      ,option [] ["logsdir", "logs-dir"]
          "The location to put log files"
          globalLogsDir (\v flags -> flags { globalLogsDir = v })
          (reqArgFlag "DIR")
@@ -446,7 +446,7 @@ globalCommand commands = CommandUI {
          globalWorldFile (\v flags -> flags { globalWorldFile = v })
          (reqArgFlag "FILE")
 
-      ,option [] ["store-dir"]
+      ,option [] ["storedir", "store-dir"]
          "The location of the nix-local-build store"
          globalStoreDir (\v flags -> flags { globalStoreDir = v })
          (reqArgFlag "DIR")
@@ -468,7 +468,7 @@ configureCommand = c
       ++ "including v1-build, v1-test, v1-bench, v1-run, v1-repl.\n"
   , commandUsage        = \pname ->
     "Usage: " ++ pname ++ " v1-configure [FLAGS]\n"
-  , commandNotes = Just $ \pname -> 
+  , commandNotes = Just $ \pname ->
     (Cabal.programFlagsDescription defaultProgramDb ++ "\n")
       ++ "Examples:\n"
       ++ "  " ++ pname ++ " v1-configure\n"
@@ -1293,9 +1293,9 @@ upgradeCommand = configureCommand {
   }
 
 cleanCommand :: CommandUI CleanFlags
-cleanCommand = Cabal.cleanCommand 
+cleanCommand = Cabal.cleanCommand
   { commandUsage = \pname ->
-    "Usage: " ++ pname ++ " v1-clean [FLAGS]\n" 
+    "Usage: " ++ pname ++ " v1-clean [FLAGS]\n"
   }
 
 checkCommand  :: CommandUI (Flag Verbosity)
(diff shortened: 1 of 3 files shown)


phadej:
I'm :-1:, we are largely inconsistent with `data-dir` but `subdir`, `source-dir` and `builddir`, `destdir`, `dist-dir`, `output-dir`... so `store-dir` is just one more thing "you have to check" everytime, until it sticks.

phadej:
Also a general comment, @ChShersh don't push branches to the main repository. CI's will waste time building both the PR and your branch.

hvr:
@ChShersh the error you observed seems like an instance of #5516 - is this deterministically reproducible?

ChShersh:
@hvr Yes, I can see this error each time by calling:

```
cabal --storedir foo/ new-build cabal-install
```

But I'm not sure that it's relevant to #5516. In my case directory `foo/` doesn't exist. It's just a dummy argument to ensure that there exist an actual option with name `--storedir`. Though, I can see that the directory `foo/ghc-8.4.3/package.db/` is created after this command. I'm not quite understand what `--storedir` should do...

hvr:
@ChShersh do you see this also when you give `--storedir` an absolute path? (we also have this other issue #5485)

23Skidoo:
> CI's will waste time building both the PR and your branch.

I think this only affects AppVeyor, since for Travis we have a [branch whitelist](https://github.com/haskell/cabal/blob/master/.travis.yml#L9).

hvr:
@23Skidoo shall we use https://www.appveyor.com/docs/branches/#white--and-blacklisting then?

23Skidoo:
@hvr Done in 1e9df97d1564db77e8a073d35d18df849177edbc.

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -1293,9 +1293,9 @@ upgradeCommand = configureCommand {
   }
 
 cleanCommand :: CommandUI CleanFlags
-cleanCommand = Cabal.cleanCommand
```
23Skidoo: In the future, please try to put whitespace changes into a separate commit.

ChShersh:
@hvr 

>  do you see this also when you give --storedir an absolute path?

No, I don't see this issue. Everything works when I specify absolute path :ok_hand:

phadej:
@hvr, note that we can
```
cabal --store-dir=/tmp/asdad --logs-dir=/tmp/asdasd new-build Cabal --dry
```

and what's more surprising, these options aren't in `cabal --help` :(

phadej:
There are some `-dir` / `-dirs` flags in http://cabal.readthedocs.io/en/latest/nix-local-build.html#commands

phadej:
Sorry for interrupting on somewhat agreed change, I just find it weird. *all* `new-build` dir configuration are without dash...

cabal-install/Distribution/Client/Setup.hs
ChShersh: @23Skidoo Just curious. Is there any particular reason to do them in separate commit?

23Skidoo: Makes life easier for the reviewer.

ChShersh: @23Skidoo GitHub supports hiding of whitespace changes. You can add `?w=1` manually to the end of the link to hide whitespace changes. Like this:

* https://github.com/haskell/cabal/pull/5527/files?w=1

Oh you can open GitHub settings

screen shot 2018-08-16 at 9 56 15 am

But it means that reviewer still needs to do something to make his own life easier. So it's a tradeoff between who does less work. I'd prefer to do less 🙂 But since you're maintainer, it's okay for me to follow this commit style guide.

ChShersh:
@hvr @23Skidoo @phadej I've addressed issues from this comment and I've added more aliases to proposed options:

* https://github.com/haskell/cabal/pull/5527#issuecomment-413231581

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -436,7 +436,7 @@ globalCommand commands = CommandUI {
          globalLocalRepos (\v flags -> flags { globalLocalRepos = v })
          (reqArg' "DIR" (\x -> toNubList [x]) fromNubList)
 
-      ,option [] ["logs-dir"]
+      ,option [] ["logsdir", "logs-dir"]
```
phadej: I *suspect* that CI fails because the new variants are now first, should we add them last?

hvr: @ChShersh it'd be worth a try; but make it a separate commit in this PR so we can easily revert it if necessary

ChShersh:
@hvr @phadej @23Skidoo CI looks much better now. I can't see why that one particular build failed, but otherwise CI results look good.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into `2.4`.

PR #5538, opened 2018-08-22
WIP: use sudo: true on travis ci

joshk:
Josh here from Travis CI

Our `sudo: true` infrastructure are full VMs and are generally faster build environments than the `false` counterpart.

Additionally, we can enable 4 core VMs, giving you faster compile times.

This change will allow me to enable the 4 core feature for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ language: c
 
 dist: trusty
 
+# This sets the default config for each job to use full VMs.
+# The VMs have 2 cores and 8 gigs of ram. Larger VMs are also available.
 sudo: true
 
 # We whitelist branches, as we don't really need to build dev-branches.
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: c
 
 dist: trusty
 
-sudo: false
+sudo: true
 
 # We whitelist branches, as we don't really need to build dev-branches.
 # Remember to add release branches, both here and to appveyor.yml.


ChShersh:
I like this change. For all my Haskell projects I usually set `sudo: true` on Travis CI. Haskell can benefit a lot from build parallelism.

ezyang:
Blaming the relevant line, I see:

```
    - When possible, we use the container infrastructure (sudo: false)
      rather than Google Compute Engine infrastructure (sudo: required).
      Unfortunately, we can't use GCE for the Linux builds, where
      reduced RAM available hoses are GHC build.
```

Indeed, the failing job appears to have failed because of this https://travis-ci.org/haskell/cabal/jobs/419385574

joshk:
@ezyang the Google VMs have more ram, and I can move cabal to our 4 core 16 gigs of ram setup to have even more yet again.

I'm going to enable the larger VMs and make a minor change to the `.travis.yml` to see what impact this has.

phadej:
@joshk thanks a lot!

23Skidoo:
@joshk Thanks!

joshk:
My pleasure. Let me know how it goes and if there is anything I can help with.

23Skidoo:
@joshk We've been lately seeing frequent timeouts on the macOS configurations ([example](https://travis-ci.org/haskell/cabal/builds/422277579?utm_source=github_status&utm_medium=notification)), would be great if the time limit could be bumped a little bit for these.

joshk:
Hey @23Skidoo, sorry I missed your comment. Sure, I can look into this. Is it still happening?

Also, how are the larger VMs going? It might be best to look into more concurrency for Haskell so that builds also process faster.

Issue #5541, opened 2018-08-23
Too many open files with cabal new-sdist and many data-files


nomeata:
A project that I was trying to build contains one package (`ghc-toolkit`) that has 1054 files in `data-files` defined 
(https://github.com/tweag/asterius/tree/caf15690c54aef3ae4eecbc7224c6163de040f7a/ghc-toolkit).

It also contains another package `genapply` that does not even depend on `ghc-toolkit`.

I can run `cabal new-build genapply` just fine. But if I run `cabal new-install genapply`, I get

```
$ cabal new-install genapply -w ~/build/haskell/ghc/inplace/bin/ghc-stage2 --allow-newer=base
Wrote tarball sdist to
/home/jojo/build/haskell/asterius/dist-newstyle/sdist/asterius-0.0.1.tar.gz
boot-libs/template-haskell/Language/Haskell/TH/LanguageExtensions.hs: openBinaryFile: resource exhausted (Too many open files)
```
Interestingly, 
```
$ cabal new-sdist genapply 
Downloading
https://github.com/haskell/primitive/archive/ea9f0426646ecf8a5a0a1659b29a8dc014ceda05.tar.gz
Wrote tarball sdist to
/home/jojo/build/haskell/asterius/dist-newstyle/sdist/genapply-0.1.tar.gz
```
works, but
```
$ cabal new-sdist all
Downloading
https://github.com/haskell/primitive/archive/ea9f0426646ecf8a5a0a1659b29a8dc014ceda05.tar.gz
Wrote tarball sdist to
/home/jojo/build/haskell/asterius/dist-newstyle/sdist/asterius-0.0.1.tar.gz
boot-libs/template-haskell/Language/Haskell/TH/Ppr.hs: openBinaryFile: resource exhausted (Too many open files)
```
does not.

So maybe there are two issues here:

 * `cabal new-install` creates more `sdists` than it has to.
 * `cabal new-sdist` does not work well with large number of `data-files`.

hvr:
It's quite possible `new-sdist` might have uncovered a long-standing bug in the sdist code...

typedrat:
This means files aren't getting closed correctly, because if `new-sdist` works for one it should work for all.

typedrat:
I've been looking at the code and thinking about this for the last bit, and I just do not see how this bug is possible.

23Skidoo:
Lazy I/O? A missing `bracket` leading to a resource leak?

typedrat:
[...]

PR #5542, opened 2018-08-23
Add a test case for #5541

nomeata:


quasicomputational:
I think this also needs a `cabal.project` in the test directory, or else the outer one (i.e., Cabal's) will be picked up.

nomeata:
Added.

phadej:
I wonder if the input files could be generated. Doesn't feel very nice to have arbitrary amount of files for a test checked-in. Someone might have a bigger ulimit, so test will pass trivially.

23Skidoo:
+1 on autogenerating test data.

PR #5561, opened 2018-09-03
Clarify build-tool-depends a bit more

domenkozar:
---

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -2063,16 +2063,22 @@ system-dependent values for these fields.
 .. pkg-field:: build-tool-depends: package:executable list
     :since: 2.0
 
-    A list of Haskell programs needed to build this component.
+    A list of Haskell executabes needed to build this component.
+
     Each is specified by the package containing the executable and the name of the executable itself, separated by a colon, and optionally followed by a version bound.
-    It is fine for the package to be the current one, in which case this is termed an *internal*, rather than *external* executable dependency.
 
-    External dependencies can (and should) contain a version bound like conventional :pkg-field:`build-depends` dependencies.
-    Internal deps should not contain a version bound, as they will be always resolved within the same configuration of the package in the build plan.
-    Specifically, version bounds that include the package's version will be warned for being extraneous, and version bounds that exclude the package's version will raise an error for being impossible to follow.
+    All executables defined in the given Cabal file are termed as *internal* dependency as opposed to the rest which are *external* dependency.
+    Each of the two is handled differently:
 
-    Cabal can make sure that specified programs are built and on the ``PATH`` before building the component in question.
-    It will always do so for internal dependencies, and also do so for external dependencies when using Nix-style local builds.
+    1. External dependencies can (and should) contain a version bound like conventional :pkg-field:`build-depends` dependencies.
+    2. Internal depenedencies should not contain a version bound, as they will be always resolved within the same configuration of the package in the build plan.
+       Specifically, version bounds that include the package's version will be warned for being extraneous, and version bounds that exclude the package's version will raise an error for being impossible to follow.
+
+    Cabal makes sure that specified programs are built and provided on the ``PATH`` before building the component in question under following conditions:
+
+    a) For Nix-style local builds, both internal and external dependencies.
+    b) For old-style builds, only for internal dependencies [#old-style-build-tool-depends]_. 
+       It's up to the user to specify needed executables in this case.
 
     :pkg-field:`build-tool-depends` was added in Cabal 2.0, and it will
     be ignored (with a warning) with old versions of Cabal.  See
@@ -3478,3 +3484,13 @@ a few options:
 
 
 .. include:: references.inc
+
+.. rubric:: Footnotes
+
+.. [#old-style-build-tool-depends] Some packages (ab)use :pkg-field:`build-depends` on old-style builds, but this has a few major drawbacks:
+
+                                   - using new-build it's considered an error if you depend on a exe-only package via build-depends: the solver will refuse it
+
+                                   - it may or may not place the executable on $PATH
+
+                                   - it does not ensure correct version of the package is installed, you might end up overwriting versions with each other
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -2063,7 +2063,8 @@ system-dependent values for these fields.
 .. pkg-field:: build-tool-depends: package:executable list
     :since: 2.0
 
-    A list of Haskell executabes needed to build this component.
+    A list of Haskell executables needed to build this component. Executables are provided during the whole duration of the build,
+    so this field can be used for executables needed during :pkg-section:`test-suite` as well.
 
     Each is specified by the package containing the executable and the name of the executable itself, separated by a colon, and optionally followed by a version bound.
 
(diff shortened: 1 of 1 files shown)


domenkozar:
@hvr I hope I have addressed all of your comments.

hvr:
>  of course only of valid build-tool-depends strings and nothing more.

well, I think it'd be more illustrative to show a minimal example which involves an `executable` and a `test-suite`, where the test-suite uses `build-tool-depends` to get access to the intra-package `executable` (without version bounds obviously), and you can combine that with having the `executable` refer to some external build-tools such as markdown-unlit (with version bounds)

Cabal/doc/developing-packages.rst
```diff
@@ -3478,3 +3488,13 @@ a few options:
 
 
 .. include:: references.inc
+
+.. rubric:: Footnotes
+
+.. [#old-style-build-tool-depends] Some packages (ab)use :pkg-field:`build-depends` on old-style builds, but this has a few major drawbacks:
+
+                                   - using Nix-style builds it's considered an error if you depend on a exe-only package via build-depends: the solver will refuse it
```
Ericson2314: This is full sentence, so I'd punctate it as such like the others

Cabal/doc/developing-packages.rst
```diff
@@ -3478,3 +3488,13 @@ a few options:
...
+
+                                   - it may or may not place the executable on $PATH
+
+                                   - it does not ensure correct version of the package is installed, you might end up overwriting versions with each other
```
Ericson2314: *the* correct version

Ericson2314: , *so* you might

Cabal/doc/developing-packages.rst
```diff
@@ -2063,21 +2063,31 @@ system-dependent values for these fields.
...
-    Specifically, version bounds that include the package's version will be warned for being extraneous, and version bounds that exclude the package's version will raise an error for being impossible to follow.
+    All executables defined in the given Cabal file are termed as *internal* dependency as opposed to the rest which are *external* dependency.
+    Each of the two is handled differently:
+
+    1. External dependencies can (and should) contain a version bound like conventional :pkg-field:`build-depends` dependencies.
+    2. Internal depenedencies should not contain a version bound, as they will be always resolved within the same configuration of the package in the build plan.
+       Specifically, version bounds that include the package's version will be warned for being extraneous, and version bounds that exclude the package's version will raise an error for being impossible to follow.
+
+    Cabal makes sure that specified programs are built and provided on the ``PATH`` before building the component in question under following conditions:
```
Ericson2314: Cabal try to makes sure that all specified programs are built and provided on the ``PATH`` before building the component in question, but can only do so for Nix-style builds. Specifically:

Ericson2314: I found something like that easier to read

Cabal/doc/developing-packages.rst
```diff
@@ -2063,21 +2063,31 @@ system-dependent values for these fields.
 .. pkg-field:: build-tool-depends: package:executable list
     :since: 2.0
 
-    A list of Haskell programs needed to build this component.
+    A list of Haskell executables needed to build this component. Executables are provided during the whole duration of the build,
+    so this field can be used for executables needed during :pkg-section:`test-suite` as well.
```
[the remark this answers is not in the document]
domenkozar: Yes :) Anything to clarify here?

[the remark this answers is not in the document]
hvr: @Ericson2314 btw, assuming we get `run-tool-depends`  at some point; we'd still need to emulate the current `build-tools-depends` which conflatedly implies the future `run-tool-depends` w/ `cabal-version:2.*` packages; so at that point, @domenkozar wording will still be valid for packages using the respective `cabal-version:`-values which still had these semantics.

Cabal/doc/developing-packages.rst
```diff
@@ -2063,21 +2063,31 @@ system-dependent values for these fields.
...
+    All executables defined in the given Cabal file are termed as *internal* dependencies as opposed to the rest which are *external* dependencies.
+    Each of the two is handled differently:
+
+    1. External dependencies can (and should) contain a version bound like conventional :pkg-field:`build-depends` dependencies.
+    2. Internal depenedencies should not contain a version bound, as they will be always resolved within the same configuration of the package in the build plan.
+       Specifically, version bounds that include the package's version will be warned for being extraneous, and version bounds that exclude the package's version will raise an error for being impossible to follow.
 
-    Cabal can make sure that specified programs are built and on the ``PATH`` before building the component in question.
-    It will always do so for internal dependencies, and also do so for external dependencies when using Nix-style local builds.
+    Cabal tries to make sure that all specified programs are built and provided on the ``PATH`` before building the component in question, but can only do so for Nix-style builds. Specifically:
```
hvr: One thing worth mentioning explicitly is that those tools' are prepended to the search `$PATH`; that's important as otherwise a globally same-called executable could interfere (concrete example: if you happen to have a system-wide installed `alex` executable version 3.1.3 in `/usr/bin`, but your `build-tools-depends: alex:alex ^>= 3.2.4` you'd certainly want to use the one provisioned by `new-build` satisfying the version constraint to take precedence.

Cabal/doc/developing-packages.rst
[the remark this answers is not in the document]
hvr: Btw, that sentence implies that running the test-suite is considered part of the build-phase; I think it's worth spelling this out as otherwise it might not be obvious that the respective `$PATH`-augmentation will also be in effect during the invocation of the test-suite (via e.g. `cabal new-test`)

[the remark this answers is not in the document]
hvr: @quasicomputational you picked up on my subtle hint... ;-)

I consider this a bit of a bug, and I seem to recall we already talked about this in another issue; i.e. ideally

- `cabal new-run test:x` behaves close to `cabal new-test x` 
- `cabal new-run bench:x` behaves close to `cabal new-bench x` 

as I'd consider it merely a different interface to invoking the test-suites; but the $PATH setup ought to be compatible.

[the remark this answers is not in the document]
Ericson2314: Right insofar that there's a bug, I consider the `new-run` behavior correct :D.

fommil:
From IRC:

     <fommil> 3. `build-tool-depends` sounds very interesting, but confused me.
         Will cabal download and build these components, on a per-project
         basis? Or is this simply a requirement check and installation is the
         responsibility of the developer?  [10:50]

Cabal/doc/developing-packages.rst
fommil: "are built" ... does this mean `cabal` will do the building automatically or that the user is expected to do the building independently?

hvr: @fommil would "are atomically built" clarify the intent?

domenkozar:
Hey all, I haven't given up on this, thanks for all the feedback. Going to revisit this today.

domenkozar:
@hvr @Ericson2314 @gilligan @fommil @ChShersh addressed comments, I'm not entirely sure about the examples, they should probably be full-blown otherwise they provide little insight.

hvr:
@domenkozar Thank you! I've merged the current state so we can get an incremental improvement in earlier rather than later (so we can start pointing to the official docs); for the examples let's do a new PR if you're up to it! :-)

PR #5567, opened 2018-09-05
Extend `cabal init` support for `cabal-version` selection

hvr:
This adds support to `cabal init` for selecting the CABAL spec
version (i.e. the `cabal-version:` field) and also takes care
of translating between pre-SPDX and SPDX license ids.

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -17,7 +17,6 @@ module Distribution.Client.Init (
 
     -- * Commands
     initCabal
-  , pvpize
   , incVersion
 
   ) where
@@ -48,7 +47,7 @@ import Control.Arrow
 import Text.PrettyPrint hiding (mode, cat)
 
 import Distribution.Version
-  ( Version, mkVersion, alterVersion
+  ( Version, mkVersion, alterVersion, versionNumbers, majorBoundVersion
   , orLaterVersion, earlierVersion, intersectVersionRanges, VersionRange )
 import Distribution.Verbosity
   ( Verbosity )
@@ -70,7 +69,8 @@ import Distribution.Client.Init.Heuristics
     scanForModules, neededBuildPrograms )
 
 import Distribution.License
-  ( License(..), knownLicenses )
+  ( License(..), knownLicenses, licenseToSPDX )
+import qualified Distribution.SPDX as SPDX
 
 import Distribution.ReadE
   ( runReadE, readP_to_E )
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Init.hs
```diff
@@ -209,15 +238,25 @@ getLicense flags = do
...
+    Just msg -> putStrLn msg >> getLicense flags
+    Nothing  -> return $ flags { license = maybeToFlag lic }
 
   where
-    isLicenseInvalid (Just (UnknownLicense t)) = any (not . isAlphaNum) t
-    isLicenseInvalid _ = False
+    displayLicense l | needSpdx  = prettyShow (licenseToSPDX l)
+                     | otherwise = display l
+
+    checkLicenseInvalid (Just (UnknownLicense t))
```
23Skidoo: Not required now, but do you think we could change the type of `InitFlags.license` to `Either License SPDX.License`?

ChShersh:
@hvr I wonder, are there any drawback for not using the latest cabal version so there's need to choose `cabal-version`?

cabal-install/Distribution/Client/Init.hs
hvr: @23Skidoo I tried that briefly but in the interest to get this implemented faster I went for the hackier approach and wanted to postpone this to a future PR; doing it properly would require quite a bit of refactoring in multiple places

cabal-install/Distribution/Client/Setup.hs
```diff
@@ -2184,9 +2184,9 @@ initCommand = CommandUI {
                           (flagToList . fmap display))
 
       , option [] ["cabal-version"]
-        "Required version of the Cabal library."
+        "Version of the CABAL specification."
```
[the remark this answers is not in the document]
hvr: Fair enough; I was struggling how to disambiguate between the library, the spec and the executable, and I assumed that all-caps "CABAL" was sometimes used to refer to the spec. I'm totally fine with changing it to "Version of the Cabal specification".

hvr:
@ChShersh There's only a minor drawback that you exclude older `cabal` clients; so if you don't need the additional features it's sometimes desirable to stick with the oldest cabal spec version which supports the features you want to use. However, I see little harm to provide the user the choice, and see how this ends up being used. The version selection also has as a subtle side effect that it highlights some new features with each major version, and tries to convey what `cabal-version:x.y` actually means, as there's a lot of confusion about it.

hvr:
(cherry-picked to 2.4 branch)

Issue #5503, opened 2018-08-06
`cabal new-haddock` fails for build-type:configure 


hvr:
Reproduction:

```
$ cabal get network-2.7.0.2 && cd network-2.7.0.2/
$ cabal new-haddock -w ghc-8.4.3 -v
$ cabal new-haddock -w ghc-8.4.3 
Build profile: -w ghc-8.4.3 -O1
In order, the following will be built (use -v for more details):
 - network-2.7.0.2 (first run)
cabal: Unrecognised flags: lib:network
Warning: Failed to build documentation for network-2.7.0.2.
```

using `-v` reveals:

```
...
configure: creating ./config.status
config.status: creating network.buildinfo
config.status: creating include/HsNetworkConfig.h
configure: WARNING: unrecognized options: --with-compiler
Reading parameters from
/tmp/network-2.7.0.2/dist-newstyle/build/x86_64-linux/ghc-8.4.3/network-2.7.0.2/build/network.buildinfo
whenReRegister: nothing to register
Using self-exec internal setup method with build-type Configure and args:
["act-as-setup","--build-type=Configure","--","haddock","--verbose=2","--builddir=/tmp/network-2.7.0.2/dist-newstyle/build/x86_64-linux/ghc-8.4.3/network-2.7.0.2","lib:network"]
/stuff/dot-cabal/bin/cabal act-as-setup --build-type=Configure -- haddock
--verbose=2
--builddir=/tmp/network-2.7.0.2/dist-newstyle/build/x86_64-linux/ghc-8.4.3/network-2.7.0.2
lib:network
cabal: Unrecognised flags: lib:network
Warning: Failed to build documentation for network-2.7.0.2-inplace.
```

hvr:
This is related to #5261 which fixed the same situation for `build-type: Custom`; however, the issue at hand is about `build-type: Configure`!

hvr:
@alexbiehl suggests the culprit to be

 - https://github.com/haskell/cabal/blob/b8c4ce3dacece1b1ac68360b48a3129d9ca440c1/cabal-install/Distribution/Client/ProjectBuilding.hs#L1049

missing the logic from

- https://github.com/haskell/cabal/blob/b8c4ce3dacece1b1ac68360b48a3129d9ca440c1/cabal-install/Distribution/Client/ProjectBuilding.hs#L1311

PR #5569, opened 2018-09-07
[#5503] Add filterHaddockFlags for new-haddock

chshersh:
---

I've tried to add additional logic suggested by the issue. But I still see the error:

```
 - network-2.7.0.2 (first run)
cabal: Unrecognised flags: lib:network
Warning: Failed to build documentation for network-2.7.0.2.
```

### DIFF
--- a/cabal-install/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/Distribution/Client/ProjectBuilding.hs
@@ -1070,7 +1070,8 @@ buildAndInstallUnpackedPackage verbosity
     buildFlags   _   = setupHsBuildFlags pkg pkgshared verbosity builddir
 
     haddockCommand   = Cabal.haddockCommand
-    haddockFlags _   = setupHsHaddockFlags pkg pkgshared
+    haddockFlags v   = flip filterHaddockFlags v $
+                       setupHsHaddockFlags pkg pkgshared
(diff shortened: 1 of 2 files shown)


hvr:
@alexbiehl the problem appears to be (haven't investigated enough to be sure) that `build-type:configure` throws the package into legacy mode and thus disables per-component  -- this appears to disable support for passing arguments such as `lib:network` to `Setup.hs haddock` 

@ChShersh I'm afraid, we need to do a bit more here; we need to e.g. extend the arg filtering to detect this case somehow

hvr:
@alexbiehl @ChShersh take a look at this here

https://github.com/haskell/cabal/blob/0c9792076b30257ff638bf6dd5a13b77ec7b22be/Cabal/Distribution/Simple.hs#L708

I'm afraid we've been looking at the wrong place to fix this...

ChShersh:
@hvr Sure, no problems. I will do the work in separate PR :ok_hand:

hvr:
As discussed w/ @ChShersh and the urgency involved I've created #5573 myself

ChShersh:
@hvr @23Skidoo Could you, please, provide some updates on what is required for this PR to be done? Am I right that this is resolved in #5573 or is something needs to be done here as well?

PR #5578, opened 2018-09-17
Add new v2-build flag '--only-configure'

DanielG:
When this flag is active we only execute the Cabal build system's 'configure'
step for each local unit and nothing else. This allows (re)generating the Setup
configuration cache which is accessed directly by tooling like ghc-mod and HIE
through cabal-helper and allows such tooling to forgoe a full rebuild.

I've added a new `BuildFlags` type for `v2-build` specific options since don't think just sticking this in any of the other flags sets makes even a lick of sense.

It seems to me that for some reason the `defaultBuildFlags` are not being applied to the flags before being passed to `buildAction`, so I have worked around that by doing `fromFlag (buildOnlyConfigure defaultBuildFlags <> buildOnlyConfigure buildFlags)` which is kind of really ugly.

@hvr turns out combining `--only-depencencies` with `--only-configure` does kind of make sense for multi package projects, though I don't imagine it would be very useful, so I'm not bothering with a check for that.

---

### DIFF
--- a/cabal-install/changelog
+++ b/cabal-install/changelog
@@ -3,6 +3,7 @@
 2.6.0.0 (current development version)
 	* New solver flag: '--reject-unconstrained-dependencies'. (#2568)
 	* "cabal new-repl" now works for indefinite (in the Backpack sense) components.
+	* New v2-build flag: '--only-configure'. (#5578)
 
 2.4.0.1
 	* 'new-sdist' now generates tarballs with file modification
--- a/Cabal/doc/nix-local-build.rst
+++ b/Cabal/doc/nix-local-build.rst
@@ -368,6 +368,12 @@ cause extra store packages to be built (for example,
 ``--enable-profiling`` will automatically make sure profiling libraries
 for all transitive dependencies are built and installed.)
 
+In addition ``cabal new-build`` accepts these flags:
+
+- ``--only-configure``: When given we will forgoe performing a full build and
+  abort after running the configure phase of each target package.
+
+
 cabal new-repl
 --------------
 
--- a/cabal-install/Distribution/Client/CmdBuild.hs
+++ b/cabal-install/Distribution/Client/CmdBuild.hs
@@ -14,13 +14,15 @@ module Distribution.Client.CmdBuild (
 import Distribution.Client.ProjectOrchestration
 import Distribution.Client.CmdErrorMessages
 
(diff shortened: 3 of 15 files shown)


hvr:
@DanielG the build failures look legit; I'd recommend heeding the advice from the README, 

> Cabal has its own Prelude, in `Distribution.Compat.Prelude`, that provides a compatibility layer and exports some commonly used additional functions. Use it in all new modules.

cabal-install/changelog
```diff
@@ -2,6 +2,7 @@
 
 2.6.0.0 (current development version)
 	* New solver flag: '--reject-unconstrained-dependencies'. (#2568)
+        * New v2-build flag: '--only-configure'. (#5578)
```
phadej: tab vs spaces :S

DanielG: The entry just below `Bugfix: "cabal new-build --ghc-option '--bogus' --ghc-option '-O1'"` also has spaces, I just copied that one ;)

DanielG:
Does the CI not build cabal-install? The 7.8 build is supposedly green but I get

```
Distribution/Client/Get.hs:203:3:
    ‘displayException’ is not a (visible) method of class ‘Exception’
```

when building it locally.

phadej:
@hvr you forgot to patch master? https://github.com/haskell/cabal/commit/0ba16f5c98ad612920f79e60d9105866096fb2e5

(I'd argue that we should have patches into master first, and then backport them to stable branches)

hvr:
@phadej I intentionally held off porting to master before I could fix the CI which doesn't properly ignore `cabal-install` when `ONLY_LIB_CABAL=True` is set

DanielG:
The [8.6.2 build](https://travis-ci.org/haskell/cabal/jobs/447379380) is failing with:

```
[...]
----
$ cabal act-as-setup --build-type=Simple -- haddock --builddir=/home/travis/build/haskell/cabal/dist-newstyle/build/x86_64-linux/ghc-8.6.2/Cabal-2.5.0.0
Run the 'configure' command first.
```

Is that a known problem or did I break something?

23Skidoo:
It's failing on master as well and no-one has gotten around to fixing it yet.

DanielG:
Ok, good. The OSX build is failing due to timeout. I assume that can also be ignored?

PR #5579, opened 2018-09-17
No readp

phadej:
This is preview PR. I'll chop this into separate PRs for each commit here, to not make too big changes at once.

There are few things not made, and maybe there are some unexpected breakages too.

cc @hvr @fgaz


phadej:
Done in #5746

PR #5580, opened 2018-09-18
Remove Text from D.Simple

phadej:
Also add Parsec/Pretty CompilerId instances

---


Cabal/Distribution/Simple/Utils.hs
```diff
@@ -513,7 +515,7 @@ infoNoWrap verbosity msg = withFrozenCallStack $
 
 -- | Detailed internal debugging information
 --
--- We display these messages when the verbosity level is 'deafening'
+-- We prettyShow these messages when the verbosity level is 'deafening'
```
phadej: ugh

Cabal/Distribution/Simple/BuildTarget.hs
```diff
@@ -183,30 +182,31 @@ readUserBuildTargets = partitionEithers . map readUserBuildTarget
...
 readUserBuildTarget targetstr =
-    case readPToMaybe parseTargetApprox targetstr of
-      Nothing  -> Left  (UserBuildTargetUnrecognised targetstr)
-      Just tgt -> Right tgt
+    case explicitEitherParsec parseTargetApprox targetstr of
+      Left _    -> Left  (UserBuildTargetUnrecognised targetstr)
+      Right tgt -> Right tgt
 
   where
-    parseTargetApprox :: Parse.ReadP r UserBuildTarget
```
phadej: this **should** manifest in some test. Let's see

phadej: and it did :P

phadej:
This is ready for review. Ping @hvr.
Essentially this is
- `s/disp/pretty/` + `s/display/prettyShow/`
- Changing some `Distribution.Simple` parser-like things to build upon `Parsec` instances

Cabal/Distribution/Simple/BuildTarget.hs
```diff
@@ -180,33 +179,59 @@ readUserBuildTargets :: [String] -> ([UserBuildTargetProblem]
...
+-- Right (UserBuildTargetDouble "lib" "comp")
+--
+-- >>> readUserBuildTarget "pkg:lib:\"comp\""
+-- Right (UserBuildTargetTriple "pkg" "lib" "comp")
+--
+-- >>> readUserBuildTarget "pkg:lib:comp:more"
+-- Left (UserBuildTargetUnrecognised "pkg:lib:comp:more")
+--
+-- >>> readUserBuildTarget "pkg:\"lib\":comp"
+-- Left (UserBuildTargetUnrecognised "pkg:\"lib\":comp")
```
quasicomputational: We're not actually doctesting as part of the test suite, right?

Cabal/Distribution/Simple/BuildTarget.hs
```diff
@@ -180,33 +179,59 @@ readUserBuildTargets :: [String] -> ([UserBuildTargetProblem]
...
-              c <- tokenQ
-              return (UserBuildTargetTriple a b c))
-
-    token  = Parse.munch1 (\x -> not (isSpace x) && x /= ':')
-    tokenQ = parseHaskellString <++ token
-    parseHaskellString :: Parse.ReadP r String
-    parseHaskellString = Parse.readS_to_P reads
+    parseTargetApprox :: CabalParsing m => m UserBuildTarget
+    parseTargetApprox = do
+        -- read one, two, or three tokens, where last could be "hs-string"
```
quasicomputational: I don't find this clearer than the previous implementation, and it's not shorter, either. Is the win that it's using applicative combinators, not monadic?

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -110,6 +107,10 @@ import Distribution.Compat.Semigroup (Last' (..))
 
 import Data.Function (on)
 
+-- To be removed
```
quasicomputational: Is there a tracking bug for this, explaining when it'll happen/what it's blocked on?

Cabal/Distribution/Types/PackageId.hs
```diff
@@ -40,15 +37,61 @@ instance Pretty PackageIdentifier where
...
 
+-- |
+--
+-- >>> simpleParse "foo-bar-0" :: Maybe PackageIdentifier
+-- Just (PackageIdentifier {pkgName = PackageName "foo-bar", pkgVersion = mkVersion [0]})
+--
+-- >>> simpleParse "foo-bar" :: Maybe PackageIdentifier
+-- Just (PackageIdentifier {pkgName = PackageName "foo-bar", pkgVersion = mkVersion []})
+--
+-- /Note:/ Broken (too lenient, doesn't require full consumption)
```
quasicomputational: Issue for this? Or do we not care because this instance is going away? In which case, the comment should probably say so.

Cabal/Distribution/Simple/Setup.hs
phadej: #5580

Cabal/Distribution/Simple/BuildTarget.hs
phadej: (There are other doctest strings elsewhere too, there are 65 examples - which count setups though too)

Cabal/Distribution/Simple/Setup.hs
quasicomputational: OK, so they're going away in a later revision of this PR? That's fine too. I just don't want mystery 'to be removed' comments living in the code with no indication of when or how that's meant to happen.

phadej: ... in the follow up PR.

Cabal/Distribution/Types/PackageId.hs
phadej: Implied by #5579 . No `ReadP` => No `Text`

PR #5598, opened 2018-10-01
Track the exact source of ghc-*-options related warnings

### LINKED ISSUE #5342: cabal check mentions ghc-options when it could be ghc-shared-options
Title is self-explanatory. All the various `ghc-(shared-|prof-|)options` get smashed together in one list and if there's something warnable in there it's reported as `ghc-options`, which could send someone astray. Better would be keeping track of which field an option comes from and then reporting that.

nkly:
Previously, when running `cabal check` all the various ghc-\*-options flags were merged together, thus losing the information about the exact place of the warning. This PR implements separate checking of ghc-\*-options, which allows us to give users more precise warnings.

Fixes #5342

Checklist:

As for the testing approach, it is the same used for all other tests in the `Check` package.


nkly:
Looks like one of the jobs failed by timeout, and another one got an error during dependency resolution. Both failures are unrelated to the changes in the PR. Is it okay to merge?

quasicomputational:
I'd say to go for it. The solver error is happening on master, so it's not due to your code, and we bump right up against time limits on the OSX builders so it's inevitable that some builds get unlucky and time out.

PR #5602, opened 2018-10-03
Warn when failing to symlink an exe in new-install

fgaz:
---


cabal-install/changelog
```diff
@@ -2,6 +2,7 @@
 
 2.6.0.0 (current development version)
 	* New solver flag: '--reject-unconstrained-dependencies'. (#2568)
+	* 'new-install' now warns when failing to symlink an exe
```
hvr: NB: this is intended for a 2.4 point release

fgaz: Should I change 2.6 to 2.4.0.1? Or should I do that in the 2.4 branch?

hvr: maybe just drop changelog entry for the `master` branch; this reduces the conflict-potential during a cherry-pick to the 2.4 branch

hvr:
PS: while testing the PR I noticed a subtle (but harmless) bug in the `targetOkToOverwrite` function which is triggered when your `${HOME}/.cabal` folder is a symlink as then the `canonicalizePath`-based test gets a false negative.

PR #5604, opened 2018-10-04
Do register monolithic packages if they contain a buildable pub-lib

### LINKED ISSUE #5379: Builds depending on cpphs fail occasionally.
Occasionally, builds with `new-build` will fail with the following:

```
cabal: The following package dependencies were requested
--dependency='cpphs=cpphs-1.20.8-df266940be4ee1f49d22445541267998ca9452a1a9acff3e608ae8cee98d4624'
however the given installed package instance does not exist.
```

Builds with GHC 8.4.3 always work, but builds with past versions (such as 8.2.2, 8.0.2, etc.) will sometimes fail if cpphs has already been depended by a project using GHC 8.4.3.

This works on e.g. [shake-literate](https://hub.darcs.net/vmchale/shake-literate) with GHC 8.0.2, viz.

```
Resolving dependencies...
Build profile: -w ghc-8.0.2 -O1
 - shake-literate-0.1.0.0 (lib) (first run)
Configuring library for shake-literate-0.1.0.0..
cabal: The following package dependencies were requested
--dependency='cpphs=cpphs-1.20.8-5073bb53f1c45a8d04e54e55ee8e2500a032fca11038320e52b495d33ca6cbbf'
however the given installed package instance does not exist.
```

I'm using `cabal-head` from the PPA.

```
cabal-install version 2.3.0.0
compiled using version 2.3.0.0 of the Cabal library
```

hvr:
As the case of `cpphs-1.20.8` has shown in #5379, in cases when a
monolithic package gets installed due to its executable components
(i.e. `exe:cpphs`) into the store we *have* to register if there's a
buildable public library (i.e. `lib:cpphs`) that was built and installed
into the same store folder as otherwise this will cause build failures
once a target actually depends on `lib:cpphs`

This ought to fix #5379

(TODO: changelog entry when cherry-picking to 2.4)

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning/Types.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning/Types.hs
@@ -355,7 +355,24 @@ elabRequiresRegistration elab =
             -- register in all cases, but some Custom Setups will fall
             -- over if you try to do that, ESPECIALLY if there actually is
             -- a library but they hadn't built it.
-            build_target || any (depends_on_lib pkg) (elabBuildTargets elab)
+            --
+            -- However, as the case of `cpphs-1.20.8` has shown in
+            -- #5379, in cases when a monolithic package gets
+            -- installed due to its executable components
+            -- (i.e. exe:cpphs) into the store we *have* to register
+            -- if there's a buildable public library (i.e. lib:cpphs)
+            -- that was built and installed into the same store folder
+            -- as otherwise this will cause build failures once a
+            -- target actually depends on lib:cpphs.
+            build_target || (elabBuildStyle elab == BuildAndInstall &&
+                             Cabal.hasPublicLib (elabPkgDescription elab))
+            -- the next sub-condition below is currently redundant
+            -- (see discussion in #5604 for more details), but it's
+            -- being kept intentionally here as a safeguard because if
+            -- internal libraries ever start working with
+            -- non-per-component builds this condition won't be
+            -- redundant anymore.
+                         || any (depends_on_lib pkg) (elabBuildTargets elab)
   where
     depends_on_lib pkg (ComponentTarget cn _) =
         not (null (CD.select (== CD.componentNameToComponent cn)


ezyang:
I was asked in private correspondence if the test `any (depends_on_lib pkg) (elabBuildTargets elab)` is redundant with the new public library check. The answer is no. Here's when it matters:

1. You are building a package with a custom setup
2. This package has an executable which depends on an internal library
3. The package is from a sufficiently old version of Cabal, such that per-component builds are not in effect

We must register and install dynamic libraries of the internal (non-public library), and this will not trigger the `Cabal.hasPublicLib (elabPkgDescription elab)` condition.

After thinking about the proposed change a bit, I think it's OK.

I do think there's a more important, structural problem with the code we edited here. Basically, we shouldn't be making registration decisions based on selected *build targets*. A build target doesn't get fed into the Nix hash... so it shouldn't have an effect on how we install the package. I guess this is related to the strategy of pruning that local builds do.... but it seems very error prone.

hvr:
@ezyang ok, but

> Here's when it matters:
> 
>     1. You are building a package with a custom setup
> 
>     2. This package has an executable which depends on an internal library
> 
>     3. The package is from a sufficiently old version of Cabal, such that per-component builds are not in effect

But is this situation currently possible? It was my understanding that internal libraries are not supported unless per-component builds can be enabled. In other words, we don't support internal libraries when using custom setups, do we?

ezyang:
You're right, I had forgotten about this. So yes, I guess if that case is impossible, the test their is redundant. But it's a sort of accidental redundancy which would go away if you ever made internal libraries work with non-per-component build (though we probably won't ever do that.)

ezyang:
You're right, I had forgotten about this. So yes, I guess if that case is impossible, the test their is redundant. But it's a sort of accidental redundancy which would go away if you ever made internal libraries work with non-per-component build (though we probably won't ever do that.)

hvr:
@ezyang btw, a test-case fails legitimately; specifically the `cabal-testsuite/PackageTests/Exec` one which has a cabal file which forces legacy mode w/ build-type simple:

```
name:           my
version:        0.1
license:        BSD3
cabal-version:  >= 1.2
build-type:     Simple

library
    exposed-modules:    Foo
    build-depends:      base


executable my-executable
    main-is:            My.hs
    build-depends:      base
```

And the problem is specifically because in this situation despite being in legacy non-component mode we can still build single components:

```
$ cabal new-build my-executable -vnormal+nowrap
Resolving dependencies...
 - my-0.1 (exe:my-executable) (first run)
Configuring my-0.1...
Preprocessing executable 'my-executable' for my-0.1..
Building executable 'my-executable' for my-0.1..
[1 of 1] Compiling Main             ( My.hs, .../cabal-testsuite/PackageTests/Exec/dist-newstyle/build/x86_64-linux/ghc-8.2.2/my-0.1/build/my-executable/my-executable-tmp/Main.o )
Linking .../cabal-testsuite/PackageTests/Exec/dist-newstyle/build/x86_64-linux/ghc-8.2.2/my-0.1/build/my-executable/my-executable ...
cabal: Failed to build my-0.1. The failure occurred during the final install step. The exception was:
  dieVerbatim: user error (cabal: '/opt/ghc/bin/ghc-pkg' exited with an error:
my-0.1: Warning: haddock-interfaces: .../cabal-testsuite/PackageTests/Exec/dist-newstyle/build/x86_64-linux/ghc-8.2.2/my-0.1/doc/html/my/my.haddock doesn't exist or isn't a file
my-0.1: Warning: haddock-html: .../cabal-testsuite/PackageTests/Exec/dist-newstyle/build/x86_64-linux/ghc-8.2.2/my-0.1/doc/html/my doesn't exist or isn't a directory
my-0.1: cannot find any of ["Foo.hi","Foo.p_hi","Foo.dyn_hi"] (use --force to override)
my-0.1: cannot find any of ["libHSmy-0.1-inplace.a","libHSmy-0.1-inplace.p_a","libHSmy-0.1-inplace-ghc8.2.2.so","libHSmy-0.1-inplace-ghc8.2.2.dylib","HSmy-0.1-inplace-ghc8.2.2.dll"] on library path (use --force to override)
)
```

Any suggestion/idea?

ezyang:
Don't drop the condition I wrote? :) But yes, I think this is exactly the case (which I had forgotten about) which required the more subtle check I added.

hvr:
@ezyang if you look at the patch, I did *not* drop the condition! :-)

hvr:
merged to 2.4 via 517b49d50c266462374cfe92bc709239400dca31

I figured it out at last how to distinguish the inplace from the store-install case; the test-suite passes

PR #5605, opened 2018-10-04
Add DerivingVia to KnownExtensions

janpath:

I compiled it and tested it on a project, that uses `DerivingVia` as a `default-extension`.


23Skidoo:
Ugh, we should've had this in 2.4 - now it's too late, alas. Checking that `KnownExtensions` is up to date should be made a part of the release checklist.

23Skidoo:
Merged, thanks!

PR #5606, opened 2018-10-06
add new .cabal file field extra-dynamic-library-flavours

alpmestan:
It serves an almost identical purpose as extra-library-flavours, except
that the latter is only concerned with static library flavours, while
extra-dynamic-library-flavours lets us, in the case of libHSrts, build
and install a dynamic+threaded flavour of it.

I tested this in the context of hadrian, which uses the Cabal library to
build & register the RTS library, for which we need all sorts of fancy flavours, including
for dynamic libraries. With this patch and an accompanying change to `rts/rts.cabal.in`,
I manage to produce and install all the fancy dynamic RTS libraries I wanted.

---


phadej:
Parser-wise looks good. I'd rather add `^^^ availableSince [2,6] []` to FieldGrammar, then Check in Check.hs is redundant

quasicomputational:
There's also a Cabal file format changelog in `Cabal/doc`, which should have a mention of this.

quasicomputational:
Also, I'd rather keep the check, even if the parser learns to reject it based on cabal-version, since a `GenericPackageDescription` can come from other sources that might conceivably want to do a `check` operation (e.g., `dhall-to-cabal` produces them directly).

alpmestan:
@quasicomputational Oh, I didn't know about that one. Added a mention to the new field under version `2.5` (for now).

phadej:
dhall-to-cabal is good point. But I'd recommend it to do pretty&parse roundtrip anyway. It's the only way to be sure. And duplicating things is not good. (And we won't be able to test Check functionality already rejected by the parser - without manually constructing "invalid" GPDs).

alpmestan:
I added an `availableSince` annotation for `extra-dynamic-library-flavours` and adjusted the versions of some other such annotations to mention the _released_ version.

Cabal/Distribution/PackageDescription/Check.hs
```diff
@@ -1277,6 +1277,13 @@ checkCabalVersion pkg =
         ++ " and 'extra-library-flavours' requires the package "
         ++ " to specify at least 'cabal-version: >= 2.1'."
 
+  {-
+  , checkVersion [2,5] (not . null $ buildInfoField extraDynLibFlavours) $
+      PackageDistInexcusable $
+           "The use of 'extra-dynamic-library-flavours' requires the package "
+        ++ " to specify at least 'cabal-version: >= 2.5'."
```
phadej: @alpmestan I'd add the contents to `buildInfoField extraDynLibFlavours` to the error message, and see what happens.

I have no idea why it fails.

alpmestan: @phadej Done. It prints... nothing. See e.g [here](https://travis-ci.org/haskell/cabal/jobs/453007656#L2926).

alpmestan: Oh I get it. The predicate should be `any (not . null)`, not `not . null`.

alpmestan:
Woohoo, CI passes now.

alpmestan:
I'm not quite sure why the tests fail on appveyor, but it doesn't seem related to my patches. On the other hand, all the travis jobs (but the stack one) were green. I rebased and solved a conflict. It should still all be green, but let's see how things go.

What would be the process like to get this change to hit GHC HEAD? IIRC there is a special relationship between GHC and Cabal branches.

23Skidoo:
Once this is `master`, GHC's submodule would need to be bumped. Right now it tracks 2.4, I believe.

alpmestan:
There's a test failing on appveyor, but I have no idea what's going on:

```
cabal-testsuite\PackageTests\CustomPlain\cabal.test.hs                                                                            FAIL (6.93s)
$ "C:\ProgramData\chocolatey\lib\ghc\tools\ghc-8.0.2\bin\runghc.exe" "--" "--ghc-arg=-i" "--ghc-arg=-no-user-package-db" "--ghc-arg=-package-db" "--ghc-arg=C:\SR\ghc-8.0.2\package.db" "--ghc-arg=-package-db" "--ghc-arg=C:\projects\cabal\dist-newstyle\packagedb\ghc-8.0.2" "--ghc-arg=-package-id" "--ghc-arg=Cabal-2.5.0.0-inplace" "--ghc-arg=-package-id" "--ghc-arg=async-2.2.1-dbe6b46733e008b378100cb5e2b5eb1ccaf7981b" "--ghc-arg=-package-id" "--ghc-arg=base-4.9.1.0" "--ghc-arg=-package-id" "--ghc-arg=cabal-testsuite-2.5.0.0-inplace" "--ghc-arg=-package-id" "--ghc-arg=clock-0.7.2-022e0e52bdfb457aa825d723faa5322c265931c0" "--ghc-arg=-package-id" "--ghc-arg=exceptions-0.10.0-bf97183ec370757687bee79acf486ce62bb68e0e" "--ghc-arg=-package-id" "--ghc-arg=filepath-1.4.1.1" "--ghc-arg=-package-id" "--ghc-arg=optparse-appl_-0.14.3.0-f1755e613d499a3a6ef106ae49c02a405b52f028" "--ghc-arg=-package-id" "--ghc-arg=process-1.4.3.0" "--ghc-arg=-package-id" "--ghc-arg=transformers-0.5.2.0" "cabal-testsuite\PackageTests\CustomPlain\cabal.test.hs" "--builddir" "C:\projects\cabal\dist-newstyle\build\x86_64-windows\ghc-8.0.2\cabal-testsuite-2.5.0.0" "cabal-testsuite\PackageTests\CustomPlain\cabal.test.hs" "--with-cabal" "dist-newstyle\build\x86_64-windows\ghc-8.0.2\cabal-install-2.5.0.0\x\cabal\build\cabal\cabal.exe"
stdout:
+ "C:\projects\cabal\dist-newstyle\build\x86_64-windows\ghc-8.0.2\cabal-install-2.5.0.0\x\cabal\build\cabal\cabal.exe" "new-build" "-vverbose +markoutput +nowrap" "--builddir" "C:\projects\cabal\cabal-testsuite\PackageTests\CustomPlain\cabal.dist\work\.\dist" "--project-file" "cabal.project" "-j1" "-v1"
Resolving dependencies...
Build profile: -w ghc-8.0.2 -O1
 - old-locale-1.0.0.7 (lib) (requires download & build)
 - old-time-1.1.0.3 (lib:old-time) (requires download & build)
 - plain-0.1.0.0 (lib:plain) (first run)
ghc-pkg.exe: cannot create: C:\Users\appveyor\AppData\Roaming\cabal\store\ghc-8.0.2\package.db already exists
stderr:
*** Exception: Command "C:\projects\cabal\dist-newstyle\build\x86_64-windows\ghc-8.0.2\cabal-install-2.5.0.0\x\cabal\build\cabal\cabal.exe" "new-build" "-vverbose +markoutput +nowrap" "--builddir" "C:\projects\cabal\cabal-testsuite\PackageTests\CustomPlain\cabal.dist\work\.\dist" "--project-file" "cabal.project" "-j1" "-v1" failed.
Output:
Resolving dependencies...
Build profile: -w ghc-8.0.2 -O1
 - old-locale-1.0.0.7 (lib) (requires download & build)
 - old-time-1.1.0.3 (lib:old-time) (requires download & build)
 - plain-0.1.0.0 (lib:plain) (first run)
ghc-pkg.exe: cannot create: C:\Users\appveyor\AppData\Roaming\cabal\store\ghc-8.0.2\package.db already exists
*** unexpected failure for cabal-testsuite\PackageTests\CustomPlain\cabal.test.hs
```

The stack build on travis fails for an unrelated reason. Everything else is green.

Does the above failure look related to this patch for others? Anything else I should address?

phadej:
@alpmestan looks like #5681 ?

alpmestan:
@phadej Indeed, which would mean it's unrelated to my patch. Alright, then my patch is really ready for review :-)

alpmestan:
@adamse I think it only solves one part.

The code from this patch will (try to find and) install any dynamic flavours of your library that you mention in the `extra-dynamic-library-flavours` field, but _it does not make sure we build them_. It is not a problem in my use case because this is intended to be used in the cabal file for the `libHSrts` library, which is not built by Cabal. I just make sure I produce all the relevant dynamic flavours and point Cabal to it when it's time to install it all in the package db.

I don't know whether Cabal is currently equipped to _build_ non-vanilla dynamic flavours of libraries, I have to defer the answer to this to Cabal devs.

alpmestan:
Do we confirm that the failure I mentionned isn't this patch's fault? In which case, is there anything else I should do?

phadej:
Merged in https://github.com/haskell/cabal/pull/5715

alpmestan:
Thanks @phadej!

PR #5609, opened 2018-10-06
Check existence of packagedb before dumping it

### LINKED ISSUE #5516: new-install fails with empty store:  ghc-pkg: .../package.db: getDirectoryContents:openDirStream: does not exist (No such file or directory)

fgaz:
Fixes #5516

To do:

* [x] Better commit message/comment explaining why not failing (even with a Maybe) is ok

/cc @hvr

### DIFF
--- a/Cabal/Distribution/Compat/Directory.hs
+++ b/Cabal/Distribution/Compat/Directory.hs
@@ -1,8 +1,17 @@
 {-# LANGUAGE CPP #-}
 
-module Distribution.Compat.Directory (listDirectory, makeAbsolute) where
+module Distribution.Compat.Directory
+( listDirectory
+, makeAbsolute
+, doesPathExist
+) where
 
+#if MIN_VERSION_directory(1,2,7)
+import System.Directory as Dir hiding (doesPathExist)
+import System.Directory (doesPathExist)
+#else
 import System.Directory as Dir
+#endif
 #if !MIN_VERSION_directory(1,2,2)
 import System.FilePath as Path
 #endif
@@ -25,3 +34,11 @@ makeAbsolute p | Path.isAbsolute p = return p
     return $ cwd </> p
 
 #endif
+
+#if !MIN_VERSION_directory(1,2,7)
+
+doesPathExist :: FilePath -> IO Bool
+doesPathExist path = (||) <$> doesDirectoryExist path <*> doesFileExist path
+
+#endif
+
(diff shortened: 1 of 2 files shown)


23Skidoo:
Can we perhaps stop in

23Skidoo:
Silently filtering invalid parameters out is a bit surprising, I'd prefer if it printed a warning at least. Why are we invoking this function incorrectly in the first place? Perhaps we can stop doing that?

fgaz:
Well, since a non-existent pkgdb technically does not contain any package I thought returning an [] was ok.

The alternative is adding a pkgdb init abstraction and creating the pkgdb when running new-install. I'll open another pr

typedrat:
> Well, since a non-existent pkgdb technically does not contain any package I thought returning an [] was ok.

That's the semantic meaning that I'd want here, perhaps with a warning printed.

fgaz:
I added a warning, but note that since cabal-install uses a reduced verbosity version of getInstalledPackages, it will only show with -v2 or more (which may be what we want)

23Skidoo:
> Well, since a non-existent pkgdb technically does not contain any package I thought returning an [] was ok.

OK, makes sense, let's merge this, but add a Haddock comment.

fgaz:
Cherry-picked on 2.4

PR #5620, opened 2018-10-14
Setup: Don't skip repl for indefinite components

### LINKED ISSUE #5619: Backpack - unable to launch `cabal new-repl` for an indefinite library
`cabal new-repl` for a signature-only component doesn't launch GHCi. I believe it should be possible.

The command just prints the following:

```
$ cabal new-repl
Resolving dependencies...
 - IndefRepl-0.1.0.0 (lib) (configuration changed)
Configuring library for IndefRepl-0.1.0.0..
Preprocessing library for IndefRepl-0.1.0.0..
```

and exits.

Test case:

`cabal.project`:  
```
packages: .
```

`IndefRepl.cabal`:  
```
name:                IndefRepl
version:             0.1.0.0
build-type:          Simple
cabal-version:       >=1.25

library
  signatures:          Sig
  build-depends:       base
```

`Sig.hsig`:  
```
signature Sig where
foo :: ()
```

Tested with cabal-install `2.4.0.0` and HEAD (a6415a8f303ea9c2713ccc6a07a3c8d3397f6e1b).

zyla:
For some reason repl invocation was skipped for indefinite components.
However, GHCi seems to handle them just fine.

Fixes #5619.

How I tested this change: Using the test case from issue #5619. Ran `cabal new-repl` and verified that it does indeed launch GHCi.
I couldn't figure out how to add this as an automated test to the suite, as it required somehow killing GHCi to avoid hanging indefinitely. The simplest options I could think of is passing `:quit` as input, but I couldn't find facilities for passing stdin in the test machinery.

---

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -655,7 +655,7 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
            unless forRepl $ whenProfLib   (runGhcProgIfNeeded   profCxxOpts)
       | filename <- cxxSources libBi]
 
-  when has_code . ifReplLib $ do
+  ifReplLib $ do
     when (null (allLibModules lib clbi)) $ warn verbosity "No exposed modules"
     ifReplLib (runGhcProg replOpts)
 
--- a/cabal-install/changelog
+++ b/cabal-install/changelog
@@ -2,6 +2,7 @@
 
 2.6.0.0 (current development version)
 	* New solver flag: '--reject-unconstrained-dependencies'. (#2568)
+	* "cabal new-repl" now works for indefinite (in the Backpack sense) components.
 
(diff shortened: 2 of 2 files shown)


zyla:
> GHCi seems to handle them just fine.

To clarify: it works "just fine" if you don't ask it to do impossible things (i.e. evaluate expressions). However, `:reload`ing and `:type` do seem to work.

ezyang:
Wow, I'm surprised this works (speaking as the person who wrote all this in the first place :) SGTM.

zyla:
Thank you! It worked so well I didn't know this use case wasn't meant to be supported :)

However, there are some rough edges on the GHC side I didn't notice at first - for example, it sometimes panics when you try to evaluate code. Should I report them on the GHC Trac?

quasicomputational:
Panic reports would be very welcome on the GHC Trac, I'm certain.

hvr:
backported to 2.4 via 3abee6332eb9918f673fb5656f2b138d36f699cf

PR #5633, opened 2018-10-21
Fix monolithic inplace build tool PATH

### LINKED ISSUE #5104: "In-place"d build-tools not made available via $PATH
It appears that `build-tools` which become in-place, don't end up being locatable via `$PATH`.

I noticed this when trying to `cabal new-build exe:cabal` in Git's cabal-2.0 branch:

```
$ cnb exe:cabal
 - pretty-show-1.6.16 (lib) (first run)
 - cabal-install-2.0.0.1 (exe:cabal) (first run)
Configuring library for pretty-show-1.6.16..
cabal: The program 'happy' is required but it could not be found.

cabal: Failed to build pretty-show-1.6.16 (which is required by cabal-install-2.0.0.1). The failure occurred during the configure step.
```

And if we look at plan.json, we can clearly see that `pretty-show` depends on the inplace `happy-1.19.8-inplace` build-tool:

```json
    {
      "type": "configured",
      "id": "pretty-show-1.6.16-inplace",
      "pkg-name": "pretty-show",
      "pkg-version": "1.6.16",
      "flags": {},
      "style": "inplace",
      "pkg-src-sha256": "dbee8476bf12ce5bd991d0a52c1340802e07bb706c68c1a7911a38db16ac0350",
      "dist-dir": "/stuff3/work/GitHub/cabal-2.0/dist-newstyle/build/x86_64-linux/ghc-8.2.2/pretty-show-1.6.16",
      "depends": [
        "array-0.5.2.0",
        "base-4.10.1.0",
        "haskell-lexer-1.0.1-e064fdcf35c9d26edcab1e4ec89f91c2f69f01338016ce35cbe105ffc3720962",
        "pretty-1.1.3.3",
        "filepath-1.4.1.2",
        "ghc-prim-0.5.1.1"
      ],
      "exe-depends": [
        "happy-1.19.8-inplace"
      ],
      "component-name": "lib"
    },
```

whose executable can be found in `dist-newstyle`:

```json
    {
      "type": "configured",
      "id": "happy-1.19.8-inplace",
      "pkg-name": "happy",
      "pkg-version": "1.19.8",
      "flags": {
        "small_base": true
      },
      "style": "inplace",
      "pkg-src-sha256": "4df739965d559e48a9b0044fa6140241c07e8f3c794c6c0a6323024fd7f0d3a0",
      "dist-dir": "/stuff3/work/GitHub/cabal-2.0/dist-newstyle/build/x86_64-linux/ghc-8.2.2/happy-1.19.8",
      "components": {
        "exe:happy": {
          "depends": [
            "base-4.10.1.0",
            "array-0.5.2.0",
            "containers-0.5.10.2",
            "mtl-2.2.1-97763c4fed4fb8db9480ab17817d7bc91517cc666c409844d1b89dd550f4d270"
          ],
          "exe-depends": [],
          "bin-file": "/stuff3/work/GitHub/cabal-2.0/dist-newstyle/build/x86_64-linux/ghc-8.2.2/happy-1.19.8/build/happy/happy"
        },
        "setup": {
          "depends": [
            "Cabal-2.0.1.1-inplace",
            "base-4.10.1.0",
            "directory-1.3.0.2",
            "filepath-1.4.1.2"
          ],
          "exe-depends": []
        }
      }
    },
```

fgaz:
Monolithic inplace packages have a directory for each exe, so we allow
multiple paths and return all the exe paths of the package.

Fixes #5104

/cc @hvr

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning.hs
@@ -1435,7 +1435,8 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
                 compExeDependencyPaths =
                     [ (annotatedIdToConfiguredId aid', path)
                     | aid' <- cc_exe_deps cc0
-                    , Just path <- [Map.lookup (ann_id aid') exe_map1]]
+                    , Just paths <- [Map.lookup (ann_id aid') exe_map1]
+                    , path <- paths ]
                 elab_comp = ElaboratedComponent {..}
 
             -- 3. Construct a preliminary ElaboratedConfiguredPackage,
@@ -1520,10 +1521,10 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
                 external_exe_dep_sids ++ external_lib_dep_sids
 
             external_exe_map = Map.fromList $
-                [ (getComponentId pkg, path)
+                [ (getComponentId pkg, paths)
                 | pkg <- external_exe_dep_pkgs
-                , Just path <- [planPackageExePath pkg] ]
-            exe_map1 = Map.union external_exe_map exe_map
+                , let paths = planPackageExePaths pkg ]
+            exe_map1 = Map.union external_exe_map $ fmap (\x -> [x]) exe_map
 
             external_lib_cc_map = Map.fromListWith Map.union
                                 $ map mkCCMapping external_lib_dep_pkgs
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1522,7 +1522,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
             external_exe_map = Map.fromList $
                 [ (getComponentId pkg, path)
                 | pkg <- external_exe_dep_pkgs
-                , Just path <- [planPackageExePath pkg] ]
+                , path <- planPackageExePaths pkg ]
```
fgaz: Hmm I just realized only the first path will be picked, since we convert to a `Map ComponentId FilePath`. Which is a problem if some build tool has multiple exes. I'll change the map to `Map ComponentId [FilePath]`.

hvr:
I can confirm this fixes the original repro-case from #5104; I'm currently dogfooding this PR locally

hvr:
@23Skidoo fwiw, I couldn't find any empirical problems yet; I'd suggest merging

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1520,10 +1521,10 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
                 external_exe_dep_sids ++ external_lib_dep_sids
 
             external_exe_map = Map.fromList $
-                [ (getComponentId pkg, path)
+                [ (getComponentId pkg, paths)
                 | pkg <- external_exe_dep_pkgs
-                , Just path <- [planPackageExePath pkg] ]
-            exe_map1 = Map.union external_exe_map exe_map
+                , let paths = planPackageExePaths pkg ]
+            exe_map1 = Map.union external_exe_map (pure <$> exe_map)
```
23Skidoo: I'd rather be a bit more explicit here and write this as `M.map singleton exe_map` (if there's no `singleton :: a -> [a]` in `cabal-install`, we can add it to `Distribution.Compat.Prelude` or a utils module. Or just use `(\a -> [a])`.

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1583,24 +1584,30 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
...
+    planPackageExePaths =
         -- Pre-existing executables are assumed to be in PATH
         -- already.  In fact, this should be impossible.
-        InstallPlan.foldPlanPackage (const Nothing) $ \elab -> Just $
+        InstallPlan.foldPlanPackage (const []) $ \elab ->
             binDirectoryFor
                 distDirLayout
                 elaboratedSharedConfig
-                elab $
+                elab <$>
```
23Skidoo: I'd use `map` instead of `<$>` here as well.

fgaz: Like `map (binDirectoryFor ...) case ...`? or inline `map`? The first is a bit hard to follow imo, and inline map is just `<$>` (it's clear that we are acting on a `[]` from the type sig)... maybe a comment instead?

Or what about a binding?
```suggestion
                elab
                <$> executables
        where
            executables =
```

23Skidoo:
@fgaz Both suggestions sound fine, use your discretion.

cabal-install/Distribution/Client/ProjectPlanning.hs
23Skidoo: I was thinking of `` `map` `` instead of `<$>`.

fgaz:
Cherry picked on 2.4

PR #5635, opened 2018-10-23
Add message to alert user to potential package casing errors.

### LINKED ISSUE #5601: we still have poor error messages for mismatched case for package names (in october 2018 / cabal 2.4)
as far as I know, on the hackage side we dont allow new package names to use uppercase ascii, but that aside, cabal-install could stand much better error messages for this class of typoes

an example from today: 
```
$ cabal v2-install earley megaparsec
Resolving dependencies...
[__0] unknown package: earley (user goal)
[__0] fail (backjumping, conflict set: earley)
goals I've had most trouble fulfilling: earley

```

i had to google to notice that the hackage package name has its first letter capitalized `Earley`.

I think our error messages when we can't find any package could be better... 
This does get tricky for off hackage packages, but at least in the context of hackage we can assume ASCII only .. 

the error message could be an analogue of the "`hatt` is an unbound/wrongly typed variable, perhaps you meant `hott`?" style error diagnostic many language implementations provide

cc @hvr thoughts?

typedrat:
As pointed out to me by @hvr, this bit of helpful error reporting was not present in new-install.

Fixes #5601.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -31,10 +31,11 @@ import Distribution.Client.Setup
 import Distribution.Solver.Types.ConstraintSource
          ( ConstraintSource(..) )
 import Distribution.Client.Types
-         ( PackageSpecifier(..), PackageLocation(..), UnresolvedSourcePackage )
+         ( PackageSpecifier(..), PackageLocation(..), UnresolvedSourcePackage
+         , SourcePackageDb(..) )
 import qualified Distribution.Client.InstallPlan as InstallPlan
 import Distribution.Package
-         ( Package(..), PackageName, mkPackageName )
+         ( Package(..), PackageName, mkPackageName, unPackageName )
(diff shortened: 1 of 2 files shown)


cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -358,14 +372,50 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, newInstal
           | Just (pkg :: PackageId) <- simpleParse pkgName = return pkg
           | otherwise = die' verbosity ("Invalid package ID: " ++ pkgName)
       packageIds <- mapM parsePkg targetStrings
+      
+      cabalDir <- getCabalDir
```
23Skidoo: I'd make this whole bit a separate function with a name like `reportCasingErrors` or `checkForUnknownPackages`.

hvr:
btw, this kinda fixes #5601

typedrat:
Cherry-picked to 2.4.

Issue #5491, opened 2018-08-02
cabal new-install silently fails if symlink target already exists


hvr:
Currently, when you e.g. `cabal new-install hsc2hs --symlink-bindir=/tmp/bin`, and something exists already in `/tmp/bin`, `new-install` will claim to

    Symlinking hsc2hs

even though it failed to update the symlink.

E.g.

```
$ cabal new-install hsc2hs-0.68.3 --symlink-bindir=/tmp/bin/
Resolving dependencies...
Up to date
Symlinking hsc2hs

$ ls -l /tmp/bin/
total 0
lrwxrwxrwx 1 hvr hvr 127 Aug  2 11:17 hsc2hs -> ../../home/hvr/.cabal/store/ghc-8.2.2/hsc2hs-0.68.3-098a68e230c98f65b69ba9c636ab16a4395deacc40aff03647d85d54a5ad7aeb/bin/hsc2hs
```

that's ok so far, as the end-result was in fact having a symlink pointing to `hsc2hs-0.68.3`; however, when I request to `new-install` again which is expected to results in a *different* symlink, `new-install` silently fails to:

```
$ cabal new-install hsc2hs-0.68.2 --symlink-bindir=/tmp/bin/
Resolving dependencies...
Build profile: -w ghc-8.2.2 -O1
In order, the following will be built (use -v for more details):
 - hsc2hs-0.68.2 (exe:hsc2hs) (requires download & build)
Downloading  hsc2hs-0.68.2
Downloaded   hsc2hs-0.68.2
Starting     hsc2hs-0.68.2 (exe:hsc2hs)
Building     hsc2hs-0.68.2 (exe:hsc2hs)
Installing   hsc2hs-0.68.2 (exe:hsc2hs)
Completed    hsc2hs-0.68.2 (exe:hsc2hs)
Symlinking hsc2hs

$ ls -l /tmp/bin/
total 0
lrwxrwxrwx 1 hvr hvr 127 Aug  2 11:17 hsc2hs -> ../../home/hvr/.cabal/store/ghc-8.2.2/hsc2hs-0.68.3-098a68e230c98f65b69ba9c636ab16a4395deacc40aff03647d85d54a5ad7aeb/bin/hsc2hs
```

Consequently, `new-install` ought to at very least warn if not even error (or even better yet overwrite), when it cannot produce the desired final state of the symlink; specifically,

I'd expect two subsequent invocations of e.g.

```
[...]

PR #5638, opened 2018-10-25
new-install: error on existing symlink, overwrite option

fgaz:
Phase 2 of the fix for #5491. Followup to #5602.

Erroring out is safer and more explicit.

Next, we'll have to check for error/no-op as early as possible (currently we build everything first).

/cc @hvr

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -632,7 +632,7 @@ symlinkBuiltExe :: Verbosity -> OverwritePolicy
                 -> UnqualComponentName
                 -> IO Bool
 symlinkBuiltExe verbosity overwritePolicy sourceDir destDir exe = do
-  notice verbosity $ "Symlinking " ++ prettyShow exe
+  notice verbosity $ "Symlinking '" <> prettyShow exe <> "'"
   symlinkBinary
     overwritePolicy
     destDir
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -625,7 +625,7 @@ symlinkBuiltPackage verbosity overwritePolicy
                    (mkSourceBinDir pkg) destDir exe
       let errorMessage = case overwritePolicy of
                   NeverOverwrite ->
-                    "Symlink for '" <> prettyShow exe <> "' already exists. "
+                    "Path '" <> (destDir </> prettyShow exe) <> "' already exists. "
                     <> "Use --overwrite-policy=always to overwrite."
                   -- This shouldn't even be possible, but we keep it in case
                   -- symlinking logic changes
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -73,7 +73,7 @@ import Distribution.Client.DistDirLayout
 import Distribution.Client.RebuildMonad
          ( runRebuild )
 import Distribution.Client.InstallSymlink
-         ( symlinkBinary )
+         ( OverwritePolicy(..), symlinkBinary )
 import Distribution.Simple.Setup
          ( Flag(Flag), HaddockFlags, fromFlagOrDefault, flagToMaybe, toFlag
          , trueArg, configureOptions, haddockOptions, flagToList )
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -149,6 +155,11 @@ newInstallOptions _ =
     "Set the environment file that may be modified."
     ninstEnvironmentPath (\pf flags -> flags { ninstEnvironmentPath = pf })
     (reqArg "ENV" (succeedReadE Flag) flagToList)
+  -- TODO choose a name. --force-overwrite, --overwrite-symlink, --overwrite...
+  , option [] ["force-overwrite"]
```
fgaz: Is `--force-overwrite` ok?

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -131,14 +131,20 @@ import System.FilePath
...
 defaultNewInstallFlags :: NewInstallFlags
 defaultNewInstallFlags = NewInstallFlags
   { ninstInstallLibs = toFlag False
   , ninstEnvironmentPath = mempty
+  , ninstForceOverwrite = toFlag False
   }
 
+boolToOverwritePolicy :: Bool -> OverwritePolicy
+boolToOverwritePolicy True  = DoOverwrite
+boolToOverwritePolicy False = DontOverwrite
```
fgaz: Still trying to figure out how to just have a `Flag OverwritePolicy` without having to reimplement trueArg

cabal-install/Distribution/Client/CmdInstall.hs
hvr: maybe we could go with a more generic 

    --overwrite-policy[=always,never]

which has the benefit of 

1. being extendable to e.g.

        --overwrite-policy[=always,always_warn,ask,never]`

and 

2. allowing for changing what the default should be (e.g. via `~/.cabal/config`) -- which seems to be a good idea given we had some controversy about in the issue discussion

fgaz: 1. good idea, doing that.
0. What about scripts? It's another flag that has to be overridden to ensure a consistent behavior.

fgaz: Didn't do the 2nd.
I'll open another issue for that.

hvr:
I gave this PR a test-run, and it appears to work reasonably.

There's some minor aesthetic issues with some messages which e.g. claim that a "Symlink" already exists even though it's not a symlink but a regular file... but there's a few more (not high prio) UI details we might want to improve/fix there that this can be addressed by a future PR

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -155,12 +151,22 @@ newInstallOptions _ =
     "Set the environment file that may be modified."
     ninstEnvironmentPath (\pf flags -> flags { ninstEnvironmentPath = pf })
     (reqArg "ENV" (succeedReadE Flag) flagToList)
-  -- TODO choose a name. --force-overwrite, --overwrite-symlink, --overwrite...
-  , option [] ["force-overwrite"]
-    "Overwrite an existing symlink."
-    ninstForceOverwrite (\v flags -> flags { ninstForceOverwrite = v })
-    trueArg
+  , option [] ["overwrite-policy"]
```
23Skidoo: We can also make it work similarly to `--allow-newer`, i.e. make `--overwrite-symlink` w/o arguments be a shortcut for `--overwrite-symlink=always`.

hvr: @23Skidoo btw, didn't we want to deprecate the option-less `--allow-newer` variant? :)

hvr: oh and btw, I'd like to point out that during dogfooding I started using an abbrev, like e.g.

    cabal v2-install alex --overwrite=always

or even

    cabal v2-install alex --over=always

hvr: @23Skidoo after thinking about this,  the decision whether to support an-argument less `--overwrite-...` hinges on what default we want to pick. I.e. if we default to `always` it'd be confusing to have a seemingly no-op alias which just reaffirms the default setting.

23Skidoo: > thus I'd suggest to postpone adding a shortcut to this PR

Sure. Let's merge.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked to 2.4.

PR #5651, opened 2018-10-30
Cabal: Allow time-1.9

bgamari:


hvr:
@bgamari shouldn't this rather be against master? Btw, we try to avoid extending bounds to include releases of dependencies we don't cover in CI

bgamari:
> shouldn't this rather be against master?

I wanted to keep the bump minimal so I applied the 

> Btw, we try to avoid extending bounds to include releases of dependencies we don't cover in CI

I'm not sure what you mean. `time-1.9` has been available on Hackage for nearly a year. Why wouldn't this be covered in CI?

23Skidoo:
Merged, thanks! Also cherry-picked into `master`.

bgamari:
Thanks!

PR #5652, opened 2018-11-01
Filter out ambiguous builddir flag

### LINKED ISSUE #5550: option `--builddir' is ambiguous
I want cabal to put its build files in a custom directory. However using `builddir` results in an error:

```
$ /opt/cabal/bin/cabal-head new-install --builddir=/tmp/dist
cabal-head: option `--builddir' is ambiguous; could be one of:
--builddir=DIR The directory where Cabal puts generated build files (default
dist)
--builddir=DIR The directory where Cabal puts generated build files (default
dist)
```

I'm unsure if `--builddir` is even a valid option for new-build, but it seems to be according to the docs.

fgaz:
It conflicted with the flag with the same name from configFlags

Fixes #5550

/cc @hvr

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -213,9 +213,9 @@ installCommand = CommandUI
                  . optionName) $
                                installOptions showOrParseArgs)
        ++ liftOptions get4 set4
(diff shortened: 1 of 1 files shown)


23Skidoo:
Looks good.

fgaz:
Cherry-picked to 2.4

PR #5657, opened 2018-11-02
Link to 'developing packages' from new-build docs.

m-renaud:
This page is often surfaces above the "Developing Packages Quickstart" section in search results (for queries including "cabal new project") but a beginner ending up here may not know how to set up their package before using the commands here. This adds a link to the relevant section in the docs that walks through how to set up a package so they don't need to go looking for it themselves.

This is a docs-only change.

[ci skip]


23Skidoo:
Merged, thanks!

PR #5658, opened 2018-11-03
Add `-w`/`--with-compiler` flag to `cabal init`

### LINKED ISSUE #4936: "cabal init" should support support --with-ghc
"cabal init" should be consistent with the other cabal commands and allow "--with-ghc":

```
$ cabal init
cabal: The program 'ghc' version >=6.11 is required but it could not be found.
$ cabal init --with-ghc=/opt/ghc-8.2.2/bin/ghc
cabal: unrecognized 'init' option `--with-ghc=/opt/ghc-8.2.2/bin/ghc'
```

### LINKED ISSUE #5654: ``cabal init`` doesn't allow specifying an alternative location for GHC executable
For many invocations of ``cabal``, I can pass a ``-w`` argument to specify where GHC can be found, which is very useful if you have local GHCs set up in your home directory. However, ``cabal init`` doesn't allow this, instead opting to fail and complain about a missing GHC. Would it be possible to have ``cabal init`` accept a ``-w`` flag too?

hvr:
Fixes #5654 
Fixes #4936

----

Besides giving you the ability to conveniently pick a different `ghc` which you could workaround by e.g.

```
PATH=/opt/ghc/8.6.2/bin:$PATH cabal init
```

via a simpler, more explicit (because it doesn't rely on the indirection via $PATH to pick the right exe), and more idiomatic

```
cabal init -w ghc-8.6.2
```

if you have a setup where all `ghc` versions are symlinked version-suffixed in your $PATH already.

However, there's also the case where you don't have *any* `ghc` in your PATH, but *only* `ghc-<version>` symlinks; this is e.g. the default setup scheme promoted by `ghcup` and for this case, there's not necessarily any workaround (other than creating a `cabal.config` file or modifying your global cabal config). Or an even more pathological case is

```
cabal init -w /work/ghc/inplace/bin/ghc-stage2
```

---

TODO:

### DIFF
--- a/cabal-install/Distribution/Client/Init/Types.hs
+++ b/cabal-install/Distribution/Client/Init/Types.hs
@@ -69,6 +69,8 @@ data InitFlags =
               , sourceDirs   :: Maybe [String]
               , buildTools   :: Maybe [String]
 
+              , initHcPath    :: Flag FilePath
+
               , initVerbosity :: Flag Verbosity
               , overwrite     :: Flag Bool
               }
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -2295,6 +2295,14 @@ initCommand = CommandUI {
         (reqArg' "TOOL" (Just . (:[]))
                         (fromMaybe []))
 
+        -- NB: this is a bit of a transitional hack and will likely be
+        -- removed again if `cabal init` is migrated to the v2-* command
+        -- framework
+      , option "w" ["with-compiler"]
+        "give the path to a particular compiler"
+        IT.initHcPath (\v flags -> flags { IT.initHcPath = v })
+        (reqArgFlag "PATH")
+
       , optionVerbosity IT.initVerbosity (\v flags -> flags { IT.initVerbosity = v })
       ]
   }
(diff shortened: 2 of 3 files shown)


hvr:
@23Skidoo is this hack tolerable? if yes, I'd vote for backporting to 2.4 as it's low-risk and trival

23Skidoo:
Looks fine and not really hacky to me, unless I'm missing something.

hvr:
@23Skidoo maybe I've just got a different threshold for what I consider a hack... :-)

23Skidoo:
How would a non-hacky version look like? It doesn't make sense for `init` to inherit all of `configure` flags IMO.

m-renaud:
Will this create a `cabal.project.local` like `cabal v2-configure -w ghc-ver` would, or does this still require running the configure to pin the version for the project?

hvr:
@m-renaud for now no `cabal.project.*` generation; I'd also keep this PR agnostic to that

However, I wanted to see support for for creating `cabal.project` skeletons added to `cabal init`; but this ought to be controlled by a question "Do you want to create a cabal.project file?", and we'd want to create a couple of comments in there (again controllable) with common fields people might want to uncomment or tweak and a pointer to the docs or a cheatsheet

hvr:
@23Skidoo Can I take your comment as approval to merge (after I've update docs & changelog)?

23Skidoo:
@hvr Sure, go ahead.

hvr:
(cherry-picked to 2.4 via 878642e1e6014ed8424cefcdfbc41879516ab62b + 08019e5c3352d1f02cfd3c371b4b7e7e776397b4)

PR #5659, opened 2018-11-03
Add 'optional-target' field to plan.json

DanielG:
Having this enables tooling to give better error messages when build output is
missing. cabal-install can choose to include optional components in the build
plan even though they weren't requested explicitly by the user. Currently
tooling would have to assume missing build output after a `v2-build all` call
means the component is currently disabled, which is just not very clean.

---

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanOutput.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanOutput.hs
@@ -157,7 +157,7 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
                     [ "depends"     J..= map (jdisplay . confInstId) ldeps
                     , "exe-depends" J..= map (jdisplay . confInstId) edeps
                     ] ++
-                    requestedByDefaultJ c ++
+                    optionalTargetJ c ++
                     bin_file c)
                   | (c,(ldeps,edeps))
                       <- ComponentDeps.toList $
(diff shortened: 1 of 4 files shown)


cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -155,7 +155,9 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
             let components = J.object $
                   [ comp2str c J..= (J.object $
                     [ "depends"     J..= map (jdisplay . confInstId) ldeps
-                    , "exe-depends" J..= map (jdisplay . confInstId) edeps ] ++
+                    , "exe-depends" J..= map (jdisplay . confInstId) edeps
+                    , "requested-by-default" J..= requestedByDefaultJ c ] ++
```
hvr: Since most of the time this will be redundantly `requested-by-default: True` and I intend `plan.json` to be manually optimized to reduce redundancy and noise, I'd rather this was something like

```suggestion
                    ] ++ [ "optional" J..= True | not (requestedByDefaultJ c) ] ++
```

cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -166,9 +168,15 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
             ["depends"     J..= map (jdisplay . confInstId) (elabLibDependencies elab)
             ,"exe-depends" J..= map jdisplay (elabExeDependencies elab)
             ,"component-name" J..= J.String (comp2str (compSolverName comp))
+            ,"requested-by-default" J..= requestedByDefaultJ (compSolverName comp)
```
hvr: dito

```suggestion
            ] ++ [ "optional" J..= True | not (requestedByDefaultJ (compSolverName comp))
```

23Skidoo:
@hvr OK, noted.

hvr:
@23Skidoo I've rebased this and renamed the flag; I've been dogfooding this a bit and it looks good to me; 

IOW, it's ready to merge from my POV

phadej:
Is there a ticket about buildinfo.json?

23Skidoo:
Merged, thanks! Also cherry-picked to `2.4`.

PR #5670, opened 2018-11-06
Enable bootstrap script to download and use local unpacked dependencies

watashi:
* `bootstrap.sh` can use local tarball, this change enhances it by using
  the unpacked local directory if that exists, so it's possible to play
  with a patch of a dependency locally.
* a flag `--no-install` is added, so we can fetch and unpack all
  dependencies without actually installing them. this will allow us to
  later patch or build standalone cabal offline.

Test plan:

`./bootstrap.sh --no-install` downloads a bunch of packages and
I can run `./bootstrap.sh` later to build cabal without network.

---

### DIFF
--- a/cabal-install/bootstrap.sh
+++ b/cabal-install/bootstrap.sh
@@ -137,6 +137,9 @@ while [ "$#" -gt 0 ]; do
     "--no-doc")
       NO_DOCUMENTATION=1
       shift;;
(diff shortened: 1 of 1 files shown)


watashi:
cc: @simonmar

23Skidoo:
Merged, thanks!

PR #5671, opened 2018-11-07
Don't install anything if --dry is used

### LINKED ISSUE #5667: `cabal new-install --dry` creates a dangling symlink
If I combine `cabal new-install` with the `--dry` flag, I would expect it to show me the build plan and do nothing else. In practice, it actually does something: it constructs a symlink to the package to be installed. Even worse, this symlink is dangling!

```
$ cabal new-install pointful -w /opt/ghc/8.4.4/bin/ghc --dry
Resolving dependencies...
Build profile: -w ghc-8.4.4 -O1
 - haskell-src-exts-simple-1.20.0.0 (lib) (requires build)
 - pointful-1.0.11.0 (lib) (requires build)
 - pointful-1.0.11.0 (exe:pointful) (requires build)
Symlinking pointful

$ ls -alh
total 106M
drwxr-xr-x  3 rgscott students 4.0K Nov  6 11:11 .
drwxr-xr-x 10 rgscott students 4.0K Nov  5 13:02 ..
<elided>
lrwxrwxrwx  1 rgscott students  114 Nov  6 11:11 pointful -> ../store/ghc-8.4.4/pointful-1.0.11.0-52b63f7a657099a01c78fd5256be119f8ffdd2c599ae90a0655dc7bf22d2283e/bin/pointful

$ cd ../store/ghc-8.4.4/pointful-1.0.11.0-52b63f7a657099a01c78fd5256be119f8ffdd2c599ae90a0655dc7bf22d2283e/bin
-bash: cd: ../store/ghc-8.4.4/pointful-1.0.11.0-52b63f7a657099a01c78fd5256be119f8ffdd2c599ae90a0655dc7bf22d2283e/bin: No such file or directory
```

fgaz:
Fixes #5667

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -26,7 +26,7 @@ import Distribution.Client.CmdErrorMessages
 import Distribution.Client.CmdSdist
 
 import Distribution.Client.Setup
-         ( GlobalFlags(..), ConfigFlags(..), ConfigExFlags, InstallFlags
+         ( GlobalFlags(..), ConfigFlags(..), ConfigExFlags, InstallFlags(..)
          , configureExOptions, installOptions, liftOptions )
 import Distribution.Solver.Types.ConstraintSource
          ( ConstraintSource(..) )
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -573,6 +573,7 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, newInstal
     globalConfigFlag = projectConfigConfigFile (projectConfigShared cliConfig)
     overwritePolicy = fromFlagOrDefault NeverOverwrite
                         $ ninstOverwritePolicy newInstallFlags
+    dryRun = fromFlagOrDefault False $ installDryRun installFlags
```
fgaz: I just found out about buildSettingDryRun, which is probably better than this. Fixing it.

fgaz:
> Maybe I'd make the when installLibs / when (not installLibs) blocks into helper procedures.

...I'll add that in the [potential refactors](https://github.com/haskell/cabal/wiki/Potential-Refactors) wiki page

fgaz:
Cherry-picked to 2.4

Issue #5644, opened 2018-10-26
Add NewInstallFlags to SavedConfig


fgaz:
Being able to configure the defaults would be nice.

Like it's done for `ConfigExFlags`.
Maybe change them from `*ExFlags` to `*ClientFlags` or `Client*Flags`

This came up in https://github.com/haskell/cabal/pull/5638#discussion_r228170109.

Also maybe do this for all other new-build commands

PR #5672, opened 2018-11-08
Prompt overwrite policy

fredefox:
When linking binaries into a globally accessible location on the machine `--overwrite-policy=always` is needed if an existing file sits where we are trying to create a link. This allows for a prompt-the-user strategy.

---

  NA: Couldn't find any existing documentation about `overwrite-policy`.

How I tested the change:

I simply ran `cabal new-install` on `cabal-install` (after having installed `cabal-install` with my feature.

## WIP
* [ ] Testing on Windows. Help wanted
* [x] A way to globally configure `--overwrite-policy=prompt` as the default. This was my idea with the patch. I need a bit of help on how to achieve this though as I'm not familiar with the code-base.

### DIFF
--- a/cabal-install/src/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/src/Distribution/Client/CmdInstall.hs
@@ -90,7 +90,7 @@ import Distribution.Client.DistDirLayout
 import Distribution.Client.RebuildMonad
          ( runRebuild )
 import Distribution.Client.InstallSymlink
-         ( symlinkBinary, trySymlink )
+         ( symlinkBinary, trySymlink, promptRun )
 import Distribution.Client.Types.OverwritePolicy
          ( OverwritePolicy (..) )
 import Distribution.Simple.Flag
@@ -140,8 +140,6 @@ import System.Directory
          , removeFile, removeDirectory, copyFile )
 import System.FilePath
          ( (</>), (<.>), takeDirectory, takeBaseName )
-import Distribution.Client.Init.Types (DefaultPrompt(MandatoryPrompt))
-import Distribution.Client.Init.Prompt (promptYesNo)
 
 installCommand :: CommandUI (NixStyleFlags ClientInstallFlags)
 installCommand = CommandUI
@@ -833,11 +831,10 @@ installBuiltExe verbosity overwritePolicy
     overwrite :: IO Bool
     overwrite = remove >> copy
     maybeOverwrite :: IO Bool
-    maybeOverwrite = do
-      a <- promptYesNo
-        "Existing file found while installing executable. Do you want to unlink that file? (y/n)"
-        MandatoryPrompt
-      if a then overwrite else pure a
+    maybeOverwrite
+      = promptRun
+        "Existing file found while installing executable. Do you want to overwrite that file? (y/n)"
+        overwrite
 
 -- | Create 'GhcEnvironmentFileEntry's for packages with exposed libraries.
 entriesForLibraryComponents :: TargetsMap -> [GhcEnvironmentFileEntry]
--- a/cabal-install/src/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/src/Distribution/Client/CmdInstall.hs
@@ -140,6 +140,8 @@ import System.Directory
          , removeFile, removeDirectory, copyFile )
 import System.FilePath
          ( (</>), (<.>), takeDirectory, takeBaseName )
+import Distribution.Client.Init.Types (DefaultPrompt(MandatoryPrompt))
+import Distribution.Client.Init.Prompt (promptYesNo)
 
 installCommand :: CommandUI (NixStyleFlags ClientInstallFlags)
 installCommand = CommandUI
@@ -781,7 +783,7 @@ installUnitExes verbosity overwritePolicy
               <> "Use --overwrite-policy=always to overwrite."
             -- This shouldn't even be possible, but we keep it in case
             -- symlinking/copying logic changes
-            AlwaysOverwrite ->
+            _ ->
               case installMethod of
                 InstallMethodSymlink -> "Symlinking"
                 InstallMethodCopy    ->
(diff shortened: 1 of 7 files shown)


fgaz:
> A way to globally configure --overwrite-policy=prompt as the default. This was my idea with the patch. I need a bit of help on how to achieve this though as I'm not familiar with the code-base.

See #5644. `SavedConfig` represents the `~/.cabal/config` file, but right now `NewInstallFlags` isn't in there

hvr:
@23Skidoo I haven't yet looked at this in detail, but this is  exactly the kind of overwrite policies I had in mind :-)

fredefox:
I've changed the PR to target `master`. I just cherry-picked the one commit I have made towards this.

fgaz:
I added the new-install flags to SavedConfig with #5870 (not working yet due to a bug in how flags are handled: #5942)

There's a conflict now though

gbaz:
I hope this can be picked up and worked on, seems like it is a nice ux improvement that everyone is on board with.

gbaz:
@fgaz can you try to resolve the conflict here? This looks like something we should make happen...

fgaz:
I resolved the conflict, but the copy+prompt case is not handled. I left a TODO, if someone wants to finish this.

fgaz:
Also sorry for force-pushing @fredefox, I remembered too late that this wasn't my branch (github kept the old ref archived though)

Mikolaj:
@fredefox: thank you for your work so far. Would you like to add the missing case or are you busy ATM?

Mikolaj:
Given that @fredefox seems busy, I'm opening this PR to a rescue operation by any noble volunteers. Please feel free to overwrite this branch (or creating a new PR if you insist).

fredefox:
Sorry @Mikolaj. Forgot to reply. My GitHub inbox is super swamped. I don't know if I'll look at this any time soon. Anyone should indeed feel free to take this over.

Mikolaj:
@fredefox: hi! may I help to get this PR merged before it rots? :)

fredefox:
Hi @Mikolaj. Feel free to take over. There is a remaining issue with my work. The prompt is not displayed eagerly enough. I'm not sure why. I suggest testing this, as I did, by running `cabal-install Cabal` twice. Once to install my change, and then once to see the prompt appear. I may return to this, but I make no promises as to when.

Mikolaj:
@fredefox: got it. No pressure. Thanks for sharing the problem --- perhaps somebody takes a look and gets a hunch as to why this happens, before you return.

fredefox:
Hi @Mikolaj. I found the issue and I wrote _a_ solution to it. Emphases on "a". Perhaps someone might have a better proposal than what I did. I figured out that in `promptDefault` the call to `putStr` did not write to stdout before the call to `getLine` triggered. I'm actually surprised by this, because if I compare it to, say, this program:
```
main :: IO ()
main = putStr "Hello" *> getLine *> pure ()
```
The output does indeed appear before blocking for input. Perhaps some strange interaction from the `Interactive` instance for `IO`? I don't really know.

Anyways. In summation, the PR is ready to go as-is as far as I can see.

Mikolaj (approved):
I don't have an explanation, but I'm not surprised at all. I routinely flush whenever I finish adding to the displayed text.

Mikolaj:
@mergifyio rebase

cabal-install/src/Distribution/Client/InstallSymlink.hs
```diff
@@ -195,12 +197,20 @@ symlinkBinary overwritePolicy publicBindir privateBindir publicName privateName
...
+        PromptOverwrite -> maybeOverwrite
   where
     relativeBindir = makeRelative publicBindir privateBindir
     mkLink = createFileLink (relativeBindir </> privateName) (publicBindir   </> publicName)
     rmLink = removeFile (publicBindir </> publicName)
+    overwrite = True <$ (rmLink *> mkLink)
+    maybeOverwrite :: IO Bool
+    maybeOverwrite = do
+      a <- promptYesNo
+        "Existing file found while installing symlink. Do you want to unlink that file? (y/n)"
```
Kleidukos, line 211: From a user perspective the phrasing of the second sentence is a bit obscure. Maybe a better phrasing would be "Do you want to overwrite that file?"

cabal-install/src/Distribution/Client/CmdInstall.hs
```diff
@@ -827,6 +830,14 @@ installBuiltExe verbosity overwritePolicy
...
       else removeFile      destination
     copy = copyFile source destination >> pure True
+    overwrite :: IO Bool
+    overwrite = remove >> copy
+    maybeOverwrite :: IO Bool
+    maybeOverwrite = do
+      a <- promptYesNo
+        "Existing file found while installing executable. Do you want to unlink that file? (y/n)"
+        MandatoryPrompt
+      if a then overwrite else pure a
```
Kleidukos, line 840: You're doing this work twice, here and in `InstallSymlink.hs`. Could this maybe be unified? :)

Kleidukos (changes requested):
Couple of changes, but it looks good overall. :)

cabal-install/src/Distribution/Client/CmdInstall.hs
fredefox: I think it definitely might be able to do some refacturing to increase sharing between `installBuiltExe` and `symlinkBinary`. I'm not sure it's warranted in the case of `maybeOverwrite`. I'm personally not convinced that the code is made more legible by just doing it for `maybeOverwrite`, but I've included it in the latest commit (9f594be790e6cc6fe19485c1d7c655bd09ecf07b) for you to be the judge :)

cabal-install/src/Distribution/Client/InstallSymlink.hs
fredefox: Accepted

Mikolaj:
Is everybody happy with how the feedback was addressed? If so, I'm going to merge (and if nobody answers, I'm merging on Tuesday).

Mikolaj:
@mergifyio rebase

Mikolaj:
CI is OK. Merging. Thank you everybody!

fredefox:
Thank you @Mikolaj!

PR #5673, opened 2018-11-08
Remove unused Semigroup imports.

DavidEichmann:
Due to a bug in ghc, some unused imports do not yield warnings.
This commit will remove such unused imports in preparation for
the ghc bug fix (see https://ghc.haskell.org/trac/ghc/ticket/13064).

---

Testing: Building in CI should be sufficient here.

### DIFF
--- a/Cabal/Distribution/Compat/Prelude.hs
+++ b/Cabal/Distribution/Compat/Prelude.hs
@@ -4,9 +4,11 @@
 {-# LANGUAGE FlexibleContexts #-}
 
 #ifdef MIN_VERSION_base
+#define MINVER_base_411 MIN_VERSION_base(4,11,0)
 #define MINVER_base_48 MIN_VERSION_base(4,8,0)
 #define MINVER_base_47 MIN_VERSION_base(4,7,0)
 #else
+#define MINVER_base_411 (__GLASGOW_HASKELL__ >= 804)
 #define MINVER_base_48 (__GLASGOW_HASKELL__ >= 710)
 #define MINVER_base_47 (__GLASGOW_HASKELL__ >= 708)
 #endif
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Compat/Prelude.hs
```diff
@@ -117,7 +118,12 @@ import Control.DeepSeq               (NFData (..))
 import Data.Data                     (Data)
 import Data.Typeable                 (Typeable)
 import Distribution.Compat.Binary    (Binary (..))
-import Distribution.Compat.Semigroup (Semigroup (..), gmappend, gmempty)
+#if !MINVER_base_411 && MINVER_base_48
```
phadej: don't we need to import `Semigroup` always before base-4.11, let's see what travis says

DavidEichmann: I've made quite a lot of these changes today, so my head is spinning. I think you're right, I'm having a closer look.

DavidEichmann: On closer investigation, `import Distribution.Compat.Semigroup (Semigroup (..))` **is** used because it is rexported as `Semigroup (..)`. Note that `Semigroup (..)` from `Distribution.Compat.Semigroup` contains only `Semigroup((<>))`. The import is still redundant (in newer base version) because the default Prelude is imported (non-qualified) which exports `Semigroup((<>))`.

I'll try and resolve the warning by hiding `Semigroup((<>))`  from prelude.

23Skidoo:
Well, it still verified that lib:Cabal builds, also the Linux 7.10.3 build passed.

23Skidoo:
Merged, thanks!

PR #5675, opened 2018-11-09
Always transliterate coding failures for stdout and stderr

watashi:
We used to only transliterate coding failures on Windows, but this can
fail on other platforms as well, e.g. when `LANG=C` is set on Linux.
We have done this in ghc for all platforms:
https://phabricator.haskell.org/D4642

Before, following commands will fail

```
$ LANG=C cabal install base-compat --verbose=3
Finalized package description:
...
(c) 2014-2018 Jo<stdout>: commitBuffer: invalid argument (invalid character)
```

and now finishes without problem, and print

```
(c) 2014-2018 Jo?o Crist?v?o,
```

instead.

---


watashi:
The only build failure in cause by network issue when downloading HTTP package on travis-ci and not related.

PR #5678, opened 2018-11-09
Suppress warnings in autogenerated signatures.

### LINKED ISSUE #5677: Spurious missing export list warning with Backpack
Reproduction repo: https://github.com/quasicomputational/cabal-backpack-export-list

With cabal-install 2.4.0.0 (using Cabal 2.4.0.1) and GHC 8.6.2, `cabal new-build all` results in this warning:

```
dist-newstyle/build/x86_64-linux/ghc-8.6.2/prog-0/build/autogen/Sig.hsig:1:1: warning: [-Wmissing-export-lists]
    The export item ‘module Sig’ is missing an export list
  |
1 | {-# LANGUAGE NoImplicitPrelude #-}
  | ^
```

The crucial part to triggering it seems to be `prog.cabal`:

```
library
  hs-source-dirs: src/
  exposed-modules:
    Prog
  ghc-options: -Wmissing-export-lists
  build-depends:
    base, sig

executable prog
  main-is: prog.hs
  ghc-options: -Wmissing-export-lists
  build-depends:
    base, impl, prog
```

Something about the library depending on the signature-only package and then the executable instantiating it is critical here; I didn't see any warnings with just the executable.

Autogenerated code really shouldn't be generating warnings under any circumstances. It looks like this is just a matter of engineering, as opposed to something fundamental (cc @ezyang just in case), so I'll have a look and see if I can do it directly.

quasicomputational:
Issue #5677 demonstrates one way for warnings to be produced by dummy
signatures. Autogenerated code really ought to never be generating
warnings, under any sets of flags provided by the user; this patch
uses the big hammer of entirely disabling warnings for the generated
signatures.

Fixes #5677.

---

### DIFF
--- a/Cabal/Distribution/Simple/Build.hs
+++ b/Cabal/Distribution/Simple/Build.hs
@@ -688,6 +688,7 @@ writeAutogenFiles verbosity pkg lbi clbi = do
                       </> ModuleName.toFilePath mod_name <.> "hsig"
             createDirectoryIfMissingVerbose verbosity True (takeDirectory sigPath)
             rewriteFileEx verbosity sigPath $
+                "{-# OPTIONS_GHC -w #-}\n" ++
                 "{-# LANGUAGE NoImplicitPrelude #-}\n" ++
                 "signature " ++ prettyShow mod_name ++ " where"
     _ -> return ()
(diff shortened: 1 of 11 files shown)


quasicomputational:
Should be straightforward enough to merge to 2.4, too, so I'll milestone it.

23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked to `2.4`.

PR #5679, opened 2018-11-11
Remove 'do' in 'instance Text ModuleRenaming'

int-index:
We have this code in `Cabal`, which I don't believe to be syntactically valid Haskell:

```
  do fmap ModuleRenaming parseRns
     <++ parseHidingRenaming
     <++ return DefaultRenaming
```

It works by a fluke. The GHC parser shifts and we parse it as

```
  do (fmap ModuleRenaming parseRns
     <++ parseHidingRenaming
     <++ return DefaultRenaming)
```

rather than

```
  do (fmap ModuleRenaming parseRns)
     (<++ parseHidingRenaming)
     (<++ return DefaultRenaming)
```

However, this behaviour is hard to support. I'm currently working on a set of changes to the parser that result in no shift in this case -- and then get the latter parse, which results in an error about no LHS for `<++`.

Furthermore, even in today's GHC such syntax is not fully supported (happens to work for the `<++` operator above, but not if the operator is `!`).

I'd appreciate if we could merge this to both `master` and the `2.4` branch. (`2.4` because that's what the GHC submodule tracks currently).

----

### DIFF
--- a/Cabal/Distribution/Types/ModuleRenaming.hs
+++ b/Cabal/Distribution/Types/ModuleRenaming.hs
@@ -112,7 +112,7 @@ instance Parsec ModuleRenaming where
 
 
 instance Text ModuleRenaming where
-  parse = do fmap ModuleRenaming parseRns
+  parse = fmap ModuleRenaming parseRns
              <++ parseHidingRenaming
              <++ return DefaultRenaming
     where parseRns = do


23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into `2.4`.

int-index:
Thanks!

PR #5682, opened 2018-11-12
RFC: Put Arbitrary instances for Cabal types in their own package.

quasicomputational:
This is with the intention of the new package,
cabal-quickcheck-instances, being the blessed location for these
orphans, as QuickCheck acquiring a Cabal dependency or vice-versa
would be unsuitable.

This reduces some duplication (some presumably deliberate, and some
apparently accidental) and then some drift between the versions of
these instances.

Due to #1575, some tests have had to move from the Cabal package to
cabal-install.

---


hvr:
So this is strictly a local/internal package *NOT* intended to be released to Hackage, right?

phadej:
There are tests in `Cabal` which depends on this instances. Before we have per-component solving, I see this as blocked.

The tests **should** be inside `Cabal` package. `cabal-install` test-suite is too "far"; and developer experience wise increases feedback loop too much.

:-1:

---

Having a package which shares the sources with `Cabal` (a bit like `aeson` and `attoparsec-iso8601` could work).

23Skidoo:
IMO the existence of `cabal.project` makes the argument for moving all test suites to a single separate package more convincing.

phadej:
FWIW, there is `Version: 3.3 2018-10-24` of SPDX license list. Try out how not nice it is to update now, when tests are mile away from `Cabal`.

quasicomputational:
OK, I have pushed a v2 that shares its source between the new package and Cabal's test suite. (It also has the side benefit on actually working; I thought I'd tested the previous version, but travis disagreed...)

phadej:
I'm trying to resurrect this in https://github.com/haskell/cabal/pull/6557 with a small first step.

phadej:
The goal is achieved. There is `Cabal-QuickCheck` now.

PR #5684, opened 2018-11-13
[WIP] Use the directory package to create new-install symlinks

YellPika:
Running `cabal new-install` on Windows currently gives the error `"Symlinking feature not available on Windows"`. Symlinks have been supported on Windows since Vista (with administrator privileges), and symlinks can currently be created without administrator privileges on Windows 10 when in developer mode. This change lets `cabal new-install` create symlinks on Windows.

I've been using this patch on my own Windows system for ~4 months (ever since haskell/directory#83 was fixed) with no issues.

One caveat is that `directory-1.3.3.0` or newer is required for in order to create symlinks on Windows 10 without administrator privileges.

---


23Skidoo:
Sounds like a good idea.

cabal-install/cabal-install.cabal
```diff
@@ -307,7 +307,7 @@ executable cabal
         containers >= 0.5      && < 0.7,
         cryptohash-sha256 >= 0.11 && < 0.12,
         deepseq    >= 1.3      && < 1.5,
-        directory  >= 1.2.2.0  && < 1.4,
+        directory  >= 1.3.3.0  && < 1.4,
```
hvr: That's a fairly recent `directory` version; can we make this conditional on `if os(windows)`?

YellPika: I don't understand the problem. Are you worried about the stability of recent versions of `directory`?

This change uses the `createFileLink` function, which was added in `directory-1.3.1.0`, so the lower bound will need to be raised anyways. Is `1.3.1.0` an acceptable lower bound? One potential downside is that the behaviour of `new-install` on Windows will differ depending on which version of `directory` was used to compile `cabal-install`.

YellPika:
Travis is reporting that the bootstrap builds are failing. I tried having `bootstrap.sh` download the appropriate `directory` version, but this caused dependency conflicts. Any advice on how to resolve this issue?

YellPika:
Before, any attempt to use `new-install` would report:
```
Symlinking feature not available on Windows
```
Running without developer mode now gives a failure message like the following:
```
C:\Users\Anthony\AppData\Roaming\local\bin\ghcid.exe: createFileLink:CreateSymbolicLink:
permission denied (A required privilege is not held by the client. Creating symbolic links
usually requires administrative rights.)
```

I don't have any machines with older versions of Windows, so unfortunately I can't test there.

Mistuke:
Yeah that error looks fine to me. No objections from my part. I'll leave it to @hvr about the directory bounds

23Skidoo:
An alternative solution/fallback would be to copy and record the checksum of the installed exe somewhere, and after that overwrite only if the checksum of the existing file matches the recorded one (or the user forced an overwrite explicitly).

cabal-install/cabal-install.cabal
hvr: Well, we have all those `*.Compat.*` modules for a reason... ;-)

See 

- https://github.com/haskell/cabal/tree/master/Cabal/Distribution/Compat
- https://github.com/haskell/cabal/tree/master/cabal-install/Distribution/Client/Compat

So I'd strongly suggest to try to follow the prior-art from those compat modules to avoid requiring such a bleeding edge `directory` version

veqryn:
I'm brand new to haskell and trying to setup my dev environment on windows. Immediately ran into the error message that `Symlinking feature not available on Windows`, despite that being false.
Any chance this PR can get traction again?

phadej:
Symlinking on windows is done in https://github.com/haskell/cabal/pull/6519, but it looks it doesn't made it into 3.2. Will be in the next major cabal-install

PR #5692, opened 2018-11-17
Fix erroneous "no exes" warning on lib+exe packages

fgaz:
f6aaeeeae3daca0240a6192024934f4b2143412e acted on units, but an exe+lib
package generates two units (when per-component building is enabled) and
it produced a warning for the lib one.

Now the check is done on all units together.

/cc @hvr


23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked to 2.4.

PR #5695, opened 2018-11-18
Add package revision and cabal file hash to plan.json

arianvp:
This is useful for external tools like plan2nix [1] which
can then create their own build plan based on the cabal
solver output, even in the presence of revised cabal files.

[1] - https://github.com/angerman/nix-tools/tree/master/plan2nix

---

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanOutput.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanOutput.hs
@@ -19,7 +19,7 @@ import           Distribution.Client.ProjectPlanning.Types
 import           Distribution.Client.ProjectBuilding.Types
 import           Distribution.Client.DistDirLayout
 import           Distribution.Client.Types (Repo(..), RemoteRepo(..), PackageLocation(..), confInstId)
-import           Distribution.Client.PackageHash (showHashValue)
+import           Distribution.Client.PackageHash (showHashValue, hashValue)
 
 import qualified Distribution.Client.InstallPlan as InstallPlan
 import qualified Distribution.Client.Utils.Json as J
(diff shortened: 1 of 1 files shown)


arianvp:
By the way, feel free to bikeshed about the field names. I just arbitrarily came up with something that sort of looked appropriate

cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -141,6 +141,12 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
         , "style"      J..= J.String (style2str (elabLocalToProject elab) (elabBuildStyle elab))
         , "pkg-src"    J..= packageLocationToJ (elabPkgSourceLocation elab)
         ] ++
+        [ "pkg-revision" J..=  J.String x
+        | Just x <- [lookup "x-revision" (PD.customFieldsPD (elabPkgDescription elab))]
```
hvr: strong :-1: from me, sorry!

I'm strongly against adding the `x-revision` field to the `plan.json` file to infer the version; it's a completely unreliable field that is only used internally by Hackage (but other package repositories do not -- the x-revision field would be totally bogus there) for bookkeeping purposes and as a failsafe; any tooling which relies on it to be accurate is plain and simple broken!

Also, the revision number is a concept that breaks down completely for e.g. Hackage overlays; the only way to accurately identify a revision is via its SHA256. I plan to add support to cabal for pinning to revisions but I do not intend to support pinning by revision number; only by either SHA256 or by timestamp (the latter addressing scheme will likely be the first to be implemented as it's the one that's close to the `--index-state` design).

arianvp:
Is there a way to get a revision from the hackage API based on the cabal
hash?  Or a planned way to do so?

cabal-install/Distribution/Client/ProjectPlanOutput.hs
arianvp: That indeed sounds like a saner approach to me.   Let me just state that I would very much be interested for the `plan.json` to contain information about what revision is being used.    The approach I implemented might be too fragile, so let's not merge it, but I would very much a less fragile approach to be available.

cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -141,6 +141,12 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
...
+        ] ++
+        [ "pkg-revised-cabal-sha256" J..= J.String hash
```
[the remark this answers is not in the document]
hvr: It's ok if we for now only have the info for overridden .cabals; we might figure out a way to make it available always in a different PR; we definitely have access to the `ByteString` *somewhere*

hvr:
As discussed with @arianvp directly, I'm taking over the PR since he's got little time and I'd like to get this simple but quite useful addition into the upcoming 2.4.1 point rls

hvr:
also cherry-picked to 2.4 via 5c49e01  (+ changelog entry)

PR #5701, opened 2018-11-21
new-clean should not delete dist-newstyle/src

### LINKED ISSUE #5699: new-clean should not delete dist-newstyle/src
I also tried with `-s`

Unfortunately deleting the `src` dir means that source dependencies (i.e. git repos) are also deleted. Which is unfortunate.

It would be highly desirable not to delete these repos. Perhaps if the repos were stored in a global location this would not be a problem?

typedrat:
Fixes #5699.

---


23Skidoo:
Cherry-picked into 2.4.

Issue #5586, opened 2018-09-22
source-repository-package seems to be treated as local


quasicomputational:
This means that `--enable-tests` will build the tests of those dependencies, `--enable-benchmarks` will build the benchmarks, `new-test` will (presumably) run their tests, and so on.

This was quite surprising to me, since my primary use case for `source-repository-package` is to swap in a broken version of a dependency with a fixed pre-release version from the VCS. But doing that means that, if the dependency's test suite is not in great condition (e.g., it doesn't even compile because I'm trying to port things to a new GHC version), I'm suddenly going to be affected by that!

The discussion in #2189 suggests that this is for implementation convenience (which is entirely reasonable), rather than a deliberate decision. Are there use cases where packages being pulled from a VCS should be local for principled reasons? I can't think of any off the top of my head, but there's a whole universe of potential weirdness out there.

Edit: Also, I suspect we could decouple the relevant aspects of 'localness': a `source-repository-package` could be treated as local for the purposes of recompilation checking and not going into the global store, but non-local for the purposes of 'should this package's tests/benchmarks be available for building/running?'.

hvr:
Fwiw, I consider having source-repository-package cached in the global store a killer feature; and I feel like in the majority of use-cases you'd want to err on caching it globally rather than not. That being said, I welcome the ability to opt-out and force it to be treated as local if the user wishes so.

Something I think I briefly discussed with Duncan in the context of the source-repo-package UI was to have a generic uniform "package" syntax where you could explicitly force/control whether a package in the build-plan is demoted to a `local` package or not.

typedrat:
How do packages that are in an index that is referenced in a specific `cabal.project` but is not globally available handled? Because honestly I think internally treating the `source-repository-package` as a source of packages rather than adding a new local package for it makes more sense.

RyanGlScott:
What is the status of this ticket after #5708 was merged into `cabal-install-2.4`? I was originally redirected to this ticket after seeing #5612 (which I was directly affected by in haskell-CI/haskell-ci#186), and after upgrading to `cabal-install-2.4`, I can confirm that I am no longer affected by #5612. Since #5612 was closed as a duplicate of this ticket, does that mean this ticket is also fixed now?

hvr:
@RyanGlScott this ticket is still relevant insofar as the `source-repository-packages` are still treated as if they were local packages whereas it's desirable instead to have the ability to treat them as non-local ones.

harendra-kumar:
[...]

PR #5708, opened 2018-11-23
Fix "source-repository-package breaks 'cabal v2-install'"

### LINKED ISSUE #5643: source-repository-package breaks 'cabal new-install all'
Example minimal project: https://github.com/hasufell/example-cabal-project

```sh
$ cabal new-install all
Wrote tarball sdist to
/home/maerwald/git/example-cabal-project/dist-newstyle/sdist/example-cabal-project-0.1.0.0.tar.gz
cabal: The impossible happened: a local package isn't local
```

And `cabal new-install` on the contrary does **nothing**.

fommil:
Fix #5643 which is a bigger bug than original reported: `v2-install` currently does not work at all if there are git packages. // @hasufell 

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdSdist.hs
+++ b/cabal-install/Distribution/Client/CmdSdist.hs
@@ -29,6 +29,9 @@ import Distribution.Client.DistDirLayout
 import Distribution.Client.ProjectConfig
     ( findProjectRoot, readProjectConfig )
 
+import Distribution.Compat.Semigroup
+    ((<>))
+
 import Distribution.Package
     ( Package(packageId) )
 import Distribution.PackageDescription.Configuration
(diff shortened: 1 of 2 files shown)


fommil:
meh, this doesn't work. For a source dependency on `free`

```
/home/fommil/.../dist-newstyle/sdist/freer-0.2.4.1.tar.gz: openBinaryFile: does not exist (No such file or directory)
```

// @typedrat any ideas?

23Skidoo:
Would be nice if there was a test case.

fommil:
@23Skidoo yeah, I would like that too. Where can I look to add one? I basically just need to run `v2-install` on a package definition that has a git source dependency.

23Skidoo:
I'd look at the existing VCS tests.

fommil:
@23Skidoo could you please point me to them in the repo?

fommil:
I can confirm that this fixes the exception for me, but unfortunately it causes a recompile of my project (e.g. the originally reported repro). I'm not sure if that is a side effect of this fix, or if it is another bug in cabal.

quasicomputational:
A PackageTest should be able to exercise this code. I'd also suggest a test for `new-sdist` with a `source-repository-package`, which was also broken.

Is installing binaries from a `source-repository-package` the right thing to do, considering that regular dependencies off Hackage don't have binaries installed? Will this surprise users? See also #5586.

fommil:
@quasicomputational adding a test seems to be the easy thing, but referencing a git repo that must permanently be available is another thing. Does one exist that I can use?

I think I've convinced myself to prefer a workflow based on `v2-build` rather than `v2-install` so I'm not sure what other people's expectations are. In my usecase I just want access to the binary, I don't care about sdists.

quasicomputational:
I think we'll want to have a local git repository during testing, rather than hitting the network. Creating one as part of the test is probably the simplest route, so shell out to `git init` etc.

To be clear - you want to have a binary from the `source-repository-package` installed (i.e., copied to the `symlink-bindir` directory)? That's what the patch as written will do - i.e., it's treating the package from the git repo as local. I don't object to this exactly, but I just want to be clear on if this is desired behaviour or an implementation detail.

fommil:
I'm travelling and won't be able to work on this PR for a few weeks.

PR #5711, opened 2018-11-23
Improve the handling of .ghc.environment files

### LINKED ISSUE #4542: Add config flag to control generation of .ghc.environment files

### LINKED ISSUE #5587: new-clean/v2-clean does not clean env files
meaning that "build; clean; ghci" breaks:

~~~~
> cabal-2.4.0.0 new-build
Resolving dependencies...
Build profile: -w ghc-8.6.1 -O1
[..]
> cabal-2.4.0.0 new-clean
> ghci
GHCi, version 8.6.1: http://www.haskell.org/ghc/  :? for help
Loaded package environment from $BUILDDIR/.ghc.environment.x86_64-linux-8.6.1
ghc: can't find a package database at $BUILDDIR/dist-newstyle/packagedb/ghc-8.6.1
> echo $?
1
~~~~

Of course plainly deleting env files is not an option, because they may have been created manually, or by other tools.

Affects the latest cabal-2.4.0.0 release with any ghc-8+.

23Skidoo:
* Add a config flag to control generation of `.ghc.environment` files. Fixes #4542.
* Change `new-clean` to also remove `.ghc.environment` files. Fixes #5587.

This is the final blocker before the 2.4.1.0 release.

I haven't updated the docs yet, will do tomorrow.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdClean.hs
+++ b/cabal-install/Distribution/Client/CmdClean.hs
@@ -61,7 +61,8 @@ cleanCommand = CommandUI
             cleanDistDir (\dd flags -> flags { cleanDistDir = dd })
             showOrParseArgs
         , option [] ["project-file"]
-            "Set the name of the cabal.project file to search for in parent directories"
+            ("Set the name of the cabal.project file"
+             ++ " to search for in parent directories")
             cleanProjectFile (\pf flags -> flags {cleanProjectFile = pf})
             (reqArg "FILE" (succeedReadE Flag) flagToList)
         , option ['s'] ["save-config"]
@@ -79,7 +80,8 @@ cleanAction CleanFlags{..} extraArgs _ = do
         mprojectFile = flagToMaybe cleanProjectFile
 
     unless (null extraArgs) $
-        die' verbosity $ "'clean' doesn't take any extra arguments: " ++ unwords extraArgs
+        die' verbosity $ "'clean' doesn't take any extra arguments: "
+                         ++ unwords extraArgs
 
     projectRoot <- either throwIO return =<< findProjectRoot Nothing mprojectFile
 
--- a/cabal-install/Distribution/Client/CmdClean.hs
+++ b/cabal-install/Distribution/Client/CmdClean.hs
@@ -22,39 +22,44 @@ import Distribution.Simple.Utils
 import Distribution.Verbosity
     ( Verbosity, normal )
 
+import Control.Monad
+    ( mapM_ )
 import Control.Exception
     ( throwIO )
 import System.Directory
-    ( removeDirectoryRecursive, doesDirectoryExist )
+    ( removeDirectoryRecursive, removeFile
+    , doesDirectoryExist, getDirectoryContents )
+import System.FilePath
+    ( (</>) )
 
 data CleanFlags = CleanFlags
-    { cleanSaveConfig :: Flag Bool
-    , cleanVerbosity :: Flag Verbosity
-    , cleanDistDir :: Flag FilePath
+    { cleanSaveConfig  :: Flag Bool
+    , cleanVerbosity   :: Flag Verbosity
+    , cleanDistDir     :: Flag FilePath
     , cleanProjectFile :: Flag FilePath
     } deriving (Eq)
 
 defaultCleanFlags :: CleanFlags
 defaultCleanFlags = CleanFlags
-    { cleanSaveConfig = toFlag False
-    , cleanVerbosity = toFlag normal
-    , cleanDistDir = NoFlag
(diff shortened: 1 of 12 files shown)


cabal-install/Distribution/Client/CmdClean.hs
```diff
@@ -22,46 +22,52 @@ import Distribution.Simple.Utils
...
+import System.FilePath
+    ( (</>) )
 
 data CleanFlags = CleanFlags
-    { cleanSaveConfig :: Flag Bool
-    , cleanVerbosity :: Flag Verbosity
-    , cleanDistDir :: Flag FilePath
+    { cleanSaveConfig  :: Flag Bool
+    , cleanVerbosity   :: Flag Verbosity
+    , cleanDistDir     :: Flag FilePath
```
m-renaud: Style curiosity question: why is this formatting so common? It results in so much whitespace noise in diffs because whenever you remove, rename, or add a longer entry you need to re-indent everything else. You also end up with inconsistency when one of the entries is really long and needs to hang on the next line (or results in the other entries being separated by a lot of whitespace).

hvr: It's subjective of course, but having things nicely aligned helps you see the structure rightaway; if it isn't aligned it's more diff friendly but otoh it adds cognitive overhead for the human programmer to see the structure. Also, it's not that bad; you can come up with heuristics for realigning which reduce the diff-noise and deal with overlong entries...

23Skidoo: GitHub diff viewer allows ignoring whitespace noise in diffs (`?w=1` IIRC). I also usually separate formatting commits from meaningful ones.

cabal-install/Distribution/Client/Types.hs
```diff
@@ -591,3 +591,19 @@ instance Monoid AllowNewer where
...
   mappend = (<>)
+
+-- ------------------------------------------------------------
+-- * --write-ghc-environment-file
+-- ------------------------------------------------------------
+
+-- | Whether 'v2-build' should write a .ghc.environment file after
+-- success. Possible values: 'always', 'never', 'ghc8.4.4+' (the
+-- default; GHC 8.4.4 is the earliest version that supports
+-- '-pkg-env -').
```
phadej: `-package-env`?

GHC-8.0.1 seems to support `-package-env`.

Am I missing something?

phadej: FWIW, `-package-env` is also the only option with `env` in it which `ghc --show-options` knows.

23Skidoo: Fixed the comment. It's about `-package-env -` (note the dash), which disables package environments, was added in GHC 8.6, and then backported to 8.4.4. See https://ghc.haskell.org/trac/ghc/ticket/13753 and https://downloads.haskell.org/~ghc/8.4.4/docs/html/users_guide/8.4.4-notes.html#compiler.

phadej: @23Skidoo thanks for clarification!

Issue #5713, opened 2018-11-24
Support leading and trailing commas in `exposed-modules` (and similar)


m-renaud:
## Overview

So, the cabal file format is kinda inconsistent regarding when commas are required and where they're optional, and the places where they are optional don't always allow leading or trailing commas (like `build-depends` supports).

## Example

For example, the following fails to parse, giving an error about the trailing comma after `B`.

```
library
  exposed-modules:
    A,
    B,
  build-depends:
    ...
```

This also fails for `other-modules`, `default-extensions`, and `other-extensions`.

## Solution

The grammar and parser should support leading and trailing commas for all repeated fields which support optional commas.

## (Potentially) Controversial Opinion

The grammar **should require commas for all module fields** (and probably most repeated fields, except for maybe GHC options?). This makes the grammar much more consistent, although would be a backwards incompatible change (but VERY easy to automate updating).

It's really strange that some repeated fields are comma separated and others are whitespace separated, I can't think of any other file format that has this inconsistency.

hvr:
If you `cabal init`, you might notice something:

```
$ cabal init
Please choose version of the Cabal specification to use:
 * 1) 1.10   (legacy)
   2) 2.0    (+ support for Backpack, internal sub-libs, '^>=' operator)
   3) 2.2    (+ support for 'common', 'elif', redundant commas, SPDX)
   4) 2.4    (+ support for '**' globbing)
```

Notice the mention of  "`redundant commas`"? ;-)

m-renaud:
@hvr I should have mentioned in the original post that this is for Cabal spec 2.4, here's the .cabal file:

```
cabal-version:       2.4
name:                myapp
version:             0.1.0.0

library
  exposed-modules:
    A,
  build-depends:       base >=4.12 && <4.13
  default-language:    Haskell2010
```

Which results in:

```
$ cabal v2-build
CabalFileParseError "./myapp.cabal" [PError (Position 7 7) "\nunexpected end of input\nexpecting white space or \"\\\"\"\n\nA,\n"] (Just (mkVersion [2,4])) []
```

Or with the new fancy error messages:

```
Errors encountered when parsing cabal file ./myapp.cabal:

myapp.cabal:7:7: 
unexpected end of input
expecting white space or "\""

    6 |   exposed-modules:
    7 |     A,
    8 |   build-depends:       base >=4.12 && <4.13
```

phadej:
[...]

PR #5714, opened 2018-11-24
Make all or none commas in optional comma lists

phadej:
WIP, comments?

Resolves, https://github.com/haskell/cabal/issues/5713

---


m-renaud:
Thanks for putting this together @phadej , this is definitely a strict improvement over what we have now! So, despite my comments below I'm +1 on this particular improvement :)

That being said, I feel like this optional comma business is a mis-feature of the cabal file syntax. I'm not sure what the original rationale was but it seems to only introduce inconsistencies. The fact that _some_ module lists (`reexported-modules`) have required commas whereas the others have optional commas is just weird.

Also, that the following is valid is also confusing since it would appear that A and B are one item, but really it's a mixing of whitespace and comma separated items:

```
A B, C
```

If we don't go all the way to making commas always required, there should at least be a requirement that it's consistent throughout the block.

In any case, I think we should get some more opinions on what to do going forward. If there's ever a time to make breaking changes it's in 3.0 :)

phadej:
> If we don't go all the way to making commas always required, there should at least be a requirement that it's consistent throughout the block.

That's what this PR does.

m-renaud:
> That's what this PR does.

Ahh, you're right, sorry. I replied before really looking into the code, I just skimmed it and mistakenly misread that this only added support for leading and trailing commas.

PR #5723, opened 2018-11-25
Refactor options fields to use PerCompilerFlavor type

phadej:
E.g. jaeger-flamegraph.cabal failed to roundtrip 'pretty . parse';

Making the list business work didn't feel right,
so I made a bit bigger refactor.

Summary of the patch:

```diff
- options :: [(CompilerFlavor,[String])]
+ options :: PerCompilerFlavor [String]
```

where

```haskell
-- | 'PerCompilerFlavor' carries only info per GHC and GHCJS
-- Cabal parses only @ghc-options@ and @ghcjs-options@, others are omitted.
data PerCompilerFlavor v
```

---


phadej:
Does `appveyor` fails because I changed `GPD` and tests assume `cabal-install` used to build project uses the same format as `Cabal` under test provides (i.e. caches). This is bizarre if it's that way.

phadej:
With this patch `hackage-tests roundtrip` passes. (we don't run it in the CI as it takes 5+ minutes)

PR #5725, opened 2018-11-25
Visibility field

phadej:
This is small step towards adding `visible` field for libraries.

This is **must** for the next release, so keep bikeshedding to the minimum. Concrete change proposals are most appreciated. I hope to merge this soon.

---


m-renaud:
High level comment: having just a boolean for visibility seems pretty limited. Most systems that I've worked with support a variety of visibility attributes ("public", "private", "test-only", "package", "custom", etc.) so it seems like limiting to just a boolean may prohibit some uses of this field in the future. The current change could remain as simple as it is today by changing the field name to `visibility` and supporting two values: `public` and `private` (corresponding to `visible: True` and `visible: False`).

phadej:
I agree that `visibility: public` or `visibility: private` would make sense.

Other visibilities are out-of-scope for 3.0.  Would `test-only` mean that only test-suites in other packages would be able to depend on it. What's the use-case, is it worth complexity? Too much to think about. Not now.

m-renaud:
> Other visibilities are out-of-scope for 3.0.

Perfectly reasonable.

> Would test-only mean that only test-suites in other packages would be able to depend on it

This wasn't the best example tbh, so lets forget about it for now. The "enumerate the list of packages that can see this" is the more interesting use case, which comes up a lot in a large code base. For example, when you only want to allow _some_ teams to depend on your code (because its not general availability yet) its incredibly useful to be able to lock down who can use the code. This would look something like:
```
package-group my-feature-internal-packages:
  my-feature-internal-lib
  my-feature-internal-tests

package-group my-feature-beta-users:
  package-a
  package-b

visibility: 
  my-feature-internal-packages
  my-feature-beta-users
```

phadej:
ping @fgaz @hvr

fgaz:
+1 for visibility: public|private

Actually I already coded this part locally, with the visibility ADT but without the tests. I wish I published it earlier :-/

hvr:
I'm +1 for a non-boolean `visibility: (public | private )` (even if it remained a binary choice)

phadej:
Changed to `visibility: public | private`

I made it so that in main library stanza `visibility` field is not even parsed (and therefore never printed either).

Cabal/Distribution/Types/LibraryVisibility.hs
```diff
@@ -0,0 +1,52 @@
...
+    pretty LibraryVisibilityPublic  = Disp.text "public"
+    pretty LibraryVisibilityPrivate = Disp.text "private"
+
+instance Parsec LibraryVisibility where
+  parsec = do
+    name <- P.munch1 isAlpha
+    case name of
+      "public"  -> return LibraryVisibilityPublic
+      "private" -> return LibraryVisibilityPrivate
+      _         -> fail $ "Unknown visiblity: " ++ name
```
phadej: typo

Cabal/Distribution/Types/Library.hs
```diff
@@ -9,22 +9,24 @@ module Distribution.Types.Library (
...
 import qualified Distribution.Types.BuildInfo.Lens as L
 
 data Library = Library
     { libName           :: Maybe UnqualComponentName
     , exposedModules    :: [ModuleName]
     , reexportedModules :: [ModuleReexport]
-    , signatures        :: [ModuleName]   -- ^ What sigs need implementations?
-    , libExposed        :: Bool           -- ^ Is the lib to be exposed by default?
+    , signatures        :: [ModuleName]       -- ^ What sigs need implementations?
+    , libExposed        :: Bool               -- ^ Is the lib to be exposed by default? (i.e. registered into package db).
```
phadej: (i.e. whether its modules available in GHCi for example)

Cabal/Distribution/Types/Library.hs
```diff
@@ -36,31 +38,41 @@ instance Binary Library
...
+    , libVisibility     = mempty
+    , libBuildInfo      = mempty
+    }
+
+-- | This instance is not good.
+--
+-- We need it for 'PackageDescription.Configuration.addBuildableCondition'.
+-- More correct method would be some kind of "create empty clone".
+--
+-- More concretely, 'addBuildableCondition' will make `libVisibility = False`
```
fgaz: s/False/LibraryVisibilityPrivate/

Cabal/Distribution/Types/Library.hs
```diff
@@ -9,22 +9,24 @@ module Distribution.Types.Library (
...
 
 data Library = Library
     { libName           :: Maybe UnqualComponentName
     , exposedModules    :: [ModuleName]
     , reexportedModules :: [ModuleReexport]
-    , signatures        :: [ModuleName]   -- ^ What sigs need implementations?
-    , libExposed        :: Bool           -- ^ Is the lib to be exposed by default?
+    , signatures        :: [ModuleName]       -- ^ What sigs need implementations?
+    , libExposed        :: Bool               -- ^ Is the lib to be exposed by default? (i.e. whether its modules available in GHCi for example)
+    , libVisibility     :: LibraryVisibility  -- ^ Whether this multilib can be dependent from outside.
```
fgaz: I think the comment should be "[...]can be depended on from outside"

hvr: We could make it even clearer what is meant by "outside" by saying something like "... can be depended on from other packages"

phadej:
I'm merging this as is, as finally CI is green. Will push `[skip ci]` textual fixes later

PR #5729, opened 2018-11-26
Fancy error printing

phadej:
```
Errors encountered when parsing cabal file ./tree-diff.cabal:

tree-diff.cabal:102:7: 
unexpected 'n'
expecting space, "&&", white space, "||", comma or end of input

   98 |       void               >=0.7.2    && <0.8
   99 | 
  100 |       -- space and comments
  101 | 
  102 |       nats               >=1.1.1    && <1.2,
      |       ^

tree-diff.cabal:86:48: version with tags

   85 |     tagged               >=0.8.5    && <0.9,
   86 |     text                 >=1.2.2.2  && <1.3-foo,
      |                                                ^

```

---

### DIFF
--- a/Cabal/Distribution/FieldGrammar/Parsec.hs
+++ b/Cabal/Distribution/FieldGrammar/Parsec.hs
@@ -262,31 +262,33 @@ instance FieldGrammar ParsecFieldGrammar where
 -- Parsec
 -------------------------------------------------------------------------------
 
-runFieldParser' :: Position -> ParsecParser a -> CabalSpecVersion -> FieldLineStream -> ParseResult a
-runFieldParser' (Position row col) p v str = case P.runParser p' [] "<field>" str of
+runFieldParser' :: [Position] -> ParsecParser a -> CabalSpecVersion -> FieldLineStream -> ParseResult a
+runFieldParser' inputPoss p v str = case P.runParser p' [] "<field>" str of
     Right (pok, ws) -> do
-        -- TODO: map pos
-        traverse_ (\(PWarning t pos w) -> parseWarning pos t w) ws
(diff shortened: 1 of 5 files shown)


m-renaud:
Beautiful, I was thinking the same thing myself :) 

Just to make sure, the improvements are:

1. It prints the previous most important line (using more than 2 if its whitespace/comments)
2. Puts a ^ under the place in the output where the parse error occurred.

I noticed that none of the lines _after_ the error are printed, is that intentional? Personally I agree that they probably aren't useful, just wanted to make sure that's intended.

Second (follow-up, may require changes to the parser), could we be even smarter about what the _full_ offending content was? In your second example (assuming I haven't understood what this error was, I haven't personally encountered it myself), the `-foo` is what's actually the issue, so it would we awesome if we could instead output:

```
tree-diff.cabal:86:48: version with tags

   85 |     tagged               >=0.8.5    && <0.9,
   86 |     text                 >=1.2.2.2  && <1.3-foo,
      |                                            ^^^^
```

This would require context-sensitive understanding of what we were parsing and how far we got before something was invalid. Just a thought, no action needed here.

phadej:
> Just to make sure, the improvements are:

exactly those

> I noticed that none of the lines after the error are printed, is that intentional? Personally I agree that they probably aren't useful, just wanted to make sure that's intended.

Correct, it's intentional. The line looked ugly (after caret lines), and didn't seem to provide useful enough context, so two -1/2 made a :-1:.

> Second (follow-up, may require changes to the parser

Yes, it requires *a lot* of changes in the parser as errors and warnings are recorded at source positions, not *source spans*. For some warnings it's a bit easier to add source spans, but `parsec` i.e. parser itself doesn't support them, so ...

m-renaud:
Cool, that's what I expected, no objections from me here!

PR #5732, opened 2018-11-26
bump base bounds

vmchale:
Currently, building `cabal-install` with GHC 7.8.4 will fail. This updates the `.cabal` file to reflect that fact. 

---


23Skidoo:
It's not supposed to fail, but Travis didn't catch it this time (see #5730). i.e. I'd rather fix the failure.

Issue #5726, opened 2018-11-26
RFC: Remove `build-tools`


phadej:
The field is deprecated, it doesn't work.

User can specify
```
build-tools: unzip
```

And that field **does nothing**. I somehow assume that `cabal-install` would check that the tool exists in `$PATH`; but it doesn't even do that.

Therefore, let's remove the field.

---

Note: this would affect only `cabal-version: 3.0` and higher, so nothing existing would break

phadej:
- https://github.com/haskell/cabal/issues/5464 an issue with `build-tools`
- https://github.com/haskell/cabal/issues/4238 confusing issue by me (almost two years old)
- https://github.com/haskell/cabal/issues/5412 (migration to `build-tool-depends`) will be naturally forced as people move to `cabal-version: 3.0` (sooner or later)

phadej:
ping @hvr @Ericson2314

hvr:
I'm sympathetic to making `build-tools` illegal with `cabal-version: 3.0`; at this point I consider `build-tools` merely redundant syntax sugar for `build-tool-depends` anyway.

Ericson2314:
Oh I just saw this. Awesome!!

PR #5733, opened 2018-11-26
Remove fields in cabal spec 3.0

phadej:
- extensions
- (singular) hs-source-dir
- build-tools

Example, with `cabal-version: 2.5`

```
[laptop] tree-diff % cabal new-build --dry
Errors encountered when parsing cabal file ./tree-diff.cabal:

tree-diff.cabal:59:3: error:
The field "extensions" is removed in the Cabal specification version 2.5. Please use 'default-extensions' or 'other-extensions' fields.

   58 | library
   59 |   extensions: CPP
      |   ^

tree-diff.cabal:60:3: error:
The field "extensions" is removed in the Cabal specification version 2.5. Please use 'default-extensions' or 'other-extensions' fields.

   59 |   extensions: CPP
   60 |   extensions: DeriveGeneric
      |   ^
```

Resolves https://github.com/haskell/cabal/issues/5726

---

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -59,6 +59,8 @@ extra-source-files:
   tests/ParserTests/errors/noVersion2.errors
   tests/ParserTests/errors/range-ge-wild.cabal
   tests/ParserTests/errors/range-ge-wild.errors
+  tests/ParserTests/errors/removed-fields.cabal
+  tests/ParserTests/errors/removed-fields.errors
   tests/ParserTests/errors/spdx-1.cabal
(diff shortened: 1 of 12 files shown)


Cabal/tests/ParserTests/regressions/Octree-0.5.format
```diff
@@ -1,3 +1,4 @@
+Octree-0.5.cabal:31:4: The field "extensions" is deprecated in the Cabal specification version 1.12. Please use 'default-extensions' or 'other-extensions' fields.
```
phadej: this is a bug

Issue #5696, opened 2018-11-18
Beginner friendly cabal-install CLI


m-renaud:
# Overview

Provide the most beginner friendly cabal-install CLI possible. This means that the `cabal` commands by default (read: no arguments) should do the thing that beginners expect to get them to do, from zero to a Haskell program they can run.

_Note: This is related, but an expansion in scope from https://github.com/haskell/cabal/issues/5661_

## Example “Hello World”

Here are the commands that a beginner should need to run to get a working “Hello World” Haskell program that will build and run now, and will always build in the future, regardless of global system changes (like ghc version updates):

```
mkdir myfirstproject && cd myfirstproject
cabal init
cabal run
```

That’s it. No interactive prompt asking them a bunch of questions about licences; if they want a library, executable, or library and executable; if they want a separate `src` directory; base language; or anything else.

## Whose developer experience does this improve?
Beginners. Let’s make it as easy as possible for someone new to the language to get started and hacking on a Haskell program. The easier the onboarding, the better the adoption rate. You could argue that you could invoke ghc directly but you quickly need to start using cabal when you want to start using other packages, so let’s not make them learn something new twice.

Casual users. I often find myself with some idea I want to play with and just want the simplest way to get a basic project up and running. The fewer commands and CLI arguments I need to remember the better.

_**”But what about the power users?”**_

I’ll make the same point I made in #5661, if you’re a power user, you know your way around cabal anyways. If you want to create a package interactively you can run `cabal init -i` (or `--interactive`), or maybe just specify the specific arguments you want to `cabal init` yourself. We should make the simplest experience the easiest in order to appeal to the widest audience possible.

This old default behaviour can also be re-enabled by setting `interactive-init: True` (module bikeshed) in `~/.cabal/config` for those that want the old interactive-by-default workflow.

# Changes

Let me first present a summary of the changes, I’ll explain the rational of them below.

- [x] `cabal init` default to not-interactive (`-n`), and `--interactive-init, -i` is added (#5902).
  - Implemented: First question is to initialize a simple project (also available as `--simple`)
  - Provide both `--interactive` and `--non-interactive` flags (#5864)
- [x] `cabal init` defaults to `--exe` _(edit: previously `--libandexe`)_ (#5854)
- [ ] `cabal init` pins the `ghc` version via `with-compiler` in `cabal.project` for executables
- [x] `cabal run` is `cabal v2-run` (aka `cabal new-run`) (#5800)
[...]

PR #5740, opened 2018-11-27
`cabal init` improvements: application-dir and creating library hs file

m-renaud:
# Overview

This adds support for specifying an "application directory" for the `Main.hs` executable entrypoint to be located in (`"app"` is a very common one).

This also generates a `MyLib.hs` file when initializing a new project that is `Library` or `LibraryAndExecutable`. For the latter, the `Main` module also imports this library and uses it.

This can be provided via the `--application-dir` command line flag, and can also be set in the `init` section of `~/.cabal/config` (and will always use that directory unless overridden by the command line flag).

## Config File

```
init
  ...
  -- application-dir:
  -- source-dir:
```

## Examples

**Interactive options:**

```
Application (Main.hs) directory:
 * 1) (none)
   2) src-exe
   3) app
   4) Other (specify)
Your choice? [default: (none)] 2
Library source directory:
 * 1) (none)
   2) src
   3) lib
   4) src-lib
   5) Other (specify)
Your choice? [default: (none)] 4
```

**`--is-library`, no application directory**

```
$ cabal-dev init --is-library -n
Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...
Generating CHANGELOG.md...
Generating MyLib.hs...
Generating inittest.cabal...

$ ls
CHANGELOG.md  inittest.cabal  MyLib.hs  Setup.hs

$ cat inittest.cabal
cabal-version:       >=1.10
-- Initial package description 'inittest.cabal' generated by 'cabal init'.
--   For further documentation, see http://haskell.org/cabal/users-guide/

name:                inittest
version:             0.1.0.0
-- synopsis:
-- description:
-- bug-reports:
-- license:
license-file:        LICENSE
author:              Matt Renaud
maintainer:          matt@m-renaud.com
-- copyright:
-- category:
build-type:          Simple
extra-source-files:  CHANGELOG.md

library
  exposed-modules:     MyLib
  -- other-modules:
  -- other-extensions:
  build-depends:       base >=4.12 && <4.13
  -- hs-source-dirs:

```

**`--is-libandexe`, source-dir, no application dir**

```
$ cabal-dev init --is-libandexe --source-dir=src -n
Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...
Generating CHANGELOG.md...
Generating src/MyLib.hs...
Generating Main.hs...
Generating inittest.cabal...

$ cabal-dev v2-run
Resolving dependencies...
<snip>
[1 of 1] Compiling MyLib            ( ... )
<snip>
[1 of 1] Compiling Main             ( ... )
<snip>
Hello, Haskell!
someFunc
```

**`--application-dir`**

```
$ cabal-dev init --is-libandexe --application-dir=app --source-dir=src -n

Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...
Generating CHANGELOG.md...
Generating src/MyLib.hs...
Generating app/Main.hs...
Generating inittest.cabal...

$ cabal-dev v2-run
Resolving dependencies...
<snip>
[1 of 1] Compiling MyLib            ( ... )
<snip>
[1 of 1] Compiling Main             ( ... )
<snip>
Hello, Haskell!
someFunc
```

## Code Hackiness

Due to the way that other-modules needs to be handled when MyLib.hs and Main.hs are in the same directory complicates the code a little bit. There's also some explicit checks for `exposedModules flags == Just [ModuleName.fromString "MyLib"]`, I could instead add a new flag which is set to `True` when we're creating a `MyLib.hs` file, will probably be a bit cleaner than checking the `exposedModules` list.

## Testing

Manual testing, see examples above.

---


m-renaud:
This is ready for review as well @23Skidoo :)

m-renaud:
Merge conflicts resolved, should be good to go now :)

hvr:
@m-renaud you say

> "app" is a very common one

which is a bit news to me... I myself have never used such a folder name nor do I remember having seen that one used on Hackage.

Personally, I go for `hs-source-dirs` named `src` or `src-...`, such as

```
src/
src-lib/
src-exe/
src-test/
src-bench/
```

m-renaud:
@hvr thanks for the input! I selected `app` because it is the default directory where `Main.hs` is placed by [Summoner](https://github.com/kowainik/summoner), `stack new`, and over half of available Stack templates that produce an executable (12/20 at time of writing). The thing I was going for is consistency.

> I myself have never used such a folder name nor do I remember having seen that one used on Hackage

I'm guessing that's because you don't often use Summoner or Stack to start packages or projects :)

It's likely that this isn't seen on Hackage because:

a) Most Hackage packages are libraries only, not binaries
b) The packages that are binaries were possible/likely created with `cabal init` which did not create `app` directory, so the creators probably didn't change the default (`/src`).

If desired I can also create a global configuration option for this so that users can set it to be what they want once and not have to think about it again (`application-dir: src-exe/` in your case). If we do this I'd like to do that in a follow-up PR though to keep the scope of this change smaller.

gbaz:
fwiw i've seen "src" for the lib and "app" for the executable used a bit at work. it seems to correspond with using "tests" for the test dir instead of e.g. "src-test". But I don't think it matters much any particular way :-)

hvr:
> I'm guessing that's because you don't often use Summoner or Stack to start packages or projects :)

I'll blame Stack then for this weird naming then, which Summoner simply cargo-culted then... ;-)

m-renaud:
I don't think we'll ever be able to please everyone unfortunately :stuck_out_tongue_winking_eye:  For better or worse, `app/` is the most commonly used directory name, and I think at this point consistency-by-default is more important than the specific name. I'd be happy to implement a global config option so that folks that don't like it can pick one they do like.

m-renaud:
A few more thoughts:

Another possible name for this flag would be `executable-dir`. This is a bit more general/common term than "application".

RE: global configuration: it would be cleaner/easier to understand if init had its own section in the global cabal config, then instead of having a top level `interactive-init` and `application-dir` (or `executable-dir`) options we would have:

```
init
  interactive-init: False
  source-dir: src
  executable-dir: app (or src/exe/ or src/bin/ or exe/ or whathaveyou
  cabal-version: 2.4
  ...etc
```

WDYT? I don't want to hold up merging this PR, just want to put the idea out there :)

23Skidoo:
I'm +1 on putting `init` settings into its own section.

m-renaud:
> I'm +1 on putting init settings into its own section.

SG, I'll start putting together a PR for that this week.

Anything else blocking this PR?

m-renaud:
@23Skidoo See #5852, that took less time than expected :stuck_out_tongue_winking_eye:

hvr:
@m-renaud at the very least I'd expect the naming convention to be consistent in itself; and if you want to stick to `app`/`test` instead of `src-*`, then for consistency you should also use `lib` instead of `src` for the library component! And just because somebody else picked an inconsistent naming scheme we don't need to promote it even further by blindly adopting it ;-)

m-renaud:
@hvr another option is to not pick a default at all in this PR, just leave it empty (`(none)`) like is done for `source-dir`.

Also, even with the current implementation this does not "default to `app/`" it only suggests it as one of the options. I'd be perfectly happy to remove that suggestion and instead have `src/` as a suggestion for both library source directory and binary/application directory.

So, to summarize you get the following prompts if you select:

`Executable`:
```
Executable (Main.hs) directory:
1) (none)
2) src
3) app
4) Other (specify)
```

`Library`:
```
Library source directory:
1) (none)
2) src
3) Other (specify)
```

`LibraryAndExecutable`:

Both of the above prompts.

Another alternative is to drop option 3 (app) from the "Executable (Main.hs) directory" prompt all together. The goal of this PR is to allow the libraries and executables to be placed in separate directories if desired. I'd also be open to renaming the flag `--executable-dir` instead of `--application-dir` but that could be confused with "the directory that built executables will be placed in.

WDYT?

gbaz:
We should very much steer people away from having the exe-dir and the src-dir the same. It leads to all sorts of issues and confusions. Herbert's proposal was to have `src-lib` (or just `src`) and `src-exe`. I.e. I don't think that the options for the lib and exe dirs should have overlap.

m-renaud:
> We should very much steer people away from having the exe-dir and the src-dir the same

I completely agree. All that I'm saying is this is the status quo (to use `src/` for both), and the discussion on what to name that directory is holding up this PR (which even permits that distinction). The next cabal-install release isn't for another month or two which is more than enough time to paint the bikeshed and then repaint it again if desired ;)

If desired, I could remove `src/` from the application directory list so its just:

```
Executable (Main.hs) directory:
1) (none)
2) Other (specify)
```

This avoids suggesting that the "library" code and Main.hs live in the same directory. Adding other options can easily be done in a follow-up PR.

gbaz:
Why not add both `app` and `src-exe` to the list? And have all of `lib`, `src` and `src-lib` in the lib-dir list?

m-renaud:
@gbaz 

That SGTM, I'll resolve the merge conflicts, make those changes and update the PR description with updated example output.

m-renaud:
Resolved merge conflicts and updated sets of recommendations in interactive prompt (see Examples section in PR description). Also made this available for configuration in `~/.cabal/config > init > application-dir`.

m-renaud:
Friendly ping :) Any other concerns or can this be merged? @23Skidoo @gbaz

23Skidoo:
Merged, thanks! Sorry for being slow.

m-renaud:
No worries, we've all got stuff going on :)

PR #5753, opened 2018-11-29
Fix `--with-compiler` failing to locate compiler on Windows

hvr:
Due to idiosyncrasies (courtesy of the `SearchPath` Win32 function) in
the Windows implementation of `findExecutable` we need to explicitly try
both, with and without appending the Windows specific executable
extensions (i.e. `.exe`).

For example, before this patch,

     cabal new-build -w ghc-8.6.2

would fail to find the `ghc-8.6.2.exe` even though it's on `%PATH%` and
invoking

     ghc-8.6.2 --version

on the `CMD.exe` shell would have been able to locate the executable.


23Skidoo:
Merged, thanks!

23Skidoo:
Also cherry-picked into 2.4.

PR #5766, opened 2018-12-03
[WIP] Port Travis fixes from the 2.4 branch

### LINKED ISSUE #5763: cabal.project.travis.libonly is missing from master

23Skidoo:
This should fail initially, but eventually will fix #5763.

Please don't merge yet.


23Skidoo:
Fails on 7.6 and 7.8, in the way expected.

23Skidoo:
Still fails (as expected)...

23Skidoo:
Mostly green, except for one unrelated failure, which I have seen before and I think can be fixed by clearing the cache (would be nice to fix the root of the error, though):

```
PackageTests/NewSdist/MultiTarget/all.test.hs                                                                     SKIP (0.06s)
GHCi runtime linker: fatal error: I found a duplicate definition for symbol
   fps_reverse
whilst processing object file
   /home/travis/.cabal/store/ghc-7.6.3/bytestring-0.10.8.2-27b2642182bd2200a97f8837630f23c9058a0a4b6bbad55ba6b46066f77d305d/lib/libHSbytestring-0.10.8.2-27b2642182bd2200a97f8837630f23c9058a0a4b6bbad55ba6b46066f77d305d.a
This could be caused by:
   * Loading two different object files which export the same symbol
   * Specifying the same object file twice on the GHCi command line
   * An incorrect `package.conf' entry, causing some object to be
     loaded twice.
```

Issue #5435, opened 2018-07-15
`cabal new-test` fails if there is no tests


vrom911:
If my package has no test stanza and I run `cabal new-test` I get:

```
cabal: Cannot run tests for the target '' which refers to the package
mypackage-0.0.0 because it does not contain any test suites.
```
with exit code 1.

Though it would be much friendly if this command would succeed with exit code 0 but produce some descriptive message. For example, it would help to make a uniform config for CI (ex: Travis).

My version of `cabal`:
```
$ cabal --version
cabal-install version 2.3.0.0
compiled using version 2.3.0.0 of the Cabal library
```

m-renaud:
Would hiding this behaviour behind a flag be an acceptable solution? I imagine that changing this behaviour for the default could very well break existing usages. I can see situations where both return values could be useful though, so it seems reasonable to support.

Maybe something like `--allow-no-test-suites` or something like that.

vrom911:
@m-renaud, yes, I agree that this is a reasonable solution!

fgaz:
@m-renaud I'd like the opposite: `--disallow-no-test-suites`, since the identity of test results is success

m-renaud:
@fgaz I'm a slight -1 on that name since it's a double negative which is confusing to begin with. It would mean to trigger the new behaviour would need to pass `--disallow-no-test-suites=False` which is just mind bending ;P

Also, for optional flags the usual default value is False, but this one would require a default value of True which is also inconsistent.

vrom911:
I also like this suggestion more as it makes more sense that "no tests" is the positive result of the tests. I'm just happy with any solution at the moment, cause I need to set up a bunch of `if`s for Cabal to overcome this issue on CI.

fgaz:
Fixed by #5768

PR #5768, opened 2018-12-04
Make v2-test succeed when no test suites found

m-renaud:
# Overview 

Change the default behaviour of `v2-test` so that it succeeds when no test suites are found.

This changes the exit code behaviour of the `v2-test` command when the "problem" is that no tests were found. Previously, when no tests were found `cabal v2-test` exit with status `1`, now prints a it friendlier message and exits with status `0`. The old behaviour can be achieved by passing `--test-fail-when-no-test-suites` to `cabal v2-test`.

This addresses #5435.

## Example Usage

**--help output**
```
$ cabal v2-test help
<snip>
    --test-fail-when-no-test-suites                      Exit with failure when no
                                                         test suites are found.
<snip>
```

**Before**
```
$ cabal v2-test
Resolving dependencies...
cabal: Cannot run tests for the target '' which refers to the package
initsample-0.1.0.0 because it does not contain any test suites.

$ echo $?
1
```

**After**
```
$ cabal-dev v2-test
No tests to run for the package initsample-0.1.0.0.

$ echo $?
0
```

```
$ cabal v2-test --test-fail-when-no-test-suites
Resolving dependencies...
cabal: Cannot run tests for the target '' which refers to the package
initsample-0.1.0.0 because it does not contain any test suites.

$ echo $?
1
```

---

/cc @vrom911 @hvr

### DIFF
--- a/cabal-install/Distribution/Client/CmdTest.hs
+++ b/cabal-install/Distribution/Client/CmdTest.hs
@@ -23,14 +23,17 @@ import Distribution.Simple.Setup
          ( HaddockFlags, TestFlags(..), fromFlagOrDefault )
 import Distribution.Simple.Command
          ( CommandUI(..), usageAlternatives )
+import Distribution.Simple.Flag
+         ( Flag(..) )
 import Distribution.Deprecated.Text
          ( display )
 import Distribution.Verbosity
          ( Verbosity, normal )
 import Distribution.Simple.Utils
-         ( wrapText, die' )
+         ( notice, wrapText, die' )
 
 import Control.Monad (when)
+import qualified System.Exit (exitSuccess)
 
 
 testCommand :: CommandUI (ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags, TestFlags)
@@ -102,7 +105,7 @@ testAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags)
 
             -- Interpret the targets on the command line as test targets
             -- (as opposed to say build or haddock targets).
-            targets <- either (reportTargetProblems verbosity) return
+            targets <- either (reportTargetProblems allowNoTestSuites verbosity) return
                      $ resolveTargets
                          selectPackageTargets
                          selectComponentTarget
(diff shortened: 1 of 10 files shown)


fgaz:
I'd prefer the opposite: https://github.com/haskell/cabal/issues/5435#issuecomment-444109546

23Skidoo:
I'm +1 on @fgaz's suggestion.

m-renaud:
@23Skidoo @fgaz see my comment on the other issue, having a double negative flag with a default value of True is quite a bit less clean IMHO.

Unless you're implying you'd be in favour of changing the default behaviour to success when no tests exist. I would actually prefer that but wasn't sure if changing this default behaviour world be breaking.

Let me know what you think.

m-renaud:
@23Skidoo gotcha, just wanted to clarify first :) I wrote the first version with the constant that default behaviour stayed the same, although I suppose this will go out with cabal 3.0 which has a host of other breaking changes.

fgaz:
fwiw old-test succeeded on no tests, so this isn't really breaking

m-renaud:
Ah yes, you're right.

So, we have a couple name options, let's pick one before I change the code then have to change it again ;P

- `--test-disallow-no-test-suites`
- `--test-fail-no-test-suites`
- `--test-fail-when-no-test-suites`

I lean towards the last one.

fgaz:
+1 for `--test-fail-when-no-test-suites`

23Skidoo:
`fail-no` or `fail-when-no` are both good IMO.

m-renaud:
PR updated, CI still running but I've tested it locally and everything seems to work fine :)

cabal-install/Distribution/Client/CmdTest.hs
```diff
@@ -206,9 +210,25 @@ data TargetProblem =
    | TargetProblemIsSubComponent   PackageId ComponentName SubComponentTarget
   deriving (Eq, Show)
 
-reportTargetProblems :: Verbosity -> [TargetProblem] -> IO a
-reportTargetProblems verbosity =
-    die' verbosity . unlines . map renderTargetProblem
+reportTargetProblems :: Flag Bool -> Verbosity -> [TargetProblem] -> IO a
```
23Skidoo: `Verbosity` normally always comes first unless there are special reasons.

23Skidoo:
Merged, thanks!

cabal-install/Distribution/Client/CmdTest.hs
m-renaud: Good point, I'll send a follow-up to clean that up.

Issue #5765, opened 2018-12-03
cabal should pass --host=$HOST to configure script when cross-compiling with build-type:Configure


vmchale:
Otherwise you have to have a `cabal.project` that has

```
package network
    configure-options: --host=arm-linux-gnueabihf
...
```

all over the place. I will hopefully get around to putting together a PR for this.

PR #5771, opened 2018-12-04
pass --host=$HOST to configure when cross-compiling

vmchale:
This fixes https://github.com/haskell/cabal/issues/5765

This has the minor downside of possibly breaking past behavior, since a `configure` script could feasibly error when passed an unknown flag. However, this would only occur when cross-compiling, and any `configure` script that doesn't accept a `--host` flag is unlikely to cross-compile anyways. 

I tested this by cross-compiling `network` (which uses `build-type: Configure`) locally.


23Skidoo:
Merged, thanks! Would be nice if this was documented somewhere in the manual. Maybe add a "Cross-compilation" section?

Issue #5764, opened 2018-12-03
Cabal should pass -x to hsc2hs when cross-compiling


vmchale:
Currently, one has to write

```
package lzma
    hsc2hs-options: -x

package memory
    hsc2hs-options: -x
```

and so on, which can be quite tedious. Ideally, cabal should pass `-x` to `hsc2hs` when it detects we are cross-compiling. 

I'd be willing to implement this feature, but I want to make sure it's done "properly" - i.e. we may wish to make a patch which is general enough to work for other preprocessors which may or may not need special options when cross-compiling.

23Skidoo:
Does

```
package *
    hsc2hs-options: -x
```

work?

PR #5772, opened 2018-12-04
Automatically pass -x to hsc2hs when cross-compiling

vmchale:
This fixes https://github.com/haskell/cabal/issues/5764

`-x` will be passed to `hsc2hs` automatically any time we are cross-compiling. 

I tested this by building the `network` package.


vmchale:
Travis seems to be failing for its own reasons.

23Skidoo:
Merged, thanks!

PR #5790, opened 2018-12-11
Refactor script-extraction code

### LINKED ISSUE #5775: "#! /usr/bin/env cabal" seems broken on windows
The same input file works on linux with cabal-2.4.

I've built cabal-install today from HEAD to make sure this wasn't an solved issue.

Input file
```
#! /usr/bin/env cabal
{- cabal:
build-depends: base ^>= 4.0
-}
{-# LANGUAGE BangPatterns #-}
{-# LANGUAGE ScopedTypeVariables #-}

import Prelude

main :: IO ()
main = return()
```

Console output.
```
$ cabal cabTest.hs
Resolving dependencies...
[1 of 1] Compiling Main             ( Main.hs, C:\ghc\msys64\tmp\cabal-repl.-28296\dist-newstyle\build\x86_64-windows\ghc-8.4.3\fake-package-0\x\script\build\script\script-tmp\Main.o )

Main.hs:7:1: error:
    Could not find module `Prelude'
  |
7 | import Prelude
  | ^^^^^^^^^^^^^^
```

hvr:
The existing implementation suffered from two issues:

 1. Didn't cope well with trailing whitespace (of which #5775 is
    special case)

 2. Silently failing in case of missing (or not properly terminated)
 script blocks

To this end, this commit refactors and improves the parsing logic and
triggers warnings and errors accordingly when appropriate.

This also fixes #5775

---


hvr:
This is all I can manage right now; somebody else will need to take care of writing tests

typedrat:
While we're touching this part of the code, we should probably put in a way to define what spec version to use when parsing.

hvr:
@typedrat I'd do this in a separate PR -- this PR fixes an acute problem which manifests on Windows and causes problems for users. So this here is intend to be cherry-picked to 2.4

typedrat:
In that case, it's mergeable pending tests.

hvr:
@typedrat I've rebased it to master; but I'm also noticing some Git weirdness as it appears that Git (maybe due to some local Git config of mine?) gets confused by files with CRLFs and shows them as "modified"

PR #5812, opened 2018-12-21
Change MungedPackageName to not be opaque

phadej:
See commit messages.

`cabal-install` is not done. Asking for the review now already. I think `cabal-install` part should be quite straight-forward fixing of type errors

PR #5815, opened 2018-12-26
Add LibraryVisibility to InstalledPackageInfo

fgaz:
This pr only modifies the datatype. The field is not filled with the correct visibility info yet. We'll need a ghc-pkg compiled with this patch to actually record this information and properly test the visibility checks.

/cc @phadej @ezyang @23Skidoo

---


phadej:
I guess this needs #5812 and corrections to GHC to properly test, so I'm in favor of merging this eagerly.

If I read properly, the field isn't ever written with this change (as we never yet set it to not *private*), so this is safe to apply.

phadej:
@23Skidoo yes, in similar way we gate on e.g. different formats of `license` field. It's not difficult to do, and I'll watch after so it's done properly.

PR #5825, opened 2019-01-04
Add global configuration option for non-interactive cabal init.

m-renaud:
## Overview

This PR adds support for specifying the `non-interactive` option for `cabal init` in `~/.cabal/config` with the default value being `False`. This does not change any default behaviour.

When the configuration file is generated (when it does not exist), it will contain the following line:

```
-- non-interactive: False
```

## Testing

Manually tested. Added `non-interactive: False` to `~/.cabal/config` and ran `cabal init`:

```
$ cabal init

Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...
Generating CHANGELOG.md...
Generating myapp.cabal...

```

Note that no `.hs` files (other than Setup) are generated because the default is `Library` which currently doesn't generate any Haskell files for (at least until #5740 is merged). Also, I'd like to change this to generate an executable by default (see follow-ups section below).

## Possible follow-ups (RFC)

- Change the flag from a negative (`non-interactive: False`) to a positive (`interactive-init: True`)
- Change `cabal init` to by default generate an `Executable` or `LibAndExe` (the more common case, see the "Default to executable" section of #5696 for rationale).
- Change the default to be non-interactive (`cabal init` requires no args and generates an executable)

---

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -53,9 +53,12 @@ import Distribution.Client.Types
          )
 import Distribution.Client.BuildReports.Types
          ( ReportLevel(..) )
+import qualified Distribution.Client.Init.Types as IT
+         ( InitFlags(..) )
 import Distribution.Client.Setup
          ( GlobalFlags(..), globalCommand, defaultGlobalFlags
          , ConfigExFlags(..), configureExOptions, defaultConfigExFlags
+         , initOptions
          , InstallFlags(..), installOptions, defaultInstallFlags
          , UploadFlags(..), uploadCommand
          , ReportFlags(..), reportCommand
@@ -147,6 +150,7 @@ import GHC.Generics ( Generic )
 
 data SavedConfig = SavedConfig {
     savedGlobalFlags       :: GlobalFlags,
+    savedInitFlags         :: IT.InitFlags,
     savedInstallFlags      :: InstallFlags,
     savedConfigureFlags    :: ConfigFlags,
     savedConfigureExFlags  :: ConfigExFlags,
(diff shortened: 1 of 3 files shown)


cabal-install/Distribution/Client/Config.hs
```diff
@@ -873,6 +914,15 @@ configFieldDescriptions src =
         configAllowNewer (\v flags -> flags { configAllowNewer = v })
        ]
 
+  ++ toSavedConfig liftInitFlag
+       (initOptions ParseArgs)
+       ["quiet", "no-comments", "minimal", "overwrite", "package-dir",
```
m-renaud: I'm wondering if `toSavedConfig` would be better if this was an explicit list of what to include in the global config instead of an exclude list. In this case `non-interactive` is the only thing (at this point) from `InitFlags` that we want to be able to configure in `~/.cabal/config`. Using an exclude list also introduces the possibility of newly added flags sneaking into the global config. Just a thought.

m-renaud:
Not sure if this needs to be in the changelog since it may be renamed before the 3.0 release. I'll update the changelog when we finalize on naming/behaviour/etc.

cabal-install/Distribution/Client/Config.hs
23Skidoo: Makes sense. We could also do something like `IncludeAll | IncludeWhiteList [FlagName]`.

23Skidoo:
Merged, thanks!

23Skidoo:
I'd change the name to something more intuitive like `non-interactive-init` though, because

```
-- non-interactive: False
```

gives me no clue about what it does.

m-renaud:
@23Skidoo 

> I'd change the name to something more intuitive like non-interactive-init though, because

I'd like to refactor/rename it to `--interactive-init` in the `cabal init` command line CLI, and change the default value to `False`. I originally proposed this in the #5696 but there was push back because some folks want the existing "interactive by default" behaviour. With this now available as a global config option it is easy to obtain the old behaviour by setting the following in ~/.cabal/config:

```
interactive-init: True
```

In general I find negative flags confusing, both in CLI usage as well as when reading/modifying the code. Now that this is I'll revisit the issue in the issue I linked above.

23Skidoo:
Sounds good!

cabal-install/Distribution/Client/Config.hs
m-renaud: Added to my TODO list :)

Issue #5705, opened 2018-11-22
Modern project template/scaffolding for `cabal init`


m-renaud:
# Overview

Provide a modern project template/scaffolding when running `cabal init`.

**Target Release:** cabal-install 3.0

A number of features (internal libraries, cabal.project files, etc.) have been added over the past several releases but most of them are hidden in the docs (I only found them because I took some time to read through all the Cabal user docs). We should showcase these features and provide a better scaffolding by default when initializing a package.

There are a few options for when to use this new template:

- For the `LibraryAndExecutable` package type.
- For a new `PackageScaffolding` package type.

# New Template

Assume `packageName: myapp`.

**Directory structure:**
```
./myapp.cabal
./cabal.project
./app/Main.hs
./src/Lib.hs
./test/LibTest.hs
./benchmark/Main.hs
```

**`myapp.cabal`**

```
cabal-version: 2.4
name: myapp
version: 0.1.0.0
synopsis: My application synopsis.
description: My application description.
license: BSD-3-Clause
license-file: LICENSE
build-type: Simple

common common-deps
  default-language: Haskell2010
  build-depends:
    base ^>=4.12,
  ghc-options:
    -Wall
    -Wincomplete-uni-patterns
    -Wincomplete-record-updates
    -Wcompat
    -Widentities
    -Wredundant-constraints
    -fhide-source-paths
    -Wmissing-export-lists
    -Wpartial-fields

library my-internal-lib
  import: common-deps
  hs-source-dirs: src
  exposed-modules:
    Lib
--  build-depends:
--    deps, here,

executable myapp
  import: common-deps
  hs-source-dirs: app
  main-is: Main.hs
  ghc-options:
    -threaded
    -rtsopts
    -with-rtsopts=-N
  build-depends:
    my-internal-lib,

test-suite my-internal-lib-test
  import: common-deps
  type: exitcode-stdio-1.0
  hs-source-dirs: test
  main-is: LibTest.hs
  build-depends:
    my-internal-lib,


benchmark my-internal-lib-benchmark
  import: common-deps
  type: exitcode-stdio-1.0
  hs-source-dirs: benchmark
  main-is: Main.hs
  build-depends:
    my-internal-lib,
```

**`cabal.project`**

```
-- packages: ./*.cabal
-- with-compiler: ghc-8.6.2
```

# Related Issues

#5005 - Cabal `init project` command?
#5696 - Beginner friendly cabal-install CLI

hvr:
[...]

PR #5826, opened 2019-01-04
Default to cabal version 2.4 when running `cabal init --simple`.

m-renaud:
## Overview

`cabal init --simple` provides a way of initializing a cabal package with sensible defaults instead of requiring one to go through the entire interactive prompt. Its goal is to generate a modern directory structure (src/MyLib.hs, app/Main.hs, tests/MyLibTests.hs, etc.) and showcase new cabal features (^>=, recent cabal spec version, common stanzas, etc).

This change defaults to using a more recent cabal spec version instead of the current default of version 1.12.

This partially addresses #5696 and #5705.

## Testing

Manually tested. After running `cabal init --simple` the `.cabal` file contains `cabal-version: 2.4` as the first line of the file.

---


23Skidoo:
Merged, thanks!

Issue #5831, opened 2019-01-09
GHCJS: run benchmark executables directly


harendra-kumar:
When building with GHCJS cabal is able to run the test executables but is not able to run the benchmark executables. I have added GHCJS support to `gauge` and we can now directly run benchmarks using `cabal new-bench`, only if cabal can support running them:

```
Linking /vol/hosts/cueball/workspace/packages/streamly/dist-newstyle/build/x86_64-linux/ghcjs-8.6.0.1/streamly-0.6.0/b/linear-async/build/linear-async/linear-async.jsexe (LinearOps,Main)
Running 1 benchmarks...
cabal: Error: Could not find benchmark program
"/vol/hosts/cueball/workspace/packages/streamly/dist-newstyle/build/x86_64-linux/ghcjs-8.6.0.1/streamly-0.6.0/b/linear-async/build/linear-async/linear-async".
Did you build the package first?
```

I can run manually using `node` though:

```
$ node /vol/hosts/cueball/workspace/packages/streamly/dist-newstyle/build/x86_64-linux/ghcjs-8.6.0.1/streamly-0.6.0/b/linear-async/build/linear-async/linear-async.jsexe/all.js

asyncly/unfoldr                          time                 138.4 μs  
asyncly/unfoldrM                         time                 58.00 ms  
asyncly/fromFoldable                     time                 2.357 ms 
asyncly/fromFoldableM                    time                 43.00 ms  
asyncly/foldMapWith                      time                 37.00 ms  
asyncly/foldMapWithM                     time                 34.00 ms  
...
```

hvr:
@harendra-kumar the problem is that cabal currently only supports invoking executables which the OS can directly execute natively (i.e. where the kernel's executable loader is able to infer which method to invoke to run an executable passed to the `exec(3)` system call). We'd first have to introduce a new abstraction into Cabal which allows to execute artifacts which require some sort of interpreter (be it `node` or `jvm` or something else) sandwiched in.

harendra-kumar:
But it does invoke the tests with `node`, so it seems to be already aware of the `node` interpreter:

```
$ cabal new-test test  --project-file cabal.project.ghcjs
...
Building test suite 'test' for streamly-0.6.0..
Linking /vol/hosts/cueball/workspace/packages/streamly/dist-newstyle/build/x86_64-linux/ghcjs-8.6.0.1/streamly-0.6.0/t/test/build/test/test.jsexe (Main)
Running 1 test suites...
[...]

PR #5832, opened 2019-01-10
Improve GHCJS support

hvr:
This PR contains two commits:

----

### Align `benchmarkExeV10asExe` with `testSuiteExeV10AsExe`'s type-sig 

For some reason, the two operations

    testSuiteExeV10AsExe :: TestSuite -> Executable

    benchmarkExeV10asExe :: Benchmark -> ComponentLocalBuildInfo -> (Executable, ComponentLocalBuildInfo)

were different even though benchmarks and test-suite should be handled
mostly the same.

This divergence was exposed by #5831.

----

### Enable `ghcjs`' `-build-runner` explicitly for exe/test/benches

This way, we automatically gain *some* level of support for
`build-tools-depends` as well as for `v2-install` for applications
which are executed via nodejs

For instance, the following now works:

    $ cabal v2-install -w /opt/ghcjs/8.4/bin/ghcjs --ghcjs alex --symlink-bindir=/tmp/bin
    Resolving dependencies...
    Build profile: -w ghcjs-8.4.0.1 -O1
     - alex-3.2.4 (exe:alex) (requires build)
    Starting     alex-3.2.4 (exe:alex)
    Building     alex-3.2.4 (exe:alex)
    Installing   alex-3.2.4 (exe:alex)
    Completed    alex-3.2.4 (exe:alex)
    Symlinking 'alex'

    $ file - < /tmp/bin/alex
    /dev/stdin: a /usr/bin/nodejs script, UTF-8 Unicode text executable

    $ /tmp/bin/alex  --version
    Alex version 3.2.4, (c) 2003 Chris Dornan and Simon Marlow

This is the best we can without teaching cabal about cross-compilation
and distinguishing local native executables (build by a native-targetting
haskell compiler) for build-tools/custom-setup vs
the main compilation goals for the target platform (which is build with
a haskell cross-compiler).

### DIFF
--- a/Cabal/Distribution/Simple/Build.hs
+++ b/Cabal/Distribution/Simple/Build.hs
@@ -306,15 +306,15 @@ buildComponent verbosity _ _ _ _
 
 buildComponent verbosity numJobs pkg_descr lbi suffixes
                comp@(CBench bm@Benchmark { benchmarkInterface = BenchmarkExeV10 {} })
-               clbi _ = do
-    let (exe, exeClbi) = benchmarkExeV10asExe bm clbi
+               clbi _distPref = do
+    let exe = benchmarkExeV10asExe bm
     preprocessComponent pkg_descr comp lbi clbi False verbosity suffixes
     extras <- preprocessExtras verbosity comp lbi
     setupMessage' verbosity "Building" (packageId pkg_descr)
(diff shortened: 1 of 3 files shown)


hvr:
Btw, if you're on Ubuntu you can use https://launchpad.net/~hvr/+archive/ubuntu/ghcjs to quickly install ghcjs

PR #5845, opened 2019-01-15
Allow list for `repl --build-depends`

typedrat:
`cabal repl --build-dep base, lens` now works as promised.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdRepl.hs
+++ b/cabal-install/Distribution/Client/CmdRepl.hs
@@ -54,7 +54,7 @@ import Distribution.Package
          ( Package(..), packageName, UnitId, installedUnitId )
 import Distribution.PackageDescription.PrettyPrint
 import Distribution.Parsec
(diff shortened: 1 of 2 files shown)


23Skidoo:
Can I do `--build-depends="foo == 0.2, bar == 0.4"`?

typedrat:
It's not done... For one, right now it only works if it is quoted.

typedrat:
As in `cabal new-repl -b "base == 4.12.0.0, filepath == 1.4.2.1"`

23Skidoo:
That's all right, I don't think that other list options work without quotes either.

hvr:
@23Skidoo btw, we should totally do this for `--constraint` as well

phadej:
about quotes: you'll need quotes anyway, as my shell notices `>` and tries to redirect - if not quotes. So I do write `--constraint='lens ^>=4.17'` anyway already.

PR #5848, opened 2019-01-16
Prevent dependency on private library

### LINKED ISSUE #5806: RFC: library visibility enforcement
So, before 3.0 we want a way to enforce library visibility, but to do that we need an extra field in IPI, which would require a GHC release and would restrict multilibs to GHC >=8.8.

Instead, what about:
* enforcing visibility at the Cabal-the-lib level on GHC >=8.8 only
* on GHC <8.8 just enforcing it at the cabal-install level
* hoping that [this doesn't happen](https://xkcd.com/1172/)

Update: we decided not to take the risk. Multilibs will only be supported from GHC 8.8

fgaz:
This will make Cabal error out when depending on a private library.

It does not give detailed info on which sublibrary is the culprit because that would require changing the `dependencySatisfiable` and `finalizePD` return types, and that touches A LOT of the codebase. It can be done later.

/cc @ezyang @phadej

TODO:
* [x] on older ghcs treat all internal libraries as private and all main libraries as public, instead of "everything is private"
  * [x] even on _older_ older ghcs (<8.2)

Closes #5806

### DIFF
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -83,6 +83,7 @@ import Distribution.Types.PkgconfigDependency
 import Distribution.Types.PkgconfigVersionRange
 import Distribution.Types.LocalBuildInfo
 import Distribution.Types.LibraryName
+import Distribution.Types.LibraryVisibility
 import Distribution.Types.ComponentRequestedSpec
 import Distribution.Types.ForeignLib
 import Distribution.Types.ForeignLibType
@@ -913,12 +914,7 @@ dependencySatisfiable
                       packageNameToUnqualComponentName depName)
                  requiredDepsMap)
 
-          || all
-               (\lib ->
-                 (depName, CLibName lib)
-                 `Map.member`
-                 requiredDepsMap)
-               sublibs
+          || all visible sublibs
 
     | isInternalDep
     = if use_external_internal_deps
@@ -949,6 +945,20 @@ dependencySatisfiable
            -- Reinterpret the "package name" as an unqualified component
            -- name
            = LSubLibName $ packageNameToUnqualComponentName depName
+    -- Check whether a libray exists and is visible.
+    -- We don't disambiguate between dependency on non-existent or private
+    -- library yet, so we just return a bool and later report a generic error.
+    visible lib = maybe
+                    False -- Does not even exist (wasn't in the depsMap)
+                    (\ipi -> Installed.libVisibility ipi == LibraryVisibilityPublic
+                          -- If it's a library of the same package then it's
+                          -- always visible.
+                          -- This is only triggered when passing a component
+                          -- of the same package as --dependency, such as in:
+                          -- cabal-testsuite/PackageTests/ConfigureComponent/SubLib/setup-explicit.test.hs
+                          || pkgName (Installed.sourcePackageId ipi) == pn)
+                    maybeIPI
+      where maybeIPI = Map.lookup (depName, CLibName lib) requiredDepsMap
 
 -- | Finalize a generic package description.  The workhorse is
 -- 'finalizePD' but there's a bit of other nattering
@@ -981,7 +991,7 @@ configureFinalizedPackage verbosity cfg enabled
                    pkg_descr0
             of Right r -> return r
                Left missing ->
-                   die' verbosity $ "Encountered missing dependencies:\n"
+                   die' verbosity $ "Encountered missing or private dependencies:\n"
                      ++ (render . nest 4 . sep . punctuate comma
                                 . map (pretty . simplifyDependency)
                                 $ missing)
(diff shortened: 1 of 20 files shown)


fgaz:
Oops, of course on ghc <8.6 all libraries are private :man_facepalming:

fgaz:
Hmm, this probably requires to add `visibility` to `Basic`, so we can turn it into `LibraryVisibilityPublic` if it's a `LMainLibName`

fgaz:
This is now complete and reviewable

(Sorry for the mass ping)

phadej:
@hvr why

> I'm not so sure anymore we should merge this;  

?

fgaz:
@phadej 

Concerns about the narrow ghc compatibility. As it is, this patch only allows multilibs on GHC<8.8 and doesn't provide an override. If it did (for example when using `--exact-config`), then `cabal-install` could make multilibs work even on older GHCs by getting the visibility info from an IPI-external source.

Also this makes cabal-install fail _after_ build plan construction, which is not pleasant, but the only way to solve that is to make the solver sublibrary-aware.

23Skidoo:
@fgaz @hvr Do you still want this in 3.0?

phadej:
This has `priority: high`. 

But in fact, i'm leaning towards disabling parsing of `visibility` field (i.e. essentially removing the multiple public libraries feature from public usage), if it won't be finalized in some way.

This is because I simply don't understand all the implications of how to do this, and want to error on the safe side.

23Skidoo:
I'm fine with postponing it until 3.2.

fgaz:
I think we can, and in fact _should_ merge this, even if we end up disabling `visibility` for 3.0, simply because we can relax the constraint later, but the opposite is difficult. As it is, the rest of multilibs is already there, unconstrained.

Now, should we, in another pr, disable it? well...
* When this pr is merged, the `Cabal` part _will_ be complete
  * The only missing piece I can think of is the error reporting I mentioned in the OP
* `cabal-install` is not multilibs-aware, but IMO this is not so bad. Changing the visibility of a library
    can be treated like changing the visibility of a module for now, like I suggest in [this blog post draft
    (see the warnings)](https://deploy-preview-7--fgaz.netlify.com/posts/2019-05-02-cabal-multiple-libraries/)
* The BC problem cannot be solved at the `Cabal` level, but we can provide an override which can be used by future `cabal-install` releases
  * This can just be `exact-config`
    * Easy to implement at the Cabal level, it's just an extra condition
      [here](https://github.com/haskell/cabal/pull/5848/files#diff-a6de9cdd16372d3e6c4ec16d4d1703a1R901)
    * I think this requires at least a basic no-deps-on-private-libs check in `cabal-install`, since we _always_ pass `exact-config`
  * Or it can be an entirely new flag (`--allow-depending-on-private-libs-yes-im-checking-it-externally`)
    * Easy-ish to add to `Cabal`, there should just be a lot of boilerplate since we have to pass it down to `dependencySatisfiable`
    * Requires no modifications to `cabal-install` 3.0
    * Probably the safest option
  * Or we can add the override in later `Cabal` versions. A minor version, even (if we use a new flag)

TL;DR: +1 for merging, -1 for disabling multilibs, I'm for adding a new flag to allow future `cabal-install`s to override visibility and expand BC

23Skidoo:
OK, I'll try to review this PR this week. Sorry for being slow.

fgaz:
np

---

Added a negative package test

hvr:
> but we can provide an override which can be used by future cabal-install releases

please do! :-)

fgaz:
The test is failing for ghc <=8.0.*

Somehow for those versions sublibraries always appear as public. I'm investigating, but I don't see any clues in the ghc changelog

fgaz:
OK, this should fix the older ghcs. I also added the `--allow-depending-on-private-libs` flag which can be used by `cabal-install` later.

Ping @23Skidoo

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -915,6 +919,7 @@ dependencySatisfiable
                  requiredDepsMap)
 
           || all visible sublibs
+          || allow_private_deps
```
fgaz: Hmm, for efficiency's sake this should probably be the first condition

fgaz: Incidentally that was also a logic error. Fixed.

fgaz:
The travis failures are unrelated to this (I did not even modify ResponseFiles.hs).

@23Skidoo can we merge?

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -703,6 +707,13 @@ configureOptions showOrParseArgs =
          configUseResponseFiles
          (\v flags -> flags { configUseResponseFiles = v })
          (boolOpt' ([], ["disable-response-files"]) ([], []))
+
+      ,option "" ["allow-depending-on-private-libs"]
```
23Skidoo: This should be added to `filterConfigureFlags` as well, though since we don't set it anywhere, it's not a blocker for merging.

23Skidoo:
Merged, thanks!

PR #5853, opened 2019-01-20
Use Haskell2010 and BSD3 for simple project initialization.

m-renaud:
# Overview

cabal init --simple provides a way of initializing a cabal package with sensible defaults instead of requiring one to go through the entire interactive prompt. Its goal is to generate a modern directory structure (src/MyLib.hs, app/Main.hs, tests/MyLibTests.hs, etc.) and showcase new cabal features (^>=, recent cabal spec version, common stanzas, etc).

This change explicitly specifies Haskell2010 and BSD3 for the license.

Testing
Manually tested. After running `cabal init --simple` the `.cabal` file contains `license: BSD-3-Clause` .

---

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -195,9 +195,11 @@ getSimpleProject flags = do
   return $ case maybeToFlag simpleProj of
     Flag True ->
       flags { nonInteractive = Flag True
-            , simpleProject = Flag True
-            , packageType = Flag LibraryAndExecutable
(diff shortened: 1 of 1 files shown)


hvr:
> I'm against any license defaults

I violently agree! the choice of a license ought to be a conscious choice and not imposed as a default by the build tool.

(btw, this also reminds me, we need to extend the licence choices to include the "or-later" variants of the GPL-class licenses)

m-renaud:
Very fair point, I'll leave it empty.

PR #5854, opened 2019-01-20
Default to initializing executable when not specified to `cabal init`

m-renaud:
# Overview

Default to initializing an executable when running `cabal init` non-interactively. This is inline with the other Haskell project initialization tools, as well as tools in other languages (such as Rust's Cargo).

# Rationale

It is more common to want a binary than a library because often one wants to simply play around with an idea then to develop a full fledged library to publish on Hackage. Additionally, the binary case is always more common than the library case because each library is depended upon by _1 or more_ executables.

# Behaviour Change

**Before:**

`cabal init --non-interactive` produces a library.

The `Library` option is choice 1 in the interactive prompt list.

**After:**

`cabal init --non-interactive` produces a binary.

The `Executable` option is choice 1 in the interactive prompt list.

 

---


23Skidoo:
Merged, thanks!

PR #5859, opened 2019-01-24
Add documentation on --config-file and CABAL_CONFIG env variable.

### LINKED ISSUE #5322: document CABAL_CONFIG env variable.
`nix-shell` works by setting a bunch of env variables. It would be nice if among those variables it could set would be one that directed `cabal` to use a config file other than in ~/.cabal/config. That way, we could tweak all of cabal's other settings at once to make it play nicely in the nix environment. This is potentially useful in other settings as well when some other tool wants to put cabal into a specific environment.

m-renaud:
Add information about the `--config-file` command line flag and the `CABAL_CONFIG` environment variable for specifying the cabal config file to section 2.1 ([Configuration](https://www.haskell.org/cabal/users-guide/installing-packages.html)) of the Cabal User Guide.

Fixes #5322.

---

### DIFF
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -15,6 +15,10 @@ explicitly ask ``cabal`` to create it for you using
 
     $ cabal user-config update
 
+You can change the location of the global configuration file by specifying
+either ``--config-file=FILE`` on the command line or by setting the
+``CABAL_CONFIG`` environment variable.
+
 Most of the options in this configuration file are also available as
 command line arguments, and the corresponding documentation can be used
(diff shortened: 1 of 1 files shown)


hvr:
I wonder if it would make sense to create a special section in https://cabal.readthedocs.io/en/latest/cabal-projectindex.html for special CLI flags such as `--config-file` which don't fit into the existing sections.

23Skidoo:
Merged, thanks!

23Skidoo:
@hvr 
> I wonder if it would make sense to create a special section [...]

SGTM

PR #5863, opened 2019-01-25
Instantiate fully instantiated dependencies even in indefinite packages.

### LINKED ISSUE #5634: Template Haskell fails when signatures stack up
Hi,

There is definitely some weird interplay between Template Haskell and backpack that I don't totally understand.  Even when the two seem completely unrelated, I'm getting errors.  Consider the following example:

## Example

Let's assume this is the cabal file:
```
name: th-backpack-failure
version: 1.0
build-type: Simple
cabal-version: >= 2.0

library
    build-depends: base, impl, sig-with-th
    hs-source-dirs: th-backpack-failure
    mixins: sig-with-th requires (Sig as Impl)
    signatures: Unused
    exposed-modules: Go

library sig-with-th
    build-depends: base
    hs-source-dirs: sig-with-th
    signatures: Sig
    exposed-modules: THFuns

library impl
    build-depends: base
    hs-source-dirs: impl
    exposed-modules: Impl
```

The individual files are quite simple.
impl/Impl.hs:
```haskell
module Impl where
```

sig-with-th/Sig.hsig
```haskell
signature Sig where
```

sig-with-th/THFuns.hs
```haskell
module THFuns where
thfun _ = return []
```

th-backpack-failure/Go.hs
```haskell
{-# LANGUAGE TemplateHaskell #-}
module Go where
import THFuns

thfun ''Int
```

th-backpack-failure/Unused.hsig
```haskell
signature Unused where
```

## Tests
Building the above fails with an error message that (I think) indicates that Template Haskell and Backpack don't play nicely together.  But, if you remove the line `signatures: Unused` from the cabal file, it builds just fine.  Alternatively, you can remove the lines `mixins: sig-with-th requires (Sig as Impl)` and `signatures: Sig` and it builds just fine too.

Another work-around is to add a new library stanza to the cabal file, like so:
```
library intermediary
    build-depends: base, sig-with-th, impl
    reexported-modules: THFuns
    mixins: sig-with-th requires (Sig as Impl)
```
Then, change the main library stanza by removing the `mixins` line and depending on `intermediary` instead of `sig-with-th` and `impl`.  This builds just fine too.

## Thoughts
It seems like signatures are being piled together in a way that doesn't support template haskell.  In other words, I really wanted it to be the case that when the main library instantiated `sig-with-th`, that import would be treated as if there were no backpack components at all (after all, all the backpack holes have been filled in).  Then, any Template Haskell imports from that import are totally fine, even if I add a signature to the main library.

The `intermediary` option is a definite work-around, but it would be better to not have to use such tedious boilerplate.

ezyang:

Didn't have time to add tests tonight, they'll come later. Affected users may want to see if this patch solves their problem. Fixes #5634

---

### DIFF
--- a/Cabal/Distribution/Backpack/ReadyComponent.hs
+++ b/Cabal/Distribution/Backpack/ReadyComponent.hs
@@ -45,6 +45,7 @@ import qualified Data.Traversable as T
 import Control.Monad
 import Text.PrettyPrint
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 
 import Distribution.Version
 import Distribution.Pretty
@@ -120,7 +121,11 @@ data IndefiniteComponent
         -- | The dependencies which need to be passed to the compiler
         -- to bring modules into scope.  These are 'OpenUnitId' because
         -- these may refer to partially instantiated libraries.
-        indefc_includes :: [ComponentInclude OpenUnitId ModuleRenaming]
+        indefc_includes :: [ComponentInclude OpenUnitId ModuleRenaming],
+        -- | Fully definite dependencies of this package.  We must
+        -- build this indefinite package after those dependencies,
+        -- in case someone wants to use it for TH.
+        indefc_inst_deps :: [ComponentInclude DefUnitId ModuleRenaming]
     }
 
 -- | Compute the dependencies of a 'ReadyComponent' that should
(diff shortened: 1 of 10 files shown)


ezyang:
Well... the patch does what it says on the tin, but it doesn't actually fix the bug.

ezyang:
Now fixed with tests! The problem was that GHC refuses to improve a fully instantiated indefinite package into the definite one unless it is transitively reachable from the given command line arguments... so just pass in the definite package, not the indefinite one!

ezyang:
cc @dwincort @LightAndLight

23Skidoo:
Merged, thanks!

PR #5864, opened 2019-01-25
Add cabal init --interactive option

m-renaud:
# Overview

Provide both positive (`-i/--interactive`) and negative (`-n/--non-interactive`) flags. This combined with the ability to set the default to either interactive or non-interactive in the ~/.cabal/config file allows you to switch between them. Previously there was only the `--nonInteractive` flag, which if you set your default to be non-interactive could not be used to turn `cabal init` back into interactive mode.

# New Flags

`-i/--interactive` - run the interactive prompt, overriding `interactive: False` in ~/.cabal/config init section.

# Behaviour Change

No change to the default behaviour of `cabal init`.

If both `-i` and `-n` are provided, then the one that occurs last is used (as per the definition of `Flag`)

# `cabal init --help` output

```
Flags for init:
 -h --help                        Show this help text
 -i --interactive                 Enable interactive mode.
 -n --nonInteractive              Disable interactive mode.
```

# ~/.cabal/config

```
init
  -- interactive: True
```

/cc @23Skidoo 

---


23Skidoo:
Merged, thanks!

PR #5869, opened 2019-02-03
bindir-method, copy-bindir, new-install flags in SavedConfig -- 2.4 version

fgaz:
Backport of #5870

TODOs:
* [x] changelog
* [x] docs
* [x] windows testing. does appveyor test new-install?
* [x] is the copy-bindir flag name ok? (it's only for 2.4 anyway)
* [x] Is it ok to rename New*Flags to Client*Flags? *ExFlags (like it's done for configure) is not very descriptive
* [x] bindir-method/install-method

/cc @hvr

### DIFF
--- a/Cabal/doc/nix-local-build.rst
+++ b/Cabal/doc/nix-local-build.rst
@@ -557,6 +557,16 @@ repository, this command will build cabal-install HEAD and symlink the
 
     $ cabal new-install exe:cabal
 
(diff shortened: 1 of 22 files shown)


fgaz:
What about renaming bindir-method to install-method? In the 3.0 version there's no *-bindir, and I think it's better to at least keep the method flag the same

PR #5870, opened 2019-02-03
bindir-method/install-method, installdir, new-install flags in SavedConfig -- 3.0 version

### LINKED ISSUE #5644: Add NewInstallFlags to SavedConfig

### LINKED ISSUE #5837: Task: Implement `v2-install --bindir-method={copy,symlink}`
There's at least two use-cases for having this code-path

1. On Windows, symlink support requires extra configuration; file copying however works without trouble (but we might need to use "ADS" to store some metadata to support store garbage-collection) (see also #5748)
2. When using Docker or other custom workflows, symlinks might easily break; there a way to copy instead of symlink is desirable as well (see e.g. #5628)

Currently, cabal already has support for two locations, 

1. `--bindir=DIR`
2. `--symlink-bindir=DIR`

All that we need is the ability to configure which of those two locations to use; right now `v2-install` is hardwired to symlink and use `--symlink-bindir`

This task is about adding a new flag `--bindir-method=...` to `v2-install` (whose default ought to be configurable via `~/.cabal/config`) which takes takes an enumeration of at least

1. `symlink`, symlink into `--symlink-bindir` location; operation fails if it cannot be symlinked
2. `copy`, copy into `--bindir` location; operation fails if it cannot be copied to target location

Optional (i.e. not critical for 2.4; can be done later): additionally, we might want a 3rd variant (maybe called `auto`) which tries to `symlink`, but falls back to `copy` if symlink fails (relevant for Windows; see also PR #5684 which tried to hardcode this scheme)

Finally,  for the implied defaults (i.e. when there's no explicit default set via `~/.cabal/config`) for `bindir-method:` we should have

1. on non-Windows platforms: `symlink`
2. on Windows: `copy` (or maybe the `auto` method, once it's implemented; however sticking to `copy` as default might result in more deterministic results)

----

Possible bikeshedding alternative:

- `--bindir-style`
- `--bindir-policy` (c.f. `--overwrite-policy`)

### LINKED ISSUE #5843: RFC: post-2.4 redesign of v2 --bindir UI
This is a bit of a follow-up to #5837 (which is intended to provide an incremental pain remedy for 2.4.x) which is expected to result in the following situation for v2-install:

We have two bindir "install" paths settings,

- `--symlink-bindir=DIR` (used when `--bindir-method=symlink`)
- `--bindir=DIR` (used when `--bindir-method=copy`)

This is mostly for historical reasons as cabal has had these two configuration properties for about a decade.

However, it doesn't seem to make much sense in the nix-style paradigm to have two separate folders; instead I suggest to redesign this for cabal 3.0 to

1. deprecate `--symlink-bindir=DIR`
2. Use a single `--bindir=DIR` configuration field
3. Use `--bindir-method={copy,symlink,auto}` to control how executables are "installed" into the folder denoted by `--bindir`

This would IMO be a less confusing UI as you need to be careful which bindir field you set; there's now only one!

----

So e.g. if you're preparing some binary dist for an executable (for which you know it is relocatable as it doesn't have any ties to the nix-style store), you can just say

```bash
cabal install --bindir=./bindist/bin --bindir-method=copy  mypackage
```

and if you instead want to prepare some local `bin`-folder with symlinks to some build tools, you can instead say

```bash
cabal install --bindir=${HOME}/buildtools-bin --bindir-method=symlink alex happy
```

and if you use the `auto` method (i.e. try to use `symlinks` and fallback to `copy` if not possible), now you can say

```bash
cabal install --bindir=${HOME}/buildtools-bin --bindir-method=auto alex happy
```

With the UI as implemented for #5837 for cabal 2.4 however, you'd have to use the following incantations respectively:

```bash
cabal v2-install --bindir=./bindist/bin --bindir-method=copy  mypackage

cabal v2-install --symlink-bindir=${HOME}/buildtools-bin --bindir-method=symlink alex happy

# NB: we need to set both bindir locations!
cabal v2-install --bindir=${HOME}/buildtools-bin --symlink-bindir=${HOME}/buildtools-bin --bindir-method=auto alex happy
```

fgaz:
TODOs:
* [x] changelog
* [x] docs
* [x] a good commit message
* [x] bindir-method vs install-method
* [x] NewFlags vs ClientFlags

Closes #5644
Closes #5837
Closes #5843

/cc @hvr

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -47,7 +47,7 @@ import Distribution.Client.ProjectConfig.Types
          , projectConfigDistDir, projectConfigConfigFile )
 import Distribution.Simple.Program.Db
          ( userSpecifyPaths, userSpecifyArgss, defaultProgramDb
-         , modifyProgramSearchPath )
+         , modifyProgramSearchPath, ProgramDb )
 import Distribution.Simple.Program.Find
          ( ProgramSearchPathEntry(..) )
 import Distribution.Client.Config
@@ -90,7 +90,8 @@ import Distribution.Simple.Command
 import Distribution.Simple.Configure
          ( configCompilerEx )
 import Distribution.Simple.Compiler
-         ( Compiler(..), CompilerId(..), CompilerFlavor(..) )
+         ( Compiler(..), CompilerId(..), CompilerFlavor(..)
+         , PackageDBStack )
 import Distribution.Simple.GHC
          ( ghcPlatformAndVersionString
          , GhcImplInfo(..), getImplInfo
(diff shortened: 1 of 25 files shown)


cabal-install/Distribution/Client/CmdInstall/ClientInstallFlags.hs
```diff
@@ -0,0 +1,106 @@
...
+                  | BindirMethodSymlink
+  deriving (Eq, Show, Generic, Bounded, Enum)
+
+instance Binary BindirMethod
+
+data ClientInstallFlags = ClientInstallFlags
+  { cinstInstallLibs :: Flag Bool
+  , cinstEnvironmentPath :: Flag FilePath
+  , cinstOverwritePolicy :: Flag OverwritePolicy
+  , cinstBindirMethod :: Flag BindirMethod --TODO bindir-method or install-method?
```
hvr: If we're going with `--installdir=...` then this should definitely be `--install-method=` or `--install-policy=` or something else with `--install-*` :-)

hvr:
I briefly tested this one empirically and it looks good so far; I haven't done a proper code-review yet though

Mistuke:
Thanks @fgaz, @hvr , it compiles now. A couple of observations:

1. Doing a `cabal new-install pony` seems to build a binary and then end with

```
Installing executable can-i-have-a-pony in R:\CabalStore\ghc-8.4.3\incoming\new-20796\CabalStore\ghc-8.4.3\pony-1.0-2eda844295defe2b7defaac8f0f9a0f26ee16330\bin
Warning: The directory
R:\CabalStore\ghc-8.4.3\incoming\new-20796\CabalStore\ghc-8.4.3\pony-1.0-2eda844295defe2b7defaac8f0f9a0f26ee16330\bin
is not in the system search path.
cabal.exe: installdir is not defined. Set it in your cabal config file or use
--installdir=<path>
```

Which led me to believe the binary was in `R:\CabalStore\ghc-8.4.3\incoming\new-20796\CabalStore\ghc-8.4.3\pony-1.0-2eda844295defe2b7defaac8f0f9a0f26ee16330\bin`, but it turns out the folder no longer exists after install. Sooo... where is it? :)

2) I then tried using `--installdir` as suggested

```
$ ./cabal.exe new-install --installdir=/r/bin pony
Resolving dependencies...
Up to date
Symlinking 'can-i-have-a-pony'
cabal.exe: Symlinking feature not available on Windows
```
So we'll need to pick a working default, maybe recommend people try a different `--bindir-method=`.

3) I then tried using `--bindir-method=copy` which failed with

```
$ ./cabal.exe new-install --installdir=/r/bin --bindir-method=copy pony
Resolving dependencies...
Up to date
Copying 'can-i-have-a-pony'
R:\CabalStore\ghc-8.4.3\pony-1.0-2eda844295defe2b7defaac8f0f9a0f26ee16330\bin\can-i-have-a-pony: copyFile:atomicCopyFileContents:withReplacementFile:copyFileToHandle:openBinaryFile: does not exist (No such file or directory)
```

Which failed because `can-i-have-a-pony` doesn't have an extension, it should be trying to copy `can-i-have-a-pony.exe`.

fgaz:
@Mistuke could you try 2f17b0ef42b5174db5d07719f27732325856e2d0?

fgaz:
Tested on windows, ready for review. Docs update coming shortly.

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -560,18 +567,27 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags
...
+  let storeDirLayout = cabalStoreDirLayout $ cabalDirLayout baseCtx
+  let mkPkgBinDir :: UnitId -> FilePath
+      mkPkgBinDir = InstallDirs.bindir .
+                    storePackageInstallDirs'
+                      storeDirLayout
+                      (compilerId compiler)
+      mkExeName :: UnqualComponentName -> FilePath
+      mkExeName exe = unUnqualComponentName exe <> ext
+        where
+          ext = case exeExtension platform of "" -> ""
```
23Skidoo: Hmm, can't we use `<.>` from `System.FilePath` here?

23Skidoo:
Rebased against master.

hvr:
@23Skidoo I assume we can consider the 2.4-backport #5869 accepted as well?

23Skidoo:
Yeah I guess, assuming it's the same patches but just backported -- I haven't looked at it.

cabal-install/Distribution/Client/CmdInstall.hs
fgaz: Didn't know about that function, thanks

fgaz:
The error is a timeout. Merging.

PR #5873, opened 2019-02-05
Append newline to the end of sdist help message

watashi:
The help message did not end with newline before and looked weird as it
can concatenate with next shell prompt.

---


23Skidoo:
Maybe this should be done once in the code that prints the command description instead of individually for each command?

23Skidoo:
Otherwise LGTM.

watashi:
> Maybe this should be done once in the code that prints the command description instead of individually for each command?

I tried this and it broke other stuff. `wrapText` will strip leading spaces as well, which is not expected for some `commandDescription` or `commandNotes`, e.g. `sandbox`'s.

We may change `wrapText` to support this, but I think the change is more complex and supposed to be a commit itself. So I will push this as it is now.

PR #5874, opened 2019-02-06
Fix sdist output when sent to stdout

watashi:
This was broken by 7a029caa7026f9b6b585a68062a4175eadb6ca4f

We can not use `ByteString.pack` and then print it as `String`, which
will get encoded again based on the system setting, and very likely,
we get different things unless the encoding is something like ISO-8859-1, e.g.
in UTF-8 environment, the magic of gz "\x1f\x8b" will get encoded again
into "\x1f\xc2\x8b".

Before:

```
$ /cabal sdist --output-dir=- | tar -t -z -f -
tar: This does not look like a tar archive
```

And now:

```
$ cabal sdist --output-dir=- | tar -t -z -f -  | tail -n 1
cabal-install-2.5.0.0/tests/README.md
```

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdSdist.hs
+++ b/cabal-install/Distribution/Client/CmdSdist.hs
@@ -77,7 +77,7 @@ import qualified Data.ByteString.Lazy.Char8 as BSL
 import Data.Either
     ( partitionEithers )
 import Data.List
-    ( find, sortOn, nub, intercalate )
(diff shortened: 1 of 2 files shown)


watashi:
> Any chance a test could be added?

Checked it a bit, it's not easy in current test framework (stdout is handled as `String`).

23Skidoo:
OK, feel free to merge.

PR #5876, opened 2019-02-09
Check for used but undefined flags already in parser

### LINKED ISSUE #4476: "cabal new-build" fails when undefined flags are present with cryptic message
I tried to build using `cabal new-build` and produced the following error:

```
Resolving dependencies...
Map.!: given key is not an element in the map
error, called at libraries/containers/Data/Map/Base.hs:489:16 in
containers-0.5.7.1:Data.Map.Base
```

It turned out that this was because I had

```cabal
if flag(llvm-fast)
    ghc-options: -threaded -O3 -optlo-O3 -fllvm
else
    ghc-options: -threaded
```

in my .cabal file without a corresponding

```cabal
Flag llvm-fast {
  Description: Enable build with llvm backend
  Default: False
}
```

If cabal could throw a nicer error, I'm sure it could save someone else a bit of time.

Thanks!

### LINKED ISSUE #5283: Attempting to de-reference a non-existing flag gives a really bad error.
If you add something like

```
   if flag(that-does-not-exist)
```

to any stanza then you get the rather unhelpful

```
Map.!: given key is not an element in the map
  error, called at libraries/containers/Data/Map/Internal.hs:610:17 in containers-0.5.11.0:Data.Map.Internal
```

with no indicator as to even which cabal file it was in, that this was an issue about a flag at all, or where the error happened. 

I found this while trying to refactor several cabal projects, and it took me the better part of a day to track down the erroneous line of configuration.

phadej:
- Fixes #4476
- Fixes #5283

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -81,6 +81,8 @@ extra-source-files:
   tests/ParserTests/errors/version-sets-3.errors
   tests/ParserTests/errors/version-sets-4.cabal
   tests/ParserTests/errors/version-sets-4.errors
+  tests/ParserTests/errors/undefined-flag.cabal
+  tests/ParserTests/errors/undefined-flag.errors
   tests/ParserTests/ipi/Includes2.cabal
   tests/ParserTests/ipi/Includes2.expr
   tests/ParserTests/ipi/Includes2.format
--- a/Cabal/Distribution/PackageDescription/Parsec.hs
+++ b/Cabal/Distribution/PackageDescription/Parsec.hs
@@ -35,39 +35,41 @@ module Distribution.PackageDescription.Parsec (
 import Distribution.Compat.Prelude
 import Prelude ()
 
-import Control.Monad                                (guard)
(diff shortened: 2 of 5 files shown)


phadej:
Fails as does `master`

```
UNEXPECTED FAIL: PackageTests/BuildTargets/UseLocalPackageForSetup/use-local-package-as-setup-dep.test.hs PackageTests/NewBuild/T4450/cabal.test.hs PackageTests/NewBuild/T4288/cabal.test.hs PackageTests/NewBuild/CustomSetup/RemotePackageWithCustomSetup/build-package-from-repo-with-custom-setup.test.hs PackageTests/NewBuild/CustomSetup/LocalPackageWithCustomSetup/build-local-package-with-custom-setup.test.hs PackageTests/NewBuild/T5164/cabal.test.hs
```

23Skidoo:
Yeah that failure seems to have been caused by @nh2's static exes patch and I haven't gotten around to fixing it yet.

nh2:
Should I have noticed this somehow?

Is it known why that test breaks?

Issue #1780, opened 2014-04-16
Turn on --reorder-goals by default


snoyberg:
I'll admit that I don't understand exactly what --reorder-goals does under the surface. It was basically explained to me as a feature that makes complicated build plans get resolved more quickly, but slows down simpler build plans. Apologies if that's completely incorrect.

In any event, I've recently needed to update the Yesod installation guide to include `--reorder-goals`, as it drastically cuts down on the number of times cabal fails to find a valid build plan. Is there a reason not to have this feature on by default? If it's just to speed up simple build plans, I'd argue that those are probably fast enough already.

thekyriarchy:
is this no longer planned? i was unable to build yi without this option

grayjay:
--reorder-goals still makes each step significantly slower, so I think that the reasons for not making it the default still apply.  The solution that @snoyberg suggested in https://github.com/haskell/cabal/issues/1780#issuecomment-41036239 could work, though.  I created a separate issue for it: #4776

It's also possible that there is a bug causing the solver to not find a solution more quickly without --reorder-goals.  Do you want to open an issue for the yi build?

Mikolaj:
Given that this ticket is linked to from the user guide, let me warn that `--reorder-goals` in some cases slows down dependency resolution from from 15s to 1200s. See #7784.

PR #5886, opened 2019-02-11
Increase `max-backjumps` default from 2000 to 4000

hvr:
The last time the max-backjumps was increased was back in 2014
(see 0229cd5252d44ca65cf33fe350852b46b1a70b20) as suggested
per #1780, and while `--reorder-goals` might often help as well,
it also often comes at a bigger runtime penalty on average than
doubling the max-backjumps.

This should also address issues such as #5882

---


23Skidoo:
Merged, thanks!

PR #5707, opened 2018-11-22
Add 'SimplePackage' package type for `cabal init`

m-renaud:
# Overview

Partially addresses #5696.

Implement's @gbaz's suggestion in https://github.com/haskell/cabal/issues/5696#issuecomment-439737949.

This new `PackageType` addresses the common use case of getting a package set up as quickly as possible without needing to go through the full interactive prompt. We move the `getLibOrExec` prompt to the front of the list, and if `SimpleProject` is selected we bypass the rest of the interactive prompt and use the defaults.

This behaviour can also be invoked via: `cabal init --simple`

# Testing

This change was manually tested, let me know if there's any automated testing for this you'd like me to update.

**New `cabal init` behaviour:**
```
$ cabal-dev init
Should I generate a simple project with sensible defaults?? [default: y] y

Guessing dependencies...

Generating LICENSE...
Warning: unknown license type, you must put a copy in LICENSE yourself.
Generating Setup.hs...
Generating CHANGELOG.md...
Generating Main.hs...
Generating myapp.cabal...

Warning: no synopsis given. You should edit the .cabal file and add one.
You may want to edit the .cabal file and add a Description field.

$ cat myapp.cabal 
cabal-version:       >=1.10
-- Initial package description 'myapp.cabal' generated by 'cabal init'.
-- For further documentation, see http://haskell.org/cabal/users-guide/

name:                myapp
version:             0.1.0.0
-- synopsis:
-- description:
-- bug-reports:
-- license:
license-file:        LICENSE
author:              Matt Renaud
maintainer:          matt@m-renaud.com
-- copyright:
-- category:
build-type:          Simple
extra-source-files:  CHANGELOG.md

library
  -- exposed-modules:
  -- other-modules:
  -- other-extensions:
  build-depends:       base >=4.12 && <4.13
  -- hs-source-dirs:
  default-language:    Haskell2010

executable myapp
  main-is:             Main.hs
  -- other-modules:
  -- other-extensions:
  build-depends:       base >=4.12 && <4.13
  -- hs-source-dirs:
  default-language:    Haskell2010
```

```
$ cabal-dev init --simple

Guessing dependencies...

Generating LICENSE...
Warning: unknown license type, you must put a copy in LICENSE yourself.
Generating Setup.hs...
Generating CHANGELOG.md...
Generating Main.hs...
Generating myapp.cabal...

Warning: no synopsis given. You should edit the .cabal file and add one.
You may want to edit the .cabal file and add a Description field.
```

# Documentation

[...]

PR #5902, opened 2019-02-27
Default to non-interactive init.

m-renaud:
## Overview

By default, running `cabal init` will be performed non-interactively (new) and will default to an executable (existing behaviour). For rationale see #5696, in summary, its much more common to want to create an executable than a library (since each library is by definition depended on by one or more binaries). This also creates a better beginner friendly experience out of the box, allowing them to simply run two commands and have an executable they can hack on right away:

```
cabal init
cabal run
```

## Getting the old behaviour

Whether or not init is interactive is controlled by a flag in `~/.cabal/config`:

```
init
  -- interactive: False
```

By changing it to the value below, all `init` invocations will be interactive (unless `-n` is passed):

```
init
  interactive: True
```

/cc @hvr @23Skidoo @gbaz 

---


byorgey:
LGTM too.

hvr:
I'm not happy about this PR; see my comment at https://github.com/haskell/cabal/issues/5696#issuecomment-468907950

harpocrates:
I'm also against this as the default.

Beginners don't usually start by making packages. Even if they do, the current interactive mode isn't that scary; most questions have defaults so it is possible to just "skip" the ones you don't understand (and still end up with something reasonable).

That said, I also want to appreciate how much effort @m-renaud has put into revamping `init` (even if I don't agree with all those changes).

23Skidoo:
> That said, I also want to appreciate how much effort @m-renaud has put into revamping `init` (even if I don't agree with all those changes).

+1, thanks a lot @m-renaud!

gbaz:
I tend to agree with harpocrates here. The interactive mode holds the hands of beginners through a lot of stuff, and I think it is a good experience. I think that it might be nice to more clearly highlight as _part_ of interactive mode that there is a flag to skip it in the future, to streamline the process for them. But certainly interactive feels more friendly for newcomers, and non-interactive for people that already know what they're doing.

m-renaud:
Thanks everyone for the input and for the recognition, it means a lot :) It an attempt to not scatter the discussion too much I've replied to hvr's comment in https://github.com/haskell/cabal/issues/5696#issuecomment-469451233 to speak to some of the points he raised as well as bringing in some of the other comments left here to have a more complete picture in one spot.

A few short answers (expanded in the other thread):

> @harpocrates : Beginners don't usually start by making packages

I don't think this is true because as soon as you try and do anything non-trivial you need to pull in some packages from hackage since haskell bundles very little in its base library. Also, using the `ghc` CLI directly is significantly less straightforward than running `cabal init && cabal v2-run`, if anything I would rather point folks directly to cabal instead of using ghc directly since you don't need to learn anything new when moving from "super simple file" to "a bit more complex/multi-file/depends on packages". 

> and still end up with something reasonable

If the goal is to end up with something reasonable then not having to do anything and still ending up with something reasonable seems like the base case :)

To speak briefly to a couple of concrete suggestions here: 

> @gbaz : I think that it might be nice to more clearly highlight as part of interactive mode that there is a flag to skip it in the future

I really like this idea and is something to be done regardless of the outcome of this PR discussion, so I'll put together a PR for that when I get a chance.

I mentioned a poll in the other comment, but wanted to ask again here to see what folks thought. LMK and I can put one together, could also post on various discussion threads to see what other people think.

Lastly, I just want to state that I completely respect whatever decision is made regarding this change, I just want to make sure that the various trade-offs have been thoroughly considered and there is diversity of input. I know the release of cabal 3.0 is coming up very quickly so it would be great to have a decision by then, but if that seems unlikely I'd like to at least make sure docs and suggestions from the tool are updated appropriately to reflect the ability to change the default.

byorgey:
I also added a comment over at #5696 .

m-renaud:
Friendly ping :) The PR is marked as blocked on "decision needed", how does this decision get made?

23Skidoo:
I'm a bit on the fence with this one. On the one hand, `-n` is clearly a better default for experienced users, but on the other hand I agree with the comments made in #5696 about the interactive mode being better for beginners.

Nevertheless, given that @byorgey (the original author of `init`) is in favour of making `-n` the default and @m-renaud has volunteered to improve the situation for beginners in other ways (by updating the tutorial and documentation), I decided to merge this.

I agree with @hvr that it probably makes sense to rip out the interactive mode altogether if it's not the default one.

tom-audm:
Count me as another person in favor of leaving it interactive. I've created more than 100 cabal packages (just used `find` to make sure this wasn't hyperbole) and still find the interactivity really helpful. I also find it strange to default to executable. If we want to make a library (which I'm much more likely to do), we have to go online to look up fields I'll need to supply (e.g. `"exposed-modules"`?

I don't think of interactivity as a case of "dumbing down a tool for newcomers." Can't "power users" just find their own ways to create a default that's to their liking (write it by hand/copy a custom template file)? Feels strange to make life difficult for most people for whom "cabal init" means "help me make a cabal package," not "guess what I want so I can type less."

tom-audm:
Also *why* not default to interactive and power users can change one line in their `~/.cabal/config`: `interactive: False`?

Frankly this feels like pulling up the drawbridge and telling users to RTFM.

tom-audm:
Lastly, doesn't defaulting to non-interactive negate https://github.com/haskell/cabal/pull/5707 ?

m-renaud:
Hey @tom-audm! Thanks for the input, a few thoughts:

> and still find the interactivity really helpful

I do as well! But not all the time :)

> I've created 100+ cabal packages ... If we want to make a library (which I'm much more likely to do)

Its true that _you_ are probably more likely to create a library but in general the executable case is by far more common since the total number of libraries would hopefully pale in comparison to the number of executables that use those libraries (see my comment in #5696).

> for most people for whom "cabal init" means "help me make a cabal package," 

I'm not sure how many of the questions are "helpful", out of the 14 questions that get asked, someone who is just learning is only interested in ~1 of them (Library or Executable or Both), the others are just noise (and arguably have bad defaults, such as Cabal spec version 1.10). 

> steepening the required learning curve to simply creating a package
> This will certainly make creating a library more difficult for some users

Previously you would need to use `--is-library` which is a super non-standard flag name, now `cabal init --lib` is all you need to type, and is in line with every other tool from other languages that I'm aware of. I'm not sure how this makes creating a package significantly more difficult, especially with improved documentation.

When starting out, being able to do `cabal init && cabal run` and then start hacking on `Main.hs` is probably the most seamless experience going, I don't see how this could cause users to give up. This would actually make cabal one of the easiest to use tools of any language for a beginner who wanted to start out.

> doesn't defaulting to non-interactive negate #5707 ?

To some degree, yes. I started on that because it was non-controversial and allowed me to get familiar with the code and come out with something useful. It wasn't clear at the time (and still isn't entirely clear) if non-interactive would become the default so I decided to build it out.

Anyways, I'm not the decider of whether or not this goes in or not. It's been sitting in the neutral zone for a while, maybe I'll make a post on Reddit and Haskell Cafe to gather some more data and then the maintainers can go from there.

byorgey:
> Anyways, I'm not the decider of whether or not this goes in or not. It's been sitting in the neutral zone for a while, 

@m-renaud  , @23Skidoo said he decided to merge this.  Though of course it hasn't actually been merged yet.

m-renaud:
Ah, I wasn't sure if that was a decision or not given that it was still open ;P I'm hoping to have some time this week or next to resolve any merge conflicts and start putting together changes to the user guide to go along with this change.

tom-audm:
It seems like we're coming up against a fundamental question we sometimes bump into about Cabal's purpose here -- is it a way to create packages for distribution, or is it more generally a wrapper around all (or nearly all) uses of GHC (i.e. it's more like stack, which more or less requires a project-based workflow)?

The idea that there are many more executables than libraries, for example: this partly rests on the idea that to create a little script for yourself, you need to make a directory and `cabal init` a boilerplate "I don't care what's in it" `.cabal` file, and then `cabal run`, instead of calling `ghci` or `cabal new-repl` with a global package DB. In the GHC-centered workflow, you may not need a `.cabal` package at all until you want to distribute it to others.

If we need to create a `.cabal` file and we don't care what's in it, I have to wonder what the point is of requiring one at all. Don't we run the risk of `.cabal` files just becoming hard drive litter like `.swp` files for vim?

If we do want to prioritize the case of "just give me something so I can start hacking," I'd be much, much happier with a compromise like @gbaz proposed here: https://github.com/haskell/cabal/issues/5696#issuecomment-439737949 -- in other words, the first interactive question allows you to choose "Simple project" and not be asked any more questions.

One other point: if we're considering a "big" user-facing change like changing how `cabal init` works, shouldn't we be checking with the wider user community before causing chaos with a mini FTP-like surprise?

tom-audm:
I also think the question about Cabal's "purpose" illuminates why people in both pro and con groups think of their option as the more beginner-friendly one:

  - For the "cabal is a tool for distribution/package management" people, the most helpful thing for beginners is to guide them through the difficult choices and possible options in a file for distribution (or local package management). In a "distribution" setting, each choice does matter, and the possible options are non-obvious. Interactivity helps, a lot.

  - For the "Cabal is a wrapper for GHC, and `.cabal` files are a key in the ignition of the cabal car" people, the goal is to get users into `cabal run`/`cabal repl` as quickly as possible. Choosing a license before writing the first line of code slows that process down. Then, way down the road if/when they want to distribute the package, the user can read the Cabal manual for each field or delete the `.cabal` file and `cabal init --interactive`

My worry with the second option is we either discourage users from creating packages for distribution by hitting them with more complexity ("do it yourself"), and/or we fill Hackage up with low-quality machine-generated `.cabal` files that beginners don't feel equipped to edit themselves.

phadej:
I’m not sure I said this: make the interactive / non-interactive choice configurable.

There will be unhappy people with either default choice. The ones who don’t like the default, could change it. And if there are a lot angry people after the release, they won’t need to wait for 3.2 when we’ll maybe switch it back

23Skidoo:
It *is* configurable, current discussion is about what should be the default.

tom-audm:
To be clear: we've already merged https://github.com/haskell/cabal/pull/5707, so the question is about whether we should default to:

  1) Interactive prompt, whose first question is:

```
Should I generate a simple project with sensible defaults?? [default: y]
```

And if you choose "y" (or don't make a choice) you don't get asked any more questions.

  2) No interactivity at all - just generate the "sensible defaults" (scare quotes mine)

My strong vote is for option 1.

23Skidoo:
Merged and cherry-picked to `3.0`.

phadej:
My opinion that any kind of migration guides should be as appendices in https://cabal.readthedocs.io/en/latest/ and not spread around in blogs.

phadej:
@m-renaud yes. for example that. producing content is the most important and difficult. It can be organised later. (If someone have good vision for overall structure, i.e. table-of-contents - i don't)

emilypi:
Just throwing my 2 cents into the ring: I really like the default `init` behavior being "hold my hand unless I know what I'm doing". If there is a flag in my cabal config I can flag to turn interactive off by default, that would be preferable. Otherwise, "sensible" means nothing to a new user (and even old users who disagree with the notion of what is sensible), and also forces new users to be aware of `--interactive`, which means they have to grep `--help` just to figure how to get any when building a project; yet another hurdle.

phadej:
As far as I know, there aren't configuration flag to toggle whether `cabal init` is interactive or not. I asked for that feature too.

But now this is merged, so ship has sailed. ping @m-renaud @23Skidoo

m-renaud:
Hey @emilypi @phadej,  I haven't had a chance to update the cabal user guide yet unfortunately, but as is mentioned in the description of this PR you can change the default behaviour via the `init` subsection of `~/.cabal/config`.

Since these are new sections in the configuration as of 3.0 you'll need to run `cabal user-config update` or add the subsection yourself. The update command should generate a new subsection that contains:

```
init
  -- interactive: False
```

If you change this to:
```
init
  interactive: True
```
it will revert to the previous interactive-by-default behaviour.

Issue #5138, opened 2018-02-12
Revert unintended spec change re version number syntax


hvr:
#5092 hghlighted an issue where a change in cabal 1.24 resulted in an unintended change to the cabal specification, specifically relaxing the version parser to unintentionally allow for denormalised version numbers (which are not isomorphic to `NonEmpty Natural` (where strictly speaking, `Natural` has an upper bound)).

Since there's little benefit to allow denormalisation, and it mostly causes subtly problems as e.g. it would violate an implicit assumption code may have head about version numbers round-tripping loss-free, causes trouble to older parsers, complicates the parser as it needs to support and keep track whether it encountered a denormalised version number, this may offer an opportunity for denial of service attacks and/or exploits (I've got more planned in this area, but I don't want to discuss this publicly before we don't have counter measures in place), but most importantly, this wasn't an intentional change to the spec to begin with, we shall consider this spec change to be a bug, and act accordingly.

## Measures that were taken already

 1. The 4 index entries on Hackage were revised (c.f. https://github.com/haskell-infra/hackage-trustees/issues/128); this means that clients which consume the `00-index.tar` index will not be exposed to the original revisions which featured the invalid `.cabal` files.
 2. cabal 2.2's parser was extended to detect denormalisation in order to have `cabal check` complain about invalid `.cabal` files
 3. Hackage-server was retrofitted with an additional validation check which prevents new `.cabal` files with the invalid syntax to enter the index from now on.

## Next steps

Since `cabal > 1.24` currently happily tolerate denormal version numbers, and hackage now actively prevents from more invalid entries entering the package index, the next steps are not "urgent", and can be done whenever convenient:

 - Add the 4 historic original `.cabal` revisions which have become part of the `01-index.tar` index, as well as are included in the original `.tar.gz` to cabal's parser quirk-DB
 - Remove support from the parsec parser for parsing (the invalid) denormal version numbers, turning them into ordinary parser failures (rather than `cabal check` warnings).

PR #5906, opened 2019-03-01
caret operator accepts a set of versions

tseenshe:
I've found the `^>=` operator to be quite verbose to use when I am supporting multiple ghcs (or stackage snapshots). I found this cool post from ages ago https://www.reddit.com/r/haskell/comments/7tykqi/should_stackage_ignore_version_bounds/dtgp0v0 which looked good. e.g.

```
  build-depends: network ^>= { 2.6.3.6, 2.7.0.2, 2.8.0.0, 3.0.1.0 }
```

I don't know how to write a test. Somebody please help?

---

### DIFF
--- a/Cabal/Distribution/Types/VersionRange.hs
+++ b/Cabal/Distribution/Types/VersionRange.hs
@@ -495,12 +495,15 @@ instance Parsec VersionRange where
         verSet :: CabalParsing m => m [Version]
         verSet = do
             _ <- P.char '{'
-            vs <- P.sepBy1 (P.spaces *> verPlain) (P.try (P.spaces *> P.char ','))
             P.spaces
+            vs <- P.sepBy1 (verPlain <* P.spaces) (P.char ',' *> P.spaces)
             _ <- P.char '}'
             pure vs
 
-        -- plain version without tags or wildcards
+        -- a plain version without tags or wildcards
+        -- note: this uses P.integral which allows redundant zeros.
+        --   This is not a problem because 'verPlain' is only used by
+        --   'verSet' which requires cabal > 3
         verPlain :: CabalParsing m => m Version
         verPlain = mkVersion <$> P.sepBy1 P.integral (P.char '.')
 
--- a/Cabal/Distribution/Types/VersionRange.hs
+++ b/Cabal/Distribution/Types/VersionRange.hs
@@ -420,8 +420,19 @@ instance Parsec VersionRange where
 
                 "==" -> do
                     P.spaces
-                    (wild, v) <- verOrWild
-                    pure $ (if wild then withinVersion else thisVersion) v
+                    (do (wild, v) <- verOrWild
+                        pure $ (if wild then withinVersion else thisVersion) v
+                     <|>
+                     (verSet' thisVersion =<< verSet))
+
+                "^>=" -> do
+                    P.spaces
+                    (do (wild, v) <- verOrWild
+                        when wild $ P.unexpected $
+                            "wild-card version after ^>= operator"
+                        majorBoundVersion' v
+                     <|>
+                     (verSet' majorBoundVersion =<< verSet))
 
                 _ -> do
                     P.spaces
@@ -431,7 +442,6 @@ instance Parsec VersionRange where
                     case op of
                         ">="  -> pure $ orLaterVersion v
                         "<"   -> pure $ earlierVersion v
-                        "^>=" -> majorBoundVersion' v
                         "<="  -> pure $ orEarlierVersion v
                         ">"   -> pure $ laterVersion v
                         _ -> fail $ "Unknown version operator " ++ show op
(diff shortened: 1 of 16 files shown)


phadej:
@tseenshe look at leading-comma tests in https://github.com/haskell/cabal/tree/master/Cabal/tests/ParserTests/errors and https://github.com/haskell/cabal/tree/master/Cabal/tests/ParserTests/regressions

(note that only `.cabal` file is written, others are `--accept`ed when run the test-suite).

And after you make a first commit, you'll need to `make gen-extra-source-files` to populate `extra-source-files` field in `Cabal`.

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -469,6 +479,31 @@ instance Parsec VersionRange where
...
+        verSet = do
+            _ <- P.char '{'
+            vs <- P.sepBy1 (P.spaces *> verPlain) (P.try (P.spaces *> P.char ','))
+            P.spaces
+            _ <- P.char '}'
+            pure vs
+
+        -- plain version without tags or wildcards
+        verPlain :: CabalParsing m => m Version
+        verPlain = mkVersion <$> P.sepBy1 P.integral (P.char '.')
```
phadej: Could you add a note that this is used in set notation only (that explains why no tags or wildcards),
and also that as set notation is `>= 3.0`, we can use `P.integral`

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -469,6 +479,31 @@ instance Parsec VersionRange where
...
+
         -- either wildcard or normal version
         verOrWild :: CabalParsing m => m (Bool, Version)
         verOrWild = do
```
phadej: @hvr if you look at this, this doesn't differentiate between `01.02.03` and `1.2.3` given different `cabal-version`.

I'll change that,  hopefully nothing slipped on Hackage

http://hackage.haskell.org/package/Cabal-2.4.1.0/docs/src/Distribution.Types.Version.html#line-99

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -469,6 +479,31 @@ instance Parsec VersionRange where
...
+                , "To use this syntax the package needs to specify at least 'cabal-version: 3.0'."
+                , "Alternatively, if broader compatibility is important then use"
+                , "a series of single version constraints joined with the || operator:"
+                , prettyShow (foldr1 unionVersionRanges (map op vs))
+                ]
+
+        verSet :: CabalParsing m => m [Version]
+        verSet = do
+            _ <- P.char '{'
+            vs <- P.sepBy1 (P.spaces *> verPlain) (P.try (P.spaces *> P.char ','))
```
phadej: this can be rewritten without `try`, if you eat spaces after the parsers here.

```
_ <- P.char '{'
P.spaces
vs <- P.sepBy (verPlain <* spaces) (P.char ',' *> P.spaces)
_ <- P.char '}'
```

Cabal/Distribution/Types/VersionRange.hs
hvr: See also #5138 where we took some measures to prevent denorm versions hitting hackage

Cabal/Distribution/Types/VersionRange.hs
hvr: small typo though, should have read

```
vs <- P.sepBy1 (verPlain <* P.spaces) (P.char ',' *> P.spaces)
```

Cabal/Distribution/Types/VersionRange.hs
hvr: this seems resolved

Cabal/tests/ParserTests/regressions/version-sets.format
```diff
@@ -0,0 +1,18 @@
...
+library
+    default-language: Haskell2010
+    build-depends:
+        network ^>=0,
+        base ==1,
+        base ==1,
+        base ==1 || ==2,
+        base ==1.2,
+        base ==1.2 || ==3.4,
+        ghc ==8.6.3 || ==8.4.4 || ==8.2.2 || ==8.0.2 || ==7.10.3 || ==7.8.4 || ==7.6.3 || ==7.4.2,
```
phadej: ooh, hmm...

@hvr should we introduce a `VersionRange` constructor for set notation? I'd say yes. And can do it after this is merged.

phadej: We'd need `NonEmpty` to do it, so it's "a bit" tricky

hvr: The main benefit would be for syntactical round tripping, wouldn't it? Don't we already have other cases such as `-none` where we have a syntactically lossy encoding?

I wonder if we should instead go for having two different datatypes, one for the concrete parsetree, and one for the desugared abstract syntax tree. The current `VersionRange` type is a bit inconsistent in this regard IMO (i.e. it has support for representing parens which are imo redundant for an expression tree, but otoh it lacks the ability to represent `-none` directly).

In any case, I think we should discuss this in a new separate "design ticket" and we also still have the issue of having to handle pkg-config version-ranges which might be related.

Issue #5819, opened 2018-12-30
Release for GHC 8.8


bgamari:
While GHC is quite behind in the alpha cycle this time around due to shuffling around of infrastructure, I would still like to make sure we have plenty of time to stabilize submodules before the final release in February. Our original plan stated that core library version numbers would be set by mid-December and core library releases for 8.8 would be ready by late January. How are we looking compared to these benchmarks?

hvr:
@bgamari btw, I was in the midst of updating lib:Cabal in  GHC HEAD to a more recent snapshot and got side-tracked cause there are some API adaptation necessary and those exposed some bugs that needed to be fixed before the update can proceed... and this prompted #5812 ... and.... long story short... we need to resolve this before it even makes any sense to bump the version (which we should be able to do soon afterwards)

bgamari:
Any updates on this?

hvr:
@bgamari #5812 just got merged, and I'm right now trying to fwd port https://gitlab.haskell.org/hvr/ghc/commit/d6428a5e3533e0c94a68f854f8b543f6a52405b1 to it

bgamari:
How is this looking?

bgamari:
How close are we to being able to pin down a final submodule commit?

23Skidoo:
@fgaz @hvr What's the state of the public sublibraries feature? Do we need anything else on the `lib:Cabal` side?

fgaz:
I need to write a fallback for ghc<8.8 in #5848, then multilibs will be complete

hvr:
@23Skidoo if we're talking about an actual hard freeze of 3.0, I think there's a few loose ends (one of which might be the 3.0 variant of #5837 which might require changes to lib:Cabal, @fgaz might know this)  which should ideally be resolvable within 1-2 weeks; weekend's coming up, so I'll be able to take a closer look a the 3.0 situation and give a more concrete answer

fgaz:
#5837 requires no changes to lib:Cabal

fgaz:
There's also this parsing bug: #5846

But It's on new functionality and there are no regressions, so it it's not super important

bgamari:
How are things going here?

fgaz:
Public sublibraries: I'm awaiting review on #5848

bgamari:
[...]

PR #5915, opened 2019-03-03
Bump Cabal versions from 2.5 -> 3.0

hvr:
NOTE: This will most certainly fail at CI because of the `cabal-testsuite.cabal` requiring a bleeding edge lib:Cabal for its Setup.hs -- I can't find the ticket, but there's a a plan to decouple `cabal-testsuite` from requiring a bleeding edge lib:Cabal (UPDATE: see #5916)

Another (less ideal) workaround would be to use a `cabal` 3.0.* executable for building `cabal-testsuite`

Addressing #5819


phadej:
`cabal-testsuite` indeed failed on appveyor

bgamari:
On what sort of timeframe can we expect this to be merged? Perhaps it would make sense for GHC to just start tracking this branch?

hvr:
@bgamari we could push the commit from this PR together with a instant-revert-followup-commit to `master`, and then you could refer to that; using this branch would be dangerous as there's no guarantee the commit  971f083 will remain reachable and then you'd end up with a broken gitlink in ghc.git

bgamari:
@hvr, we could do that but we'll need to sort out [#16378](https://ghc.haskell.org/trac/ghc/ticket/16378) before we can do so.

hvr:
@bgamari can that very backpack failure be reproduced with GHC 8.6? or is it really a function of the GHC version and specific to GHC 8.8?

23Skidoo:
This is green, time to merge?

PR #5922, opened 2019-03-06
Fix pull URLs [ci skip]

snoyberg:
Opening as a PR instead of pushing as I'm not sure which branch to best push to.

I'm also now thoroughly annoyed with Github for having a URL scheme with `/issues/1234` (plural) versus `/pull/5678` (singular), not sure how I never noticed that before.

---


phadej:
`master` will do. We don't have release branch for `3.0` now, and a new release for 2.4 is highly unlikely at this point. Thanks!

PR #5929, opened 2019-03-09
Add bkpcabal01 regression test-case

hvr:
This exemplifies the bpkcabal01 regression (introduced via #5863); @ezyang is aware of this

See e.g. failure at

- https://travis-ci.org/haskell/cabal/jobs/504058651#L4431
- https://travis-ci.org/haskell/cabal/jobs/504058649#L3944

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanning.hs
@@ -2229,6 +2229,41 @@ instantiateInstallPlan plan =
 
     indefiniteComponent :: UnitId -> ComponentId -> InstM ElaboratedPlanPackage
     indefiniteComponent _uid cid
+      -- Only need Configured; this phase happens before improvement, so
+      -- there shouldn't be any Installed packages here.
+      | Just (InstallPlan.Configured epkg) <- Map.lookup cid cmap
+      , ElabComponent elab_comp <- elabPkgOrComp epkg
+      = do -- We need to do a little more processing of the includes: some
+           -- of them are fully definite even without substitution.  We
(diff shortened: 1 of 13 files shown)


23Skidoo:
> bkpcabal01

This implies the existence of `bkpcabal02`, `03`, and so on. Why not pull in them all?

hvr:
> This implies the existence of bkpcabal02, 03, and so on.

Not necessarily... it was primarily to make a connection to the same-name testcase in GHC's testsuite :-)

> Why not pull in them all?

We should do it; I just haven't reviewed them yet to see if they're actually distinct from backpack tests we already have in cabal's testsuite....

ezyang:
Many of the GHC tests are intended to exercise specific things that GHC is more likely to break, like recompilation avoidance. It would be nice if we could easily run tests in both places, but this has to be balanced with copy pasting.

ezyang:
OK, I have a pretty good guess for what the problem is: ISTR in #5863 I only updated Cabal, but I feel like maybe I also have to update the logic in cabal-install too. Looking.

ezyang:
Fixing this is a little tricky. Will look at it more tomorrow.

ezyang:
Fix pushed. It passes my local CI, let's see what Travis says.

hvr:
I really wish we had a GHC 8.2.3 point-release to fix stuff like the "unknown package: hole" regressions, as I'm stuck on GHC 8.2.* for the foreseeable future for many of the projects I work on... and so far my option is to use an unofficially patched GHC 8.2.3 :-/

ezyang:
Thank you! Looks like this is good to go.

PR #5931, opened 2019-03-10
Fix some `v2-install` options

### LINKED ISSUE #5788: cabal new-install doesn't support --ghc-options
```
$ cabal new-install --ghc-options='-fllvm -optlo-O2 -j +RTS -A64m -n2m -RTS' all
cabal: unrecognized 'new-install' option `--ghc-options=-fllvm -optlo-O2 -j
```

harpocrates:
Fixes #5788 by

  - adding `--ghc-options` (actually the full suite of configure-related options
    already available to other `v2-*` commands, but `--ghc-options` is the one
    most folks care about)
  - prefixing haddock related options with `haddock-`

The fact that `Distribution.Client.Setup` and `Distribution.Simple.Setup` both
export a bunch of functions that are identically named _and_ have similar
functionality is probably going to continue to cause bugs of this sort...

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -28,7 +28,7 @@ import Distribution.Client.CmdSdist
 import Distribution.Client.Setup
          ( GlobalFlags(..), ConfigFlags(..), ConfigExFlags, InstallFlags(..)
          , configureExOptions, haddockOptions, installOptions, testOptions
-         , liftOptions )
+         , configureOptions, liftOptions )
 import Distribution.Solver.Types.ConstraintSource
          ( ConstraintSource(..) )
 import Distribution.Client.Types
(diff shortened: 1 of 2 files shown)


23Skidoo:
Merged, thanks!

hasufell:
Will there be a point release for this?

23Skidoo:
Well, 3.0 should be out fairly soon now, but maybe. I've cherry-picked this patch to the `2.4` branch.

23Skidoo:
I noticed that `--ghc-options` is missing from the `help v2-install` output, unlike `v1-install`, for which we have:

```
The flags --with-PROG and --PROG-option(s) can be used with the following programs:
  alex ar c2hs cpphs doctest gcc ghc ghc-pkg ghcjs ghcjs-pkg greencard haddock
  happy haskell-suite haskell-suite-pkg hmake hpc hsc2hs hscolour jhc ld
  pkg-config runghc strip tar uhc
```

hvr:
@phadej @23Skidoo I've been wondering about backporting this to the 2.4 branch; I'm just a bit hesitant as this can reasonably be considered a "breaking change" in the CLI, even though it's fixing something that was clearly wrong in the CLI... any opinions whether we should or should not consider this for the `cabal` 2.4.2 point rls?

23Skidoo:
I'm +1 on backporting.

PR #5937, opened 2019-03-13
Tidy up AppVeyor slightly

Mistuke:
The new `cabal-head` packages now automatically detect AppVeyor and set the required paths. (You can see this in the build log)
also use AppVeyor time better by not building branches to open PRs since the PR itself is also built.

I play to push this out to the stable branch with the next `cabal-install` release.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ install:
   - choco install -y ghc --version 8.0.2 --ignore-dependencies
   - choco install -y cabal-head -pre
   - refreshenv
-  - set
+  - set PATH
   - cabal --version
   - cabal %CABOPTS% update
   - cabal %CABOPTS% v1-install happy alex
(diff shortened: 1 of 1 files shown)


23Skidoo:
Patch looks good, but the AppVeyor build seems to have failed.

23Skidoo:
Apparently it can't build `network` because Cygwin gets in the way.

Mistuke:
Yeah, I'm wondering why. Was gonna debug it later today.

23Skidoo:
My guess is that removing `C:\msys64\usr\bin` from `PATH` did. `ghc\tools` probably doesn't contain every GNU tool needed to run the `network`'s `configure` script, so it picks the Cygwin versions that come preinstalled on AppVeyor images.

23Skidoo:
Or maybe your package install scripts should prepend to `PATH` instead of appending.

Mistuke:
Hmm you're right it probably should prepend. I can't change the package
until Tuesday, will update it then!

Mistuke:
Ok that's better, seems that I had to override the global path not the user one.

23Skidoo:
Merged, thanks!

PR #5943, opened 2019-03-16
Fix some markdown in the local builds docs

dwijnand:
[ci skip]

---


23Skidoo:
Merged, thanks!

PR #5985, opened 2019-04-05
Disable .ghc-environment file generation by default

mpickering:
See https://mail.haskell.org/pipermail/ghc-devs/2019-March/017351.html for discussion about this default.

---

### DIFF
--- a/cabal-install/Distribution/Client/Types.hs
+++ b/cabal-install/Distribution/Client/Types.hs
@@ -597,8 +597,8 @@ instance Monoid AllowOlder where
 -- ------------------------------------------------------------
 
 -- | Whether 'v2-build' should write a .ghc.environment file after
--- success. Possible values: 'always', 'never', 'ghc8.4.4+' (the
--- default; GHC 8.4.4 is the earliest version that supports
+-- success. Possible values: 'always', 'never' (the default), 'ghc8.4.4+'
+-- (8.4.4 is the earliest version that supports
 -- '-package-env -').
 data WriteGhcEnvironmentFilesPolicy
   = AlwaysWriteGhcEnvironmentFiles
--- a/cabal-install/changelog
+++ b/cabal-install/changelog
@@ -1,6 +1,7 @@
 -*-change-log-*-
 
 3.0.0.0 (current development version)
+	* '--write-ghc-environment-files' now defaults to 'never' (#4242)
 	* Fix `sdist`'s output when sent to stdout. (#5874)
 	* Allow a list of dependencies to be provided for `repl --build-depends`. (#5845)
 	* Legacy commands are now only accessible with the `v1-` prefixes, and the `v2-`
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -405,7 +405,7 @@ runProjectPostBuildPhase verbosity
           $ projectConfig
 
         shouldWriteGhcEnvironment =
-          case fromFlagOrDefault WriteGhcEnvironmentFilesOnlyForGhc844AndNewer
+          case fromFlagOrDefault NeverWriteGhcEnvironmentFiles
                writeGhcEnvFilesPolicy
           of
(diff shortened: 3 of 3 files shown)


cabal-install/changelog
```diff
@@ -1,6 +1,7 @@
 -*-change-log-*-
 
 3.0.0.0 (current development version)
+	* write-ghc-environment-files now defaults to never (#4242)
```
23Skidoo:
```suggestion
	* '--write-ghc-environment-files' now defaults to 'never' (#4242)
```

cabal-install/Distribution/Client/Types.hs
```diff
@@ -597,8 +597,8 @@ instance Monoid AllowOlder where
 -- ------------------------------------------------------------
 
 -- | Whether 'v2-build' should write a .ghc.environment file after
--- success. Possible values: 'always', 'never', 'ghc8.4.4+' (the
--- default; GHC 8.4.4 is the earliest version that supports
+-- success. Possible values: 'always', 'never', 'ghc8.4.4+'
```
23Skidoo:
```suggestion
-- success. Possible values: 'always', 'never' (the default), 'ghc8.4.4+'
```

mpickering:
@23Skidoo I made your suggested changes.

vaibhavsagar:
Thank you for doing this @mpickering!

hvr:
Just as a reminder, we need to make a couple of follow-up changes to prevent bad user-experience as a result of this change of defaults, as ironically changing the default this way will guarantee a couple of regression for cabal users if caught off guard.

gbaz:
Let me expand on that -- in an irc discussion with someone who had been quite frustrated with env files, it turns out the problem was precisely that we'd limited their auto generation to 8.4.4 and after. Hence, they had files for 8.4.3, etc projects that were out of date, and which weren't getting auto-regenerated (due to changing defaults), and this was what made the env files fail for them.

So, in particular, we need to handle the case where someone updates cabal, then env files _stop_ being generated, and then workflows break without their realizing it.

What this suggests to me is we need some behavior when env files already exist that A) detects them and B) either warns, offers to delete, or says "if they already exist, then update them" -- i.e. the default is not "don't create or update" but rather "don't create, but do update".

hsenag:
"don't create but do update" sounds good for the long-term behaviour as it's then opt-in - particularly if it can detect the cabal-generated files and only update those and not one created directly by a user.

In common with many others I'm against the current behaviour: it's both a layering violation (a high-level tool cabal modifying the behaviour of a low-level tool ghci) and a silent side-effect.

kcsongor:
@gbaz I like your proposal about issuing a warning when the environment file is not being generated, as I've run into a similar situation myself.
However, I think it might be better to always show the notice, even if old env files don't exist. My travis CI broke after the cabal-head package was updated, and my doctests (which relied on the env files) suddenly stopped working. Since travis runs a clean build every time, the warning would not have been shown. If I got a warning about the env file not being generated, it would have been a little easier to get to the bottom of the problem.

Would always issuing a warning cause issues elsewhere?

phadej:
Showing the notice that “environment files not generated” would be acceptable only for verbose output, but then it would be good to be implicit; especially if “don’t create, but update” variant is added.

phadej:
I meant explicit as in always be present, whatever the setting is

gbaz:
One other suggestion I would like to make -- I think we should have the "disable" setting in general control if the files are created in the project directory. However, we should probably either also unconditionally (or at least when files are not created in the project directory) create an appropriately named file (perhaps without a dot) in the dist-newstyle (or dist, i don't remember if we're renaming dist-newstyle to dist or not) directory, so that tools which start to make use of these files explicitly can know where to find one regardless of other settings.

Since part of the whole issue has been to make things as seamless as possible for various tooling, I think keeping that design consideration in mind going forward as well could be important.

phadej:
I merged this. It's already an improvement, and following refinements can be done separately.

23Skidoo:
Also cherry-picked into 2.4.

PR #5987, opened 2019-04-07
Convert all new- to v2- in docs [ci skip]

phadej:
---

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -1475,7 +1475,7 @@ dependencies. The ``cabal outdated`` command is designed to help with
 that. It will print a list of packages for which there is a new
 version on Hackage that is outside the version bound specified in the
 ``build-depends`` field. The ``outdated`` command can also be
-configured to act on the freeze file (both old- and new-style) and
+configured to act on the freeze file (both old- and v2-style) and
 ignore major (or all) version bumps on Hackage for a subset of
 dependencies.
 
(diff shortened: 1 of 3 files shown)


Cabal/doc/developing-packages.rst
```diff
@@ -1486,14 +1486,14 @@ The following flags are supported by the ``outdated`` command:
     instead of the package description file (``$PACKAGENAME.cabal``).
     ``--v1-freeze-file`` is an alias for this flag starting in Cabal 2.4.
 ``--new-freeze-file``
-    Read dependency version bounds from the new-style freeze file
+    Read dependency version bounds from the v2-style freeze file
     (by default, ``cabal.project.freeze``) instead of the package
-    description file. ``--v2-freeze-file`` is an alias for this flag
+    description file. ``--new-freeze-file`` is an alias for this flag
```
quasicomputational: This one should still be `v2`, presumably. Or maybe the head-term should be `--v2-freeze-file` and this should be reworded to say that before 2.4 it was only `--new-freeze-file`, but both are accepted now.

Cabal/doc/developing-packages.rst
```diff
@@ -1475,7 +1475,7 @@ dependencies. The ``cabal outdated`` command is designed to help with
 that. It will print a list of packages for which there is a new
 version on Hackage that is outside the version bound specified in the
 ``build-depends`` field. The ``outdated`` command can also be
-configured to act on the freeze file (both old- and new-style) and
+configured to act on the freeze file (both old- and v2-style) and
```
quasicomputational: s/old-/v1-/?

phadej: Won't do. Someone else can do another `sed`.

phadej: As in, I'm not sure that `old-` won't match something non valid, and I have no time to check. Sorry. Take or leave.

Cabal/doc/developing-packages.rst
phadej: There are no `--v2-freeze-file`, at least not in

```
% cabal outdated --help
Check for outdated dependencies

Usage: cabal outdated [FLAGS]

Checks for outdated dependencies in the package description file or freeze
file

Flags for outdated:
 -h --help                     Show this help text
 -v --verbose[=n]              Control verbosity (n is 0--3, default verbosity
                               level is 1)
    --freeze-file              Act on the freeze file
    --new-freeze-file          Act on the new-style freeze file (default:
                               cabal.project.freeze)
    --project-file=PROJECTFILE Act on the new-style freeze file named
                               PROJECTFILE.freeze rather than the default
                               cabal.project.freeze
    --simple-output            Only print names of outdated dependencies, one
                               per line
    --exit-code                Exit with non-zero when there are outdated
                               dependencies
 -q --quiet                    Don't print any output. Implies '--exit-code'
                               and '-v0'
    --ignore=PKGS              Packages to ignore
    --minor[=PKGS]             Ignore major version bumps for these packages
```

PR #5992, opened 2019-04-08
Support window 8.8

phadej:
---


23Skidoo:
Should this be backported to 2.4, btw?

phadej:
I don't see a point. `Cabal-3.0` will be bundled with GHC-8.8. So I'd rather concentrate on making that release solid.

Also, are you 100% sure, Cabal-2.4 can actually handle GHC-8.8? It's unusual that GHC-8.8.1 doesn't introduce new language extensions e.g. Anyway, I'd not spend any time on making old Cabal support new GHC. They are released in in lock-step anyway.

PR #5994, opened 2019-04-09
Don't export 'read': panic explicitly

phadej:
Motivated by failing test, which now gives more hints why it failed:

```
panic! read @TestSuiteLog ""
  error, called at ./Distribution/Simple/Test/LibV09.hs:127:34 in Cabal-3.0.0.0-inplace:Distribution.Simple.Test.LibV09
  runTest, called at ./Distribution/Simple/Test.hs:69:19 in Cabal-3.0.0.0-inplace:Distribution.Simple.Test
  doTest, called at ./Distribution/Simple/Test.hs:115:24 in Cabal-3.0.0.0-inplace:Distribution.Simple.Test
```

### DIFF
--- a/Cabal/Distribution/Compat/Prelude.hs
+++ b/Cabal/Distribution/Compat/Prelude.hs
@@ -91,10 +91,15 @@ module Distribution.Compat.Prelude (
 
     -- * Text.PrettyPrint
     (<<>>),
+
+    -- * Text.Read
+    readMaybe,
     ) where
 -- We also could hide few partial function
 import Prelude                       as BasePrelude hiding
   ( IO, mapM, mapM_, sequence, null, length, foldr, any, all
+  -- partial functions
+  , read
 #if MINVER_base_411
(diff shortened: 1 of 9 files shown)


validate.sh
```diff
@@ -196,7 +196,7 @@ BUILDDIR=dist-newstyle-validate-$BASEHC
 CABAL_TESTSUITE_BDIR="$(pwd)/$BUILDDIR/build/$ARCH/$BASEHC/cabal-testsuite-${CABAL_VERSION}"
 
 CABALNEWBUILD="${CABAL} v2-build $JOBS -w $HC --builddir=$BUILDDIR --project-file=$PROJECTFILE"
-CABALPLAN="${CABALPLAN} --builddir=$BUILDDIR"
+CABALPLANLISTBIN="${CABALPLAN} list-bin --builddir=$BUILDDIR"
```
23Skidoo: Can you please split the changes to `validate.sh` into a separate commit?

phadej: I can, I wasn't supposed to commit `validate.sh` changes at all.

PR #5995, opened 2019-04-10
Add test-wrapper option to wrap test execution

angerman:
Sometimes we need to run the test ins some prebuild context, for
these cases we now allow a script/application to be passed as
`--test-wrapper`, which will obtain the execution arguments and
be responsible for running the test in the environment.

A test-wrapper could for example do the following to execute a
test through wine:

```
set -euo pipefail
WINEDLLOVERRIDES="winemac.drv=d" \
  WINEDEBUG=-all+error \
  LC_ALL=en_US.UTF-8 \
  WINEPREFIX=$(mktemp -d) \
  /path/to/bin/wine64 $@*
```

---

### DIFF
--- a/Cabal/Distribution/Simple/Setup.hs
+++ b/Cabal/Distribution/Simple/Setup.hs
@@ -1864,6 +1864,7 @@ data TestFlags = TestFlags {
     testMachineLog  :: Flag PathTemplate,
     testShowDetails :: Flag TestShowDetails,
     testKeepTix     :: Flag Bool,
+    testWrapper     :: Flag FilePath,
     testFailWhenNoTestSuites :: Flag Bool,
     -- TODO: think about if/how options are passed to test exes
     testOptions     :: [PathTemplate]
@@ -1877,6 +1878,7 @@ defaultTestFlags  = TestFlags {
     testMachineLog  = toFlag $ toPathTemplate $ "$pkgid.log",
     testShowDetails = toFlag Failures,
     testKeepTix     = toFlag False,
+    testWrapper     = NoFlag,
     testFailWhenNoTestSuites = toFlag False,
     testOptions     = []
   }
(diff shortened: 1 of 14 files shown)


23Skidoo:
Looks fine in principle, but we should think about what happens when `cabal test` passes `--test-wrapper` to a custom `Setup` script compiled againt Cabal < 3.0. We may need add a variant of `filterConfigureFlags` for the `test` command.

Also needs a changelog note and documentation, you can add the latter here:
* https://github.com/haskell/cabal/blob/master/Cabal/doc/nix-local-build.rst#L516
* https://github.com/haskell/cabal/blob/master/Cabal/doc/installing-packages.rst#setup-test

cabal-install/Distribution/Client/ProjectPlanning.hs
```diff
@@ -1884,6 +1884,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
         elabTestHumanLog        = perPkgOptionMaybe pkgid packageConfigTestHumanLog
         elabTestShowDetails     = perPkgOptionMaybe pkgid packageConfigTestShowDetails
         elabTestKeepTix         = perPkgOptionFlag pkgid False packageConfigTestKeepTix
+        elabTestWrapper         = perPkgOptionFlag pkgid False packageConfigTestWrapper
```
harpocrates: Did you mean `perPkgOptionMaybe pkgid packageConfigTestWrapper`?

Cabal/Distribution/Simple/Setup.hs
```diff
@@ -1942,6 +1944,11 @@ testOptions' showOrParseArgs =
         "keep .tix files for HPC between test runs"
         testKeepTix (\v flags -> flags { testKeepTix = v})
         trueArg
+  , option [] ["test-wrapper"]
+        "Run test through a wrapper."
+        testWrapper (\v flags -> flags { testWrapper = v })
+        (reqArg' "FILE" (toFlag :: FilePath -> Flag FilePath)
+            (flagToList :: Flag FilePath -> [FilePath]))
```
harpocrates: You need to add this to at least `Distribution.Cabal.Setup.testOptions` if you want it to actually make it through to a user-facing option.

angerman:
> I think this is still WIP, right? Even after fixing some more obvious issues, I wasn't able to reproduce the expected behaviour (the wrapper script I passed in was being ignored).
> 
> Might be helpful to mark as WIP until ready for actual review. 🙂

This is actually being used with the low level plumbing of the `Setup.hs`, so yes there might be something missing for the `cabal-install` frontend :-)

JoshMeredith:
> We may need add a variant of `filterConfigureFlags` for the `test` command.

What would be the version number used in the code for `filterTestFlags`, given that the project files use a placeholder `3.0.0` version number?

23Skidoo:
Using 3.0.0.0 is fine.

23Skidoo:
@JoshMeredith There are some CI failures still unfortunately.

23Skidoo:
Looks like some QuickCheck tests are now failing.

JoshMeredith:
Are the current changes good to merge?

23Skidoo:
@JoshMeredith Yes, feel free to push the big green button!

PR #6008, opened 2019-04-21
Cabal: don't use normalise for copyFile on Windows

Mistuke:
`normalise` from `System.FilePath` has an annoying behavior in that it doesn't simplify `FilePath`. The documented behavior on Windows is to normalise `Z://` into `Z:\\\\` which in the `Win32` namespace won't matter but since we're using the `device` namespace `\\?\Z:\\` is an invalid path to the filesystem.

This results in cabal throwing an illegal path exception for such cases which happen to be fairly common when building `GHC`.  This fixes it by inlining some more code from `System.Directory` which has a function which correctly preserves the semantics of a path when converted to a device path.

Tested with a full GHC bootstrap and testrun. Fixes https://gitlab.haskell.org/ghc/ghc/merge_requests/619

---

### DIFF
--- a/Cabal/Distribution/Compat/CopyFile.hs
+++ b/Cabal/Distribution/Compat/CopyFile.hs
@@ -123,6 +123,8 @@ copyFile fromFPath toFPath =
 -- | Add the @"\\\\?\\"@ prefix if necessary or possible.  The path remains
 -- unchanged if the prefix is not added.  This function can sometimes be used
 -- to bypass the @MAX_PATH@ length restriction in Windows API calls.
+--
+-- See Note [Path normalization].
 toExtendedLengthPath :: FilePath -> FilePath
 toExtendedLengthPath path
   | isRelative path = path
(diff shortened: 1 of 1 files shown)


phadej:
Is there a corresponding patch to `filepath`?

Mistuke:
No, because `FilePath` 's `normalise` doesn't claim to simplify paths and doesn't prepare paths for use in any namespace other than the `win32` one. So what it's doing is at least according to it's documentation.

Usually it won't matter as once the path hits the RTS we'll do any fixups needed.  The issues here is that we call the `Win32 CopyFile` API directly, in order not to go through the GHC I/O manager, so we have to deal with the path fixups.

phadej:
If you put your explanation as a Note in the code, then this is good to go :+1:

Mistuke:
All good @phadej?

phadej:
Feel free to merge

Mistuke:
Thanks!

PR #6019, opened 2019-04-27
Add static flag to cabal executable

vmchale:
Adds a new flag to the .cabal file for cabal-install. This makes cross-compilation a little easier.

---

### DIFF
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -102,6 +102,11 @@ source-repository head
   location: https://github.com/haskell/cabal/
   subdir:   cabal-install
 
+Flag static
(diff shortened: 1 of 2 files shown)


23Skidoo:
Isn't this what the new `--enable-executable-static` `configure` flag is for?

23Skidoo:
Also the meta check fails because you need to update the `cabal-install.pp` file.

23Skidoo:
@vmchale Any reason you can't use `configure --enable-executable-static` instead? (besides that, IIRC, it's only available in `master`)

vmchale:
Nope, should work for me. Closing :)

PR #6033, opened 2019-05-05
Implement {cmm.asm}-{options,sources} for real

hvr:
While those buildinfo fields were added to the parser some time
ago via 57d7f28ebf9e513030f2c1d0ebf164a1151dbe06 and
4a2876594ce1029d3d84c18d4d41c73c312fa2e0 that work was never completed
by implementing the necessary build/sdist logic in Cabal.

This commit remedies this oversight by implementing and wiring up the
missing build logic.

cc @erikd

----

Things that work now (i.e. that I tested manually; these should be turned into testsuite tests long-term):

- `cmm-sources` and `asm-sources` are compiled (in a separate phase like `c-sources` are) and the resulting object files are linked into the respective component (i.e. like `c-sources`)
- `cabal sdist` properly includes `cmm-sources` and `asm-sources` in the src-dist
- `cmm-options` are passed (directly) to GHC when compiling `cmm-sources`
- `asm-options` are passed (via `-opta`) to GHC when compiling `asm-sources` files
- `ghc-options` are *NOT* passed to compilation of `cmm-sources`/`asm-sources`


angerman:
> oversight

🤔

23Skidoo:
TIL about Draft PRs, nice feature.

hsyl20:
I've added some missing C-- bits to @hvr's patch: https://github.com/hsyl20/cabal/tree/hsyl20-asm-cmm-sources
Hopefully we could have this patch for GHC 8.8

hvr:
@hsyl20 thank you! I'll look at your branch asap and will cherry-pick the delta into this PR if it looks good

angerman:
🎉

hvr:
@hsyl20  I tested the code a bit, and it works partly; specifically `cmm-options` are not passed to the `ghc` commandline when compiling the `.cmm` module

Cabal/Distribution/Simple/GHC/Internal.hs
```diff
@@ -396,15 +434,50 @@ componentGhcOptions verbosity implInfo lbi bi clbi odir =
...
+                                          ++ [buildDir lbi </> dir | dir <- PD.includeDirs bi],
+      ghcOptCppOptions     = cppOptions bi,
+      ghcOptCppIncludes    = toNubListR $
+                             [autogenComponentModulesDir lbi clbi </> cppHeaderName],
+      ghcOptHideAllPackages= toFlag True,
+      ghcOptPackageDBs     = withPackageDB lbi,
+      ghcOptPackages       = toNubListR $ mkGhcOptPackages clbi,
+      ghcOptOptimisation   = toGhcOptimisation (withOptimization lbi),
+      ghcOptDebugInfo      = toFlag (withDebugInfo lbi),
+      ghcOptExtra          = hcOptions GHC bi,
```
hsyl20: Oops I forgot to fix this while copy-pasting from the other cases: s/hcOptions GHC bi/cmmOptions bi/
I guess it should fix cmm-options passing.

hvr: yeah... this is the kind of bugs I was worried about... :-)

I've rebased and updated the PR to include your fixup suggestion; I'm currently testdriving this locally

I'm also going to un-DRAFT this PR to invite more code-review

Cabal/doc/developing-packages.rst
```diff
@@ -2550,8 +2551,14 @@ system-dependent values for these fields.
     command-line arguments with the :pkg-field:`cc-options` and the
     :pkg-field:`cxx-options` fields.
 
+.. pkg-field:: cmm-options: token list
```
hvr: The thing that worries me here is that this should rather be called `ghc-cmm-options`, because we don't really have an actual C-- compiler -- we're just invoking GHC; and `cmm-options` would imply that the flags you can pass to the `C--` compiler are somewhat universal to C-- compilation.

hsyl20: I would keep `cmm-options` here. Mainly for consistency and also because the fact that we use GHC as a `C--` compiler is incidental and might change in the future if we split GHC into smaller components.

23Skidoo: Agreed that `ghc-cmm-options` would be a better name.

Cabal/Distribution/Simple/Build.hs
```diff
@@ -356,6 +362,24 @@ addExtraCxxSources bi extras = bi { cxxSources = new }
...
+  where new = Set.toList $ old `Set.union` exs
+        old = Set.fromList $ cmmSources bi
+        exs = Set.fromList extras
+
+
+-- | Add extra ASM sources generated by preprocessing to build
+-- information.
+addExtraAsmSources :: BuildInfo -> [FilePath] -> BuildInfo
+addExtraAsmSources bi extras = bi { asmSources = new }
+  where new = Set.toList $ old `Set.union` exs
```
23Skidoo: If I have both auto-generated `asm-sources` and `cmm-sources`, won't this add the union of these to the LBI's `asmSources`?

23Skidoo: Actually, this probably will break if I have any preprocessor-generated C files at all.

23Skidoo:
Would be nice to see a test case, would be even nicer if the test case had a custom preprocessor that generates `.s` and `.cmm` files.

Cabal/Distribution/Simple/Build.hs
```diff
@@ -216,7 +216,13 @@ buildComponent verbosity numJobs pkg_descr lbi suffixes
     setupMessage' verbosity "Building" (packageId pkg_descr)
       (componentLocalName clbi) (maybeComponentInstantiatedWith clbi)
     let libbi = libBuildInfo lib
-        lib' = lib { libBuildInfo = addExtraCxxSources (addExtraCSources libbi extras) extras }
```
hvr: @23Skidoo based on your observation, wasn't this line already broken? i.e. the same FilePaths being added to cxx *and* c sources?

23Skidoo: Yes, looks like it.

hvr:
@23Skidoo as discussed I've backported the current state of this PR to the 3.0 branch via c5395b853c0c348955c44dd54e2c4a5887b06e1d and this PR remains open for the express purpose to fix the outstanding issues with the preprocessor generated sources

Cabal/doc/developing-packages.rst
phadej: This ship has unfortunately sailed.

phadej:
https://github.com/haskell/cabal/pull/6376 merged, https://github.com/haskell/cabal/issues/6377 is tracking leftovers

Issue #6039, opened 2019-05-07
Make the cabal-install solver multilibs-aware


fgaz:
The Cabal part is done, but cabal-install's solver is not aware of the multiple public libraries feature yet, leading to late errors and issues like #6038

grayjay:
I think this would be pretty straightforward, at least for source packages, because it could be implemented similarly to enforcing build tool dependencies.  The visibility field could probably be implemented similarly to the current handling of unbuildable components.  I have a few questions about how visibility should behave, though.

- Should the solver enforce visibility now?  I wasn't sure if the comments about disabling the visibility field in #5848 applied here.
- Should the solver just check the visibility fields in `Library` and `InstalledPackageInfo`?  Can the main library be private?

There may be an issue with enforcing visibility for installed packages, because there is a hack where the solver treats installed internal libraries as separate packages with munged names (https://github.com/haskell/cabal/blob/5f57d4d0ee14592cc488385151c15edef59f9bd6/cabal-install/Distribution/Solver/Modular/IndexConversion.hs#L106-L131).  I'm not sure how difficult it will be to distinguish between inter- and intra-package dependencies.

fgaz:
> Should the solver enforce visibility now? I wasn't sure if the comments about disabling the visibility field in #5848 applied here.

Yes. In #5848 we're talking about disabling it at the syntax level, but internally it's still used

> Should the solver just check the visibility fields in Library and InstalledPackageInfo? Can the main library be private?

Yes and no, but I'll have to check the second one.

grayjay:
@fgaz Thanks.  I made a PR (#6047), but it only checks the visibility of source packages so far.  It reads the libVisibility field for all libraries, though I can easily change it to only check sub-libraries if that is the correct behavior.

grayjay:
#6836 enforces visibility for source packages, so the only remaining task is to handle installed packages.

grayjay:
@gbaz The only thing that is missing in the dependency solver is enforcing visibility restrictions for installed packages, so that one package cannot depend on a private installed sub-library from another package. The change could be difficult due to the hack mentioned in https://github.com/haskell/cabal/issues/6039#issuecomment-491195495.

I also noticed that #7270 may relate to the solver, though I haven't looked into it yet.

grayjay:
[...]

PR #6047, opened 2019-05-13
Solver: Support dependencies on sub-libraries (issue #6039).

### LINKED ISSUE #6038: cabal-install: cannot depend on public sublibrary when the package does not contain a main library too
Repro:

cabal.project:
```cabal.project
packages: d, p
```

d.cabal:
```cabal
cabal-version: 3.0
name: d
version: 0.1.0.0

library sublib
  visibility: public
  default-language: Haskell2010
```

p.cabal:
```cabal
cabal-version: 3.0
name: p
version: 0.1.0.0

library
  build-depends: d:sublib
  default-language: Haskell2010
```

`cabal v2-build p` output:
```
[__0] trying: d-0.1.0.0 (user goal)
[__1] next goal: p (user goal)
[__1] rejecting: p-0.1.0.0 (requires library from d, but the component does
not exist)
[__1] fail (backjumping, conflict set: d, p)
goals I've had most trouble fulfilling: d, p
```

Adding an empty main library to `d` fixes it.:w

grayjay:
This commit tracks dependencies on sub-libraries by extending the functionality
for tracking executables that was added in
e86f83890d93068f7c27faea5bf07146b5452c23.  It also starts adding support for
library visibility, though it currently only works for source packages.  There
is a TODO for handling installed packages.

Fixes #6038.

/cc @fgaz @kosmikus 

---

### DIFF
--- a/cabal-install/Distribution/Solver/Modular/IndexConversion.hs
+++ b/cabal-install/Distribution/Solver/Modular/IndexConversion.hs
@@ -7,6 +7,7 @@ import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as M
 import Data.Maybe (mapMaybe, fromMaybe, maybeToList)
 import Data.Monoid as Mon
+import qualified Distribution.Compat.NonEmptySet as NonEmptySet
 import qualified Data.Set as S
 
 import qualified Distribution.InstalledPackageInfo as IPI
@@ -571,7 +572,7 @@ unionDRs (DependencyReason pn' fs1 ss1) (DependencyReason _ fs2 ss2) =
 convLibDeps :: DependencyReason PN -> Dependency -> [LDep PN]
 convLibDeps dr (Dependency pn vr libs) =
     [ LDep dr $ Dep (PkgComponent pn (ExposedLib lib)) (Constrained vr)
-    | lib <- S.toList libs ]
+    | lib <- NonEmptySet.toList libs ]
 
 -- | Convert a Cabal dependency on an executable (build-tools) to a solver-specific dependency.
 convExeDep :: DependencyReason PN -> ExeDependency -> LDep PN
--- a/cabal-install/Distribution/Solver/Modular/IndexConversion.hs
+++ b/cabal-install/Distribution/Solver/Modular/IndexConversion.hs
@@ -25,6 +25,7 @@ import Distribution.PackageDescription.Configuration
 import qualified Distribution.Simple.PackageIndex as SI
 import Distribution.System
 import Distribution.Types.ForeignLib
+import Distribution.Types.LibraryVisibility
 
 import           Distribution.Solver.Types.ComponentDeps
                    ( Component(..), componentNameToComponent )
@@ -92,11 +93,18 @@ convIP :: SI.InstalledPackageIndex -> IPI.InstalledPackageInfo -> (PN, I, PInfo)
 convIP idx ipi =
   case mapM (convIPId (DependencyReason pn M.empty S.empty) comp idx) (IPI.depends ipi) of
         Nothing  -> (pn, i, PInfo [] M.empty M.empty (Just Broken))
-        Just fds -> ( pn
-                    , i
-                    , PInfo fds (M.singleton ExposedLib (IsBuildable True)) M.empty Nothing)
+        Just fds -> ( pn, i, PInfo fds components M.empty Nothing)
  where
+  -- TODO: Handle sub-libraries and visibility.
+  components =
+      M.singleton (ExposedLib LMainLibName)
+                  ComponentInfo {
+                      compIsVisible = IsVisible True
+                    , compIsBuildable = IsBuildable True
+                    }
+
   (pn, i) = convId ipi
+
   -- 'sourceLibName' is unreliable, but for now we only really use this for
   -- primary libs anyways
   comp = componentNameToComponent $ CLibName $ IPI.sourceLibName ipi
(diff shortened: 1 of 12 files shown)


grayjay:
I just realized that this PR probably doesn't handle visibility correctly.  It only reads the top-level visibility field, but the field can also appear in conditionals.

Handling conditional visibility would probably require something similar to the "component on/off" flags in https://github.com/haskell/cabal/issues/4087#issuecomment-287665861, though that would be a significant change.

Another option would be to handle visibility similarly to the way buildable is handled after #5337.  At the start of dependency solving, we could check whether a component is private with the current environment and flag constraints, which would allow the solver to reject most dependencies on private libraries.  A visibility field controlled by an automatic flag could still cause a build failure at the end of solving.

phadej:
Is there a use case for visibility controlled by an automatic flag? If there aren’t, then cabal check could reject such definitions?

23Skidoo:
> If there aren’t, then cabal check could reject such definitions?

I think it'd be totally fine to do this, we can always relax this restriction later.

grayjay:
I like the idea of disallowing visibility controlled by automatic flags, if it would be easy to reverse later.  I could update this PR to try to determine whether libraries are visible at the start of solving, as I described above, which should always be correct when automatic flags aren't allowed.

23Skidoo:
@grayjay SGTM.

grayjay:
I finally updated this PR to determine whether each library is visible in the current environment at the start of solving.  The solver checks the visibility of the main library and sub-libraries, though I'm not sure if it is possible for the main library to be private.

cc @fgaz

grayjay:
I'm planning to merge this next weekend if no one objects.

phadej:
I was going to (try to) stabilize `master` for 3.2 release this week. How high risk that this introduces regressions?

phadej:
Yet, this doesn't change anything in `Cabal` the library, so as far as CI is happy, I'm happy too.

grayjay:
Yes, it only affects the solver.  I think there is a small risk of a regression in cabal-install, especially since this PR only contains unit tests, and, as far I know, there isn't a positive integration test for depending on sublibraries yet.

phadej:
Wouldn't a positive test be a copy of `cabal-testsuite/PackageTests/MultipleLibraries` with a `visibility` toggled?

grayjay:
I'm not sure.  It looks like there some complexity with the GHC support for sublibraries in the discussion in #5848.

phadej:
We can skip that test for GHC <8.8, I'll try this today as well.

phadej:
I'm milestoning this for 3.2. Let's get this in. If I introduce conflicts I can do the merging.

fgaz:
Sorry, I've been busy lately. I'll take a look before the weekend

grayjay:
@phadej @fgaz Thanks for looking into the integration testing.  I'm fine with merging this before or after the release.

cabal-install/Distribution/Solver/Modular/IndexConversion.hs
```diff
@@ -93,11 +94,18 @@ convIP :: SI.InstalledPackageIndex -> InstalledPackageInfo -> (PN, I, PInfo)
 convIP idx ipi =
   case mapM (convIPId (DependencyReason pn M.empty S.empty) comp idx) (IPI.depends ipi) of
         Nothing  -> (pn, i, PInfo [] M.empty M.empty (Just Broken))
-        Just fds -> ( pn
-                    , i
-                    , PInfo fds (M.singleton ExposedLib (IsBuildable True)) M.empty Nothing)
+        Just fds -> ( pn, i, PInfo fds components M.empty Nothing)
  where
+  -- TODO: Handle sub-libraries and visibility.
```
fgaz: What does this entail?

cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs
```diff
@@ -281,23 +281,79 @@ tests = [
...
 
-        , let exe = ExExe "exe" []
+        , let exe = exExe "exe" []
               db = [ Right $ exAv "A" 1 [ExAny "B"]
                    , Right $ exAvNoLibrary "B" 2 `withExe` exe
                    , Right $ exAv "B" 1 [] `withExe` exe ]
           in runTest $ mkTest db "choose version of build-depends dependency that has a library" ["A"] $
              solverSuccess [("A", 1), ("B", 1)]
         ]
+    , testGroup "sub-library dependencies" [
```
fgaz: This is great :)

fgaz:
Almost forgot: would it be possible to give a better error when using ghc<8.8 and depending on a public library?

grayjay:
@fgaz Thanks for the review!

> Almost forgot: would it be possible to give a better error when using ghc<8.8 and depending on a public library?

How should the solver handle ghc<8.8?  Should it disallow inter-package dependencies on sublibraries?

cabal-install/Distribution/Solver/Modular/IndexConversion.hs
grayjay: I forgot to look into handling installed packages.  It could be tricky because of the hack where the solver treats installed internal libraries as packages with munged names:

https://github.com/haskell/cabal/blob/8f42d3fa9b1257f80b55cc9b35e9da1a2a8195f3/cabal-install/Distribution/Solver/Modular/IndexConversion.hs#L106-L131

I would need to figure out how to distinguish between inter- and intra-package dependencies when enforcing visibility.

fgaz:
There are two options:
* disallow inter-package dependencies on sublibraries, since they will fail Cabal's checks
* override Cabal's checks with the `--allow-depending-on-private-libs` flag (see https://github.com/haskell/cabal/pull/5848#pullrequestreview-209916319 and later comments), so that cabal-install can enable the multilibs functionality even on older ghcs
I think it was agreed to go with the second
/cc @hvr @ezyang

grayjay:
@fgaz Does the second option (override with --allow-depending-on-private-libs) mean that the solver doesn't need to give a different error for ghc<8.8?

fgaz:
Yes

phadej:
@grayjay I noticed you worked on this recently. Do you think we can get this merged in the beginning of June 2020?

I optimistally milestone this for 3.4. But if the above deadline is too tight, feel free to de-milestone.

grayjay:
@phadej I was only rebasing the branch.  I think that the only thing blocking this PR was the integration test, but I see that you added tests in #6409.  Thanks for adding them!  This PR still has a TODO for handling installed packages, but I would rather do that in a separate PR.

grayjay:
This was merged in #6836.

Issue #5862, opened 2019-01-25
Cabal disagrees on the `datadir` in the `Path_` module, and where it ends up installing data files.


angerman:
The issue seems to be the handling of components. If we explicitly ask for the component, we end up with a fully qualified `$pkgid`, if we don't we get the plain pkg name.  This can be illustrated on the `alex` packaged quite neatly:

```
$ runghc Setup.hs configure exe:alex 
$ runghc Setup.hs build exe:alex 
$ cat dist/build/alex/autogen/Paths_alex.hs |grep datadir
bindir, libdir, dynlibdir, datadir, libexecdir, sysconfdir :: FilePath
datadir    = "/usr/local/share/x86_64-osx-ghc-8.6.3/alex-3.2.4-BnUEMKTG1CcF5xN1sTDkoW-alex"
getDataDir = catchIO (getEnv "alex_datadir") (\_ -> return datadir)
```
vs.
```
$ runghc Setup.hs configure
$ runghc Setup.hs build
$ cat dist/build/alex/autogen/Paths_alex.hs |grep datadir
bindir, libdir, dynlibdir, datadir, libexecdir, sysconfdir :: FilePath
datadir    = "/usr/local/share/x86_64-osx-ghc-8.6.3/alex-3.2.4"
getDataDir = catchIO (getEnv "alex_datadir") (\_ -> return datadir)
```

Now this wouldn't be an issue, if `runghc Setup.hs install` installed the data files into `/usr/local/share/x86_64-osx-ghc-8.6.3/alex-3.2.4-BnUEMKTG1CcF5xN1sTDkoW-alex` for the component configured case, but instead it will install them in `/usr/local/share/x86_64-osx-ghc-8.6.3/alex-3.2.4`, which then gives rise to amazing errors at runtime :(

This seems mildly related to #5543, but I'm not sure it's the identical situation.

angerman:
Could this be the patch I'm looking for?
```diff
diff --git a/Cabal/Distribution/Simple/Build/PathsModule.hs b/Cabal/Distribution/Simple/Build/PathsModule.hs
index f6a203874..2aa406f74 100644
--- a/Cabal/Distribution/Simple/Build/PathsModule.hs
+++ b/Cabal/Distribution/Simple/Build/PathsModule.hs
@@ -192,10 +192,14 @@ generate pkg_descr lbi clbi =
           bindir     = flat_bindir,
           libdir     = flat_libdir,
           dynlibdir  = flat_dynlibdir,
[...]

PR #6055, opened 2019-05-22
Fix datadir path in generated `Paths_` file

JoshMeredith:
The generated datadir path can be incorrect if a project is configured for a single component as per https://github.com/haskell/cabal/issues/5862

Since the install command uses the non-component variant of the paths for the datadir, https://github.com/haskell/cabal/blob/27ef964abfa78e9b2013e51bd197f6c2321a36d0/Cabal/Distribution/Simple/Install.hs#L98-L114 this patch changes the generated path to use the same function.

---

### DIFF
--- a/Cabal/Distribution/Simple/Build/PathsModule.hs
+++ b/Cabal/Distribution/Simple/Build/PathsModule.hs
@@ -192,10 +192,18 @@ generate pkg_descr lbi clbi =
           bindir     = flat_bindir,
           libdir     = flat_libdir,
           dynlibdir  = flat_dynlibdir,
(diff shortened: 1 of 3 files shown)


phadej:
Can we refactor so "consistency" is obvious via code sharing?

JoshMeredith:
> Can we refactor so "consistency" is obvious via code sharing?

Yes, I think it would be a good idea to make for example `absoluteMixedInstallDirs`, although I don't know which variant would be a better default for the fields of `InstallDirs` that are unused by the install command and the generated file.

PR #6058, opened 2019-05-26
Improve error message for 'new-test' command

erikd:
Suggest adding '--enable-tests'. [ci skip]

---


erikd:
Hmm, yeah. For projects with a single cabal this is the right message, but maybe it should be changed for multi cabal file projects.

erikd:
Had private IM discussion with @dcoutts about this issue and this is the change we came to agree on.

PR #6063, opened 2019-05-31
Pass `@...` as a literal argument if the response file does not exist and support recursive response files

### LINKED ISSUE #6061: cabal --index-state broken
Another `@`-related regression in HEAD I stumbled over:

```
cabal build --index-state @1474732068
```

used to work for various v2 commands... but now it fails with

```
Error while expanding response file: 1474732068: openFile: does not exist (No such file or directory)
```

/cc @angerman @JoshMeredith

JoshMeredith:
Fixes #6061

Arguments to `--index-state` also use the `@` syntax, so instead of exiting when a non-existent response file is passed as an argument, print a warning and pass it as a literal argument. `--index-state=@1234` continues to work without warnings.

---

### DIFF
--- a/Cabal/Distribution/Compat/ResponseFile.hs
+++ b/Cabal/Distribution/Compat/ResponseFile.hs
@@ -1,13 +1,11 @@
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE ScopedTypeVariables #-}
 
 -- Compatibility layer for GHC.ResponseFile
 -- Implementation from base 4.12.0 is used.
 -- http://hackage.haskell.org/package/base-4.12.0.0/src/LICENSE
 module Distribution.Compat.ResponseFile (expandResponse) where
 
-import Control.Exception
-import System.IO
+import System.IO.Error
 
 #if MIN_VERSION_base(4,12,0)
 import GHC.ResponseFile (unescapeArgs)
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Compat/ResponseFile.hs
```diff
@@ -57,8 +56,6 @@ expandResponse = fmap concat . mapM expand
     expand x = return [x]
 
     readFileExc f =
-      readFile f `catch` \(e :: IOException) -> do
-        hPutStrLn stderr $ "Error while expanding response file: " ++ show e
-        exitFailure
-
-#endif
+      readFile f `catch` \(_e :: IOException) -> do
+        hPutStrLn stderr $ "Response file `@" ++ f ++ "` does not exist, assuming literal argument."
```
hvr: do we really have to be that noisy?

23Skidoo: Hmm yeah, we can't use `notice` et al. here though since we don't know the verbosity level yet. Let's just be silent.

hvr: For reference, here's GNU Binutils' implementation of `@file`-expansion:

http://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=libiberty/argv.c;h=6444896f995a8e4abfb6a1edc510f3a4bbe1a30b;hb=HEAD#l342

Cabal/Distribution/Compat/ResponseFile.hs
```diff
@@ -47,8 +50,18 @@ unescape args = reverse . map reverse $ go args NoneQ False [] []
 #endif
 
 expandResponse :: [String] -> IO [String]
-expandResponse = fmap concat . mapM expand
+expandResponse = go recursionLimit "."
   where
-    expand :: String -> IO [String]
-    expand arg@('@':f) = unescapeArgs <$> readFile f `catchIOError` (const $ return arg)
-    expand x = return [x]
+    recursionLimit = 100
```
23Skidoo: `libiberty` uses 2000, but I'm okay with using 100.

JoshMeredith: I think they expand iteratively, so the 2000 is the total number of response files encountered in their expansion tree, whereas this 100 is tree depth.

Issue #6068, opened 2019-06-05
Cabal-3 breaks GHC's cabal01 test


bgamari:
GHC's `cabal01` test has a trivial `Setup.lhs` file:
```haskell
> import Distribution.Simple
> main = defaultMain
```

Bumping Cabal to 15675844bb36929448c189d6b4aabf7e853b3ee1 breaks this test:
```
=====> cabal01(normal) 1 of 3 [0, 0, 0]
cd "cabal/cabal01/cabal01.run" && $MAKE -s --no-print-directory cabal01 VANILLA=--enable-library-vanilla PROF=--disable-library-profiling DYN=--enable-shared CLEANUP=1  
Wrong exit code for cabal01()(expected 0 , actual 2 )
Stdout ( cabal01 ):
install1:
bin
lib
install2:
bin
lib
Stderr ( cabal01 ):
fromFlag NoFlag. Use fromFlagOrDefault
CallStack (from HasCallStack):
  error, called at libraries/Cabal/Cabal/Distribution/Simple/Flag.hs:96:21 in Cabal-3.0.0.0:Distribution.Simple.Flag
  fromFlag, called at libraries/Cabal/Cabal/Distribution/Simple/SrcDist.hs:126:20 in Cabal-3.0.0.0:Distribution.Simple.SrcDist
  sdist, called at libraries/Cabal/Cabal/Distribution/Simple.hs:412:5 in Cabal-3.0.0.0:Distribution.Simple
  sdistAction, called at libraries/Cabal/Cabal/Distribution/Simple.hs:188:50 in Cabal-3.0.0.0:Distribution.Simple
  defaultMainHelper, called at libraries/Cabal/Cabal/Distribution/Simple.hs:116:27 in Cabal-3.0.0.0:Distribution.Simple
  defaultMain, called at Setup.lhs:2:10 in main:Main
make[2]: *** [Makefile:35: cabal01] Error 1
*** unexpected failure for cabal01(normal)
```

bgamari:
For the record the `Setup configure` command line is 
```
[...]

PR #6073, opened 2019-06-06
Simple.SrcDist: Fix search for DistPref

bgamari:
Previously we would look at directly at the sDistDistPref flag. However,
this may not be set, resulting in #6068. We should rather be using
findDistPrefOrDefault to find the DistPref directory.

---


23Skidoo:
LGTM

PR #6078, opened 2019-06-14
Unpack in existing directory if empty

### LINKED ISSUE #6053: cabal get should work if target directory if empty
`cabal get` should work like `git clone`. If the directory doesn't exist then it is created (like now) but it should also work if the target directory is empty and unpack the package into it.

fendor:
---

Tested locally, I'd like to try to add tests for get, though.
closes #6053

### DIFF
--- a/cabal-install/Distribution/Client/Get.hs
+++ b/cabal-install/Distribution/Client/Get.hs
@@ -24,7 +24,8 @@ module Distribution.Client.Get (
 
 import Prelude ()
 import Distribution.Client.Compat.Prelude hiding (get)
-
+import Distribution.Compat.Directory
+         ( listDirectory )
 import Distribution.Package
          ( PackageId, packageId, packageName )
 import Distribution.Simple.Setup
@@ -56,8 +57,7 @@ import Control.Monad
          ( mapM, forM_, mapM_, unless )
 import qualified Data.Map as Map
 import System.Directory
-         ( createDirectoryIfMissing, doesDirectoryExist, doesFileExist
-         , listDirectory )
+         ( createDirectoryIfMissing, doesDirectoryExist, doesFileExist )
 import System.Exit
          ( ExitCode(..) )
 import System.FilePath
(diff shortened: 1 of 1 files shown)


cabal-install/Distribution/Client/Get.hs
```diff
@@ -162,12 +163,16 @@ unpackPackage :: Verbosity -> FilePath -> PackageId
               -> PackageDescriptionOverride
               -> FilePath  -> IO ()
 unpackPackage verbosity prefix pkgid descOverride pkgPath = do
-    let pkgdirname = display pkgid
-        pkgdir     = prefix </> pkgdirname
-        pkgdir'    = addTrailingPathSeparator pkgdir
+    let pkgdirname               = display pkgid
+        pkgdir                   = prefix </> pkgdirname
+        pkgdir'                  = addTrailingPathSeparator pkgdir
+        emptyDirectory directory = null <$> listDirectory directory
```
phadej: `listDirectory` is since `1.2.5.0`, tet we have `listDirectory` in `Distribution.Compat.Directory`.

cabal-install/Distribution/Client/Get.hs
```diff
@@ -162,12 +163,16 @@ unpackPackage :: Verbosity -> FilePath -> PackageId
...
     existsDir  <- doesDirectoryExist pkgdir
-    when existsDir $ die' verbosity $
-     "The directory \"" ++ pkgdir' ++ "\" already exists, not unpacking."
+    when existsDir $ do
+      isEmpty <- emptyDirectory pkgdir
+      unless isEmpty $
+        die' verbosity $
+        "The directory \"" ++ pkgdir' ++ "\" already exists, not unpacking."
```
phadej: should we change an error message to "... already exists and is not empty, not unpacking."

cabal-install/Distribution/Client/Get.hs
phadej: Hmm. this is `cabal-install` which is GHC-7.10+; yet `directory >=1.2.2.0` only.

I think we need to bump lower-bounds of boot libraries. @23Skidoo opinions?

fendor: Ok, so, I will use `listDirectory` from there.

fendor: For now, switch it to use compat module.

fendor: I changed the import to use `listDirectory` from the `Distribution.Compat.Directory` module.

jneira: Ddddddsssddddddddssddddddddddddrsdraaddasdasdaddddddsdsrssddsdrsd

jneira:
780

fendor:
Failing test case seems to be timeout related.

23Skidoo:
Merged, thanks!

Issue #6082, opened 2019-06-17
Prelude.foldr1: empty list on pkgconfig-depends in sub-library using cabal 3


ekmett:
If I uncomment the `pkgconfig-depends` line at [fontconfig.cabal](https://github.com/ekmett/codex/blob/178b48960bdf67d59a37beea227905aeb352dcb0/fontconfig/fontconfig.cabal#L67) then I get the following rather helpful error message from cabal:

```
$ cabal new-build
Prelude.foldr1: empty list
```

I don't believe I have done anything wrong here. The sub-library worked just fine as a standalone package. At worst, I think the dependency list may be redundant with the upstream packages, which both incur the same constraints, and therefore hopefully would ensure I get linked against the appropriate libs, but this isn't the right way to say that if that is what is being indicated. =)

hvr:
@phadej does this ring any bells related to your commit at 01df85eda2a1e986b47c9ca83cdc511b49fb62f2 ?

phadej:
Argh, yes. I should used foldr or sepBy1

Can we have NonEmpty in Cabal?

hvr:
@phadej we could bite the bullet at some point (maybe Cabal 3.2 ?), and add a conditional dependency such as

```
if !impl(ghc >= 8.0)
   build-depends: semigroups >= 0.16.1 && < 0.19
```

to Cabal -- assuming there's no `build-type:custom` in semigroups' trans-deps... :-)

and also get rid of our fake `Distribution.Compat.Semigroup` in the process... :-)

PR #6084, opened 2019-06-17
Replace fake `Semigroup`/`NonEmpty` with proper ones from `semigroups`

hvr:
This is a proof of concept for depending on `semigroups` instead of defining a fake local `class Semigroup` as suggested in https://github.com/haskell/cabal/issues/6082#issuecomment-502598220

/cc @phadej

---


Cabal/Cabal.cabal
```diff
@@ -285,7 +285,7 @@ library
   else
     -- at least one of lib:Cabal's dependency (`parsec`)
     -- already depends on `fail` transitively
```
23Skidoo: The comment also needs to be updated.

PR #6088, opened 2019-06-18
Fix #6082: Use pseudo-NonEmpty

### LINKED ISSUE #6082: Prelude.foldr1: empty list on pkgconfig-depends in sub-library using cabal 3

phadej:
- [x] cabal-install

### DIFF
--- a/Cabal/Distribution/Compat/Parsing.hs
+++ b/Cabal/Distribution/Compat/Parsing.hs
@@ -26,7 +26,7 @@ module Distribution.Compat.Parsing
   , many     -- from Control.Applicative
   , sepBy
   , sepBy1
-  -- , sepByNonEmpty
+  , sepByNonEmpty
   , sepEndBy1
   -- , sepEndByNonEmpty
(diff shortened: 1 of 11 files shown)


hvr:
...so what about #6084? :-)

phadej:
refactoring to use proper `NonEmpty` after #6084 is something *beginner friendly*

Issue #6014, opened 2019-04-25
cabal v2-install does not work with conditionally buildable executable


andrewthad:
Here's an example of how to observe this problem:

```
$ git clone https://github.com/andrewthad/ping
$ cd ping
$ cabal v2-install -f+application
Resolving dependencies...
Up to date
Warning: You asked to install executables, but there are no executables in
targets: . Perhaps you want to use --lib to install libraries instead.
```

But there is an executable. It happens to be hidden behind a flag like this:

```
executable multiping
  if flag(application)
    buildable: True
  else
    buildable: False
  main-is: Multiping.hs
```

I don't want users to incur all the extra dependencies of the executable, since it's just a CLI application exposing the library's functions. With `cabal v2-build -f+application`, the application does successfully get built. So I suspect that `cabal v2-install` may be forgetting to apply flags.

phadej:
Try

```
cabal v2-install -f+application .
```

(Note the dot).

If that helps, than this is a duplicate of https://github.com/haskell/cabal/issues/5978

PR #6092, opened 2019-06-20
Add default for v2-install

### LINKED ISSUE #5978: local v2-install requires to specify executable name(s)
```
% cabal v2-install
Resolving dependencies...
Up to date
Warning: You asked to install executables, but there are no executables in
targets: . Perhaps you want to use --lib to install libraries instead.
```

but

```
% cabal v2-install hackage-cli
Wrote tarball sdist to
/home/ogre/Documents/other-haskell/hackage-cli/dist-newstyle/sdist/hackage-cli-0.1.0.0.tar.gz
...
```

This is already annoying with `Agda` where you have to say `cabal install agda agda-mode`

typedrat:
Fixes #5978, #6014.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -247,6 +247,7 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags
   let
     installLibs = fromFlagOrDefault False (cinstInstallLibs clientInstallFlags)
     targetFilter = if installLibs then Just LibKind else Just ExeKind
+    targetStrings' = if null targetStrings then ["."] else targetStrings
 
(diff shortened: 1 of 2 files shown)


typedrat:
I'll merge this if there's no objection.

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -273,7 +274,7 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags
         then return (packageSpecifiers, packageTargets, projectConfig localBaseCtx)
```
23Skidoo: > `if null targetStrings'`

Shouldn't this be changed to `if null targetStrings''`?

PR #6095, opened 2019-06-20
Update issue template

phadej:
Make a new bug report template.

I noticed that we need to ask whether it's `v1-` or `v2-` build. Maybe a template will help remove single roundtrip.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
+```
+$ cabal v2-build ...
+```
+
+```
+```
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
...
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
+```
+$ cabal v2-build ...
+```
+
+Please use versions prefixed commands (e.g. `v2-build` or `v1-build`) to avoid ambiguity.
```
23Skidoo:
```suggestion
Please use version-prefixed commands (e.g. `v2-build` or `v1-build`) to avoid ambiguity.
```

23Skidoo:
I cancelled the Travis build (you forgot `[ci skip]`).

hvr:
Let's say a non-expert user submits a bug report, and they only know they've invoked `cabal build` or `cabal install` and they never heard about `v1-*` or `v2-*`.

So I wonder if wouldn't be easier to require the exact commands invoked **together** with the output of `cabal --version`.

phadej:
Bikeshedding can be done in `master`.

PR #6103, opened 2019-06-22
Add support for `v2-install`-ing foreign libraries.

### LINKED ISSUE #6046: Add support for foreign-library to v2-install
`cabal v2-install .` does install foreign libraries (`foreign-library`) into the store but doesn't symlink/copy them into `~/.cabal/lib` directory.

[3.3.2.8.Foreign libraries, Cabal User's Guide](https://www.haskell.org/cabal/users-guide/developing-packages.html#foreign-libraries) says

> ... we install foreign libraries in `~/.cabal/lib`, much like we install executables in `~/.cabal/bin`.

but it only applies to `v1-install`.

I'd like to `v2-install` foreign libraries, not just `v2-build`, because the former embeds proper rpaths that refer to `~/.cabal/store/`, whereas the latter leaves references to `dist-newstyle/`.

Currently I'm doing something like:

```
find ~/.cabal/store -name 'libmyforeignlib.so' -exec ls -t {} + |
head -n1 |
xargs cp -t /path/to/lib/
```

but it'd like to be able to do something like this instead:

```
cabal v2-install --installdir-lib=/path/to/lib/ .
```

typedrat:
Closes #6046.

Still needs documentation updates, plus shakedown testing from anyone who isn't me.

With more than two options of "thing that can be installed", existing interface doesn't make sense. Instead of exclusive modes, we now have a flag for each kind of thing that can be installed, allowing the user to select what they want explicitly.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -53,7 +53,7 @@ import Distribution.Simple.Program.Db
          ( userSpecifyPaths, userSpecifyArgss, defaultProgramDb
          , modifyProgramSearchPath, ProgramDb )
 import Distribution.Simple.BuildPaths
-         ( exeExtension )
+         ( dllExtension, exeExtension )
 import Distribution.Simple.Program.Find
          ( ProgramSearchPathEntry(..) )
 import Distribution.Client.Config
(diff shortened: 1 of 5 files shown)


typedrat:
I don't understand why there's the test failure... Where is that extra space coming from?

jneira:
It woould be nice if someonr could continue this (it needs a rebase, docs ans tests but is a relative small pr)

ulysses4ever:
@Mergify rebase

Kleidukos:

ffaf1:

PR #6127, opened 2019-07-03
Make `Setup.hs copy/install` work when data-files uses **.

### LINKED ISSUE #6125: Cabal install fails when ** wildcards are used
See https://github.com/NixOS/nixpkgs/issues/64173#issuecomment-507997029

quasicomputational:
Treating globs like filenames was always illegitimate, but this code
was broken further by the addition of recursive globs.

I had a look around for other dubious code along these lines, and it
looks like this site is the only problematic one.

Fixes #6125.

---

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -34,6 +34,9 @@
   * Uniformly provide 'Semigroup' instances for `base < 4.9` via `semigroups` package
   * Setting `debug-info` now implies `library-stripping: False` and
     `executable-stripping: False) ([#2702](https://github.com/haskell/cabal/issues/2702))
+  * `Setup.hs copy` and `install` now work in the presence of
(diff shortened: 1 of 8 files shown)


quasicomputational:
@23Skidoo: this isn't too late for 3.0, right? It's a pretty bad bug, since it's possible for the package author to never run these code paths before publishing and breaking users.

23Skidoo:
Hopefully we can squeeze this in `3.0`.

quasicomputational:
OK - the failure is a time-out on one of the OSX builders, which are known to be flaky. Since this comes with a test and the fix is easy to eyeball-verify, I'll go ahead and merge to master. I'll leave the 3.0 merging up to you?

23Skidoo:
Cherry-picked to 3.0.

PR #6128, opened 2019-07-03
Properly expand globs when looking for headers in extra-src-files.

quasicomputational:
This ought to be the right fix, but I'm not actually sure what the original code is achieving here. When would headers go in `extra-source-files` rather than `includes`?

---

### DIFF
--- a/cabal-install/Distribution/Client/SourceFiles.hs
+++ b/cabal-install/Distribution/Client/SourceFiles.hs
@@ -17,6 +17,7 @@ import Distribution.Client.RebuildMonad
 
 import Distribution.Solver.Types.OptionalStanza
 
+import Distribution.Simple.Glob
 import Distribution.Simple.PreProcess
 
 import Distribution.Types.PackageDescription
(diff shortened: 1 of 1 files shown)


phadej:
The documentation says

> A list of header files to be included in any compilations via C. This field applies to both header files that are already installed on the system and to those coming with the package to be installed. The former files should be found in absolute paths, while the latter files should be found in paths relative to the top of the source tree or relative to one of the directories listed in include-dirs.

But I'm really unsure what that means. And if it should or should not affect `{-# LANGUAGE CPP #-}`

quasicomputational:
Looking again, this PR is a strict improvement - the previous code was illegitimate - so I guess we can put off figuring out what it actually *should* be doing; I'll go ahead and merge now.

phadej:
> Looking again, this PR is a strict improvement - the previous code was illegitimate - so I guess we can put off figuring out what it actually should be doing; I'll go ahead and merge now.

When "we" will figure it out? Did you add a comment in code, and issue for that (if anything else to alert people who might now something about).

---

And merging when CI is failing is :-1:, **never**. You could had a fix in this PR already. Be careful.

phadej:
I'm enabling more required jobs for CI, i guess that's the only way to be sure there are no ninja-merges

phadej:
The commit in this PR is reverted

Issue #4743, opened 2017-09-05
Using a local hackage mirror with HTTP Auth stopped working in 2.0


zeha:
In previous versions (verified with cabal-install 1.22.6.0 library 1.22.5.0, but pretty sure also with some 1.24.x), this worked:

```
% grep hackage ~/.cabal/config 
repository hackage.haskell.org
  url: http://build:xxxxxxxxxxxxxxxx@nexus.namespace.at/nexus/repository/hackage-all/
```

With cabal-install 2.0.0.0 library 2.0.0.2, on OS X, the various cabal commands now fail with:

```
Warning: No mirrors found for
http://build:...@nexus.namespace.at/nexus/repository/hackage-all/
Unexpected response 401for http://build:...@nexus.namespace.at/nexus/repository/hackage-all/root.json
```

A quick tcpdump reveals that `build:...` is not just for display purposes, but actually gets sent in the HTTP request!

```
	GET /nexus/repository/hackage-all/root.json HTTP/1.1
	Host: nexus.namespace.at
	Authorization: Basic YnVpbGQ6Li4u
	User-Agent: cabal-install/2.0.0.0 (osx; x86_64)
	Accept: */*
	Cache-Control: no-transform
```

Note that `YnVpbGQ6Li4u` decodes to `build:...`.

zeha:
Forcing the `plain-http` transport using `cabal --http-transport=plain-http update` works; with the default `curl` transport it fails as described above.

phadej:
To clarify, `build:...` is the actual contents of `YnVpbGQ6Li4u`, but we should encode the password, i.e. `xxxxxxxxxxxxxxxx`. Shouldn't be hard to fix.

zeha:
Yes. Sorry that wasn't clear enough :-)

hvr:
- temporary workaround for this one: `secure: False` in custom `repository` section
- related: #2761

gbaz:
resolved by above.

PR #6130, opened 2019-07-05
Allow curl to use .netrc

dsturnbull:
If using a private hackage repo in conjunction with the main one, and the private one requires HTTP Basic Auth, then you are forced to use the wget transport. The curl transport doesn't understand http auth credentials embedded within the repo uri.

e.g.
```
repository private
  url: https://user:pass@hackage.private.net/
```

`curlTransport` doesn't understand this, but `wgetTransport` does.

Unfortunately, `wgetTransport` doesn't do range requests so updating from hackage is really slow.

This PR simply adds a `curlnetrc` transport which runs `curl -n` instead of just `curl`. If a user has a private repo configured, and if it requires auth, and if they have provided credentials in `~/.netrc`, then this will allow using the curl transport to grab updates.

If the user doesn't have a `~/.netrc`, then behaviour is unchanged.

== Testing

Added a private repository entry in `~/.cabal/config` without embedded credentials:

```
repository private
  url: https://hackage.private.net/

http-transport: curlnetrc
```

And specified the credentials in `~/.netrc` instead:
```
machine hackage.private.net
login uuu
password xxx
```

```
$ dist-newstyle/build/x86_64-osx/ghc-8.6.5/cabal-install-3.0.0.0/build/cabal/cabal v2-update
Resolving dependencies...
Up to date
Downloading the latest package lists from:
- hackage.haskell.org
- private
To revert to previous state run:
    cabal v2-update 'private,2019-07-04T05:42:06Z'
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2019-07-04T23:22:51Z'
```

---

### DIFF
--- a/cabal-install/Distribution/Client/HttpUtils.hs
+++ b/cabal-install/Distribution/Client/HttpUtils.hs
@@ -59,6 +59,7 @@ import System.IO.Error
 import Distribution.Simple.Program
          ( Program, simpleProgram, ConfiguredProgram, programPath
          , ProgramInvocation(..), programInvocation
+         , programOverrideArgs, programPostConf
          , ProgramSearchPathEntry(..)
          , getProgramInvocationOutput )
 import Distribution.Simple.Program.Db
@@ -252,6 +253,12 @@ supportedTransports :: [(String, Maybe Program, Bool,
 supportedTransports =
     [ let prog = simpleProgram "curl" in
       ( "curl", Just prog, True
+      , \db -> curlTransport <$> lookupProgram prog db )
+
+    , let prog = (simpleProgram "curl") { programPostConf = post }
+          args = ["-n"]
+          post = \_ p -> return p { programOverrideArgs = args } in
+      ( "curlnetrc", Just prog, True
       , \db -> curlTransport <$> lookupProgram prog db )
 
     , let prog = simpleProgram "wget" in
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -405,7 +405,7 @@ globalCommand commands = CommandUI {
          trueArg
 
       ,option [] ["http-transport"]
-         "Set a transport for http(s) requests. Accepts 'curl', 'wget', 'powershell', and 'plain-http'. (default: 'curl')"
+         "Set a transport for http(s) requests. Accepts 'curl', 'curlnetrc', 'wget', 'powershell', and 'plain-http'. (default: 'curl')"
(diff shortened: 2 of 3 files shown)


hvr:
Is there *any* downside to passing `-n` unconditionally to `curl`?

I'm slightly worried this would lead to cabal automatically authenticating on `cabal upload` even though I don't expect it to. I.e. I do have  a `.netrc` in place which happens to contain credentials for various Hackage servers in place, but I use it for *other* tooling, and I don't want cabal to pick it up to avoid accidentally uploading something to Hackage. So I wonder if we shouldn't have a more generic approach to this, by e.g. having a `http-transport-curl-options: ...` feature in place; this could even be per-repository in case you need/want to have different curl flags for your various repositories

23Skidoo:
I'd prefer if we just had per-repo username/password fields:

```
remote-repo private.hackage
    username: u
    password: pass
    ...
```

But I guess this can be a useful workaround in the meantime.

dsturnbull:
@hvr the downside is exactly the scenario you portray but it's worth mentioning that `wget` already defaults to using `.netrc`. If you had used `wget` because you wanted credentials to work in the first place, you would already be suffering from this (arguably) implicit magic.

`http-transport-curl-options` seems reasonable though.

@23Skidoo we already kind of do. You can specify e.g. `https://u:p@private.hackage.com`. It's just that the curl transport encodes your password as `...` for security purposes, but then actually submits that to the server. https://github.com/haskell/cabal/issues/4743

I thought to just bypass that issue by using the tools' built-in password stuff.

23Skidoo:
> `https://u:p@private.hackage.com`

This is not ideal because of the security issue you mention (username/password shows up in `ps` output). It'd be nicer if there was a uniform way to specify this info in `~/.cabal/config` that worked for all transports (including new ones we might add in the future).

`--http-transport-*-options` sounds like a good idea too, we should be able to implement via the existing `ProgramDb` abstraction.

dsturnbull:
> `--http-transport-*-options` sounds like a good idea too, we should be able to implement via the existing `ProgramDb` abstraction.

I'm struggling with it. 

I've tried adding a new field to `Distribution.Client.Config.remoteRepoFields` but that config is not passed in to `Distribution.Client.HttpUtils.configureTransport` from numerous places, e.g. `Distribution.Client.ProjectConfig` has forgotten anything about the `RemoteRepo`. Same path goes for adding a new field to `GlobalFlags`.

I don't know how to get any config into the `ProgramDb`.

dsturnbull:
ping @hvr / @23Skidoo -- @AlexeyRaga and I have bypassed the compatibility issue by introducing a `curlnetrc` transport.

PR #6133, opened 2019-07-05
Literate run

phadej:
An example file 

```latex
%cabal:
%build-depends: base, lens
%endcabal
\documentclass{article}
%include polycode.fmt
\begin{document}

This is an example document

\begin{code}
import Control.Lens

main :: IO ()
main = putStrLn $ toListOf traverse "hello world"
\end{code}

\end{document}
```

Comments on "syntax" are welcome. Having distinct start and end markers helps parsing. `%endcabal` is distinctive enough end marker, but should start marker be `%cabal:` or `%cabal` (without colon), or some completely different markers?

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdRun.hs
+++ b/cabal-install/Distribution/Client/CmdRun.hs
@@ -104,7 +104,7 @@ import qualified Text.Parsec as P
 import System.Directory
          ( getTemporaryDirectory, removeDirectoryRecursive, doesFileExist )
 import System.FilePath
-         ( (</>), isValid, isPathSeparator )
+         ( (</>), isValid, isPathSeparator, takeExtension )
 
 
 runCommand :: CommandUI (ConfigFlags, ConfigExFlags, InstallFlags, HaddockFlags, TestFlags)
@@ -171,8 +171,10 @@ runAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags)
     let
       scriptOrError script err = do
         exists <- doesFileExist script
+        let pol | takeExtension script == ".lhs" = LiterateHaskell
+                | otherwise                      = PlainHaskell
         if exists
-          then BS.readFile script >>= handleScriptCase verbosity baseCtx tempDir
+          then BS.readFile script >>= handleScriptCase verbosity pol baseCtx tempDir
           else reportTargetSelectorProblems verbosity err
 
     (baseCtx', targetSelectors) <-
@@ -339,9 +341,9 @@ parseScriptBlock str =
 readScriptBlock :: Verbosity -> BS.ByteString -> IO Executable
 readScriptBlock verbosity = parseString parseScriptBlock verbosity "script block"
 
-readScriptBlockFromScript :: Verbosity -> BS.ByteString -> IO (Executable, BS.ByteString)
-readScriptBlockFromScript verbosity str = do
-    str' <- case extractScriptBlock str of
+readScriptBlockFromScript :: Verbosity -> PlainOrLiterate -> BS.ByteString -> IO (Executable, BS.ByteString)
+readScriptBlockFromScript verbosity pol str = do
+    str' <- case extractScriptBlock pol str of
               Left e -> die' verbosity $ "Failed extracting script block: " ++ e
               Right x -> return x
     when (BS.all isSpace str') $ warn verbosity "Empty script block"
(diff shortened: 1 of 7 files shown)


23Skidoo:
@typedrat, can you please review this?

hvr:
Syntax seems reasonable;  however, what's the motivation? is this merely syntax sugar so one doesn't need to wrap the syntax within `\begin{code}`?

```latex
\begin{code}
{- cabal:
...
-}
\end{code}

**Comment by 23Skidoo** [comment]
Re: syntax, can't we just parse the special comment block until the first line that doesn't start with `%` and omit the end marker?

**Comment by phadej** [comment]
@hvr Yes, more TeX friendly syntax.

```latex
\begin{code}
{- cabal:
...
-}
\end{code}
```

needs to be in `\begin{document}`, so probably somewhat far in the document.

OTOH, it would work with any literate processors, e.g. `markdown-unlit`.

---

@23Skidoo 

```latex
%This document can be run directly with
%
% $ cabal v2-run script.lhs
%
% Following are needed buildinfo definition
%
%cabal:
%build-depends: base, lens
%
% some random comments
% to describe this document
\documentclass{article}
%include polycode.fmt
\begin{document}
```

case would be ambiguous. IMO having and `endmarker` is clearer. Something like

```
%This document can be run directly with
%
% $ cabal v2-run script.lhs
%
% Following are needed buildinfo definition
%
%cabal:
%  build-depends: base, lens
%
% some random comments
% to describe this document
\documentclass{article}
%include polycode.fmt
\begin{document}
```

would work, i.e. look up to the next non-empty line and its indentation, takeWhile indentation is at least as large. But that's complicated.

---

I'l split the PR to only add "if input file is `.lhs`, then use `main-is: Main.lhs`". I.e. make @hvr approach work already. That's small, yet good enough to get me forward.

23Skidoo:
With `.hs` scripts the following works:

```
{- cabal:
   unknown-field: unknown
   build-depends: base
   This is a comment.
   This is another comment.
-}
```

Doesn't work if you put the comment first, so I guess it stops after the first parse error. `unknown-field` line just results a warning. So I think that for consistency reasons we should either add end marker support to `.hs` scripts too, or drop it from `.lhs` ones.

phadej:
@23Skidoo the end marker for `.hs` files is `-}`

```
   This is a comment.
   This is another comment.
```

is actually "valid" cabal lexical form.

This fails:

```haskell
{- cabal:
build-depends: base, lens

garbage {
-}

import Control.Lens

main = print "foo"
```

with

```
Warning: script block:3:9: "the input" (line 3, column 9):
unexpected end of input
expecting field or section name or "}"
cabal: Failed parsing "script block".
```

23Skidoo:
So your example

```
%cabal:
%build-depends: base, lens
%
% some random comments
% to describe this document
\documentclass{article}
```

will work fine if you treat the first non-comment line as an end marker?

phadej:
@23Skidoo empty line is arbitrary, for "reasons", people may add empty lines there:

```
%caba
%
% build-depends:
%   , base
%   , lens
%   , more-deps
%
% ghc-options: -Wall
% build-tool-depends: ...
%
```

and rules like "first block of non-empty lines" is IMHO limiting.

23Skidoo:
```
%cabal:
%
% build-depends:
%   , base
%   , lens
%   , more-deps
%
% ghc-options: -Wall
% build-tool-depends: ...
%
```

will be treated as a single `cabal:` block under my proposal, but:

```
%cabal:
%
% build-depends:
%   , base
%   , lens
%   , more-deps

% ghc-options: -Wall
% build-tool-depends: ...
%
```

will not. Basically, a continuous `%`-comment block will be treated the same way as a `{- ... -}` block.

phadej:
@23Skidoo Yes, the consecutive % is an option, but I think it's just too fragile. (as something latter might be accidentally glued to in). I.e. *separator is very subtle: single-character whitespace*

23Skidoo:
What will you do when there's no `endcabal` marker? Syntax error?

phadej:
@23Skidoo yes, as we do now:

```haskell
    goBody _ [] = Left $ "`" ++ endMarker ++ "` end marker not found"
```

23Skidoo:
OK, it's not too bad. The error message should tell you how the end marker looks like.

I guess we could also use
```
\iffalse
cabal:
build-depends: base
...
\fi
```
but it feels too hacky.

phadej:
Hmm, I don't think `\iffalse` is actually that hacky...

phadej:
I'm fine with `\iffalse` approach. No new syntax is good.

Issue #6107, opened 2019-06-23
As `Cabal` now depends on `semigroups` remove partial functions


phadej:
e.g. `groupBy` + `head` can be made safe with `NonEmpty` stuff, and so one.

As most modules use `Distribution.Compat.Prelude`, hiding partial `Prelude` stuff should be quite straight forward.

Bonus: clean up `(a, [a])` stuff in `Distribution.Compat.Parsing`.

dwijnand:
@phadej Anything left?  I could have a look if there is.

phadej:
Thanks for helping, I think we are good now

PR #6142, opened 2019-07-12
Use Data.List.NonEmpty and avoid some partial functions

edmundnoble:
Add safe alternatives to some partial functions in Distribution.Utils.Generic, and removes the originals from Distribution.Compat.Prelude. See #6107.

Uncomment versions of functions in Distribution.Compat.Parsing that return non-empty lists.

There are still many partial functions in use like `maximum` after this PR, and there are a few uses of `head`, `last`, `tail` and `init` that still remain because I didn't find any easy way to remove them.

---

- I ran `cabal new-run cabal-tests`, `cabal new-test Cabal:unit-tests`, etc.

### DIFF
--- a/Cabal/Distribution/Backpack/ComponentsGraph.hs
+++ b/Cabal/Distribution/Backpack/ComponentsGraph.hs
@@ -8,7 +8,7 @@ module Distribution.Backpack.ComponentsGraph (
     componentCycleMsg
 ) where
 
-import Prelude ()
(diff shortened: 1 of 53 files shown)


23Skidoo:
There are some `-Werror` failures on older versions of GHC.

dwijnand:
Can be closed now, this work was recovered and landed in #6320.

PR #6134, opened 2019-07-05
cabal run works with Literate Haskell file

phadej:
A lite variant, without additional syntax. There's codepath so we could add it, if we comeup with something good. Until then `\iffalse` (by @23Skidoo ) idiom is IMHO good enough.

```latex
\iffalse
{- cabal:
build-depends: base >= 4.3 && <5
-}
\fi
\documentclass{article}
\begin{document}
\begin{code}

{-# LANGUAGE BangPatterns #-}
{-# LANGUAGE ScopedTypeVariables #-}

import Prelude

main :: IO ()
main = putStrLn "Hello World"
\end{code}
\end{document}
```

EDIT: there's no new syntax, so I hope this is an uncontroversial change.

Issue #5977, opened 2019-04-01
cabal v2-haddock exits with 0 even when it fails to build the docs


phadej:
With `3.0+git20190317.1.b3d6d0b-6~18.04`

```
% cabal v2-haddock && echo $?
...
Warning: Failed to build documentation for github-0.21.
0
```

phadej:
Even it says a *warning* that failed to build a documentation. I need a way to make it fail, even the default is lenient. That is needed for automated scripts, like CI.

phadej:
ping @RyanGlScott, atm `haskell-ci`s haddock building doesn't fail the builds.

phadej:
I assigned this to @hvr, as he did tweak something related. Again, this is high priority issue, as there are no way to check in CI that haddocks build.

isovector:
@phadej suggests https://github.com/haskell/cabal/pull/5459 might be to blame. I'm going to investigate.

isovector:
For a quick CI workaround:

```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
- - cabal new-haddock
+ - "! cabal new-haddock 2>&1 >/dev/null | grep 'Failed to build documentation'"
```

phadej:
Fixed with https://github.com/haskell/cabal/pull/6137, thanks @isovector!

PR #6148, opened 2019-07-20
Add changelog for #5977 amd #6134 fixes [ci skip]

phadej:


phadej:
Also I think, these should be backported to `cabal-install-3.0`, but I'm not sure what are the current plans with `3.0` branch? ping @hvr @23Skidoo

23Skidoo:
Cherry-picked to `3.0`.

PR #6151, opened 2019-07-21
Clearly state that invalid fields are ignored

bgamari:
Previously we would issue a warning of the following form when
we encountered a field that isn't supported by the cabal-version
declared in the cabal file:

    Warning: ghc-heap.cabal:30:3: The field "cmm-sources" is available
    only since the Cabal specification version 3.0.

The message, however, doesn't tell the user the consequence of this
warning. Consequently a user could easily be confused when the build
subsequently fail.

Avoid this possibility by clearly stating that the field value will be
ignored.

---


bgamari:
This arose out of the discussion on #5819.

phadej:
One need to `--accept` `parser-tests`

phadej:
Merged in https://github.com/haskell/cabal/pull/6173

Issue #5420, opened 2018-07-08
Uses of 'extra-prog-path' from cabal config need to format it for the specific platform


randen:
The cabal config uses "," (comma) and the platform-specific newline (at least on Windows, it is CR/LF) to format multi-line entries.  For example, if cabal-install is used to --augment the user's config file with extra-prog-path entries, this is the result (note: the "\n" below would be the actual newline 0x0D):

````
extra-prog-path: F:\Program Files\Haskell Platform\8.4.3\msys\usr\bin,\n
                 F:\Users\name\AppData\Roaming\cabal\bin
````
Uses of this entry by the Cabal code (specifically, during a 'cabal build') is taken without sufficient conversion.  When used in command line arguments or environment variables (e.g., %PATH%), a conversion to remove any linefeeds (0x0A) and change newline (0x0D) into the two characters "\n" seems to be done, and the comma(s) remains.

However, for use in a Windows PATH variable, the commas need to be converted to semi-colons, and the newlines should be stripped completely rather than being converted to the two characters '\\' and 'n'.

This issue results in the extra programs not being found on these extra paths due to these syntactic errors, failing the use case of being able to build configure-style projects which require a unix-like system such as MSys.

yaxu:
Thanks for clarifying this @randen. As the maintainer of the tidalcycles library I probably help someone with this issue every day.

yaxu:
Oh, I thought that this was recent, but the probem was identified 8th July 2018. I'd love to see this fixed.

PR #6155, opened 2019-07-24
write multiline fields as multiline fields in config

gbaz:
---

This changes `fieldGet` of `FieldDescr` to have `a -> [Doc]` and almost everything to not care about the extra list functionality. The one exception is fields that are instantiated with `multiListField`, a new drop-in replacement for `newLineListField`. These fields are able to be monoidally combined regardless afaik. However, by being broken out across multiple entries, now we don't have to worry about different path conventions (a la #5420) for different systems.

Everything still passes the test suite.

### DIFF
--- a/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
@@ -23,7 +23,7 @@ module Distribution.Client.ProjectConfig.Legacy (
 import Prelude ()
 import Distribution.Client.Compat.Prelude
 
-import Distribution.Deprecated.ParseUtils (parseFlagAssignment)
+import Distribution.Deprecated.ParseUtils (parseFlagAssignment, runP)
 
 import Distribution.Client.ProjectConfig.Types
 import Distribution.Client.Types
@@ -914,7 +914,7 @@ legacySharedConfigFieldDescrs =
           showTokenQ parseTokenQ
           (fromNubList . globalLocalRepos)
           (\v conf -> conf { globalLocalRepos = toNubList v }),
-         newLineListField "extra-prog-path-shared-only"
+         multiListField "extra-prog-path-shared-only"
           showTokenQ parseTokenQ
           (fromNubList . globalProgPathExtra)
           (\v conf -> conf { globalProgPathExtra = toNubList v })
@@ -1000,23 +1000,23 @@ legacyPackageConfigFieldDescrs =
       legacyConfigureFlags
       (\flags conf -> conf { legacyConfigureFlags = flags })
   . addFields
-      [ newLineListField "extra-include-dirs"
+      [ multiListField "extra-include-dirs"
           showTokenQ parseTokenQ
           configExtraIncludeDirs
           (\v conf -> conf { configExtraIncludeDirs = v })
-      , newLineListField "extra-lib-dirs"
+      , multiListField "extra-lib-dirs"
           showTokenQ parseTokenQ
           configExtraLibDirs
           (\v conf -> conf { configExtraLibDirs = v })
-      , newLineListField "extra-framework-dirs"
+      , multiListField "extra-framework-dirs"
           showTokenQ parseTokenQ
           configExtraFrameworkDirs
           (\v conf -> conf { configExtraFrameworkDirs = v })
-      , newLineListField "extra-prog-path"
+      , multiListField "extra-prog-path"
           showTokenQ parseTokenQ
           (fromNubList . configProgramPathExtra)
           (\v conf -> conf { configProgramPathExtra = toNubList v })
-      , newLineListField "configure-options"
+      , multiListField "configure-options"
           showTokenQ parseTokenQ
           configConfigureArgs
           (\v conf -> conf { configConfigureArgs = v })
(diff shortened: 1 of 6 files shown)


gbaz:
I do wonder, now having written all this, if we couldn't instead change the _parser_ to parse newLineListField into a list properly instead? Hrmm..

phadej:
One just should rewrite parsers using `Distribution.Parsec` framework... than this kind of tweaking won't be necessary...

cabal-install/Distribution/Deprecated/ParseUtils.hs
```diff
@@ -177,7 +177,7 @@ warning s = ParseOk [PWarning s] ()
 data FieldDescr a
   = FieldDescr
       { fieldName     :: String
-      , fieldGet      :: a -> Doc
+      , fieldGet      :: a -> [Doc]
```
phadej: the `Distrubution.Fields.Pretty` cheats. It renders `Doc` and sees if it's a multiline, and then indents

https://github.com/haskell/cabal/blob/0d2d9ca7b7a1689dba9c3a22a8d9966b05b2957f/Cabal/Distribution/Fields/Pretty.hs#L107-L111

I'd use here the same approach, i.e. keep `fieldGet` as `a -> Doc`. It won't make migration *harder*.

cabal-install/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -1402,6 +1402,16 @@ newLineListField :: String -> (a -> Doc) -> ReadP [a] a
                  -> (b -> [a]) -> ([a] -> b -> b) -> FieldDescr b
 newLineListField = listFieldWithSep Disp.sep
 
+multiListField :: String -> (a -> Doc) -> ReadP [a] a
```
phadej: newLineListField has a comment, `

> like
-- commaNewLineListField it pretty prints on multiple lines

ins't it working? What does this PR changes? Examples, tests missing.

phadej:
> I do wonder, now having written all this, if we couldn't instead change the parser to parse newLineListField into a list properly instead? Hrmm..

It does parse into a list?

gbaz:
Let me explain the desired change, and we can see if there's a better way:

before: 

```
extra-prog-path: F:\Program Files\Haskell Platform\8.4.3\msys\usr\bin,\n
                 F:\Users\name\AppData\Roaming\cabal\bin

```
after:

```
extra-prog-path: F:\Program Files\Haskell Platform\8.4.3\msys\usr\bin
extra-prog-path:  F:\Users\name\AppData\Roaming\cabal\bin

```

The former does _not_ get parsed into a list, as per #5420, but instead just goes into a single thing that augments the path, thus going splat on windows.

phadej:
> after:

```
extra-prog-path: F:\Program Files\Haskell Platform\8.4.3\msys\usr\bin
extra-prog-path:  F:\Users\name\AppData\Roaming\cabal\bin
```

I'm against this change.

monoidal fields **MUST** have single line output. If the comma breaks things, than require no comma for that field.

gbaz:
cabal accepts the latter -- you believe it should not? that would break a lot of existing files...

gbaz:
that said, i can look for another way to fix this issue that doesn't introduce this format -- i'm digging in more to understand what the genuine underlying cause is.

phadej:
I'm not sure what are the semantics of `extra-prog-path` field.

If it's a *monoidal* fields, i.e. can be specified multiple times, and it's arguments are appended; 
then it should be parsed as such. And if Windows / Unix require different ways of concatenating those (`:: [ ...]`) fields, that's not a problem of parsing&pretty-printing code.

---

I can believe that field is parsed as is, i.e. not split on comma, but this all is old code, so I wont spend any more time on this.

gbaz:
> I can believe that field is parsed as is, i.e. not split on comma, but this all is old and deprecated code, so I wont spend any more time on this.

This code is currently used by cabal on the standard path, and it causes problems for windows users uniformly. So just saying "its deprecated" isn't an answer. I understand that you don't want to dig in more at the moment, but I will continue looking. This _needs_ to be fixed.

gbaz:
Just verified that the parser doesn't work. So I can change the output, or I can change the parser.

Parsing a conf file with three lines with commas gives: `"baz,\nfoo,\nbar"`.

gbaz:
Ok so the issue is the following. All that nice parsing stuff in `Distribution.Client.ProjectConfig.Legacy` for handling commas works fine. However this is _only_ on the project parsing path. On the config file parsing path, we just use the `parseFields` from here

https://github.com/haskell/cabal/blob/ed3ae1332ca7ef4d1d09b6da6494a44f3e40f5a8/cabal-install/Distribution/Client/ParseUtils.hs#L139

So, absent/until rewriting absolutely everything about config file parsing, we have two options: A) improve the parser in some special-casing way, despite it being legacy.
Or B) emit two lines, which is what this patch does.

Thoughts?

gbaz:
So this reveals the more general issue, which is that in cabal config files, currently most (all?) fields which are monoidal on input (i.e. able to merge multiple lines) _cannot_ be given as single line lists on input. The code for handling this... just isn't there at all.

gbaz:
@23Skidoo What do you think the right approach is here? The one here (emit multiple entries) or hack up the legacy `parseFields` parser to special case roughly four fields?

gbaz:
Submitted #6161 which takes the alternative approach. This should please @phadej since it maintains the "monoidal fields get written as a single line" invariant and just does a spot-patch to config file parsing to make it work how people apparently already thought it should.

gbaz:
Closing in favor of #6161

PR #6164, opened 2019-07-27
Fix cabal-install fighting over index cache with old versions

DanielG:
Currently 2.4.1.0 and master/3.0 fight over the global index cache. Every time one switches between the versions you get:

```
Warning: Parsing the index cache failed (Unknown encoding for constructor).
Trying to regenerate the index cache...
[takes 30sec or so]
```

and it takes just long enough to be really annoying.

Here's my proposed solution: just rename the index cache file.

---

### DIFF
--- a/cabal-install/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/Distribution/Client/IndexUtils.hs
@@ -549,8 +549,8 @@ indexFile (RepoIndex _ctxt repo) = indexBaseName repo <.> "tar"
 indexFile (SandboxIndex index)   = index
 
 cacheFile :: Index -> FilePath
-cacheFile (RepoIndex _ctxt repo) = indexBaseName repo <.> "cache"
-cacheFile (SandboxIndex index)   = index `replaceExtension` "cache"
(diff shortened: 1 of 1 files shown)


phadej:
Use the cabal major version than adhoc v2, cache format changes every major release anyway?

DanielG:
Does it though? I'd rather do this only when it is necessary and not litter everything with new cache files on each release.

phadej:
my point is: the versioning should be automatic, if it's manual, one will forget to update it.

If version of `cabal-install` (and/or `Cabal`?) is too conservative, then a structure hash of index structure (like in https://hackage.haskell.org/package/binary-tagged) could be used.

(FWIW, `binary-tagged` for caches in `dist-newstyle` would make sense too. `cabal` gets confused when e.g. `Extension` is added)

DanielG:
Totally agree this should be automatic. I didn't realise this was fesible though but since there's an existing solution to that problem with Binary instances I'll see if I can spin a PR to integrate this.

DanielG:
I think this was fixed for good by #6255

amigalemming:
cabal-3.2 and cabal-2.4 still fight over the index cache.
What can I do as a `cabal-install` user?

Issue #4402, opened 2017-03-20
Specify remote repository target with cabal upload


ezyang:
If you have multiple remote repositories, it might be convenient to call out specifically which one you want to upload to. Right now, we always take the "last" repository listed.

This is not urgent because you can always upload the tarball by hand.

ezyang:
Something that makes this a little more urgent is that there is no web UI for uploading Haddock documentation.

ezyang:
Also, a simpler but still useful thing to fix, would be for upload to tell you WHERE you're uploading the tarball to.

mitchellwrosen:
@ezyang Would you mind explaining how to upload a tarball by hand? Or point me at some relevant documentation? Thanks!

nomeata:
I am also bit by this. In fact, I didn’t even realize that `cabal upload` broke because I have a repository configured in `~/.cabal/config`.

This is actually a security issue: If I tell someone to add my repository to their config (which I have…), I could get the password the server there.

So when asking for the password, `cabal` really should tell the user for which server it wants the password!

gbaz:
I believe cabal enforces digest auth, which means that the password is not
recoverable from such an attack. We should improve the situation anyway.
Perhaps even a setting for default upload repo distinct from the main
package repo setting?

peti:
This feature is very much needed, IMHO. `cabal update` takes the name of the repository as an argument, too, and so should `upload`.

ezyang:
@mitchellwrosen yes

kindaro:
I want to work on this. Is there maintainer consensus on the design? If I make a patch, will it be merged?

1. I think a command line option `repository` should be added to the `upload` command.
1. I think it is a security issue that a package gets silently uploaded _somewhere_ when there are multiple repositories configured. When there are multiple repositories to choose from, and the command line does not say which to pick, Cabal should error out.

Is this agreeable?

ulysses4ever:
@kindaro I don't know of a consensus on design but there's some activity around this isue I believe here:

- https://github.com/haskell/cabal/pull/11478

kindaro:
Oh, how timely! Hopefully it gets merged.

PR #6172, opened 2019-08-02
Upload to specific repo

dsturnbull:
Closes https://github.com/haskell/cabal/issues/4402

You can now specify `cabal upload --repository-name [repo name]` to tell cabal to upload to a specific hackage repo.

```
cabal upload dist-newstyle/sdist/dummy-lib-0.2.0.0.tar.gz --repository-name private
Uploading dist-newstyle/sdist/dummy-lib-0.2.0.0.tar.gz...
Package successfully uploaded as candidate. You can now preview the result at
'https://private/package/dummy-lib-0.2.0.0/candidate'. To
publish the candidate, use 'cabal upload --publish'.
```

---

### DIFF
--- a/cabal-install/Distribution/Client/BuildReports/Upload.hs
+++ b/cabal-install/Distribution/Client/BuildReports/Upload.hs
@@ -33,7 +33,7 @@ import Distribution.Client.Setup
 type BuildReportId = URI
 type BuildLog = String
 
-uploadReports :: Verbosity -> RepoContext -> (String, String) -> URI -> [(BuildReport, Maybe BuildLog)] -> IO ()
+uploadReports :: Verbosity -> RepoContext -> Maybe (String, String) -> URI -> [(BuildReport, Maybe BuildLog)] -> IO ()
 uploadReports verbosity repoCtxt auth uri reports = do
   forM_ reports $ \(report, mbBuildLog) -> do
      buildId <- postBuildReport verbosity repoCtxt auth uri report
@@ -41,11 +41,11 @@ uploadReports verbosity repoCtxt auth uri reports = do
        Just buildLog -> putBuildLog verbosity repoCtxt auth buildId buildLog
        Nothing       -> return ()
 
-postBuildReport :: Verbosity -> RepoContext -> (String, String) -> URI -> BuildReport -> IO BuildReportId
+postBuildReport :: Verbosity -> RepoContext -> Maybe (String, String) -> URI -> BuildReport -> IO BuildReportId
 postBuildReport verbosity repoCtxt auth uri buildReport = do
   let fullURI = uri { uriPath = "/package" </> display (BuildReport.package buildReport) </> "reports" }
   transport <- repoContextGetTransport repoCtxt
-  res <- postHttp transport verbosity fullURI (BuildReport.show buildReport) (Just auth)
+  res <- postHttp transport verbosity fullURI (BuildReport.show buildReport) auth
   case res of
     (303, redir) -> return $ undefined redir --TODO parse redir
     _ -> die' verbosity "unrecognized response" -- give response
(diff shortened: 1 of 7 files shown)


cabal-install/Distribution/Client/HttpUtils.hs
```diff
@@ -252,7 +259,13 @@ supportedTransports :: [(String, Maybe Program, Bool,
 supportedTransports =
     [ let prog = simpleProgram "curl" in
       ( "curl", Just prog, True
-      , \db -> curlTransport <$> lookupProgram prog db )
+      , \db -> curlTransport Prompt <$> lookupProgram prog db )
+
+    , let prog = (simpleProgram "curl") { programPostConf = post }
+          args = ["-n"]
```
phadej: use long `--netrc` argument, it's more self-documenting.

phadej (dismissed):
What happens if you use `curlnetrc` transport, more specifically what `curl` does when
- there aren't `.netrc` file
- the file doesn't contain entry for the server

---

I don't think this is right approach.

I think we need
- a way to pass additional options to transport programs
- turn off unneeded prompt

Note: that `wget` seems to use `netrc` by default

```
       --no-netrc
           Do not try to obtain credentials from .netrc file. By default .netrc file is searched for credentials in case none have been passed on command line and authentication is required.
```

AlexeyRaga:
@phadej In this case `curl` won't be able to authenticate and you'll see an authentication error. So no harm is done.

The first attempt/PR was to use `netrc` by default (as you pointed, `wget` is already doing that), but this was rejected because it would interfere with someone's workflows: 

people DO:
- use `.netrc` (have `hackage` credentials there)
- use `curl` transport
- but rely on `curl` NOT using `.netrc`

> a way to pass additional options to transport programs

I agree, that would be ideal. We may try to make it happen (and I believe @dsturnbull looked at it before), but it feels like much more involved change.
Also it looks like different transports would have to have different options (`curl` and `wget` have different defaults, `plain-http` probably doesn't support `.netrc` at all, etc).

So we thought that having `curlnetrc` is an option that has less risk of impacting peoples' workflows, but also easy to add.

But I will have a look at this closer.

dsturnbull:
> What happens if you use `curlnetrc` transport, more specifically what `curl` does when
> 
>     * there aren't `.netrc` file
> 
>     * the file doesn't contain entry for the server

cabal will fail and tell you there's a 401 Unauthorised error.

> 
> I don't think this is right approach.
> 
> I think we need
> 
>     * a way to pass additional options to transport programs
> 
>     * turn off unneeded prompt

I would like this, but getting options to transport programs via config is inordinately complicated.

> 
> 
> Note: that `wget` seems to use `netrc` by default
> 
> ```
>        --no-netrc
>            Do not try to obtain credentials from .netrc file. By default .netrc file is searched for credentials in case none have been passed on command line and authentication is required.
> ```

We are using wget now because it is able to authenticate using embedded credentials within the repo url. But wget is unable to do partial downloads, so it's much slower.

dsturnbull:
I have remove the curl bits here. Now it just allows choosing a repo to upload to.

dsturnbull:
@phadej all the transport options are removed from here, what do you think?

gbaz:
If this is rebased to master, I would support merging. Sorry it got lost in the backlog!

Kleidukos:
@dsturnbull Would you mind fixing the conflicts? :)

Mikolaj:
@dsturnbull: a hopeful ping? ;)

Kleidukos:
@dsturnbull Would you like us to take on this PR? It's okay if you have low bandwidth these days. :)

dsturnbull:
@Kleidukos please do! Sorry, didn’t see the thread comments here.

Kleidukos:

PR #6188, opened 2019-08-10
Remove dead code

sgraf812:
This would trip up the improved pattern match checker from [!963](https://gitlab.haskell.org/ghc/ghc/merge_requests/963).

### DIFF
--- a/Cabal/Distribution/PackageDescription/PrettyPrint.hs
+++ b/Cabal/Distribution/PackageDescription/PrettyPrint.hs
@@ -126,17 +126,10 @@ ppCondTree2 v grammar = go
         thenDoc = go thenTree
 
     ppIf (CondBranch c thenTree (Just elseTree)) =
-          case (False, False) of
- --       case (isEmpty thenDoc, isEmpty elseDoc) of
-              (True,  True)  -> mempty
-              (False, True)  -> [ ppIfCondition c thenDoc ]
-              (True,  False) -> [ ppIfCondition (cNot c) elseDoc ]
-              (False, False) -> [ ppIfCondition c thenDoc
-                                , PrettySection () "else" [] elseDoc
-                                ]
-      where
-        thenDoc = go thenTree
-        elseDoc = go elseTree
+      -- See #6193
+      [ ppIfCondition c (go thenTree)
+      , PrettySection () "else" [] (go elseTree)
+      ]
 
 ppCondLibrary :: CabalSpecVersion -> Maybe (CondTree ConfVar [Dependency] Library) -> [PrettyField ()]
 ppCondLibrary _ Nothing = mempty


phadej:
I'm on the fence. The code there is in the state to have roundtrip property for code like

```cabal
if flag(some-condition)
  buildable: True
else
  buildable: False
```

which is ugly when pretty-printed as:

```cabal
if flag(some-condition)
else
  buildable: False
```

That's because `buildable: True` is no-op, and there could be other a like.

One way, is to pretty-print that as

```cabal
if !flag(some-condition)
  buildable: False
```

But that violates roundtrip property.

---

If the code is removed, then there should be an issue to think how to resolve (or not) the above.

sgraf812:
Well, I agree that you probably don't want to print empty `if` branches (IIUC), but currently it's just dead code, no? What I did was purely a refactoring. The line that actually did something has been commented out for 2 years. Unless there is some ugly templating mechanism involved in the build process that replaces the `case (True, True) of` by the line below, I don't think keeping the other branches makes sense at all.
If on the other hand that dead code makes actual sense, then we should probably think about the scrutinised expression...

phadej:
That dead code reminds me of an unsolved issue, but when it's gone there won't be anything reminding of that ugliness...

sgraf812:
Soo... When is that issue going to be resolved? Why does it mean we can't currently dead code? I want my patch to hit master within the next month, which is quite impossible when the code here stays this way. Of course, we could always attach a `{-# OPTIONS_GHC -woverlapping-patterns #-}` as a short term solution...

DanielG:
How about replacing the dead code with a comment and an issue reference which everyone remember rather than storing obscure information only in @phadej's brain?

sgraf812:
Is there any news on this, @phadej? Should I add a comment referencing this MR, expressing your plans to get back to it at some point? The deleted code is still in version control, so it's not like we could never recover it... Quite the contrary, leaving dead code seems like a code smell to me.

sgraf812:
I opened #6193 to track the issue here.

sgraf812:
I can't manually restart the failing jobs, short of force pushing with a different commit message. I don't know why the same two jobs keep running into the timeout.

sgraf812:
Could anyone merge this please? This is still holding back https://gitlab.haskell.org/ghc/ghc/merge_requests/963 to land on master... Apparently CI turned green thanks to some kind stranger restarting the failing jobs for me - thank you!

DanielG:
I would love to but you still haven't addressed the comments I gave you on IRC i.e. adding a comment with a link to the issue.

sgraf812:
Ah, I thought you meant I should link from the commit message, which I did. Fair enough.

sgraf812:
Thank you! :) What is the story for updating the GHC's Cabal submodule? Should I just submit a bump along with my MR? I suspect there will be some breakage...

PR #6195, opened 2019-08-17
Fix v2-install ProgramDb confusion

DanielG:
Commit 273dacf ("Extract installLibraries and InstallExes from
installAction") makes 'getInstalledPackages' get passed the
pre-`configCompilerEx` ProgramDb which will cause pattern match failures when
the code tries to lookup ghc-pkg in the progdb.

As far as I can tell on master currently any `v2-install` invocation would fail with 

```
Distribution/Simple/GHC.hs:1967:5-56: Irrefutable pattern failed for pattern Just ghcPkgProg
```

but I only tested with `cabal v2-install --lib Cabal-3.0.0.0`.

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -468,7 +468,8 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags
     hcPath   = flagToMaybe projectConfigHcPath
     hcPkg    = flagToMaybe projectConfigHcPkg
 
-    progDb =
+    -- ProgramDb with directly user specified paths
+    preProgDb =
         userSpecifyPaths (Map.toList (getMapLast packageConfigProgramPaths))
       . userSpecifyArgss (Map.toList (getMapMappend packageConfigProgramArgs))
       . modifyProgramSearchPath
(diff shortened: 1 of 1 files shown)


DanielG:
FYI this also needs a backport to the 3.0 branch if we're intending to cut a cabal-install release from that eventually.

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -477,7 +477,7 @@ installAction (configFlags, configExFlags, installFlags, haddockFlags, testFlags
       $ defaultProgramDb
 
   (compiler@Compiler { compilerId =
-    compilerId@(CompilerId compilerFlavor compilerVersion) }, platform, progDb') <-
+    compilerId@(CompilerId compilerFlavor compilerVersion) }, platform, progDb) <-
```
phadej: If this shadows, then it will cause `-Wall` warning, won't it?

```
Distribution/Client/CmdInstall.hs:480:73: warning: [-Wname-shadowing]
    This binding for ‘progDb’ shadows the existing binding
      bound at Distribution/Client/CmdInstall.hs:471:5
<no location info>: error: 
Failing due to -Werror.
```

I'd rename original `progDb` to e.g. `preProgDb` and adding some comments how they differ

```
-- ProgramDb with directly user specified paths
preProgDb = ...

-- progDb is a program database with compiler tools configured properly
(compiler@ ... progDb) <- ...
```

DanielG: Yeah, I wasn't sure if we have -Wall and -Werror on by default even.

Honestly I vehemently disagree with the `-Wname-shadowing` warning. In fact in my opinion because this warning discourages shadowing names in do blocks like it does was _the_ reason for this bug to even get introduced in the first place. People are just used to imperative style of doing things and especially in a do block that should just be something Haskell lets you do. I digress -- should write a GHC proposal about that instead of ranting ;)

23Skidoo:
Cherry-picked to `3.0`.

PR #6207, opened 2019-08-23
Honour suffix and prefix on install

### LINKED ISSUE #6179: v2-install ignores  --program-suffix
**Describe the bug**

```
cabal v2-install foo -w ghc-8.4.4 --program-suffix=-ghc-8.4.4
```

will install `foo` not `foo-ghc-8.4.4` as expected :disappointed: 

(replace `foo` with anything valid, e.g. for sake of argument `stylish-haskell`)

```
cabal-install version 2.4.1.0
compiled using version 2.4.1.0 of the Cabal library 
```

fendor:
Previously, the flags --program-suffix and --program-prefix
were ignored when given to v2-install.
To fix this, both installation cases, symlink and copy need to be
aware of the new installation name.
Closes #6179 

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -107,7 +107,7 @@ import Distribution.System
 import Distribution.Types.UnitId
          ( UnitId )
 import Distribution.Types.UnqualComponentName
-         ( UnqualComponentName, unUnqualComponentName, mkUnqualComponentName )
+         ( UnqualComponentName, unUnqualComponentName )
 import Distribution.Verbosity
          ( Verbosity, normal, lessVerbose )
 import Distribution.Simple.Utils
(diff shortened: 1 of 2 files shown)


hvr:
just to be clear; the semantics are such that *only* the installed name of the executable (i.e. the symlink or copy) gets modified by --program-suffix and --program-prefix but *not* the name within the nix-style .cabal/store, right?

fendor:
Yes, we assume:
`cabal v2-install exe:foo --installdir=/home/user/.cabal/bin --installmethod=symlink --program-suffix=-8.8.1 --program-prefix=bar`
results in a symlink
`barfoo-8.8.1 -> /home/user/.cabal/store/.../bin/foo`
and copying renames the executable only in the ~/.cabal/bin folder.
Everything else would be an unintentional change.

fendor:
Since I forgot to mention it:
I tested this by manually executing in a project `cabal install exe:foo --installdir=/home/user/.cabal/bin --installmethod=symlink --program-suffx=-8.6.4 --program-prefix=bar`
and checked that the executable linked to ~/.cabal/bin/ is named "barfoo-8.6.4" and links into the store to some executable "~/.cabal/store/.../bin/foo".
I repeated the same for --installmethod=copy.

fendor:
So, is anything blocking this from getting merged?

phadej:
Thanks, sorry for long wait

PR #6216, opened 2019-08-31
Recognise flags that change error message format as simple

sergv:
I.e. don’t rebuild everything when a ghc flag that affects how error messages are presented is added.

Tested with `cabal test` and by hand on a project where adding `-fprint-potential-instances` to `cabal.project` caused all dependencies to be rebuilt.

---

### DIFF
--- a/Cabal/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/Distribution/Simple/Program/GHC.hs
@@ -211,7 +211,7 @@ normaliseGhcArgs (Just ghcVersion) PackageDescription{..} ghcArgs
 
     simpleFlags :: Set String
     simpleFlags = Set.fromList . mconcat $
-      [ [ "-n", "-#include", "-Rghc-timing", "-dsuppress-all", "-dstg-stats"
+      [ [ "-n", "-#include", "-Rghc-timing", "-dstg-stats"
         , "-dth-dec-file", "-dsource-stats", "-dverbose-core2core"
         , "-dverbose-stg2stg", "-dcore-lint", "-dstg-lint", "-dcmm-lint"
         , "-dasm-lint", "-dannot-lint", "-dshow-passes", "-dfaststring-stats"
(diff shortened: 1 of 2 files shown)


phadej:
@sergv I assume you didn't test with GHC-8.8, as it *shouldn't work* with a patch I see. (See #6218 )

phadej:
... also I think this should be backported to `3.0`

sergv:
Yes, I didn't test with `8.8`, I just glimpsed the bit in the manual. I'll remove `print-axiom-incomps` then.

Cabal/Distribution/Simple/Program/GHC.hs
```diff
@@ -172,11 +172,15 @@ normaliseGhcArgs (Just ghcVersion) PackageDescription{..} ghcArgs
               , "break-on-exception", "print-bind-result"
               , "print-bind-contents", "print-evld-with-show"
               , "implicit-import-qualified", "error-spans"
+              , "print-expanded-synonyms", "print-potential-instances"
```
merijn: It seems these should be under some specific GHC version? I went through all the flags in the GHC source at the time this code was written and since I didn't see them then I assume these are new in ghc 8.6/8.8? Other than that this seems fine.

sergv: I did this just now and they're already in 8.0

```
sergey@debian:~$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.0.2
sergey@debian:~$ ghc --show-options | awk '/-fprint/'
-fprint-explicit-foralls
-fprint-explicit-kinds
-fprint-explicit-coercions
-fprint-explicit-runtime-reps
-fprint-equality-relations
-fprint-unicode-syntax
-fprint-expanded-synonyms
-fprint-potential-instances
-fprint-typechecker-elaboration
```

Which means that my placement of `print-unicode-syntax` is not entirely correct and `-fprint-explicit-runtime-reps` seems to be missing. Perhaps the better way is to treat all flags starting with `-fprint` as simple and safe to ignore?

phadej:
superseded by https://github.com/haskell/cabal/pull/6422, where I guard the flags by version.

sergv:
Thanks a lot for picking this up and actually going through with it.

PR #6225, opened 2019-09-07
Document hs-source-dirs

toonn:
I replaced the note about `hs-source-dir` being deprecated in the
description of `hs-source-dirs` by a field of its own to include the
`deprecated` and `removed` fields because it's no longer in 3.0.

Source directories can be shared among components but there's a
non-obvious caveat which I've tried to clarify with a note for `hs-source-dirs`.
This definitely needs review because I'm not sure I understood it and whether I'm getting the point across.

---

### DIFF
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -2184,14 +2184,20 @@ system-dependent values for these fields.
        :pkg-field:`other-modules`, :pkg-field:`library:exposed-modules` or
        :pkg-field:`executable:main-is` fields.
 
+.. pkg-field:: hs-source-dir: directory list
+    :deprecated: 2.0
+    :removed: 3.0
+    :default: ``.``
+
+    Root directories for the module hierarchy.
+
+    Deprecated in favor of :pkg-field:`hs-source-dirs`.
+
 .. pkg-field:: hs-source-dirs: directory list
 
     :default: ``.``
 
     Root directories for the module hierarchy.
-
-    For backwards compatibility, the old variant ``hs-source-dir`` is
-    also recognized.
 
 .. pkg-field:: default-extensions: identifier list
 
--- a/Cabal/doc/developing-packages.rst
+++ b/Cabal/doc/developing-packages.rst
@@ -2199,6 +2199,13 @@ system-dependent values for these fields.
 
     Root directories for the module hierarchy.
 
+    .. note::
+
+      Components can share source directories but modules will be compiled if
+      they can be found there, i.e., if a library and an executable share a
+      source directory and the executable depends on the library's ``Foo``
+      module it will be recompiled even if the library has already been built.
+
 .. pkg-field:: default-extensions: identifier list
 
     A list of Haskell extensions used by every module. These determine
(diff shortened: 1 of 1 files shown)


Cabal/doc/developing-packages.rst
```diff
@@ -2184,14 +2184,27 @@ system-dependent values for these fields.
...
 
     :default: ``.``
 
     Root directories for the module hierarchy.
 
-    For backwards compatibility, the old variant ``hs-source-dir`` is
-    also recognized.
+    .. note::
+
+      Components can share source directories but modules will be compiled if
```
DanielG: "will be **re**compiled"?

Cabal/doc/developing-packages.rst
```diff
@@ -2184,14 +2184,27 @@ system-dependent values for these fields.
...
+      they can be found there, i.e., if a library and an executable share a
+      source directory and the executable depends on the library's ``Foo``
+      module it will be recompiled even if the library has already been built.
```
DanielG: You could add a note that internal libraries are a way to solve that problem.

Cabal/doc/developing-packages.rst
toonn: I did intend "compiled" here. Should I add "even if other components have already built these same modules?" Or would recompiled be clear/accurate enough?

Cabal/doc/developing-packages.rst
toonn: Isn't the real solution not to share a tree with dependencies? Or would internal libraries's modules in the src dir not be recompiled?

DanielG: I'm not sure what you mean by "sharing a tree with dependencies"?

What you would do is add the internal library as a build-depends to the components where you want to share modules so then they only get compiled once, namely when the internal lib is built but you can still use them all over.

Cabal/doc/developing-packages.rst
DanielG: It just sounds kind of weird because what else would I expect cabal to do if not compile these modules ;)

I think the combination of both, "modules will be recompiled even if other components already built them" makes even more sense.

toonn: Will rewrite this weekend.

Cabal/doc/developing-packages.rst
toonn: The tree I'm talking about is a file hierarchy. No matter what you do if a module you import is in your source dir you'll compile it again. Even if a dependency, internal lib or no, already built it.

toonn: @DanielG, I'm not clear on the problem you suppose would be solved here using internal libraries? I can imagine people might share a source dir if they wanted multiple parts depending on something without exposing it but the reason for sharing the dirs isn't really relevant to the note. The note's just supposed to point out a gotcha people might run into. Another way you could end up in this situation is simply desiring a simpler repository setup, with a single src directory for all your code.

DanielG: I think I missed the fact that you were talking specifically about (public) libraries already, so nevermind.

I was talking about that other common problem where you want your test suite to have access to internals without exporting them in the public library but that doesn't really fit here, you're right.

DanielG:
Thanks!

Issue #1906, opened 2014-05-28
improve c2hs support


abbradar:
Continuing https://github.com/haskell/cabal/issues/55 -- I'm stuck in this bug now (using {#import's in C2HS, which fails without dependency resolution).

vmchale:
@abbradar I wrote [chs-cabal](http://hackage.haskell.org/package/chs-cabal) which polishes it up a bit - it's kind of hacky on the Cabal side, but it works.

poscat0x04:
@vmchale Is it possible to use this package with stack?

Issue #55, opened 2012-05-24
improve c2hs support


bos:
(Imported from [Trac #48](http://hackage.haskell.org/trac/hackage/ticket/48), reported by @dcoutts on 2006-01-16)

To properly support c2hs in Cabal we need a couple enhancements:
1. dependency resolution.
   Not only do the .hs files depend on the .chs files but .chs files can depend on other .chs files (or strictly speaking on their corresponding .chi files).

This is because c2hs supports its own module import mechanism to get information about bindings defined in one module to be used in other modules.

From the c2hs [manual](http://www.cse.unsw.edu.au/~chak/haskell/c2hs/docu/c2hs-1.html):

<pre class="wiki">If the compiled binding module contains import hooks, C-&gt;Haskell needs to
find the .chi (C-&gt;Haskell interface files) produced while compiling the
corresponding binding modules. By default, they will be searched for in
the current working directory. If they are located elsewhere, the
--include=INCLUDE option has to be used to indicate the location, where
INCLUDE is a colon-separated list of directories. Multiple such options
are admissible. Paths specified later are searched first.
</pre>The Gtk2Hs project has a [script](http://darcs.haskell.org/gtk2hs/mk/chsDepend.in) to find the dependencies of a .chs file. This should be translated into Haskell and included in Cabal's future pluggable dependency resolution mechanism(!).
1. The .chi files need to be installed

If one Cabal package uses .chs files then another package that wants to import that module into another .chs module using c2hs's import mechanism will need access to the .chi file of the imported module. This is basically the same as how ghc needs the .hi file of imported modules.

One difference is that .chi modules are transitive (unlike ghc's .hi files) so only the .chi files corresponding to exposed modules need to be installed (where as for ghc every .hi file needs to be installed).

Both of these enhancements are essential to be able to package [Gtk2Hs](http://haskell.org/gtk2hs/) using Cabal. Gtk2Hs has many .chs modules and with dependencies between them, including many dependencies between .chs modules in different packages. For example all of the Gtk+ extension packages (OpenGL, SourceView, Glade, Mozilla embeding widget) import type definitions from .chs modules in the base gtk &amp; glib packages.

PR #6233, opened 2019-09-11
Fix c2hs handling in Cabal

vmchale:
This fixes a 13 year-old bug, viz. https://github.com/haskell/cabal/issues/55 aka https://github.com/haskell/cabal/issues/1906

I tested it by building my own [library](https://github.com/vmchale/libarchive).

### DIFF
--- a/Cabal/Distribution/C2Hs.hs
+++ b/Cabal/Distribution/C2Hs.hs
@@ -1,0 +1,74 @@
+module Distribution.C2Hs ( fixDeps ) where
+
+import qualified Data.Map                        as M
+import qualified Data.Set                        as S
+import           Distribution.C2Hs.Lexer
+import           Distribution.ModuleName         (ModuleName, components)
+import           Distribution.PackageDescription (PackageDescription)
+import qualified Distribution.PackageDescription as PD
+import           Distribution.Simple.Utils       (dieNoVerbosity,
+                                                  findFileWithExtension,
+                                                  withUTF8FileContents)
+import           Distribution.Text               (simpleParse)
+import           System.FilePath                 (joinPath)
+
+fixDeps :: PackageDescription -> IO PackageDescription
+fixDeps pd@PD.PackageDescription {
+          PD.library = Just lib@PD.Library {
+            PD.exposedModules = expMods,
+            PD.libBuildInfo = bi@PD.BuildInfo {
+              PD.hsSourceDirs = srcDirs,
+              PD.otherModules = othMods
+            }}} = do
+
+  let findModule = findFileWithExtension [".chs"] srcDirs . joinPath . components
+
+  mExpFiles <- traverse findModule expMods
+  mOthFiles <- traverse findModule othMods
+
+  let modDeps = zipWith (ModDep True []) expMods mExpFiles ++
+                zipWith (ModDep False []) othMods mOthFiles
+  modDeps <- traverse extractDeps modDeps
+  let (othMods, expMods) = span (not . mdExposed) $ sortTopological modDeps
+  return pd { PD.library = Just lib {
+    PD.exposedModules = map mdOriginal (reverse expMods),
+    PD.libBuildInfo = bi { PD.otherModules = map mdOriginal (reverse othMods) }
+  }}
+fixDeps pd = pure pd
+
+data ModDep = ModDep {
+  mdExposed  :: Bool,
+  mdRequires :: [ModuleName],
+  mdOriginal :: ModuleName,
+  mdLocation :: Maybe FilePath
+}
+
+instance Show ModDep where
+  show x = show (mdLocation x)
+
+instance Eq ModDep where
+  ModDep { mdOriginal = m1 } == ModDep { mdOriginal = m2 } = m1==m2
+instance Ord ModDep where
+  compare ModDep { mdOriginal = m1 } ModDep { mdOriginal = m2 } = compare m1 m2
+
+extractDeps :: ModDep -> IO ModDep
+extractDeps md@ModDep { mdLocation = Nothing } = return md
+extractDeps md@ModDep { mdLocation = Just f } = withUTF8FileContents f $ \con -> do
+  mods <- case getImports con of
+        Right ms -> case traverse simpleParse ms of
+            Just ms -> pure ms
+            Nothing -> dieNoVerbosity ("Cannot parse module name in c2hs file " ++ f)
+        Left err -> dieNoVerbosity ("Cannot parse c2hs import in " ++ f ++ ": " ++ err)
+  return md { mdRequires = mods }
+
+sortTopological :: [ModDep] -> [ModDep]
+sortTopological ms = fst $ foldl visit (([]), S.empty) (mdOriginal <$> ms)
+  where
+  set = M.fromList (map (\m -> (mdOriginal m, m)) ms)
+  visit (out,visited) m
+    | m `S.member` visited = (out,visited)
+    | otherwise = case m `M.lookup` set of
+        Nothing -> (out, m `S.insert` visited)
+        Just md -> (md:out', visited')
+          where
+            (out',visited') = foldl visit (out, m `S.insert` visited) (mdRequires md)
--- a/Cabal/Distribution/C2Hs.hs
+++ b/Cabal/Distribution/C2Hs.hs
@@ -1,74 +1,53 @@
-module Distribution.C2Hs ( fixDeps ) where
+-- Based on https://github.com/gtk2hs/gtk2hs/blob/master/tools/src/Gtk2HsSetup.hs#L414
+module Distribution.C2Hs ( reorderC2Hs ) where
 
-import qualified Data.Map                        as M
-import qualified Data.Set                        as S
-import           Distribution.C2Hs.Lexer
-import           Distribution.ModuleName         (ModuleName, components)
-import           Distribution.PackageDescription (PackageDescription)
-import qualified Distribution.PackageDescription as PD
-import           Distribution.Simple.Utils       (dieNoVerbosity,
-                                                  findFileWithExtension,
-                                                  withUTF8FileContents)
-import           Distribution.Text               (simpleParse)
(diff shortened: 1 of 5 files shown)


cabal-install/bootstrap.sh
```diff
@@ -430,8 +432,9 @@ do_Cabal_pkg () {
     then
         if need_pkg "Cabal" ${CABAL_VER_REGEXP}
         then
+            ALEX_PATH="$(pwd)/$(find . -name alex -perm 755 | tail -n1)"
             echo "Cabal-${CABAL_VER} will be installed from the local Git clone."
-            (cd ../Cabal && install_pkg ${CABAL_VER} ${CABAL_VER_REGEXP})
+            (cd ../Cabal && ${ALEX_PATH} -g Distribution/C2Hs/Lexer.x && install_pkg ${CABAL_VER} ${CABAL_VER_REGEXP})
```
phadej: I'm on the fence with these. The lexer for `.cabal` files is generated in manual preprocessing step (see `Makefile`). And rather keep it that way, i.e. let's not have compile-time dependency on alex.

vmchale: Will do.

Makefile
```diff
@@ -3,6 +3,7 @@
 .PHONY : cabal-install-dev cabal-install-prod
 
 LEXER_HS:=Cabal/Distribution/Fields/Lexer.hs
+LEXER_CHS:=Cabal/Distribution/C2Hs/Lexer.hs
```
phadej: We still need to commit in `Lexer.hs` file too. So the people who do not modify `Lexer.x` shouldn't need to be aware of it.

vmchale: Should be done.

Cabal/Distribution/C2Hs.hs
```diff
@@ -0,0 +1,56 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+-- Based on https://github.com/gtk2hs/gtk2hs/blob/master/tools/src/Gtk2HsSetup.hs#L414
```
phadej: That's GPLd repository, we cannot "base" anything on that code.

boot/C2HsLexer.x
```diff
@@ -0,0 +1,90 @@
...
+module Distribution.C2Hs.Lexer ( getImports ) where
+
+import Control.Applicative ((<$>))
+}
+
+%wrapper "monad"
+
+$module = [A-Za-z\.]
+
+tokens :-
```
phadej: This file needs comments, what it does. Is there a reason why you chose `alex`, and not writing something using `parsec`?

Cabal/Distribution/C2Hs.hs
```diff
@@ -0,0 +1,56 @@
...
+import Distribution.Compat.Prelude
+import qualified Data.Map as M
+import qualified Data.Set as S
+import Distribution.C2Hs.Lexer
+import Distribution.ModuleName (ModuleName, components)
+import Distribution.Parsec (simpleParsec)
+import Distribution.Simple.Utils (dieNoVerbosity, findFileWithExtension, withUTF8FileContents)
+import System.FilePath (joinPath)
+
+reorderC2Hs :: [FilePath] -> [ModuleName] -> IO [ModuleName]
```
phadej: This need a proper comment, what it does, and why.

Cabal/Distribution/C2Hs.hs
```diff
@@ -0,0 +1,56 @@
...
+                     , moduleOriginal :: ModuleName
+                     }
+
+extractDeps :: (ModDep, Maybe FilePath) -> IO ModDep
+extractDeps (md, Nothing) = pure md
+extractDeps (md, (Just f)) = withUTF8FileContents f $ \con -> do
+  mods <- case getImports con of
+        Right ms -> case traverse simpleParsec ms of
+            Just ms' -> pure ms'
+            Nothing -> dieNoVerbosity ("Cannot parse module name in c2hs file " ++ f)
```
phadej: This dieing here feels wrong. There should be way to fail in pure way...

We need tests.

Cabal/Distribution/C2Hs.hs
```diff
@@ -0,0 +1,56 @@
...
+        Left err -> dieNoVerbosity ("Cannot parse c2hs import in " ++ f ++ ": " ++ err)
+  pure (md { moduleRequires = mods })
+
+sortTopological :: [ModDep] -> [ModDep]
```
phadej: Is there a way to `Distribution.Compat.Graph` to not reinvent the wheel?

Cabal/Distribution/Simple/PreProcess.hs
```diff
@@ -168,7 +169,8 @@ preprocessComponent pd comp lbi clbi isSrcDist verbosity handlers = do
   (CLib lib@Library{ libBuildInfo = bi }) -> do
     let dirs = hsSourceDirs bi ++ [autogenComponentModulesDir lbi clbi
                                   ,autogenPackageModulesDir lbi]
-    for_ (map ModuleName.toFilePath $ allLibModules lib clbi) $
+    mods <- reorderC2Hs dirs (allLibModules lib clbi)
```
phadej: Why only for librayr components, shouldn't this be done for all components?

Cabal/Distribution/C2Hs.hs
vmchale: Oh, good point.

boot/C2HsLexer.x
vmchale: I used `alex` because it seems to be the best choice for lexers and I've used it in the past for identifying dependencies in e.g. C.

Cabal/Distribution/C2Hs.hs
vmchale: Will do.

Cabal/Distribution/C2Hs.hs
vmchale: I have changed the `dieNoVerbosity` to a `warn`. 

I've added some tests to ensure it works as I need it to.

Cabal/Distribution/C2Hs.hs
vmchale: Yes; I've fixed this.

Cabal/Distribution/Simple/PreProcess.hs
vmchale: Yes, this should be fixed now.

vmchale:
Ready for further review.

vmchale:
@phadej is this ready to be merged?

boot/C2HsLexer.x
phadej: Well, I don't agree that we need to write any new `alex` code. It's write-only for almost any other contributor. You should been asked before making tech choice

Cabal/tests/C2Hs/src/Codec/Archive/Types/Foreign.chs
```diff
@@ -0,0 +1,119 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric  #-}
```
phadej: Is this file actually compiled, or is it only an input for tests?

Can we place it's somewhere where it doesn't look like real file.

I value that we have "real life" tests inputs, but we also should have something more minimal.

Also negative examples.

vmchale:
@hvr thoughts?

I'm quite keen on getting this merged (/fixed), and I'm open to rearranging most of it based on the experience of others (I am quite partial to the alex-based lexer, though).

vmchale:
(Also, if someone else wants to rewrite the lexer I would not be offended - but I would not know how to do this without alex or a similar tool/library for lexers)

DanielG:
@vmchale  I'm also interested in improving dependency tracking for cabal preprocessors, but in general, not just for c2hs. I don't really understand in detail what exactly c2hs needs here, maybe you cain explain what is going on?

Also is there a reason you can see that would prevent doing this more generally, by say adding an eplicit interface for any preprocessors to communicate dependencies to Cabal instead of Cabal having to parse the preprocessor's input which seems like a pretty nasty violation of the separation of concerns?

vmchale:
@DanielG That's a good point. It might be better to expose a general interface to allow preprocessors to reorder dependencies. 

As for what `c2hs` needs here: I think the easiest way to explain is an example: https://github.com/vmchale/libarchive/blob/master/src/Codec/Archive/Foreign/Archive.chs#L327

Briefly, `c2hs` can `{# import #}` another `.chs` module which gives that module some "context" for marshaling Haskell types. 

If you have any more questions about the specific `c2hs` case, feel free to ask. 

My 2¢ is that `Cabal` should include a special case for `c2hs`, since it tries to includes support for `.chs` files out of the box.

DanielG:
I'm not against shipping special support code for the builtin preprocessors, I just think it should be possible for external ones to work to the same extent. Possibly with some custom Setup.hs code but ideally I'd like to just have some protocol that preprocessors can choose to expose which Cabal can automatically use to give you improved dependency tracking.

vmchale:
Right, my guess is it's easier to require custom `Setup.hs` code - designing such a protocol is more complex.

phadej:
I have thinking about this, and I won't accept this approach. There are two options:

- Either `c2hs` is tought a variant where `Cabal` can pass it all the files to process
- Or there is some discovery option, so `Cabal` could ask `c2hs` (again passing all the files) in which order to process them.

cf. that with `ghc --make` or possible single-shot compilation with initial dependency discovery.

To say it directly: I won't accept `cabal` "learning" about the contents of source files. `c2hs` might be quite stable, but there's no reason to assume it won't change. Let's not couple `c2hs` and `Cabal`.

vmchale:
@phadej So you think it would be sensible to provide build tools a mechanism to pass dependency reorderings to cabal-install?

Should a spec be decided on on the cabal side before making PRs to c2hs and so on?:

Kleidukos:

ffaf1:

PR #6108, opened 2019-06-23
show-build-info (lib:Cabal part)

hvr:
This is the lib:Cabal part extracted from PR #5954 by @fendor and squashed into two commits.

---
Please include the following checklist in your PR:

* [ ] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ ] Any changes that could be relevant to users have been recorded in the changelog.
* [ ] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

Issue #6106, opened 2019-06-23
v2-build --only-dependencies solver should never fail, if v2-build --dry doesn't.


phadej:
```
servant release-0.16 % cabal new-build --dry all | tail
 - cookbook-generic-0.1 (exe:cookbook-using-custom-monad) (cannot read state cache)
 - cookbook-db-sqlite-simple-0.1 (exe:cookbook-db-sqlite-simple) (cannot read state cache)
 - cookbook-db-postgres-pool-0.1 (exe:cookbook-db-postgres-pool) (cannot read state cache)
 - cookbook-curl-mock-0.1 (exe:cookbock-curl-mock) (cannot read state cache)
 - cookbook-basic-streaming-2.1 (exe:cookbook-basic-streaming) (cannot read state cache)
 - cookbook-basic-auth-0.1 (exe:cookbook-basic-auth) (cannot read state cache)
 - tutorial-0.10 (test:spec) (cannot read state cache)
 - cookbook-testing-0.0.1 (exe:cookbook-testing) (cannot read state cache)
 - cookbook-pagination-2.1 (exe:cookbook-pagination) (cannot read state cache)
 - cookbook-file-upload-0.1 (exe:cookbook-file-upload) (cannot read state cache)
```

```
servant release-0.16 % cabal new-build --dep all       
cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the packages servant-0.16.1,
servant-foreign-0.15, servant-docs-0.11.3, servant-server-0.16.1 and
servant-client-0.16 are required by a dependency of one of the other targets.
```

Solver should pick as much as it can, there are dependencies to be built, even there are so called in-place packages (here: `tutorial` depends on `servant-js` which depends on `servant-js`)

---

More concretely, given there are four categories of packages in a plan

- **builtin** Lives in global (non-nix-style) package db
- **global** Lives in Nix-store cache
- **local** Local package
- **inplace** Local in-place package

I'd need a flag to build all packages which will end in the nix-store (which IMO we should rename to v2-build store: https://github.com/haskell/cabal/issues/6105, because nix-store is `/nix/store`). I think that `--only-dependencies` is close to that. To make more people happy, the `v2-build --only-dependencies` could also *download* inplace packages.

Given that, if you go on a flight, `v2-build --only-dependencies` will prepare you for an offline hacking session.

23Skidoo:
[...]

PR #6241, opened 2019-09-14
show-build-info (exe:cabal part plus tests)

fendor:
---

Since #6108 merged parts of #5954, the original pr #5954 got obsolete.
This now is the second part that contains only changes to cabal-install and the test-suite.
This is a push to get s-b-i merged for `exe:cabal`, enabling further improvements to it.

cc @hvr, @DanielG, @mpickering


cabal-install/Distribution/Client/CmdShowBuildInfo.hs
```diff
@@ -0,0 +1,327 @@
...
+import Distribution.Client.JobControl (newLock, Lock)
+import Distribution.Simple.Configure (tryGetPersistBuildConfig)
+import qualified Distribution.Client.CmdInstall as CmdInstall
+
+import System.Directory (getTemporaryDirectory)
+import System.FilePath ((</>))
+
+showBuildInfoCommand :: CommandUI ShowBuildInfoFlags
+showBuildInfoCommand = CmdInstall.installCommand {
+  commandName         = "new-show-build-info",
```
fendor: This should probably just be "show-build-info", right?

cabal-install/main/Main.hs
```diff
@@ -304,7 +305,8 @@ mainWorker args = do
       , hiddenCmd  win32SelfUpgradeCommand win32SelfUpgradeAction
       , hiddenCmd  actAsSetupCommand actAsSetupAction
       , hiddenCmd  manpageCommand (manpageAction commandSpecs)
-
+      , hiddenCmd CmdShowBuildInfo.showBuildInfoCommand
```
fendor: should it be a hidden command?

cabal-install/Distribution/Client/CmdShowBuildInfo.hs
DanielG: I think so, the default was already switched to v2 in 3.0.

fendor:
CI fail seems to be unrelated to this pr.

cabal-install/main/Main.hs
```diff
@@ -447,40 +448,67 @@ reconfigureAction flags@(configFlags, _) _ globalFlags = do
...
+  where verbosity = fromFlagOrDefault normal (buildVerbosity buildFlags)
+
+-- showBuildInfoAction :: (Cabal.ShowBuildInfoFlags, BuildExFlags) -> [String] -> Action
+-- showBuildInfoAction (showBuildInfoFlags, buildEx) = buildActionForCommand
+--   (Cabal.showBuildInfoCommand defaultProgramDb)
+--   showBuildInfoFlags
+--   verbosity
+--   (Cabal.buildInfoBuildFlags showBuildInfoFlags, buildEx)
+--     -- Default silent verbosity so as not to pollute json output
+--   where verbosity = fromFlagOrDefault silent (buildVerbosity (Cabal.buildInfoBuildFlags showBuildInfoFlags ))
```
DanielG: I think you moved this to a different file so you can probably just remove this?

DanielG: That also seems to make the whole buildActionForCommand refactoring unused, can you have another look at what's going on here?

cabal-install/main/Main.hs
```diff
@@ -522,8 +550,8 @@ replAction (replFlags, buildExFlags) extraArgs globalFlags = do
       -- be done to support sandboxes.
       _ <-
         reconfigure configureAction
-        verbosity distPref useSandbox noAddSource NoFlag
-        mempty [] globalFlags config
+          verbosity distPref useSandbox noAddSource NoFlag
+          mempty [] globalFlags config
```
DanielG: Unrelated whitespace change, just leave the code as is. This doesn't belong in this PR.

cabal-testsuite/PackageTests/ShowBuildInfo/A/build-info-unknown.out
```diff
@@ -0,0 +1,12 @@
+# cabal show-build-info
+cabal: Internal error in target matching. It should always be possible to find a syntax that's sufficiently qualified to give an unambiguous match. However when matching 'exe:B'  we found exe:B (unknown-component) which does not have an unambiguous syntax. The possible syntax and the targets they match are as follows:
```
DanielG: Um, this sounds like a cabal bug to me. Certainly nothing that should be expected test output or ever seen by users! Please find out what's going on here.

cabal-testsuite/PackageTests/ShowBuildInfo/A/cabal.project
```diff
@@ -0,0 +1 @@
+packages: .
```
DanielG: No newline at end of file :)

There's a bunch of other files in here that have this problem too. Mainly cosmetic I'll grant you but should still be fixed.

cabal-testsuite/PackageTests/ShowBuildInfo/B/B.cabal
```diff
@@ -0,0 +1,24 @@
...
+-- synopsis:
+-- description:
+-- bug-reports:
+license:             BSD-3-Clause
+license-file:        LICENSE
+author:              Foo Bar
+maintainer:          cabal-dev@haskell.org
+-- copyright:
+-- category:
+extra-source-files:  CHANGELOG.md
```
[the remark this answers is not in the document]
DanielG: I'd also appreciate it if you could strip all the comments out of the test cabal files.

cabal-testsuite/PackageTests/ShowBuildInfo/Complex/lib.test.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE DeriveGeneric #-}
+import           Test.Cabal.Prelude
+
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
+import           Data.Aeson
+import           GHC.Generics
+
+main = cabalTest $ do
+    r <- cabal' "show-build-info" ["lib:Complex", "-v0"]
```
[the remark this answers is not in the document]
DanielG: Ah I saw we have the output file redirection in Setup.hs for that, so I guss `-v0` isn't really necessary. Why is it used for all the tests then?

cabal-testsuite/PackageTests/ShowBuildInfo/A/build-info-unknown.out
grayjay: That error message looks like #5081.

cabal-install/main/Main.hs
fendor: Someone else needs to define this, I dont know enough of the process and the ramifications of such promises.

cabal-install/Distribution/Client/CmdShowBuildInfo.hs
```diff
@@ -0,0 +1,327 @@
...
+                  globalFlags configFlags configExFlags
+                  installFlags clientInstallFlags
+                  haddockFlags
+                  testFlags
+
+-- Pretty nasty piecemeal out of json, but I can't see a way to retrieve output of the setupWrapper'd tasks
+showTargets :: Maybe FilePath -> Maybe [String] -> Verbosity -> ProjectBaseContext -> ProjectBuildContext -> Lock -> IO ()
+showTargets fileOutput unitIds verbosity baseCtx buildCtx lock = do
+  tempDir <- getTemporaryDirectory
+  withTempDirectory verbosity tempDir "show-build-info" $ \dir -> do
```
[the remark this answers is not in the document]
fendor: Cabal s-b-i vomits a single json object. We capture it here and wrap it in a JSON Array, so that output is valid json.

cabal-testsuite/PackageTests/ShowBuildInfo/Complex/lib.test.hs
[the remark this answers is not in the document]
fendor: There is no reason why I preferred "-v0" over `output` flag. Just felt easier to deal with.

cabal-install/Distribution/Client/CmdShowBuildInfo.hs
```diff
@@ -0,0 +1,327 @@
...
+     ++ "    Shows build information about the package located in './pkgname'\n"
+     ++ cmdCommonHelpTextNewBuildBeta,
+  commandOptions = \showOrParseArgs ->
+      Client.liftOptions buildInfoInstallCommandFlags (\pf flags -> flags { buildInfoInstallCommandFlags = pf }) (commandOptions CmdInstall.installCommand showOrParseArgs)
+      ++
+      [ option [] ["buildinfo-json-output"]
+              "Write the result to the given file instead of stdout"
+              buildInfoOutputFile (\pf flags -> flags { buildInfoOutputFile = pf })
+              (reqArg' "FILE" Just (maybe [] pure)),
+        option [] ["unit-ids-json"]
```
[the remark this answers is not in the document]
fendor: Right, makes sense.

cabal-install/Distribution/Client/CmdShowBuildInfo.hs
```diff
@@ -0,0 +1,327 @@
...
+              False
+              lock
+          configureFlags = setupHsConfigureFlags (ReadyPackage pkg) shared verbosity buildDir
+          configureArgs = setupHsConfigureArgs pkg
+
+      -- check cabal version is corrct
+      (cabalVersion, _, _) <- getSetupMethod verbosity scriptOptions
+                                            (elabPkgDescription pkg) buildType'
+      when (cabalVersion < mkVersion [3, 0, 0,0])
+        ( die' verbosity $ "Only a Cabal version >= 3.0.0.0 is supported for this command.\n"
```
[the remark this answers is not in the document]
phadej: I think that those packages should have some dummy entry in the output, i.e. not `die`.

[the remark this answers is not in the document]
bubba: @phadej I've noticed that `cabal show-build-info` doesn't work on `build-type: Custom` without first running `cabal build` to generate the package db. `cabal configure` doesn't seem to generate it. Should the package db be needed?

bubba:
I've been trying to integrate this with `hie-bios` and it's been going well so far. There are some things I would like to see addressed first though:

- `runghc Setup.hs show-build-info` returns one JSON object, whereas `cabal show-build-info` returns an array of JSON objects, one for each component/unit it plans on building, namely due to how it is calling the wrapper several times. But this seems to me to defeat the purpose of the `components` array, as it is possible to get `runghc Setup.hs show-build-info` to return one object containing all the components needed
- With `runghc Setup.hs show-build-info`, you need to configure it first with `--enable-tests`/`--enable-benchmarks` to see the build info for tests and benchmarks respectively. This makes sense for the Setup.hs method, but you also need to do the same for `cabal show-build-info`. For the sake of tooling integration I think `cabal show-build-info` should default to including tests and benchmarks without having to pass `--enable-tests/--enable-benchmarks`

Cabal/Distribution/Utils/Json.hs
```diff
@@ -1,7 +1,8 @@
--- | Utility json lib for Cabal
--- TODO: Remove it again.
-module Distribution.Simple.Utils.Json
+-- | Extremely simple JSON helper. Don't do anything too fancy with this!
```
fendor, line 1: I think it was the idea to replace this with a dependency such as microaeson https://hackage.haskell.org/package/microaeson

phadej: `Cabal` cannot take any dependencies easily.

bubba: It looked like `Cabal` only depends on boot packages, hence why I thought it would just be easier to expose this module

bubba:
@phadej @fendor I've changed this to avoid calling Setup.hs to get the show-build-info result (It still uses it for configuring the components). That way we can collect the component information separately and use the correct compiler information, and return just one build info object rather than several, whilst also avoiding the temporary files. So there's a bit of poking about done with `LocalBuildInfo` here which I'm aware seems to be taboo within cabal-install

test.hs
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env cabal
+{- cabal:
+build-depends: base
+with-compiler: ghc-8.6.5
+-}
+main = putStrLn "hey"
```
bubba, line 6: Ugh accidentally added this and hie.yaml during a rebase. Will fix

bubba:
I think `cabal show-build-info` should implicitly build dependencies, much like `cabal repl`, otherwise it will fail on a clean project when trying to configure. It should behave equivalently to

```
cabal build --only-dependencies <targets>  && cabal show-build-info <targets>
```

Passing the targets to the build part will be important, as with `cabal repl` only the dependencies of the target need to be buildable before ghci can be launched with the right flags

phadej:
`--only-dependencies` could fail easily: https://github.com/haskell/cabal/issues/6106

If project contains packages `A` and `C`, and there is third package `B` (on Hackage),
then if dependency graph is `A -> B -> C`, you essentially need to build also `C` to build `B` to configure `A`.

We don't have such more of operation currently. Would be good to have. Actually both:
- build all dependencies you can without building local packages
- build all dependencies, incl. some local packages, so all external dependencies are built. I.e. you can configure packages.

bubba:
I ran into this whilst testing out integrating this with hie-bios, https://github.com/haskell/cabal/issues/6874 and I think there should be a mechanism to allow for ambiguous target selectors to be resolved, e.g. by just selecting the first one. Namely, hie-bios isn't going to care what component's flags `cabal show-build-info` returns, as long as it can actually get the flags in the first place

bubba:
### Todo 

- [ ] Make `AmbiguousTargetResolver` a command line flag so other commands can use it. `--pick-first-target`

bubba:
Tricky situation with dependencies: Given this simple cabal file
```
cabal-version: 2.4
name:          C
version:       0.1.0.0
license:       BSD-3-Clause

library
  exposed-modules:  Lib
  build-depends:    base
  default-language: Haskell2010

test-suite tests
  type: exitcode-stdio-1.0
  main-is: Test.hs
  build-depends: base, C
  default-language: Haskell2010
```

If we are to run `cabal show-build-info` verbatim without passing any targets to it at all, it will fail at configuring the test-suite because the library C will not have been built. `cabal show-build-info Test.hs` does work though, as it seems to count the library C as a dependency in this scenario

bubba:
@DanielG Phew, reworked it to now use the nix stuff. Hopefully ci passes

bubba:
New and noteworthy things in this PR:

- Added cabal-install show-build-info
- `--pick-first-target` flag to resolve ambiguous target selectors to just the first target
- show-build-info now returns the location of the cabal file as well as the unpacked source directory of the package (not the hs-src-directory, which is different)
- show-build-info now uses Text rather than String internally
- show-build-info has the `buildinfo-components-only` flag which spits out each component's json line by line. Used by cabal-install to piece together components that are split across multiple per-component packages

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,65 @@
...
+data Json = JsonArray [Json]
+          | JsonBool !Bool
+          | JsonNull
+          | JsonNumber !Int
+          | JsonObject [(Text, Json)]
+          | JsonRaw !Text
+          | JsonString !Text
+
+-- | A type to mirror 'ShowS'
+type ShowT = Text -> Text
```
phadej, line 22: I spotted this via your comment. Why this and not https://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-Lazy-Builder.html

The `ShowT = Text -> Text` is worse than `ShowS`, as it copies all intermediate values, causing quadratic behavior.

Yet, if you doing this for performance reasons, you should use `ByteString` and its builder, like `aeson`.

cabal-install/Distribution/Client/CmdShowBuildInfo.hs
```diff
@@ -0,0 +1,209 @@
...
+      components = map JsonRaw componentBuildInfos
+      fields = mkBuildInfo' compilerInfo components
+      json = JsonObject $ fields <>
+        [ ("project-root", JsonString (T.pack (distProjectRootDirectory (distDirLayout baseCtx))))
+        ]
+      res = renderJson json ""
+
+  case fileOutput of
+    Nothing -> T.putStrLn res
+    Just fp -> T.writeFile fp res
```
phadej, line 157: This is locale dependent, better to serialise  to `ByteString` and write it.

phadej:
Separate JSON utility into own PR, it is independent and can be reviewed and committed independently.

Cabal/Distribution/Utils/Json.hs
bubba: Accidentally quadratic, how embarrassing! I'll switch this all over to `ByteString`, didn't realise `Text` was also locale dependent

masaeedu:
@jneira @fendor Sorry, would you mind expanding on what this has been superseded with? There seems to now be a network of issues about `show-build-info` that are all pointing at each other saying X is superseded by Y and Y is superseded by Z and so on :)

Is there a "can i run `cabal show-build-info` yet" issue somewhere?

fendor:
This issue summarises the current ideas, discussions and efforts: https://github.com/haskell/cabal/issues/7489

Specifically this is a PR overview: https://github.com/haskell/cabal/issues/7489#issuecomment-894387350

PR #6255, opened 2019-09-24
Structured binary

phadej:
---

### DIFF
--- a/boot/SPDX.LicenseExceptionId.template.hs
+++ b/boot/SPDX.LicenseExceptionId.template.hs
@@ -11,9 +11,11 @@ module Distribution.SPDX.LicenseExceptionId (
 import Distribution.Compat.Prelude
 import Prelude ()
 
+import Distribution.Compat.Lens (set)
 import Distribution.Pretty
 import Distribution.Parsec
 import Distribution.Utils.Generic (isAsciiAlphaNum)
+import Distribution.Utils.Structured (Structured (..), nominalStructure, typeVersion)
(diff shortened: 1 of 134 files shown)


cabal.project
```diff
@@ -1,4 +1,5 @@
-packages: Cabal/ cabal-testsuite/ cabal-install/ solver-benchmarks/
+packages: Cabal/ cabal-testsuite/
+-- packages: cabal-install/ solver-benchmarks/
```
phadej: hups, will undo.

boot/SPDX.LicenseExceptionId.template.hs
```diff
@@ -39,6 +40,10 @@ instance Binary LicenseExceptionId where
         then fail "Too large LicenseExceptionId tag"
         else return (toEnum (fromIntegral i))
 
+-- note: remember to bump version each time the definition changes
```
DanielG: This seems to defeat the purpose of this patch. Any way this datatype can be included in the structure hash too?

phadej: Manual `Binary` instance => Manual `Structured` instance.

Yet, here `Binary` instance is generated by templating, so yes, we should generate `Structured` instance too. I'll take the *license list version* and use that.

Good catch.

phadej:
> Would be nice if there was some documentation on what exactly this is trying to do. Your commits don't have any messages, pretty hypocritical to complain when other people do this and then not have any yourself :P

This is WIP PR, I will rewrite commit messages.

PR #6257, opened 2019-09-24
Allow Diff-0.4

phadej:


23Skidoo:
Looks OK, let's merge when it's green.

phadej:
Appveyor fails due the reason it fails in `master` too.

PR #6258, opened 2019-09-25
buildinfo-reference-generator

phadej:
So we can generate from code:

(I didn't specify monoidal fields, thus it's very incomplete)

Screenshot from 2019-09-25 11-37-30


phadej:
Current status
Screenshot from 2019-09-27 14-02-36

Issue #6232, opened 2019-09-10
Extended preprocessors support


boris-stepanov:
Hey, Cabal team.

Some time ago, I had to integrate [Apache Thrift](https://thrift.apache.org/) in one of my projects. But, I faced issues with [HookedPreprocessors](https://hackage.haskell.org/package/Cabal-3.0.0.0/docs/Distribution-Simple.html#t:UserHooks).

Let's say, that we have a preprocessor that can translate files with the `.foo` extension. If you declare some module in your `exposed-modules` (f.e. Test.Module), then Cabal will look for the file named `Test.Module.foo` and run the preprocessor for it. It is one-to-one correspondence.

Now, we have Apache Thrift. The preprocessor can parse Thrift DSL and generate protocols and RPC servers/clients. But, it will translate single `Module.thrift` file in many others: `Module_Types.hs`, `Module_Consts.hs`, `RPCPoint_Iface.hs`, `RPCPoint_Client.hs`. The first two files correspond to the source file, while the rest correspond to a single RPCPoint service inside the `Module.thrift`. There can be other interface/client files, if you declare more services in the `Module.thrift`.

So, here is a problem: Cabal finds `Test.Module` in exposed-modules sections, finds `Test/Module.thrift` among source files and thinks that's it, but after the preprocessing stage, there is no `Module.hs` among the build files. On the other hand, if you declare `Module_Types/Module_Consts` and others in your `exposed-modules` section, that Cabal will ignore `Module.thrift` source file.

What I did in my project:
  1. I've placed the single pattern `thrift/*.thrift` in the `extra-source-files` field.
  2. I've written custom `Setup.hs` file with overloaded `buildHook/testHook`. Before running the default hook, they will call the thrift preprocessor for every file matching the pattern in `extraSrcFiles` with the destination `buildDir`.

This approach works, but I don't like it:
  1. Files specific for the build are listed in a global parameter.
  2. Part of preprocessing stage is placed in the `buildHook`. Also, I have duplicated code in the `testHook` (and I would have it in the `benchHook`, if I needed to).
  3. There are other preprocessors with similar problems.

I suggest to remove pattern matching on preprocessor stage at all. I think, time savings in case of dropping unnecessary files doesn't make a difference.

ulysses4ever:
I wonder if this issue has been addressed by either code-generators https://github.com/haskell/cabal/pull/7688 or the Hooks build type  #9292.

sheaf:
[...]

PR #6261, opened 2019-09-26
Bump cabal-version allowed in .cabal files

vmchale:
This is related to https://github.com/haskell/cabal/pull/6233. When `c2hs` handling is fixed, some packages will only build with a recent `Cabal`. 

Tested it by running on [libarchive](https://github.com/vmchale/libarchive).

### DIFF
--- a/Cabal/Distribution/PackageDescription/Parsec.hs
+++ b/Cabal/Distribution/PackageDescription/Parsec.hs
@@ -106,7 +106,7 @@ parseGenericPackageDescription bs = do
     setCabalSpecVersion ver
     -- if we get too new version, fail right away
     case ver of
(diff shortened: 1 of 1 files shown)


phadej:
Next `cabal-version: ` will be 3.2 not 3.1. It’s easy to confuse Cabal library version with cabal-version version of the spec, as they are the same for released versions.

Do we have anything casing on CabalSpecVersion3_2 yet?

phadej:
I checked, neither #6232 nor nothing already existing cases on `CabalSpecV3_2` yet.

vmchale:
I suppose that depends on whether https://github.com/haskell/cabal/pull/6233 is merged, then.

(I have updated the file to use 3.2 instead of 3.1)

Issue #6005, opened 2019-04-18
Problems using cabal-install with backpack


vmchale:
Hi,

I'm trying to use `cabal` with a backpack-ified library, viz. [`archive-tar`](http://hackage.haskell.org/package/archive-tar) (which provides a signature for [`archive-sig`](http://hackage.haskell.org/package/archive-sig). Unfortunately, the build fails with

```
Installing   archive-sig-0.1.0.0 (lib)

Failed to build archive-sig-0.1.0.0. The failure occurred during the final
install step.
Build log (
/home/vanessa/.cabal/logs/ghc-8.6.4/archive-sig-0.1.0.0-63b09de7c791f4fc3a113a853baaa9866da61a06d6904f7a1606df18d4d93fe8.log
):
Configuring library for archive-sig-0.1.0.0..
Preprocessing library for archive-sig-0.1.0.0..
Building library instantiated with Archive = <Archive>
for archive-sig-0.1.0.0..
[1 of 2] Compiling Archive[sig]     ( src/Archive.hsig, nothing )
[2 of 2] Compiling Archive.Generic  ( src/Archive/Generic.hs, nothing )
Installing library in /home/vanessa/.cabal/store/ghc-8.6.4/incoming/new-24224/home/vanessa/.cabal/store/ghc-8.6.4/archive-sig-0.1.0.0-63b09de7c791f4fc3a113a853baaa9866da61a06d6904f7a1606df18d4d93fe8/lib
cabal: '/opt/ghc/bin/ghc' exited with an error:
Failed to load interface for ‘Data.DList’
no unit id matching
‘dlist-0.8.0.6-8fa4f62b6f8ca69e1a8182b60bd05c108ae5778932e24195d5646fce42a11a17’
was found

cabal: Failed to build archive-sig-0.1.0.0 (which is required by exe:atspkg
from ats-pkg-3.2.5.12). See the build log above for details.
```

You can see a branch with a failing project [here](https://github.com/vmchale/atspkg/tree/backpack-fail/ats-pkg). I'm happy to provide more debug information. 

Thanks!

andrewthad:
I’m getting the same problem when I try to build `sockets-0.4.0.0`. I can build the library fine  if I just clone of the source and build it.  However, installing the library from hackage doesn’t work.  I’m using cabal-install 3.0 from hvr’s ppa.  The build failure I get mentions A missing unit I’d for attoparsec. However, sockets does not depend on attoparsec even transitively.

[...]

PR #6264, opened 2019-09-30
Warn that dependencies using backpack are broken when installed from hackage

andrewthad:
Documents the deficiency from https://github.com/haskell/cabal/issues/6005 in the user manual. In the event that the issue does not get fixed for some time, documenting this should help prevent people from wasting space on hackage by uploading packages that no build tool can currently build.

### DIFF
--- a/Cabal/doc/nix-local-build.rst
+++ b/Cabal/doc/nix-local-build.rst
@@ -656,7 +656,7 @@ the custom publishing of Haddock documentation to Hackage.
 .. warning::
 
   Packages that use Backpack will stop working if uploaded to
-  Hackage, due to https://github.com/haskell/cabal/issues/6005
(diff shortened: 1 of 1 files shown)


ezyang:
Sorry about not having had time to look into the issue. I think it would be more useful for users if there was a link to the issue in question?

ezyang:
I reworded the doc to this effect. If you like it we can merge this.

andrewthad:
That looks good. My restructured text knowledge isn't great, but I've attempted to do a little formatting of the hyperlink to make it more consistent with the way the other ones look.

For what it's worth, I suspect that backpack hasn't seen enough uptake in libraries yet for this issue to cause much pain for many users. The feeling I get is that most people using backpack are writing applications with a mega-repo structure, where the workaround using a `cabal.project` file is not bad at all.

PR #6274, opened 2019-10-01
Upgrade cabal-install to 3.0.0.0 on appveyor

phadej:
TBW.

~~An experiment, I think appveyor fails (but it shouldn't).~~

2nd hypothesis: disabled tests fail


phadej:
Let's upgrade it later, when `cabal-install-3.0` matures itself :)

PR #6284, opened 2019-10-06
[WIP] Fix cabal-install bootstrapping with GHC 8.8

DanielG:
This is still blocked by https://github.com/haskell/hackage-security/issues/229 so I'm not sure if I cought all necessary bounds changes. I think we should issue a 3.0.0.1 patch release for this since currently bootstrapping on 8.8 is just completely broken.

This probably should also be cherry picked onto master eventually.


hvr:
> I think we should issue a 3.0.0.1 patch release for this since currently bootstrapping on 8.8 is just completely broken.

That's not the reason we need a new release; we rather need it because only a new release of cabal will be allowed to link to the next major version of hackage-security; we don't want a `cabal-install-3.0.0.0` whose behaviour subtly differs depending on the major versions of its dependencies...

phadej:
This is outdated.

PR #6286, opened 2019-10-06
Exe dependency less lexeme

phadej:
This is tricky: There are no `happy :happy` (note the space) uses of `build-tool-depends` on Hackage, so if we are quick to make Cabal-3.0.x.y with this patch and deploy `hackage-server` using it, it can be as simple as this.

Note: this is different than with spaces in `mixins` as here I'm *restricting* the grammar: existing `cabal-install` / tools using old `Cabal` will continue to work fine.

Edit: look at the `Make ExeDependency parser stricter` commit to see how the example output change.


phadej:
ping @hvr, @23Skidoo

phadej:
Backported to 2.4 and 3.0 branches

PR #6287, opened 2019-10-06
Fix dropExeExtension on Windows

DanielG:
On Windows dropExeExtension doesn't actually drop the exe extension.

We have

    splitExtension "foo.exe" == ("foo", ".exe")

but the code is expecting just `"exe"` for the extension field.

Among other things this makes `guessToolFromGhcPath` behave
unexpectedly. Since `takeVersionSuffix` can't see past the extension the
`null suf` case in `mkGuesses` will trigger which causes the versioned
program path candidates to not be looked for.

---

### DIFF
--- a/Cabal/Distribution/Compat/FilePath.hs
+++ b/Cabal/Distribution/Compat/FilePath.hs
@@ -1,0 +1,27 @@
+{-# LANGUAGE CPP #-}
+
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+
+module Distribution.Compat.FilePath
+( isExtensionOf
+, stripExtension
+) where
+
+import Data.List ( isSuffixOf, stripPrefix )
+import System.FilePath
+
+#if !MIN_VERSION_filepath(1,4,2)
+isExtensionOf :: String -> FilePath -> Bool
+isExtensionOf ext@('.':_) = isSuffixOf ext . takeExtensions
+isExtensionOf ext         = isSuffixOf ('.':ext) . takeExtensions
+#endif
+
+#if !MIN_VERSION_filepath(1,4,1)
+stripExtension :: String -> FilePath -> Maybe FilePath
+stripExtension []        path = Just path
+stripExtension ext@(x:_) path = stripSuffix dotExt path
+ where
+  dotExt = if isExtSeparator x then ext else '.':ext
+  stripSuffix :: Eq a => [a] -> [a] -> Maybe [a]
+  stripSuffix xs ys = fmap reverse $ stripPrefix (reverse xs) (reverse ys)
+#endif
--- a/Cabal/Distribution/Simple/Utils.hs
+++ b/Cabal/Distribution/Simple/Utils.hs
@@ -177,6 +177,7 @@ import Distribution.Version
 import Distribution.Compat.CopyFile
 import Distribution.Compat.Internal.TempFile
 import Distribution.Compat.Exception
+import Distribution.Compat.FilePath as FilePath
 import Distribution.Compat.Stack
 import Distribution.Verbosity
 import Distribution.Types.PackageId
@@ -212,7 +213,7 @@ import System.Environment
     ( getProgName )
 import System.Exit
     ( exitWith, ExitCode(..) )
-import System.FilePath
+import System.FilePath as FilePath
     ( normalise, (</>), (<.>)
     , getSearchPath, joinPath, takeDirectory, splitExtension
     , splitDirectories, searchPathSeparator )
@@ -1411,11 +1412,25 @@ shortRelativePath from to =
 -- unchanged.
 dropExeExtension :: FilePath -> FilePath
 dropExeExtension filepath =
-  case splitExtension filepath of
-    (filepath', extension) | extension `elem` exeExtensions -> filepath'
-                           | otherwise                      -> filepath
+  -- System.FilePath's extension handling functions are horribly
+  -- inconsistent, consider:
+  --
+  --     isExtensionOf "" "foo"  == False but
+  --     isExtensionOf "" "foo." == True.
+  --
+  -- On the other hand stripExtension doesn't remove the empty extension:
+  --
+  --    stripExtension "" "foo." == Just "foo."
+  --
+  -- Since by "" in exeExtensions we mean 'no extension' anyways we can
+  -- just always ignore it here.
+  let exts = [ ext | ext <- exeExtensions, ext /= "" ] in
+  fromMaybe filepath $ do
+    ext <- find (`FilePath.isExtensionOf` filepath) exts
+    ext `FilePath.stripExtension` filepath
 
--- | List of possible executable file extensions on the current platform.
+-- | List of possible executable file extensions on the current build
+-- platform.
 exeExtensions :: [String]
 exeExtensions = case buildOS of
   -- Possible improvement: on Windows, read the list of extensions from the
--- a/Cabal/tests/UnitTests/Distribution/Simple/Utils.hs
+++ b/Cabal/tests/UnitTests/Distribution/Simple/Utils.hs
@@ -2,13 +2,16 @@ module UnitTests.Distribution.Simple.Utils
     ( tests
     ) where
 
+import Distribution.Simple.BuildPaths ( exeExtension )
 import Distribution.Simple.Utils
+import Distribution.System ( buildPlatform )
 import Distribution.Verbosity
 
 import Data.IORef
 import System.Directory ( doesDirectoryExist, doesFileExist
                         , getTemporaryDirectory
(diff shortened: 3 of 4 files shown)


Cabal/tests/UnitTests/Distribution/Simple/Utils.hs
```diff
@@ -98,4 +103,6 @@ tests ghcPath =
       withTempDirRemovedTest
     , testCase "rawSystemStdInOut reports text decoding errors" $
       rawSystemStdInOutTextDecodingTest ghcPath
+    , testCase "dropExeExtension actually drops exe extension" $
```
phadej: `s/actually//`

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1412,8 +1412,10 @@ shortRelativePath from to =
...
   case splitExtension filepath of
-    (filepath', extension) | extension `elem` exeExtensions -> filepath'
-                           | otherwise                      -> filepath
+    (filepath', extension)
+      | extension `elem` exts -> filepath'
+      | otherwise             -> filepath
+  where exts = [ "." ++ ext | ext <- exeExtensions ]
 
 -- | List of possible executable file extensions on the current platform.
 exeExtensions :: [String]
```
phadej: Could we improve the haddocks here to say

> on the build platform

as current is ambiguous, could mean target platform too.

That would explain why `Ghcjs` entry doesn't (and shouldn't?) have `jsexe` listed.

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1415,14 +1415,14 @@ dropExeExtension filepath =
...
+-- | List of possible executable file extensions on the current build platform.
 exeExtensions :: [String]
 exeExtensions = case buildOS of
   -- Possible improvement: on Windows, read the list of extensions from the
   -- PATHEXT environment variable. By default PATHEXT is ".com; .exe; .bat;
   -- .cmd".
-  Windows -> ["", "exe"]
-  Ghcjs   -> ["", "exe"]
+  Windows -> ["", ".exe"]
+  Ghcjs   -> ["", ".exe"]
```
DanielG: So technically this is breaking API compatibility since this is exported, but there don't seem to be any usages outside of Cabal oh hackage: https://codesearch.aelve.com/haskell/search?query=exeExtensions&insensitive=off&space=off&precise=on&sources=on

Furthermore `<.>` doesn't care about the leading dot in the extension, so `"foo" <.> "bar" == "foo" <.> ".bar" == "foo.bar"`.

phadej:
```
tests/UnitTests/Distribution/Simple/Utils.hs:12:27:
    Module `System.Directory' does not export `exeExtension'
```

on old GHCs. There's `exeExtension` in `Distribution.Simple.BuildPaths` though

DanielG:
Argh, more CPP needed :)

Cabal/tests/UnitTests/Distribution/Simple/Utils.hs
```diff
@@ -84,6 +85,10 @@ rawSystemStdInOutTextDecodingTest ghcPath
     Left err | isDoesNotExistError err -> Exception.throwIO err -- no ghc!
              | otherwise               -> return ()
 
+dropExeExtensionTest :: Assertion
+dropExeExtensionTest =
+  assertBool "dropExeExtension didn't drop exeExtension!" $
+    dropExeExtension ("foo" ++ exeExtension) == "foo"
```
phadej: you don't use `<.>` here?

DanielG: I suppose I could use `<.>` if you prefer? Since it swallows the second dot anyways.

phadej: You **should**. You are adding extension to a `FilePath`, even it's a `String` (there's a dream that we could move to opaque path type)

DanielG: Haha, sure :)

phadej: TL;DR


IMHO you should use `Distribution.Simple.BuildPaths.exeExtension Distribution.Simple.buildPlatform`

and avoid CPP

DanielG:
How many revisions can a PR that essentially adds a `"."` possibly need!? ;-)

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1415,14 +1415,14 @@ dropExeExtension filepath =
...
--- | List of possible executable file extensions on the current platform.
+-- | List of possible executable file extensions on the current build platform.
 exeExtensions :: [String]
 exeExtensions = case buildOS of
   -- Possible improvement: on Windows, read the list of extensions from the
   -- PATHEXT environment variable. By default PATHEXT is ".com; .exe; .bat;
   -- .cmd".
-  Windows -> ["", "exe"]
-  Ghcjs   -> ["", "exe"]
+  Windows -> ["", ".exe"]
```
phadej: and as I think it, it's confusing that we have **both**, this and https://hackage.haskell.org/package/Cabal-3.0.0.0/docs/src/Distribution.Simple.BuildPaths.html#exeExtension without any cross-reference.

It would be confusing if one has dot, and another doesn't.

phadej:
Cabal codebase has the following property: if something is not fixed immediately as spotted, it will take years to be spotted again (and this fixed).

DanielG:
Well pretty much any codebase has that property ;) That's why I try to shoot patches as soon as I find issues.

phadej:
```
tests\UnitTests\Distribution\Simple\Utils.hs:92:33: error:
    * Couldn't match type `Distribution.System.Platform -> String'
                     with `[Char]'
      Expected type: String
        Actual type: Distribution.System.Platform -> String
    * Probable cause: `exeExtension' is applied to too few arguments
      In the second argument of `(<.>)', namely `exeExtension'
      In the first argument of `dropExeExtension', namely
        `("foo" <.> exeExtension)'
      In the first argument of `(==)', namely
        `dropExeExtension ("foo" <.> exeExtension)'
   |
92 |     dropExeExtension ("foo" <.> exeExtension) == "foo"
   |                                 ^^^^^^^^^^^^
cabal.exe: Failed to build test:unit-tests from Cabal-3.1.0.0.
```

Cabal/Distribution/Simple/Utils.hs
DanielG: This whole extensions buissness is just inconsistent as hell, consider this:

```
splitExtension "foo." == ("foo", ".")
"." `isExtensionOf` "foo." == True -- So far so good

splitExtension "foo" == ("foo", "")
"" `isExtensionOf` "foo" == False -- Ugh...
```
If the dot in an extension wasn't optional things could be nice and consistent, where `""` stands for no extension and `"."` stands for the empty extension. So I'm not sure what to do here, pretty much all the options are gross and ad-hoc.

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1411,11 +1412,24 @@ shortRelativePath from to =
...
+  --     isExtensionOf "" "foo." == True.
+  --
+  -- On the other hand stripExtension doesn't remove the empty extension:
+  --
+  --    stripExtension "" "foo." == Just "foo."
+  --
+  -- Since by "" in exeExtensions we mean 'no extension' anyways we can
+  -- just always ignore it here.
+  let exts = [ ext | ext <- exeExtensions, ext /= "" ] in
+  fromMaybe filepath $
```
phadej: I'd wrote this as

```haskell
fromMaybe filepath $ do
  ext <- find (`isExtensionOf` filepath)
  stripExtension ext filepath
```

IMHO, pointfree is pointless here.

DanielG: Sure thing.

DanielG:
Travis is green except for Mac, with one timeout and one network error :)

Cabal/Distribution/Compat/FilePath.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE CPP #-}
+
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+
+module Distribution.Compat.FilePath
+( isExtensionOf
+, stripExtension
+) where
+
+import Data.List
```
phadej: I'd use explicit import lists

DanielG: For `Data.List` ok, but for FilePath I'd have to use CPP in the import list which I'd rather not.

phadej: Yes.

PR #6296, opened 2019-10-14
Use ghc as the linker with hsc2hs

simonmar:
(Note that this depends on https://github.com/haskell/cabal/pull/6295 and includes that commit)

This enables hsc2hs to depend on Haskell code via foreign export. I have
an upcoming use case that requires this.

Should we make this optional? I'm hoping we can get away without adding
an option, because it shouldn't break any existing uses: using GHC as
the linker only provides access to more symbols, which shouldn't break
anything provided the libraries that GHC adds don't shadow anything,
which should be the case.

---


ffaf1:

simonmar:
I created this PR, and I don't need it any more so feel free to delete.

PR #6300, opened 2019-10-18
Backport fix for dropExeExtension to 3.0 branch

DanielG:
Backport of #6287 onto 3.0. 

I would appreciate it if we could get a `lib:Cabal` point release for this out. I depend on it over in cabal-helper.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -306,6 +306,7 @@ library
     Distribution.Compat.Directory
     Distribution.Compat.Environment
     Distribution.Compat.Exception
+    Distribution.Compat.FilePath
     Distribution.Compat.Graph
     Distribution.Compat.Internal.TempFile
     Distribution.Compat.Newtype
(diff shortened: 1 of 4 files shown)


hvr:
@DanielG are you aware of a0d7bc804bda16bee97ada5ef9fa7f6c9c472a40 ? I.e. are we sure this doesn't regress the thing I workarounded with said commit?

DanielG:
I did not know about that commit. I don't see why this fix would regress it though? Doesn't look like it's even using dropExeExtension anywhere. You might be thinking of one of the earlier revisions of my fix where I modify `exeExtensions`, I didn't end up doing that in the end.

I looks like unlike my fix you didn't include a regression test so who can say. If you want your fixes to stay fixed that might well be a thing to do ;)

DanielG:
Ping. CI Looks green apart from a windows failure which exists on `3.0` as well and some OSX timeouts.

Windows:
```
cabal.exe: filepath wildcard 'tests/README.md' does not match any files.
Command exited with code 1
```

phadej:
```
cabal.exe: filepath wildcard 'tests/README.md' does not match any files.
```

looks like a failure in `master` we have atm, weird.

phadej:
Backported as in bf862b8d219b2c2ef7bb8d61118d05f7b5aed5f2

PR #6301, opened 2019-10-18
Accept "linux-androideabi" as an alias for Android

Ericson2314:
Google uses `armv7-none-linux-androideabi`, analogous to
`armv7-unknown-linux-gnueabi` for e.g. a 32-bit regular linux.

The existing alias it not a mistake, it is the right one for 64 bit.
Google for that use `aarch64-linux-android`, analogous to
`aarch64-unknown-linux-android` for a 64-bi regular linux.

See https://developer.android.com/ndk/guides/standalone_toolchain for
evidence/details.

---

### DIFF
--- a/Cabal/Distribution/System.hs
+++ b/Cabal/Distribution/System.hs
@@ -122,7 +122,8 @@ osAliases Permissive FreeBSD = ["kfreebsdgnu"]
 osAliases Compat     FreeBSD = ["kfreebsdgnu"]
 osAliases Permissive Solaris = ["solaris2"]
 osAliases Compat     Solaris = ["solaris2"]
-osAliases _          Android = ["linux-android"]
+osAliases Permissive Android = ["linux-android", "linux-androideabi", "linux-androideabihf"]
+osAliases Compat     Android = ["linux-android"]
 osAliases _          _       = []
 
 instance Pretty OS where


Cabal/Distribution/System.hs
```diff
@@ -122,7 +122,7 @@ osAliases Permissive FreeBSD = ["kfreebsdgnu"]
 osAliases Compat     FreeBSD = ["kfreebsdgnu"]
 osAliases Permissive Solaris = ["solaris2"]
 osAliases Compat     Solaris = ["solaris2"]
-osAliases _          Android = ["linux-android"]
+osAliases _          Android = ["linux-android", "linux-androideabi" ]
```
phadej: I think we should gate these based on `cabal-version` / `CabalSpecVersion`. This affects cabal file format.

cc @hvr

Ericson2314: I think in practice this is more likely to fix old packages than break them. `linux-android` was added without such a gate https://github.com/haskell/cabal/pull/5199/files, and before that GHC broke things on Cabal's end when it got rid of its notion of ios and android, so way back when `os(android)` probably worked for both OSes.

phadej: The fact that we were sloppy in the past doesn't justify being sloppy in the future.

```
Warning: The following errors will cause portability problems on other
environments:
Warning: Unknown operating system name 'linux-androideabi'
Warning: Hackage would reject this package.
```

The `osAliases` are part of *cabal file format* and I'm going to be the strict about their changes. Let's do things properly, even it requires more extensive changes. The more we do it so, the easier future changes will be.

---

I think the **right** solution is to have separate `osAliases`: the one used in `.cabal` files, and the ones used to match against `System.Info.os`.

- `.cabal` files list should not be changed easily, and should be documented. We really don't need be able to use `mingw32` to mean `windows`
- latter ones can be changed on need.

Many purposes, tricky to change. Decouple: profit!

Ericson2314: Yeah decoupling sound good. I only expect anyone to use anything but `os(android)`. I wouldn't even spend any time advocating anything else to work.

`os(linux-android)` was just something people started using as a workaround to `android` not being recognized -- a sort of workdaround that I don't even think works anymore as `os(android) || os(linux-androideabi)` didn't work when I tried it. In fact, I would recommend you revision that away on Hackage if there's any up there.

Ericson2314: Also, does `ClassificationStrictness` already handle this split, albiet obliquely?

Ericson2314:
I used `Permissive` and `Compat` for now to avoid changing the Cabal file format, as far as I can tell, and also added `linux-androideabihf`, which means hardware floating point support.

3noch:
What's holding this up now?

phadej:
Merged rebased in https://github.com/haskell/cabal/pull/6949

Ericson2314:
Thanks!

Issue #5698, opened 2018-11-19
Add support for (some) `cabal.project`-fields to cabal script-metadata


hvr:
The basic script feature (see #3843) implemented has support for including package-description/buildinfo fields but didn't yet implement support for project metadata.

At the very least fields such as `index-state`, `with-compiler`, `allow-{newer,older}` and `constraints` shall be supported as exemplified in the example below

```hs
#! /usr/bin/env cabal
{- cabal:
index-state: 2016-09-15T12:00:00Z
with-compiler: ghc-8.0.2
allow-newer: shelly:base
constraints: filepath==1.4.1.1, containers==0.5.7.1, Cabal installed, cryptonite -none

build-depends: base ^>= 4.9
             , shelly ^>= 1.6.8
-}

main :: IO ()
main = do
   ...
```

/cc @typedrat

23Skidoo:
We should also think about the backwards compat story for `{- cabal:` blocks while we're at it.

hvr:
@phadej good catch!

@23Skidoo `cabal:` afaik is versioned; i.e. it's actually `{- cabal:<script-spec-version>`

23Skidoo:
@hvr
>  `cabal:` afaik is versioned; i.e. it's actually `{- cabal:[<script-spec-version>]` (and the spec ver is optional)

Is it documented somewhere which `script-spec-version` is the implicit one?

23Skidoo:
Not against this proposal, but wouldn't something like

```
#! /usr/bin/env cabal --allow-newer=foo,bar,baz
```

work too?

hvr:
@23Skidoo Unfortunately, what is allowed under POSIX  for portable shebang scripts is *very* limited; see https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability for the sad story

23Skidoo:
okay.jpg

cartazio:
does this require the file be named Main.hs or have a module main? (i'm kinda interested in just making sure its easy to check / validate them?)

maoe:
+1 to this feature. I also would like to be able to set `verbose: 0` to suppress build messages.

fendor:
@23Skidoo 
> Is it documented somewhere which `script-spec-version` is the implicit one?

According to the code:
https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/CmdRun.hs#L473
It looks like `2.2`

[...]

PR #6310, opened 2019-10-26
Add with-compiler field for cabal scripts

bubba:
Adds a new data type and field grammar for parsing script metadata.
Now it is possible to have scripts like:

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends: base
with-compiler: ghc-8.6.5
-}
main = putStrLn "hey"
```

Progresses towards #5698

---

### DIFF
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -555,3 +555,4 @@ scriptFieldGrammar = Script
     <*> optionalFieldAla "with-compiler" FilePathNT L.hcPath
 {-# SPECIALIZE scriptFieldGrammar :: ParsecFieldGrammar' Script #-}
 {-# SPECIALIZE scriptFieldGrammar :: PrettyFieldGrammar' Script #-}
+
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -37,6 +37,8 @@ module Distribution.PackageDescription.FieldGrammar (
     setupBInfoFieldGrammar,
     -- * Component build info
     buildInfoFieldGrammar,
+    -- * Script
+    scriptFieldGrammar
     ) where
 
 import Distribution.Compat.Lens
@@ -57,6 +59,7 @@ import Distribution.Types.ExecutableScope
 import Distribution.Types.ForeignLib
 import Distribution.Types.ForeignLibType
 import Distribution.Types.LibraryVisibility
+import Distribution.Types.Script
 import Distribution.Types.UnqualComponentName
 import Distribution.Version                   (anyVersion)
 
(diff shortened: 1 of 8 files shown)


Cabal/doc/nix-local-build.rst
```diff
@@ -462,6 +462,7 @@ a script that looks like:
     {- cabal:
     build-depends: base ^>= 4.11
                 , shelly ^>= 1.8.1
+    with-compiler: ghc-8.8.1
```
DanielG: This makes it sound like with-compiler is mandatory now. Might it be better to just have another paragraph here that mentions which fields are allowed?

Cabal/Distribution/PackageDescription/FieldGrammar.hs
```diff
@@ -539,3 +542,16 @@ setupBInfoFieldGrammar def = flip SetupBuildInfo def
...
+-------------------------------------------------------------------------------
+
+scriptFieldGrammar
+    :: (FieldGrammar g, Applicative (g Script), Applicative (g Executable), Applicative (g BuildInfo))
+    => g Script Script
+scriptFieldGrammar = Script
+    <$> blurFieldGrammar L.executable (executableFieldGrammar "script")
+    <*> optionalFieldAla "with-compiler" FilePathNT L.hcPath
+{-# SPECIALIZE scriptFieldGrammar :: ParsecFieldGrammar' Script #-}
+{-# SPECIALIZE scriptFieldGrammar :: PrettyFieldGrammar' Script #-}
```
DanielG: I don't really get why this pragma is used twice here. This does seem to be the case in the rest of the definitions above as well though, weird.

DanielG: Oh, I can't read. The first one is for "Parsec" and the other one for "Pretty". nvm.

phadej: @DanielG indeed, I'm trying had to name dual things with nicely alignable names! :)

phadej: There's no EOL newline. Not biggie, but if you can configure your editor for this project, it would be nice next time!

bubba:
@phadej @DanielG ping

hvr:
At this point, we should really start "versioning" the "script spec format" and make use of the version field right after the `cabal:` marker, e.g.

```
{- cabal: 3.2
build-depends: base ^>= { 4.12.0, 4.13.0 }
with-compiler: ghc-8.6.5
-}
```

phadej:
I’d use `cabal-version: 3.2` everywhere. We have machinery for that, also finding static`cabal:` marker is easier.

Motivation: less ad-hoc syntax/grammar, even it’s slightly more verbose

mpickering:
I would also like to be able to specify the compiler to use on the command line. For example

```
cabal File.hs -wghc-8.6.5
```

It seems the option is currently just ignored?

hvr:
> I would also like to be able to specify the compiler to use on the command line. For example
> 
> ```
> cabal File.hs -wghc-8.6.5
> ```
> 
> It seems the option is currently just ignored?

@mpickering It's *not* ignored; you're invoking the special shebang interface;

If you had used the normal designated CLI invocation for cabal scripts via `cabal v2-run -wghc-8.6.5 File.hs` you would have been fine.

Consider this simple script:

```
#! /usr/bin/env cabal

{- cabal:
build-depends: base
-}

import System.Environment

main = print =<< getArgs
```

Then,  compare the outputs of the three invocations:

```
$ cabal ./foo.hs -wghc-7.4.2 -- --help

["-wghc-7.4.2","--","--help"]

$ cabal v2-run ./foo.hs -wghc-7.4.2 -- --help

["--help"]

$ ./foo.hs -wghc-7.4.2 -- --help

["-wghc-7.4.2","--","--help"]
```

Does it make sense now?

mpickering:
Thanks. I did not know you could use `v2-run` with an executable script.

phadej:
I think @hvr raised the valid concern of need to versioning. I'll assign this to myself and merge after 3.0.1.0 and 3.2.0.0 dust is settled and add `cabal` versioning to this. If we version libraries, we definitely should version the tool that reads the definitions, as they will evolve (and possibly in non backward compatible ways).

gbaz:
@phadej is this something you still intended to get to?

gbaz:
is this superceded by https://github.com/haskell/cabal/pull/7997 ?

Mikolaj:
In that case, let me close. 

@bubba: thank you for the PR and apologies that it dragged on for so long. If you think there's anything in the PR that can get rebased, scavenged and put to good use, please kindly open a new PR and we'll do our best not to lose it again.

PR #6311, opened 2019-10-27
Start GHCi with the main module in scope

### LINKED ISSUE #5103: new-repl on test-suite does not bring Main module into scope
Steps to  reproduce:

```
$ git clone -b cabal-new-build https://github.com/hspec/hspec-example
$ cd hspec-example/
$ cabal new-repl spec
*Data.String.StripSpec> 
```

This will result in `Data.String.StripSpec` being in scope.

My naive assumption was that `Main` should be in scope.  Is there a rational for the current behavior?

dwijnand:

Fixes #5103.

I tested this change by locally building cabal-install, running it in a project whose `test-suite` has `other-modules` and asserting that the main module was in scope in the REPL.  I would love to make that a test, but I would need a little guidance (and/or prior art).

I'm also not sure if this is changelog worthy and/or if there is documentation to change or add to.  Please advise as I'd be happy to add these.

More generally, this is my first cabal PR (edit: nvm, second, after #5943 🙂) and I'm still ramping up in Haskell, so any feedback is welcome.

### DIFF
--- a/Cabal/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/Distribution/Simple/Program/GHC.hs
@@ -745,8 +745,10 @@ renderGhcOptions comp _platform@(Platform _arch os) opts
   ---------------
   -- Inputs
 
-  , [ prettyShow modu | modu <- flags ghcOptInputModules ]
+  -- Specify the input file(s) first, so that in ghci the `main-is` module is
+  -- in scope instead of the first module defined in `other-modules`.
   , flags ghcOptInputFiles
+  , [ prettyShow modu | modu <- flags ghcOptInputModules ]
 
   , concat [ [ "-o",    out] | out <- flag ghcOptOutputFile ]
   , concat [ [ "-dyno", out] | out <- flag ghcOptOutputDynFile ]


Cabal/Distribution/Simple/Program/GHC.hs
```diff
@@ -745,8 +745,8 @@ renderGhcOptions comp _platform@(Platform _arch os) opts
   ---------------
   -- Inputs
 
-  , [ prettyShow modu | modu <- flags ghcOptInputModules ]
   , flags ghcOptInputFiles
+  , [ prettyShow modu | modu <- flags ghcOptInputModules ]
```
phadej: I'm really blid, but I don't see any difference even GitHub is highlighting it to me?

dwijnand: The change is in the ordering.  By loading the main module first (the input file) the ghci session has that module in scope, rather than the first module defined in `other-modules`.

So for instance with this change and a test-suite of

```
  main-is:             Main.hs
  other-modules:
      FooSpec
```

then `Main` will be in scope rather than `FooSpec`.

phadej: I see now. This is so subtle, we need a comment here.

dwijnand: Oh, I agree.  I can do a comment but I'd prefer a test the captures the over-arching desire.

Are there any existing tests that make assertions on `cabal ghci`'s behaviour?

phadej: I'm afraid we don't have any, as testing interactive behaviour is tricky.

dwijnand: Hmm, what about testing to the point of asserting how ghc is invoked?  `renderGhcOptions` is just seems too generic removed from "I need ghci to be invoked this way".  But maybe I should just test around that?

dwijnand: Hmm, what about testing to the point of asserting how ghc is invoked?  `renderGhcOptions` is just seems too generic removed from "I need ghci to be invoked this way".  But maybe I should just test around that?

phadej: Generally, having being able to write test "`cabal repl` loads my package" would be great, but it's not a blocker for this PR.

phadej:
Thanks!

PR #6312, opened 2019-10-27
.travis.yml: Allow forks to build

dwijnand:
But then exclude builds in forks from the notifications.

In terms of testing, the `if` condition I'm fairly confident about, as it's a simpler version of a version I use at work.  Also it triggered https://travis-ci.com/dwijnand/cabal/builds/133725400 to build.

Regarding the notification changes, I believe them to be correct, using https://docs.travis-ci.com/user/notifications#conditional-notifications as reference.  However, using https://config.travis-ci.com/explore it complains that:

    [alert] on notifications.slack.rooms: expected an encrypted string

So perhaps someone should double-check that builds are still notified if this PR lands.

(Btw, I'm getting a "Failed! token_revoked" error message when trying to use https://haskell-cabal.herokuapp.com/.  May I join the Slack team some other way?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,9 @@ dist: trusty
 sudo: true
 
 # We whitelist branches, as we don't really need to build dev-branches.
+# Except if those branches are in a fork.
 # Remember to add release branches, both here and to appveyor.yml.
-branches:
-  only:
-    - master
-    - "3.0"
-    - "2.4"
-    - "2.2"
-    - "2.0"
-    - "1.24"
-    - "1.22"
-    - "1.20"
-    - "1.18"
+if: branch IN (master, 3.0, 2.4, 2.2, 2.0, 1.24, 1.22, 1.20, 1.18) OR repo != haskell/cabal
 
 # The following enables several GHC versions to be tested; often it's enough to
 # test only against the last release in a major GHC version. Feel free to omit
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -177,6 +168,9 @@ after_success:
 
 notifications:
   irc:
+    if: repo = haskell/cabal
```
phadej: are this conditionals documented somewhere, some recent Travis functionality? (single equals is confusing!)

dwijnand: Yep, https://docs.travis-ci.com/user/notifications#conditional-notifications mentions them in the context of notifications and https://docs.travis-ci.com/user/conditions-v1 goes more in detail on them.

phadej: Nice, and "oh dear, soon .travis.yml will be turing complete (if it's not already)"

phadej: It would be nice to be able to say

```yaml
script:
  - if: os = linux
    then: "bash bla bla"
  - do this on osx and linux still
```

But I guess we cannot. (it's not so useful for `Cabal`, but elsewhere).

dwijnand: I think you should be able to do that with the `TRAVIS_OS_NAME` env var.  So something like

```yaml
script:
  - [ "$TRAVIS_OS_NAME" = "linux" ] && bla bla
  - do this on osx and linux still
```

phadej: Yes, that what I do now; it gets tedious :)

PR #6313, opened 2019-10-27
Mention "password-command" as a config file option

### LINKED ISSUE #5224: cabal.config should have a field for password-command.
Currently, according to `cabal upload --help` it's possible to specify a username and password in the cabal.config file. Of course, storing passwords in config files is a horrible idea, ESPECIALLY when one wants to share/publish those. Thankfully `cabal upload` has the helpful `--password-command` flag to read the password from the stdout of another application, however as far as I know this password-command can't be specified in cabal.config?

dwijnand:

Fixes #5224 

I tested this by building cabal-install and looking at the output of `upload --help`.  On it's own, I'm not sure if that's worth wiring up a test for.  Perhaps the output of `upload --help` as a whole could be tested?  WDYT?

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -2189,7 +2189,7 @@ uploadCommand = CommandUI {
     commandDescription  = Nothing,
     commandNotes        = Just $ \_ ->
          "You can store your Hackage login in the ~/.cabal/config file\n"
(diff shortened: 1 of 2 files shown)


hvr:
@dwijnand tbh, I think it's overkill to add a test for such a trival UI aspect; however, a brief mention in `cabal-install`'s changelog pointing to #5224 would be desirable

PR #6315, opened 2019-10-27
Reduce CI matrix for PRs

### LINKED ISSUE #4384: Proposal: Reduce CI matrix for PRs
At the moment, we run builds under 14 different environments when we do CI. I have noticed that for the vast majority of commits, this is wasted effort. Usually, the important things are to (1) test every GHC version, (2) test Windows, and (3) test Mac OS X.

So, I propose to reduce the set of configurations we test during PRs to Linux (once for every GHC version), Windows, and the latest GHC with Mac OS X. We'll continue to run all tests for commits to master, and we should have some mechanism for triggering the extra tests in PR if a user wants them (e.g., a high risk commit.)

dwijnand:
Fixes #4384

The ticket (#4384) mentions:

> we should have some mechanism for triggering the extra tests in PR if a user wants them (e.g., a high risk commit.)

I'm thinking that for high risk commits perhaps calling the API to trigger an "api"-type build (as opposed to a "pull_request"-type build) would run the additional jobs.  WDYT?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,15 +69,19 @@ matrix:
    - env: GHCVER=8.4.4 SCRIPT=solver-debug-flags USE_GOLD=YES
      sudo: required
      os: linux
+     if: type != pull_request
    - env: GHCVER=8.4.4 SCRIPT=script DEBUG_EXPENSIVE_ASSERTIONS=YES TAGSUFFIX="-fdebug-expensive-assertions" USE_GOLD=YES
      os: linux
      sudo: required
+     if: type != pull_request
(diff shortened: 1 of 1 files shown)


phadej:
No way. It takes time, but CI catches problems especially with older GHCs.

phadej:
EDIT: to expand. The `master` have been red for extensive amount of time, and **no one** had time to fix it. I won't make it any easier to accidentally turn red again.

Test locally, don't spam Travis.

dwijnand:
Make sense.  Then I take it #4384 should closed.  I'll see if I'm able to help by fixing master.

PR #6320, opened 2019-10-30
Use Data.List.NonEmpty and avoid some partial functions

dwijnand:
(Rebased & fixed resubmission of #6142.)

Add safe alternatives to some partial functions in Distribution.Utils.Generic, and removes the originals from Distribution.Compat.Prelude. See #6107.

Uncomment versions of functions in Distribution.Compat.Parsing that return non-empty lists.

There are still many partial functions in use like `maximum` after this PR, and there are a few uses of `head`, `last`, `tail` and `init` that still remain because I didn't find any easy way to remove them.

Tested by running `cabal new-run cabal-tests`, `cabal new-test Cabal:unit-tests`, etc.
(Also tested by running the PR validation in a fork.)

### DIFF
--- a/Cabal/Distribution/Compat/Parsing.hs
+++ b/Cabal/Distribution/Compat/Parsing.hs
@@ -25,12 +25,9 @@ module Distribution.Compat.Parsing
   , some     -- from Control.Applicative, parsec many1
   , many     -- from Control.Applicative
   , sepBy
-  , sepBy1
   , sepByNonEmpty
-  , sepEndBy1
   , sepEndByNonEmpty
   , sepEndBy
-  , endBy1
   , endByNonEmpty
   , endBy
   , count
@@ -97,26 +94,14 @@ between bra ket p = bra *> p <* ket
 --
 -- >  commaSep p  = p `sepBy` (symbol ",")
 sepBy :: Alternative m => m a -> m sep -> m [a]
-sepBy p sep = sepBy1 p sep <|> pure []
+sepBy p sep = toList <$> sepByNonEmpty p sep <|> pure []
 {-# INLINE sepBy #-}
-
--- | @sepBy1 p sep@ parses /one/ or more occurrences of @p@, separated
--- by @sep@. Returns a list of values returned by @p@.
-sepBy1 :: Alternative m => m a -> m sep -> m [a]
-sepBy1 p sep = toList <$> sepByNonEmpty p sep
-{-# INLINE sepBy1 #-}
 
 -- | @sepByNonEmpty p sep@ parses /one/ or more occurrences of @p@, separated
 -- by @sep@. Returns a non-empty list of values returned by @p@.
 sepByNonEmpty :: Alternative m => m a -> m sep -> m (NonEmpty a)
 sepByNonEmpty p sep = (:|) <$> p <*> many (sep *> p)
 {-# INLINE sepByNonEmpty #-}
-
--- | @sepEndBy1 p sep@ parses /one/ or more occurrences of @p@,
--- separated and optionally ended by @sep@. Returns a list of values
--- returned by @p@.
-sepEndBy1 :: Alternative m => m a -> m sep -> m [a]
-sepEndBy1 p sep = toList <$> sepEndByNonEmpty p sep
 
 -- | @sepEndByNonEmpty p sep@ parses /one/ or more occurrences of @p@,
 -- separated and optionally ended by @sep@. Returns a non-empty list of values
@@ -130,14 +115,8 @@ sepEndByNonEmpty p sep = (:|) <$> p <*> ((sep *> sepEndBy p sep) <|> pure [])
 --
 -- >  haskellStatements  = haskellStatement `sepEndBy` semi
 sepEndBy :: Alternative m => m a -> m sep -> m [a]
-sepEndBy p sep = sepEndBy1 p sep <|> pure []
+sepEndBy p sep = toList <$> sepEndByNonEmpty p sep <|> pure []
 {-# INLINE sepEndBy #-}
-
--- | @endBy1 p sep@ parses /one/ or more occurrences of @p@, separated
--- and ended by @sep@. Returns a list of values returned by @p@.
-endBy1 :: Alternative m => m a -> m sep -> m [a]
-endBy1 p sep = some (p <* sep)
-{-# INLINE endBy1 #-}
 
 -- | @endByNonEmpty p sep@ parses /one/ or more occurrences of @p@, separated
 -- and ended by @sep@. Returns a non-empty list of values returned by @p@.
--- a/Cabal/Distribution/Compat/Parsing.hs
+++ b/Cabal/Distribution/Compat/Parsing.hs
@@ -28,10 +28,10 @@ module Distribution.Compat.Parsing
   , sepBy1
   , sepByNonEmpty
   , sepEndBy1
-  -- , sepEndByNonEmpty
+  , sepEndByNonEmpty
   , sepEndBy
   , endBy1
-  -- , endByNonEmpty
+  , endByNonEmpty
   , endBy
   , count
   , chainl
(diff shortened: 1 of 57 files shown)


hvr:
Thanks for picking up @edmundnoble's PR!

Cabal/Distribution/Simple/Glob.hs
```diff
@@ -121,7 +127,7 @@ data Glob
 
 data GlobFinal
   = FinalMatch IsRecursive MultiDot String
-    -- ^ First argument: Is this a @**/*.ext@ pattern?
+    -- ^ First argument: Is this a @** / *.ext@ pattern?
```
dwijnand: I had to change this to avoid CPP interpreting `/*` as a comment and failing with an 'unterminated comment' error.  I couldn't find any way to escape it or something, so I just went with this.

hvr: Ideally, that file has no CPP at all; in fact, now that I see it, the conditional import of `Data.List.NonEmpty` looks supicous -- that's something that ought to be moved into the Compat.Prelude module which might make the use of CPP redundant

Cabal/Distribution/Compat/Parsing.hs
```diff
@@ -102,8 +108,7 @@ sepBy p sep = sepBy1 p sep <|> pure []
 -- | @sepBy1 p sep@ parses /one/ or more occurrences of @p@, separated
 -- by @sep@. Returns a list of values returned by @p@.
 sepBy1 :: Alternative m => m a -> m sep -> m [a]
-sepBy1 p sep = (:) <$> p <*> many (sep *> p)
```
phadej: FWIW, we should probably justt remove `*By1` functions. I don't recall they are really used anywhere (or then the `toList` can be put there).

Doesn't need to be done, if you don't feel like that.

Cabal/Distribution/FieldGrammar/Parsec.hs
```diff
@@ -155,37 +160,37 @@ instance FieldGrammar ParsecFieldGrammar where
...
-            Nothing -> parseFatalFailure zeroPos $ show fn ++ " field missing"
-            Just [] -> parseFatalFailure zeroPos $ show fn ++ " field missing"
-            Just [x] -> parseOne v x
-            Just xs -> do
-                warnMultipleSingularFields fn xs
-                last <$> traverse (parseOne v) xs
+            Nothing     -> parseFatalFailure zeroPos $ show fn ++ " field missing"
+            Just []     -> parseFatalFailure zeroPos $ show fn ++ " field missing"
+            Just [x]    -> parseOne v x
+            Just (x:xs) -> do
```
phadej:
```
(_:x:xs) ->
```

The first one is definitely not the last one here! :) (I don't know why i didn't `last xs`

Cabal/Distribution/Simple/GHCJS.hs
```diff
@@ -72,6 +72,9 @@ import Distribution.Types.ForeignLib
 import Distribution.Types.ForeignLibType
 import Distribution.Types.ForeignLibOption
 import Distribution.Types.UnqualComponentName
+#if MIN_VERSION_base(4,9,0)
```
phadej: why this is `#if`d?

Cabal/Distribution/Simple/Glob.hs
```diff
@@ -29,6 +30,9 @@ import Prelude ()
 import Distribution.Compat.Prelude
 
 import Control.Monad (guard)
+#if MIN_VERSION_base(4,9,0)
+import Data.List.NonEmpty (NonEmpty(..))
```
phadej: Let's re-export `NonEmpty` from `Distribution.Compat.Prelude`

Cabal/Distribution/Types/MungedPackageName.hs
```diff
@@ -137,7 +137,7 @@ zdashcode s = go s (Nothing :: Maybe Int) []
 
 parseZDashCode :: CabalParsing m => m [String]
 parseZDashCode = do
-    ns <- P.sepBy1 (some (P.satisfy (/= '-'))) (P.char '-')
+    ns <- toList <$> P.sepBy1 (some (P.satisfy (/= '-'))) (P.char '-')
```
phadej: should there be `P.sepByNonEmpty` or no `toList`?

cabal-install/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -609,12 +614,12 @@ resolveTargets selectPackageTargets selectComponentTarget liftProblem
                            -> [AvailableTarget k]
                            -> Either err [k]
     selectComponentTargets subtarget =
-        either (Left . head) Right
+        either (Left . NonEmpty.head) Right
```
phadej: I'm leaning towards re-exporing `NonEmpty.head` and `last`  from `Compat.Prelude`

@hvr, opinions?

hvr: @phadej I'm sympathetic to that :-)

phadej: Let's do that afterwards.

Cabal/Distribution/FieldGrammar/Parsec.hs
dwijnand: Don't I need to keep the first one to emit the warning?  I guess I can drop the `x` in `x:|xs`.

phadej: Ah true, then it will be

```
Just xs@(_:y:ys) -> do
    warnMultipleSingularFields fn xs
    NonEmpty.last <$> traverse (parseOne v) (y:|ys)

File: Cabal/Distribution/Types/MungedPackageName.hs (hunk 6 above)
Comment (dwijnand) [reply Cabal/Distribution/Types/MungedPackageName.hs]: Yep, the `toList` here is redundant.

File: cabal-install/Distribution/Client/ProjectOrchestration.hs (hunk 7 above)
Comment (dwijnand) [reply cabal-install/Distribution/Client/ProjectOrchestration.hs]: Happy to do so, here or as a follow-up.

Where moving to `NonEmpty`'s head/tail/init/last isn't feasible, should the partial versions be scoped?  Otherwise, you'd have to refer to the imports to see which one is being brought into scope and which is being hidden.

File: Cabal/Distribution/Simple/GHCJS.hs (hunk 4 above)
Comment (dwijnand) [reply Cabal/Distribution/Simple/GHCJS.hs]: To make it work on 7.x versions of the compiler, for unclear reasons.  My guess now is that Edmund initially started importing Data.List.NonEmpty per file and then later added the re-export in Cabal's Prelude, so these are redundant.

File: Cabal/Distribution/Compat/Parsing.hs (hunk 2 above)
Comment (dwijnand) [reply Cabal/Distribution/Compat/Parsing.hs]: Yep, looks like it's only a few instances.

Personally I feel like the `*By1` naming conventions is nicer than trailing with `NonEmpty`, but I've gone for with it for now.

File: Cabal/Distribution/Simple/PreProcess/Unlit.hs [hunk 8]
```diff
@@ -35,10 +36,10 @@ classify ('>':s) = BirdTrack s
 classify ('#':s) = case tokens s of
                      (line:file:_) | all isDigit line
```
hvr:
```suggestion
                     (line:file@('"':_:_)_) | all isDigit line
```

fwiw, if we had the pattern match above, we could omit the inefficient `length`-check as well as the 'safeHead` call

cabal-install/Distribution/Client/ProjectOrchestration.hs
phadej: then one can

```haskell
-- imports
import Distribution.Compat.Prelude

import qualified Prelude -- add that if you need list head

-- usage

foo = head xs -- NonEmpty variant
bar = Prelude.head ys -- List variant
```

Issue #6011, opened 2019-04-23
build-tool-depends with custom setup tries to build the wrong component


bubba:
- The test case contains a sub-package `foo` that exposes an executable called `bar`
- The parent package has a `build-tool-depends` on `foo:bar`
- `foo` also has `build-type: Custom`
- During `cabal new-build`, cabal tries to build `exe:foo` when it should build `exe:bar`

[cabal-subexe-bug.tar.gz](https://github.com/haskell/cabal/files/3109995/cabal-subexe-bug.tar.gz)

```bash
$ cabal new-build
Resolving dependencies...
Build profile: -w ghc-8.6.4 -O1
In order, the following will be built (use -v for more details):
 - foo-0.1.0.0 (lib:foo, exe:foo) (configuration changed)
 - cabal-subexe-bug-0.1.0.0 (lib) (configuration changed)
 - cabal-subexe-bug-0.1.0.0 (lib) (configuration changed)
Configuring foo-0.1.0.0...
setup: Unknown build target 'exe:foo'.
There is no executable component 'foo' or component 'exe'.

cabal: Failed to build foo-0.1.0.0 (which is required by
cabal-subexe-bug-0.1.0.0).
```

23Skidoo:
Thanks for providing a repro case!

PR #6323, opened 2019-10-31
Add test for build tool depends with per-component builds disabled

alex-rozenshteyn-leapyear:
This seems to be the underlying issue in #6011

I'm not sure how to fix it (yet), but I'm hoping to solicit suggestions. My best guess is that the bug is in `Distribution.Client.ProjectPlanning`.

---

### DIFF
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
@@ -1,0 +1,4 @@
+import Test.Cabal.Prelude
+-- Test build-tool-depends between two packages
+main = cabalTest $ do
+    cabal "v2-build" ["client", "--disable-per-component"]
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/cabal.project
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/cabal.project
@@ -1,0 +1,2 @@
+packages: client
+optional-packages: pre-proc
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/client/Hello.hs
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/client/Hello.hs
@@ -1,0 +1,8 @@
+{-# OPTIONS_GHC -F -pgmF zero-to-one #-}
+module Main where
+
+a :: String
+a = "0000"
+
+main :: IO ()
+main = putStrLn a
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/client/client.cabal
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/client/client.cabal
@@ -1,0 +1,13 @@
+name:                client
+version:             0.1.0.0
+synopsis:            Checks build-tool-depends are put in PATH
+license:             BSD3
+category:            Testing
+build-type:          Simple
+cabal-version:       >=1.10
+
+executable             hello-world
+  main-is:             Hello.hs
+  build-depends:       base
+  build-tool-depends:  pre-proc:zero-to-one
+  default-language:    Haskell2010
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/pre-proc/MyCustomPreprocessor.hs
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/pre-proc/MyCustomPreprocessor.hs
@@ -1,0 +1,11 @@
+module Main where
+
+import System.Environment
+import System.IO
+
+main :: IO ()
+main = do
+  (_:source:target:_) <- getArgs
+  let f '0' = '1'
+      f c = c
+  writeFile target . map f  =<< readFile source
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/pre-proc/pre-proc.cabal
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/pre-proc/pre-proc.cabal
@@ -1,0 +1,17 @@
+name:                pre-proc
+version:             999.999.999
+synopsis:            Checks build-tool-depends are put in PATH
+license:             BSD3
+category:            Testing
+build-type:          Simple
+cabal-version:       >=1.10
+
+executable             zero-to-one
+  main-is:             MyCustomPreprocessor.hs
+  build-depends:       base, directory
+  default-language:    Haskell2010
+
+executable             bad-do-not-build-me
+  main-is:             MyMissingPreprocessor.hs
+  build-depends:       base, directory
+  default-language:    Haskell2010
--- a/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.out
+++ b/cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.out
@@ -1,0 +1,12 @@
+# cabal v2-build
+Resolving dependencies...
+Build profile: -w ghc-<GHCVER> -O1
+In order, the following will be built:
+ - pre-proc-999.999.999 (exe:zero-to-one) (first run)
+ - client-0.1.0.0 (exe:hello-world) (first run)
+Configuring executable 'zero-to-one' for pre-proc-999.999.999..
+Preprocessing executable 'zero-to-one' for pre-proc-999.999.999..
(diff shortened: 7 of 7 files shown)


cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
```diff
@@ -0,0 +1,4 @@
+import Test.Cabal.Prelude
+-- Test build-tool-depends between two packages
+main = cabalTest $ do
+    cabal "v2-build" ["client", "--disable-per-component"]
```
phadej: Even the test fails now, it will still fail as this flag says that `bad-do-not-build-me` have to be built too.

Will this test still fail as expected, if `bad-do-not-build-me` executable is removed?

cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
```diff
@@ -0,0 +1,4 @@
+import Test.Cabal.Prelude
+-- Test build-tool-depends between two packages
```
phadej: Could you name the test file `cabal.test.hs`, as it's testing `cabal-install` not `Setup`.

Also we have issue tests as e.g. `./PackageTests/Regression/T4202/cabal.test.hs` but I have no strong opinion on that.

cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
alex-rozenshteyn-leapyear: Yes. Removing the `bad-do-not-build-me` executable results in the following:

```
% cabal new-clean && cabal new-build client --disable-per-component
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
 - pre-proc-999.999.999 (exe:pre-proc) (first run)
 - client-0.1.0.0 (exe:hello-world) (first run)
Configuring pre-proc-999.999.999...
cabal: Unknown build target 'exe:pre-proc'.
There is no executable component 'pre-proc'.

cabal: Failed to build pre-proc-999.999.999 (which is required by
client-0.1.0.0).
```

I will make the change.

cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
alex-rozenshteyn-leapyear: I will rename the test file and move it to issue tests.

alex-rozenshteyn-leapyear:
I'm a little stuck trying to fix the bug.

For some reason, with `--disable-per-component`, `exe:pre-proc` is passed to setup: 
```
Using self-exec internal setup method with build-type Simple and args:
["act-as-setup","--build-type=Simple","--","build","--verbose=3","--builddir=.../dist-newstyle/build/x86_64-linux/ghc-8.6.5/pre-proc-999.999.999","exe:pre-proc"]
```

It seems that [`external_cc_map`](https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/ProjectPlanning.hs#L1534-L1535) in `Distribution.Client.ProjectPlanning` has an entry of

```
( PackageName "pre-proc" 
, fromList 
    [ 
        ( CExeName ( UnqualComponentName "zero-to-one" )
        , AnnotatedId 
            { ann_pid = PackageIdentifier 
                { pkgName = PackageName "pre-proc" 
                , pkgVersion = mkVersion 
                    [ 999
                    , 999   
                    , 999   
                    ]       
                }       
            , ann_cname = CExeName ( UnqualComponentName "zero-to-one" )
            , ann_id = ComponentId "pre-proc-999.999.999-inplace" 
            }
        )       
    ]
) 
```

The `ComponentId` looks suspicious to me, but that could be a result of me not understanding what's supposed to be happening.

Comparing the `plan.json` files with and without `--disable-per-component` also might yield some insight.

Thoughts?

phadej:
`-inplace` in component id  looks correct, it's "package-name-version-inplace" for "local" packages

alex-rozenshteyn-leapyear:
> `-inplace` in component id looks correct, it's "package-name-version-inplace" for "local" packages

So how does the `build` step get its parameters? It seems to be getting an `exe:[component name]` parameter, which is correct for per-component builds, but not for per-package builds if the component name is the package name.

alex-rozenshteyn-leapyear:
Found the source of the problem (turns out there was a TODO about it). I fixed it, but I'm not sure if the fix is the right way to do it. In a per-package build, I explicitly added all executables of a package to the build dependencies if any of the executables was a dependency.

alex-rozenshteyn-leapyear:
Note that the single failure seems to be unrelated (https://travis-ci.org/haskell/cabal/jobs/606495569):
```
$ Installing caching utilities
/home/travis/.rvm/scripts/cli: line 872:  4372 Segmentation fault      "$rvm_scripts_path/list" "${rvm_ruby_args[@]}"
```

alex-rozenshteyn-leapyear:
@phadej I've been dogfooding the this patch, and it seems to work. CI passed except for an unrelated failure. Is there anything else I need to do to get this merged?

alex-rozenshteyn-leapyear:
Ping. Is there any action needed on my part to help merge this in?

phadej:
Sorry, there were a bit of hassle in preparing the release branches. Could you rebase to the current master (the CI setup should be more robust there)

cabal-testsuite/PackageTests/Regression/T6011/pre-proc/pre-proc.cabal
```diff
@@ -0,0 +1,12 @@
+name:                pre-proc
+version:             999.999.999
+synopsis:            Checks build-tool-depends are put in PATH
+license:             BSD3
+category:            Testing
+build-type:          Simple
```
phadej: This is not `Custom`, so it's not what described in #6011

I'll check what happens with non-Custom builds, there we should be able to build just an executable (of possibly multiple in the package). I don't defaulting to build every exe, I'd expect someone to complain about that too.

cabal-testsuite/PackageTests/BuildToolDependsNonComponent/setup.test.hs
phadej: Ok, now I see why there was second executable. We should not build it. Argh, this is tricky. I'll take this over to see how to solve this without compromises.

cabal-testsuite/PackageTests/Regression/T6011/cabal.test.hs
```diff
@@ -0,0 +1,4 @@
+import Test.Cabal.Prelude
+-- Test build-tool-depends between two packages
+main = cabalTest $ do
+    cabal "v2-build" ["client", "--disable-per-component"]
```
phadej: Looking at this, we probably should do something special in `ProjectPlanning` based on whether per-component build is disabled (either by flag or `build-type: Custom`). We **should not** default to do more in all scenarios just in case if per component build is disabled: everyone would pay for a corner-case.

cabal-testsuite/PackageTests/Regression/T6011/pre-proc/pre-proc.cabal
alex-rozenshteyn-leapyear: Correct. It's closely related but not a direct reproduction. It _appears_ to be the case that custom builds are not the core issue, but the fact that custom turns off per-component. Whether or not it effects custom builds, this is an issue.

cabal-testsuite/PackageTests/Regression/T6011/cabal.test.hs
alex-rozenshteyn-leapyear: Could you elaborate on this? What would the "something special" look like? In what sense is everyone paying for the corner case in the current situation?

alex-rozenshteyn-leapyear:
> Sorry, there were a bit of hassle in preparing the release branches. Could you rebase to the current master (the CI setup should be more robust there)

No worries; I understand release-related stresses. Since you're taking over the issue, I'll leave the rebasing to you to avoid conflicting?

alex-rozenshteyn-leapyear:
@phadej Is any work being done on this?

phadej:
Sorry, I have paged out the details already. I'll come back to this eventually.

alex-rozenshteyn-leapyear:
@phadej No worries!

Kleidukos:

ffaf1:

PR #6325, opened 2019-11-01
CI: Add travis_retry when uploading binaries to S3

### LINKED ISSUE #6324: CI: Could not resolve host: haskell-cabal.s3.amazonaws.com
@phadej mentioned in #6315 that master has been broken for a number of days.  I'm looking at the latest build https://travis-ci.org/haskell/cabal/builds/601490658 and 8/16 jobs are due to failing to upload to S3 the binaries to test, due to:

    curl: (6) Could not resolve host: haskell-cabal.s3.amazonaws.com

There's also a 9th failure of "Could not resolve host: hackage.haskell.org", which I've also seen in the past in other CI jobs.

Does anyone know what might've happened?

I noticed the `travis_retry` usage just lower down in `travis/upload.sh` so I'm tempted to PR adding that to the curl.  Would that be appreciated?

dwijnand:
Fixes #6324

Tested in my fork.

### DIFF
--- a/travis/upload.sh
+++ b/travis/upload.sh
@@ -86,7 +86,7 @@ rm $BINARIES # Don't check me in!
 
 # Upload to S3
 S3_URL=$(curl -X POST "https://s3-bouncer.herokuapp.com/put")
-curl "$S3_URL" --upload-file binaries.tgz
+travis_retry curl "$S3_URL" --upload-file binaries.tgz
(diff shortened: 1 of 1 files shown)


dwijnand:
Ironically failed elsewhere due to "curl: (56) LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54" 😄

PR #6326, opened 2019-11-02
Remove Slack notifications

hvr:
They don't seem to have worked for a long time nor is there a relevant presence of Cabal devs on Slack.


hvr:
It might make more sense to enable notifications for say Discord; we also still have notifications on IRC (via `freenode/##haskell-cabal`) enabled (I'm currently testing whether they still work).

PR #6329, opened 2019-11-05
Re-enable regenerating the AUTHORS file

dwijnand:

No testing done.

### DIFF
--- a/travis-meta.sh
+++ b/travis-meta.sh
@@ -6,9 +6,9 @@
 # Check that auto-generated files/fields are up to date.
 # ---------------------------------------------------------------------
 
-# Regenerate the CONTRIBUTORS file.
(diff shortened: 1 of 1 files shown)


phadej:
@23Skidoo wouldn't checking this in meta make builds fail each time there's new contributor? I think it's not nice to force new contributors to do some extra stuff, satisfying our CI is quite a task already.

dwijnand:
Given it's easily automatable I'd agree that it's probably best move this from PR validation to being part of the releasing process.

I'll leave this open a second to let @23Skidoo weigh in.

23Skidoo:
Yeah, I think it makes more sense for this to be a part of the release process. Also the script that generates `AUTHORS` is not complete currently, the file also needs to be formatted by hand in Emacs.

PR #6330, opened 2019-11-06
Replace Prelude.head/tail/init/last with NonEmpty

dwijnand:

Tested my branch in my fork first.  But even though it passes CI, please double-check my refactorings as I found some mistakes in my own work while reviewing it myself.

Refs #6107
Refs https://github.com/haskell/cabal/pull/6320#discussion_r341719136

### DIFF
--- a/Cabal/Distribution/Simple.hs
+++ b/Cabal/Distribution/Simple.hs
@@ -57,7 +57,7 @@ module Distribution.Simple (
 
 import Control.Exception (try)
 
-import Prelude (head)
+import Prelude ()
 import Distribution.Compat.Prelude
 
 -- local
@@ -524,9 +524,9 @@ sanityCheckHookedBuildInfo verbosity
       ++ "but the package does not have a library."
 
 sanityCheckHookedBuildInfo verbosity pkg_descr (_, hookExes)
-    | not (null nonExistant)
+    | exe1 : _ <- nonExistant
     = die' verbosity $ "The buildinfo contains info for an executable called '"
-      ++ prettyShow (head nonExistant) ++ "' but the package does not have a "
+      ++ prettyShow exe1 ++ "' but the package does not have a "
       ++ "executable with that name."
   where
     pkgExeNames  = nub (map exeName (executables pkg_descr))
--- a/Cabal/Distribution/Simple/Configure.hs
+++ b/Cabal/Distribution/Simple/Configure.hs
@@ -54,7 +54,7 @@ module Distribution.Simple.Configure
   , platformDefines,
   ) where
 
-import Prelude (head, tail, last)
+import qualified Prelude (tail)
 import Distribution.Compat.Prelude
 
 import Distribution.Compiler
@@ -102,6 +102,7 @@ import Distribution.Backpack.DescribeUnitId
 import Distribution.Backpack.PreExistingComponent
 import Distribution.Backpack.ConfiguredComponent (newPackageDepsBehaviour)
 import Distribution.Backpack.Id
+import Distribution.Utils.Generic
 import Distribution.Utils.LogProgress
 
 import qualified Distribution.Simple.GHC   as GHC
@@ -112,6 +113,7 @@ import qualified Distribution.Simple.HaskellSuite as HaskellSuite
 import Control.Exception
     ( ErrorCall, Exception, evaluate, throw, throwIO, try )
 import Control.Monad ( forM, forM_ )
+import Data.List.NonEmpty            ( nonEmpty )
 import Distribution.Compat.Binary    ( decodeOrFailIO, encode )
 import Distribution.Compat.Directory ( listDirectory )
 import Data.ByteString.Lazy          ( ByteString )
@@ -1313,18 +1315,21 @@ selectDependency pkgid internalIndex installedIndex requiredDepsMap
 
     -- It's an external package, normal situation
     do_external_external =
-        case PackageIndex.lookupDependency installedIndex dep_pkgname vr of
-          []   -> Left (DependencyNotExists dep_pkgname)
-          pkgs -> Right $ head $ snd $ last pkgs
+        case pickLastIPI $ PackageIndex.lookupDependency installedIndex dep_pkgname vr of
+          Nothing  -> Left (DependencyNotExists dep_pkgname)
+          Just pkg -> Right pkg
 
     -- It's an internal library, being looked up externally
     do_external_internal
       :: LibraryName -> Either FailedDependency InstalledPackageInfo
     do_external_internal ln =
-        case PackageIndex.lookupInternalDependency installedIndex
+        case pickLastIPI $ PackageIndex.lookupInternalDependency installedIndex
                 (packageName pkgid) vr ln of
-          []   -> Left (DependencyMissingInternal dep_pkgname (packageName pkgid))
-          pkgs -> Right $ head $ snd $ last pkgs
+          Nothing  -> Left (DependencyMissingInternal dep_pkgname (packageName pkgid))
+          Just pkg -> Right pkg
+
+    pickLastIPI :: [(Version, [InstalledPackageInfo])] -> Maybe InstalledPackageInfo
+    pickLastIPI pkgs = safeHead . snd . last =<< nonEmpty pkgs
 
 reportSelectedDependencies :: Verbosity
                            -> [ResolvedDependency] -> IO ()
@@ -1773,7 +1778,7 @@ checkForeignDeps pkg lbi verbosity =
         findOffendingHdr =
             ifBuildsWith allHeaders ccArgs
                          (return Nothing)
-                         (go . tail . inits $ allHeaders)
+                         (go . Prelude.tail . inits $ allHeaders) -- inits always contains at least []
             where
               go [] = return Nothing       -- cannot happen
               go (hdrs:hdrsInits) =
@@ -1782,8 +1787,9 @@ checkForeignDeps pkg lbi verbosity =
                       -- If that works, try compiling too
                       (ifBuildsWith hdrs ccArgs
                         (go hdrsInits)
-                        (return . Just . Right . last $ hdrs))
-                      (return . Just . Left . last $ hdrs)
+                        (return . fmap Right . safeLast $ hdrs))
+                      (return . fmap Left . safeLast $ hdrs)
+
 
               cppArgs = "-E":commonCppArgs -- preprocess only
               ccArgs  = "-c":commonCcArgs  -- don't try to link
@@ -2004,7 +2010,7 @@ checkRelocatable verbosity pkg lbi
     -- database to which the package is installed are relative to the
     -- prefix of the package
     depsPrefixRelative = do
-        pkgr <- GHC.pkgRoot verbosity lbi (last (withPackageDB lbi))
+        pkgr <- GHC.pkgRoot verbosity lbi (registrationPackageDB (withPackageDB lbi))
         traverse_ (doCheck pkgr) ipkgs
       where
         doCheck pkgr ipkg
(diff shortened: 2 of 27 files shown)


Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -460,19 +456,21 @@ listInvocation hpi _verbosity packagedb =
...
-                                       : ("--no-user-" ++ packageDbFlag hpi)
-                                       : map specific dbs
-  _                                   -> ierror
-  where
-    specific (SpecificPackageDB db) = "--" ++ packageDbFlag hpi ++ "=" ++ db
-    specific _ = ierror
-    ierror :: a
-    ierror     = error ("internal error: unexpected package db stack: " ++ show dbstack)
+packageDbStackOpts hpi dbstack
+  | noPkgDbStack hpi = [packageDbOpts hpi (registrationPackageDB dbstack)]
```
dwijnand: Pulling this predicate into `packageDbStackOpts` also affects `invoke` (line 102), which I think is correct, but I thought I would mention it.

Cabal/Distribution/Simple.hs
```diff
@@ -524,9 +524,9 @@ sanityCheckHookedBuildInfo verbosity
       ++ "but the package does not have a library."
 
 sanityCheckHookedBuildInfo verbosity pkg_descr (_, hookExes)
-    | not (null nonExistant)
+    | (exe1:_) <- nonExistant
```
phadej: this parens are unnecessary, i.e.

```haskell
    | exe1 : _ <- ...
    = ...
```

(then you need to add spaces around `:`, otherwise it will look weird :)

Cabal/Distribution/Simple/Configure.hs
```diff
@@ -1313,18 +1315,21 @@ selectDependency pkgid internalIndex installedIndex requiredDepsMap
...
     do_external_internal ln =
-        case PackageIndex.lookupInternalDependency installedIndex
+        case pick $ PackageIndex.lookupInternalDependency installedIndex
                 (packageName pkgid) vr ln of
-          []   -> Left (DependencyMissingInternal dep_pkgname (packageName pkgid))
-          pkgs -> Right $ head $ snd $ last pkgs
+          Nothing  -> Left (DependencyMissingInternal dep_pkgname (packageName pkgid))
+          Just pkg -> Right pkg
+
+    pick :: [(Version, [InstalledPackageInfo])] -> Maybe InstalledPackageInfo
```
phadej: either change the name to `pickLastIPI` or the name&type to `pickLast :: [(a,b)] -> Maybe b`

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -1074,22 +1074,20 @@ exeMainModuleName Executable{buildInfo = bnfo} =
 -- https://github.com/haskell/cabal/pull/4539#discussion_r118981753.
 decodeMainIsArg :: String -> Maybe ModuleName
 decodeMainIsArg arg
-  | not (null main_fn) && isLower (head main_fn)
-                        -- The arg looked like "Foo.Bar.baz"
+  | any isLower (safeHead main_fn)
```
phadej: this is use of `any` makes me think, than smile.

On the edge of over-cleverness.

I'd rather define a helper `headOf main_fn isLower` function.

Cabal/Distribution/Simple/GHCJS.hs
```diff
@@ -926,12 +926,12 @@ exeMainModuleName Executable{buildInfo = bnfo} =
 -- https://github.com/haskell/cabal/pull/4539#discussion_r118981753.
 decodeMainIsArg :: String -> Maybe ModuleName
 decodeMainIsArg arg
-  | not (null main_fn) && isLower (head main_fn)
-                        -- The arg looked like "Foo.Bar.baz"
+  | any isLower (safeHead main_fn)
```
phadej: same as in `GHC` module

Cabal/Distribution/Simple/Program/Run.hs
```diff
@@ -243,13 +243,13 @@ multiStageProgramInvocation simple (initial, middle, final) args =
...
+        []        -> [ simple ]
 
-        [c]    -> [ simple  `appendArgs` c ]
+        [c]       -> [ simple  `appendArgs` c ]
 
-        (c:cs) -> [ initial `appendArgs` c ]
-               ++ [ middle  `appendArgs` c'| c' <- init cs ]
-               ++ [ final   `appendArgs` c'| let c' = last cs ]
+        (c:c2:cs) -> [ initial `appendArgs` c ]
+                  ++ [ middle  `appendArgs` c'| c' <- init (c2:|cs) ]
```
phadej: Don't we have `unsnoc :: NonEmpty a -> ([a], a)` ? feels boring to traverse the list twice (and mention it twice). Then we could written

```haskell
(c:c2:cs) | (xs, x) <- unsnoc (c2:|cs) -> [ ... ] ++ [ ... | c' <- xs ] ++ [ final `appendArgs` x ]
```

```haskell
unsnocNE :: NonEmpty a -> ([a], a)
unsnocNE (x:|xs) =  go x xs where
    go y []     = ([], y)
    go y (z:zs) = let ~(ws, w) = go z zs in (y : ws, w)
```

Cabal/Distribution/Types/VersionRange.hs
```diff
@@ -130,7 +130,15 @@ withinRange v = foldVersionRange
...
-    \lowerBound -> init lowerBound ++ [last lowerBound + 1]
+    \lowerBound -> case unsnoc lowerBound of
+        Nothing      -> []
+        Just (xs, x) -> xs ++ [x + 1]
+
+unsnoc :: [a] -> Maybe ([a], a)
+unsnoc []     = Nothing
+unsnoc [h]    = Just ([], h)
+unsnoc (h:tl) = Just (h:mid, last1)
+    where Just (mid, last1) = unsnoc tl
```
phadej: partial match, see my `unsnocNE` and adapt it.

Cabal/tests/UnitTests/Distribution/Version.hs
```diff
@@ -342,7 +344,15 @@ prop_foldVersionRange range =
...
 
-    upper = alterVersion $ \numbers -> init numbers ++ [last numbers + 1]
+    upper = alterVersion $ \numbers -> case unsnoc numbers of
+      Nothing      -> []
+      Just (xs, x) -> xs ++ [x + 1]
+
+unsnoc :: [a] -> Maybe ([a], a)
+unsnoc []     = Nothing
+unsnoc [h]    = Just ([], h)
+unsnoc (h:tl) = Just (h:mid, last1)
```
phadej: `unsnoc` and `unsnocNE` could go to `D.Generic.Utils`.

cabal-install/Distribution/Client/Outdated.hs
```diff
@@ -197,14 +198,15 @@ listOutdated deps pkgIndex (ListOutdatedSettings ignorePred minorPred) =
     lookupLatest :: Dependency -> [Version]
     lookupLatest dep@(Dependency pname vr _)
       | minorPred (depPkgName dep) =
-        map packageVersion $ lookupDependency pkgIndex  pname (relaxMinor vr)
+        map packageVersion $ lookupDependency pkgIndex  pname =<< maybeToList (relaxMinor vr)
       | otherwise                  =
         map packageVersion $ lookupPackageName pkgIndex (depPkgName dep)
 
-    relaxMinor :: VersionRange -> VersionRange
+    relaxMinor :: VersionRange -> Maybe VersionRange
```
phadej: I think this i better to still return only `VersionRange`. I.e. if there isn't `last`, just return `vr` as is.

cabal-install/solver-dsl/UnitTests/Distribution/Solver/Modular/DSL.hs
```diff
@@ -719,13 +720,13 @@ extractInstallPlan :: CI.SolverInstallPlan.SolverInstallPlan
...
-    confPkg (CI.SolverInstallPlan.Configured pkg) = Just $ srcPkg pkg
+    confPkg (CI.SolverInstallPlan.Configured pkg) = srcPkg pkg
     confPkg _                               = Nothing
 
-    srcPkg :: SolverPackage UnresolvedPkgLoc -> (String, Int)
+    srcPkg :: SolverPackage UnresolvedPkgLoc -> Maybe (String, Int)
     srcPkg cpkg =
       let C.PackageIdentifier pn ver = packageInfoId (solverPkgSource cpkg)
-      in (C.unPackageName pn, head (C.versionNumbers ver))
+      in (\vn -> (C.unPackageName pn, vn)) <$> safeHead (C.versionNumbers ver)
```
phadej: Oh, this is subtle.

Almost asks for having two `Version` variants, the one which can be `nullVersion`, and one which cannot. But that's a huge change. cc @hvr

Cabal/Distribution/Simple/GHC.hs
dwijnand: Perhaps.  I've found it works fine as "any head of main_fn isLower" or "the head of main_fn, if any, isLower".

I'll put a little helper to... help it :)

Cabal/Distribution/Types/VersionRange.hs
dwijnand: Is it a partial match?  I thought so too but I couldn't get it to fail in a REPL, but maybe I didn't try the right things?

Cabal/Distribution/Simple/Program/Run.hs
dwijnand: Thanks!  What's the `~` called, so I can read up about it?

phadej: irrefutable pattern, here it's not strictly needed as `let` bindings are irrefutable, but I added it to make it more explicit.

In general it affects lazyness, the strict pattern match on pair, e.g.

```haskell
go y (z:zs) = case go z zs of (ws, w) -> (y : ws, w)
```

would force the traversal of original list way to early. But not we can say:

```haskell
> take 20 $ fst $ uncons (0 :| [1..])
[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19]
```

(though, if we ask for the last value, that would diverge).

cabal-install/solver-dsl/UnitTests/Distribution/Solver/Modular/DSL.hs
dwijnand: In terms of the diff here, is the change fine?  My interpretation is that when extracting the install plans it would skip any package with the null version, rather than blowing up.  Does that sound like a change in the right direction?

dwijnand:
Thank you for the wonderful review.

Cabal/Distribution/Types/VersionRange.hs
phadej: it doesn't fail, but it's still partial. There's no comment saying that `unsnoc tl` will always return `Just something`. You have to think, oh yes, `tl` is non-empty...

cabal-install/solver-dsl/UnitTests/Distribution/Solver/Modular/DSL.hs
phadej: Your current change is fine.

There shouldn't be any `nullVersion` there, but we cannot be sure without a very big refactoring (which payoff is questionable).

Cabal/Distribution/Simple/GHC.hs
phadej: My brain simplified that to `any isLower main_fn` which made me think *this cannot be right*. `safeHead` is not well known function, so it makes you think.

dwijnand:
Duplicate of #6340

PR #6333, opened 2019-11-06
validate.sh: Fix cabal-plan invocation

### LINKED ISSUE #6332: ./validate.sh might not work with current `cabal-plan`
I was made aware that `./validate.sh` might not work with current `cabal-plan` versions due to the placement of the `--builddir=` flag (which in recent versions of `cabal-plan` has become a non-global flag; i.e. it needs to be placed after the sub-command rather than before)

https://github.com/haskell/cabal/blob/48a8aa2bc9d11a27ce6be1c9eb75c9f6eda0e060/validate.sh#L199

cc @dwijnand  @phadej

dwijnand:
In recent versions of cabal-plan the --builddir= flag has become a
non-global flag; i.e. it needs to be placed after the sub-command rather
than before.

`validate.sh` isn't tested in CI so...
[ci skip]

Tested on... phadej's machine.

Fixes #6332


phadej:
thanks

PR #6335, opened 2019-11-06
tests: avoid blocking on git tagging

dwijnand:
I haven't understood 100%, but something about having the following in
my ~/.gitconfig:

    [tag]
      gpgsign = true
      forceSignAnnotated = true

means that creating these tags block on my editor exiting and providing
the annotated tag commit message.  I also didn't investigate all the
options, but I tested that this one fixes the issue for me.

Tested by running `validate.sh`.


phadej:
I wonder if broken meta is due new `alex` release?

dwijnand:
Sounds reasonable to me.  Is someone on it already?  Otherwise, I can send a PR for it in a bit.

PR #6336, opened 2019-11-06
validate.sh: Allow using short-version ghc binaries, like ghc-8.6

dwijnand:
Previously this would work until CABAL_TESTSUITE_BDIR was needed, at
which point it would fail because it looked for things under
.../ghc-8.6/... while cabal v2-build actually put them in
.../ghc-8.6.5/...

Hard-coding "ghc" is unfortunate, but I didn't know if there was an easy
way to avoid it.  I also didn't want to over-complicate it, until it was
necessary.

validate.sh isn't tested in CI so...
[ci skip]

Tested by running `validate.sh`.

### DIFF
--- a/validate.sh
+++ b/validate.sh
@@ -191,7 +191,7 @@ else
     PROJECTFILE=cabal.project.validate
 fi
 
-BASEHC=$(basename $HC)
+BASEHC=ghc-$($HC --numeric-version)
 BUILDDIR=dist-newstyle-validate-$BASEHC
 CABAL_TESTSUITE_BDIR="$(pwd)/$BUILDDIR/build/$ARCH/$BASEHC/cabal-testsuite-${CABAL_VERSION}"
 


dwijnand:
Alternatively, if we can find a way to fast-fail with something like `ghc-8.6` that works too: the outcome is it shouldn't spuriously fail half-way through because of the directory name is wrong.

hvr:
Fwiw, there's a good chance naming your symlinks this way might break tool-naming heuristics and assumptions across the board which have been in place for over a decade.

dwijnand:
That's unfortunate.  Would you be able to represent that view in an issue in https://gitlab.haskell.org/haskell/ghcup/issues?  That's where I got the names from.

phadej:
can you rebase on top of `master`.

This is an improvement. It would allow (in theory to try) to build with `ghc-head`, e.g. or whatever `ghc-weird` one might have.

PR #6364, opened 2019-11-19
Distribution.Compat.Async

phadej:


Cabal/Distribution/Compat/Async.hs
```diff
@@ -0,0 +1,139 @@
...
+    void (waitCatch a)
+
+-- | The exception thrown by `cancel` to terminate a thread.
+data AsyncCancelled = AsyncCancelled
+  deriving (Show, Eq
+    , Typeable
+    )
+
+instance Exception AsyncCancelled where
+#if __GLASGOW_HASKELL__ >= 708
```
hvr: I'd rather consider this a `MIN_VERSION_base` thing; i.e. it's an API feature depending on `base` version rather than GHC's version IMO

btw, I was trying to understand why the async package does this this way and this is the commit which provides some level of explanation (might be worth including as source-comment for other people wondering about this thing): https://github.com/ghc/ghc/commit/756a970eacbb6a19230ee3ba57e24999e4157b09

Cabal/Distribution/Compat/Async.hs
```diff
@@ -0,0 +1,139 @@
...
+import Data.Typeable           (Typeable)
+import GHC.Exts                (inline)
+
+-- TODO: base?
+#if __GLASGOW_HASKELL__ >= 708
+import Control.Exception (asyncExceptionFromException, asyncExceptionToException)
+#endif
+
+-- | Async, but based on 'MVar', as we don't depend on @stm@.
+data AsyncM a = Async
```
hvr: What's the motivation for not just naming it `Async`? Isn't the MVar vs STM thing merely an internal implementation detail that doesn't affect the public API?

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -829,53 +828,52 @@ rawSystemStdInOut verbosity path args mcwd menv input outputMode = withFrozenCal
...
-
-      -- Check if we we hit an exception while consuming the output
-      -- (e.g. a text decoding error)
-      reportOutputIOError mberr1
-      reportOutputIOError mberr2
-
-      return (out, err, exitcode)
+      let force :: NFData a => a -> IO a
+          force str = do
+            evaluate (rnf str)
```
hvr: it's a bit confusing as this clashes with `Control.DeepSeq.force` (which is available since deepseq-1.2; and Cabal requires deepseq>=1.3); if we're touching this code anyway, why not simply import `force` and use `(evaluate . force)` below? or at least call this local helper `evaluateNF` or `forceIO` instead :-)

alternatively, if this is a common pattern (it's quite common you'd want the backgrounded task to do the actual computation than to end up with some lazy thunk returned rightaway which would render the backgrounding pointless...) you could define a variant `withAsyncNF` which encodes this common use-pattern

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -829,53 +828,52 @@ rawSystemStdInOut verbosity path args mcwd menv input outputMode = withFrozenCal
...
+            -- todo: ignoreSigPipe
+            IOData.hPutContents inh inputData
+            --TODO: this probably fails if the process refuses to consume
+            -- or if it closes stdin (eg if it exits)
+
+        -- wait for both to finish
+        mberr1 <- waitCatch outA
+        mberr2 <- waitCatch errA
+
+        err <- reportOutputIOError mberr2
```
hvr: any particular reason you moved this bit to an earlier point in time? if so, might be worth commenting why it's checked before waiting for the program to terminate as well as  `mberr1`

Cabal/Distribution/Compat/Async.hs
phadej: To make it clear that it's different.

Cabal/Distribution/Simple/Utils.hs
phadej: The name `force` was there already, and I didn't tried too hard to come with new one.

`withAsyncNF` is a good point, but won't do it for this sole use case.

Cabal/Distribution/Simple/Utils.hs
phadej: Ah, true. Have to move it after. The error message printing spits out `err`.

Cabal/Distribution/Compat/Async.hs
phadej: ... in some sense, I'd like to go over the use sites if/when we change to STM based implementation.

Cabal/Distribution/Compat/Async.hs
phadej: changed to use `MIN_VERSION_base`. I had a forgotten todo there in fact.

PR #6434, opened 2019-12-15
Update solver Hackage benchmark for v2-install.

### LINKED ISSUE #6417: Update solver Hackage benchmark for v2 commands
The Hackage benchmark currently measures the performance of the solver by running the "cabal install" command with --dry-run.  Now that cabal uses v2 commands by default, "install" only works for executables.  The benchmark needs to be updated to run a command that can invoke the solver for any package on Hackage.

install command: https://github.com/haskell/cabal/blob/b46d7820009784a878ae89b6810f750e2fb733d3/solver-benchmarks/HackageBenchmark.hs#L175

grayjay:
Fixes #6417.

Changes in this commit:

- Set the --lib flag so that the benchmark works for packages with libraries or
  executables.
- Set --ignore-project to prevent cabal projects from affecting the results.
- Handle four types of cabal errors that were introduced by v2-install.
- Use +nowrap to simplify parsing of cabal errors.

---

### DIFF
--- a/solver-benchmarks/HackageBenchmark.hs
+++ b/solver-benchmarks/HackageBenchmark.hs
@@ -56,6 +56,10 @@ data CabalResult
   = Solution
   | NoInstallPlan
   | BackjumpLimit
+  | Unbuildable
+  | UnbuildableDep
+  | ComponentCycle
+  | ModReexpIssue
   | PkgNotFound
   | Timeout
   | Unknown
@@ -81,14 +85,14 @@ hackageBenchmarkMain = do
   -- "trial" or "summary".
   when argPrintTrials $ putStr $ printf "%-16s " "trial/summary"
   putStrLn $
-      printf "%-*s %-13s %-13s %11s %11s %11s %11s %11s"
+      printf "%-*s %-14s %-14s %11s %11s %11s %11s %11s"
              nameColumnWidth "package" "result1" "result2"
              "mean1" "mean2" "stddev1" "stddev2" "speedup"
 
   forM_ pkgs $ \pkg -> do
     let printTrial msgType result1 result2 time1 time2 =
             putStrLn $
-            printf "%-16s %-*s %-13s %-13s %10.3fs %10.3fs"
+            printf "%-16s %-*s %-14s %-14s %10.3fs %10.3fs"
                    msgType nameColumnWidth (unPackageName pkg)
                    (show result1) (show result2)
                    (diffTimeToDouble time1) (diffTimeToDouble time2)
@@ -125,7 +129,7 @@ hackageBenchmarkMain = do
       if isSignificantResult result1 result2
           || isSignificantTimeDifference argPValue ts1 ts2
       then putStrLn $
-           printf "%-*s %-13s %-13s %10.3fs %10.3fs %10.3fs %10.3fs %10.3f"
+           printf "%-*s %-14s %-14s %10.3fs %10.3fs %10.3fs %10.3fs %10.3f"
                   nameColumnWidth (unPackageName pkg)
                   (show result1) (show result2) mean1 mean2 stddev1 stddev2 speedup
       else when (argPrintTrials || argPrintSkippedPackages) $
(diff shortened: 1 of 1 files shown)


phadej:
Nice, `--ignore-project` immediately in use!

grayjay:
I noticed your PR when I first pushed my branch.  Thanks for adding the flag!

phadej:
When I try 

```
cabal run hackage-benchmark -- --cabal1 /opt/cabal/3.0/bin/cabal --cabal2 $(cabal-plan list-bin cabal) --packages tree-diff --print-trials
```

I get

```
trial/summary    package   result1        result2              mean1       mean2     stddev1     stddev2     speedup
trial            tree-diff Unknown        Solution            0.014s      5.325s
```

I think that `Unknown` is a failure, how can I check why it's unknown?

phadej:
Oh, I know, because `cabal-install-3.0` doens't yet know about `--ignore-project`; that makes sense.

phadej:
Another thing:

On my system, the `~/.cabal/store` is huge, which affects the solver "significantly", in the sense that the relative speedups are considerably less.

Is there some mechanism to prevent store from affecting the benchmarks?

Could we create some temporary directory, and use `--storedir=tempstore`, then `cabal` won't spend much time reading it.

phadej:
Something like

```diff
diff --git a/solver-benchmarks/HackageBenchmark.hs b/solver-benchmarks/HackageBenchmark.hs
index 145b49f93..df79d361e 100644
--- a/solver-benchmarks/HackageBenchmark.hs
+++ b/solver-benchmarks/HackageBenchmark.hs
@@ -176,7 +176,7 @@ runCabal timeoutSeconds cabal flags pkg = do
     let timeout = "timeout --foreground -sINT " ++ show timeoutSeconds
         cabalCmd =
             unwords $
-            [cabal, "install", "--ignore-project", "--lib", unPackageName pkg, "--dry-run", "-vsilent+nowrap"] ++ flags
+            [cabal, "--store-dir=/tmp/store", "v2-install", "--ignore-project", "--lib", unPackageName pkg, "--dry-run", "-vsilent+nowrap"] ++ flags
         cmd = (shell (timeout ++ " " ++ cabalCmd)) { std_err = CreatePipe }
 
     -- TODO: Read stdout and compare the install plans.
```

- We should create temporary directory using [`temporary`](https://hackage.haskell.org/package/temporary), rather then hard coding one
- I'd also prefer `v2-install`.

phadej:
In fact, we probably should specify some non-existing environment, as `--lib` reads the `default` one.

```
diff --git a/solver-benchmarks/HackageBenchmark.hs b/solver-benchmarks/HackageBenchmark.hs
index 145b49f93..ed2bdd5f8 100644
--- a/solver-benchmarks/HackageBenchmark.hs
+++ b/solver-benchmarks/HackageBenchmark.hs
@@ -176,7 +176,7 @@ runCabal timeoutSeconds cabal flags pkg = do
     let timeout = "timeout --foreground -sINT " ++ show timeoutSeconds
         cabalCmd =
             unwords $
-            [cabal, "install", "--ignore-project", "--lib", unPackageName pkg, "--dry-run", "-vsilent+nowrap"] ++ flags
+            [cabal, "--store-dir=/tmp/store", "v2-install", "--package-env=hackage-benchmarks-dont-use", "--ignore-project", "--lib", unPackageName pkg, "--dry-run", "-vsilent+nowrap"] ++ flags
         cmd = (shell (timeout ++ " " ++ cabalCmd)) { std_err = CreatePipe }
 
     -- TODO: Read stdout and compare the install plans.
```

note: `--package-env`; I think it's ok to hardcode one which is unlikely to exist, as far as it's not the `default`.

Sorry for the comment spam.

grayjay:
@phadej Thanks for the suggestions!  I didn't realize that so much time was spent reading the store, so I definitely want to skip that.  I just specified a directory that shouldn't exist, because cabal shouldn't need to create the store with --dry-run.  I also added v2- and --package-env and documented the flags.

phadej:
Merged in https://github.com/haskell/cabal/pull/6447, thanks!

Issue #6144, opened 2019-07-14
Project without packages


bgamari:
**Describe the bug**
Currently `cabal-install` expects projects to have at least one local package listed in `cabal.project`. However, there are use-cases where one might want a project without any local packages (e.g. to create an environment for using a project-specific repository, e.g. to build a package from `head.hackage` with `cabal new-install`).

**To Reproduce**
```
$ cat >cabal.project <<EOF
repository head.hackage.ghc.haskell.org
   url: https://ghc.gitlab.haskell.org/head.hackage/
   secure: True
   key-threshold: 3
   root-keys:
       7541f32a4ccca4f97aea3b22f5e593ba2c0267546016b992dfadcd2fe944e55d
       26021a13b401500c8eb2761ca95c61f2d625bfef951b939a8124ed12ecf07329
       f76d08be13e9a61a377a85e2fb63f4c5435d40f8feb3e12eb05905edb8cdea89
$ cabal v2-install --lib aeson
cabal: There is no <pkgname>.cabal package file or cabal.project file. To
build packages locally you need at minimum a <pkgname>.cabal file. You can use
'cabal init' to create one.

For non-trivial projects you will also want a cabal.project file in the root
directory of your project. This file lists the packages in your project and
all other build configuration. See the Cabal user guide for full details.
```

**Expected behavior**

`cabal new-install` would build `aeson`.

phadej:
Work-around: create a dummy package. fake-package and z are popular names (and AFAIK blacklisted from Hackage)

PR #6436, opened 2019-12-15
Resolve #6013: Warn if packages and optional-packages are both empty

### LINKED ISSUE #6013: Inscrutable error message if cabal.project has no packages field
If I create a project with a `cabal.project` file that does not list the packages field then I get the error:

```
cabal: There is no <pkgname>.cabal package file or cabal.project file. To
build packages locally you need at minimum a <pkgname>.cabal file. You can use
'cabal init' to create one.

For non-trivial projects you will also want a cabal.project file in the root
directory of your project. This file lists the packages in your project and
all other build configuration. See the Cabal user guide for full details.
```

I think this is not a very helpful message for the problem at hand.

phadej:
Now there's a warning

```
Warning: There are no packages or optional-packages in the project
cabal: There is no <pkgname>.cabal package file or cabal.project file. To
build packages locally you need at minimum a <pkgname>.cabal file. You can use
'cabal init' to create one.

For non-trivial projects you will also want a cabal.project file in the root
directory of your project. This file lists the packages in your project and
all other build configuration. See the Cabal user guide for full details.
```

It's subtle, but that's the issue with how we output warnings, not the warning itself.

I'm only making it a warning at this point, as I'm not 100% sure if that setup may still be useful for something. Playing safe for 3.2

### DIFF
--- a/cabal-install/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/Distribution/Client/ProjectOrchestration.hs
@@ -222,6 +222,10 @@ establishProjectBaseContext verbosity cliConfig currentCommand = do
                           verbosity cabalDirLayout
                           projectConfig
 
+    -- https://github.com/haskell/cabal/issues/6013
+    when (null (projectPackages projectConfig) && null (projectPackagesOptional projectConfig)) $
+        warn verbosity "There are no packages or optional-packages in the project"
+
     return ProjectBaseContext {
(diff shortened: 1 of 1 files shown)


phadej:
Related: https://github.com/haskell/cabal/issues/6144

phadej:
Backported in https://github.com/haskell/cabal/pull/6430

PR #6447, opened 2019-12-18
Grayjay update solver benchmarks

phadej:
One commit on top of #6434 which allows to run benchmarks as part of `validate.sh` (to see that they work).

Probably because of asserts in development version, `cabal-install-3.0.1.0` seems to be slightly faster, on my machine :)

cc @grayjay, I'll merge this when travis turns green.


grayjay:
Thanks!

Issue #6139, opened 2019-07-08
How does cabal select a package from multiple global repositories?


TerrorJack:
When multiple repositories (e.g. both head.hackage and regular hackage) are present in `~/.cabal/config`, a single package might come from different repos, how does `cabal` decide from which repo is the package picked up? The [documentation](https://github.com/haskell/cabal/blob/master/Cabal/doc/installing-packages.rst#repository-specification) only says:

> If you want, you can configure multiple repositories, and cabal will combine them and be able to download packages from any of them.

The intuition would be selecting from the first repo which contains the package, but it's actually not the case, e.g. when running `cabal new-install dlist` or `cabal new-build` for a local package depending on `dlist`, the `dlist` package from head.hackage is *not* picked up, even when head.hackage comes before regular hackage in the global config.

gbaz:
Note that there is a ghc proposal pending review that addresses the lack of control over this: https://github.com/ghc-proposals/ghc-proposals/pull/115

23Skidoo:
IIRC it depends on the order the repos are declared in the config file, I think the last one wins.

phadej:
Yes, last one wins. I'd already implemented a `priority` field, but that effort is blocked on e.g. https://github.com/haskell/cabal/issues/6101 (I really don't want to touch any of `ReadP` parsers at this point of time).

fgaz:
i think the recently released `active-repositories` fixes this

PR #6448, opened 2019-12-18
Implement file+noindex:///local/repositories

### LINKED ISSUE #6359: RFC: New local fs index.tar-less repository style to simplify "offline" source-distribution
## Problem statement

In various situations, there's the desire to provide a self-contained offline buildable source-distribution that (assuming you have a working toolchain installed, i.e. `ghc` + `cabal`) doesn't require any network access to build and install a cabal project. This basically requires to include all the dependencies of a project as part of the source-distribution.

In fact, I've hacked up some tooling in the past to accomplish this for my own needs at https://github.com/haskell-hvr/cabal-offline but never got around to polish it and release it properly, as I'm not satisfied with the approach yet.

The way `cabal-offline`'s approach is supposed to work is basically to start from the `plan.json` file which describes what ingredients the cabal solver deemed necessary for the most recently computed build-plan, and collect the respective `.tar.gz` files from the local package repo cache (+ any revised `.cabal` metadata files from the respective `01-index.tar`).

Once we have the dependencies' `.tar.gz` and `.cabal` files, we need to provide those to cabal's solver. There's basically 3 options here:

1. Point to the tarballs via `packages: deps/*.tar.gz`
2. Unpack and update the .cabal files, and point cabal to them via `packages: deps/*/*.cabal`
3. Construct a file-local package repository (with e.g. `hackage-repo-tool`)

Neither of these options is ideal, and I'll quickly go through these:

### Approach 1: Point to the tarballs via `packages: deps/*.tar.gz`

This only works conveniently for the case where

- (A) There's no revised `.cabal` metadata required (this can be workarounded by with a `allow-newer: *:*`+`allow-older: *:*` combined with a `cabal.plan.freeze` file which needs to pin down all cabal flag choices as allow-newed/older might break those otherwise), **and**
- (B) the build-plan doesn't require multiple versions of the same package (this can occur via [qualified goals](https://www.well-typed.com/blog/2015/03/qualified-goals/), i.e. for build-tools and custom-setups)

The former (i.e. (A)) can be workarounded (while harming the property (C) mentioned in the next paragraph by repackaging .tar.gz with their revised .cabal file), but the latter (i.e. (B)) is often a real blocker.

(C) On the bright side, dependencies will be properly shared in the Nix-style store, as their sha256sums will be exactly the same you'd have if you had build your project with the real Hackage remote repo. I.e. no recompilation of dependencies needed when you switch between your offline source-distro build and your normal "development" build which might have access to a larger package repo and the build-plan doesn't change.

### Approach 2: Unpack and update the .cabal files, and point cabal to them via `packages: deps/*/*.cabal`

This addresses (A) but not (B), but completely sacrifices (C), as the dependencies' artefacts are no longer cacheable in the Nix-style store, but instead are treated as local packages and lack proper sha256 sums.

### Approach 3: Construct a file-local package repository (with e.g. `hackage-repo-tool`)

This approach nicely addressed (A), (B) while retaining (C). But it's got a few annoyances (which I'll denote as (D)); I'll explain:

So we currently support a number styles of package repositories via the `repository` directive in `~/.cabal/config` and `cabal.project` files (c.f. https://cabal.readthedocs.io/en/latest/installing-packages.html#repository-specification)

1. Remote non-TUF repositories (this is also what the legacy syntax `remote-repo` denotes)
2. Remote TUF-secured repositories
3. Local TUF-secured repositories
4. Local non-TUF repositories (c.f. legacy syntax `local-repo`)

As a side-remark: for the Nix-style build paradigm, the non-TUF repositories are very inconvenient, as they don't provide the tarball sha256sums in their metadata, which the solver phase needs to compute nix-style hashes for the build-plan. This would require the solver to unnecessarily fetch .tar.gzs for the purpose of computing their sha256sums even if they were already cached in the nix-style store and thus we wouldn't have needed to download their source to begin with...

(D) So, one annoyance is that you have to run `hackage-repo-tool` and construct a redundant `01-index.tar.gz` from the set of `.tar.gz` and revised `.cabal` files. In order to do that, you also need to create some dummy TUF secret keys to sign the dummy TUF metadata with, and construct a `repository`-directive with the appropriate `root-keys` and `key-threshold` settings in your distribtued `cabal.project` file. Moreover, another annoyance is  that the TUF signatures would expire at some point, and then cabal would refuse to build the source-dist...

So ultimately I don't think that local TUF-secured repos make much sense for this use-case as they come with accidental complexity, and why would you even want to TUF-signatures in the first place for a source-distribution whose local sources you'd typically sign externally anyway by signing the tar'ed up aggregate deliverable source-distro wholesale?

## Proposal

Consequently, I propose to combine the benefits the 3 approaches (i.e. addressing (A)+(B), while achieving (C) and avoiding (D)) by providing a new style of index/TUF-less local package repositories, by introducing a new uri schema `file+noindex:` like

```
repository my-local-repo
  url: file+noindex:/path/to/local/repo
```

`file+noindex:` must point to a folder in the filesystem. The folder `/path/to/local/repo/` in the example above (which could also be a relative path such as `file+noindex:./deps/`) would contain files of the form 

- `${pkgid}.tar.gz`,
- `${pkgid}.cabal` (optional; a corresponding `${pkgid}.tar.gz` must exist!), or
- `preferred-versions` (optional)

Then cabal would scan that folder and compute sha256sums on the fly (which is quite cheap) and construct its package index data-structure (which it could also cache easily).

So the minimal form would be to have the folder contain merely `${pkgid}.tar.gz` files, which would already be able to address (B) in the absence of revised .cabal files; i.e. allow for multiple versions of the same package.

If there's also a `${pkgid}.cabal` for a given `${pkgid}.tar.gz`, then this `.cabal` file is taken as the package description for `${pkgid}`. If it's missing, cabal will extract the `.cabal` file from the respective `${pkgid}.tar.gz` (and ideally cache it).

If there's a `preferred-versions` file, then do cabal will treat it as if it was a combined `preferred-versions` file from the `index-01.tar.gz` (or how it was represented in the legacy `00-index.tar.gz` file).

This would considerably simplify the implementation of tooling such as `cabal-offline`, maybe to the point where we could even simply integrate it directly into `cabal` proper.

## Possible future work

(NB: this is not part of the proposal proper; this merely loud thinking to express possible directions on how this feature could be leveraged)

...imagine a `cabal sdist --offline` operation; but having this in an companion tool is fine too as I can think of various ways to tweak the offline bundle generation (e.g. take a set of plan.json files for various computed build-plans for different configurations; e.g. for different GHC versions or platforms; and construct an offline source-distro that is able to work on all of these;  or have a planjson2bash conversion which is able to emit a bash script for a given plan.json buildplan, and include those in the source-dist  -- think cabal's `bootstrap.sh`; there's lots of ideas to explore here, and these might be better to explore in a decoupled tool rather than bloat cabal and add to its maintenance overhead...)

phadej:
Resolve #6359

`preferred-versions` are left out for now.
It shouldn't be difficult to add, but needs work nevertheless.

Also change the index cache to use `Distribution.Utils.Structured`,
making Binary instances generically derived.

Almost as a feature generated 01-index.cache is never updated.
If you change the contents of the directory, you have to purge
01-index.cache file yourself.

Screenshot from 2019-12-19 14-32-56

### DIFF
--- a/Cabal/Distribution/Utils/Structured.hs
+++ b/Cabal/Distribution/Utils/Structured.hs
@@ -49,8 +49,10 @@ module Distribution.Utils.Structured (
     -- | These functions operate like @binary@'s counterparts,
     -- but the serialised version has a structure hash in front.
     structuredEncode,
+    structuredEncodeFile,
     structuredDecode,
     structuredDecodeOrFailIO,
+    structuredDecodeFileOrFail,
     -- * Structured class
     Structured (structure),
     MD5,
@@ -262,6 +264,10 @@ structuredEncode
   => a -> LBS.ByteString
 structuredEncode x = Binary.encode (Tag :: Tag a, x)
 
+-- | Lazily serialise a value to a file
+structuredEncodeFile :: (Binary.Binary a, Structured a) => FilePath -> a -> IO ()
+structuredEncodeFile f = LBS.writeFile f . structuredEncode
+
 -- | Structured 'Binary.decode'.
 -- Decode a value from a lazy 'LBS.ByteString', reconstructing the original structure.
 -- Throws pure exception on invalid inputs.
@@ -279,6 +285,10 @@ structuredDecodeOrFailIO bs =
 #else
     handler (ErrorCall str) = return $ Left str
 #endif
+
+-- | Lazily reconstruct a value previously written to a file.
+structuredDecodeFileOrFail :: (Binary.Binary a, Structured a) => FilePath -> IO (Either String a)
+structuredDecodeFileOrFail f = structuredDecodeOrFailIO =<< LBS.readFile f
 
 -------------------------------------------------------------------------------
 -- Helper data
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -57,7 +57,7 @@ The name of the repository is given on the first line, and can be
 anything; packages downloaded from this repository will be cached under
 ``~/.cabal/packages/hackage.haskell.org`` (or whatever name you specify;
 you can change the prefix by changing the value of
-``remote-repo-cache``). If you want, you can configure multiple
+:cfg-field:`remote-repo-cache`). If you want, you can configure multiple
 repositories, and ``cabal`` will combine them and be able to download
 packages from any of them.
 
@@ -97,7 +97,32 @@ received were the right ones. How that is done is however outside the
 scope of ``cabal`` proper.
 
 More information about the security infrastructure can be found at
-https://github.com/well-typed/hackage-security.
+https://github.com/haskell/hackage-security.
+
+Local no-index repositories
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It's possible to use a directory of `.tar.gz` package files as a local package
+repository.
+
+::
+
+    repository my-local-repository
+      url: file+noindex:///absolute/path/to/directory
+
+``cabal`` will construct the index automatically from the
+``package-name-version.tar.gz`` files in the directory, and will use optional
+corresponding ``package-name-version.cabal`` files as new revisions.
+
+The index is cached inside the given directory. If the directory is not
+writable, you can append ``#shared-cache`` fragment to the URI,
+then the cache will be stored inside the :cfg-field:`remote-repo-cache` directory.
+The part of the path will be used to determine the cache key part.
+
+.. note::
+    The URI scheme ``file:`` is interpreted as a remote repository,
+    as described in the previous sections, thus requiring manual construction
+    of ``01-index.tar`` file.
 
 Legacy repositories
 ^^^^^^^^^^^^^^^^^^^
@@ -120,7 +145,7 @@ although, in (and only in) the specific case of Hackage, the URL
 ``http://hackage.haskell.org/packages/archive`` will be silently
 translated to ``http://hackage.haskell.org/``.
 
-The second kind of legacy repositories are so-called “local”
+The second kind of legacy repositories are so-called “(legacy) local”
 repositories:
 
 ::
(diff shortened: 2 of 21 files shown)


phadej:
I used the current state to build hadrian in "offline" mode, see https://gitlab.haskell.org/ghc/ghc/merge_requests/2348

I'll merge this soon to backport to 3.2 branch. Please review.

typedrat:
This is exactly functionality that I was sorely missing the last time that I worked on trying to redesign/fix `v2-install`! Nice.

phadej:
@typedrat could you expand on that? What breaks with `packages: **/*.tar.gz`? Could you open issue describing that problem. We might want to extend what can be specified in the project file still. There is overlap, and drawing a line is difficult.

Creating a temporary `LocalRepo` with a sdists of a `cabal.project` is simple and elegant solution, **if** we have a way to force cabal install versions from that `LocalRepo`, and not other configured repositories. The repository overriding / preference is not developed yet. https://github.com/haskell/cabal/issues/6139 is related. Yet we have to think about the problem better. I guess we
- often want package wide shadowing, i.e. "later" repository with `pkg-verA` removes all `pkg` from previous repositories
- but sometimes only per-version shadowing is enough, i.e. `pkg-verA` shadows only `pkg-verB`
- and with case of remote repositories we might want to selectively pick only some packages (or package-versions`) from overlays

I.e. we have develop a small configuration language for `mconcat` in https://github.com/haskell/cabal/blob/787b1f22739d950f70617ecd0e0fc39353408b87/cabal-install/Distribution/Client/IndexUtils.hs#L276-L278 i.e. parametrise `PackageIndex` `merge` https://github.com/haskell/cabal/blob/787b1f22739d950f70617ecd0e0fc39353408b87/Cabal/Distribution/Simple/PackageIndex.hs#L237-L256

phadej:
Note: the overriding is not important for *offline usage*, as there you would specify the sole `LocalRepo`. Single repository is simple case :)

phadej:

PR #6450, opened 2019-12-20
Fix typos

bwignall:
Should be non-semantic. I would particularly appreciate it if the "version" changes are verified, as they may have been intentional.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos, with https://github.com/bwignall/typochecker to help automate the checking.


phadej:

PR #6456, opened 2019-12-22
Resolve #6375: Let export safe* from both Utils.Generic and Simple.Utils

### LINKED ISSUE #6375: safeHead seems to be exported from both Utils.Generic and Simple.Utils
One place would be enough?

cc @dwijnand

phadej:
So we do with many other stuff


phadej:

Issue #6271, opened 2019-10-01
cabal-testsuite failures on Windows


phadej:
TBD

phadej:
@Mistuke

Does 
```
C:\WORK\cabal-testsuite\PackageTests\Backpack\Includes3\cabal-internal.dist\work\.\dist\build\x86_64-windows\ghc-8.6.5\Includes3-0.1.0.0\l\indef\Includes3-0.1.0.0-inplace-indef+AZNSCvlbjiwGW3rUT5hCYh\build\Includes3-0.1.0.0-inplace-indef+AZNSCvlbjiwGW3rUT5hCYh\autogen\: openBinaryTempFileWithDefaultPermissions:
```

Look like path length limitation? Is there anything one can do to make it not fail?

Mistuke:
Yeah that path is way too long. Which ghc and Cabal is this?

Mistuke:
You can try this before the build/testsuite https://github.com/raaz-crypto/raaz/commit/78a8459a65a04b72cfa45c0c6ccdff5a4cdfff6c

For newer cabal and ghc you can install `ghc-jailbreak` which would take care of it too.

phadej:
@Mistuke, Cabal is `master`. In `cabal-testsuite`, we indeed have quite long CWD to begin with. Making `X:` is a nice trick, have to try that.

But the GHC used, is 8.6.5, will `ghc-jailbreak` work with it, or does one need 8.8?

phadej:
I'd try `ghc-jailbreak` if it works, before making `X:` trick. I recall you blogged about it, but I cannot find that anymore, do you have a link?

Mistuke:
It depends. I'm not quite sure who is issuing this error.. If it's coming from cabal or ghc. `openBinaryTempFileWithDefaultPermissions` sounds like it's coming from base `openBinaryTempFileWithDefaultPermissions` in which case only the X: trick will work until ghc 8.12 because the temp file function in base doesn't use the MAX_PATH workaround and ghc-jailbreak only patches the tools ghc depends on and not ghc itself.

Blog post is at https://hub.zhox.com/posts/bypassing-max_path-limit-on-windows/ but I think it won't work in this case.

Mistuke:
Alternative is to backport the temp file creation from 8.12 to cabal As a compat fix.

phadej:
If `openBinaryTempFileWithDefaultPermissions` can be backported without FFI, then I'm open to that idea.

[...]

PR #6463, opened 2019-12-25
T5677 rename prog to program, case-sensitive type-systems break

phadej:
One case in #6271


phadej:
Backported manually in 364fe0de5

PR #6469, opened 2019-12-28
Use GitHub Actions

phadej:
I'm not removing `.travis.yml` just yet. Let's have it on until 3.2 is done.

I'll cleanup the generation (there's todo about documenting it), will do that early 2020 and then merge. Please comment before that. The setup is simpler, but I'm not 100% if it has same the coverage as current travis setup

- *pro* simpler
- *pro* uses `validate.sh`, so you have better chance to reproduce on own machine
- *con* even closer integration with GitHub
    - yet, the new setup is less GitHub specific than the current setup is Travis specific, so I guess it's not that bad


hvr:
@phadej did you intentionally not include any generated gh-workflow script in the PR (so that we could see the script in action)?

PR #6473, opened 2020-01-02
Doc: v2-building every package requires "all" [ci skip]

Rufflewind:
Calling `v2-build` by itself doesn't work.  Need `all`.


phadej:
Backported in https://github.com/haskell/cabal/pull/6503

PR #6494, opened 2020-01-20
GitHub actions windows

phadej:


phadej:
@Mistuke what might cause

```
      Exception: tests\IntegrationTests2\regression/3324\q\q.cabal.backup: renameFile:renamePath:MoveFileEx "\\\\?\\D:\\a\\cabal\\cabal\\cabal-install\\tests\\IntegrationTests2\\regression\\3324\\q\\q.cabal.backup" Just "\\\\?\\D:\\a\\cabal\\cabal\\cabal-install\\tests\\IntegrationTests2\\regression\\3324\\q\\q.cabal": permission denied (Access is denied.)
```

Writing to that file is ok, is it somehow held open, so cannot be `renameFile`d to?

Mistuke:
> 
> 
> @Mistuke what might cause
> 
> ```
>       Exception: tests\IntegrationTests2\regression/3324\q\q.cabal.backup: renameFile:renamePath:MoveFileEx "\\\\?\\D:\\a\\cabal\\cabal\\cabal-install\\tests\\IntegrationTests2\\regression\\3324\\q\\q.cabal.backup" Just "\\\\?\\D:\\a\\cabal\\cabal\\cabal-install\\tests\\IntegrationTests2\\regression\\3324\\q\\q.cabal": permission denied (Access is denied.)
> ```
> 
> Writing to that file is ok, is it somehow held open, so cannot be `renameFile`d to?
> 
> And how one can force closing etc. properly, so file can be restored?

Yeah, It's usually an indication that the file is open somewhere with an exclusive lock. The lock is only released when the last handle to the file is closed.

The handle may not be opened by you, it may be held by another process (antivirus, or indexing service for instance). I would look at whatever produced the backup file to see if closeHandle was actually called. 

Things that remove/move files we tend to add a retry loop for https://github.com/haskell/hsc2hs/commit/0f68aff2180729ca5ba9569093a35e779230d293 it's short and in most cases won't loop, but if something has it locked it'll usually get freed pretty quickly.

phadej:
@Mistuke thanks for the `hsc2hs` link, that helpful (currently I just do nothing, as it's only tests, so if file isn't not reverted to the original, it's not a big issue).

Windows really makes some "easy" things really cumbersome.

Mistuke:
> 
> Windows really makes some "easy" things really cumbersome.

Yeah, it's all a consequence of layering over time. I tried to get some support for this in stuff like `directory` but some people have been rejecting the idea. I might just do it in Win32.

phadej:
there's `removePathForcibly`, that's already something. If there's similar-ish "best-effort" thing for moving files, it *should* be in `directory` (maybe not as `renameFile`). Reinventing things is very consuming.

PR #6499, opened 2020-01-22
Try to make test for 3324 more robust

phadej:
On Windows moving files fails often enough,
let's try another tactic for recovering modified file.


phadej:
Backported in https://github.com/haskell/cabal/pull/6503

PR #6506, opened 2020-01-23
Copy executables by default on windows.

AndreasPK:
---

* [ x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ x] Any changes that could be relevant to users have been recorded in the changelog.
* [ x] The documentation has been updated, if necessary.
* [ x] If the change is docs-only, `[ci skip]` is used to avoid triggering the build bots.

Running cabal install <something> on windows.

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -140,6 +140,8 @@ import System.Directory
          , removeFile, removeDirectory, copyFile )
 import System.FilePath
          ( (</>), (<.>), takeDirectory, takeBaseName )
+import System.Info
+         ( os )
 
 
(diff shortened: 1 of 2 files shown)


AndreasPK:
CI error seems unrelated to the patch as far as I can tell.

phadej:
I'll take over this. I'd add a `auto` mode, which would behave as copy on windows and symlinks everywhere else, that's so you could **configure** explicitly the default behavior.

Later that can be extended so windows mode could symlink if that's possible (pending @Mistuke's advice on how to recognize that).

I'll also fix the changelog, but first have to try to fix that failing test, as it fails way too often.

phadej:
Merged as part of #6519

Issue #6525, opened 2020-02-04
Can not build cabal with stack


flip111:
```
$ git clone https://github.com/haskell/cabal.git
$ cd cabal
$ stack build --ghc-options="-Wall"

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for cabal-install-3.3.0.0:
    lukko must match >=0.1 && <0.2, but the stack configuration has no specified version  (latest matching version is 0.1.1.1)
    windns must match >=0.1.0 && <0.2, but the stack configuration has no specified version  (latest matching version is 0.1.0.1)
needed since cabal-install is a build target.

In the dependencies for cabal-testsuite-3:
    Cabal-3.3.0.0 from stack configuration does not match ((==2.2.* || ==2.4.* || ==3.0.*) && ==3.3.0.0) && (==2.2.* || ==2.4.* || ==3.0.*)
    aeson-1.2.4.0 from stack configuration does not match ^>=1.4.2.0  (latest matching version is 1.4.6.0)
    async-2.1.1.1 from stack configuration does not match >=2.2.1 && <2.3  (latest matching version is 2.2.2)
    exceptions-0.8.3 from stack configuration does not match >=0.10.0 && <0.11  (latest matching version is 0.10.4)
    optparse-applicative-0.14.2.0 from stack configuration does not match >=0.14.3.0 && <0.15 || >=0.15.1.0 && <0.16  (latest matching version is 0.15.1.0)
    regex-tdfa-1.2.3 from stack configuration does not match ^>=1.2.3.1 || ^>=1.3.1.0  (latest matching version is 1.3.1.0)
    temporary-1.2.1.1 from stack configuration does not match ^>=1.3  (latest matching version is 1.3)
    text-1.2.3.0 from stack configuration does not match ^>=1.2.3.1  (latest matching version is 1.2.4.0)
needed since cabal-testsuite is a build target.

In the dependencies for hackage-security-0.5.3.0:
    Cabal-3.3.0.0 from stack configuration does not match >=1.14 && <2.4  (latest matching version is 2.2.0.1)
needed due to cabal-install-3.3.0.0 -> hackage-security-0.5.3.0

Some different approaches to resolving this:

  * Recommended action: try adding the following to your extra-deps in C:\Users\philippe\haskell\cabal\stack.yaml:

- Cabal-2.2.0.1@sha256:2a80d8fb655474f0eaeb20434c47f64f84e6302e55973056f00df8ca050b9683,20037
[...]

PR #6527, opened 2020-02-04
Update stack resolver

flip111:
For https://github.com/haskell/cabal/issues/6525

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ ctags
 
 # stack artifacts
 /.stack-work/
+stack.yaml.lock
 
 # Shake artifacts
 .shake*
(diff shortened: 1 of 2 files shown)


m-renaud:
The referenced issue was closed before this PR was merged in, was that a mistake? Is this still a problem or can we close this PR?

flip111:
you can merge this PR

m-renaud:
I'll defer to @phadej here, although I noticed that the change uses a nightly resolver and specific hashes for the deps. Would it be possible to select a more recent LTS and released versions of the libraries?

flip111:
> Would it be possible to select a more recent LTS and released versions of the libraries?

I couldn't get this to work. But it might be possible

gbaz:
closing this as outdated. whatever resolver this was set to work against is not nearly the current one anymore. (and in general i think when prs to update resolvers come in, they should be against lts and not nightly if at all possible)

PR #6529, opened 2020-02-06
Use process jobs when spawning subprocesses

bgamari:
Many toolchain tools written for POSIX systems rely on the exec system
call. Unfortunately, it is not possible to implement `exec` in a
POSIX-compliant manner on Windows. In particular, the semantics of the
`exec` implementation provided by the widely-used `msvcrt` C runtime
will cause process's waiting on the `exec`'ing process to incorrectly
conclude that the process has successfully terminated when in fact it is
still running in another process.

For this reason, the `process` library exposes the `use_process_jobs`
flag to use a more strict (although still not POSIX-compliant) mechanism
for tracking process completion. This is explained in this comment [2].

Unfortunately, job support in the `process` library is currently quite
broken and was only recently fixed [1]. Consequently, we only enable job
object support for process releases >= 1.6.8.

[1] https://github.com/haskell/process/pull/168
[2] https://github.com/haskell/process/blob/master/System/Process.hs#L399

### DIFF
--- a/Cabal/Distribution/Simple/Utils.hs
+++ b/Cabal/Distribution/Simple/Utils.hs
@@ -230,13 +230,14 @@ import System.IO.Unsafe
 import qualified Control.Exception as Exception
 
 import Foreign.C.Error (Errno (..), ePIPE)
+import Data.Maybe (fromJust)
 import Data.Time.Clock.POSIX (getPOSIXTime, POSIXTime)
 import Control.Exception (IOException, evaluate, throwIO, fromException)
 import Numeric (showFFloat)
 import qualified System.Process as Process
-         ( CreateProcess(..), StdStream(..), proc)
+         ( CreateProcess(..), StdStream(..), proc, createProcess )
 import System.Process
-         ( ProcessHandle, createProcess, rawSystem, runInteractiveProcess
+         ( CreateProcess, ProcessHandle
          , showCommandForUser, waitForProcess)
 
 import qualified GHC.IO.Exception as GHC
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Simple/Utils.hs
```diff
@@ -777,7 +795,7 @@ createProcessWithEnv ::
 createProcessWithEnv verbosity path args mcwd menv inp out err = withFrozenCallStack $ do
     printRawCommandAndArgsAndEnv verbosity path args mcwd menv
     hFlush stdout
-    (inp', out', err', ph) <- createProcess $
+    (inp', out', err', ph) <- createProcess $ enableProcessJobs
```
phadej: import `System.Process` qualified, and redefine local `createProcess` which would always `enableProcessJobs` if it seems to be the goal.

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -823,7 +841,15 @@ rawSystemStdInOut verbosity path args mcwd menv input _ = withFrozenCallStack $
   printRawCommandAndArgs verbosity path args
 
   Exception.bracket
-     (runInteractiveProcess path args mcwd menv)
+     (do (Just inh,Just outh, Just errh,pid) <-
+           createProcess $ enableProcessJobs
```
phadej: define inline `runInteractiveProcess`, which would be copy of what in `process` with redefined `createProcess`. Add the comment explaining why it's redefined.

Cabal/Distribution/Simple/Utils.hs
```diff
@@ -236,7 +236,7 @@ import Numeric (showFFloat)
 import qualified System.Process as Process
          ( CreateProcess(..), StdStream(..), proc)
 import System.Process
-         ( ProcessHandle, createProcess, rawSystem, runInteractiveProcess
```
phadej: `rawSystem` is not problematic, or should it be redefined as well?

phadej:
```
Distribution/Simple/Utils.hs:689:22: error:
    Not in scope: type constructor or class ‘CreateProces’
    Perhaps you meant ‘Process.CreateProcess’ (imported from System.Process)
    |
689 | enableProcessJobs :: CreateProces -> CreateProcess
```

phadej:
There's small failure in

```haskell
Distribution/Simple/Utils.hs:711:56:
    `Process.delegate_ctlc' is not a (visible) constructor field name
```

but I can take care of it, if the (restarted) OSX job turns out happy.

phadej:
I won't merge this until `process-1.6.8.0` is on Hackage, I have no means to verify this patch works when CPP conditionals are satisfied.

phadej:
Also, is there a way to see difference between `use_process_jobs = True` and `False`, i.e. can we write a test that `createProcess` behaves correctly?

bgamari:
What action is required here? I believe the `process` release is out.

> Also, is there a way to see difference between `use_process_jobs = True` and `False`, i.e. can we write a test that `createProcess` behaves correctly?

You could spawn a process which `exec`s another image which sleeps for a while and then performs some observable side effect (e.g. writing to a file). If `use_process_jobs` works correctly then `waitForProcess` won't return until the process finishes. This isn't perfectly race-free but I'm afraid it's the best we can do.

phadej:
Merged as part of #6536

PR #6531, opened 2020-02-08
Update bootstrap dependency versions.

bryant:
These were needed to work with base-4.13.0.0 which comes with ghc-8.8.2.

Testing done: Successful invocation of `./bootstrap.sh`.

Not sure if this should go into changelog.

---

### DIFF
--- a/cabal-install/bootstrap.sh
+++ b/cabal-install/bootstrap.sh
@@ -230,19 +230,19 @@ TRANS_VER="0.5.5.0";   TRANS_VER_REGEXP="0\.[45]\."
                        # >= 0.2.* && < 0.6
 MTL_VER="2.2.2";       MTL_VER_REGEXP="[2]\."
                        #  >= 2.0 && < 3
-HTTP_VER="4000.3.12";  HTTP_VER_REGEXP="4000\.(2\.([5-9]|1[0-9]|2[0-9])|3\.?)"
+HTTP_VER="4000.3.14";  HTTP_VER_REGEXP="4000\.(2\.([5-9]|1[0-9]|2[0-9])|3\.?)"
                        # >= 4000.2.5 < 4000.4
-ZLIB_VER="0.6.2";      ZLIB_VER_REGEXP="(0\.5\.([3-9]|1[0-9])|0\.6)"
+ZLIB_VER="0.6.2.1";    ZLIB_VER_REGEXP="(0\.5\.([3-9]|1[0-9])|0\.6)"
                        # >= 0.5.3 && <= 0.7
-TIME_VER="1.9.1"       TIME_VER_REGEXP="1\.[1-9]\.?"
+TIME_VER="1.9.3"       TIME_VER_REGEXP="1\.[1-9]\.?"
                        # >= 1.1 && < 1.10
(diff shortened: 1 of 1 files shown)


phadej:
This breaks bootstrap with non-GHC-8.8.

```
Configuring resolv-0.1.1.3...
Setup: Encountered missing or private dependencies:
base ==4.13.*
```

The split of versions in `resolv` "make sense" if you look at the diff between 0.1.1.2 and 0.1.1.3:
```diff
--- resolv-0.1.1.2/src/Network/DNS/Message.hs	2018-10-28 01:43:14.000000000 +0300
+++ resolv-0.1.1.3/src/Network/DNS/Message.hs	2001-09-09 04:46:40.000000000 +0300
@@ -93,7 +93,7 @@
 
 instance Binary CharStr where
     put (CharStr bs)
-      | BS.length bs > 0xff = fail "putString: string too long"
+      | BS.length bs > 0xff = error "putString: string too long"
       | otherwise = do
             putWord8 (fromIntegral $ BS.length bs)
             putByteString bs
@@ -593,7 +593,7 @@
         rdataLen = BSL.length rdata
 
     unless (rdataLen < 0x10000) $
-        fail "rdata too large"
+        error "rdata too large"
 
     putWord16be (fromIntegral rdataLen)
     putLazyByteString rdata
@@ -740,7 +740,7 @@
 
 encodeNsecTypeMap :: Set Type -> Put
 encodeNsecTypeMap bmap = do
-    when (Set.null bmap) $ fail "invalid empty type-map"
+    when (Set.null bmap) $ error "invalid empty type-map"
     -- when (Set.member 0 bmap) $ fail "invalid TYPE0 set in type-map"
```

but I'd advice to collaborate with `resolv` maintainer to rather use CPP then making per-version split. It does complicate `cabal-install` boostrapping, and therefore I cannot accept this patch yet.

phadej:
Note: the changelog actually says

> 0.1.1.3
>
> * GHC 8.8 / base-4.13 only compat hotfix release; the next release will support
>  older GHC/base versions again

So I'd guess that the patch will be very welcome there.

bryant:
I saw that you've asked a related question in https://github.com/haskell-hvr/resolv/issues/7.

My understanding from some very brief research is that ghc-8.8.2 throws hard errors on non-MFP-conforming code, and the version of binary that resolv-0.1.1.3 depends on does not `instance MonadFail PutM`. It's unclear why the preprocessor is not used instead (as you pointed out), and it does not seem like an answer will be provided in reasonable time.

Would it be amenable to work around the resolv dependency? Something like setting `RESOLV_VER` conditionally on `GHC_VER`, for instance.

phadej:
> and it does not seem like an answer will be provided in reasonable time.

This is disrespectful, until this issue https://github.com/haskell-hvr/resolv/issues/7 weren't objectively important.

Rather then adding a hack to cabal, offer a patch to `resolv`. Based on changelog entry I think it will be welcomed.

(FYI, Herbert often misses GitHub notifications, it's relatively easy to read the issue and forget it completely).

phadej:
`resolv` was updated. Amended its version in #6636. CI is happy.

Thanks for writing this PR.

PR #6536, opened 2020-02-12
use process jobs

phadej:
Supersedes #6529


phadej:

PR #6538, opened 2020-02-14
Disallow spaces around colon in Parsec Dependency

phadej:
cc @hvr @fgaz


phadej:
Backported in https://github.com/haskell/cabal/pull/6542

PR #6541, opened 2020-02-14
Accept any pkg-config --modversion output

phadej:
This results e.g. in

    cabal: The pkg-config package 'blas' version ==0.3 || >0.3 is required but the
    version installed on the system is version 0.2.20+ds

output instead of failing to parse 0.2.20+ds as a PkgconfigVersion.
We don't need to parse the `pkg-config --modversion` output
as strictly as the versions in `.cabal` file.
I.e. at all, as `PkgconfigVersion` comparison works somehow
on any possible version strings.

cc @amigalemming if you want to try this out. I'll backport it to 3.2 branch.
You won't be able to write `+` in `.cabal` file, but at least `Cabal` will accept whatever `pkg-config --modversion` gives it.


phadej:
Backported in https://github.com/haskell/cabal/pull/6542

PR #6548, opened 2020-02-19
Change manpage command to man

phadej:
And make it pipe output to `man -l -`.


andreasabel:
Also, on macOS it does not work:
```
$ cabal man 
man: invalid option -- l
man, version 1.6c

usage: man [-adfhktwW] [section] [-M path] [-P pager] [-S list]
	[-m system] [-p string] name ...
...
fd:13: commitBuffer: resource vanished (Broken pipe)

```

Issue #6554, opened 2020-02-23
macOS CI job fails


phadej:
```
PackageTests/Configure/setup.test.hs                                                                              FAIL (2.52s)
$ /opt/ghc/8.8.1/bin/runghc-8.8.1 -- '--ghc-arg=-i' '--ghc-arg=-no-user-package-db' '--ghc-arg=-package-db' '--ghc-arg=/Users/runner/.cabal/store/ghc-8.8.1/package.db' '--ghc-arg=-package-db' '--ghc-arg=/Users/runner/runners/2.165.2/work/cabal/cabal/dist-newstyle-validate-ghc-8.8.1/packagedb/ghc-8.8.1' '--ghc-arg=-package-id' '--ghc-arg=Cabal-3.3.0.0-inplace' '--ghc-arg=-package-id' '--ghc-arg=sync-2.2.2-2f09e089' '--ghc-arg=-package-id' '--ghc-arg=base-4.13.0.0' '--ghc-arg=-package-id' '--ghc-arg=cabal-testsuite-3-inplace' '--ghc-arg=-package-id' '--ghc-arg=clck-0.8-e11eb1cc' '--ghc-arg=-package-id' '--ghc-arg=xcptns-0.10.4-941ff665' '--ghc-arg=-package-id' '--ghc-arg=filepath-1.4.2.1' '--ghc-arg=-package-id' '--ghc-arg=ptprs-pplctv-0.15.1.0-a44bb4db' '--ghc-arg=-package-id' '--ghc-arg=process-1.6.5.1' '--ghc-arg=-package-id' '--ghc-arg=transformers-0.5.6.2' PackageTests/Configure/setup.test.hs --builddir /Users/runner/runners/2.165.2/work/cabal/cabal/dist-newstyle-validate-ghc-8.8.1/build/x86_64-osx/ghc-8.8.1/cabal-testsuite-3 PackageTests/Configure/setup.test.hs --with-ghc ghc-8.8.1
stdout:
+ autoreconf -i
Can't exec "aclocal": No such file or directory at /usr/local/Cellar/autoconf/2.69/share/autoconf/Autom4te/FileUtils.pm line 326.
autoreconf: failed to run aclocal: No such file or directory

stderr:
*** Exception: Command autoreconf -i failed.
Output:
Can't exec "aclocal": No such file or directory at /usr/local/Cellar/autoconf/2.69/share/autoconf/Autom4te/FileUtils.pm line 326.
[...]

PR #6558, opened 2020-02-24
What's wrong with macOS CI

phadej:
Let's see if the #6554 is easily fixable on our side.


phadej:
Looks like macOS iamge doesn't have autotools, yet I'm 100% sure they existed

phadej:
Opened https://github.com/haskell/cabal/pull/6559

PR #6566, opened 2020-02-29
[ci skip] Enable opensearch in Cabal documentation

### LINKED ISSUE #6537: Add OpenSearch description to cabal documentation
[OpenSearch MDN docs](https://developer.mozilla.org/en-US/docs/Web/OpenSearch)

Other haskell systems: GHC User Guide or hackage have it.

This allows to add a search engine to a webrowser (in firefox they appear in 'about:preference#search' in the `One-Click Search Engines` section).

coot:
---

This PR fixed #6537.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -94,7 +94,7 @@ html_theme_options = {
 # If true, an OpenSearch description file will be output, and all pages will
 # contain a <link> tag referring to it.  The value of this option must be the
 # base URL from which the finished HTML is served.
-#html_use_opensearch = ''
+html_use_opensearch = 'https://cabal.readthedocs.io/en/stable'
 
 # This is the file name suffix for HTML files (e.g. ".xhtml").
 #html_file_suffix = None


Cabal/doc/conf.py
```diff
@@ -82,7 +82,7 @@
 # If true, an OpenSearch description file will be output, and all pages will
 # contain a <link> tag referring to it.  The value of this option must be the
 # base URL from which the finished HTML is served.
-#html_use_opensearch = ''
+html_use_opensearch = 'https://haskell.org/cabal/users-guide'
```
phadej: So this won't work on https://cabal.readthedocs.io/en/latest/ ?

coot: Yes, it won't; one needs to set a different value.

coot: This can be made configurable, and expose two `make` commands:
```
make users-docs

make users-docs-readthedocs
```
Any other suggestions?

phadej: readthedocs doesn't call `make`, if i interpret the settings right we can have different `conf.py` for readthedocs. If it's possible to have an override one, for rtd then this would be fine.

Screenshot from 2020-02-29 18-07-38

Yet, I don't have access to cabal's readthedocs configs, so I can't experiment.
ping @hvr @23Skidoo

coot: One can overwrite config parameters on `sphinx-build` command line:
```
sphinx-build -D html_use_opensearch='https://cabal.readthedocs.io/en/latest' doc dist/doc/users-guide
```

ulysses4ever:
@coot do you want to finish this PR? It was blocked because of two URLs being used for docs, but now only readthedocs.io is used (`https://haskell.org/cabal/users-guide` is a simple redirect). Should be pretty easy to update to current `master` and change the URL.

coot:
I updated the PR.  I think the right URL to use is `https://cabal.readthedocs.io/en/stable`, dropping `en/stable` might not work (I am not really sure).

ulysses4ever (approved):
Let’s try it, thanks!

ulysses4ever:
It's been languished for too long and I take the liberty to apply the label. There's still the 2 day delay when people can change something.

PR #6567, opened 2020-02-29
Optimize file generation

sergv:
I noticed that temporary files are generated through `String`s. I thought it might be a good idea to switch to bytestrings and especially builders for building up long strings. Small benchmark for typicall input to `generateCabalMacrosHeader` shows about 3x speedup from `125.3 μs` to `42.47 μs`. Please shout if you'd like to see the benchmark online.

---


Cabal/Distribution/Simple/Utils.hs
```diff
@@ -1380,21 +1381,25 @@ withTempDirectoryEx _verbosity opts targetDir template f = withFrozenCallStack $
 --     ASCII-representable. Since Cabal-3.0 the file is assumed to be
 --     UTF-8 encoded.
 rewriteFileEx :: Verbosity -> FilePath -> String -> IO ()
-rewriteFileEx verbosity path newContent =
+rewriteFileEx verbosity path =
+  rewriteFileExBS verbosity path . toUTF8LBS
+
+-- | Same as `rewriteFileEx` but for 'ByteString's.
+rewriteFileExBS :: Verbosity -> FilePath -> BS.ByteString -> IO ()
```
phadej: This is lazy bytestring, isn't it?

We unfortunately do `import qualified Data.ByteString.Lazy as BS` in this module, but the name of function should have `LBS` prefix.

Cabal/Distribution/Types/PackageName.hs
```diff
@@ -22,7 +22,7 @@ import Distribution.Parsec
 --
 -- @since 2.0.0.2
 newtype PackageName = PackageName ShortText
-    deriving (Generic, Read, Show, Eq, Ord, Typeable, Data)
+    deriving (Generic, Read, Show, Eq, Ord, Typeable, Data, NFData)
```
phadej: Unnecessary change.

I'd rather not do it, as we have to support `deepseq` versio nspan where Generic derivation does different things. Therefore, I'd rather see all `NFData` instances be written explicitly.

Cabal/Distribution/Simple/Build/Macros/Z.hs
```diff
@@ -1,6 +1,12 @@
-{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveGeneric     #-}
```
phadej: This file is autogenerated. I'll add a header to it asap. See https://hackage.haskell.org/package/zinza

This is why CI *Quick jobs* fails.

Cabal/Distribution/Simple/Utils.hs
sergv: Fixed

Cabal/Distribution/Types/PackageName.hs
sergv: I've dropped the commit that introduced the change.

sergv:
> There are good bits: SetupWrapper and `rewriteFileLBS`.
> 
> `cabal_macros.h` generation OTOH have to be rethought. The module is generated for a reason. Technically you could change `ZinzaPrelude` as you did, except have `tell` stil ltake `String` as an argument. I'm not sure that internal conversion to bytestring and then builder would justify added complexity. Generation of `cabal_macros.h` is not that slow, and not in the hot loop.
> 
> I'll cherry-pick the good bits myself. Let's think what to do with the rest.

I've run a quick benchmark across three implementations of current function on a pair of typical inputs to `render`:
```
ZinzaString 1                        mean 127.0 μs  ( +- 2.159 μs  )
ZinzaString 2                        mean 139.7 μs  ( +- 2.632 μs  )
ZinzaBuilder 1                       mean 42.51 μs  ( +- 767.1 ns  )
ZinzaBuilder 2                       mean 45.76 μs  ( +- 885.6 ns  )
ZinzaBuilderStringAPI 1              mean 60.75 μs  ( +- 925.1 ns  )
ZinzaBuilderStringAPI 2              mean 65.54 μs  ( +- 1.044 μs  )
```
`String` is current implementation, `Builder` is my suggested in this PR and `BuilderStringAPI` is your suggestion to keep builder inside but have `tell` only accept strings. I'd rather stay away from `String` as much as possible since just keeping  it in the api is a 30% slowdown.

I don't claim that generation of header is slow and neither it's in the hot loop. However, it's something that always needs to be done when configuring a package and the job is a mundane one of string concatenation. The less time will be spent on it, the more time will be available to other parts of Cabal to do something interesting.

I see your point about compicating the api. But maybe there's a way to at least have something like `tell :: ByteString -> Writer ()` in the ZinzaPrelude?

phadej:
I guess I’m convinced. But then we need to change `zinza` package to work (generate) ByteString/Builder.

sergv:
Okay, I'll change the `zinza` package.

I'm trying to find out what exactly needs to be changed. AFAIU the package just provides a way to turn template into a Haskell function `render :: Type -> a` where meaning of the function and the parameter `a` will be determined by what's in scope. If in proper `ZinzaPrelude` is in scope then any concrete type can be produced instead of `a` and this is independent from the `zinza` package.

On the other hand, the generation of the Haskell function `render` happens within `zinza` using Strings. Should that part be amended to operate on bytestrings and builders? Or only scaffolding for generated modules should be amended, e.g. the Zinza.Writer module, the Zinza.simpleConfig function, etc?

sergv:
I amended `gen-cabal-macros` to generate builder-based `Cabal.Distribution.Simple.Build.Macros.Z` and managed to stick to interface that `zinza` expects. But that required changing template a bit to have a conversion of strings to builders. I have also rebased on latest master. Please let me know your toughts.

boot/cabal_macros.template.h
```diff
@@ -28,12 +28,12 @@
 
 {% if notNull packageKey %}
 #ifndef CURRENT_PACKAGE_KEY
-#define CURRENT_PACKAGE_KEY "{{ packageKey }}"
+#define CURRENT_PACKAGE_KEY "{{ mkBuilder packageKey }}"
```
phadej: I won't accept this. This makes template harder to read/edit for the benefit of non-noticeable speedup (microseconds! not even milliseconds).

phadej:
I don't see a way to get this merged. You try to workaround stuff to make it faster makes it harder to maintain. I cannot accept that trade-off.

sergv:
Ok, I don't have much desire to swim against the stream here.

PR #6580, opened 2020-03-12
Fix js-sources passing in exe builds

yorickvP:
Commit c470154 (#6025) removed buildOrReplExe, but did not port the `js-sources` functionality to the new `gbuild` function. This means `js-sources` was only being passed to GHC on lib builds, but js files were not being linked in on exe builds.

---

I tested building my ghcjs exe with or without this patch and confirmed that the js-bits are now being linked in.

### DIFF
--- a/Cabal/Distribution/Simple/GHCJS.hs
+++ b/Cabal/Distribution/Simple/GHCJS.hs
@@ -950,6 +950,7 @@ decodeMainIsArg arg
 -- | A collection of:
 --    * C input files
 --    * C++ input files
+--    * JS input files
 --    * GHC input files
 --    * GHC input modules
 --
@@ -957,6 +958,7 @@ decodeMainIsArg arg
 data BuildSources = BuildSources {
         cSourcesFiles      :: [FilePath],
         cxxSourceFiles     :: [FilePath],
+        jsSourceFiles      :: [FilePath],
         inputSourceFiles   :: [FilePath],
         inputSourceModules :: [ModuleName]
     }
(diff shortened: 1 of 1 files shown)


yorickvP:
cc: @luite

phadej:
I feel bad accepting pull request which don't introduce any tests. It's impossible to see what's fixed, i.e. this looks arbitrary to me.

phadej:
ping @luite

jneira:
@yorickvP hi, thanks for your pr, any chance to complete it with tests as requested?

yorickvP:
Hi. We no longer use js-sources in our exe's to work around this problem, and I'm not sure how to write tests for this. It looks like the patch no longer applies cleanly.

Mikolaj:
Indeed, I can't see any GHCJS tests in out test suites. The tests probably belong more in GHCJS itself than here.

Can the patch be rescued anyway? Is the patch really arbitrary or is `jsSources` added everywhere it can be? If not, why?

yorickvP:
It looks like the file moved from `Cabal/Distribution/Simple/GHCJS.hs` to `Cabal/src/Distribution/Simple/GHCJS.hs`.
jsSources is currently added arbitrarily to the library codepath but not the exe codepath, this pr fixes that.

Mikolaj:
This is great but, e.g., is it added to all libraries (in case of multiple names libraries), and to tests and benchmarks? In all contexts? I can look at the patch, code-dive and answer such questions, but I can't easily tell if any choice was on purpose and what the purpose was, hence I'm humbly asking, trusting that @phadej indeed saw something there worth testing/documenting/discussing.

Kleidukos:

ffaf1:

yorickvP:
I'm not sure about the current state of the ghcjs ecosystem. This may or may not still be a solution to a problem that may or may not still exist. In any case, I no longer have push rights to the source repository. Closing.

Issue #6575, opened 2020-03-10
Inform user the full path where the executable is copied/symlinked with v2-install


jneira:
Actually `cabal install` output ends with (you can replace `Copying` with `Symlinking`)
```
Copying 'executable1.exe'
Copying 'executable2.exe'
...
```
 and `cabal install -v3` with:
```
creating D:\bin 
Copying 'executable1.exe'
Copying 'executable2.exe'
....
```
being `D:\bin` the dir where the executables are being copied/symlinked. Even when the directory already exists (but that is not the goal of this issue)!

I think it would be great show the absolute path where the artifacts are being created even with the default verbosity:
```
Copying 'executable1.exe' to 'D:\bin'
Copying 'executable2.exe' to 'D:\bin'
```
or, if all of them only could be created in one location (not sure about that)
```
Copying executables to 'D:\bin'
Copying 'executable1.exe'
Copying 'executable2.exe'
```

It seems a small change but the directory value can come from several sources and beginners not always are aware of them.
I think bug fixes should be the priority here but this seems to be easy to implement (ha! :stuck_out_tongue_closed_eyes:)

emlautarom1:
Could you point to where these messages are generated? I would like to look into this. Seems fairly easy

jneira:
I think this is the main functions where several executables are copying/symlinking:
https://github.com/haskell/cabal/blob/2e03231974a5be0e4d1a64ad8481230e21d87a45/cabal-install/Distribution/Client/CmdInstall.hs#L798-L812
It calls this for each one:
https://github.com/haskell/cabal/blob/2e03231974a5be0e4d1a64ad8481230e21d87a45/cabal-install/Distribution/Client/CmdInstall.hs#L826-L834

emlautarom1:
I'll take a look into it

phadej:
> What about Symlinking? should we also show were we are linking files?

Yes

> Displaying the...

Let's not parameterize that output by verbosity. Then it should be one line change (if tests don't need to be updated).

emlautarom1:
I think it's done, please, take a look at: https://github.com/haskell/cabal/pull/6582

phadej:
Fixed in #6582 + #6590

jneira:
Thanks @emlautarom1 and @phadej (it is clear that maintaining a package like cabal is really, really hard)!

PR #6582, opened 2020-03-15
Show resulting executable path when running 'cabal install'

emlautarom1:
Tested the changes by installing and symlinking some programs. The output looks like the discussed in #6575 

Don't know where should I add the `ci skip` to avoid triggering a rebuild, so a maintainer should probably take a look into that

---

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -835,7 +835,7 @@ installBuiltExe
 installBuiltExe verbosity overwritePolicy
                 sourceDir exeName finalExeName
                 installdir InstallMethodSymlink = do
-  notice verbosity $ "Symlinking '" <> exeName <> "'"
+  notice verbosity $ "Symlinking '" <> exeName <> "' to '" <> installdir <> "'"
   symlinkBinary
     overwritePolicy
     installdir
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -845,7 +845,7 @@ installBuiltExe verbosity overwritePolicy
 installBuiltExe verbosity overwritePolicy
                 sourceDir exeName finalExeName
                 installdir InstallMethodCopy = do
-  notice verbosity $ "Copying '" <> exeName <> "'"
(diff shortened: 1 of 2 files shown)


jneira:
LGTM! Thanks for taking care of it.  😄

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -835,7 +835,7 @@ installBuiltExe
 installBuiltExe verbosity overwritePolicy
                 sourceDir exeName finalExeName
                 installdir InstallMethodSymlink = do
-  notice verbosity $ "Symlinking '" <> exeName <> "'"
+  notice verbosity $ "Symlinking '" <> exeName <> "' to '" <> installdir <> "'"
```
phadej: I would rather print `installdir </> finalExeName` (i.e. `destination` in `installBuiltExe`). 

This is because one might link `foo` into `/some/dir/foo-1.0.0` if there is an executable executable suffix asked for.

cabal-install/changelog
```diff
@@ -16,6 +16,7 @@
 	  --package-env` (#6298)
 	* `cabal v2-run` works with `.lhs` files (#6134)
 	* `subdir` in source-repository-package accepts multiple entries (#5472)
+	* 'cabal install' shows the resulting executable full path (symlinking and copying) (#6575)
```
phadej: It won't be in `3.2.0.0`. Omit changelog entry, I'll add it afterwards myself.

cabal-install/Distribution/Client/CmdInstall.hs
emlautarom1: Should we also include the `finalExeName` when copying, or only when symlinking?

cabal-install/changelog
emlautarom1: Removed as requested

cabal-install/Distribution/Client/CmdInstall.hs
phadej: Yes, by the same reasoning.

emlautarom1: Done!

phadej:
Merged squashed in https://github.com/haskell/cabal/pull/6590. Thanks!

PR #6603, opened 2020-03-22
Don't write Setup.hs file during cabal init.

m-renaud:
When `build-type: Simple` the contents of Setup.hs MUST BE standard. It is only
in advanced cases where a custom setup file is needed.

This also has the positive consequence of simplifying the directory contents created during package initialization.

---

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -130,7 +130,6 @@ initCabal verbosity packageDBs repoCtxt comp progdb initFlags = do
   case license initFlags' of
     Flag PublicDomain -> return ()
     _                 -> writeLicense initFlags'
-  writeSetupFile initFlags'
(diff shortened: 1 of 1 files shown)


DanielG:
Could you explain what exactly you're trying to accomplish here?

While `Setup.hs` isn't technically required to build a package with `cabal` some distributions such as Debian still use it to build haskell packages directly from tarballs, without `exe:cabal` involved.

phadej:
@DanielG they could (and should) use standard `Setup.hs`. There are packages on Hackage which don't have `Setup.hs` inside them.

DanielG:
Huh, I'd assumed we're enforcing `Setup.hs` existing in hackage-serever. Well nevermind then.

m-renaud:
Any other concerns or should I go ahead and merge this in? CI is all green.

phadej:
@m-renaud go ahead.

PR #6605, opened 2020-03-22
Mark public libs as an experimental feature

phadej:
Needs backport to whatever branch `hackage-server` would pick up.


phadej:
bonkers, I added `CABAL_CONFIG` to `testEnvironment` to isolate test runner a bit, and now GHC-7.10 tests fail - as they need `hackage.haskell.org` (for `old-time`). Argh. I'll come back to this.

Issue #6194, opened 2019-08-17
RFC: Improve and restructure Cabal user guide


m-renaud:
Hey all, I wanted to start a discussion about ways the [current user guide](https://www.haskell.org/cabal/users-guide/index.html) can be improved. I felt like now may be a good time since the release of cabal-install 3.0 is imminent, and comes with a bunch of usability improvements from the v2-* commands.

### Summary of Observed Issues

- Lack of a to-the-point "Quick Start" at the beginning of the user guide
- Dual meaning of "Cabal"
- Lack of focus on "private/project-based workflows"
- Redundant/outdated information across old-style/new-style docs

#### Quick Start

Currently, if I'm simply looking for a 1-2 sentence summary of what cabal (read: cabal-install) is and how to get started with it, I can't find one. This is especially troublesome for folks new to Haskell and its build tools who have very little if any pre-requisite knowledge.

To start, a web search for "haskell cabal quick start" gives you:
1) A link to section 3.1 (Developing Packages) which is definitely not a "quick start" since it already "assume(s) we have created a project directory and already have a Haskell module or two".
2) A link to this repo
3) Blog post
4) Wiki page (unrelated)

None of these are particularly useful. Playing around with search terms you can eventually find the [Introduction](https://www.haskell.org/cabal/users-guide/intro.html#introduction) section which has tonnes of details and gets into the weeds pretty quickly (like that "Cabal packages use the standard compressed tarball format"), with no information for "how to get cabal" or "how to use cabal". You may also find the [Nix-style local builds quickstart](https://www.haskell.org/cabal/users-guide/nix-local-build.html#quickstart) section which neither sounds "quick start"-ey or is worded as such. It starts with "Suppose that you are in a directory containing a single Cabal package which you wish to build", how am I supposed to get that directory?

In comparison, a search for "haskell stack quick start" prominently features https://docs.haskellstack.org/en/stable/README/ which has all the info you need: a short summary of _what it is_, _how to get it_, and _how to use it_ which all fits within a couple screens and doesn't have superfluous information. Similar docs can be found for build tools of other languages as well (Rust, Go, Elm, etc.)

#### Dual meaning of "Cabal"

I'm sure its been mentioned before, but the fact that "Cabal" can refer to either "a package system for Haskell software" or "a tool for working with packages" can definitely cause confusion, especially those new to the ecosystem. I obviously am not proposing changing the name of either of these, just pointing out that there's ambiguity which makes it very important how content is introduced.

[...]

PR #6606, opened 2020-03-23
Update installing-packages.rst to no longer use setup script.

m-renaud:
Update command references throughout to use `cabal <cmd>` instead of `runhaskell
Setup.hs <cmd>`.

It looks like these docs were imported >4 years ago from the previous Pandocs,
which means these docs are almost certainly out of date.

This partially addresses #6194.

[ci skip]

---

### DIFF
--- a/Cabal/doc/installing-packages.rst
+++ b/Cabal/doc/installing-packages.rst
@@ -263,14 +263,23 @@ or
 
     $ runhaskell Setup.hs --help
 
+
+Installing libraries locally
+----------------------------
+
+If you are developing a Haskell package and you need to use a library from
+Hackage_ all you need to do is add it to the ``build-depends`` attribute in
+your cabal file.
+
+
 Building and installing a system package
 ----------------------------------------
 
 ::
 
-    $ runhaskell Setup.hs configure --ghc
-    $ runhaskell Setup.hs build
-    $ runhaskell Setup.hs install
+    $ cabal configure --ghc
+    $ cabal build
+    $ cabal install
 
 The first line readies the system to build the tool using GHC; for
 example, it checks that GHC exists on the system. The second line
@@ -282,26 +291,24 @@ Building and installing a user package
 
 ::
 
-    $ runhaskell Setup.hs configure --user
-    $ runhaskell Setup.hs build
-    $ runhaskell Setup.hs install
+    $ cabal configure --user
+    $ cabal build
+    $ cabal install
 
 The package is installed under the user's home directory and is
-registered in the user's package database (:option:`setup configure --user`).
+registered in the user's package database (:option:`cabal configure --user`).
 
-Installing packages from Hackage
---------------------------------
+Installing executable packages from Hackage
+-------------------------------------------
 
 The ``cabal`` tool also can download, configure, build and install a
-Hackage_ package and all of its
-dependencies in a single step. To do this, run:
+Hackage_ package and all of its dependencies in a single step:
 
 ::
 
    $ cabal install [PACKAGE...]
 
-To browse the list of available packages, visit the
-Hackage_ web site.
+To browse the list of available packages, visit the Hackage_ web site.
 
 
 Creating a binary package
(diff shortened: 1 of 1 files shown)


m-renaud:
Friendly ping @phadej, any concerns here? Going forward, would you like me to wait for your review on doc changes or just go ahead and merge them?

m-renaud:
Agreed that there should be some mention of setup somewhere, but the places that I changed *are* the command references for `cabal <cmd>` (they're just super outdated and used the name `setup <cmd>`). The Nix style build docs do have some mention of these commands but most users likely go to the [Building and installing packages] section to look for information. For example, `cabal haddock` appears exactly zero times in the docs outside of the Nix integration sections; previously it was `setup haddock` (which no one is going to search for), same with `cabal clean`.

For the places where there are differences between the v1- and v2- commands I was going to update them in my next PR, this is just a disambiguation step.

[Building and installing packages]: https://www.haskell.org/cabal/users-guide/installing-packages.html#building-and-installing-packages

phadej:
No, I think `setup` refers to the program you get by `ghc -o setup ./Setup.hs`. It **does** document those commands. (And `v1-...` commands are close supersets).

I cannot accept this PR simply because

```
program:: cabal configure
```

means now the `v2-configure`, not `v1-configure`.

The only option's I'll accept are either:
- Change `setup` to `runhaskell Setup.hs` everywhere
- Clarify that `setup` means the program resulting from compiling `Setup.hs` (e.g. `ghc -o setup Setup.sh`)
    - With (or maybe better without) mention that `v1-` commands closely mimic `Setup.hs` interface, but there might be subtle differences.

---

There are three interfaces:
1. `./Setup.hs` interface
2. `v1-build`
3. `v2-build`

Two first are similar, but not exactly the same. First one is accessible from `cabal` via `cabal act-as-setup`. Let's not waste any effort on documenting second one, it's going away.

m-renaud:
Ah I see what you're saying. I guess I was just surprised that many of the `cabal <cmd>` commands have no command-reference docs at all if you do a search for the term (here's one for "cabal haddock": https://www.google.com/search?q=%22cabal+haddock%22+site%3Ahttps%3A%2F%2Fwww.haskell.org%2Fcabal%2Fusers-guide%2F).

I don't think changing `setup` to `runhaskell Setup.hs` is an improvement since that shouldn't be the first piece of documentation that readers see about the commands. It just seems strange to first talk about `ghc -o setup Setup.sh` before covering the basic commands like `cabal build`, `cabal test`, and `cabal clean` which is the common user-facing interface.

It looks like this whole page needs to be re-written (from ~scratch?) to focus on `cabal <cmd>` where `<cmd>` is the `v2-` variant, and then move the info about setup to some other section.

That being said, for some of these the wording appears mostly correct for both versions: `cabal hscolor`, `cabal haddock`, `cabal clean` (modulo `dist-newstyle`), `cabal v1-copy` (doesn't appear to exist in v2, is that intentional?) all seem like there's nothing `setup` specific, they're just general descriptions of the commands.

> There are three interfaces ... Two first are similar, but not exactly the same

Are any of their differences actually called out in the docs here? If not I think it would be clearer just to switch the docs here to refer to the v1 commands, especially since the default Setup.hs is the predominant approach in use now. From there we can update the command docs from v1->v2 on a command-by-command basis.

So, counter-proposal (not necessarily in this PR):
1) Change PR to explicitly prefix these with `v1-`, since afaik there isn't anything that actually exists in the docs here that is "setup" specific.
2) Remove the `v1-` prefix from commands that are about the same (such as hscolor and haddock).
3) Tackle the common commands first to update them to v2 (build, test, bench, clean).
4) Deal with the less common commands.

phadej:
> I don't think changing setup to runhaskell Setup.hs is an improvement since that shouldn't be the first piece of documentation that readers see about the commands. 

I agree, starting from `Setup.hs` is wrong. But `Setup.hs` docs **must** exist. They document low-level interface.

> Are any of their differences actually called out in the docs here? If not I think it would be clearer just to switch the docs here to refer to the v1 commands, especially since the default Setup.hs is the predominant approach in use now.

As I said. No new documentation for `v1-` commands.

phadej:
I split the reference parts from tutorial parts in https://github.com/haskell/cabal/commit/3752c3e5edc3c19b7262cdb0b70f4e0539b46362, which is mostly just moving code around.

The 

```rst
Building and installing packages
================================

To be written
```

section looks like that now.

PR #6607, opened 2020-03-23
Default to cabal version 2.4 when running cabal init.

m-renaud:
Cabal version 2.4 is over 2 years old at this point and has numerous additions
compared to the current default 1.10.

There are numerous references to the `^>=` operator throughout the docs (nix-local-build.rst and developing-packages.rst) but this does not work in version 1.10. This results in unnecessary confusion for users which then must manually edit the cabal file to change the version. To add to the confusion, the format written to the cabal file (`cabal-version: >=1.10`) only works for this version, to use a newer version you need to remove the `>=` which is undocumented and confusing.

Lets just default to a more recent version. 

---

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -850,7 +850,7 @@ commentSavedConfig = do
             },
         savedInitFlags       = mempty {
             IT.interactive     = toFlag False,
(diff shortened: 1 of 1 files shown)


m-renaud:
Hmm, not sure why one of the CI checks failed (was cancelled) :/ This PR should be safe to merge since it only changes a value within the `cabal init` command. @phadej should the CI be re-run (seems not useful) or can it just be merged in?

PR #6614, opened 2020-03-25
PackageOutput: Dump path to local dir for source-repository-package

DanielG:
This allows external tools to find the directory the source repository got
cloned into.

Needed to fix DanielG/cabal-helper#99

---

### DIFF
--- a/cabal-install/Distribution/Client/ProjectPlanOutput.hs
+++ b/cabal-install/Distribution/Client/ProjectPlanOutput.hs
@@ -193,10 +193,13 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
             J.object [ "type" J..= J.String "repo-tar"
                      , "repo" J..= repoToJ repo
                      ]
-          RemoteSourceRepoPackage srcRepo _ ->
-            J.object [ "type" J..= J.String "source-repo"
-                     , "source-repo" J..= sourceRepoToJ srcRepo
-                     ]
+          RemoteSourceRepoPackage srcRepo local ->
+            J.object $
+              [ "type" J..= J.String "source-repo"
+              , "source-repo" J..= sourceRepoToJ srcRepo
+              ] ++
+              [ "path" J..= J.String path | path <- maybeToList local ]
+
 
       repoToJ :: Repo -> J.Value
       repoToJ repo =


cabal-install/Distribution/Client/ProjectPlanOutput.hs
```diff
@@ -193,10 +193,13 @@ encodePlanAsJson distDirLayout elaboratedInstallPlan elaboratedSharedConfig =
...
-          RemoteSourceRepoPackage srcRepo _ ->
-            J.object [ "type" J..= J.String "source-repo"
-                     , "source-repo" J..= sourceRepoToJ srcRepo
-                     ]
+          RemoteSourceRepoPackage srcRepo local ->
+            J.object $
+              [ "type" J..= J.String "source-repo"
+              , "source-repo" J..= sourceRepoToJ srcRepo
+              ] ++
+              [ "path" J..= J.String path | path <- maybeToList local ]
```
phadej: is this cloned or `sdist`d path?

phadej: I think the field should be `local-tarball`, and it should be also added for `RemoteTarballPackage`.

DanielG: It's the path to the locally cloned repository. It's got nothing to do with tarballs, here's an example:

```
  "pkg-src": {
    "type": "source-repo",
    "source-repo": {
      "type": "git",
      "location": "../../../../",
      "tag": "HEAD",
      "subdir": "tests/exelib"
    },
    "path": "/home/dxld/share/dev/hs/cabal-helper/tests/src-repo/dist-newstyle/src/.-6e043ec6becad825/tests/exelib"
  },
```

DanielG: For tarballs what I'd want is the directory they get unpacked into, which I'm not so sure is available at this stage. I can at least add the path to the tarball though if you like.

DanielG:
Ah, I didn't realize that I'll have to keep digging then.

phadej:
Oh wait. Maybe it's how we should do things (so `source-repository-package` are "non local"), but we don't yet.

DanielG:
Yeah, the bug I mentioned happens because I filter for `"style": "local"` but then get these source-repo packages. That's probably why.

PR #6618, opened 2020-03-27
4267 Add --exact flag to the list command

### LINKED ISSUE #4267: better for cabal list to default to exact match or have an --exact option
I would like to suggest changing the behaviour of `cabal list PKG` to list only the info for the exact match `PKG` by default and not all packages containing the substring `PKG`, and adding a `--all` option to get back the current behaviour.  Alternatively for better backward compatibility an `--exact` option could be added for this effect, leaving the current default behaviour of listing all matches.

Looking at https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/List.hs this seems possible though the code will need a little tweaking.

v0d1ch:
fixes #4267 

This PR adds `--exact` flag to the cabal list command which
enables exact match on the package name while searching.

The cabal list output looks like this for `mtl` package:
```
cabal list mtl

* AbortT-mtl
    Synopsis: mtl instances for the AbortT monad transformer
    Default available version: 1.0.0.1
    Installed versions: [ Not installed ]
    Homepage: http://github.com/gcross/AbortT-mtl
    License:  BSD3

* Hsmtlib
    Synopsis: Haskell library for easy interaction with SMT-LIB 2 compliant
              solvers.
    Default available version: 2.8.8.8
    Installed versions: [ Not installed ]
    Homepage: https://github.com/MfesGA/Hsmtlib
    License:  BSD3

* MonadCatchIO-mtl
    Synopsis: Monad-transformer version of the Control.Exception module
    Default available version: 0.3.1.0
    Installed versions: [ Not installed ]
    Homepage: http://darcsden.com/jcpetruzza/MonadCatchIO-mtl
    License:  BSD3
...

```
And when using the `--exact` flag we are matching the exact package name like this:
```
cabal list mtl --exact
* mtl
    Synopsis: Monad classes, using functional dependencies
    Default available version: 2.2.2
    Installed versions: 2.2.2
    Homepage: http://github.com/haskell/mtl
    License:  BSD3

```

### DIFF
--- a/cabal-install/Distribution/Client/List.hs
+++ b/cabal-install/Distribution/Client/List.hs
@@ -112,12 +112,12 @@ getPkgList verbosity packageDBs repoCtxt comp progdb listFlags pats = do
           [(PackageName, [Installed.InstalledPackageInfo], [UnresolvedSourcePackage])]
         pkgsInfoMatching =
           let matchingInstalled = matchingPackages
-                                  InstalledPackageIndex.searchByNameSubstring
+                                  ipiSearch
                                   installedPkgIndex
               matchingSource   = matchingPackages
                                  (\ idx n ->
                                    concatMap snd
-                                   (PackageIndex.searchByNameSubstring idx n))
+                                   (piSearch idx n))
                                  sourcePkgIndex
           in mergePackages matchingInstalled matchingSource
 
@@ -131,6 +131,11 @@ getPkgList verbosity packageDBs repoCtxt comp progdb listFlags pats = do
     return matches
   where
     onlyInstalled = fromFlag (listInstalled listFlags)
+    exactMatch = fromFlag (listExactMatch listFlags)
+    ipiSearch | exactMatch = InstalledPackageIndex.searchByNameExact
+              | otherwise  = InstalledPackageIndex.searchByNameSubstring
+    piSearch  | exactMatch = PackageIndex.searchByNameExact
+              | otherwise  = PackageIndex.searchByNameSubstring
     matchingPackages search index =
       [ pkg
       | pat <- pats
--- a/cabal-install/Distribution/Solver/Types/PackageIndex.hs
+++ b/cabal-install/Distribution/Solver/Types/PackageIndex.hs
@@ -39,6 +39,7 @@ module Distribution.Solver.Types.PackageIndex (
   searchByName,
   SearchResult(..),
   searchByNameSubstring,
+  searchByNameExact,
 
   -- ** Bulk queries
   allPackages,
@@ -312,9 +313,23 @@ data SearchResult a = None | Unambiguous a | Ambiguous [a]
 --
 searchByNameSubstring :: PackageIndex pkg
                       -> String -> [(PackageName, [pkg])]
-searchByNameSubstring (PackageIndex m) searchterm =
+searchByNameSubstring =
+  searchByNameInternal False
+
+searchByNameExact :: PackageIndex pkg
+                  -> String -> [(PackageName, [pkg])]
+searchByNameExact =
+  searchByNameInternal True
+
+searchByNameInternal :: Bool
+                     -> PackageIndex pkg
+                     -> String -> [(PackageName, [pkg])]
+searchByNameInternal exactMatch (PackageIndex m) searchterm =
     [ pkgs
     | pkgs@(pname, _) <- Map.toList m
-    , lsearchterm `isInfixOf` lowercase (unPackageName pname) ]
+    , if exactMatch
+        then searchterm == unPackageName pname
+        else lsearchterm `isInfixOf` lowercase (unPackageName pname)
+    ]
   where
     lsearchterm = lowercase searchterm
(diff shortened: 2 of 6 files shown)


cabal-testsuite/Test/Cabal/Monad.hs
```diff
@@ -412,7 +412,7 @@ mkNormalizerEnv = do
     -- Arguably we should use Cabal's APIs but I am too lazy
     -- to remember what it is
     list_out <- liftIO $ readProcess (programPath ghc_pkg_program)
-                      ["list", "--global", "--simple-output"] ""
+                      ["list", "--global", "--simple-output", "--exact"] ""
```
phadej: `ghc-pkg` doesn't have `--exact` argument.

Cabal/Distribution/Simple/Program/HcPkg.hs
```diff
@@ -473,7 +473,7 @@ listInvocation hpi _verbosity packagedb =
       progInvokeOutputEncoding = IOEncodingUTF8
     }
   where
-    args = ["list", "--simple-output", packageDbOpts hpi packagedb]
+    args = ["list", "--simple-output", "--exact", packageDbOpts hpi packagedb]
```
phadej: ditto, `ghc-pkg` doesn't have `--exact` flag

cabal-install/Distribution/Client/List.hs
```diff
@@ -106,11 +106,11 @@ getPkgList verbosity packageDBs repoCtxt comp progdb listFlags pats = do
                         (         PackageIndex.allPackages sourcePkgIndex)
 
             -- gather info for packages matching search term
-          | otherwise = pkgsInfoMatching
+          | otherwise = pkgsInfoMatching exactMatch
 
         pkgsInfoMatching ::
-          [(PackageName, [Installed.InstalledPackageInfo], [UnresolvedSourcePackage])]
-        pkgsInfoMatching =
+          Bool -> [(PackageName, [Installed.InstalledPackageInfo], [UnresolvedSourcePackage])]
```
phadej: You don't need to pass `Bool` here. `exactMatch` would be visible as the `where` is shared across all of `getPkgList`

cabal-install/Distribution/Client/List.hs
```diff
@@ -131,6 +134,13 @@ getPkgList verbosity packageDBs repoCtxt comp progdb listFlags pats = do
     return matches
   where
     onlyInstalled = fromFlag (listInstalled listFlags)
+    exactMatch = fromFlag (listExactMatch listFlags)
+    filterPackages :: [(PackageName, a, b)] -> [(PackageName, a, b)]
+    filterPackages =
+      filter
```
phadej: Don't filter, rather define

```haskell
ipiSearch | exactMatch = InstalledPackageIndex.searchByNameExact
          | otherwise  = InstalledPackageIndex.searchByNameSubstring
```

and similarly for PackageIndex variant.

Note: `PackageIndex.searchByName` uses case-insensitive matching. I think we want exact, therefore need to do special variant. (Should be easy).

v0d1ch:
Pinging Dr. @phadej  again

cabal-install/changelog
```diff
@@ -1,6 +1,9 @@
 -*-change-log-*-
 
 3.2.0.0 Herbert Valerio Riedel <hvr@gnu.org> April 2020
+3.2.0.0 Someone <mailto:somewhere@example.com> February 2020
```
phadej, line 4: Add fine to `changelog.d` directory, see the `issue-*` files for an examples

It's recent change, so we won't introduce conflicts to changelog files.

solver-benchmarks/HackageBenchmark.hs
```diff
@@ -164,7 +164,7 @@ hackageBenchmarkMain = do
           if null argPackages
           then do
             putStrLn $ "Obtaining the package list (using " ++ argCabal1 ++ ") ..."
-            list <- readProcess argCabal1 ["list", "--simple-output"] ""
+            list <- readProcess argCabal1 ["list", "--simple-output", "--exact"] ""
```
phadej, line 167: This doesn't seem to be related to the reported issue. Why this change? Does `exact` makes it somehow faster?

cabal-install/Distribution/Solver/Types/PackageIndex.hs
```diff
@@ -312,9 +313,23 @@ data SearchResult a = None | Unambiguous a | Ambiguous [a]
...
+
+searchByNameInternal :: Bool
+                     -> PackageIndex pkg
+                     -> String -> [(PackageName, [pkg])]
+searchByNameInternal exactMatch (PackageIndex m) searchterm =
     [ pkgs
     | pkgs@(pname, _) <- Map.toList m
-    , lsearchterm `isInfixOf` lowercase (unPackageName pname) ]
+    , if exactMatch
+        then searchterm `isInfixOf` unPackageName pname
```
phadej, line 331: why still `isInfixOf` ?

phadej:
Merged in #6681. Thanks!

PR #6619, opened 2020-03-27
Cabal init prints error message forever when using invalid license

### LINKED ISSUE #6227: cabal init -l BSD-3-Clause prints error message forever
**Describe the bug**

```
$ cabal init -l BSD-3-Clause

The license must be alphanumeric. If your license name has many words, the convention is to use camel case (e.g. PublicDomain). Please choose a different license.

The license must be alphanumeric. If your license name has many words, the convention is to use camel case (e.g. PublicDomain). Please choose a different license.

The license must be alphanumeric. If your license name has many words, the convention is to use camel case (e.g. PublicDomain). Please choose a different license.

...
```
`cabal init -l a-1` gives the same repeating error.
`cabal init -l a-1-a` gives the same repeating error.
`cabal init -l a-a` does not, it returns
```
Cannot parse license: a-a-1
  error, called at ./Distribution/ReadE.hs:42:24 in Cabal-3.0.0.0-fb617b94c4973e1d5149e5c5439c47b4e57443a0819f354a288475aa6cef5eae:Distribution.ReadE
```

**System information**
 - Ubuntu `19.04`
 - cabal-install `3.0.0.0`
 - ghc `8.6.5`

v0d1ch:
fixes #6227 
When choosing non adequate license error message is printed in a loop.
I fixed that behavior and now we get something like this:
```
The license must be alphanumeric. If your license name has many words, the convention is to use camel case (e.g. PublicDomain). Please choose a different license.

Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...

```
The issue where the user sees the exception on failed parsing is still there:
```
Cannot parse license: a-a-1
  error, called at ./Distribution/ReadE.hs:42:24 in Cabal-3.0.0.0-fb617b94c4973e1d5149e5c5439c47b4e57443a0819f354a288475aa6cef5eae:Distribution.ReadE
```
I don't know if we want to prevent this from happening. 
In my mind users should't see these kinds of exceptions but maybe it should be a separate issue.

### DIFF
--- a/cabal-install/Distribution/Client/Init.hs
+++ b/cabal-install/Distribution/Client/Init.hs
@@ -278,9 +278,8 @@ getLicense flags = do
                   (maybePrompt flags
                     (promptList "Please choose a license" listedLicenses
                      (Just BSD3) displayLicense True))
-
   case checkLicenseInvalid lic of
-    Just msg -> putStrLn msg >> getLicense flags
+    Just msg -> putStrLn msg >> return (flags { license = maybeToFlag lic })
     Nothing  -> return $ flags { license = maybeToFlag lic }
 
   where


m-renaud:
Hey @v0d1ch, thanks for taking a look at this (and sorry for the delay)! With your change what's the behaviour if init is run in interactive mode without a license being specified from the CLI? The current behaviour is it will re-prompt if you don't select a valid choice, for example:

```
Please choose a license:
   1) GPL-2.0-only
  ...
  13) NONE
  14) Other (specify)
Your choice? [default: BSD-3-Clause] 15
15 is not a valid choice.
Your choice? [default: BSD-3-Clause] 
```

Agreed that hiding the parse error message and showing something more user friendly is out of scope here, it would require changing the command line argument parsing in [`Client/Setup.hs`]. I recently split up `Init.hs` into sub-modules so you'll need to rebase your changes on top of master and update [`Client/Init/Command.hs`].

[`Client/Setup.hs`]: https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Setup.hs#L2337
[`Client/Init/Command.hs`]: https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Init/Command.hs#L279

phadej:
I think https://github.com/haskell/cabal/pull/6661 fixes the looping issue as a side-effect.

The command line parser doesn't produce an `UnknownLicense str` anymore which caused later code to loop.

Issue #5734, opened 2018-11-26
`cabal format` inlines and removes `common` stanzas


m-renaud:
# Overview

When running `cabal format`, all `common` stanzas are inlined into the blocks they're referenced from.

## Example

**Before:**

```
cabal-version: 2.4
name: inittest
version: 0.1.0.0
build-type: Simple

common common-deps
       default-language: Haskell2010
       build-depends: base >=4.12 && <4.13

library
    import: common-deps
    exposed-modules:
        MyLib
    hs-source-dirs: src

executable inittest
    import: common-deps
    main-is: Main.hs
    hs-source-dirs: app
    build-depends:
        inittest -any
```

**After:**

```
cabal-version: 2.4
name: inittest
version: 0.1.0.0
build-type: Simple

library
    exposed-modules:
        MyLib
    hs-source-dirs: src
    default-language: Haskell2010
    build-depends:
        base >=4.12 && <4.13

executable inittest
    main-is: Main.hs
    hs-source-dirs: app
    default-language: Haskell2010
    build-depends:
        base >=4.12 && <4.13,
        inittest -any
```

/cc @phadej @hvr

phadej:
This is by design. Parser produces a `GPD` to be consumed by solver/`Setup.hs`/...

Asking to preserve common stanzas is in the same spirit as asking GHC to be able to pretty-print *Core* into *surface Haskell language* (analogy: type class instances, there are no such things in Core).

That said, having more intermediate structures would be reasonable; so e.g. `cabal check` could work on more "surface syntax looking" data (and importantly preserve the source locations). But that should not affect performance of parsing. This is massive task to do, I don't see clear iterative path.

phadej:
I label this as *wontfix*. Please don't be discouraged, If someone have time and willingness to pursuit this task, be my guest. But I don't feel this is important enough to have any priority.

m-renaud:
Thanks for the info. I feel like this behaviour living under the command `cabal format` doesn't make any sense then, the name implies it should "format my cabal files" not "reprint my cabal files after reducing them to a GPD, then make me undo the change so that they're actually well structured again".

phadej:
So what should we do, any suggestions?

m-renaud:
My 2cents: 

1) Rename the `format` command to something more specific/targeted to its use case, maybe something like `simplify-cabal-files` (bikeshed away) and keep it hidden. Update the docs to reflect that this is not meant for re-formatting your cabal files.

[...]

Issue #6187, opened 2019-08-10
Use Distribution.Fields.Pretty in `cabal init`.


phadej:
This will make output of `cabal init` and `cabal format` closer to each other.

phadej:
One way to solve this is to first generate `GenericPackageDescription` of a new package,
process it down to `[PrettyField ()]` and attach comments to relevant fields in ad-hoc post-processing step.

This won't if `cabal init` ought to generate common stanzas, or some other structure not visible in GPD.

This way `cabal format` on non-commented `cabal init` file would be no-op.

---

An alternative, to achieve `cabal format` and `cabal init` compatibility, is to change `cabal format` to work like `cabal-fmt`, i.e. only on `Field`/`PrettyField`; not going down to GPD at all. If someone thinks that's a way to go, please open an issue and argue why "through GPD" is not required.

Issue #5555, opened 2018-08-31
cabal init generates badly formatted file


osa1:
(tried with cabal installed from master branch)

`cabal init` generates .cabal file with very long lines for fields like `other-extensions`, `exposed-modules` and `build-depends`. Fixing this by hand is tedious. Cabal format helps, but that's hard to discover (#2460). Ideally I think cabal should format the file before automatically.

ChShersh:
I personally prefer formatting of `cabal init` command so I wouldn't call it a badly formatted file. And in all my projects I format `.cabal` files in the style of `cabal init` (with alignment). I never heard of `cabal format` command before but after I tried this command now I see that there's no benefit in the `cabal format` formatting for me and I don't like the style of `cabal format`. So I'm voting for leaving `cabal init` formatting as it is.

osa1:
Just to clarify, this is what `cabal init` generates:

```
build-depends:       base >=4.11 && <4.12, QuickCheck >=2.11 && <2.12, hashable >=1.2 && <1.3, containers >=0.5 && <0.6, fingertree >=0.1 && <0.2, stm >=2.4 && <2.5, random >=1.1 && <1.2, text >=1.2 && <1.3, bytestring >=0.10 && <0.11, async >=2.2 && <2.3, transformers >=0.5 && <0.6, directory >=1.3 && <1.4, filepath >=1.4 && <1.5, free >=5.1 && <5.2
```

This is unusable without formatting first. Compare with `cabal format` output:

```
    build-depends:
        base >=4.11 && <4.12,
        QuickCheck >=2.11 && <2.12,
        hashable >=1.2 && <1.3,
        containers >=0.5 && <0.6,
        fingertree >=0.1 && <0.2,
        stm >=2.4 && <2.5,
        random >=1.1 && <1.2,
        text >=1.2 && <1.3,
        bytestring >=0.10 && <0.11,
        async >=2.2 && <2.3,
        transformers >=0.5 && <0.6,
        directory >=1.3 && <1.4,
        filepath >=1.4 && <1.5,
        free >=5.1 && <5.2
```

23Skidoo:
@osa1 Can you please provide examples of the current and the desired output?

ChShersh:
@osa1 My bad, I haven't understand your ticket good enough. Yeah, I don't like long lists with dependencies as well.

byorgey:
[...]

Issue #5343, opened 2018-05-24
Module listing discovery


parsonsmatt:
Now that `cabal` has common stanzas, the only thing keeping me using `hpack` is that `hpack` will, by default, autodiscover all of your modules and stuff them into `exposed-modules` (you can disable this by explicitly defining that stanza). This feature is *really* convenient to me.

Mimicing that behavior seems like it might be against the `Cabal` philosophy, as it's implicit -- "If `exposed-modules` is not defined, collect all modules not listed in `other-modules` and make them `exposed-modules`."

A middle ground solution that provides convenience with a degree of explicitness is glob patterns for the module listings. Consider this syntax:

    exposed-modules: *

This would find all modules in the source directory and add them as `exposed-modules`. Kinda like `hpack`'s autodiscovery, but we're explicitly writing "please find all the modules for me."

You could also write:

    exposed-modules:
        Control.Monad.*

which would only collect modules under the `Control.Monad` namespace into `exposed-modules` -- any other modules would need to be either explicitly added to `exposed-modules` or `other-modules`.

23Skidoo:
It'd be also convenient to be able to specify module names to be treated as exceptions from auto-exposing with glob patterns: `other-modules: Control.Monad.Internal.*`.

gbaz:
Right. I'd think we'd want some logic relating the two. Explicit always takes precedence over glob. So other-modules explicit takes precedence over glob in exposed-modules, and exposed-modules explicit takes precedence over glob in other-modules. If both are globs or both are explicit, I'd suggest an error, rather than trying to be too clever and pick a potentially nonobvious direction to resolve the ambiguity?

gbaz:
I also don't know if there are interaction effects with backpack that we'd have to worry about -- i.e is it genuinely always the case that given a glob we can simply scan a file-listing without inspecting any contents to determine what it expands to?

parsonsmatt:
Yeah, given the following modules:

    Foo
    Foo.Bar
    Baz
    Baz.Internal

I'd expect the following:

```
exposed-modules: *
-- other-modules: [] -- no other-modules clause

-- expands to:

exposed-modules:
    Foo
    Foo.Bar
    Baz
    Baz.Internal
```

```
exposed-modules:
    Foo
    Foo.*
 other-modules:
    Baz.*

-- expands to:

exposed-modules:
    Foo
    Foo.Bar
other-modules:    
    Baz.Internal
```
[...]

PR #6621, opened 2020-03-29
WIP/RFC: Parse .cabal files to a source code representation

m-renaud:
> **NOTE:** This is a work in progress, looking for input on direction/structuring (see Open Questions below).

**TL;DR;** Don't perform all simplification steps _during_ parsing, do them as a follow-up transformation.

## Why

- Allows `cabal format` to maintain key pieces of source code (common stanzas for example #5734)

- Separation of concerns; simplifies parser logic (no longer need to strange `FromBuildInfo` typeclass while parsing)

## How

- Create a source representation in the AST (`PackageSourceDescription`) for common stanza sections (`CommonStanza`) and for `import` directives within sections (`CommonStanzaImports`).

- Don't perform inlining of sections during parsing, leave that to another step. The transformation from `PackageSourceDescription` is not yet implemented, but should be pretty trivial.

Another alternative is to update `GenericPackageDescription` in place, but I thought it may be useful (if not more code) to have a separate representation. May be too complicated but we could look at the "trees that grow" approach that is used in GHC as well.

> **NOTE:** For simplicity I forked the code from GenericPackageDescription instead of modifying in place to keep things separate, so there's a lot of duplication right now. We'll likely want to refactor several modules as pre-steps to this. 

> **NOTE:** Ignore the `HasCommonStanzaImports` typeclass, that's an artifact from another approach I took and isn't used currently.

## Testing

I've updated the `cabal format` command to round trip thorough `PackageSourceDescription` instead of `GenericPackageDescription` and it appears to work for all the common cases I've tested (including `common` stanzas that `import` other common stanzas).

## Open Questions

1) What's up with `PackageDescription` and `GenericPackageDescription`? Why do they both exist? Historical reasons or separate concerns?

2) Currently I've added a `CommonStanzaImports` field to each of the section types that can have a common stanza, this results in having to insert `mempty`s in places where these sections are created manually (and thus will never have `import`s). A more principled way would be to create separate types for `ExecutableSource`, `LibrarySource` etc. which has the `CommonStanzaImports` fields, and leave the existing section types used in GenericPackageDescription unchanged.

3) As a first step should we keep this separate from the `GenericPackageDescription` parser and just use it for `cabal format`?

4) Is there good test coverage for parsing to `GenericPackageDescription`? In other words, if I replace the `String -> GenericPackageDescription` parser with a `String -> PackageSourceDescription` parser and then a separate `PackageSourceDescription -> GenericPackageDescription` transform, how confident can I be that it works?

/cc @phadej @gbaz 

---

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -447,6 +447,7 @@ library
     Distribution.Types.MungedPackageId
     Distribution.Types.PackageId
     Distribution.Types.UnitId
+    Distribution.Types.CommonStanza
     Distribution.Types.Executable
     Distribution.Types.ExecutableScope
     Distribution.Types.Library
@@ -474,6 +475,7 @@ library
     Distribution.Types.TestSuiteInterface
     Distribution.Types.TestType
     Distribution.Types.GenericPackageDescription
+    Distribution.Types.PackageSourceDescription
     Distribution.Types.Condition
     Distribution.Types.CondTree
     Distribution.Types.HookedBuildInfo
@@ -543,9 +545,11 @@ library
     Distribution.Types.Lens
     Distribution.Types.Benchmark.Lens
     Distribution.Types.BuildInfo.Lens
+    Distribution.Types.CommonStanza.Lens
     Distribution.Types.Executable.Lens
     Distribution.Types.ForeignLib.Lens
     Distribution.Types.GenericPackageDescription.Lens
+    Distribution.Types.PackageSourceDescription.Lens
     Distribution.Types.InstalledPackageInfo.Lens
     Distribution.Types.Library.Lens
     Distribution.Types.PackageDescription.Lens
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -448,6 +448,7 @@ library
     Distribution.Types.PackageId
     Distribution.Types.UnitId
     Distribution.Types.CommonStanza
+    Distribution.Types.CommonStanzaImports
     Distribution.Types.Executable
     Distribution.Types.ExecutableScope
     Distribution.Types.Library
@@ -522,6 +523,8 @@ library
     Distribution.FieldGrammar.Parsec
     Distribution.FieldGrammar.Pretty
     Distribution.PackageDescription.FieldGrammar
+    Distribution.PackageDescription.PackageSourceDescriptionParser
+    Distribution.PackageDescription.PackageSourceDescriptionPrettyPrint
     Distribution.PackageDescription.Parsec
     Distribution.PackageDescription.Quirks
     Distribution.Parsec
@@ -546,6 +549,7 @@ library
     Distribution.Types.Benchmark.Lens
     Distribution.Types.BuildInfo.Lens
     Distribution.Types.CommonStanza.Lens
+    Distribution.Types.CommonStanzaImports.Lens
     Distribution.Types.Executable.Lens
     Distribution.Types.ForeignLib.Lens
     Distribution.Types.GenericPackageDescription.Lens
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -129,7 +129,8 @@ commonStanzaFieldGrammar
     :: (FieldGrammar g, Applicative (g CommonStanza), Applicative (g BuildInfo))
     => UnqualComponentName -> g CommonStanza CommonStanza
 commonStanzaFieldGrammar n = CommonStanza n
-    <$> blurFieldGrammar L.buildInfo buildInfoFieldGrammar
+    <$> optionalFieldDef  "import"  L.commonStanzaRecursiveImports emptyCommonStanzaImports
+    <*> blurFieldGrammar L.buildInfo buildInfoFieldGrammar
 {-# SPECIALIZE commonStanzaFieldGrammar :: UnqualComponentName -> ParsecFieldGrammar' CommonStanza #-}
 {-# SPECIALIZE commonStanzaFieldGrammar :: UnqualComponentName -> PrettyFieldGrammar' CommonStanza #-}
 
--- a/Cabal/Distribution/PackageDescription/FieldGrammar.hs
+++ b/Cabal/Distribution/PackageDescription/FieldGrammar.hs
@@ -3,6 +3,8 @@
 module Distribution.PackageDescription.FieldGrammar (
     -- * Package description
     packageDescriptionFieldGrammar,
+    -- * Common Stanza
+    commonStanzaFieldGrammar,
     -- * Library
     libraryFieldGrammar,
     -- * Foreign library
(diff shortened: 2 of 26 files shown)


phadej:
We already have `Field` structure which can be used to format .cabal files. See how `cabal-fmt` is implemented.

m-renaud:
> We already have `Field` structure which can be used to format .cabal files. See how `cabal-fmt` is implemented.

Yeah, that's exactly what I'm using. `cabal-fmt` round trips through `GenericPackageDescription` via `readGenericPackageDescription verbosity path >>= writeGenericPackageDescription path` (see [formatAction code].

The problem is that `GenericPackageDescription` and the concrete Field instances that are used **do not** have any representation for imports, and the parser used to back [`readGenericPackageDescription`] never actually parses `common` stanzas or `imports` into any structure, it deeps them as internal state to fold into parsing of other sections (see [`processImports`]). Here I've created a new section type and updated the `FieldGrammar` definition to understand common stanzas and import directives.

The open questions outline some various trade-offs we can make regarding embedding the source structure within the existing field types (`Executable`, `Library`, `Benchmark`, etc) or creating a parallel structure.

I imagine that this source representation will grow to include comments as well for preservation, and this could also be used to provide an intermediary form of the package description which could have processing steps performed on them (maybe via "cabal plugins" or similar). This could provide a possible place to address things like automatic module discovery (#5343) in a more generic way and also integrate with `cabal format` (since you won't want the `*`s expanded during formatting, only during interpretation).


[formatAction code]: https://github.com/haskell/cabal/blob/5392ac8c984eb3aedb0f0aca86309a53b7cbcab8/cabal-install/main/Main.hs#L1037-L1039
[`readGenericPackageDescription`]: https://github.com/haskell/cabal/blob/5392ac8c984eb3aedb0f0aca86309a53b7cbcab8/Cabal/Distribution/PackageDescription/Parsec.hs#L103
[`processImports`]: https://github.com/haskell/cabal/blob/5392ac8c984eb3aedb0f0aca86309a53b7cbcab8/Cabal/Distribution/PackageDescription/Parsec.hs#L601-L634

phadej:
To clarify, `https://hackage.haskell.org/package/cabal-fmt` is not the `cabal format`. So 

> Yeah, that's exactly what I'm using. cabal-fmt round trips through GenericPackageDescription via readGenericPackageDescription verbosity path >>= writeGenericPackageDescription path (see formatAction code).

is incorrect.

m-renaud:
Ahhhh I see, I wasn't aware of that package, I misread that that as `cabal format`. Is there any reason to not fold the functionality into cabal proper? Or is the plan to deprecate `cabal format` and just move all formatting functionality to `cabal-fmt`?

It seems like it would be beneficial to have this functionality as part of the cabal itself, this way as the Cabal format changes the formatting can be kept in sync. It also removes the discover-ability barrier of having to find and install another package.

phadej:
> It seems like it would be beneficial to have this functionality as part of the cabal itself, this way as the Cabal format changes the formatting can be kept in sync. 

I agree. Yet, I'm not ready for bikeshedding. In `cabal-fmt` I can end it by saying that the tool is opinionated. `cabal-install` have to be configurable. I'm optimistic that I get to the rework of `cabal.config` and `cabal.project` parsers this year, so there could be a way to add configution for formatting too.

Otherwise there will be issues like https://github.com/phadej/cabal-fmt/issues/11, in a sense the fact no-one uses `cabal format` is a blessing.

m-renaud:
> In cabal-fmt I can end it by saying that the tool is opinionated. cabal-install have to be configurable

That's fair, but I would also imagine that a package named `cabal-fmt` would be a general tool and  `cabal-fmt-for-phadej` would be your personal flavour :stuck_out_tongue_winking_eye: 

> I'm optimistic that I get to the rework of cabal.config and cabal.project parsers this year, so there could be a way to add configuration for formatting too.

I don't think it's reasonable to wait until some unspecified point in the future where a parser rewrite happens before making incremental improvements now. It's true that we'll eventually want some degree of configuration when formatting Cabal files (and some small subset of the community may have incredibly specific needs), but I imagine there is a pretty large group of the community that would be happy with cabal coming with a formatter with a couple of nobs (even just indentation). It doesn't make sense to me to tie creating a better internal representation to rewriting the parser for it.

Also, I'd be willing to build out the configuration infrastructure as well, it's not a particularly difficult problem to solve in the grand scheme of things. There are some improvements that I think would be useful as well such as separating the folding in of common stanza data into the other sections, imho that should be a separate step from parsing.

> in a sense the fact no-one uses cabal format is a blessing. Syntax and style is too easy to tell opinion about

As a generally available tool that we're going out and saying "everyone start using this now!" I agree. But, at the same time I disagree because it fills a real hole in the current tooling for those who want it (even if its not feature complete or fully configurable now). We could get it to a point where it handles most use cases (really just the fact that it inlines common stanzas right now is probably the only reason more folks aren't using it), and then make an announcement that it is available as an early access tool, and to expect changes to come in the future until a point at which all the configurability is available.

So, what I'm looking for are concrete next steps here, I don't want to block all improvements in this space on waiting for the perfect end state to be defined, and as you said if no one is using the tool we should feel free to make changes and see what we like and don't like. Are there specific implementation details that you don't like here? Would you prefer if I do more prefactoring work to clean up the code before making these changes (to make it more clear how much is actually changing)? Would you like to see a write-up of a longer term plan for formatting in cabal? Can I help with the parser changes you mentioned above? Is there a group of people I should reach out to for wider input on this? I'd be happy to do any and all of these, I'm just looking for guidance and actionable input to move this forward.

phadej:
>>> I don't think it's reasonable to wait until some unspecified point in the future where a parser rewrite happens before making incremental improvements now.

Most functionality of `cabal-fmt` is already present in `Cabal`. `cabal-fmt` starts to be opinionated in how it renders individual fields. Better version of `readGPD >>= writeGPD` can be already done with  `parseFields >>= showFields`.

The internal representation, i.e. `Field` is already there.

>>> .. and to expect changes to come in the future until a point at which all the configurability is available.

Please take a lead, champion the functionality and be responsive when follow up issues and requests are reported and created. The progress is not forbidden. But there have to be a plan.

It's true that since `readGPD >>= writeGPD` was fixed (https://github.com/haskell/cabal/issues/4719, check out who fixed that) there are new opportunities,. I have also since extending the parsing functionality of `Cabal` based on the lessons from writing various experiments for formatting. `cabal-fmt` the latest (but not the only one).

E.g. comments were added so 

Yet, I don't have time to deal with formatting related functionality in `cabal-install`.

---

Some alternative to `GPD` representation is required. With `Field` (which already  exists) you are  be able to preserve ordering of fields, and (some) comments (as illustrated by `cabal-fmt`). With  GPD-like variant which don't inline common stanzas, you'll loose ordering information and combine fields defined in multiple steps.

Importantly, `GPD` is an input structure for `cabal-install`'s solver (or rather `CondTree BuildInfo` like structures). Thus there a huge gap between source code and ~~first~~ second internal representation.

I don't think the latter (GPD-like structure losing ordering and comments) is what people would want, even as incremental "update".

Some people would also like to have `ghc-exactprint` like level of control of the output, i.e. that you can round trip the `.cabal` file **completely**, without destroying any formatting if you don't want to. To allow writing general refactoring tools. `Field` is too coarse for that. That would require completely new parser. It might make sense to have that as separate codepath, where one could deal with ill-formed input too, in `.cabal` files error-recovery is relatively easy.

---

I'd suggest you to look into https://github.com/haskell/cabal/issues/6187 and https://github.com/haskell/cabal/issues/5555, the annotations support I mention in #5555 is now in `Cabal`. You are the `cabal init` lead now.

Kleidukos:

ffaf1:

Issue #6149, opened 2019-07-21
cabal v2-repl scriptfile.hs


phadej:
`cabal v2-repl scriptfile.hs` outside project to load a `scriptfile.hs` into repl (GHCi).

---

Suppose I have a script file like:

```haskell
{- cabal:
build-depends: base, splitmix
-}

import Data.List (unfoldr)
import Data.Word (Word64)
import System.Random.SplitMix (nextWord64, mkSMGen)

numbers :: [Word64]
numbers = unfoldr (Just . nextWord64) (mkSMGen 42)

main :: IO ()
main = print $ take 10 numbers
```

`cabal v2-run -v0 scriptfile.hs` works perfectly, however it takes quite a lot of time, 10s on my machine!

```
% time cabal v2-run -v0 ExRun.hs
[1275548033995301424,10417309031967933079,2112719111588962399,2726445820918627087,4609004260007739600,11074269206337254866,9733109294540808081,6450629194222476823,8084222243108663947,16253064923614931039]
cabal v2-run -v0 ExRun.hs  10,70s user 1,01s system 99% cpu 11,724 total
```

Slowness is one issue, after all there's compiling involved. It however makes development of scriptfiles extremely painful.

Yet, if the following worked, the development would be a lot nicer (repl ftw):
```
% cabal v2-repl ExRun.hs
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
In order, the following will be built (use -v for more details):
 - fake-package-0 (exe:script) (first run)
Configuring executable 'script' for fake-package-0..

...

GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( ExRun.hs, interpreted )
Ok, one module loaded.
*Main> 

*Main> main
[1275548033995301424,10417309031967933079,2112719111588962399,2726445820918627087,4609004260007739600,11074269206337254866,9733109294540808081,6450629194222476823,8084222243108663947,16253064923614931039]
```

worked, it would be super great.

---

For now, I try to `cabal install --lib` the dependencies, and use raw `ghci`.

mpickering:
I'm interested in this feature. Could you give some pointers about how hard it would be to implement? (cc @fendor)

fendor:
I intend to work on this.
The implementation of the scriptfile feature seems to be mainly in `CmdInstall`. It sets up a fake project by essentially writing a *.cabal file to a temporary location and then building this fake project.

[...]

PR #6623, opened 2020-03-29
Prepend hs-source-dir to match-component #6622

### LINKED ISSUE #6622: Bug: `cabal repl file` fails for executable component
**To Reproduce**
Set up an example project:
```
$  cabal init -p example --libandexe --tests --main-is=Main.hs --source-dir=src --application-dir=app --test-dir=test
```
Load library component:
```
$ cabal repl src/MyLib.hs
# Succeeds to load
$ cabal repl src/Lib.hs
# Fails to load as expected 
$ cabal repl app/Main.hs
# Fails to load although
cabal.exe: Unknown target './app/Main.hs'.
The package example has no file target 'app/Main.hs'.
$ cabal repl test/MyTestLib.hs
cabal.exe: Unknown target './test/MyLibTest.hs'.
The package example has no file target 'test/MyLibTest.hs'.
```
**Expected behavior**

The file targets `app/Main.hs` and `test/MyTestLib.hs` should successfully load.

**System information**
 - `cabal`: `3.0.0.0`, but also HEAD.

**Benefit**

Directly benefits [hie-bios](https://github.com/mpickering/hie-bios) which uses `cabal repl <filename>` to find compilation flags.
cc @mpickering

fendor:
Why this change? 
The function `matchFile` expects arguments of a form such as:
```haskell
matchFile [("app/Main.hs", ()), ("test/Test.hs", ())]  "app/Main.hs"
```

but the old implementation produces the following:
```haskell
[("Main.hs", ()), ("Test.hs", ())]
```
since it doesnt prepend in the `hsSourceDirs` for each `other-module`.

---

* ~~[ ] Add tests for ambiguous filepath targets~~

closes #6622

### DIFF
--- a/Cabal/doc/cabal-commands.rst
+++ b/Cabal/doc/cabal-commands.rst
@@ -86,6 +86,12 @@ A target can take any of the following forms:
      - ``tests``,
      - ``benches``, ``benchmarks``.
 
+-  A module target: ``[package:][ctype:]module``, which specifies that the
+   component of which the given module is a part of will be built.
+
+-  A filepath target: ``[package:][ctype:]filepath``, which specifies that the
+   component of which the given filepath is a part of will be built.
+
 In component targets, ``package:`` and ``ctype:`` (valid component types
 are ``lib``, ``flib``, ``exe``, ``test`` and ``bench``) can be used to
 disambiguate when multiple packages define the same component, or the
@@ -99,9 +105,12 @@ Some example targets:
 
     $ cabal v2-build lib:foo-pkg       # build the library named foo-pkg
     $ cabal v2-build foo-pkg:foo-tests # build foo-tests in foo-pkg
-
-(There is also syntax for specifying module and file targets, but it
-doesn't currently do anything.)
+    $ cabal v2-build src/Lib.s         # build the library component to
+                                       # which "src/Lib.hs" belongs
+    $ cabal v2-build app/Main.hs       # build the executable component of
+                                       # "app/Main.hs"
+    $ cabal v2-build Lib               # build the library component to
+                                       # which the module "Lib" belongs
 
 Beyond a list of targets, ``cabal v2-build`` accepts all the flags that
 ``cabal v2-configure`` takes. Most of these flags are only taken into
--- a/cabal-install/tests/IntegrationTests2.hs
+++ b/cabal-install/tests/IntegrationTests2.hs
@@ -247,9 +247,15 @@ testTargetSelectors reportSubCase = do
                                  ":pkg:p:lib:p:file:P.y"
                      , "q/QQ.hs", "q:QQ.lhs", "lib:q:QQ.hsc", "q:q:QQ.hsc",
                                   ":pkg:q:lib:q:file:QQ.y"
+                     , "q/Q.hs", "q:Q.lhs", "lib:q:Q.hsc", "q:q:Q.hsc",
+                                  ":pkg:q:lib:q:file:Q.y"
+                     , "app/Main.hs", "p:app/Main.hs", "exe:ppexe:app/Main.hs", "p:ppexe:app/Main.hs",
+                                  ":pkg:p:exe:ppexe:file:app/Main.hs"
                      ]
        ts @?= replicate 5 (TargetComponent "p-0.1" (CLibName LMainLibName) (FileTarget "P"))
            ++ replicate 5 (TargetComponent "q-0.1" (CLibName LMainLibName) (FileTarget "QQ"))
+           ++ replicate 5 (TargetComponent "q-0.1" (CLibName LMainLibName) (FileTarget "Q"))
+           ++ replicate 5 (TargetComponent "p-0.1" (CExeName "ppexe") (FileTarget "app/Main.hs"))
        -- Note there's a bit of an inconsistency here: for the single-part
        -- syntax the target has to point to a file that exists, whereas for
        -- all the other forms we don't require that.
@@ -1330,7 +1336,7 @@ testExceptionInConfigureStep config = do
     (_pkga1, failure) <- expectPackageFailed plan res pkgidA1
     case buildFailureReason failure of
       ConfigureFailed _ -> return ()
-      _ -> assertFailure $ "expected ConfigureFailed, got " ++ show failure 
+      _ -> assertFailure $ "expected ConfigureFailed, got " ++ show failure
     cleanProject testdir
   where
     testdir = "exception/configure"
--- a/cabal-install/tests/IntegrationTests2.hs
+++ b/cabal-install/tests/IntegrationTests2.hs
@@ -361,6 +361,16 @@ testTargetSelectorAmbiguous reportSubCase = do
                       , mkexe "other2" `withCFiles`  ["Foo"] ]
       ]
 
+    -- File target is ambiguous, part of multiple components
+    reportSubCase "ambiguous: file in multiple comps"
+    assertAmbiguous "Bar.hs"
+      [ mkTargetFile "foo" (CExeName "bar")  "Bar"
+      , mkTargetFile "foo" (CExeName "bar2") "Bar"
+      ]
+      [ mkpkg "foo" [ mkexe "bar"  `withModules` ["Bar"]
+                    , mkexe "bar2" `withModules` ["Bar"] ]
+      ]
+
     -- non-exact case packages and components are ambiguous
     reportSubCase "ambiguous: non-exact-case pkg names"
     assertAmbiguous "Foo"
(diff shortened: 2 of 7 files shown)


mpickering:
Is this feature of the repl documented in the manual? Now might be a good time to do so.

fendor:
I am happy to update any documentation for `cabal repl`. Where should the documentation be added?

phadej:
re docs, there are sections for `v2` commands. For example https://cabal.readthedocs.io/en/latest/nix-local-build.html#cabal-v2-repl

There are also `TargetSelector` tests in `cabal-install/tests/IntegrationTests2.hs`. Please add one testing new scenario.

fendor:
target-syntax
This should be updated, since `cabal build File.hs` works.

fendor:
The windows error seems to be that the following works on windows:
```bash
$ cabal repl p:app\Main.hs
```
but this doesnt:
```bash
$ cabal repl p:app/Main.hs
```

However, both 
```bash
$ cabal repl app/Main.hs

$ cabal repl app\Main.hs
```
work.

Is it ok to fix that in this PR, too?

Cabal/doc/nix-local-build.rst
```diff
@@ -357,6 +357,16 @@ A target can take any of the following forms:
...
 
+-  A module target: ``[package:][ctype:]module``, which specifies that the
+   component of which the given module is a part of will be built. If the
+   target is ambiguous, e.g. only ``module`` is specified and the module name
+   is part of multiple components, then no component will be built.
+
+-  A filepath target: ``[package:][ctype:]filepath``, which specifies that the
+   component of which the given filepath is a part of will be built. If the
+   target is ambiguous, e.g. only ``filepath`` is specified and the filepath
+   is part of multiple components, then no component will be built.
```
fendor, line 368: This needs to be proven with a test.
Currently, an ambiguous target produces this error message:
```
TargetSelector: internal error: NoMatch
  error, called at .\\Distribution\\Client\\TargetSelector.hs:669:3 in main:Distribution.Client.TargetSelector
```

fendor:
So, currently there is an internal error when the file target is ambiguous, e.g. target is part of two executables, therefore hard to actually test for. The bug-fix for that is relatively simple, however, I feel like this PR is already bigger than planned. 
Should I push the bug-fix here, or rather create another issue and another PR, so at least the very helpful changes in this PR are merged?

fendor:
ping

phadej:
> So, currently there is an internal error when the file target is ambiguous, e.g. target is part of two executables,

Shouldn't `v2-build` just build both?

phadej:
I'm on the fence with this. Something tells me that `cabal repl Foo.hs` should not pick the component `Foo.hs` is included but build all of the project (or none if outsde) and load `Foo.hs`.

https://github.com/haskell/cabal/issues/6149

fendor:
> Shouldn't `v2-build` just build both?

Yeah, but it fails. I assume it is just some minor bug, somewhere in target matching.

> I'm on the fence with this. Something tells me that `cabal repl Foo.hs` should not pick the component `Foo.hs` is included but build all of the project (or none if outsde) and load `Foo.hs`.

What is all of the project in the case of a library? Wouldnt "library" be the component to build?
Or do you mean additionally, we should load the module `Foo` into the ghci session? 
Afaict, we could add `:add Foo` to the ghci-script passed to ghci to achieve this.

However, please note, that not behaviour is introduced by this patch. This patch only fixes two bugs regarding finding the component associated with the module.

phadej:
I imagine use case:

```haskell
cd some-library            # go to library sources
vim ExampleBug.hs          # quickly hack some example file
cabal repl ExampleBug.hs   # load it into repl
```

This is very common workflow for myself, but I have to rely on local
environment files now.

Yet, I don't understand what is the motivation to be able to select
a component by an arbitrary source-file. Module name is somewhat in unclear,
also `main-is: EntryPoint.hs`. But arbitrary source file: not really.

OTOH out-of-package source file and in package source files should
be disjoint, so there is no ambiguity.

mpickering:
@phadej The motivation to select a component by a source file is to implement an IDE which can create the right session given any FilePath. 

This patch just fixes the existing behavior as far as I understand.

fendor:
> This is very common workflow for myself, but I have to rely on local
environment files now.

Would this be accomplished by #6149? Or rather, could we accomplish it that way?

phadej:
@mpickering I see. That's enough for me for a motivation.

fendor:
So, is there anything blocking this for getting merged? I would be happy if this *bug*-fix can make it into the next release.

Cabal/doc/cabal-commands.rst
```diff
@@ -99,6 +105,12 @@ Some example targets:
...
     $ cabal v2-build foo-pkg:foo-tests # build foo-tests in foo-pkg
+    $ cabal v2-build src/Lib.s         # build the library component to
+                                       # which "src/Lib.hs" belongs
+    $ cabal v2-build app/Main.hs       # build the executable component of
+                                       # "app/Main.hs"
+    $ cabal v2-build Lib               # build the library component to
+                                       # which the module "Lib" belongs
 
 (There is also syntax for specifying module and file targets, but it
 doesn't currently do anything.)
```
DanielG, line 116: Isn't this no longer true now?

fendor: Why not? It was at least a month ago.

fendor: Nothing here that this PR would have changed, did I miss another change?

fendor: Ok, got it what you mean, thanks for catching this!

fendor:
Is it ok if I merge this? Ideally before #6774 is merged so this PR does not need to be rebased.

m-renaud:
Assuming no major delays here I can hold off until this PR is merged.

fendor:
I am just waiting for approval. Dont want to go over anyone's head.

phadej:
what happens if executable name is `HLint` (note case), and there is `HLint` module in a library.

Is that ambiguous, how to resolve ambiguity?

phadej:
I see

```
                                  ":pkg:q:lib:q:file:Q.y"
                                  ":pkg:p:exe:ppexe:file:app/Main.hs"
```

selectors in text but have no idea what they mean (by looking at the documentation changes).

cabal-install/tests/IntegrationTests2.hs
```diff
@@ -247,9 +247,15 @@ testTargetSelectors reportSubCase = do
                                  ":pkg:p:lib:p:file:P.y"
                      , "q/QQ.hs", "q:QQ.lhs", "lib:q:QQ.hsc", "q:q:QQ.hsc",
                                   ":pkg:q:lib:q:file:QQ.y"
+                     , "q/Q.hs", "q:Q.lhs", "lib:q:Q.hsc", "q:q:Q.hsc",
+                                  ":pkg:q:lib:q:file:Q.y"
+                     , "app/Main.hs", "p:app/Main.hs", "exe:ppexe:app/Main.hs", "p:ppexe:app/Main.hs",
```
phadej, line 252: Should we also add a test that `Main.hs` would be ambiguous?

fendor: Yeah, an ambiguous target selector makes sense, going to update it asap!

fendor:
> what happens if executable name is HLint (note case), and there is HLint module in a library.

I dont know a lot about the module syntax, but the current (3.2.0.0) behaviour is that the executable is chosen. If this is not clear enough, we can remove the module syntax from the documentation.

> selectors in text but have no idea what they mean (by looking at the documentation changes).

That is mainly copied from the other test-cases. I read the latter as the most verbose version of a file-target syntax while the former is the same for `.y` files (which are generated from happy, I think?)

cabal-install/tests/IntegrationTests2.hs
fendor: I am sorry, I am not sure I get what you are saying?
Do you mean to add a test for the case that `Main.hs` is ambiguous? Since it should be a filepath, I dont think it can be ambiguous?

fendor:
> what happens if executable name is HLint (note case), and there is HLint module in a library.

There exists a test for it already:
```haskell 
    -- local components shadow modules and files
    reportSubCase "unambiguous: cwd comp vs module, file"
    assertUnambiguous "Foo"
      (mkTargetComponent "bar" (CExeName "Foo"))
      [ mkpkg "bar" [mkexe "Foo"]
      , mkpkg "other" [ mkexe "other"  `withModules` ["Foo"]
                      , mkexe "other2" `withCFiles`  ["Foo"] ]
      ]
```

fendor:
Merged via #6826

PR #6628, opened 2020-03-30
Reset to 7fda66f60481adb36d6ca18ef8e98a00b4e2316a

phadej:
Let's see what CI says


phadej:
So CI still fails. I suspect something changed in Windows environment. I wonder what.

PR #6637, opened 2020-04-01
Remove v1-sdist

### LINKED ISSUE #6635: Remove v1-sdist
There are `sdist` related issues and fixing them twice is not fun.

phadej:
Resolves #6635


phadej:
Two things broke, it seems we
- might need `--ignore-project` for `cabal v2-sdist` too (makes sense)
- need `--list-sources=FILE`, the current `--list-only` is too limited.

PR #6643, opened 2020-04-02
setupAndCabalTest uses cabal act-as-setup for cabal part

phadej:


phadej:
I opened this PR with improper name first. It should been "setupAndCabalTest uses cabal act-as-setup for cabal part" .

PR #6654, opened 2020-04-05
Require cabal-versions >=1.25 to be exact

phadej:


phadej:

PR #6661, opened 2020-04-06
Improve `cabal init` code a bit

phadej:
ping @m-renaud

### DIFF
--- a/cabal-install/Distribution/Client/Init/Command.hs
+++ b/cabal-install/Distribution/Client/Init/Command.hs
@@ -31,8 +31,6 @@ import System.Directory
 import System.FilePath
   ( (</>), takeBaseName, equalFilePath )
 
-import Data.List
-  ( (\\) )
 import qualified Data.List.NonEmpty as NE
 import Data.Function
   ( on )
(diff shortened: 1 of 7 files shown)


cabal-install/Distribution/Client/Init/Command.hs
```diff
@@ -269,39 +267,45 @@ getVersion flags = do
...
-         ?>> fmap (fmap (either UnknownLicense id))
-                  (maybePrompt flags
-                    (promptList "Please choose a license" listedLicenses
-                     (Just BSD3) displayLicense True))
-
-  case checkLicenseInvalid lic of
-    Just msg -> putStrLn msg >> getLicense flags
-    Nothing  -> return $ flags { license = maybeToFlag lic }
-
+  elic <-     return (fmap Right $ flagToMaybe $ license flags)
```
m-renaud, line 270: Nit: de-indent return (...) by one char to match below.

cabal-install/Distribution/Client/Init/Defaults.hs
```diff
@@ -21,11 +21,11 @@ import Distribution.ModuleName
...
   ( fromString )
-import Distribution.Version
-  ( Version, mkVersion )
+import Distribution.CabalSpecVersion
+  ( CabalSpecVersion (..))
 
-defaultCabalVersion :: Version
-defaultCabalVersion = mkVersion [1,10]
+defaultCabalVersion :: CabalSpecVersion
+defaultCabalVersion = CabalSpecV1_10
```
m-renaud, line 28: Optional: can you change this to `CabalSpecV2_4` to match other places? I missed the default value in the prompt and only defaulted it under non-interactive init when unspecified.

phadej: I'll change non-interactive to actually use values from `Defaults`

m-renaud: SG.

PR #6665, opened 2020-04-06
6610 Add pijul to known repository type

v0d1ch:
Unfortunately I could not install pijul on my machine and test out this pr so any help would be appreciated.

### DIFF
--- a/cabal-install/Distribution/Client/VCS.hs
+++ b/cabal-install/Distribution/Client/VCS.hs
@@ -28,6 +28,7 @@ module Distribution.Client.VCS (
     vcsGit,
     vcsHg,
     vcsSvn,
+    vcsPijul,
   ) where
 
 import Prelude ()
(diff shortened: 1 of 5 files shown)


cabal-install/tests/UnitTests/Distribution/Client/VCS.hs
```diff
@@ -693,3 +718,45 @@ vcsTestDriverDarcs mtimeChange verbosity vcs repoRoot =
                            }
     darcs = runProgramInvocation verbosity . darcsInvocation
 
+
+vcsTestDriverPijul :: Verbosity -> VCS ConfiguredProgram
```
phadej, line 722: Nice. Having test is great. I'll take a look at CI so it won't fail

(Looks like there is no `apt-get install pijul` easy way of installing pijul on ubuntu)

cabal-install/Distribution/Client/VCS.hs
```diff
@@ -498,3 +499,84 @@ svnProgram = (simpleProgram "svn") {
...
+vcsPijul :: VCS Program
+vcsPijul =
+    VCS {
+      vcsRepoType = KnownRepoType Pijul,
+      vcsProgram  = pijulProgram,
+      vcsCloneRepo,
+      vcsSyncRepos
+    }
+  where
+    vcsCloneRepo :: Verbosity -- ^ it seams that pijul does not have verbose flag
```
phadej, line 515: I'm quite sure it does, let us find out.

phadej: ... but indeed, it looks like there are no flags to affect output of `pijul` atm. Surprising :)

phadej:
I got `pijul` installed, and it seems to work. I can continue from here!

Thanks for doing this @v0d1ch .

phadej:
Merged in #6684

PR #6666, opened 2020-04-06
Fix sdist permissions

phadej:
---


ramsdell:
I have just been informed that Cabal sdist will make all files in Extra-Source-Files read-only.  This is very bad for the CPSA distribution because Extra-Source-Files contains executable scripts.  How do you propose executable scripts be included in a distribution?

phadej:
What kind of executable scripts. If they are truly *scripts*, execute them using explicitly calling an interpreter.

ramsdell:
We deliver several scripts.  One script runs our tests suite.  One script finds the largest amount of memory that can be safely used on a Linux system by consulting /proc/meminfo.  It exports its result by setting GHCRTS.  Another uses Python to perform a complex transformation on CPSA input.  They are programs to be executed.

Why should one not be able to include executable scripts in a distribution?  Seems arbitrary to fail to allow this.

phadej:
Because it's very hard for `Cabal` to know which files needs to be executable and which shouldn't. As your original issue mentioned, just marking all `extra-source-files` as executables is wrong as well.

I don't want rely on the existing file permissions, as those are very implicit. E.g. on Windows.

Additionally the packages may be unpacked on filesystems with `noexec` so just trying to execute scripts may fail even on windows. Discovering `python` and running the interpreter explicitly is a bit more work, but is not impossible, and overall better.

See e.g. how `Cabal` itself calls `configure` script: https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple.hs#L692-L758

ramsdell:
Hmm.  I thought Cabal was inferring permissions correctly in pre 3.0 versions of Cabal, but then I had strange problems with sdist on Macs, so I was always building distributions on Linux. (There was some weird issue with the size of uuid because my company uses employee numbers.)  In any event, that is the past.  Let me ask for something different to solve the problem.

Please add a new field called `scripts`  which is behaves as does `extra-source-files`, with the only exception that it adds execute permissions to the files within the archive.  Surely this must be easy to do.

gbaz:
Can you please confirm why oleg's proposed solution (calling the script file explicitly with an interpreter) does not work?

ramsdell:
The user has to know what interpreter to call.  A user should not need to know that.  That is what the `#!` notation at the beginning of an executable script file is for.

ramsdell:
It just occurred to me that some people many not know about shebang in Unix.  It's a standard feature that allows a text file to used as if it were an executable.  The feature is enabled when the file is given execute permission and the first two characters of the file are `#!`.  The characters that follow the shebang must specify the absolute path of an executable.  Some typical examples are:

* `#!/bin/sh` - execute the file using a Bourne shell or something compatible. 
* `#!/usr/bin/env python` - execute the file using the version of Python found by env.
* `#!/bin/false` - ensure the file fails to execute even if it is given execute permissions.

As it says in [Shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)), the purpose is to "allow scripts and data files to be used as commands, hiding the details of their implementation from users and other programs, by removing the need to prefix scripts with their interpreter on the command line."

A CPSA distribution contains several Bourne shell scripts and a few Python scripts.

PR #6669, opened 2020-04-06
Create "Getting started" tutorial.

m-renaud:
Create a Getting Started guide focused on newcomers to the Haskell ecosystem and Cabal. This page sits at the front of the Cabal user guide and walks through creating a new package, building and running it, and adding a dependency on a package from Hackage.

Rendered docs can be found at: https://m-renaud-haskell-cabal.readthedocs.io/en/mrenaud-docs-newcomer-tutorial/getting-started.html

Follow-up PRs:

- Link to `cabal (v2-)build` docs (when they are written)
- Write and link to section about `^>=` (there's no place that really explains it currently)
- More commentary about `:target` syntax (and write docs discussing target syntax to link to)
- Add section about "projects"

---

### DIFF
--- a/Cabal/doc/getting-started.rst
+++ b/Cabal/doc/getting-started.rst
@@ -1,0 +1,169 @@
+Getting Started with Haskell and Cabal
+======================================
+
+Installing the Haskell toolchain
+--------------------------------
+
+To install the Haskell toolchain follow the `ghcup instructions
+<https://www.haskell.org/ghcup/>`__ if you're using Linux or Mac, or follow
+`this guide <https://hub.zhox.com/posts/introducing-haskell-dev/>`__ if you're
+using Windows.
+
+
+Creating a new application
+--------------------------
+
+Let's start by creating a simple Haskell application from scratch where we'll
+learn about a Haskell package's directory structure, how to run the executable,
+and how to add external dependencies.
+
+
+Initializing the application
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Start by creating a ``myfirstapp`` directory to hold the project files, these
+instructions work in unix shells and PowerShell (if you're on Windows).
+
+.. code-block:: console
+
+    $ mkdir myfirstapp
+    $ cd myfirstapp
+
+Once you have an empty directory we can initialize our package:
+
+.. code-block:: console
+
+    $ cabal init --cabal-version=2.4 --license=NONE -p myfirstapp
+
+This will generate the following files:
+
+.. code-block:: console
+
+    $ ls
+    CHANGELOG.md
+    Main.hs
+    myfirstapp.cabal
+    Setup.hs
+
+
+``Main.hs`` is where your package's code lives. By default ``cabal init``
+creates an executable with the same name as the package ``myfirstapp`` in this
+case, you can instruct ``cabal init`` to generate just a library (with
+``--lib``) or both a library and executable with (``--libandexe``); for the full
+set of options see ``cabal init --help``.
+
+``myfirstapp.cabal`` is Cabal's metadata file which describes your package and
+its dependencies. We'll be updating this file in a little bit when we add an
+external dependency to our package.
+
+
+Running the application
+^^^^^^^^^^^^^^^^^^^^^^^
+
+As mentioned above, ``cabal init`` with no arguments generates a package with a
+single executable that prints ``"Hello, Haskell!"`` to the terminal. To run the
+executable enter the following command:
+
+``cabal run :myfirstapp``
+
+You should see the following output in the terminal:
+
+.. code-block:: console
+
+     $ cabal run :myfirstapp
+     ...
+     Hello, Haskell!
+
+.. note:: The ``:`` prefix in ``:myfirstapp`` signifies that the the
+	  ``myfirstapp`` target is part of the current package.
+
+Notice that we didn't need to run a `build` command before ``cabal run``, this
+is because ``cabal run`` first determines if the code needs to be re-built
+before running the executable. If you just want to build a target you can do so
+with ``cabal build``:
+
+``cabal build :myfirstapp``
+
+
+Adding dependencies
+^^^^^^^^^^^^^^^^^^^
+
+Next we'll add an external dependency to our application. `Hackage
+<https://hackage.haskell.org/>`__ is the Haskell community's central `package`
+archive of open source software.
+
+In our application, we'll use a package called `haskell-say
+<https://hackage.haskell.org/package/haskell-say>`__ to print text to the
+terminal with some embellishment.
+
+.. TIP::
+   If you installed ``cabal`` a while ago but haven't used it recently you may
+   need to update the package index, you can do this by running ``cabal
+   update``.
+
+In our ``myfirstapp.cabal`` file we'll update the ``build-depends`` attribute of
+the ``executable myfirstapp`` section to include ``haskell-say``:
+
+.. code-block:: cabal
+
+   executable myfirstapp
+       main-is: Main.hs
+       build-depends:
+           base >=4.11 && <4.12,
+           haskell-say ^>=1.0.0.0
+
+.. NOTE::
+   ``^>=1.0.0.0`` means use version 1.0.0.0 of the library or any more recent
+   minor release with the same major version.
+
+Next we'll update ``Main.hs`` to use the ``HaskellSay`` library:
+
+.. code-block:: haskell
+
+   module Main where
+
+   import HaskellSay (haskellSay)
+
+   main :: IO ()
+   main =
+     haskellSay "Hello, Haskell! You're using a function from another package!"
+
+``import HaskellSay (haskellSay)`` brings the ``haskellSay`` function from the
+module named ``HaskellSay`` into scope. The ``HaskellSay`` module is defined in
+the ``haskell-say`` packages that we added a dependency on above.
+
+Now you can build and re-run your code to see the new output:
+
+.. code-block:: console
+
+   $ cabal run
+       ________________________________________________________
+      /                                                        \
+     | Hello, Haskell! You're using a function from another     |
+     | package!                                                 |
+      \____       _____________________________________________/
+           \    /
+            \  /
+             \/
+       _____   _____
+       \    \  \    \
+        \    \  \    \
+         \    \  \    \
+          \    \  \    \  \-----------|
+           \    \  \    \  \          |
+            \    \  \    \  \---------|
+            /    /  /     \
+           /    /  /       \  \-------|
+          /    /  /    ^    \  \      |
+         /    /  /    / \    \  \ ----|
+        /    /  /    /   \    \
+       /____/  /____/     \____\
+
+
+What Next?
+----------
+
+Now that you know how to set up a simple Haskell package using Cabal, check out
+some of thee resources on the Haskell website's `documentation page
+<https://www.haskell.org/documentation/>`__ or read more about packages and
+Cabal on the `introduction <intro.html>`__ page.
--- a/Cabal/doc/index.rst
+++ b/Cabal/doc/index.rst
@@ -6,6 +6,7 @@ Welcome to the Cabal User Guide
    :maxdepth: 2
    :numbered:
 
+   getting-started
    intro
    config-and-install
    concepts-and-development


jneira:
Thanks for writing this, it will be a great improvement for beginners.
One note: it seems to me that, in general, it could be more friendly for windows users: 
* Installation method refer to an only *nix method (I know that it redirects to haskell platform but still)
* All console commands are posix, although some of them could be executed in cmd and/or powershell. 

Otoh not sure if the tutorial needs some changes to be executed.

I've got an error following the instructions cause the cabal file generated has not the adequate `cabal-version` with support for the caret operator (i've tried it with `3.0.0.0` and `3.2.0.0`):

```
PS D:\dev\ws\haskell\myapp> cabal-3.0.0.0 run :myapp
Errors encountered when parsing cabal file .\myapp.cabal:

myapp.cabal:24:46: error:
unexpected major bounded version syntax (caret, ^>=) used. To use this syntax the package need to specify at least 'cabal-version: 2.0'. Alternatively, if broader compatibility is important then use: >=1.0.0.0 && <1.1
expecting "." or "-"

   23 |   build-depends:       base >=4.12 && <4.13,
   24 |                        haskell-say ^>=1.0.0.0
      |                                              ^
```

I am getting errors with the target for `cabal-3.2.0.0` (is it a cabal bug??!?) and the lib `haskell-say` is not in hackage (i suppose it will be before pushing this to master):

```
PS D:\dev\ws\haskell\myapp> cabal-3.2.0.0 run :myapp
cabal-3.2.0.0.exe: Unrecognised target ':myapp'.
Expected a meta-target 'all', rather than 'myapp'.

PS D:\dev\ws\haskell\myapp> cabal-3.0.0.0 run :myapp # same for cabal-3.2.0.0 run all
Resolving dependencies...
cabal-3.0.0.0.exe: Could not resolve dependencies:
[__0] trying: myapp-0.1.0.0 (user goal)
[__1] unknown package: haskell-say (dependency of myapp)
[__1] fail (backjumping, conflict set: haskell-say, myapp)
goals I've had most trouble fulfilling: myapp, haskell-say
```

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. code-block:: console
+
+    $ mkdir myfirstapp && cd myfirstapp
+
+Once you have an empty directory we can initialize our package:
+
+.. code-block:: console
+
+    $ cabal init
```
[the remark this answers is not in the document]
phadej: In other words, we want something like

```
cabal init --cabal-version=2.4 --license=NONE
```

for now. The new defaults are not yet released.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+This will generate the following files:
+
+.. code-block:: console
+
+    $ ls
+    CHANGELOG.md
```
phadej, line 39: The `Setup.hs` is still generated.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+case, you can instruct ``cabal init`` to generate just a library (with
+``--lib``) or both a library and executable with (``--libandexe``); for the full
+set of options see ``cabal init --help``.
+
+``myfirstapp.cabal`` is Cabal's metadata file which describes your package and
+its dependencies. We'll be updating this file in a little bit when we add an
+external dependency to our package.
+
+.. note:: By default the name of the current directory will be used for the
+	  package name, you can change this with ``-p <package-name>``.
```
phadej, line 55: I think it's better to supply package name explicitly...

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+
+Running the application
+^^^^^^^^^^^^^^^^^^^^^^^
+
+As mentioned above, ``cabal init`` with no arguments generates a package with a
+single executable that prints ``"Hello, Haskell!"`` to the terminal. To run the
+executable enter the following command:
+
+``cabal run :myfirstapp``
```
phadej, line 65: I think it's better to supply package name explicitly...

... and then this command can be literal.

Now I get

```
cabal: Unknown target ':myfirstapp'.
The package getting-started has no component 'myfirstapp'.
```

error.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+with ``cabal build``:
+
+``cabal build :myfirstapp``
+
+
+Adding dependencies
+^^^^^^^^^^^^^^^^^^^
+
+Next we'll add an external dependency to our application. The unit of
+shareability in the Haskell ecosystem is a `package`, the database of available
```
phadej, line 90: Cabal term is *repository*, or

> Hackage is the Haskell community's central package archive of open source software.

as Hackage itself puts it.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+Haskell packages is called `Hackage <https://hackage.haskell.org/>`__.
+
+In our application, we'll use a package called `haskell-say
+<https://hackage.haskell.org/package/haskell-say>`__ to print text to the
```
phadej, line 94: I wonder if there are any "better" example package. If users go to see that package they'll find `package.yaml` and `stack.yaml` and no CI setup and ...

If we put something at the very front (*the first 3rd party package a new user may see*), we have to put something of the highest quality. I'm biased, so I won't suggest any specific package here.

phadej:
As I commented, we should rather collect what we can simplify once 3.4 is released in separate issue, but write the guide with `3.2` in mind. Then this will be useful immediately.

jneira:
Yeah, i guessed it would work with latest version. 
But i would try to make the guide retrocompatible as much as possible, without adding nothing about to be obsolete or introduce any complexity. I would use `cabal-3.0if there is no strong reasons to disallow it, but `3.2` would be nice too. The guide must note the required version of cabal in a visible place.

F.e. the caret operator is great but simply changing the example to be

```cabal
executable myfirstapp
       main-is: Main.hs
       build-depends:
           base >=4.11 && <4.12,
           haskell-say >=1.0.0.0 && < 1.1
```

You could remove the note about the caret operator and it would look more simple (uniformed?) and it would be compatible with `cabal-3.0` at once.
Hopefully users will be able to discover the caret operator later.

phadej:
`cabal-install-3.0` supports caret operator as nicely. The `cabal-version: 2.4` is the important part in the generated `.cabal` file itself.

> Hopefully users will be able to discover the caret operator later.

They won't. It's safer to assume that they won't read beyond *Getting started* if things just work.

jneira:
Mmm, yeah, cabal supports it even before, but `cabal-3.0 init` by default does not generate a `.cabal` file compatible, and without the caret operator `cabal init --cabal-version=2.4` would not be needed, but it is fine.

> They won't. It's safer to assume that they won't read beyond Getting started if things just work.

Fair enough.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
+Getting Started with Haskell and Cabal
+======================================
+
+Installing the Haskell toolchain
+--------------------------------
+
+To install the Haskell toolchain follow the `ghcup instructions
+<https://www.haskell.org/ghcup/>`__.
```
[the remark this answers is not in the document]
m-renaud: Done.

Cabal/doc/getting-started.rst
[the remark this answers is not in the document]
m-renaud: Yeah, I had forgotten that there hasn't been a release since 2.4 was the default. Added the explicity `--cabal-version=2.4 --license=NONE -p myfirstapp`.

Cabal/doc/getting-started.rst
m-renaud: Added back, can remove after the next release.

Cabal/doc/getting-started.rst
m-renaud: Done.

Cabal/doc/getting-started.rst
m-renaud: Done. I guess I was assuming they were in a directory called `myfirstapp` in which case `cabal init` would create an executable named `myfirstapp` which you could then run with the `:<target>` syntax.

Cabal/doc/getting-started.rst
m-renaud: Inlined definition from Hackage.

Cabal/doc/getting-started.rst
m-renaud: I was looking for something very small, with zero dependencies, that did something fun, that works entirely in the terminal. We could easily add a test or two and a CI setup to that package if we want it to showcase everything (I can send a PR this week).

m-renaud:
Hey @jneira, thanks for the feedback!

> Installation method refer to an only *nix method ...

Added a link to a Windows walkthrough for getting Haskell set up.


> All console commands are posix, although some of them could be executed in cmd and/or powershell.

tbh I haven't used Windows in years, is there anything specific I should change here?

> ...cabal-version with support for the caret operator...

Yeah, oops, I had forgotten that the change to switch `cabal init`s Cabal spec version to 2.4 isn't in a release yet, updated instructions to pass `--cabal-version=2.4` explicitly.

Cabal/doc/getting-started.rst
m-renaud: And I think as far as `package.yaml` and `stack.yaml` go, these are both very widely used tools in the community today (regardless of personal opinions of them), so exposing a newcomer to them isn't necessarily a bad idea.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+   import HaskellSay (haskellSay)
+
+   main :: IO ()
+   main =
+     haskellSay "Hello, Haskell! You're using a function from another package!"
+
+``import HaskellSay (haskellSay)`` brings the ``haskellSay`` function from the
+module named ``HaskellSay`` into scope. The ``HaskellSay`` module is defined in
+the ``haskell-say`` packages that we added a dependency on above.
```
phadej, line 131: Related to your previous comment. I think it would be actually good to find some library which has few transitive dependencies.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Now you can build and re-run your code to see the new output:
+
+.. code-block:: console
+
+   $ cabal run
+       ________________________________________________________
```
phadej, line 138: there is output in between, we should use opportunity to explain what happens i.e. building and caching of dependencies. That's the whole point of using `cabal` and not just raw GHC!

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+          \    \  \    \  \-----------|
+           \    \  \    \  \          |
+            \    \  \    \  \---------|
+            /    /  /     \
+           /    /  /       \  \-------|
+          /    /  /    ^    \  \      |
+         /    /  /    / \    \  \ ----|
+        /    /  /    /   \    \
+       /____/  /____/     \____\
+
```
phadej, line 159: I think we should have a section *What next*. I don't know what to put there yet, I guess https://www.haskell.org/documentation/ is a good to have at the very least. Maybe we want highlight something from there, if it's good to know for beginner cabal users.

Cabal/doc/getting-started.rst
phadej: ... but I don't have anything specific in mind yet, *fun* is tricky requirement. So this is good for now.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Initializing the application
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Start by creating a ``myfirstapp`` directory to hold the project files:
+
+.. code-block:: console
+
+    $ mkdir myfirstapp && cd myfirstapp
```
jneira, line 28: in windows powershell (present by default in last windows versions) could be
```
>mkdir myfirstapp ; cd myfirstapp
```
in raw dos
```
> md myfirstapp && cd myfirstapp
```

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. code-block:: console
+
+    $ cabal init --cabal-version=2.4 --license=NONE -p myfirstapp
+
+This will generate the following files:
+
+.. code-block:: console
+
+    $ ls
```
jneira, line 40: `ls` would work in powershell, in dos we have to use `dir` (the output will be different in any case but whatever)

phadej: It looks like that if we ask windows users to use *PowerShell* than there're only small tweaks needed? casing for `cmd` would make it all too messy and hard to follow.

Cabal/doc/getting-started.rst
phadej:
```
mkdir myfirstapp ; cd myfirstapp
```

would work in unix shells to, but I think it's better to split into two separate commands

Cabal/doc/getting-started.rst
jneira: yeah, powershell is fine nowadays imo, it is the default console in windows 10

Cabal/doc/getting-started.rst
jneira: That is great, it could contain some short list of awesome things you can do with cabal, to motivate users to continue reading more advanced stuff, like:
```
* Create libraries, executables, test suites and benchmarks for your package
* Set up a project with multiple packages using `cabal.project`
* Upload your package to a common repository to share your code
* ....
```

Cabal/doc/getting-started.rst
m-renaud: I was thinking that in a (yet to be written) walkthrough on working with multiple packages in a project is where we could pull in more deps, wanted to keep this as simple as possible.

Cabal/doc/getting-started.rst
m-renaud: Yeah, lets discuss in a follow-up PR, would like to get a baseline checked in and then we can make tweaks from there.

Cabal/doc/getting-started.rst
m-renaud: Created a short blurb, lets bikeshed content on a follow-up PR.

Cabal/doc/getting-started.rst
m-renaud: Split into two separate commands.

Cabal/doc/getting-started.rst
m-renaud: Added some words about it working on unix shells and PowerShell if you're a Windows user.

Cabal/doc/getting-started.rst
```diff
@@ -0,0 +1,169 @@
...
+         /    /  /    / \    \  \ ----|
+        /    /  /    /   \    \
+       /____/  /____/     \____\
+
+
+What Next?
+----------
+
+Now that you know how to set up a simple Haskell package using Cabal, check out
+some of thee resources on the Haskell website's `documentation page
```
jneira, line 167: Typo: the

Cabal/doc/getting-started.rst
phadej: it's not about multiple packages. It's about showing that `cabal-install` does work for you. You *don't have to worry (as a beginner) if a dependency has own dependencies*, it **just works**  :)

Cabal/doc/getting-started.rst
m-renaud: Oops, will fix in follow-up, good catch :)

PR #6676, opened 2020-04-09
Default cabal init application-dir to app, and source-dir to src.

### LINKED ISSUE #6150: `cabal init` should not use same hs-source-dirs when choosing "Library and Executable"
This is bad practice as if an exe component depends on a lib component, GHC will recompile the modules, and not use the compiled library: double compilation of same modules.

---

An alternative is to implement a "don't compile modules we didn't told you to" to GHC, but I think this small step is still a good engineering practice, even `cabal` could tell (future) GHC to be less smart.

---

See https://github.com/haskell/cabal/issues/6143#issuecomment-513531999

m-renaud:
This sets the default directory for the main module as `app/` (`--application-dir=app`) and library source files as `src/` (`--source-dir=src`).

This avoids the problem described in #6150 of having a library and executable share the same `hs-source-dirs`. Even for just the executable case its best to have a separate directory because its likely that eventually code will be moved into internal libraries for testing (which should also have a separate `hs-source-dirs`).

Tested manually, both interactively and non-interactively.

This resolves #6150. 

---

### DIFF
--- a/cabal-install/Distribution/Client/Config.hs
+++ b/cabal-install/Distribution/Client/Config.hs
@@ -851,8 +851,8 @@ commentSavedConfig = do
             IT.cabalVersion    = toFlag IT.defaultCabalVersion,
             IT.language        = toFlag Haskell2010,
             IT.license         = NoFlag,
-            IT.sourceDirs      = Nothing,
-            IT.applicationDirs = Nothing
+            IT.sourceDirs      = Just [IT.defaultSourceDir],
+            IT.applicationDirs = Just [IT.defaultApplicationDir]
             },
         savedInstallFlags      = defaultInstallFlags,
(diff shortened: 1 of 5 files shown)


cabal-install/Distribution/Client/Config.hs
```diff
@@ -852,7 +852,7 @@ commentSavedConfig = do
             IT.language        = toFlag Haskell2010,
             IT.license         = NoFlag,
             IT.sourceDirs      = Nothing,
```
phadej, line 854: Should ` sourceDirs` have default set too?

m-renaud: Yes it should, thanks.

m-renaud:
Updated changelog.d/cabal-init as well.

PR #6678, opened 2020-04-09
Add FileCreators.generateCabalFile unit tests.

m-renaud:
I'll be switching `cabal init` to use `PrettyField` in an upcoming PR, this will make it easier to verify that the output remains approximately the same (modulo better formatting).

### DIFF
--- a/cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
@@ -1,0 +1,151 @@
+module UnitTests.Distribution.Client.Init.FileCreators (
+  tests
+  ) where
+
+import Distribution.Client.Init.FileCreators
+  ( generateCabalFile )
+
+import Test.Tasty
+import Test.Tasty.Golden (goldenVsString)
+
+import System.FilePath
+  ( (</>) )
+import qualified Data.ByteString.Lazy as BS
+import qualified Data.ByteString.Lazy.Char8 as BS8
+import qualified Data.Set as Set
+
+import Distribution.Client.Init.Types
+  ( InitFlags(..), PackageType(..), defaultInitFlags )
+import Distribution.Simple.Setup
+  ( Flag(..) )
+
+import Distribution.CabalSpecVersion
+  ( CabalSpecVersion(CabalSpecV2_4) )
+import Distribution.Types.Dependency
+  ( Dependency, mkDependency )
+import Distribution.Types.LibraryName
+  ( LibraryName(LMainLibName) )
+import Distribution.Types.PackageName
+  ( mkPackageName )
+import Distribution.Types.VersionRange
+  ( majorBoundVersion )
+import Distribution.Types.Version
+  ( mkVersion )
+import qualified Distribution.ModuleName as ModuleName
+  ( fromString )
+import qualified Distribution.SPDX as SPDX
+import Language.Haskell.Extension ( Language(..) )
+
+tests :: [TestTree]
+tests = [ testGroup "cabal init goldens"
+          [ checkCabalFileGolden exeFlags "exe-only-golden.cabal"
+          , checkCabalFileGolden libExeAndTestFlags "lib-exe-and-test-golden.cabal"
+          ]
+        ]
+
+checkCabalFileGolden :: InitFlags -> FilePath -> TestTree
+checkCabalFileGolden flags goldenFileName =
+  goldenVsString goldenFileName goldenFilePath generatedCabalFile
+  where
+    goldenFilePath :: FilePath
+    goldenFilePath = "tests" </> "fixtures" </> "init" </> goldenFileName
+
+    generatedCabalFile :: IO BS.ByteString
+    generatedCabalFile = pure . BS8.pack $ generateCabalFile goldenFileName flags
+
+-- ==================================================
+-- Base flags to set common InitFlags values.
+
+baseFlags :: InitFlags
+baseFlags = defaultInitFlags {
+  -- Values common to all (or most) test flags.
+    packageName = Flag (mkPackageName "foo")
+  , noComments = Flag False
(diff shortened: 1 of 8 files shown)


cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
```diff
@@ -0,0 +1,171 @@
...
+
+exeFlags :: InitFlags
+exeFlags = baseTestFlags {
+    packageType = Flag Executable
+  , applicationDirs = Just ["app"]
+  }
+
+expectedExeOutput :: String
+expectedExeOutput =
+     "cabal-version:       2.4\n"
```
phadej, line 60: Use https://hackage.haskell.org/package/tasty-golden

Updating these by hand will be PITA. `tasty-golden` allows run tests with `--accept`. Cabal have some test using golden-tests already (roundtrip in `parser-tests`)

m-renaud: TIL, thanks! That's so much better.

cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
```diff
@@ -0,0 +1,157 @@
...
+    goldenFilePath = inGoldensDir goldenFileName
+
+    generatedCabalFile :: IO BS.ByteString
+    generatedCabalFile = pure . BS8.pack $ generateCabalFile goldenFileName flags
+
+-- ==================================================
+-- Simple exe.
+
+exeFlags :: InitFlags
+exeFlags = InitFlags {
```
phadej, line 66: could this rather be specified in terms of `defaultInitFlags`, or some `commonInitFlags` derived from `defaultInitFlags` (to set package name etc.). Adding new test shouldn't require 40 lines.

cabal-install/cabal-install.cabal
```diff
@@ -90,6 +90,8 @@ Extra-Source-Files:
   tests/IntegrationTests2/targets/tests-disabled/q/q.cabal
   tests/IntegrationTests2/targets/variety/cabal.project
   tests/IntegrationTests2/targets/variety/p.cabal
+  tests/UnitTests/Distribution/Client/Init/goldens/generateCabalFile/exe-only.cabal
+  tests/UnitTests/Distribution/Client/Init/goldens/generateCabalFile/lib-exe-and-test.cabal
```
phadej, line 94: they could be somewhere less burried. `tests/fixtures/init/*.cabal` is "deep" enough.

I'd like to avoid mixing "data" and "src" directories.

cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+
+-- ==================================================
+-- Simple exe.
+
+exeFlags :: InitFlags
+exeFlags = baseFlags {
+  -- Create an executable only, with main living in app/Main.hs.
+    packageType = Flag Executable
```
phadej, line 107: This is great :+1:

phadej:
Please merge when CI is green

cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
```diff
@@ -0,0 +1,152 @@
...
+    goldenFilePath = "tests" </> "fixtures" </> "init" </> goldenFileName
+
+    generatedCabalFile :: IO BS.ByteString
+    generatedCabalFile = pure . BS8.pack $ generateCabalFile goldenFileName flags
+
+-- ==================================================
+-- Base flags to set common InitFlags values.
+
+baseFlags :: InitFlags
+baseFlags = InitFlags {
```
phadej, line 60: Ah, sorry. I'd still like this to use `defaultInitFlags`, but I'll could make that change myself.

I want to see "how much" non-defaults there are.

cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
m-renaud: Done.

cabal-install/cabal-install.cabal
m-renaud: Done.

cabal-install/tests/UnitTests/Distribution/Client/Init/FileCreators.hs
m-renaud: Ahhh gotcha, switched to use `defaultInitFlags` as the base (it only sets `initVerbosity` which I've removed from the list here).

m-renaud:
Alright, lets see if I can get those metadata checks to pass. It's currently not possible to run the `make gen-extra-source-files` tool, there's a base version conflict since `cabal-dev-scripts.cabal` has the constraints `base >=4.10 && <4.13` but the rest of the packages have `<4.14`. Should I just bump the upper bound to `<4.14` there as well?

phadej:
Let's allow only one-two GHC versions for development tools, I guess GHC-8.8 is "safe" to use now, yet I (and CI) still use 8.6. So `base >=4.12 && <4.14`.

That said. I'm not keen to upgrade *the* GHC used for `cabal` development tools too often. So we'll lag behind the "newest and shinest" GHC. I'm happy to add `with-compiler: ghc-8.6.5` to `cabal.project.meta`.

m-renaud:
Gotcha, I got the tests passing anyways so lets leave the tools as is. Just waiting for the last checks to finish then I'll merge.

m-renaud:
Hmmmmmmmm, it appears that "Merge" on the official GitHub mobile app does not do the same thing as "Merge" from the web interface :confused: Seems to not have created a merge commit for the PR, bizarre, guess I won't be doing that again.

PR #6693, opened 2020-04-11
Add analytics to user guide.

m-renaud:
This uses Google Analytics to track site views and interactions.

## Benefits

- Determining which locales access the docs the most often
  - Benefit: Informs us if it would be useful to invest in an i18n version of the docs, and if so which languages
- Detecting dead links and duplicate pages, info missing from sitemap, etc.
  - Benefit: Improves SEO, letting users find what they're looking for more easily
- Bounce rate on individual pages to give insight into when users aren't finding what they're looking for on a page
  - Benefit: Informs us where attention should be focused in the existing docs
- Referrer
  - Benefits: Lets us determine if there's other locations that we want links to the docs to be visible, but is being missed (ex. we expect users to get to the user guide from https://www.haskell.org/cabal/, but we don't know if users are actually finding it on the page, so we could try out different placements and formats)

### DIFF
--- a/Cabal/doc/_templates/layout.html
+++ b/Cabal/doc/_templates/layout.html
@@ -1,5 +1,18 @@
 {% extends "!layout.html" %}
 
+{%- block extrahead %}
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-42745779-5"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
+    gtag('js', new Date());
+
+    gtag('config', 'UA-42745779-5');
+  </script>
+{% endblock %}
+
+
 {% block menu %}
   {{ super() }}
   <a href="cabal-projectindex.html">Reference</a>


phadej:
Who can access the dashboard?

Yet even if that's controlled, I'm against trackers, so there should be really really good need and perfectly not owned by a "big evil corporation".

This should been first discussed as an issue, this is not "small thing".

m-renaud:
> Who can access the dashboard?

Was going to add everyone who's an admin of the Cabal ReadTheDocs project. 

> This should been first discussed as an issue, this is not "small thing".

Yes, you're right, if you'd like I can open a separate issue, or we can discuss on this PR, doesn't matter to me. I should have listed the rationale below in the PR description (I'll also copy it above).

The main reasons for enabling analytics would be:

- Determining which locales access the docs the most often
  - Benefit: Informs us if it would be useful to invest in an i18n version of the docs, and if so which languages
- Detecting dead links and duplicate pages, info missing from sitemap, etc.
  - Benefit: Improves SEO, letting users find what they're looking for more easily
- Bounce rate on individual pages to give insight into when users aren't finding what they're looking for on a page
  - Benefit: Informs us where attention should be focused in the existing docs
- Referrer
  - Benefits: Lets us determine if there's other locations that we want links to the docs to be visible, but is being missed (ex. we expect users to get to the user guide from https://www.haskell.org/cabal/, but we don't know if users are actually finding it on the page, so we could try out different placements and formats)

To address privacy concerns, from the [ReadTheDocs analytics integration](https://docs.readthedocs.io/en/stable/guides/google-analytics.html): "Read the Docs takes some extra precautions with analytics to protect user privacy. As a result, users with Do Not Track enabled will not be counted for the purpose of analytics." In addition, User-Id tracking (which allows cross-device session attribution) is disabled.

phadej:
The page you linked: https://docs.readthedocs.io/en/stable/guides/google-analytics.html

> Going to Admin > Advanced Settings in your project.
> Fill in the Analytics code heading with your Google Tracking ID (example UA-123456674-1)

i.e. not by adding snippet to the source. Please open an issue.

m-renaud:
Done, see #6695.

PR #6694, opened 2020-04-11
Add sitemap.xml to Cabal User Guide.

m-renaud:
This gives search engines more info about the structure of the pages, and which
pages to give more weight to, such as "latest" and "stable" versions of the docs
compared to older versions, or preferring a canonical latest/ page instead of
latest/index.html to avoid duplication in search results.

### DIFF
--- a/Cabal/doc/_static/sitemap.xml
+++ b/Cabal/doc/_static/sitemap.xml
@@ -1,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset
+      xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9
+            http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
+
+<url>
+  <loc>https://cabal.readthedocs.io/en/latest/</loc>
+  <lastmod>2020-04-11T12:00:00+00:00</lastmod>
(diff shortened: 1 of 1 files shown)


phadej:
Does this have to be maintained by hand? Then it's no-go.

m-renaud:
I'm not sure I follow? You need some way of specifying metadata about the content of the site, the alternative could be some script that reads metadata from individual files that you tag in RST somehow, or to use some other config file which contains the same information. In general, site structure changes pretty infrequently so I don't think its too onerous to add a new entry to the sitemap if a new page is added. Otherwise it can remain the same. Even a slightly out of date sitemap is significantly better than no sitemap at all.

m-renaud:
If you're asking about the `<lastmod>` tag that's not particularly important, but can be useful if you want to force a re-crawl of the content.

phadej:
I'm afraid adding "one more thing" which can cause stuff to work or not. I guess there are tools to verify that sitemaps are up to date. Such development documentation have to be written before I'll be comfortable.

Also, there is https://docs.readthedocs.io/en/stable/guides/sitemaps.html

> Read the Docs automatically generates a sitemap for each project that hosts to improve results when performing a search on these search engines. 

Which generates: https://cabal.readthedocs.io/sitemap.xml

What's wrong with that? If you want to recommend latest over stable, that's indeed something we can discuss in an issue.

phadej:
@m-renaud ping

m-renaud:
Sorry for the delay, the one that's auto-generated doesn't have any of the sub-pages, although it appears search engines still do the right thing with the sub-pages so maybe it's fine :shrug: 

I'm fine with closing this issue since I don't have tonnes of spare time right now.

Issue #6213, opened 2019-08-29
Update docs for 3.x release: remove `v2-` prefix and unflag as experimental 


gbaz:
The cabal users guide still describes the v2- stuff with the prefix and says it is experimental.

https://cabal.readthedocs.io/en/latest/nix-local-build.html

Now that it is the standard path, shouldn't we at least do a small update to the docs to get rid of all the v2- prefix stuff and indicate that this is the normal mode of operation?

m-renaud:
See #6194 for some ideas I had to improve the docs, this would include at the minimum updating everything for `v2-` as the default without the prefix.

gbaz:
I think that's a great proposal. But we have a cabal out _today_ that the docs don't explain correctly. We should fix that first, and independently worry about making the docs better. For now let's just make sure they're correct :-)

m-renaud:
Reviving this thread :) @gbaz are there specific concerns you have about updating docs to move to 3.0 now considering how long it has been the default?

To me fixing any incorrectness in existing docs and switching to 3.0 seem completely orthogonal to one another. If you can give me some specific actionable changes you'd like to see I can take a look at those, otherwise I'd like to decouple them and move forward with updating the references to v2. Let me know!

ulysses4ever:
Looks like it will be closed by #9087.

Issue #6105, opened 2019-06-23
s/nix style/v2-build/ in the documentation


phadej:
I conjecture that mentioning Nix only adds confusion. Almost all people I know use `v2-build` (well actually they use `new-build`, but maybe `v2-build` is a bit better).

E.g.

- One of the primary innovations of Nix-style local builds is the distinction between local packages, which users edit and recompile and must be built per-project, versus external packages, which can be cached across projects.
+ One of the primary innovations of `v2-build` local builds is the distinction between local packages, which users edit and recompile and must be built per-project, versus external packages, which can be cached across projects.

phadej:
An example of confusion: Nix store can (and does) mean `/nix/store` for a lot of people.

23Skidoo:
Agreed.

fgaz:
+1, but it should also mention the other temporary names (nix-style local builds, new-*) so that at least it shows up in searches

hvr:
But do we want to keep referring to `v2-build` forever? I assumed we'd phase out the term `v2-build` to become `cabal build`, at which point referring to it as `v2-build` will be more confusing than using the term we've been advertising this feature ever since cabal 1.24 (the term "new-build" and "v2-build" have been rather ephemeral in contrast to the term "nix-style local builds" which has been a constant throughtout time). And I've myself been referring to new-build and v2-build frequently also as "Nix-style" where it made sense (specifically when talking about the nix-style store)

phadej:
I think @hvr did veto this. Nix-style it is.

phadej:
Actually no.

https://cabal.readthedocs.io/en/latest/nix-local-build-overview.html starts with

> Nix-style local builds are a new build system implementation inspired by Nix. The Nix-style local build system is commonly called “v2-build” for short.

That needs to be changed in the way or another. Either

1. we avoid "v2-build" and use nix-style in non-command-name context
2. or "v2-build" is used in place of informal "nix-style"

Having consistent language would make documentation more polished almost for free.

hvr:
well, note that

> ... commonly called “v2-build” for short....

was changed into that wording in 2017 via c6928316a5f72b4c4f367f157f1e8dafe71c13ea

```diff
-``cabal new-build``, also known as Nix-style local builds, is a new
-command inspired by Nix that comes with cabal-install 1.24. Nix-style
-local builds combine the best of non-sandboxed and sandboxed Cabal:
+Nix-style local builds are a new build system implementation inspired by Nix.
[...]

Issue #6104, opened 2019-06-23
Documentation: v1 to v2 migration guide.


phadej:
cabal-install-3.0 changelog / release notes should point out to that part of documentation as the first things.

There will be a lot of confusion when 3.0 is released, and there are no migration guide explaining how `v2-build` or `v2-install` are different from `v1-build` or `v1-install`. The QuickStart section is not the right one, as it *doesn't explain the differences*.

23Skidoo:
BTW, I've started writing a 3.0 migration guide for custom Setup authors: https://github.com/haskell/cabal/wiki/3.0-migration-guide. Help appreciated...

okeuday:
Please describe in the migration guide what to do with the v2 (Nix-style local builds) commands to get the same sandbox functionality that was present with the v1 sandbox command.

phadej:
There are no sandboxes in `v2` workflows. They are unneeded.

23Skidoo:
Right, but the migration guide should explain why.

GeorgeCo:
I think it is important to get this into 3.2 and if possible something on ghc-pkg,  if only a reference to the ghc issue: haskell/cabal#6342

phadej:
I de-milestone this issue.

okeuday:
Still not seeing how to use local filesystem dependencies without sandboxes.  I noticed that compiling cabal-install 3.2.0.0 with GHC 8.10.1 required manually fixing dependency versions in bootstrap.sh (cabal-install 3.0.0.0 had a similar problem), so Cabal 3 still needs more testing?

ulysses4ever:
I think this ship has sailed. Let's close this umbrella issue and solicit concrete improvements (like description of how v2 replace sandboxes) if anyone is interested?

PR #6703, opened 2020-04-12
Remove v2- prefix from command, package, and project docs.

m-renaud:
The v2- style commands have been default since version 3.0.0.0, so when looking at the latest version of the docs it should assume the defaults. I left the explicit v2- prefix in the nix-style docs since it is explicitly discussing the difference between the old and new style commands, and also because there is (was?) ongoing discussion about what prefix (if any) to use in #6105.

Also fix some small errors in the text.

This partially addresses #6194 and #6213.

### DIFF
--- a/Cabal/doc/cabal-commands.rst
+++ b/Cabal/doc/cabal-commands.rst
@@ -4,24 +4,24 @@ cabal-install Commands
 We now give an in-depth description of all the commands, describing the
 arguments and flags they accept.
 
-cabal v2-configure
+cabal configure
 -------------------
 
-``cabal v2-configure`` takes a set of arguments and writes a
+``cabal configure`` takes a set of arguments and writes a
 ``cabal.project.local`` file based on the flags passed to this command.
-``cabal v2-configure FLAGS; cabal new-build`` is roughly equivalent to
-``cabal v2-build FLAGS``, except that with ``new-configure`` the flags
-are persisted to all subsequent calls to ``v2-build``.
+``cabal configure FLAGS; cabal build`` is roughly equivalent to
+``cabal build FLAGS``, except that with ``configure`` the flags
+are persisted to all subsequent calls to ``build``.
 
-``cabal v2-configure`` is intended to be a convenient way to write out
+``cabal configure`` is intended to be a convenient way to write out
 a ``cabal.project.local`` for simple configurations; e.g.,
-``cabal v2-configure -w ghc-7.8`` would ensure that all subsequent
-builds with ``cabal v2-build`` are performed with the compiler
+``cabal configure -w ghc-7.8`` would ensure that all subsequent
+builds with ``cabal build`` are performed with the compiler
 ``ghc-7.8``. For more complex configuration, we recommend writing the
 ``cabal.project.local`` file directly (or placing it in
 ``cabal.project``!)
 
-``cabal v2-configure`` inherits options from ``Cabal``. semantics:
+``cabal configure`` inherits options from ``Cabal``. semantics:
 
 -  Any flag accepted by ``./Setup configure``.
 
@@ -42,10 +42,10 @@ to happen if a flag actually applied to every transitive dependency). To
 apply options to an external package, use a ``package`` stanza in a
 ``cabal.project`` file.
 
-cabal v2-update
+cabal update
 ----------------
 
-``cabal v2-update`` updates the state of the package index. If the
+``cabal update`` updates the state of the package index. If the
 project contains multiple remote package repositories it will update
 the index of all of them (e.g. when using overlays).
 
@@ -53,13 +53,13 @@ Some examples:
 
 ::
 
-    $ cabal v2-update                  # update all remote repos
-    $ cabal v2-update head.hackage     # update only head.hackage
+    $ cabal update                  # update all remote repos
+    $ cabal update head.hackage     # update only head.hackage
 
-cabal v2-build
+cabal build
 ---------------
 
-``cabal v2-build`` takes a set of targets and builds them. It
+``cabal build`` takes a set of targets and builds them. It
 automatically handles building and installing any dependencies of these
 targets.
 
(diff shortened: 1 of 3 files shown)


phadej:
Does all documentation which talks about `v1-` commands, explicitly has `v1-` prefix? I think that should be checked first.

m-renaud:
I went through all the docs and didn't see anything that was clearly v1 that should be prefixed, afaik the usages were all pretty generic. Of course there's tonnes of options documentation which may or may not all work for v2 commands, but trying out every single option I'd say is out of scope.

phadej:
If it doesn't say explicitly either, and is not clearly `v2-` then I think we should first make a change to prefix everything we are not sure about to `v1-`.

I don't want end in situation where we have (implicitly) `v2-` documentation which actually doesn't work. I don't remember anything outside specifically nix-style documentation to be so.

**we should emphasize** (using different words though) that similarities between `v1-build` and `v2-build` are almost as accidental as similarities between `cabal build` and `stack build`. You cannot just change one to another and expect that things continue to work. They might, but it is far from guaranteed.

---

I think that we should be explicit about the prefix for as long as possible (i.e. `v1-` commands are around). It's not pretty, but it is explicit.

phadej:
I also want mention that a person ranted today on `#hackage` that `cabal` broke everything by changing what `build` command does. So there are people "stuck" in `v1-build`. I think it's better if there is longer period when Cabal manual doesn't talk about prefixless `build` command at all. (We should have done that edit many years ago, but ...). Not to mention that there are no progress on #6104. We should not concentrate solely on beginner (who start "fresh") needs

m-renaud:
> If it doesn't say explicitly either, and is not clearly v2- then...

Most of the pages in the docs don't reference commands at all, the commands in:

* Sections 1-4 all work with the v2 commands (and are relatively surface level commands, see [here](https://cabal.readthedocs.io/en/latest/intro.html#a-tool-for-working-with-packages), [here](https://cabal.readthedocs.io/en/latest/intro.html#similar-systems), and [here](https://cabal.readthedocs.io/en/latest/installing-packages.html#installing-packages-from-hackage), 
* Section 5 explicitly has prefixes because it's explicitly talking about the difference with the old system
* Section 6 is specifically about the cabal-install commands of the **current version of the cabal-install tool you're looking at the docs for**
* Section 7 is about the cabal package format and only has references to bare `cabal build|run|repl|test|benchmark` with no options.
* Section 8 is cabal project file reference (very few cabal command references, all simple invocations)
* Section 9 is Setup.hs (remains unchanged since it refers explicitly to `runhaskell Setup.hs`)
* Section 10 is file format history (no commands)
* Section 11 is field syntax description (no commands)
* Section 12 is how to report bugs (no commands)
* Section 13 is Nix Integration, which is affected but has an explicit note at the top
  * "This functionality doesn’t work with nix-style builds. Nix-style builds are not related to Nix integration."

> I also want mention that a person ranted today on #hackage that cabal broke everything by changing what build command does

I sympathize, but I don't see exactly how its relevant to how the docs for the most recent release of the tool should be written. If someone is using an older version of cabal (or at least an older version's functionality) then they should (and can) look at docs for a previous release, that's why docs are versioned as well.

> We should not concentrate solely on beginner (who start "fresh") needs

This doesn't really have anything to do with beginners, if you're looking at **`v: latest`** docs, the docs should reflect the current state of the **latest** release of the tool, not a conglomeration of all previous versions and mismatches in semantics, historical notes, etc.

> we should emphasize (using different words though) that similarities between...

How about this: I think one way of doing this would be to put a NOTE at the start of Section 6 (cabal-install Commands, since that's the only one that actually has any details about the commands or their arguments) that states something along the lines of: _"If you are using the old/v1-style build infrastructure, please take a look at the docs for version X.Y (and link)."_ I agree there should be some place that calls out the difference between the old and new style, but that legacy shouldn't be littered throughout every single page of the docs.

The docs have been languishing for months (years?) and it seems to be because there's a desire to have the docs be 100% perfect for v2-style before making any changes, but I'd argue the current incredibly outdated version of the docs at head is more harmful. Yes, documentation gets out of date with code, and as a result we should be reactive to fix it whenever issues are discovered. I'd really like to see the docs get better, but without concrete feedback on what you'd like to see I'm not sure if there's much else I can do.

phadej:
I don't think that having `v2-` prefixes is *littering with legacy*.

I'd like @hvr and @23Skidoo to comment, as we two are not in agreement.

m-renaud:
Yeah that sounds good, I'm fine with whatever the decision is, just wanted to lay out my line of thinking. Agreed that Herbert and Mikhail should definitely be involved, are there other individuals or groups that we should reach out to about general structure of the docs? How widely do you think we should open up the discussion?

m-renaud:
It's been over a year since this PR was opened. At this point the v2- commands are ubiquitous so having them be explicit seems irrelevant now. Wdyt?

Mikolaj:
@m-renaud: I'm afraid you are right. I have scripts that mention v2- explicitly so, while I'm fine to change the scripts, it's too late to prevent creating such scripts by not mentioning v2- in the docs or discouraging their use.

OTOH, obviously at some point v1- commands will vanish and, probably later or perhaps at the same time, the v2- aliases will too, so perhaps in-between is a good time to remove v2- from documentation.

jneira:
maybe it is time to reconsider merge this @Mikolaj ?

Mikolaj:
> maybe it is time to reconsider merge this @Mikolaj ?

I won't object.

@andreasabel, @gbaz, @fgaz, other wonderful devs, what do you think?

andreasabel:
I am rather with @phadej (in his April 2020 incarnation).  Could we do the following?
- Change the docs so that _every_ command is either prefixed with `v1-` or `v2-`.
- Wait some more.
- Then (after `v1` has disappeared or truly marginalized) find-and-replace `v2-` by nothing.

fgaz:
@andreasabel there's basically no trace of v1- in the docs though

Mikolaj:
>     * Then (after `v1` has disappeared or truly marginalized) find-and-replace `v2-` by nothing.

I don't think `v1` is going away any time soon, so I don't think we should let it block stuff. Whether it's marginalized would be a hotly debated topic as long as it stays in.

I guess `v1` should almost never appear in documentation (just in lists of commands, marked as a backward compatibility for legacy workflows that can't be (easily or at all) expressed with the new idiom). So rather than marking its appearances in docs with `v1-`, the documentation fragment in question should be modernized to how `v2-` operates IMHO.

ulysses4ever:
My 2c. Working in academia, I care about newcomer experience a lot. From this standpoint, my desire is to have v2 prefixes hidden in the docs most of the time. This change can be augmented with a separate section on history and explanation of v1 vs v2, and a section on v1. The rest of docs mention no prefixes and describing the today’s default (which is v2) without mentioning the legacy (v2 implicitly references the legacy in today’s doc).

Mikolaj:
So what's the action plan? Could the patrons that have reservations about this PR have a look at the docs and suggest concrete additional improvements that need to be done before this can be merged? Would some disclaimers and links in strategic places informing about the existence of v1- and that v2- is now the default be a good enough precaution?

gbaz:
At this point, I think v2- usage is widespread enough that we can merge this PR. Anyone that still wants to use v1 by now knows it exists and knows how it works.

gbaz (approved):

jneira:
@m-renaud thanks for the pr and your persistence, would be too much asking for a final rebase?

jneira (approved):
lgtm, thanks again!

Mikolaj:
Right, we haven't heard any loud objections so once this is rebased, we should merge and handle any fallout afterwards. Move fast and break things. ;)

jneira:
Superseeded by #8020 to merge changes done here asap, thanks again @m-renaud!

Mikolaj:
Citing @jneira from the new PR: Like all prs a rendered version of changed docs can be chacked here: https://cabal--8020.org.readthedocs.build/en/802

We are going to merge soon (1 more review required). Please kindly have a look and start fixing whatever's missing in the new version.

PR #6707, opened 2020-04-13
Resolve #6691: Remove upgrade, uninstall and win32selfupgrade commands

### LINKED ISSUE #6691: `cabal upgrade` leads to odd diagnostic
**Describe the bug**
I inadvertently tyoed `cabal upgrade` instead of `cabal update` and got a strange diagnostic

**To Reproduce**

```
$ cabal upgrade
fromFlag NoFlag. Use fromFlagOrDefault
  error, called at ./Distribution/Simple/Flag.hs:98:21 in Cabal-3.2.0.0 w73VV9T25vCb2mhueKgfz:Distribution.Simple.Flag
  fromFlag, called at main/Main.hs:909:15 in main:Main

$ cabal v2-upgrade
cabal: unrecognised command: v2-upgrade (try --help)
```

**Expected behavior**
So far as I aware, `upgrade` is not a valid `cabal` command so I should get an `unrecognised command` diagnostic.

**System information**
 - Operating system: MacOS Catalina
 - `cabal`, `ghc` versions: cabal 3.2.0.0, ghc 8.8.3

**Additional context**
This is obviously not a deal breaker, but it may point to a deeper bug in `cabal` command line parsing.

phadej:
All three were hidden commands, first two were unimplemented
and the last one no-one knows about.

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -30,8 +30,6 @@ module Distribution.Client.Setup
     , defaultSolver, defaultMaxBackjumps
     , listCommand, ListFlags(..)
     , updateCommand, UpdateFlags(..), defaultUpdateFlags
-    , upgradeCommand
-    , uninstallCommand
     , infoCommand, InfoFlags(..)
     , fetchCommand, FetchFlags(..)
     , freezeCommand, FreezeFlags(..)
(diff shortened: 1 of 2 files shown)


23Skidoo:
What's the situation with symlinks on Windows? I had the impression that `win32selfupgrade` is still used by `cabal install cabal-install`.

phadej:
By `cabal v1-install cabal-install`, yes. Symlinking is used on windows when it works, not universally.

23Skidoo:
Right, so I think we shouldn't remove it. v2 code might also need it for the no-symlink case (I think there was a PR or ticket about adding a file-copying fallback).

PR #6718, opened 2020-04-19
Use PrettyField to format cabal file in `cabal init` 

### LINKED ISSUE #5555: cabal init generates badly formatted file

m-renaud:
This PR replaces the hard-rolled cabal file formatting with PrettyField.

**Notable differences:**
  - Uses 4 space indent instead of 2
  - Long lists (extra-source-files, build-depends, etc.) are formatted nicely (not just one long list on a single line)

This resolves #5555.

## Open Questions

**Unspecified/any version range formatting**

It appears not to be possible to pretty-print a dependency of the form: `<package-name>` with no version constraints, the only matching constructor for `VersionRange` is `AnyVersion` which is printed as `<package-name> -any`. This "unspecified" version range is needed to express a dependency from the package executable to the library within the same package (as is created in `--libandexe`).

```
executable foo
    build-depends:
        foo -any
```

Looking for input on a solution here, we could introduce a new `UnspecifiedVersion` constructor such that:

```
parsecSimple "" :: VersionRange
Just UnspecifiedVersion

pretty UnspecifiedVersion
Doc.empty
```

**How to represent standalone comments in `PrettyField`**

Currently (see below) I've created a new constructor in `PrettyField` for a "commented out" field that can also have comments. Should this be generalized to just `PrettyComment`? Thoughts on implementation here?

## Implementation Notes

In order to re-use the pretty-printing logic I needed to expose the specific list format functions from `FieldGrammar.hs`. These functions mostly wrap list fields in newtypes so the pretty printer knows how to lay them out.

The one thing that was not easy to express with the existing `PrettyField` construct was commented out fields that also had comments associated with them. To address this I added a new constructor to the `PrettyField` type, namely `PrettyFieldCommentedOut`, this is pretty specific and should probably just be renamed to `PrettyComment` or similar, lmk what you think about this.

## Testing

- Cabal golden file (see updates)
- Manual testing

---


Cabal/Distribution/Fields/Pretty.hs
```diff
@@ -34,6 +34,7 @@ import qualified Text.PrettyPrint as PP
 
 data PrettyField ann
     = PrettyField ann FieldName PP.Doc
+    | PrettyFieldCommentedOut ann FieldName
```
phadej, line 37: This doesn't look right.

I think we should rather extend `ann` to carry notion whether thing is "commented" out. E.g. whole `PrettySection` could be commented out - that sounds as "useful".

I.e. maybe there should be `showFields'` variant which also takes `(ann -> Bool)` argument and comments out stuff if the annotation says so.

Cabal/Distribution/PackageDescription/FieldGrammar.hs
```diff
@@ -542,3 +552,28 @@ setupBInfoFieldGrammar def = flip SetupBuildInfo def
     <$> monoidalFieldAla "setup-depends" (alaList CommaVCat) L.setupDepends
 {-# SPECIALIZE setupBInfoFieldGrammar :: Bool -> ParsecFieldGrammar' SetupBuildInfo #-}
 {-# SPECIALIZE setupBInfoFieldGrammar :: Bool ->PrettyFieldGrammar' SetupBuildInfo #-}
+
+-------------------------------------------------------------------------------
+-- Field formatters - Define how field values should be formatted for 'pretty'.
```
phadej, line 557: This comment is not right. These functions also tell how fields are parsed as well. There is `unpack` for them as well.

cabal-install/Distribution/Client/Init/FileCreators.hs
```diff
@@ -327,241 +339,277 @@ findNewName oldName = findNewName' 0
...
+     , fieldP "other-extensions" (maybeToFlag (formatOtherExtensions <$> otherExts c))
+       ["LANGUAGE extensions used by modules in this package."]
+       True
+
+     , fieldP "build-depends" (maybeToFlag (formatDependencyList <$> buildDependencies))
+       ["Other library packages from which modules are imported."]
+       True
+
+     , fieldP "hs-source-dirs"
+       (maybeToFlag (formatHsSourceDirs <$> case buildType of
```
phadej, line 454: I'd make `fieldPAla` variant which takes `Newtype a b =>`. and `(a -> b)` "pack" function.

Yet, I wonder what we should do so we don't need to copy `formatHsSourceDirs`. The `hs-source-dirs`  + `formatHsSourceDirs` coupling is already done in `FieldGrammar` module, we shouldn't need to copy & paste that.

Cabal/Distribution/Fields/Pretty.hs
phadej: Related to this. Could you add a golden file which generates commented out fields before this PR.

cabal-install/Distribution/Client/Init/FileCreators.hs
m-renaud: Yeah, I was thinking about that as well. The problem with the `FieldGrammar` module is it also ties it to where the field exists in the large structure (the `ALens' s a'` part of most of the typeclass functions). I'm wondering if we should re-package the field name and (optional) newtype pack into some structure that could then be shared, that's definitely a more invasive change though.

I went with the approach here of exporting the `formatXyz` (which as pointed out isn't exactly accurate because you could also use it to `unpackXyz`) because it was the least invasive and kept the newtype packing and formatting as a unit without having to expose the implementation details (aka the newtype wrapper) of any given "formatting". We could instead of exporting `formatHsSourceDirs` export the `FSep FilePathNT` part of it (as you mentioned), but then you lose the semantic level of "this is how you format the hs-source-dirs field" and instead you leak the low level implementation which is "you pack them using `FSep`.

Also, if at some point in the future you wanted to change how hs-source-dirs was formatted you wouldn't need to change anything outside of `FieldGrammar.hs` in the current form. Conversely, if you used the `fieldPAla` approach then you'd need to know that when you change the `FSep` newtype wrapper in `FieldGrammar.hs` then you also need to change `fieldPAla` in `Init/FileCreators.hs` which is bound to be forgotten.

Cabal/Distribution/PackageDescription/FieldGrammar.hs
m-renaud: Any suggestions for a better name which encompasses both "format" and "unpack"? Or maybe just change the comment to:

"Define how field values should be formatted and parsed."

And rename the functions to something like `annotateHsSourceDirs` (name tbd)?

cabal-install/Distribution/Client/Init/Utils.hs
```diff
@@ -28,6 +29,10 @@ eligibleForTestSuite flags =
   Flag True == initializeTestSuite flags
   && Flag Executable /= packageType flags
 
+-- | Witness the isomorphism between Maybe and Flag.
+maybeToFlag :: Maybe a -> Flag a
```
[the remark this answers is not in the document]
m-renaud: Ahh cool, updated.

cabal-install/Distribution/Client/Init/FileCreators.hs
```diff
@@ -327,241 +339,277 @@ findNewName oldName = findNewName' 0
...
-  renderStyle style { lineLength = 79, ribbonsPerLine = 1.1 } $
-  -- Starting with 2.2 the `cabal-version` field needs to be the first line of the PD
-  (if specVer < CabalSpecV1_12
-   then fieldS "cabal-version" (Flag $ ">=" ++ showCabalSpecVersion specVer)
-   else fieldS "cabal-version" (Flag $ showCabalSpecVersion specVer))
-      Nothing
+generateCabalFile fileName c =
+    showFields id $ catMaybes
+  [ (if specVer < CabalSpecV1_12
+       then fieldS "cabal-version" (Flag $ ">=" ++ showCabalSpecVersion specVer)
```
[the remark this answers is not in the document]
m-renaud: Gotcha, updated to use the newtype wrapper.

Cabal/Distribution/Fields/Pretty.hs
phadej: It's not as invasive change as adding a new constructor.  See https://github.com/haskell/cabal/compare/master...phadej:pretty-post for a PoC

```haskell
> putStr $ showFields' fst snd 2 [PrettyField (["-- comm comm"], id) "foo" (PP.text "value"), PrettyField ([], map ("-- " ++)) "bar" (PP.text "value2") ]
-- comm comm
foo: value
-- bar: value2
```

Cabal/Distribution/PackageDescription/FieldGrammar.hs
phadej: I think `format` is ok name. The `hsSourceDirsGrammar` is also an option. I don't feel strongly about these (I'm thinking of a way for not needing them).

cabal-install/Distribution/Client/Init/FileCreators.hs
phadej: I was thinking more of type like:

```haskell
fieldP
   :: (Pretty b, Newtype b a)
   => FieldName
   -> Flag a
   -> Doc
```

which would be used as

```diff
-   , fieldP "hs-source-dirs"
-       (maybeToFlag (formatHsSourceDirs <$> case buildType of
+   , fieldPAla "hs-source-dirs" formatHsSourceDirs
+       (maybeToFlag (case buildType of
```

phadej: The above decouples "what" from "how it looks", when how it looks doesn't come from the type directly.

phadej:
About *Unspecified/any version range formatting*,

I think we can change how `Dependency` as pretty-printed. Try to modify its `Pretty` instance.

```
cabal run hackage-tests roundtrip packagenameprefix
```

will round roundtrip on Hackage, so if it passes, I'm confident it's ok.

cabal-install/Distribution/Client/Init/FileCreators.hs
m-renaud: Ahh gotcha, yeah that makes sense and is much cleaner.

Cabal/Distribution/PackageDescription/FieldGrammar.hs
m-renaud: Updated comment, going to leave names as is.

Cabal/Distribution/Fields/Pretty.hs
m-renaud: Makes sense, updated the code and introduced a new `FieldAnnotation` type in `FileCreators.hs` to hold the comments and the decision on whether or not to comment out the field. No diffs in goldens (as expected).

phadej:
7.10.3 failure is legit

m-renaud:
Huh, interesting. The failure is on line 278 in the `prependPrefix` helper function in `mainHs` which wasn't touched at all in this PR, even transitively :/ Do you know what's up with that @phadej ?

Also, when trying to build with ghc-7.10.3 locally `cryptohash-sha256-0.11.101.0` fails to build:

<details>
  <summary>Build log</summary>
```
Failed to build cryptohash-sha256-0.11.101.0.
Build log (
/home/mrenaud/.cabal/logs/ghc-7.10.3/cryptohash-sha256-0.11.101.0-541559c14cc76365a19f26bf3fd7c60613e1870c1ce0f5cd3363cf2a9b76d027.log
):
Configuring library for cryptohash-sha256-0.11.101.0..
Warning: Packages using 'cabal-version: >= 1.10' must specify the
'default-language' field for each component (e.g. Haskell98 or Haskell2010).
If a component uses different languages in different modules then list the
other ones in the 'other-languages' field.
Preprocessing library for cryptohash-sha256-0.11.101.0..
Building library for cryptohash-sha256-0.11.101.0..
[1 of 2] Compiling Crypto.Hash.SHA256.FFI ( src/Crypto/Hash/SHA256/FFI.hs, dist/build/Crypto/Hash/SHA256/FFI.o )
/usr/bin/ld: -r and -pie may not be used together
collect2: error: ld returned 1 exit status
cabal: Failed to build cryptohash-sha256-0.11.101.0 (which is required by
test:unit-tests from cabal-install-3.3.0.0, test:solver-quickcheck from
cabal-install-3.3.0.0 and others). See the build log above for details.
```
</details>

m-renaud:
Looks like a poor interaction between OverloadedStrings and ghc 7.10.3.

phadej:
I added changelog and merged (squashed)  as https://github.com/haskell/cabal/commit/10051c3c6d91f6b879b47a742a30664a09f55002

Issue #6700, opened 2020-04-12
Rewrite bootstrap.sh


phadej:
It should have:
- hardcoded lists of dependencies per GHC (and platform)
- bail out for unsupported combinations

These lists should be autogenerated, with help of `cabal-plan` (I have done something close in spirit in https://github.com/phadej/cabal-extras/tree/master/cabal-bundler so I know it's possible).

It should also check the hashes of `.tar.gz` and revision `.cabal` files downloaded (`cabal-bundler` does that).

Why so? Because `bootstrap.sh` is largely not maintained. We test it with single ghc-compiler version. Much better would be if it was "correct by construction".

---

The script should always work with self-created `ghc-pkg init` package database (which is called sandbox in current script). 

---

It should not generate documentation, when there's working `cabal-install`, one could iterate to get it. Same applies for anything extra. `bootstrap.sh` purpose is to get first working version on platform where there are no (older) `cabal-install`. All other use-cases are out-of-scope.

bgamari:
I would imagine it would also be helpful to write it in a real language. This is the sort of thing I usually fall back onto Python for.

bgamari:
I gave this a quick shot in #6719.

phadej:
Continued in #6979

PR #6719, opened 2020-04-19
WIP: Rewrite bootstrap

bgamari:
Given I've complained about this in the past, I felt compelled to pick up #6700. This currently requires a reasonably modern Python (Python 3.5 or newer) but if this is a problem we could likely relax this to Python >= 3.2.

# The idea

A JSON file (generated with `bootstrap.py --extract-plan`) containing dependency and source information is built from `plan.json`. This is then used to fetch and build the bootstrapping dependencies. Since this is only to be used for bootstrapping on new architectures, we advise the users to only use the resulting `cabal-install` executable to build a proper `cabal-install`.

# To-do

* [x] Figure out how to determine revision information; unfortunately this doesn't appear to be available from `plan.json`
* [x] Figure out story for distributing `bootstrap-deps.json`; should it be in the repository?
* [x] Document usage
* [x] Test in CI


cabal-install/bootstrap.py
```diff
@@ -0,0 +1,299 @@
...
+
+    return [from_json(dep) for dep in deps]
+
+def write_bootstrap_deps(deps: List[BootstrapDep]):
+    def to_json(dep: BootstrapDep) -> object:
+        return {
+            'package': dep.package,
+            'version': dep.version,
+            'source': dep.source.value,
+            'revision': dep.revision,
```
phadej, line 121: revision file should have sha256 too.

cabal-install/bootstrap.py
```diff
@@ -0,0 +1,299 @@
...
+import hashlib
+import json
+from pathlib import Path
+import shutil
+import subprocess
+from textwrap import dedent
+from typing import Set, Optional, Dict, List, Tuple, \
+                   NewType, BinaryIO, NamedTuple
+
+ghc_path = Path('ghc')
```
phadej, line 35: One of problems with old bootstrap was that we were somewhat sensitive to the `ghc` version used. We should use `ghc --version` info.

Or better we should parse `ghc --info` and find `ghc-pkg` using `LibDir`, as perfectly boostrapping ghc should be configurable (i.e. we wouldn't need to rely on PATH).

cabal-install/bootstrap.py
```diff
@@ -0,0 +1,299 @@
...
+            source = PackageSource.LOCAL
+        elif unit['type'] == 'configured':
+            source = PackageSource.HACKAGE
+        elif unit['type'] == 'pre-existing':
+            source = PackageSource.PREEXISTING
+
+        return BootstrapDep(package = unit['pkg-name'],
+                            version = unit['pkg-version'],
+                            source = source,
+                            revision = None,
```
phadej, line 216: This is TODO?

bgamari: Indeed. This is the largest issue at the moment. I don't believe this information is accessible from `plan.json.

23Skidoo:
`bootstrap.sh` is also supposed to work when it's run from a Git checkout, I guess that this change means dropping that feature? Overall I'm +1 on making it more robust.

bgamari:
> `bootstrap.sh` is also supposed to work when it's run from a Git checkout, I guess that this change means dropping that feature? 

Not necessarily. We could decide that we want to include a bootstrap dependencies description for a couple of supported bootstrap compilers in the repository. I'll leave this decision to you and @phadej.

cabal-install/bootstrap.py
bgamari: @phadej, how best to go about addressing this?

cabal-install/bootstrap.py
bgamari: > Or better we should parse `ghc --info` and find `ghc-pkg` using `LibDir`, as perfectly boostrapping ghc should be configurable (i.e. we wouldn't need to rely on PATH).

This suggestion seems very fragile to me. The `ghc-pkg` wrapper in `bin/` arranges to pass `--global-package-db` to the `ghc-pkg` executable in `LibDir`. We would need to duplicate this logic if we were to use the `LibDir` executable directly.

Moreover, on some platforms (e.g. NixOS) the executable in `PATH` is actually a distribution-provided wrapper which sets up other state. IMHO we shouldn't start poking around in `LibDir`. If anything we should just look for a `ghc-pkg` in the same directory as we found the `ghc` executable.

cabal-install/bootstrap.py
bgamari: Fair enough. However, I am going to hold off on addressing this until we have a scheme for figuring out the current revision of a package.

cabal-install/bootstrap.py
bgamari: It actually turns out the that the `--global-package-db` flag that I noted above is added by the NixOS wrapper. Nevertheless, the point stands.

cabal-install/bootstrap.py
phadej: The only approach I know (which doesn't require changing `cabal`) is to scan through `01-index.tar` and find the info there. That's how `cabal-bundler` works.

The `cabal` doesn't really know which revision it's working with, as the number is not defining one. We probably know the sha256 of the revision, and I'd rather change hackage-server to provide revisions based on the hash. The revision number is for humans, not the machines.

cabal-install/bootstrap.py
phadej: I'm sad to learn about this.

The simple heuristics don't work:

```
% $(dirname $(which ghc-8.10.1))/ghc-pkg --version
GHC package manager version 8.8.3
```

we have better heuristics for these, but they don't work if suffix is arbitrary. I'd rather not to use heuristics, and for special (and I consider NixOS special) situations users would rather required to be explict.

NixOS probably uses wrapper because `--global-package-db` because they fake global installs, not actually do that. That doesn't affect bootstrap.sh because it doesn't use "extras".

cabal-install/bootstrap.py
phadej: As said below, cleanest approach would be to able directly ask `hackage-server` for `package.cabal` with known sha256.

cabal-install/bootstrap.py
bgamari: Alright, let me know how this looks.

bgamari:
We still need to sort out if/how we want to package the build dependency sets in the tree.

bgamari:
I cannot tell what the meta check is worried about.

cabal-install/bootstrap.py
```diff
@@ -0,0 +1,436 @@
...
+        if path is None:
+            raise ValueError("Couldn't find ghc in PATH")
+        ghc = Compiler(Path(path))
+    else:
+        ghc = Compiler(args.with_compiler)
+
+    print(f'Bootstrapping cabal-install with GHC {ghc.version} at {ghc.ghc_path}...')
+
+    if args.extract_plan:
+        subprocess.run(['cabal', 'v2-configure', '-w', ghc.ghc_path])
```
phadej, line 409: 'v2-build', '--dry'

Configure creates / modifies `cabal.project.local` - that's not great.

Also we probably should use `cabal.project.release` so dev stuff don't accidentally affect buildplans. Or even own project file (if we want to use `Cabal` bundled with GHC).

cabal-install/bootstrap.py
```diff
@@ -0,0 +1,436 @@
...
+
+PackageSpec = Tuple[PackageName, Version]
+
+def find_pkg_revisions(index_tarball: Path,
+                       pkgs: Set[PackageSpec],
+                       ) -> Dict[PackageSpec, Tuple[SHA256Hash, int]]:
+    """
+    Find the current revision number and hash of the most recent revision of
+    the given package name/versions present in an index tarball. Ideally we
+    would request this information from Hackage, but sadly this isn't currently
```
blackgnezdo, line 109: Missing some words, something like "possible, as"

cabal-install/bootstrap.py
```diff
@@ -0,0 +1,436 @@
...
+        subprocess.run([str(ghc.ghc_pkg_path), 'describe', f'{dep.package}-{dep.version}'],
+                       check=True, stdout=subprocess.DEVNULL)
+        print(f'Using {dep.package}-{dep.version} from GHC...')
+        return
+
+    elif dep.source == PackageSource.HACKAGE:
+        assert dep.src_sha256 is not None
+        tarball = fetch_package(dep.package, dep.version, dep.src_sha256,
+                                dep.revision, dep.cabal_sha256)
+        subprocess.run(['tar', 'xf', tarball.resolve()],
```
blackgnezdo, line 231: Please use `zxf` to support non-GNU tar versions.

I confirmed that I can this patched script with ghc 8.10.1 on OpenBSD.

bgamari: Fair point.

bgamari:
Addressed comments.

@phadej, how would you like to handle merging this? Should I drop the `bootstrap-deps.json` files and then we can reintroduce them in the release branch?

bgamari:
Ping @phadej.

blackgnezdo:
@phadej, I'd love to use this for building the OpenBSD port. It would be awesome to merge before 3.4.

blackgnezdo:
Quick observation: after rebasing the `boostrap*json` files have to be regenerated due to new dependencies on `regex-base` and `regex-posix` at HEAD.

phadej:
I'm looking at this. One note immediately, we have to record flag selection.

phadej:
Continued in #6979

PR #6723, opened 2020-04-21
Add cabal init golden test for cabal file with field comments.

m-renaud:


phadej:
Thanks!

tests/fixtures/init/lib-exe-and-test-with-comments-golden.cabal
```diff
@@ -0,0 +1,105 @@
...
+-- The name of the package.
+name:               foo
+
+-- The package version.  See the Haskell package versioning policy (PVP)
+-- for standards guiding when and how versions should be incremented.
+-- https://pvp.haskell.org
+-- PVP summary:      +-+------- breaking API changes
+--                   | | +----- non-breaking API additions
+--                   | | | +--- code changes with no API change
+version:            3.2.1
```
vrom911, line 16: PVP implies 4 components. Maybe in the example, it makes sense to have the most iconic one with the description?
Especially when it could be important for boundaries (for example, `2.0.1.0 > 2.0.1`)

```suggestion
--                   | | | +--- code changes with no API change
--                   | | | | +--- bug fixes, documentation changes with no API change
version:            3.2.1.0
```

phadej: PVP doesn't imply 4 components.

> A package version number SHOULD have the form A.B.C, and MAY optionally have any number of additional components, for example 2.1.0.4 (in this case, A=2, B=1, C=0).

PR #6736, opened 2020-04-24
VersionIntervals don't destroy MajorBoundVersion

phadej:


phadej:
I have no plans to continue working on this.

PR #6740, opened 2020-04-25
Change 'optional-packages' default on missing cabal.project

### LINKED ISSUE #6739: Cabal looks inside subdirectories, but only with later GHC
**Describe the bug**

I have a "top-level" directory that contains `opaleye.cabal`.  The directory contains a subdirectory called `opaleye-sqlite` which contains `opaleye-sqlite.cabal`.  The latter is, by design, not buildable.

If I run `cabal v2-configure --with-ghc=<8.4.4>` then all is well. It builds `opaleye.cabal` as expected.

If I run `cabal v2-configure --with-ghc=<8.6.5>` then it also tries to build `opaleye-sqlite.cabal`.  I don't want it to!  The build fails, as expected.

```
...
[__0] trying: opaleye-0.6.7004.2 (user goal)
[__1] trying: transformers-0.5.6.2/installed-0.5.6.2 (dependency of opaleye)
[__2] trying: contravariant-1.5.2 (dependency of opaleye)
[__3] trying: base-4.12.0.0/installed-4.12.0.0 (dependency of opaleye)
[__4] next goal: opaleye-sqlite (user goal)
[__4] rejecting: opaleye-sqlite-0.0.1.0 (conflict: contravariant==1.5.2,
opaleye-sqlite => contravariant>=1.2 && <1.5)
[__4] skipping: opaleye-sqlite-0.0.0.1, opaleye-sqlite-0.0.0.0 (has the same
characteristics that caused the previous version to fail: excludes
'contravariant' version 1.5.2)
[__4] fail (backjumping, conflict set: contravariant, opaleye-sqlite)
goals I've had most trouble fulfilling: contravariant, transformers, base,
opaleye, opaleye-sqlite
Try running with --minimize-conflict-set to improve the error message.
````

**To Reproduce**

* Clone Opaleye from https://github.com/tomjaguarpaw/haskell-opaleye
* Run `cabal v2-configure --with-ghc=<your GHC 8.4.4>`. It will succeed.
* Run `cabal v2-configure --with-ghc=<your GHC 8.6.5>`. It will fail with the error above.
* `rm opaleye-sqlite/opaleye-sqlite.cabal`
* Run `cabal v2-configure --with-ghc=<your GHC 8.6.5>`. It will succeed, because it no longer tries to build `opaleye-sqlite/opaleye-sqlite.cabal`.

**Expected behavior**

I expected that when using GHC 8.6.5 it only try to build `opaleye.cabal` (like it did with GHC 8.4.4) not `opaleye-sqlite/opaleye-sqlite.cabal` too.

It seems to be looking into subdirectories, and I'm not sure why. I'm also not sure why this depends on GHC version!

**System information**
Debian Linux 10.3, `cabal` 3.2.0.0 via ghcp, `ghc` 8.4.4 from Debian, `ghc` 8.6.5 from ghcup

hasufell:
Fixes #6739

### DIFF
--- a/Cabal/doc/cabal-project.rst
+++ b/Cabal/doc/cabal-project.rst
@@ -67,7 +67,7 @@ project are:
 .. cfg-field:: optional-packages: package location list (space or comma-separated)
     :synopsis: Optional project packages.
 
-    :default: ``./*/*.cabal``
+    :default: empty
 
     Like :cfg-field:`packages`, specifies a list of package locations
     containing local packages to be built. Unlike :cfg-field:`packages`,
@@ -90,13 +90,14 @@ project are:
 
 All local packages are *vendored*, in the sense that if other packages
 (including external ones from Hackage) depend on a package with the name
-of a local package, the local package is preferentially used.  This
-motivates the default settings::
+of a local package, the local package is preferentially used.
+For subdirectories to be considered local packages, the following setting
+can be used::
 
     packages: ./*.cabal
     optional-packages: ./*/*.cabal
 
-...any package can be vendored simply by making a checkout in the
+...then any package can be vendored simply by making a checkout in the
 top-level project directory, as might be seen in this hypothetical
 directory layout::
 
--- a/cabal-install/Distribution/Client/ProjectConfig.hs
+++ b/cabal-install/Distribution/Client/ProjectConfig.hs
@@ -544,9 +544,6 @@ readProjectLocalConfigOrDefault verbosity distDirLayout = do
         -- We expect a package in the current directory.
         projectPackages         = [ "./*.cabal" ],
 
-        -- This is to automatically pick up deps that we unpack locally.
-        projectPackagesOptional = [ "./*/*.cabal" ],
-
         projectConfigProvenance = Set.singleton Implicit
       }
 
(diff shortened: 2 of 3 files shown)


phadej:
Thanks.

I’m not sure whether the behavior we are changing is documented (doesn’t seem so, could you double check). This definitely requires changelog entry. Add a file to `changelog.d`, see other files for examples.

phadej:
I’d remove the field completely. Commit messages tell history, let’s not leave/repeat it in the source :)

hasufell:
Apparently there are multiple references to this default: https://github.com/haskell/cabal/blob/master/Cabal/doc/cabal-project.rst

tomjaguarpaw:
Thanks for responding to my confusion.  I agree that this API is better.  On the other hand, I have been bitten by cabal changing its public API a few times in the past and for the sake of others I would be cautious about this going in without a very obvious warning/deprecation cycle.

phadej:
@tomjaguarpaw We don't have resources to manage warnings&deprecations. Either stuff is changed, or it's not.

phadej:
Specifically, when people upgrade from ancient, say `cabal-install-2.0` to `cabal-install-3.0` they do miss deprecation warnings etc. I repeat, there is just no resources to manage warnings & deprecations with 2-3 year cycles, so people have chance to notice them. People do not upgrade `cabal-install` eagerly, so the 2-3 years is justified so they actually could see those warnings.

tomjaguarpaw:
As you wish. All in all, I do believe the behaviour in this PR is the better one.

PR #6741, opened 2020-04-25
Remove "-any" from AnyVersion pretty-print output.

m-renaud:
## Behaviour Change

**Old behaviour:**

```
λ> pretty AnyVersion
-any
```

**New behaviour:**

```
λ> pretty AnyVersion

```

## Testing

```
cabal run hackage-tests roundtrip
```

There is one failure in the `Workflow` package version 0.7.0.0 which uses the `-any` syntax (see [Workflow.cabal](https://hackage.haskell.org/package/Workflow-0.7.0.0/src/Workflow.cabal)) in addition to a version bound check: `TCache -any && <1.0`. Is this even valid? It would be equivalent to `TCache <1.0`.

```
TCache -any && <1.0
```

```
TCache && <1.0
       ^ Parse error
```

How would we go about fixing this @phadej ?

Also, let me know if there's other places you'd like to see updated, I just updates the `pretty` function as suggested in the other PR.

### TODO

- Fix lib-tests
- Update changelog
- Update docs

---

### DIFF
--- a/Cabal/Distribution/Types/VersionRange/Internal.hs
+++ b/Cabal/Distribution/Types/VersionRange/Internal.hs
@@ -235,7 +235,7 @@ hyloVersionRange f g = h where h = f . fmap h . g
 instance Pretty VersionRange where
     pretty = fst . cataVersionRange alg
       where
-        alg AnyVersionF                     = (Disp.text "-any", 0 :: Int)
+        alg AnyVersionF                     = (Disp.empty, 0 :: Int)
         alg (ThisVersionF v)                = (Disp.text "==" <<>> pretty v, 0)
         alg (LaterVersionF v)               = (Disp.char '>'  <<>> pretty v, 0)
         alg (OrLaterVersionF v)             = (Disp.text ">=" <<>> pretty v, 0)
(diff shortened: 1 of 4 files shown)


m-renaud:
Alright, @phadej as expected the round-trip property tests fail since `print . parse "-any"` does not equal "`-any"`. Looking at the parsing code its not clear exactly how a lack of version range gets turned into the `AnyVersion` constructor, do you know how that happens? Is there a different parser for the version range other than what is in [`Cabal/Distribution/Types/VersionRange/Internal.hsCabal/Distribution/Types/VersionRange/Internal.hs`](https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Types/VersionRange/Internal.hs#L331) ? The only place `AnyVersion` is constructed in the parser is [here](https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Types/VersionRange/Internal.hs#L357) :confused:

phadej:
See #6742

Cabal/Distribution/Types/VersionRange/Internal.hs
```diff
@@ -235,7 +235,7 @@ hyloVersionRange f g = h where h = f . fmap h . g
 instance Pretty VersionRange where
     pretty = fst . cataVersionRange alg
       where
-        alg AnyVersionF                     = (Disp.text "-any", 0 :: Int)
+        alg AnyVersionF                     = (Disp.empty, 0 :: Int)
```
phadej, line 238: This is wrong. The `Dependency` output have to be modified.

See what's done in #6742. That PR is not complete, I run out of time on Friday, so didn't put it up. Sorry for that. I changed the `Dependency` output as I was updating `parser-tests` outputs anyway.

PR #6743, opened 2020-04-29
Split up CmdInstall (v2-install) monolithic module.

m-renaud:
1/n pre-factorings to split out `cabal install --lib` into its own command.

Adds the following modules:
- CmdInstall.TargetProblem
- CmdInstall.Utils
- Env.Install
- Env.Utils

This first PR does not attempt to make any structural or code organization changes, it simply splits out functions into logical groups. Future PRs will introduce more separation between the exe and lib install paths, and centralize as much as the environment specific logic into `Distribution.Client.CmdEnv...`.

---


typedrat:
As the original author of a lot of the former code, I'll try and give this a review soon.

phadej:
I'd pause on this.

Could you refactor `TargetProblem` into `Distribution.Client.TargetProblem`. There is 

```
cabal-install/Distribution/Client/CmdBench.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdBuild.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdHaddock.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdInstall.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdRepl.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdRun.hs:data TargetProblem =
cabal-install/Distribution/Client/CmdSdist.hs:data TargetProblem = AllComponentsOnly ComponentKind
cabal-install/Distribution/Client/CmdTest.hs:data TargetProblem =
```

and I suspect they are all very similar. That would be more atomic first (and very great!) step.

cabal-install/Distribution/Client/Env/Utils.hs
```diff
@@ -0,0 +1,53 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+module Distribution.Client.Env.Utils
```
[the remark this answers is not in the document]
m-renaud: I was originally thinking that there would be some "environment" related utils shared between CmdEnv and CmdInstall but taking a closer look at the code it appears it can be refactored such that it doesn't need them. Renamed/moved everything into the an `CmdEnv` module group.

m-renaud:
Ack on `TargetProblem`, since this is a bit more involved change and I'd like to start more substantial refactorings of the install code in parallel I'll do that in a follow-up PR.

cabal-install/Distribution/Client/CmdInstall/Utils.hs
```diff
@@ -0,0 +1,41 @@
...
+import Distribution.Client.ProjectOrchestration
+       ( CannotPruneDependencies(..), ComponentName(..)
+       , ComponentTarget(..) , ProjectBuildContext(..) )
+import Distribution.Simple.Utils
+       ( die', warn )
+import Distribution.Verbosity
+       ( Verbosity )
+
+-- | Emits a warning if the build context does not contain any executables.
+warnIfNoExes :: Verbosity -> ProjectBuildContext -> IO ()
```
phadej, line 22: I think this should be just in `CmdInstall` implementation. It's used only there.

cabal-install/Distribution/Client/CmdInstall/Utils.hs
```diff
@@ -0,0 +1,41 @@
...
+  where
+    targets    = concat $ Map.elems $ targetsMap buildCtx
+    components = fst <$> targets
+    selectors  = concatMap snd targets
+    noExes     = null $ catMaybes $ exeMaybe <$> components
+
+    exeMaybe (ComponentTarget (CExeName exe) _) = Just exe
+    exeMaybe _                                  = Nothing
+
+reportCannotPruneDependencies :: Verbosity -> CannotPruneDependencies -> IO a
```
phadej, line 39: On the other hand, this seems to be used in both `CmdBuild` and `CmdInstall` so having it in the module with `CmdInstall` in the name doesn't feel right.

cabal-install/Distribution/Client/CmdInstall/Utils.hs
m-renaud: See comment See top level review comment.

m-renaud: TL;DR; a lot of the other "implementation detail" code will be moving to this module, but I also realize this module should likely be named `Internal` instead of `Utils`.

cabal-install/Distribution/Client/CmdInstall/Utils.hs
phadej: Yes, that's fine. Please don't create `.Utils` modules though. They are *impossible* to get rid off.

m-renaud:
@phadej I hear you on the Utils, what I have in mind is moving the majority of the logic and helper functions out of the top level `Distribution.Client.CmdInstall` cmd so it just contains the top level logic, and delegates to some internal module. In other words, `Utils` here should probably be renamed `Internal`, which will eventually host most of the details that are currently in the top level `CmdInstall` module.

> We have CmdBuild + CmdInstall importing one another

`CmdInstall` importing `CmdEnv` is only temporary, currently the logic for the two is intermingled but it doesn't have to be, but splitting it requires a bit more refactoring. I could have sent one large PR which makes all the change I have in mind, but I would strongly prefer keeping file moves and simply extracting code used once into a function separate from changes that restructure the code. I would rather send several smaller PRs because they're easier to reason about while writing them (which is especially useful when working on something part time since you need to keep less state in your head for any given change), easier to review (clearly separates renames and code cleanup from true logic changes), and if there happens to be an issue it is easier to pinpoint which commit caused it.

So, don't take the current structure in this PR as the final structure that this will be in, this is just the first of several (probably 4 or 5?) PRs to decompose the logic. It's possible that after everything is split apart it becomes apparent that some particular utility should be moved to another place and deduped, but that can be done afterwards. Once this PR is in, there are about 3 other things that can happen in parallel, so I'm happy to implement any suggestions you have but if they can easily be done as follow-up PRs to unblock further work I'd request that you allow that. Now, a valid concern is that you may be worried that the code will be left in a half-finished state but I can assure you (to as much certainty that I have) that I'll actually make the follow up changes described above :)

cabal-install/Distribution/Client/CmdInstall/Utils.hs
m-renaud: I don't know if I'd say _impossible_ ;) In any case renamed both of them to `Internal`.

phadej:
`Internal` is not any better then `Utils.` Please, if you have some concrete `CmdInstall.Functionality` module in mind do that. Let's no go through **completely arbitrary** `Internal` or `Utils` module.

Either you outline the plan you have in mind, or make "good standalone steps". And I'd prefer latter.

phadej:
Moving code to then move it again is pointless. That doesn't add any value.

I'd suggest you tackle `TargetProblem` next.

m-renaud:
> Moving code to then move it again is pointless

I'm really not sure what you're referring to here, the only thing that _may_ be moved later is the duplication between `CmdInstall` and `CmdBuild` that you mentioned, but that duplication already exists. I made a comment that there _may_ be better places for things at the end, but that it doesn't make sense to not do anything until the absolute ideal module structuring is determined.

> or make "good standalone steps",

I'm confused, I described above what the standalone steps were above. Most of the "executable install" specific logic will move into `CmdInstall.Internal`, `CmdInstall` will become the high level logic for the command and will mostly call into `CmdInstall.Internal`. `CmdEnv.Internal` will hold the logic for manipulating the environment (currently `installLibraries`), `CmdEnv` will eventually be the command entrypoint for the `env` command. The logic in the current `CmdInstall` also needs to be split into smaller functions in `CmdInstall.Internal`.

Simply closing the PR feels very handed, there is a real user bug that is repeatedly being encountered and I'm trying to make progress on `env` since you're opposed to fixing `cabal install` in place. `TargetProblem` is completely irrelevant, I'm all for cleaning up the code but progress **needs** to be made towards env and fixing the lib install issue.

phadej:
As I said, I don't like `CmdInstall.Internal` naming and aproach.

If you don't like to do `TargetProblem` refactoring next, then maybe `CmdInstall.CommandUI` would be good one. the `installCommand` is used by `v2-test` and `v2-bench` at least. Maybe `NixStyleCommandUI` would be even better, and then all above would specialize that one.

There are ways to make structure objectively better. This PR wasn't.

m-renaud:
> As I said, I don't like CmdInstall.Internal naming and aproach.

Sure, I suggested `Internal` since its the pattern used throughout the rest of the Cabal codebase, its easy to rename it to something different in the same PR if that's not what you wanted to see.

> then maybe CmdInstall.CommandUI would be good one. the installCommand is used by v2-test and v2-bench

I hadn't planned to move `installCommand` at this point in time since it still doesn't get us any closer to the `env` command, its the internals of `installCommand` that need to get refactored, so I'm not sure what moving this to `CmdInstall.CommandUI` would buy us.

> There are ways to make structure objectively better

The suggestions above don't get at the heart of the problem is the internal implementation details of `installCommand` need to be deconstructed so that the logic needed for `env` can be extracted. I was trying to do it in smaller steps to separate the boring moving of code from the actual deconstruction, but I guess I'll just send a larger PR has the finished product if that's what you'd prefer.

phadej:
>  so I'm not sure what moving this to CmdInstall.CommandUI would buy us.

Making the `cabal-install` codebase less spaghetti.

I'm sorry, but I'm the one who require a bit more than is strictly necessary to implement a feature.

m-renaud:
> Making the cabal-install codebase less spaghetti.

Yes, I agree that making the codebase less spaghetti is very desirable, but that doesn't buy us anything towards splitting out the env code, and now introduces an ordering dependency since the PR moving `installCommand` would need to land before making changes to the internals unless you want to deal with integrating merge conflicts (which I don't because they're a pain). I'd prefer to structure it in such a way that the individual changes can be made independently.

Also, unless you're referring to `Client.installCommand` I don't see a reference to `CmdInstall.installCommand` in `v2-test` or `v2-bench`, and if you are that's even further departed from `CmdInstall` refactoring.

> I'm sorry, but I'm the one who require a bit more than is strictly necessary to implement a feature.

This is beyond "more than is strictly necessary", it's a revamp of how the various commands are structured across the cabal codebase, if we're going to block other work on this there should be a justification on why that has to happen first and slow other changes down. I disagree that this is the yak that needs shaving right now.

I have neither unlimited time or mental energy to spend on this, and it really is not clear to me how the changes you suggest help clean up CmdInstall internals and move us towards an `env` command (or at least why they need to be done first). This is far from my first time refactoring code, I'm not asking you to blindly stamp everything, I'm just asking that you have some trust that I have a vision in mind to put the codebase into a better state then the code is in now. There are obviously different orderings for all of these things, but I'm trying to optimize for independent changes, my time, and fixing the `cabal install --lib` bug.

Issue #6445, opened 2019-12-18
Remove sandbox functionality


phadej:
Related is e.g. https://github.com/haskell/cabal/issues/6444#issuecomment-567002066

Maybe it's still a bit too early to remove `v1-build`, but sandboxes could IMHO be removed. If someone really needs them, they can be simulated.

Yet we need to do some cleanup, as v2-build is the default now.

CMCDragonkai:
Just a question. In my Gitlab CI/CD I used `.cabal-sandbox` as the CI cache. Now if sandboxes are removed or I switch over to V2 commands. What is the appropriate thing to cache to speed up CI builds? Is it `.cabal/store`?

phadej:
@CMCDragonkai yes. And `.cabal/packages`, when you clean a bit from there. See e.g. https://github.com/haskell-CI/haskell-ci/blob/f3e046c5ba5ba252ae44f82b2ecfb1e2d8760c0a/.travis.yml#L17-L31

Mistuke:
> Maybe it's still a bit too early to remove v1-build, but sandboxes could IMHO be removed. If someone really needs them, they can be simulated.

How? :) I still use them to reproduce GHC issues people submit so I don't mess up my global config but genuinely unsure how to simulate them. The biggest thing I use them for is to have a `package-db` with all the dependencies of the repro the user submitted.

Since I often repro with ghc development builds I have to run with `--allow-newer` so I don't want these packages in my global store-dir as as soon as the ghc version changes I gotta rebuild again anyway (and the space isn't cleaned up automatically).

But new-install 's dist folder only contains a package-db with the package only, not it's dependencies.. so I end up having to use `--store-dir` to point it locally and then use two `package-dbs` where sandbox just has one so it's easier to use.

hasufell:
Sandboxes are still used by some users and are very robust. Recovering from problems/failure is easy, while with v2-build, it basically boils down to removing the entire store. Semi-reproducible bugs like https://github.com/haskell/cabal/issues/6483 are a good example

phadej:
My opinion is: people who need sandboxes could keep using old `cabal-install` versions. The less code there is blocking new development work, the better.

hasufell:
> The less code there is blocking new development work, the better.

I think that's a little quick. In case the code is really a problem, there should at least be a call for maintainers picking that up. We should know clearly whether there is no interest in this feature or not.

Mistuke:
[...]

PR #6747, opened 2020-05-01
Remove sandbox

phadej:


m-renaud:
One thought: there likely exists some users who still require sandbox functionality, if we remove support in future versions of cabal should we have a process for backporting bugfixes to previous versions of cabal that still support sandboxes? Or are we just saying: if you want bugfixes and new features, you need to figure out how to migrate off of the sandbox functionality? If its the later, is there a guide we can point to for how to get off of sandboxes?

hvr:
> One thought: there likely exists some users who still require sandbox functionality, if we remove support in future versions of cabal should we have a process for backporting bugfixes to previous versions of cabal that still support sandboxes?

I don't think that's realistic given the current available resources. Our support model is currently to have people install the most recent stable release of cabal and we try hard to support a wide range of GHC releases with the most recent releases. Even backporting a fix to previous releases of cabal requires  a non-neglibile amount of cognitive overhead which we just don't have the luxury for right now.

> Or are we just saying: if you want bugfixes and new features, you need to figure out how to migrate off of the sandbox functionality?

Yeah, as frustrating as this may come across to the affected parties, that's the state of affairs. And we do want to make it as easy as reasonable to migrate away from this legacy feature...

> If its the later, is there a guide we can point to for how to get off of sandboxes?

...I'm not aware of such a guide (maybe there exists one already though). Are you interested to write such a migration guide (I guess it'd be a small chapter/section in the cabal user guide)? In its simplest form it's just describing the basic legacy `sandbox` workflow, and how to map the previously used `cabal sandbox` invocations to the new respective `v2-build` mechanisms...

m-renaud:
Yeah, that all makes sense and sounds reasonable. I guess one thing that could be useful to do is put out a pre-announcement on the main channels (mailing list, reddit, twitter, etc) to see if anyone is critically relying on this behaviour, and if there's enough response we could invest some time in putting together a migration guide, otherwise it may not be worth the time if there's only a handful of users.

jneira:
I am happy using v2 commands but i think this feature had been useful for lot of users for years so i agree with @m-renaud in it deserves a careful announcement in all possible channels.

phadej:
I'm locking discussion on this PR. Discuss issue on https://github.com/haskell/cabal/issues/6445

emilypi:
🎉

PR #6754, opened 2020-05-04
Update docs to build with Python 3 and Sphinx 3

bubba:
As per suggestion in https://github.com/haskell/cabal/pull/6668
Also update documentation to mention the virtualenv now (maybe a requirements.txt should be created as well to pin the version of sphinx?)
This will only work assuming that the python used to build `virtualenv` was >3. Will this be a problem for people?
Also not sure if anything needs to be done to the CI/readthedocs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,4 +4,6 @@ sphinx:
   configuration: Cabal/doc/conf.py
 
 python:
-  version: 2.7
+  version: 3.7
+  install:
+    - requirements: Cabal/doc/requirements.txt
--- a/Cabal/Makefile
+++ b/Cabal/Makefile
@@ -53,16 +53,17 @@ $(HADDOCK_STAMP) : $(CONFIG_STAMP) $(BUILD_STAMP)
 
 SPHINX_HTML_OUTDIR=dist/doc/users-guide
 
+# do pip install everytime so we have up to date requirements when we build
 users-guide: .python-sphinx-virtualenv $(USERGUIDE_STAMP)
 $(USERGUIDE_STAMP) : doc/*.rst
 	mkdir -p $(SPHINX_HTML_OUTDIR)
-	(. ./.python-sphinx-virtualenv/bin/activate && $(SPHINXCMD) doc $(SPHINX_HTML_OUTDIR))
+	(. ./.python-sphinx-virtualenv/bin/activate && pip install -r doc/requirements.txt && $(SPHINXCMD) doc $(SPHINX_HTML_OUTDIR))
 
 docs: haddock users-guide
 
 .python-sphinx-virtualenv:
-	virtualenv .python-sphinx-virtualenv
-	(. ./.python-sphinx-virtualenv/bin/activate && pip install sphinx && pip install sphinx_rtd_theme)
+	python3 -m venv .python-sphinx-virtualenv
+	(. ./.python-sphinx-virtualenv/bin/activate)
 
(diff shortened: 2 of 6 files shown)


m-renaud:
Could you set up a temporary ReadTheDocs project which is built from your branch to verify that this builds (and renders) as expected? Specifically, we should verify that the updates `cabaldomain.py` work correctly (I don't know why they wouldn't, but I also don't write much Python).

bubba:
I set one up [here](https://readthedocs.org/projects/cabal-luke/builds/10993888/) and it turned out I needed to specify the sphinx version used, so I created a requirements.txt file. This should work with the virtualenv in the makefile as well!

m-renaud:
So sorry for the incredibly delayed response here. Cloning your repo and running `make users-guide` succeeded for me locally (although with some warnings, which may just be due to an old CL your based on though).

Since this has been open for a month and there haven't been any objections I'm in favour of merging this, could you squash down to one commit?

bubba:
Not at all, I don't think anyone is desperately waiting on this PR :)
Squashed the commits, but just to double check the warnings you're getting are these ones right?
```
/Users/luke/Source/cabal2/Cabal/doc/setup-commands.rst:891: WARNING: unknown option: --allow-newer
/Users/luke/Source/cabal2/Cabal/doc/setup-commands.rst:891: WARNING: unknown option: --allow-older
/Users/luke/Source/cabal2/Cabal/doc/setup-commands.rst:928: WARNING: unknown option: --allow-newer
/Users/luke/Source/cabal2/Cabal/doc/setup-commands.rst:946: WARNING: unknown option: --allow-newer
/Users/luke/Source/cabal2/Cabal/doc/setup-commands.rst:1001: WARNING: cabal:pkg-field reference target not found: setup-depends
```

m-renaud:
Yup, those are the ones. I wonder if it's because Sphinx 3 is more strict about unresolvable symbols? Do you think you might have time to look into fixing the warnings?

bubba:
Yeah they were bugging me too, I'll take a quick peek

bubba:
A lot of those warnings are from the setup-command options not supporting the syntax used in setup-commands.rst:
```
.. option:: --allow-newer[=pkgs], --allow-older[=pkgs]
```
`option` from the sphinx standard domain seems to only parse things like `--allow-newer <pkgs>`

bubba:
I think these are just warnings being picked up by the new version of Sphinx. I've created a PR upstream for it https://github.com/sphinx-doc/sphinx/pull/7770 to help address some of the warnings, but I don't think they're anything to worry about

bubba:
Ping @phadej @DanielG

bubba:
Preview is here https://cabal-luke.readthedocs.io/en/latest/

jhrcek:
@bubba thanks for this PR.
I tried locally and it works well for me. It would be nice to have this merged, because python2 no longer ships on the latest of fedora workstation which I'm using and I don't want to install it just to build these docs.

There's a bunch of warnings about " reference target not found", but I'd be happy to fix those should we decide to merge this.

Is there any obstacle to merging this?

bubba:
@jhrcek I think those warnings were there previously, or maybe the new version of sphinx is just picking them up. But yeah they don't seem to be too major, so I'm happy for this to go in if others are!

phadej:
Merged as part of https://github.com/haskell/cabal/pull/6954

Cleaned up the sphinx warnings there.

PR #6759, opened 2020-05-06
Add shrinker, so writing big non-generic product shrinkers is easier

phadej:
This is suboptimal, as applicative stuff have to in order. But we suffer from that in `arbitrary` already.


phadej:
8.4.4 failure is legit and doesn't seem to be related to #6760 I'll check when fixing #6760 though.

PR #6768, opened 2020-05-08
Resolve #5570: Use PackageVersionConstraint more

### LINKED ISSUE #5570: Dependency as constraint vs Dependency as dependency
Just opening this to keep track of how Dependency is used throughout the codebase for when we'll split it

See https://github.com/fgaz/gsoc/blob/master/2018/final-report.md#the-dependency-as-constraint-problem

* `Dependency` as constraint
  * `configConstraints`
  * `Client/Outdated` and `packageConstraintToDependency`
  * maaaybe the two `lookupDependency`s
* `Dependency` as dependency on library
  * `build-depends` obviously
  * `setup-depends`
  * `DependencyMap` and similar
* `Dependency` as "dependency" (more like... package source? hint? it's not even used..) on package
  * `ProjectConfig.projectPackagesNamed`

Other things to remove when doing the refactor:
* `Types/Dependency.hs`:
  * `thisPackageVersion`
  * `notThisPackageVersion`

phadej:


Cabal/Distribution/Types/PackageVersionConstraint.hs
```diff
@@ -21,8 +22,8 @@ import           Text.PrettyPrint                ((<+>))
 -- | A version constraint on a package. Different from 'ExeDependency' and
 -- 'Dependency' since it does not specify the need for a component, not even
 -- the main library.
--- There are a few places in the codebase where 'Dependency' is used where
--- 'PackageVersionConstraint' should be used instead (#5570).
+-- There are a few places in the codebase where 'Dependency' was used where
+-- 'PackageVersionConstraint' is not used instead (#5570).
```
fgaz, line 26: Isn't this sentence a contradiction now?

PR #6774, opened 2020-05-10
Refactor shared TargetProblem data types into their own module.

m-renaud:
This adds a new 'ExtensibleTargetProblem e' type which can be used uniformly
across the commands, supporting the most common use case of errors being one of:
'common', 'none enabled', or 'no targets'. In cases where custom target errors
are needed, the base set can be expanded with a custom problem type, and custom
reporting/rendering is supported.

---


cabal-install/Distribution/Client/TargetProblem.hs
```diff
@@ -0,0 +1,83 @@
...
+
+-- | Type alias for a 'TargetProblem' with no user-defined problems/errors.
+--
+-- Can use the utilities below for reporting/rendering problems.
+type TargetProblem = ExtensibleTargetProblem ()
+
+-- | The various error conditions that can occur when matching a
+-- 'TargetSelector' against 'AvailableTarget's which can be extended
+-- with command specific target problems as described by 'e'.
+data ExtensibleTargetProblem e
```
phadej, line 46: Can `TargetProblemCommon` and `ExtensibleTargetProblem` be merged into own?

m-renaud: That would be doable, would just require some more code shuffling. One question would be if we want to inline all of the cases of `TargetProblemCommon` into the top level `ExtensibleTargetProblem` or if we want to keep the hierarchy as is.

At the least we could move the `TargetProblemCommon` definition into `TargetProblem.hs`, I'd prefer to leave the change of flattening the hierarchy (if that's what we decide to do) to a follow-up PR to keep this one as simply extracting the repeated code into one location.

phadej: Please shuffle code.

- `TargetProblemCommon` is a good name
- `ExtensibleTargetProblem` is bad name
    - `ExtensibleTargetProblemNoTarget` is also bad. First we rename it from `TargetProblemNoTarget` to then rename back to `TargetProblemNoTarget` again.

---

This commit is good start. But I won't merge it. Please at least try `TargetProblemCommon` extraction as well.

m-renaud: See most recent commit, to summarize:
- Rename `ExtensibleTargetProblem` -> `TargetProblem` (and change alias)
- Shorten data constructor names in `TargetProblem`
- Extracts `TargetProblemCommon` into `TargetProblem.hs`

PTAL!

cabal-install/Distribution/Client/TargetProblem.hs
```diff
@@ -0,0 +1,87 @@
...
+  ( UnqualComponentName,
+  )
+import Distribution.Simple.LocalBuildInfo
+  ( ComponentName(..),
+  )
+import Prelude ()
+
+-- | Target problems that occur during project orchestration.
+data TargetProblemCommon
+   = TargetNotInProject                   PackageName
```
phadej, line 32: 3 spaces,  cabal is truly diverse!

m-renaud: Haha! I really wish we could just run a formatter over the whole code base (I don't care which one!) so that I can stop worrying about indentation.

m-renaud:
Want me to squash commits in the meantime or will you handle that when you make the tweaks and merge later?

phadej:
merged as #6821. Thanks!

PR #6779, opened 2020-05-11
Add weeder

phadej:
```
make weeder
```

to run weeder.

There are plenty of false positives. But still, useful. Thanks @ocharles for the tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,11 @@ validate-via-docker-8.10.1:
 validate-via-docker-old:
 	docker build -t cabal-validate -f .docker/validate-old.dockerfile .
 
+# Weeder
+weeder :
+	cabal build all --project-file=cabal.project.weeder
+	weeder | less
+
 # tags
 .PHONY : tags
 tags :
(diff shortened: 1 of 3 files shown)


ocharles:
How exciting! If you didn't know, you can add those false positives to your Weeder config. I'd also be interested to know which of those are not truly dead, as those would be bugs in Weeder. Sounds like I can grab this branch and have a play myself to get that, though. Maybe a ZuriHac project!

phadej:
Some stuff is used in tests, or is there just for completeness.

ocharles:
Weeder should pick all test uses up as test suites produce a Main.main root (but you need to build tests with hie files, of course)

phadej:
Can one list everything on [`hackage-security`](https://hackage.haskell.org/package/hackage-security) as root? (It's a local library, as it depends on `Cabal`, so it's is built).

ocharles:
Currently a root is a regular expression matching symbol names, and that only mentions module names and variable names at the moment. Perhaps everything there is under a common module prefix? E.g., would `^Hackage.Security..*` work?

Issue #6281, opened 2019-10-02
Parse error when trying to use multiple public libraries within mixins


chshersh:
**Describe the bug**

I have the following lines in my `.cabal` file:

```cabal
  build-depends:       containers-backpack:{ int-strict, ordered-strict, unordered-strict, contrib }

  mixins:              containers-backpack:contrib (Map.Contrib.Group as Map.Contrib.Group.Int)
                                          requires (Map as Map.Int)
```

Here `containes-backpack` is a package with the public library `contrib` inside it. However, when trying to build this package with `cabal-install` I see the following error:

```
Errors encountered when parsing cabal file ./containers-backpack.cabal:

containers-backpack.cabal:142:43: error:
unexpected ':'
expecting "-", white space, "(", "hiding", "requires", comma or end of input

  140 |   build-depends:       containers-backpack:{ int-strict, ordered-strict, unordered-strict, contrib }
  141 | 
  142 |   mixins:              containers-backpack:contrib (Map.Contrib.Group as Map.Contrib.Group.Int)
      |                                           ^
```

**To Reproduce**

You can find full code here:
* https://github.com/kowainik/containers-backpack/pull/46

1. Clone the repo.
2. Checkout to branch from linked PR.
3. Do `cabal v2-build`

**Expected behavior**

No parse error.

**System informataion**
 - Ubuntu-18.04
 - `cabal-3.0.0.0`
- `ghc-8.6.5` or `ghc-8.8.1`

fgaz:
Might be an instance of #5846.

chshersh:
@fgaz Thanks for pointing to this issue! #5846 is another problem I have. For some reason the following doesn't parse:

```cabal
build-depends: containers-backpack:sig
             , QuickCheck
```

But when I write dependencies in a single line, it parses:

```cabal
build-depends: containers-backpack:sig, QuickCheck
```

> Adding `-any` also parses

Which is for now only a minor UI inconvenience.

fgaz:
Weird… I wonder what's wrong with the parsing code. I wasn't able to find the issue last time I tried to debug this. If you want to take a look at it yourself it's in Cabal/Distribution/Types/Dependency.hs

phadej:
@fgaz this is missing functionality. `Mixin` (**not** `Dependency`) is atm defined as 

```haskell
data Mixin = Mixin { mixinPackageName :: PackageName
                   , mixinIncludeRenaming :: IncludeRenaming }
```

There are no sublibs.

PR #6800, opened 2020-05-14
Described Mixin

phadej:
Related to #6281

Screenshot from 2020-05-14 20-29-27

(black dots are where space is required).


phadej:
I think

```
mixins: pkg(Module)requires(Module2)
```

is accepted by the parser but not precisely `Described`. The whitespace is tricky when we have infix keywords *and* empty productions. E.g.

```
mixins: pkg requires(Module2 as Module3)
mixins: pkg requires hiding{Module4}
```

Issue #6083, opened 2019-06-17
Cabal 3: in build-depends, package name gets treated as internal library name ("legacy" syntax) even when using the pkg:lib qualified syntax


ekmett:
I had 3 libraries:

```
fontconfig -- provides fontconfig bindings
fontconfig-freetype -- provides supplementary support for freetype
freetype -- provides freetype bindings
```

I've consolidated the first two into one using public sublibraries. I had to rename the new sublibrary for `fontconfig-freetype` to `fcfreetype` to hack it into place. 

Now I'd like to be able to use the unmangled name, so that users would incur a dependency on `fontconfig:freetype` or `fontconfig:{freetype}`, and all that really changes is a character or so in their imports, but I'd be able to maintain a single package on hackage. While users who do not want or need `freetype` can depend on `fontconfig:core` and bypass the extras.

Done right this sort of thing seems like a path to having many many little fine-grained internal dependencies, so you can incur dependencies on just the functionality you need out of a larger package.

But it seems no matter how I try to refer to the external library `freetype`, I'm getting a cycle detected as if the top level `fontconfig` library which re-exports its own sub-library `freetype`s contents is in a cycle with its member.

I'm guessing either I have no way to refer to the external `freetype` in the current naming convention, or something is one step too simple in the cycle detector? What is the correct name to use to refer to a library that isn't the one in the current package being built, but rather the name in the top level namespace?

See this [associated commit](https://github.com/ekmett/codex/commit/a53cdec2974d5a0e374accb43432ab8c1c96f3e3
) to a side-branch for this issue.

The repository should compile with ghc 8.8 and cabal 3. I apologize for not distilling a smaller issue, but basically any example of library name shadowing should do.

This arises for me at least, because I also have about a dozen more of these, each of which would need distinct name mangling conventions.(e.g. `harfbuzz-freetype`, `harfbuzz-icu`) and it seems a shame to make the user use `harfbuzz:hb-ft` or something rather than remain as clear as the current naming.

fgaz:
That's the legacy syntax :-/

Internal library names shadow external package names.

I was planning to leave it for another .cabal version or so, but I guess we could just disable it now since conflicts like this are likely to happen

/cc @hvr

fgaz:
Btw you may want to add visibility:public to those library stanzas. The Cabal shipped with ghc 8.8 alpha does not enforce it, but from rc1 onwards it will

hvr:
@fgaz what do you consider "legacy" syntax?

[...]

PR #6802, opened 2020-05-14
Resolve #6409: Add public multilib test(s)

### LINKED ISSUE #6409: Add positive test for multilibs.
As we have GHC-8.8 on travis now, there shouldn't be blockers.

phadej:
Also add a test for current behaviour
of https://github.com/haskell/cabal/issues/6083

And variant with mixin, for https://github.com/haskell/cabal/issues/6281

The tests are disable for GHC older than 8.8
Should they work? (cc @fgaz)


fgaz:
They should not, at least until #6047 is in

PR #6822, opened 2020-05-19
Update GHC environment location (fixes #6565)

### LINKED ISSUE #6565: "cabal v2-install --lib" uses the wrong directory for the global package environment on Windows
**Describe the bug**
cabal writes the package environment file in the directory returned by `getHomeDirectory`.
GHC uses `getAppUserDataDirectory`.
These are the same on Linux, but different on Windows (%HOMEPATH% and %APPDATA% respectively)

**To Reproduce**

```
cabal v2-install --lib vector
ghci
```

**Expected behavior**
The package environment file should be created in the correct directory and ghci should load it on startup.

**System information**
 - Windows 10
 - cabal 3.0.0.0, ghc 8.6.5

vituscze:
cabal-install currently assumes that the GHC environment files are always located in `$HOME/.ghc/`. However, GHC itself doesn't query the home directory directly. Instead, it uses `getAppUserDataDirectory "ghc"` ([GHC code](https://gitlab.haskell.org/ghc/ghc/-/blob/4898df1cc25132dc9e2599d4fa4e1bbc9423cda5/compiler/main/DynFlags.hs#L1514), [getAppUserDataDirectory](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:getAppUserDataDirectory)) which happens to coincide with `$HOME/.ghc/` on UNIX systems. On Windows, however, GHC ends up looking in `%APPDATA%/ghc/`.

As a result, `cabal install --lib` doesn't change the correct environment files on Windows.

This PR changes the environment file locations to match GHC. On UNIX, there is no change since `getAppUserDataDirectory "ghc"` is defined as `(</> ".ghc") <$> getHomeDirectory`. On windows, I have confirmed that cabal-install changes the correct GHC environment file with this PR.

---

### DIFF
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -2060,7 +2060,7 @@ pkgRoot verbosity lbi = pkgRoot'
       let ghcProg = fromMaybe (error "GHC.pkgRoot: no ghc program") $ lookupProgram ghcProgram (withPrograms lbi)
       in  fmap takeDirectory (getGlobalPackageDB verbosity ghcProg)
     pkgRoot' UserPackageDB = do
-      appDir <- getAppUserDataDirectory "ghc"
+      appDir <- getGhcAppDir
       let ver      = compilerVersion (compiler lbi)
           subdir   = System.Info.arch ++ '-':System.Info.os
                      ++ '-':prettyShow ver
--- a/Cabal/Distribution/Simple/GHC.hs
+++ b/Cabal/Distribution/Simple/GHC.hs
@@ -51,6 +51,7 @@ module Distribution.Simple.GHC (
         registerPackage,
         componentGhcOptions,
         componentCcGhcOptions,
+        getGhcAppDir,
         getLibDir,
         isDynamic,
         getGlobalPackageDB,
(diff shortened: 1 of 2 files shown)


phadej:
FWIW, the GHC documentation is then *incorrect*, https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html#package-environments explicitly mentions $(HOME)

cf. the package databases section which says https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html#package-databases

> The user package database private to each user. On Unix systems this will be $HOME/.ghc/arch-os-version/package.conf.d, and on Windows it will be something like C:\Documents And Settings\user\ghc\package.conf.d.

cc @bgamari

I suspect that GHC indeed relies on `getAppUserDataDirectory`, but that should be somehow be reflected in the manual. (And I have to check what we say about `~/.cabal` in our manual).

---

That said, this PR is good.

We should move the logic of finding GHC environment file directory somewhere in `Distribution.Simple.GHC`, https://hackage.haskell.org/package/Cabal-3.2.0.0/docs/Distribution-Simple-GHC.html#g:1 has already some logic there. If you could do that, that would be great. Otherwise I'll do that later.

vituscze:
Did you mean something like this? https://github.com/vituscze/cabal/commit/a57bb3396db1042666d8225ebd3bf54ce2e1dc9c

phadej:
@vituscze I'd rather go for simpler `getGhcAppDirectory :: IO FilePath`. We already have two `getAppUserDataDirectory "ghc"` in the code. replace that one with single `getGhcAppDirectory` (and use it in your patch).

if GHC sometimes decide to change its mind about where to place files, we'd need to change it only in one place.

PR #6826, opened 2020-05-20
Prepend hs-source-dir to match-component

phadej:
Squashed and rebased #6623 

---

Add Integration tests for File Target Selector

Add Integration test for Target Selecetors for files that
are part of other-modules in sub-directories.

Normalise Filepaths for target matching

Add filepath and module target syntax to readthedocs

Add ambiguous file target test-case


mpickering:
Thanks for doing the rebase @phadej

fendor:
Yes, thank you!

PR #6831, opened 2020-05-22
Fix Windows GitHub Action

phadej:
- msys2 should be installed
- there is cabal and GHC, but we want to be sure about
  versions, so we reinstall.

---

Let's see if this works.


phadej:
Looks like uninstalling GHCs and then reinstalling is not an option. `choco -y uninstall ghc` took so 15 minutes so far.

PR #6812, opened 2020-05-18
Solver: Support dependencies on sub-libraries (issue #6039)

phadej:
Rebased #6047

PR #6836, opened 2020-05-22
Solver: Support dependencies on sub-libraries (issue #6039) (3rd iteration)

fgaz:
Adds check overrides to #6812 / #6047

/cc @phadej @grayjay


fgaz:
Oh, [ci skip] doesn't work anymore :-/

phadej:
No. There is no easy check that [ci skip] tagged commit would be safe. CI is fast and reliable enough to just build everything.

fgaz:
I do see that it's waaay faster than before, thanks for that! And just to be clear, I'm merging this once it's green

grayjay:
Thanks for finishing my PR!

PR #6838, opened 2020-05-23
Backport filter --allow-depending-on-private-libs

fgaz:
In #6836 I noticed that I forgot to filter out the --allow-depending-on-private-libs flag for old Cabals. With that patch it becomes necessary to have the filtering, but I guess having it in 3.2 would be good too?

### DIFF
--- a/cabal-install/Distribution/Client/Setup.hs
+++ b/cabal-install/Distribution/Client/Setup.hs
@@ -540,6 +540,8 @@ filterConfigureFlags flags cabalLibVersion
               convertToLegacyInternalDep (GivenComponent pn LMainLibName cid) =
                 Just $ GivenComponent pn LMainLibName cid
           in catMaybes $ convertToLegacyInternalDep <$> configDependencies flags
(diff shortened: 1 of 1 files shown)


phadej:
Sorry, I didn't notice this patch was to 3.2 branch and not `master`. Is this change in `master`. Which commit introduced it?

fgaz:
It isn't in master yet. It's part of #6836. By the way, could you restart https://ci.appveyor.com/project/23Skidoo/cabal/builds/33067141 so we can merge that too? I'm pretty sure that failure is just random

phadej:
Let's avoid backporting stuff which is not merged yet. Can #6836 merged as CI is done?

Also you merged this even GitHub actions Windows 8.6.5 job was broken. Which is something I'm actively working on in #6828. I work hard to have especially release branches green, please don't merge if CI is not green.

I should enforced it for 3.2 branch too.

fgaz:
Yes, when #6836 is green we can add the changelog entry and merge.

Issue #6841, opened 2020-05-23
GHC 8.12 (master) wants `RankNTypes` pragma in Distribution/Client/Sandbox.hs


vdukhovni:
```
Distribution/Client/Sandbox.hs:713:7: error:
    • Illegal qualified type:
        GHC.Stack.Types.HasCallStack => IO WereDepsReinstalled
      Perhaps you intended to use RankNTypes
    • In the expansion of type synonym ‘Distribution.Compat.Stack.WithCallStack’
      In the expansion of type synonym ‘Cabal-3.3.0.0:Distribution.Compat.Prelude.IO’
      When checking the inferred type
        topHandler' :: Cabal-3.3.0.0:Distribution.Compat.Prelude.IO
                         WereDepsReinstalled
                       -> IO WereDepsReinstalled
    |       
713 |       topHandler' = topHandlerWith $ \_ -> do
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```
Enabling the pragma resolves the issue (assuming not a GHC bug).

phadej:
Please make a PR.

Issue #6840, opened 2020-05-23
GHC 8.12 does not like whitespace around `@` in `var@pattern`


vdukhovni:
Cabal-install does not build with GHC 8.12, because of bare `@` in https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Dependency.hs#L724
The trivial fix is to remove the surrounding whitespace.

phadej:
Please make a PR.

jneira:
The linked pr was finally merged so this one might be closed. I dont checked it but i am gonna close it optimistically.
@vdukhovni feel free to reopen if that is not the case.

PR #6844, opened 2020-05-24
Cabal-install GHC master compat

vdukhovni:
Bootstrapping cabal-install with ghc-master fails without these minor patches.
(Some of the package versions in bootstrap.sh also need updates, separate PR later).


vdukhovni:
This should address #6841 and #6840

phadej:
I won't act on this before https://gitlab.haskell.org/ghc/ghc/issues/18251 is addressed.

vdukhovni:
> I won't act on this before https://gitlab.haskell.org/ghc/ghc/issues/18251 is addressed.

Understood, though for the record, I mostly agree with the folks who say that you just need to deal with the incompatible change. :-(

PR #6846, opened 2020-05-24
More portable sed syntax in cabal-install

vdukhovni:
The sed syntax for munging GHC_ARCH in bootstrap.sh was not portable
outside Linux.  The "-E" flag variant was tested and works on at least recent
versions of Linux, MacOS and FreeBSD.


vdukhovni:
The MacOS build failure is spurious (some download randomly failed).

phadej:
Obsoleted by work on #6700 (#6979 #6719)

Issue #6847, opened 2020-05-25
Prepare Cabal for new Windows I/O manager in GHC 8.12


Mistuke:
**Describe the bug**

GHC 8.12 will have  a new asynchronous I/O manager for Windows. This new I/O manager will require changes to any library that uses GHC internal I/O modules such as `GHC.IO.FD`.

When the new I/O manager is enabled any use of `FD` will lead to an error.  So there are two options, avoid the use of GHC internals and stick to end user interfaces (i.e. only use `Handle`).

Or detect the new I/O manager support with the macro `__IO_MANAGER_WINIO__` and provide a codepath for versions of functions that use `HANDLE` instead of `FD`.

GHC provides some helper utilities to switch between the two. Namely the helper operator `<!>`.

Cabal seems to only use `FD` in `Cabal/Distribution/Compat/CreatePipe.hs`, so is this module still needed?

If it is, anonymous pipes can't be used and instead named pipes opened with `OVERLAPPED` support with proper security descriptors should be used.

**System information**
 - Windows
 - GHC 8.12

**Additional context**

Also the minimum supported versions of binaries compiled with GHC 8.12 will be moved up to Vista SP1. For while the user will be able to switch between the two managers at startup.

phadej:
I merged #6865, so the #6848 would resolve this issue. Am I right, @Mistuke ?

Mistuke:
> I merged #6865, so the #6848 would resolve this issue. Am I right, @Mistuke ?

yes, I have an update for it but haven't posted it yet as I'm trying to get clean bootstrap + testsuite run so I know all libraries are good. Should be done this week.

phadej:
This is resolved as far as I can see

PR #6848, opened 2020-05-25
WINIO: Add support for WINIO to Cabal.

Mistuke:
---

This makes `openNewBinaryFile` use `openBinaryTempFileWithDefaultPermissions` on Windows when the new I/O manager is used. It's my understanding that the only difference between the two is the file mode? `666` vs `600`. However `base` has supprted `openBinaryTempFileWithDefaultPermissions` for a while now with `666` as permissions.

Without this any calls to `openNewBinaryFile` in GHC 8.12 and the new I/O manager enabled will cause a panic.

Related to #6847

### DIFF
--- a/Cabal/Distribution/Compat/Internal/TempFile.hs
+++ b/Cabal/Distribution/Compat/Internal/TempFile.hs
@@ -10,16 +10,21 @@ module Distribution.Compat.Internal.TempFile (
 import Distribution.Compat.Exception
 
 import System.FilePath        ((</>))
-import Foreign.C              (CInt, eEXIST, getErrno, errnoToIOError)
 
 import System.IO              (Handle, openTempFile, openBinaryTempFile)
-import Data.Bits              ((.|.))
-import System.Posix.Internals (c_open, c_close, o_CREAT, o_EXCL, o_RDWR,
-                               o_BINARY, o_NONBLOCK, o_NOCTTY,
-                               withFilePath, c_getpid)
-import System.IO.Error        (isAlreadyExistsError)
-import GHC.IO.Handle.FD       (fdToHandle)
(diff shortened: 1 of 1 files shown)


Cabal/Distribution/Compat/Internal/TempFile.hs
```diff
@@ -40,6 +44,12 @@ import qualified System.Posix
 -- needs to be changed to make this better.
```
phadej, line 44: Do you mind updating the comment above, I guess `base` library *is changed*?

Mistuke: yeah, will do

Cabal/Distribution/Compat/Internal/TempFile.hs
```diff
@@ -36,10 +40,17 @@ import qualified System.Posix
 -- TODO: This file should probably be removed.
 
 -- This is a copy/paste of the openBinaryTempFile definition, but
--- if uses 666 rather than 600 for the permissions. The base library
--- needs to be changed to make this better.
+-- if uses 666 rather than 600 for the permissions. Newer versions
```
phadej, line 43: `s/if/it/` ?

phadej:
The Quick jobs failed, (#6890 will fix them soon).

phadej:
@Mistuke Could you fix a typo and rebase on top of `master`. Then we can merge this, I think.

Mistuke:
@phadej  not sure what that CI failure is.. seems to have failed checkout?

phadej:
CI was green in https://github.com/haskell/cabal/pull/6908 (with the same commit).

I merged it. If there is still something please open a followup!

Issue #6832, opened 2020-05-22
cabal v2-install fails late if symlink target already exists


fgaz:
(continuing from #5491)

Right now `cabal v2-install x` first builds x and then attempts to symlink its executables, failing if they already exist.

Instead, we should check for existing symlinks before building anything, and fail as early as possible. See https://github.com/haskell/cabal/issues/5491#issuecomment-433032100 and previous comments

jneira:
@fgaz hi, have you still plans to work on this? it seems a not too hard a nice change so i am gonna label with `pr welcome`

Issue #6481, opened 2020-01-15
Separate `cabal install --lib` into own command


phadej:
... e.g. `cabal env`.

People do confuse what "installing libraries" and "installing executables" means, and after answering on few issues, I think stronger that separation of commands will make UX clearer.

An alternative is to drop `--lib` flag, and teach `cabal install` to do **the** right thing every time. I don't see how it would work.

typedrat:
The only alternative that I can see is to make it so that `cabal install` just installs every public executable or library (or foreign library if someone who actually has a project to test with wants to land that nascent patch). I don't like that, but personally I don't like the idea of adding a whole new command, either.

phadej:
@typedrat the drawback of *install everything*, is that when you e.g. `cabal install hlint`, or `cabal install [`stylish-haskell`](https://hackage.haskell.org/package/stylish-haskell)` (which has more dependencies), all these will be in the *default* environment, therefore causing the "environment cabal hell" earlier.

That said, I guess there's a way to tell solver to solve for all components in the target package. I don't fully understand what will happen if the component sets change across versions: e.g. `yaml` package, with public libraries they might come and go: these are major version changes, but if you say `cabal install lens` it's unclear whether you want the latest version or just some version.

One way is to have `cabal install pkg` to always default to the library, and if you want executable, then you must say `cabal install pkg:exename`, which is "just" flipping the `--lib` default; but it's more explicit (and also related to https://github.com/haskell/cabal/issues/6369). One variation is to default to executable installation, if that's the only component; but that's not true for many exes (e.g. mentioned `hlint` and `stylish-haskell`, and [`cabal-install-2.2.0.0`](https://hackage.haskell.org/package/cabal-install-2.2.0.0), so `cabal install cabal-install` could not default to installing executable: there **is** a library).

This is a tough UX question.

Also I motivate `cabal env` command because then you can have subcommands, to
- list what's in the environment
- list available environments
- manage them in some other ways

One could argue that "it's just moving and editing files", but editing files is very bad UX; nor is well scriptable (c.f. `cabal user-config` for updating `~/.cabal/config`; way better than `sed`ing).

[...]

Issue #6393, opened 2019-11-30
cabal install https://.../package.tar.gz


phadej:
Should "just" work.

hvr:
This used to work with `v1-install` and got lost in `v2-install`

PR #6857, opened 2020-05-28
Fix #6856 Make `--lib` warning louder and clearer

### LINKED ISSUE #6856: People do not read (the Warning about `--lib` is regularly ignored)
**Describe the bug**

People show up at various places around the internet with complaints about cabal install not installing their library (see this recent example: https://stackoverflow.com/questions/62073719/why-is-installing-a-haskell-package-not-working).

**To Reproduce**

1. Make a useful tool
2. Gain popularity
3. Change it's behavior
4. Let humans be humans

**Expected behavior**
People read.

**Additional context**

I think we need the line:

```
Warning: You asked to install executables, but there are no executables in
target: binary-bits. Perhaps you want to use --lib to install libraries
instead
```

To be bold and red when the terminal allows.  Thoughts?

I'm not sure how to do this portably, but direct use of escape sequences does work:

```
Prelude> putStrLn "\x1b[31m Warning: You asked to install executables \x1b[0m"
 Warning: You asked to install executables
--- ^^^ That's red
```

TomMD:

In so far as I didn't even add a newline, just some output characters.

* [X] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).

I tested it by running the binary and observing the warning is now red on my screen.  I googled and found claims the color works on windows terminals as well.

### DIFF
--- a/cabal-install/Distribution/Client/CmdInstall.hs
+++ b/cabal-install/Distribution/Client/CmdInstall.hs
@@ -678,10 +678,14 @@ warnIfNoExes :: Verbosity -> ProjectBuildContext -> IO ()
 warnIfNoExes verbosity buildCtx =
   when noExes $
     warn verbosity $
-    "You asked to install executables, but there are no executables in "
-    <> plural (listPlural selectors) "target" "targets" <> ": "
-    <> intercalate ", " (showTargetSelector <$> selectors) <> ". "
-    <> "Perhaps you want to use --lib to install libraries instead."
+    "\n" <>
+    "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\n" <>
+    "@ WARNING: Installation might not be completed as desired! @\n" <>
+    "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\n" <>
+    "Without flags, the command \"cabal install\" doesn't expose" <>
+    " libraries in a usable manner.  You might have wanted to run" <>
+    " \"cabal install --lib " <>
+    unwords (showTargetSelector <$> selectors) <> "\". "
   where
     targets    = concat $ Map.elems $ targetsMap buildCtx
     components = fst <$> targets
--- a/changelog.d/issue-6856
+++ b/changelog.d/issue-6856
@@ -1,0 +1,4 @@
+synopsis: Adjust message indicating `--lib` is likely desired
+packages: cabal-install
+issues: #6856
+prs: #6857


changelog.d/issue-6856
```diff
@@ -0,0 +1,4 @@
+synopsis: Use PrettyField to format cabal file in `cabal init`
```
phadej, line 1: Hmm, this is not correct.

cabal-install/Distribution/Client/CmdInstall.hs
```diff
@@ -678,10 +678,10 @@ warnIfNoExes :: Verbosity -> ProjectBuildContext -> IO ()
 warnIfNoExes verbosity buildCtx =
   when noExes $
     warn verbosity $
-    "You asked to install executables, but there are no executables in "
+    "\x1b[31m You asked to install executables, but there are no executables in "
```
phadej, line 681: If we go for colors, we should do it properly by at least checking whether terminal supports them.

phadej: FWIW, also it's a warning. Not an error. Warnings are magenta (like in GHC :) )

phadej:
But generally. I'm not sure if this is a joke. Shouldn't we try to find and work (i.e. discuss) on proper solutions like #6481.

cabal-install/Distribution/Client/CmdInstall.hs
jneira: > If we go for colors, we should do it properly by at least checking whether terminal supports them.

Yeah, i guess in win dos only we'll see the funny escape codes. Well maybe that would attract the eye of users too 😄

TomMD: According to the internet ansi escape codes work in modern windows terminals.  I have no way to check.  Regardless, this doesn't mean the current terminal has support, I'd just expect most do.

TomMD:
It's not a joke. People consistently aren't reading the messaging and going to the internet for help.  An ASCII banner would be more portable and might have enough impact on behavior to be worth considering.

yaxu:
"You asked to install executables" - doesn't this come up when you just do e.g. `cabal install tidal`? `tidal` is a library, without any executables, in which case the warning would be incorrect. Making it red or magenta wouldn't make it any less confusing.

TomMD:
@yaxu Good point, the wording itself might be more of the problem.  It seems useful to re-iterate that there really is a problem and users get stuck regularly.  Hopefully everyone open to both fast and slow solutions to this issue.

Proposal: A bit of word-smithing and  ascii-banner to call out the message to get users on the right track.  This should be faster and less contentious than colors while not impeding the alternative linked by @phadej .

We currently say:

```
Warning: You asked to install executables, but there are no executables in
target: entropy. Perhaps you want to use --lib to install libraries instead.
```

How about instead we borrow from SSH and say:

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: Installation might not be completed as desired!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Without flags, command "cabal install" only installs executables.
You might have wanted "cabal install --lib ${TARGET}"
```

It's a fast change, can make the next release, and no one will feel like it was wasted work if the whole issue is better fixed before the release.

yaxu:
Agreed that's better. Howabout rather than `WARNING: Installation might not be completed as desired!` it gets to the point and says `WARNING: Nothing has been installed!`.

(I'm still unclear why it doesn't just install what's listed in the .cabal file like the old days, but understand there's some technical reason..)

phadej:
`WARNING: Nothing has been installed!.` is incorrect. Something is installed (namely library in the store).

I don't like this, but I'll accept a patch which
- Doesn't use ansi colors
- Isn't oversimplifying the message

We want to respect users and not assume that they are completely dumb and cannot read *at all*.

---

It would be nicer to cut the installation early, similarly to https://github.com/haskell/cabal/issues/6832
Then if `cabal-install` doesn't actually build anything but just output some (moderate amount of) stuff with a warning or error at the end, it will be visible.

But as I don't expect #6832 to be fixed for 3.4, I'll merge (the better version of) this PR to 3.4 branch when it's cut.

yaxu:
If the library isn't usable, I'd say it absolutely hasn't been installed.

I'd also say it's not the end-user that's stupid, but the software, for not installing a library when asked. Especially when this is undocumented, changed behaviour.. `cabal install` used to install libraries fine. 

Well this is worse than undocumented - it seems the old (working) behaviour is documented, and the current (non-working) behaviour is not. Here's some results from a google search result for "cabal install". Some results look authoritative and up-to-date. None explain that "cabal install" doesn't install libraries, or give details of '--lib'.

* https://hackage.haskell.org/package/cabal-install
* https://cabal.readthedocs.io/en/latest/intro.html
* https://wiki.haskell.org/Cabal-Install
* https://www.schoolofhaskell.com/user/simonmichael/how-to-cabal-install

Actually, I can't find any reference material that says how to install libraries with cabal.

phadej:
@yaxu, It's not like we (or myself) not trying to improve situation like I pointed in previous comments referring to other issues in this space.

If it is how you think, then I'll merge this PR, but don't assume that we (well, me) will have any motivation to fix the root cause. This warning should be then enough for unforeseeable future.

yaxu:
Apologies for my tone there @phadej, in my defence I hadn't yet eaten lunch.

I'm super grateful for this and everything, just a little frustrated sometimes with the teething problems, but thanks a lot!

TomMD:
> Doesn't use ansi colors

Done. Using the banner instead, which is very SSH in appearance.

> Isn't oversimplifying the message

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@ WARNING: Installation might not be completed as desired! @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Without flags, the command "cabal install" doesn't expose libraries in a
usable manner. You might have wanted to run "cabal install --lib tidal".
```

This appears to be saying really only as much as we know

* We're talking about installing something
* It might or might not have been done how the user wants (and perhaps they do want a copy to the store without link to package env, it's just the rare case)
* Any installed libraries are not directly "usable", by typical developers using GHCi, without "--lib".  The word "usable" masks a world of nuance with which we are walking a fine line of respecting peoples time and existing knowledge, not forcing them to read a long winded explanation.

> Well this is worse than undocumented - it seems the old (working) behaviour is documented, and the current (non-working) behaviour is not.

Interesting and terrifying statement, @yaxu. While it's not related to this PR it is related to the issue I'm trying to solve (public education and cold-start ability re: Haskell tooling). Looking deeper:

* @phadej was the one who made the best documentation in exactly the right spot in the official readthedocs, 2 months ago (3752c3e5ed)
* The intro you link should probably mention `--lib.  Care to make a PR?  The file is `Cabal/doc/intro.rst`.
* The wiki links is flags as out of date right at the top.  Should probably be deleted if you care to take a moment I'd encourage it.
* The hackage page at least isn't incorrect as far as I looked but could probably be revamped to link to readthedocs, mention ghcup etc.

Care to make an issue to discuss and push through a PR?

yaxu:
I think intro probably needs the attention of someone more expert than me, it's currently documenting v1- instructions so probably needs more updates than just adding --lib. For example it gives commands for installing packages from tarballs and URLs which as far as I know isn't supported by v2-install.

phadej:
Installing packages from tarballs is supported in `master`: https://github.com/haskell/cabal/issues/6393

PR #6875, opened 2020-06-05
Fix ambiguous file target selectors causing an internal error

### LINKED ISSUE #6874: Ambiguous TargetSelector crash
**Describe the bug**
In the project [haskell-language-server](https://github.com/haskell/haskell-language-server), calling `cabal repl exe/Arguments.hs` from the root directory will give this error:
```
$ cabal repl exe/Arguments.hs 
TargetSelector: internal error: NoMatch
  error, called at ./Distribution/Client/TargetSelector.hs:669:3 in main:Distribution.Client.TargetSelector
```

This is because a subproject `ghcide` which is included in the packages field of `cabal.project` also has a module in a folder `ghcide/exe/Arguments.hs`.
This doesn't happen when calling `cabal repl exe/Arguments.hs` from inside `ghcide`.

**Steps to reproduce**
Working on getting a minimal test case

**Expected behavior**
Presumably in this scenario, `cabal repl src/Arguments.hs` should resolve to the `src/Arguments.hs` in the root project.

**System information**
 - cabal HEAD

bubba:
These should have been returning an error message but instead were
causing an internal error because disambiguateTargetSelectors was
rendering syntax and rematching on it, which isn't equivalent. Due to
the way syntaxForm1File renders, it does not add a FileStatus to its
TargetStringFileStatus and so cannot be matched upon again.
The fix is to just copy over the FileStatus from the match input.
This fixes #6874

### DIFF
--- a/Cabal/Distribution/Compat/Prelude.hs
+++ b/Cabal/Distribution/Compat/Prelude.hs
@@ -137,7 +137,7 @@ module Distribution.Compat.Prelude (
     readMaybe,
 
     -- * Debug.Trace (as deprecated functions)
-    traceShow, traceShowId,
+    trace, traceShow, traceShowId,
     ) where
 
 -- We also could hide few partial function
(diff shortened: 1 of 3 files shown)


bubba:
Not sure why the macOS validate job keeps on getting cancelled. Can you retrigger it?

PR #6877, opened 2020-06-05
Tell users about cabal.project.local~

bubba:
---

### DIFF
--- a/cabal-install/Distribution/Client/CmdConfigure.hs
+++ b/cabal-install/Distribution/Client/CmdConfigure.hs
@@ -95,7 +95,7 @@ configureAction flags@NixStyleFlags {..} _extraArgs globalFlags = do
     -- before overwriting
     exists <- doesFileExist "cabal.project.local"
     when exists $ do
-        notice verbosity "'cabal.project.local' file already exists. Now overwriting it."
+        notice verbosity "'cabal.project.local' already exists, backing it up to 'cabal.project.local~'."
         copyFile "cabal.project.local" "cabal.project.local~"
     writeProjectLocalExtraConfig (distDirLayout baseCtx)
                                  cliConfig
--- a/cabal-install/Distribution/Client/CmdConfigure.hs
+++ b/cabal-install/Distribution/Client/CmdConfigure.hs
@@ -10,6 +10,7 @@ import Distribution.Client.Compat.Prelude
 import Prelude ()
 
 import System.Directory
+import System.FilePath
 import qualified Data.Map as Map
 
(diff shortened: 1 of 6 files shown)


cabal-testsuite/PackageTests/NewConfigure/LocalConfigOverwrite/cabal.out
```diff
@@ -1,5 +1,5 @@
 # cabal v2-configure
-'cabal.project.local' file already exists. Now overwriting it.
+'cabal.project.local' already exists, copying it to 'cabal.project.local~'.
```
phadej, line 2: `s/copying/backupin/`.

Though I don't remember what happens when `cabal.project.local~` already exists.

bubba: It just gets overwritten it looks like

phadej: Ok, let us fix that. Otherwise we could just overwrite `cabal.project.local` directly

bubba: Fix by adding a message or by backing up cabal.project.local~ somewhere further?

cabal-install/Distribution/Client/CmdConfigure.hs
```diff
@@ -95,7 +95,7 @@ configureAction flags@NixStyleFlags {..} _extraArgs globalFlags = do
     -- before overwriting
     exists <- doesFileExist "cabal.project.local"
     when exists $ do
```
phadej, line 97: `when exists` let's search for the first one of `cabal.project.local~`, `cabal.project.local~0` `cabal.project.local~1` etc. until one doesn't.

Note: we need to change this to take into account `--project-file` as we on it. I wonder if there's an issue about that too.

cabal-install/Distribution/Client/TargetProblem.hs
```diff
@@ -45,8 +45,8 @@ data TargetProblem a
     | TargetProblemNoSuchPackage           PackageId
     | TargetProblemNoSuchComponent         PackageId ComponentName
 
-    | CustomTargetProblem a
       -- | A custom target problem
+    | CustomTargetProblem a
```
bubba, line 49: This was to fix a haddock failure

phadej:
CI fails for real

bubba:
CI looks green now

Issue #4746, opened 2017-09-06
cabal new-build does not track files added via TH’s addDependentFile


cocreature:
### Steps to reproduce
* Download https://gist.github.com/cocreature/d07ef6f35f7bd9043db50794fe31e977. 
* Build using `cabal new-build`
* Edit `fileinput`
* Try rebuilding using `cabal new-build`.

### Expected behavior
`cabal new-build` rebuilds because `fileinput` has changed.

### Actual behavior
`cabal new-build` exits without rebuilding after saying `Up to date`.

This issue affects both `cabal-2.0` as well as the master branch.

phadej:
One hack solution is to watch `extra-source-files` too. But that won't solve more dynamic (e.g. what `inline-c` uses?) stuff.

23Skidoo:
BTW, is there a way to make `ghc --make` always rerun a TH function? E.g. if I'm doing some kind of I/O at compile-time, like reading from an environment variable.

phadej:
@23Skidoo that would force GHC to always build that module, that doesn't sound like a very good idea.

23Skidoo:
Yep, but sometimes you want that in a build system. For example, Shake rules can be marked `alwaysRerun`.

Re: `extra-source-files`, that field is package-global and commonly used for things like `README` and `changelog`. So we'd need to allow per-component `extra-source-files` to avoid unnecessary recompilation

23Skidoo:
An alternative way to fix this is to get the list of files to be monitored from `ghc --make` via `-ddump-module-graph`.

phadej:
Does `-ddump-module-graph` dump `addDependentFile` & co stuff?

23Skidoo:
@phadej If you click that link, you'll see that I asked that exact question there.

cocreature:
Given that this feature will only be in `GHC 8.4` I don’t think that’s a reasonable solution at this point.

23Skidoo:
@cocreature With old versions we can simulate that feature with `ghc -M` (which will be slower because we'll have to do dependency resolution twice - once for `ghc -M` and once for `ghc --make`, but for legacy GHC versions we can afford to not care). BTW, `ghc -M` does support `addDependentFile` dependencies:

```
# DO NOT DELETE: Beginning of Haskell dependencies
Data/FileEmbed.so : Data/FileEmbed.hs
Main.so : Main.hs
Main.so : Data/FileEmbed.shi
[...]

PR #6889, opened 2020-06-11
Rebuild on changes in all extra-source-files

kmicklas:
This should alleviate #4746 for most use cases, if the dependent files
are declared in `extra-source-files`. The downside is that `cabal`
will not be able to avoid invoking GHC in as many cases, but GHC's
`--make` should be smart enough to avoid any "actual" rebuilding that
isn't necessary.

----


phadej:
Stuff in `extra-source-files` are sources, so most likely you want a rebuild when they change.

And stuff like `README` or `CHANGELOG` don't change *that* often.

This is simpler and there are no real drawbacks either.

Thanks!

PR #6907, opened 2020-06-15
Fix #6083: allow depending on public sublibrary with pkg:lib qualified syntax

### LINKED ISSUE #6083: Cabal 3: in build-depends, package name gets treated as internal library name ("legacy" syntax) even when using the pkg:lib qualified syntax

phadej:


phadej:
This still fails due missing `build-tool-depends`. Will fix that ASAP.

PR #6923, opened 2020-06-21
Fix ghci being launched before other sources are built

bubba:
The following case previously failed
```
$ cat foo.c
int foo() { return 42; }
$ cat Lib.hs
module Lib where

foreign import ccall "foo" foo :: Int

bar = foo
$ cat cabal-csrc-repl.cabal
cabal-version:      2.4
name:               cabal-csrc-repl
version:            0.1.0.0
library
    exposed-modules:  Lib
    build-depends:    base ^>=4.14.0.0
    C-sources:        foo.c
    default-language: Haskell2010
$ cabal v2-repl
Resolving dependencies...
Build profile: -w ghc-8.10.1 -O1
 - cabal-csrc-repl-0.1.0.0 (lib) (first run)
Configuring library for cabal-csrc-repl-0.1.0.0..
Preprocessing library for cabal-csrc-repl-0.1.0.0..
GHCi, version 8.10.1: https://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Lib              ( Lib.hs, interpreted )
Ok, one module loaded.
*Lib> foo
ghc: ^^ Could not load '_foo', dependency unresolved. See top entry above.
```

Looks like an accident from a6e427a


bubba:
ping @phadej

cabal-testsuite/src/Test/Cabal/Run.hs
```diff
@@ -56,13 +56,24 @@ run _verbosity mb_cwd env_overrides path0 args = do
...
     withAsync drain $ \sync -> do
 
-    -- NB: do NOT extend this to take stdin; then we will
-    -- start deadlocking on AppVeyor.  See https://github.com/haskell/process/issues/76
-    pid <- runProcess path args mb_cwd mb_env Nothing {- no stdin -}
-                (Just writeh) (Just writeh)
+    let prc = (proc path args)
+          { cwd = mb_cwd
+          , env = mb_env
+          , std_in = case input of { Just _ -> CreatePipe; Nothing -> Inherit }
```
phadej, line 62: Inheriting `stdin` is a bad idea?

cabal-testsuite/src/Test/Cabal/Prelude.hs
```diff
@@ -675,7 +679,7 @@ recordNormalizer f =
 assertOutputContains :: MonadIO m => WithCallStack (String -> Result -> m ())
 assertOutputContains needle result =
     withFrozenCallStack $
-    unless (needle `isInfixOf` (concatOutput output)) $
+    unless (needle `isInfixOf` (concatOutput output)) $ do
```
phadej, line 682: undo

cabal-testsuite/src/Test/Cabal/Prelude.hs
```diff
@@ -255,13 +256,16 @@ cabal :: String -> [String] -> TestM ()
...
+cabalWithStdin :: String -> [String] -> String -> TestM Result
+cabalWithStdin cmd args input = cabalG' [] cmd args (Just input)
 
 cabalG :: [String] -> String -> [String] -> TestM ()
-cabalG global_args cmd args = void (cabalG' global_args cmd args)
+cabalG global_args cmd args = void (cabalG' global_args cmd args Nothing)
 
-cabalG' :: [String] -> String -> [String] -> TestM Result
-cabalG' global_args cmd args = do
+cabalG' :: [String] -> String -> [String] -> Maybe String -> TestM Result
```
phadej, line 267: This breaks informal naming rule, the primed version returns `Result`, but otherwise is the same.

Rather make a newly named function (and thus avoid editing unrelated test cases).

Cabal/Distribution/Simple/GHC.hs
```diff
@@ -772,6 +768,9 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
   -- TODO: problem here is we need the .c files built first, so we can load them
```
phadej, line 768: what about this todo?

Sadly it looks like that we don't have test for that. 

But this TODO makes me worry what we might break.

cabal-testsuite/src/Test/Cabal/Run.hs
bubba: This mirrors the existing behaviour of [`runProcess`](https://hackage.haskell.org/package/process-1.6.10.0/docs/System-Process.html#v:runProcess)

> Handle to use for stdin (Nothing => use existing stdin)

bubba: Happy enough to just create a pipe for it anyway regardless of whether or not its used

Cabal/Distribution/Simple/GHC.hs
bubba: This comment was here before a6e427a, so presumably it wasn't fixed given that c files that didn't depend on the ffi generated .h files didn't even load. As far as I can tell, nothing that depended on a c source would have worked, so there's nothing really to break

DanielG: It seems the comment is talking about this "feature": https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/ffi.html#using-foreign-export-and-foreign-import-ccall-wrapper-with-ghc

The C sources could be depending on these `_stub.h` headers GHC generates for `foregin export` decls.

cabal-testsuite/src/Test/Cabal/Run.hs
DanielG: Since this is mirroring existing behaviour this should be fine.

phadej:
I have troubles writing a changelog entry for this fix.

The title of this PR doesn't make sense standalone.

bubba:
@phadej maybe something like

> Fix an issue where `cabal repl` would fail to build C sources before launching the GHCI session

phadej:
I'm suspecting changes in `run` cause https://github.com/haskell/network/issues/474 regression.

If it so, I'll will revert this PR.

phadej:
Update, it wasn't that. With `cabal-install-3.2.0.0` the

```
cabal install --lib network 0<&-
```

also fails.

I.e. inheriting stdin makes cabal fail `configure`.

phadej:

PR #6927, opened 2020-06-27
Don't use `--detach` when cloning, fixes #6888

### LINKED ISSUE #6888: source-repository-package branch failure
With:

```
source-repository-package
  type: git
  location: git://github.com/georgefst/pretty-simple.git
  branch: experimental
```

at the top of my `cabal.project`, I get the failure `fatal: '--detach' cannot be used with '-b/-B/--orphan'`. Unclear what's causing this. 

`git` version is `2.7.4`, which is pretty old (can't upgrade right now on my work machine), so if that's the issue, I'll just work around it.

liftM:
---

  - **Help needed:** Should add a changelog file for this fix?
  - I have not made any documentation updates. This PR does not change intended behaviour - it merely fixes a bug in an existing implementation.

---

I think this fixes #6888 but I'm not sure how to write an automated test for this.

#6888 happens because we use `--detach` even when we are checking out branch names. This causes `git` to fail with the error message `fatal: '--detach' cannot be used with '-b/-B/--orphan'`. To the best of my knowledge, there's no reason to ever prefer `--detached` over regular tag names, so I've removed that flag.

I tested this fix by using the binary built from this commit to successfully build and run my project that depends on a non-`master` branch of a Git repository for an external package. [Here](https://github.com/liftM/hipsterfy/tree/02751f04848d26d4d4ab5615d67b495e36dc2252) is the exact commit that does not compile on with cabal-install 3.2.0.0, but does compile with this patch.

### DIFF
--- a/cabal-install/src/Distribution/Client/VCS.hs
+++ b/cabal-install/src/Distribution/Client/VCS.hs
@@ -384,7 +384,7 @@ vcsGit =
                            Just peerLocalDir -> ["--reference", peerLocalDir]
                       ++ verboseArg
                          where loc = srpLocation
-        checkoutArgs   = "checkout" : verboseArg ++ ["--detach", "--force"
+        checkoutArgs   = "checkout" : verboseArg ++ ["--force"
                          , checkoutTarget, "--" ]
         checkoutTarget = fromMaybe "HEAD" (srpBranch `mplus` srpTag)
         verboseArg     = [ "--quiet" | verbosity < Verbosity.normal ]


liftM:
This is now ready for review.

liftM:
I'd love to help, but I'm not really sure where to start on adding tests.

The errors that this PR fix occur when you run `git checkout --detach B` on a newly cloned repository without first running `git branch B` (i.e. when `B` does not yet exist as branch).

When you first `git clone` a remote repository, none of the remote branches have their equivalent local branches created yet. The fully expanded way to check out a remote branch `B` is:

```bash
# Clone the repository.
git clone repo

# Construct a new local branch named "B" that tracks the branch named "B" from the remote "origin".
git branch B --track origin/B

# Switch to the local branch named "B".
git checkout B
```

Running `git checkout B` when there exists a branch `B` on the remote `origin` is actually shorthand for `git checkout -b B --track origin/B`, and `git checkout -b B --track origin/B` is actually shorthand for `git branch B --track origin/B && git checkout B`. See the documentation on [shorthand for remotes](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutemltbranchgt) and [shorthand for `-b`](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnewbranchgtltstartpointgt).

This is why an error occurs when passing `--detach` to `git checkout` when checking out a remote branch from a freshly cloned repository: the shorthand `git checkout --detach B` (which is what Cabal currently runs) expands to an illegal command.

I'm not sure how to construct this scenario in the tests. I think the right answer might be to modify `vcsTestDriverGit` to initialize a freshly cloned repository, or to construct a new set of git-specific property tests. I'd appreciate any advice here.

avieth:
Would it suffice to add a test which verifies that checking out a branch name (rather than a commit hash) works as expected? Seems to me we can be very confident that this patch won't break anything... cabal won't ever make any commits to the repositories it clones, so it doesn't matter whether they are in detached HEAD state, right?

jneira:
@liftM hi, thanks for the pr, are you planning to continue it? not sure if you readed @liskin suggestion above to correct the fix and the @avieth one on how to add a test (which would be definitely needed)

jneira:
I was revising conflicts and `git checkout` has been replaced by `git reset` and it does not have `-detach` anymore. But it continue failing with branches other than master.
So i think it is better to open a new pr following @liskin suggestion. Thanks for the pr anyways!

PR #6932, opened 2020-06-29
WIP rebuild

phadej:
The goal is to run planning of `cabal v2-build` completely in `Rebuild` monad, and not do `runRebuild` here and  there.


phadej:
I have no plans to continue working on this.

PR #6943, opened 2020-07-07
Remove redundant call to 'ifReplLib'

fendor:
---

I did not test it, but since the function definition for `ifReplLib` is as follows 
```haskell
      forRepl = maybe False (const True) mReplFlags
      ifReplLib = when forRepl
```
I argue that it is safe due to referential transparency and idempotency of the `when b` function.


DanielG:
https://github.com/haskell/cabal/pull/6923/files#diff-9ce58632269448787c6e6e270277e03fR773 already seems to include this change :)

Issue #6667, opened 2020-04-06
Introduce Path abstraction


phadej:
Like the one already used in `hackage-security`.

We want to have `Absolute` rooted `Path`s, but also more specific like `StoreDir`, or `PackageDir` and `ProjectDir` which all would make handling paths a bit more robust.

It would helped a lot in my recent sdist refactor. Getting relative to `PackageDir` paths is tricky, not sure if I haven't missed any.

I however suggest to store these (file paths) in native way. (Not in POSIX like `paths` library does), because I'm quite sure that conversion from Windows to Posix and back won't roundtrip on fancier paths there might be. Therefore we might need separate PosixPath for tarballs etc if we end up needing such notion (and we might, index files...)

phadej:
https://oleg.fi/gists/posts/2020-09-13-a-design-for-paths.html

tbidne:
I am interested in this issue, motivated by https://github.com/haskell/cabal/issues/11004. I see `hackage-security` has a [custom Path type](https://github.com/haskell/hackage-security/blob/master/hackage-security/src/Hackage/Security/Util/Path.hs), which would work, though I wonder if there are any libs that would save some work. I am aware of:

|  feature/lib | [path](https://hackage.haskell.org/package/path) | [paths](https://hackage.haskell.org/package/paths) | [strong-path](https://hackage.haskell.org/package/strong-path) | [hpath](https://hackage.haskell.org/package/hpath) |
|:-------------|-----:|------:|------------:|------:|
| Dir vs. File |   ✅ |    ❌ |          ✅ |    ❌ |
|  Abs vs. Rel |   ✅ |    ✅ |          ✅ |    ✅ |
|    Typed Rel |   ❌ |    ✅ |          ✅ |    ❌ |
|       OsPath |   ✅ |    ❌ |          ❌ |    ❌ |
|     Portable |   ✅ |    ❔ |          ✅ |    ❌ |
|   Maintained |   ✅ |    ❔ |          ❔ |    ❔ |

First, I've only really used `path`, so do correct me if there are mistakes. Some notes:

- Support (or at least a potential migration) to `OsPath` would be really nice. Because `strong-path` wraps the type from `path` -- which does support `OsPath` -- at least `strong-path` could likely support `OsPath` relatively easily.
- `Portable`: Meaning supports windows and posix.
- `Maintained`: I just looked at recent commits / hackage revisions.

[...]

PR #6948, opened 2020-07-08
Convert Json module from String to ByteString

bubba:
Taken from https://github.com/haskell/cabal/pull/6241


bubba:
cc @phadej

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,54 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- | Extremely simple JSON helper. Don't do anything too fancy with this!
+
+module Distribution.Utils.Json
+    ( Json(..)
+    , (.=)
+    , renderJson
+    ) where
+
```
phadej, line 9: Use `Distribution.Compat.Prelude` (look in other modules), that might make CI work (on older GHCs).

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,54 @@
...
+import Data.ByteString.Lazy (ByteString)
+import qualified Data.ByteString.Lazy as L
+import qualified Data.ByteString.Lazy.Char8 as C
+import Data.ByteString.Builder
```
phadej, line 14: Either add an import list, or import qualified.

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,54 @@
...
+
+data Json = JsonArray [Json]
+          | JsonBool !Bool
+          | JsonNull
+          | JsonNumber !Int
+          | JsonObject [(ByteString, Json)]
+          | JsonString !ByteString
```
phadej, line 21: What kind of strings are used here. I'd use either `ShortText` or `String` inside `Json` type.

E.g. package-ids, unit-ids are backed up by `ShortText` (and they are in UTF8, so if they don't need escaping they can be used as is).

Yet, currently filepaths are `FilePath` i.e. `String`.

Here are two issues:
- The keys and strings are textual
- double conversions. Thus I'd use `ShortText` for keys in `JsonObject` (as these are most likely static), and would use `String` for `JsonString` (as it looks like most producers are `prettyShow` or alike).
  - I'd like it to be `ShortText`, but if most producers atm are producing `String`, that would be silly. Yet, there could be a comment to eventually change it to `ShortText` (Path abstraction is related: #6667)

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,54 @@
...
+import Distribution.Compat.Prelude
+import Data.ByteString.Lazy (ByteString)
+import qualified Data.ByteString.Lazy as L
+import qualified Data.ByteString.Lazy.Char8 as C
+import Data.ByteString.Builder
+
+data Json = JsonArray [Json]
+          | JsonBool !Bool
+          | JsonNull
+          | JsonNumber !Int
```
phadej, line 19: Add comment that we don't support floating point, because we don't need that (I suppose).

Cabal/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,54 @@
...
+          | JsonObject [(ByteString, Json)]
+          | JsonString !ByteString
+
+renderJson :: Json -> L.ByteString
```
phadej, line 23: Should have some haddock (All exported functions should have). Ditto for `.=`

Cabal/Distribution/Utils/Json.hs
bubba: Is it really a double conversion if this all gets converted to a `ByteString` at the end of the day? As far as I can see for `JsonString`, all the filepaths etc. will go from `String` -> `ByteString`, and then once its in a `ByteString` the builder means it will be cheap to insert it and move it around

Cabal/Distribution/Simple.hs
```diff
@@ -104,6 +104,7 @@ import Distribution.Compat.Directory        (makeAbsolute)
 import Distribution.Compat.Environment      (getEnvironment)
 import Distribution.Compat.GetShortPathName (getShortPathName)
 
+import qualified Data.ByteString.Lazy as B
```
phadej, line 107: as LBS or BSL. We use both, but LBS is more common. (I haven't got to normalise that yet).

Cabal/Cabal.cabal
```diff
@@ -349,7 +349,7 @@ library
 
   if !impl(ghc >= 7.8)
     -- semigroups depends on tagged.
-    build-depends:  tagged >=0.8.6 && <0.9
+    build-depends:  tagged >=0.8.6 && <0.9, bytestring-builder >= 0.10.8 && <0.11
```
phadej, line 352: This is wrong as is, but I'll fix it before merge. One can have older bytestring on newer GHC, though that is contrived corner-case. But Cabal should be exemplary in own definition :)

bubba: Don't we need a newer bytestring on an older GHC here though?

phadej:
I'll postpone merging this after I cut the 3.4 branch, which I hopefully will manage to do today.

bubba:
Failing test looks flakey

DanielG:
Now that 3.4 has been released, what's the status of this PR?

jneira:
uh the tag is in this repo but the download page has not been updated?
https://www.haskell.org/cabal/download.html
still 3.2 there afaics

fgaz:
It may be because cabal-install 3.4 is still unreleased

jneira:
oops sorry even tag is for Cabal and no cabal-install

fgaz:
@bubba could you rebase?

gbaz:
Closing in favor of the rebased pr.

PR #6949, opened 2020-07-09
Accept "linux-androideabi" as an alias for Android

phadej:
rebased #6301 (added a changelog entry).


phadej:

Issue #6961, opened 2020-07-14
Cannot depend on internal library from tests


phadej:
```
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cabal-install-3.5.0.0 (user goal)
[__1] rejecting: cabal-install:!test (constraint from config file, command
line flag, or user target requires opposite flag selection)
[__1] rejecting: cabal-install:*test (requires library
'cabal-install-solver-dsl' from cabal-install, but the component is private)
[__1] fail (backjumping, conflict set: cabal-install, cabal-install:test)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cabal-install, cabal-install:test
```

with `cabal-install-3.4.0.0-rc1`

cc @grayjay @fgaz 

EDIT: for easy reproduction check https://github.com/phadej/cabal-fmt

```
% cabal build all --dry            
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cabal-fmt-0.1.3 (user goal)
[__1] rejecting: cabal-fmt:!test (constraint from config file, command line
flag, or user target requires opposite flag selection)
[__1] rejecting: cabal-fmt:*test (requires library 'cabal-fmt-internal' from
cabal-fmt, but the component is private)
[__1] fail (backjumping, conflict set: cabal-fmt, cabal-fmt:test)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cabal-fmt, cabal-fmt:test
```

If I comment a test-suite there, then the solver will find the solution, i.e. dependency from an executable in the same package is fine!

phadej:
If this cannot be solved, I'm afraid we need to revert the solver patch for multilibs.

grayjay:
I think the issue is that the solver merges all components within a package and doesn't try to handle intra-package dependencies, but the multilibs change checked the visibility for all dependencies.  The solver filters out intra-package dependencies in the build phase (https://github.com/haskell/cabal/blob/6f22f2a789fa23edb210a2591d74ea6a5f767872/cabal-install/Distribution/Solver/Modular/Builder.hs#L76-L85), but that doesn't affect the component check in Validate.hs.

@phadej Thanks for making a fix and adding more tests!  I think  #6964 will fix this issue, though a more general fix would be to filter out intra-package dependencies earlier in dependency solving.

phadej:
@grayjay Do you mind opening an issue for 

[...]

PR #6964, opened 2020-07-14
Add tests and fix for #6961

phadej:


cabal-testsuite/PackageTests/Regression/T6961/DepInternal/pkg-bar/cabal.out
```diff
@@ -0,0 +1,12 @@
...
+Build profile: -w ghc-<GHCVER> -O1
+In order, the following will be built:
+ - issue6961-0 (lib:internal-lib) (first run)
+ - issue6961-0 (exe:example) (first run)
+Configuring library 'internal-lib' for issue6961-0..
+Preprocessing library 'internal-lib' for issue6961-0..
+Building library 'internal-lib' for issue6961-0..
+Configuring executable 'example' for issue6961-0..
+Preprocessing executable 'example' for issue6961-0..
+Building executable 'example' for issue6961-0..
```
grayjay, line 12: I'm not sure I understand why there are several cabal.out files per test.  Shouldn't this log show a failure, since pkg-bar depends on pkg-foo's internal library?

phadej: Goot catch. This `caba.out` is just copied here as I made variants of the test, it's not actually a result of test run (see `cabal.out` in a directory one up).

phadej:
@fgaz, most likely they do.

In fact I'd preferred to just extend `unit-tests` with these scenarios, but it seems that framework doesn't allow asking for test-suites stanzas. I might be wrong.

cabal-testsuite/PackageTests/Regression/T6961/DepInternal/pkg-bar/cabal.out
phadej: I removed these extra files.

phadej:
Backported in https://github.com/haskell/cabal/pull/6967

PR #6970, opened 2020-07-16
Move doc/ to the top-level of the repository

phadej:

PR #6971, opened 2020-07-19
Fix broken links in user's guide + few haddock fixes

jhrcek:
I finished going through the user's guide and here's the last batch of fixes + couple of minor haddock fixes.

The build of Cabal library is broken at the moment as commented [here](https://github.com/haskell/cabal/pull/6970#issuecomment-660450922)


phadej:
Thanks!

Issue #6820, opened 2020-05-19
cabal-testsuite: add reasons for SKIP


phadej:
When running test-suite, why some tests are skipped?

Currently I'm guessing there's some environment setup issues so some tests are not run, but it's impossible to tell.

Issue #6952, opened 2020-07-09
Teach `cabal.project` about "dev dependencies"


michaelpj:
Many build tools have the concept of a "dev dependency". These are dependencies which are managed by the build tool, but which are not used by any of the actual build products. Instead, they are there to provide tools that developers of the project will use.

Why have such things managed by the build tool? Can't users just install them separately? There are a few advantages, such as convenience and consistency between developers, but I think there is one that is very important in our case: the build tool may have information that makes it much easier for it to get the "right" tool. For example, many development tools themselves expect to interact with the language toolchain and build tool. It can therefore be important to pick a tool version that's compatible with the toolchain that you're using.

This is becoming very pertinent in the Haskell tooling ecosystem as more things depend on the `ghc` library. A tool which depends on `ghc`, like `haskell-language-server`, must be built with a version of `ghc` that exactly matches the compiler the user is using for a particular project (which may indeed differ between projects!). As it stands, this creates a lot of work for users and developers. Users must carefully manage their tool installations to make sure they have precisely the right version of the tool installed for the project in question.

But we already have a tool that has all the information about what version of `ghc` you are using and how that affects what versions of `haskell-language-server` you can use: namely `cabal`!

----

So much for the motivation, what am I concretely suggesting?

- `cabal.project` learns a new field: `dev-tools`, which is a list of component specifications a la `build-tool-depends`.
- Some `cabal` command will build all the `dev-tools` and install them into some project-local location.
    - I don't know what an appropriate command is or an appropriate location (although I think it should be project-local): suggestions welcome!

I don't *think* the `dev-tools` build should include constraints from the components in the project - the implicit constraints on `ghc`, `Cabal` etc. should be enough.

----

Partial duplicate of https://github.com/haskell/cabal/issues/5588, but I'm making quite a different case, which I think is much more compelling than just convenience.

Many of the other build tools which have "dev dependencies" also use them for things that have better solutions in `cabal`, e.g. for build tools, test dependencies etc. I think there's still a useful role for dev dependencies that are really for developers only.

A short list of things I expect people might want to put into `dev-tools`:
- Language servers: `haskell-language-server`/`ghcide` etc.
- Formatters: `ormoulu`/`stylish-haskell`/`brittany` etc.
- Linters: `hlint`/`stan` (`stan` does actually rely on `.hie` files and hence depends on `ghc`!)

[...]

PR #6972, opened 2020-07-19
Consider Installed packages when pruning the install plan

alexbiehl:
The discussion in #6952 indicated that extra-packages stanzas wouldn't quite work yet. It turns out in order for cabal to find exes for already installed extra-packages we need to also consider `installed` packages when pruning the install plan.

Here is what you can do with it: 

```patch
diff --git a/cabal.project b/cabal.project
index af1d8a383..b1d724259 100644
--- a/cabal.project
+++ b/cabal.project
@@ -7,6 +7,8 @@ packages: Cabal/Cabal-tree-diff/
 packages: Cabal/Cabal-described
 packages: cabal-benchmarks/
 
+extra-packages: stylish-haskell
+
 -- Uncomment to allow picking up extra local unpacked deps:
 --optional-packages: */
```

Using HEAD

```
$ cabal run cabal-install -- run -- stylish-haskell --version
Up to date
cabal: Unknown executable stylish-haskell in package
stylsh-hskll-0.9.3.0-cdf9f8e3
```

With this patch applied

```
$ cabal run cabal-install -- run -- stylish-haskell --version
Up to date
stylish-haskell 0.9.3.0
```

* [ ] Come up with a test case that doesn't need to depend on an external package :S 

---

* [X] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).

### DIFF
--- a/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
+++ b/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
@@ -1,0 +1,12 @@
+# cabal v2-update
+Downloading the latest package list from test-local-repo
+# cabal v2-run
+Resolving dependencies...
+Build profile: -w ghc-<GHCVER> -O1
+In order, the following will be built:
+ - some-exe-0.0.1.0 (exe:some-exe) (requires build)
+Configuring some-exe-0.0.1.0...
+Preprocessing executable 'some-exe' for some-exe-0.0.1.0..
+Building executable 'some-exe' for some-exe-0.0.1.0..
+Installing executable some-exe in <PATH>
+Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/new-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/sm-x-0.0.1.0-<HASH>/bin is not in the system search path.
--- a/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
+++ b/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
@@ -9,4 +9,4 @@ Configuring some-exe-0.0.1.0...
 Preprocessing executable 'some-exe' for some-exe-0.0.1.0..
 Building executable 'some-exe' for some-exe-0.0.1.0..
 Installing executable some-exe in <PATH>
-Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/new-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/sm-x-0.0.1.0-<HASH>/bin is not in the system search path.
+Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/<PACKAGE>-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/<PACKAGE>-<HASH>/bin is not in the system search path.
--- a/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
+++ b/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
@@ -43,8 +43,8 @@ normalizeOutput nenv =
   . resub "/incoming/new-[0-9]+"
           "/incoming/new-<RAND>"
     -- look for PackageHash directories
-  . resub "/(([A-Za-z0-9]+)(-[A-Za-z0-9\\.]+)*)-[0-9a-f]{4,64}/"
-          "/\\1-<HASH>/"
+  . resub "/(([A-Za-z0-9_]+)(-[A-Za-z0-9\\._]+)*)-[0-9a-f]{4,64}/"
+          "/<PACKAGE>-<HASH>/"
     -- Normalize architecture
   . resub (posixRegexEscape (display (normalizerPlatform nenv))) "<ARCH>"
     -- Some GHC versions are chattier than others
--- a/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
+++ b/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
@@ -30,7 +30,7 @@ normalizeOutput nenv =
     -- This is dumb but I don't feel like pulling in another dep for
     -- string search-replace.  Make sure we do this before backslash
     -- normalization!
-  . resub (posixRegexEscape (normalizerGblTmpDir nenv) ++ "[a-z0-9.-]+") "<GBLTMPDIR>" -- note, after TMPDIR
+  . resub (posixRegexEscape (normalizerGblTmpDir nenv) ++ "[a-z0-9\\.-]+") "<GBLTMPDIR>" -- note, after TMPDIR
   . resub (posixRegexEscape (normalizerRoot nenv)) "<ROOT>/"
   . resub (posixRegexEscape (normalizerTmpDir nenv)) "<TMPDIR>/"
   . appEndo (F.fold (map (Endo . packageIdRegex) (normalizerKnownPackages nenv)))
@@ -39,6 +39,12 @@ normalizeOutput nenv =
     -- Apply this before packageIdRegex, otherwise this regex doesn't match.
   . resub "[0-9]+(\\.[0-9]+)*/installed-[A-Za-z0-9.+]+"
           "<VERSION>/installed-<HASH>"
+    -- incoming directories in the store
+  . resub "/incoming/new-[0-9]+"
+          "/incoming/new-<RAND>"
+    -- look for PackageHash directories
+  . resub "/(([A-Za-z0-9]+)(-[A-Za-z0-9\\.]+)*)-[0-9a-f]{4,64}/"
+          "/\\1-<HASH>/"
     -- Normalize architecture
   . resub (posixRegexEscape (display (normalizerPlatform nenv))) "<ARCH>"
     -- Some GHC versions are chattier than others
--- a/cabal-install/src/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/src/Distribution/Client/ProjectPlanning.hs
@@ -2680,16 +2680,22 @@ pruneInstallPlanPass1 pkgs =
                 setDocumentation
               $ addOptionalStanzas elab
 
-    find_root (InstallPlan.Configured (PrunedPackage elab _)) =
-        if not $ and [ null (elabConfigureTargets elab)
-                     , null (elabBuildTargets elab)
-                     , null (elabTestTargets elab)
-                     , null (elabBenchTargets elab)
-                     , isNothing (elabReplTarget elab)
-                     , null (elabHaddockTargets elab)
-                     ]
-            then Just (installedUnitId elab)
-            else Nothing
+    is_root :: PrunedPackage -> Maybe UnitId
+    is_root (PrunedPackage elab _) =
+      if not $ and [ null (elabConfigureTargets elab)
+                   , null (elabBuildTargets elab)
+                   , null (elabTestTargets elab)
+                   , null (elabBenchTargets elab)
+                   , isNothing (elabReplTarget elab)
+                   , null (elabHaddockTargets elab)
+                   ]
+          then Just (installedUnitId elab)
+          else Nothing
+
+    find_root (InstallPlan.Configured pkg) = is_root pkg
+    -- When using the extra-packages stanza we need to
+    -- look at installed packages as well.
+    find_root (InstallPlan.Installed pkg)  = is_root pkg
     find_root _ = Nothing
 
     -- Note [Sticky enabled testsuites]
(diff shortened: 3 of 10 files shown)


phadej:
For tests, `run --dry-run` should already do something. There are tools (e.g. `alex`, `happy`) which are buildable with ~any GHC version we ever test with.

phadej:
I also don't see changelog file. Did you forget to `git add` it?

alexbiehl:
I am having troubles with the patch. The integration test wants me to exactly specify the unit id of the happy package. But that's pretty much environment dependent. Need to investigate more. And maybe I am blind but is there maybe simpler test suite already that let's me run cabal and have a fixture for the expected output?

phadej:
Look at `cabal-testsuite` tests. There are plenty. Remember to add `cabal.project` file (otherwise this repositories top-level is picked).

alexbiehl:
I am seeing 
```
[__0] unknown package: alex (user goal)
[__0] fail (backjumping, conflict set: alex)
After searching the rest of the dependency tree exhaustively, these were the goals I've had most trouble fulfilling: alex (1)
``` 

with my test case. Are we executing tests with an index?

phadej:
@alexbiehl looks like so. Grep for `withRepo` for examples of tests with package repositories.

phadej:
Hmm, tests pass because

```
PackageTests/ExtraPackages/cabal.test.hs                                                                          SKIP (0.02s)
```

you actually need to add alex-... package to `repo/` directory.

I have to fix #6820 and install `hackage-repo-tool` onto CI runners.

alexbiehl:
@phadej Do you think we can move this patch forward? I don't think we should wait for the test suite to be refactored to get this change in. I can drop the failed attempt at tests for now. It would be beneficial to have this patch in the upcoming cabal-install release.

phadej:
Sorry. I have experienced too many times that tests are not written retroactively.

alexbiehl:
Alright, do we have a clear path for a test? It seems that my attempts resulted in the tests being skipped as `withRepo` doesn't seem to work properly on the test machines?

cabal-testsuite/PackageTests/ExtraPackages/cabal.out
```diff
@@ -0,0 +1,12 @@
...
+# cabal v2-run
+Resolving dependencies...
+Build profile: -w ghc-<GHCVER> -O1
+In order, the following will be built:
+ - some-exe-0.0.1.0 (exe:some-exe) (requires build)
+Configuring some-exe-0.0.1.0...
+Preprocessing executable 'some-exe' for some-exe-0.0.1.0..
+Building executable 'some-exe' for some-exe-0.0.1.0..
+Installing executable some-exe in <PATH>
+Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/new-52499<ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/sm-x-0.0.1.0-b845c0b5/bin is not in the system search path.
```
alexbiehl, line 12: Arggl this line is failing the test because of how package directories are being shortened on OSX vs. Linux. @phadej any idea where this monstrous path is coming from in the first place?

alexbiehl: Ah, I guess it's this note in ProjectBuilding.hs

```
            -- Note that the copy command has put the files into
            -- @$tmpDir/$prefix@ so we need to return this dir so
            -- the store knows which dir will be the final store entry.
```

phadej: From `CmdInstall`, we warn if people try to install stuff somewhere which is not in PATH. That is reasonable.

In tests we obviously don't use global paths.

The current diff though says you are using `v2-run` so I'm confused. (force pushes confuse me, better to just push WIP commits until the final squash).

alexbiehl: I guess that's how extra-packages work: You tell cabal-install to run an exe from an extra-package and it will make sure that package is being build and installed before running.

alexbiehl:
Alright! I think tests are working properly now. @phadej mind giving this another look?

cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
```diff
@@ -54,6 +54,10 @@ normalizeOutput nenv =
         else id)
   -- hackage-security locks occur non-deterministically
   . resub "(Released|Acquired|Waiting) .*hackage-security-lock\n" ""
+  -- Install path frequently has architecture specific elements, so
+  -- nub it out
+  . resub "^Warning: The directory (.+) is not in the system search path."
+          "Warning: The directory <PATH> is not in the system search path."
```
phadej, line 60: I think this hides a bug which you don't address.

Why `extra-packages` is build as local package, and run from somewhere inside `dist-newstyle` and not installed to the store and run from there.

Or so it looked like in previous CI errors.

I'm :-1: on this change. It shouldn't be required.

phadej (changes requested):

cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
alexbiehl: > Why extra-packages is build as local package, and run from somewhere inside dist-newstyle and not installed to the store and run from there.

Hm, that is not what's happening. `extra-packages` are not treated as local. extra-package are installed globally into the store which displays the store-path in above message. Unfortunately store-paths are platform dependent, hence the above change.

phadej: I think we should rather teach the store-dir location and replace that with `<STOREDIR>` (as we already do for other platform dependent things), rather then hiding big parts of the output

phadej: So if path is 

```
.../store/ghc-8.6.5/Agda-2.6.2-18fb8b49b1f47de3065c0ae3a191d1060c3959e9293aeb9e2aa88c3705c7a92d/bin/agda
```

we know how to replace `ghc-8.6.5`, matching against `/package-id-hash/` and replacing just `hash` part with `<HASH>` should work. An approximation of package-id is not tricky. See `unqual-name` in https://cabal.readthedocs.io/en/3.4/buildinfo-fields-reference.html#non-terminals

alexbiehl: Right, for other store path occurrences we are using the same patterns: 

```
    -- Install path frequently has architecture specific elements, so
    -- nub it out
  . resub "Installing (.+) in .+" "Installing \\1 in <PATH>"
```

phadej:
I tried this patch locally, and it works.  don't like what we print

```
Installing executable cabal-fmt in /cabal/store/ghc-8.6.5/incoming/new-52431/cabal/store/ghc-8.6.5/cabal-fmt-0.1.2-e-cabal-fmt-c0266ba350e909c32b666048e3fc6b2c0db976621f3da582284e8b390d97c14b/bin
Warning: The directory
/cabal/store/ghc-8.6.5/incoming/new-52431/cabal/store/ghc-8.6.5/cabal-fmt-0.1.2-e-cabal-fmt-c0266ba350e909c32b666048e3fc6b2c0db976621f3da582284e8b390d97c14b/bin
is not in the system search path.
```

as that *incoming* path shouldn't be printed. The command is run from `/cabal/store/ghc-8.6.5/cabal-fmt-0.1.2-e-cabal-fmt-c0266ba350e909c32b666048e3fc6b2c0db976621f3da582284e8b390d97c14b/bin/cabal-fmt`

which is more sensible path (btw it might make sense to recognise just `unqual-name-hash`, as looks like we make fancy unit-ids, sorry I forgot about that :) )

That "not in the system path" comes from `Distribution.Simple.Install` and is irrelevant for the users. *Luckily* it's printed only on higher verbosity levels, but it's still confusing.

phadej:
```diff
diff --git a/cabal-testsuite/PackageTests/ExtraPackages/cabal.out b/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
index 7820c030c..6584a39a7 100644
--- a/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
+++ b/cabal-testsuite/PackageTests/ExtraPackages/cabal.out
@@ -9,4 +9,4 @@ Configuring some-exe-0.0.1.0...
 Preprocessing executable 'some-exe' for some-exe-0.0.1.0..
 Building executable 'some-exe' for some-exe-0.0.1.0..
 Installing executable some-exe in <PATH>
-Warning: The directory <PATH> is not in the system search path.
+Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/new-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/some-<HASH>/bin is not in the system search path.
diff --git a/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs b/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
index e4bb63f61..d16567f5c 100644
--- a/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
+++ b/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs
@@ -30,7 +30,7 @@ normalizeOutput nenv =
     -- This is dumb but I don't feel like pulling in another dep for
     -- string search-replace.  Make sure we do this before backslash
     -- normalization!
-  . resub (posixRegexEscape (normalizerGblTmpDir nenv) ++ "[a-z0-9.-]+") "<GBLTMPDIR>" -- note, after TMPDIR
+  . resub (posixRegexEscape (normalizerGblTmpDir nenv) ++ "[a-z0-9\\.-]+") "<GBLTMPDIR>" -- note, after TMPDIR
   . resub (posixRegexEscape (normalizerRoot nenv)) "<ROOT>/"
   . resub (posixRegexEscape (normalizerTmpDir nenv)) "<TMPDIR>/"
   . appEndo (F.fold (map (Endo . packageIdRegex) (normalizerKnownPackages nenv)))
@@ -39,6 +39,12 @@ normalizeOutput nenv =
     -- Apply this before packageIdRegex, otherwise this regex doesn't match.
   . resub "[0-9]+(\\.[0-9]+)*/installed-[A-Za-z0-9.+]+"
           "<VERSION>/installed-<HASH>"
+    -- incoming directories in store
+  . resub "/incoming/new-[0-9]+"
+          "/incoming/new-<RAND>"
+    -- look for PackageHash directories
+  . resub "/([A-Za-z0-9]+)(-[A-Za-z0-9\\.]+)*-[0-9a-f]{4,64}/"
+          "/\\1-<HASH>/"
     -- Normalize architecture
   . resub (posixRegexEscape (display (normalizerPlatform nenv))) "<ARCH>"
     -- Some GHC versions are chattier than others
@@ -54,10 +60,6 @@ normalizeOutput nenv =
         else id)
   -- hackage-security locks occur non-deterministically
   . resub "(Released|Acquired|Waiting) .*hackage-security-lock\n" ""
-  -- Install path frequently has architecture specific elements, so
-  -- nub it out
-  . resub "^Warning: The directory (.+) is not in the system search path."
-          "Warning: The directory <PATH> is not in the system search path."
   where
     packageIdRegex pid =
         resub (posixRegexEscape (display pid) ++ "(-[A-Za-z0-9.-]+)?")
```

phadej:
I'm not sure whether `new-[0-9]+` is good enough for all architectures, but I hope it is.

alexbiehl:
Cool! Unfortunately we have to escape the package name as well as it differs on Linux and OSX: 

On my mac machine it is

```
sm-x-0.0.1.0-<HASH>/bin
```

On CI it is 

```
some-exe-0.0.1.0-HASH/bin
```

cabal-testsuite/PackageTests/ExtraPackages/cabal.out
```diff
@@ -9,4 +9,4 @@ Configuring some-exe-0.0.1.0...
 Preprocessing executable 'some-exe' for some-exe-0.0.1.0..
 Building executable 'some-exe' for some-exe-0.0.1.0..
 Installing executable some-exe in <PATH>
-Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/new-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/sm-x-0.0.1.0-<HASH>/bin is not in the system search path.
+Warning: The directory <ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/incoming/<PACKAGE>-<HASH><ROOT>/cabal.dist/home/.cabal/store/ghc-<GHCVER>/<PACKAGE>-<HASH>/bin is not in the system search path.
```
alexbiehl, line 12: Argl it escapes too mucho :(

phadej:
hmm. that `some-exe` vs `sm-x` is tricky. Let me think what can be done there.

alexbiehl:
What about trimming away the whole path in that log message? Its a lot easier than surgically replacing butchered package names..

phadej:
> What about trimming away the whole path in that log message? Its a lot easier than surgically replacing butchered package names.

No. I said it already. We want to catch cases where we e.g. reorganise something.

phadej:
I made more tweaks to OutputNormalizer, and got tests to be green in https://github.com/haskell/cabal/pull/7099

PR #6980, opened 2020-07-21
Add cabal-version: 3.6 values

phadej:


phadej:
@jhrcek not really. `hackage-server` accepts `3.0` now, and even then it doesn't accept public libraries, i.e. there is little point to `init` with newer versions. 

Or maybe there is, but that's `init` "owner", i.e. @m-renaud call.

PR #6982, opened 2020-07-21
More warnings

phadej:


phadej:
Backported in https://github.com/haskell/cabal/pull/6987

PR #6984, opened 2020-07-22
Generate Paths module using zinza template

phadej:
This breaks `absolute` and `relocatable` variants. Let me see if we have any tests for them (maybe on windows).

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -603,6 +603,7 @@ library
     Distribution.Lex
     Distribution.Utils.String
     Distribution.Simple.Build.Macros.Z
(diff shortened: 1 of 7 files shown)


angerman:
O_o

phadej:
@angerman 

> O_o

?_?

phadej:
This is a step cleaninh up the `Paths_<pkg>` module implementation, so it can be changed. Concatenating strings manually is not 2020.

angerman:
I only read
> This breaks `absolute` and `relocatable` variants.

and was confused. I still believe the Paths module should not exist, but I'm hard pressed to offer a sensible alterntive.

PR #6979, opened 2020-07-20
Bootstrap rewrite by bgamari

phadej:
Continuation to #6700. cc @bgamari.

PR #6985, opened 2020-07-22
Bump random to 1.2

phadej:


phadej:
Seems that I broke bootstrap CI, as `splitmix` is needed for `random`. The #6979 will fix that though.

phadej:
Backported in https://github.com/haskell/cabal/pull/6987

Issue #6986, opened 2020-07-22
Cabal requires tighter bounds on process on windows.


AndreasPK:
To my knowledge all but the latest version of process (1.6.10.0) run the risk of producing segfaults on windows.

**To Reproduce**

I can reproduce this reliably by building cabal-install with ghc-8.10.1.

The resulting cabal binary will then segfault on most operations.

**Expected behavior**

**System information**
 - Windows
 - Cabal-head, ghc-8.10.1

phadej:
Is `1.6.10.0`  required, the `1.6.9.0` bundled with GHC-8.8.4 is not enough?

AndreasPK:
I'm not sure. @Mistuke do you know if the fix is also in 1.6.9?

jneira:
Wow, i did not know that executables compiled against process lib versions with the bug were causing segfault *at runtime*.
I thought that was [ghc itself that was throwing the segfaults](https://gitlab.haskell.org/ghc/ghc/-/issues/17926) when compiling packages. So thery should be solved using a proper ghc version (ghc-8.8.4 or ghc-8.10.2) indpendently of the build tool (cabal or stack) and the ghc used to compile that build tool.

Mistuke:
The fix is in 1.6.9 https://github.com/haskell/process/blob/master/changelog.md#169-may-2020 1.6.10 has an unrelated fix for pipes.

@jneira GHC 8.10.1 is broken as well. 8.10.1 was released with same process as 8.8.3.

Mistuke:
That said, Cabal currently deals with this with a compile time check https://github.com/haskell/cabal/blob/95c5e40116e178b6f003c9d464a49ea16268e2df/Cabal/src/Distribution/Compat/Process.hs#L33 but it's checking for `1.6.8` since we thought that fixed the issue.

Without process jobs cabal would be intermittently unreliable but still work. We thought at the time it would be a nicer trade-off.

jneira:
@mistuke
> GHC 8.10.1 is broken as well. 8.10.1 was released with same process as 8.8.3.

Thanks for the clarification.

> That said, Cabal currently deals with this with a compile time check 

Then maybe we should update the process version in the cpp check to 1.6.9, instead change process bounds (that maybe would break building cabal with older ghc versions?)

AndreasPK:
There is this comment surrounding the CPP:

```
-- | Enable process jobs to ensure accurate determination of process completion
-- in the presence of @exec(3)@ on Windows.
--
[...]

PR #7004, opened 2020-08-04
Bump process version required for job support to 1.6.9

AndreasPK:
The CPP used to enable it even for version 1.6.8 which
turned out to still be buggy which causes segfaults when
building Cabal with ghc-8.10. See #6986.

This does not fully fix the issue as without job support
build failures can still sporadically occur.

---

Haven't tested it locally yet but will do so.

### DIFF
--- a/Cabal/src/Distribution/Compat/Process.hs
+++ b/Cabal/src/Distribution/Compat/Process.hs
@@ -26,11 +26,11 @@ import           System.Process (waitForProcess)
 -- in the presence of @exec(3)@ on Windows.
 --
 -- Unfortunately the process job support is badly broken in @process@ releases
--- prior to 1.6.8, so we disable it in these versions, despite the fact that
+-- prior to 1.6.9, so we disable it in these versions, despite the fact that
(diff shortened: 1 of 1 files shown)


AndreasPK:
Seems to fix the segfaults for me locally at least.

jneira:
@AndreasPK do you mind to describe briefly the error you got without this patch? Are they similar to the described in this issue: https://github.com/ennocramer/floskell/issues/50?

AndreasPK:
> 
> 
> @AndreasPK do you mind to describe briefly the error you got without this patch? Are they similar to the described in this issue: [ennocramer/floskell#50](https://github.com/ennocramer/floskell/issues/50)?

I added an example to the [ticket here](https://github.com/haskell/cabal/issues/6986#issuecomment-668915278)

It's similar as in also a segfault. But it's cabal itself that is segfaulting. But ghc-8.8.3 is known to suffer from a similar issue.

AndreasPK:
Anything left to do on my side to move this along?

phadej:
Backported in https://github.com/haskell/cabal/pull/7026

PR #7009, opened 2020-08-09
Provide license migration message for LGPL-2.0

hasufell:


Cabal/src/Distribution/SPDX/LicenseId.hs
```diff
@@ -497,14 +497,15 @@ licenseIdMigrationMessage = go where
     go l | gnuVariant l    = "SPDX license list 3.0 deprecated suffixless variants of GNU family of licenses. Use " ++ l ++ "-only or " ++ l ++ "-or-later."
     go "BSD3"              = "Do you mean BSD-3-Clause?"
     go "BSD2"              = "Do you mean BSD-2-Clause?"
+    go "LGPL-2"            = "Do you mean LGPL-2.0-only?"
```
phadej, line 500:
```
LGPL-2.1-only or LGPL-3.0-only or -or-later variants
```

for old formats cabal says

```
Warning: These warnings may cause trouble when distributing the package:
Warning: 'license: LGPL-2' is not a known version of that license. The known
versions are 2.1, 3. If this is not a mistake and you think it should be a
known version then please file a ticket.
```

Cabal/src/Distribution/SPDX/LicenseId.hs
```diff
@@ -497,14 +497,15 @@ licenseIdMigrationMessage = go where
...
     go "AllRightsReserved" = "You can use NONE as a value of license field."
     go "OtherLicense"      = "SPDX license list contains plenty of licenses. See https://spdx.org/licenses/. Also they can be combined into complex expressions with AND and OR."
     go "PublicDomain"      = "Public Domain is a complex matter. See https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files. Consider using a proper license."
 
     -- otherwise, we don't know
     go _ = ""
 
-    gnuVariant = flip elem ["GPL-2.0", "GPL-3.0", "LGPL-2.1", "LGPL-3.0", "AGPL-3.0" ]
+    gnuVariant = flip elem ["GPL-2.0", "GPL-3.0", "LGPL-2.0", "LGPL-2.1", "LGPL-3.0", "AGPL-3.0" ]
```
phadej, line 508: undo

hasufell: Why?

Cabal/src/Distribution/SPDX/LicenseId.hs
hasufell: can't really follow what a known or unknown version is. Is it not known per this? https://github.com/haskell/cabal/blob/master/Cabal/src/Distribution/SPDX/LicenseId.hs#L237

Also, why would we mention `LGPL-3.0` when the user typed `LGPL-2`? We don't do that for `BSD2` either.

hasufell:
Even after regenerating the template etc, it still throws:

```hs
cabal: Error parsing config file /home/jule/.cabal/config:143:
Parse of field 'license' failed (Cannot parse license: LGPL-2): LGPL-2
```

I looked at the parser code and cannot see why this would be an invalid parse. It seems something is messed up wrt parsing vs validating.

phadej:
Ah, so this is about `init` section in `$HOME/cabal/config`?

It looks like `cabal-install-3.2` accepted whatever garbage here. I'm not sure how it's parsed in 3.4.0.0 now.

cc @m-renaud

phadej:
I.e. the migration message is not shown anyway, even with known values:

```
% cabal user-config diff
cabal: Error parsing config file /cabal/config:34:
Parse of field 'license' failed (Cannot parse license: GPL-3): GPL-3
```

PR #7015, opened 2020-08-20
Data.ByteString.Lazy.Builder -> Data.ByteString.Builder

bgamari:
The former was deprecated in 0.10.10.

---

### DIFF
--- a/Cabal/src/Distribution/Utils/Structured.hs
+++ b/Cabal/src/Distribution/Utils/Structured.hs
@@ -87,7 +87,11 @@ import GHC.Generics
 
 import qualified Data.ByteString              as BS
 import qualified Data.ByteString.Lazy         as LBS
+#if MIN_VERSION_bytestring(0,10,6)
+import qualified Data.ByteString.Builder      as Builder
+#else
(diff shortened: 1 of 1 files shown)


phadej:
Why introduce CPP, is there some `-Werror` annoyance somewhere?

bgamari:
Because the module is not present in earlier versions of `bytestring`.

phadej:
Yes, I'm wondering what is the problem with using old name for as long as it exists?

bgamari:
> Yes, I'm wondering what is the problem with using old name for as long as it exists?

The module is marked as deprecated and consequently `Cabal` is not `-Werror` clean with `bytestring-0.10.12.0`.

phadej:
Fun fact. PVP says (atm)

> 7. Deprecation. Deprecated entities (via a DEPRECATED pragma) SHOULD be counted as removed for the purposes of upgrading the API, because packages that use -Werror will be broken by the deprecation. In other words the new A.B SHOULD be greater than the previous A.B.

TL;DR, don't deprecate, just remove straight away.

phadej:
Edited and merged in https://github.com/haskell/cabal/pull/7049

PR #7024, opened 2020-09-02
Use optparse-applicative-0.16

phadej:


phadej:
Backported in https://github.com/haskell/cabal/pull/7026

PR #7030, opened 2020-09-03
Fix #7028. Use cwd when looking for data-files in sdist

### LINKED ISSUE #7028: cabal sdist fails in packages with data-files
To reproduce create a project,
add a package with data-files, *importantly* not at the root of project (e.g. inside directory).

`cabal sdist` will fail to find `data-files`.

phadej:


phadej:
Backported in https://github.com/haskell/cabal/pull/7042

PR #7049, opened 2020-09-10
Data.ByteString.Lazy.Builder -> Data.ByteString.Builder

phadej:
Like #7015, but rebased and lowered the cut bound to `0.10.4` (which ships with GHC-7.8)


phadej:
Backported in https://github.com/haskell/cabal/pull/7052

PR #7053, opened 2020-09-10
Add windows GHC-8.10.2 job

phadej:
I'm not sure whether I want GHC-8.10.2 to violate access (like 8.8.4) or not. But lets have this job anyway.


phadej:
So GHC-8.10.2 is busted. Have to wait until GHC-8.10.3.

jneira:
Yeah the info about ld in the ghc settings is incorrect: there is a [relatively easy workaround](https://gitlab.haskell.org/ghc/ghc/-/issues/18550#note_293297) to make it work locally but agree it is better to wait to 8.10.3

Issue #6421, opened 2019-12-12
GHC-8.10 has -optcxx, should Cabal do something with that?


phadej:

phadej:
ping @cblp

cblp:
I'm not sure this belongs to Cabal or cabal-install, but I will write it down anyway for somebody competent:

1. `cxx-options` should be passed to `-optcxx`
2. `cc-options` should be passed to `-optc`
3. `c-sources` should use `cc-options` but not `cxx-options`
4. `cxx-sources` should use `cxx-options` but not `cc-options`

hvr:
We also need to come up with well-defined `cabal-version`-guards to make sure the semantics well-defined for a given `.cabal` file and don't fluctuate based on which version of `cabal` you're building it with...

phadej:
`cxx-sources` and `cxx-options` are since `cabal-version: 2.2`.

My feeling is that we should make this too complicated, if there's a package which really needs
the separation as in https://github.com/haskell/cabal/issues/6421#issuecomment-564968409,
that it ought to require newer `cabal-version: ...`, but I don't think we need to simulate older behaviour in newer `Cabal`s.

In other words: What would break if above separation is changed. My gut feeling is that nothing. (at least on Hackage). Or more bluntly, whatever is on Hackage and using `cxx-sources` now is broken anyways.

bgamari:
This came up in [GHC #18687](https://gitlab.haskell.org/ghc/ghc/-/issues/18687).

ejconlon:
Hi folks, I reported the linked GHC ticket and included a simple [repro](https://github.com/ejconlon/ffi-repro). I agree with @phadej above that we probably don't have to overcomplicate the fix with advanced `cabal-version` checking. As far as I can tell it's impossible now to pass options to C++ builds through ANY means (`cc-options`, `cxx-options`, `ghc-options`) on 8.10. I believe that if we change this behavior nothing would break that isn't already broken now.

ejconlon:
AFAICT, this line should be `-optcxx` for 8.10 and above: https://github.com/haskell/cabal/blob/90c4695667ccf9281ab570b76a1c9241216c9e9a/Cabal/src/Distribution/Simple/Program/GHC.hs#L689 .

PR #7072, opened 2020-09-14
Pass -optcxx for GHC >= 8.10

ejconlon:
Intends to resolve [this issue](https://github.com/haskell/cabal/issues/6421). To summarize, cabal passes all C and C++ flags through GHC to the underlying C or C++ compiler using `-optc`. This works for GHC < 8.10, but now GHC expects C++ flags to come through `-optcxx`. This means that anything through `-optc` is ignored, so we cannot pass any flags to the C++ compiler. This change simply detects the GHC version and uses the correct arguments.

This PR has been tested manually on this [repro](https://github.com/ejconlon/ffi-repro) and two `PackageTests` have been added to `cabal-testsuite`. They pass under GHC 8.8.4 and GHC 8.10.2.

### DIFF
--- a/Cabal/src/Distribution/Simple/Program/GHC.hs
+++ b/Cabal/src/Distribution/Simple/Program/GHC.hs
@@ -686,7 +686,11 @@ renderGhcOptions comp _platform@(Platform _arch os) opts
   , concat [ [ "-optP-include", "-optP" ++ inc]
            | inc <- flags ghcOptCppIncludes ]
   , [ "-optc" ++ opt | opt <- ghcOptCcOptions opts]
-  , [ "-optc" ++ opt | opt <- ghcOptCxxOptions opts]
+  , -- C++ compiler options: GHC >= 8.10 requires -optcxx, older requires -optc
+    let cxxflag = case compilerCompatVersion GHC comp of
+                Just v | v >= mkVersion [8, 10] -> "-optcxx"
+                _ -> "-optc"
+    in [ cxxflag ++ opt | opt <- ghcOptCxxOptions opts]
   , [ "-opta" ++ opt | opt <- ghcOptAsmOptions opts]
 
   -----------------
--- a/cabal-testsuite/PackageTests/FFI/ForeignOptsC/Main.hs
+++ b/cabal-testsuite/PackageTests/FFI/ForeignOptsC/Main.hs
@@ -1,0 +1,16 @@
+{-# LANGUAGE ForeignFunctionInterface #-}
+
+module Main where
+
+import Foreign.C (CInt (..))
+
+foreign import ccall "clib.h meaning_of_life_c"
+  meaning_of_life_c :: IO CInt
+
+main :: IO ()
+main = do
+    secret <- meaning_of_life_c
+    -- The value 11 comes from __TESTOPT_C__ - see the cabal file.
+    if (secret == 11)
+        then putStrLn ("The secret is " ++ show secret)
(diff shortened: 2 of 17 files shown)


phadej (changes requested):
We need tests.

Also "negative" tests, that cxx-options are not passed to C-files, and C-options to C++-files as mentioned in a comment in the issue. (For GHC >=8.10).

ejconlon:
Yep, let me see what I can do about tests. In the meantime, I verified that this fixes the issue as reproduced [here](https://github.com/ejconlon/ffi-repro) with GHC 8.8.4 and 8.10.2.

ejconlon:
@phadej would https://github.com/haskell/cabal/tree/master/cabal-testsuite/PackageTests be the right place to put test cases? I haven't found a suitable unit test suite for this, so I think the integration tests there are appropriate.

If so, I will put in two examples lifted from the repro, both configured with `cc-options: -D__TESTOPT_C__` and `cxx-options: -D__TESTOPT_CXX__`:
1. A project including C sources that will `#error` if `__TESTOPT_CXX__` is defined or `__TESTOPT_C__` is NOT defined
2. A project including C++ sources doing the opposite

Is that comprehensive enough?

phadej:
@ejconlon yes, `PackageTests` is the place for such things.

ejconlon:
@phadej I put in two `PackageTests`: `ForeignOptsC` and `ForeignOptsCxx`. Assuming I ran them right, they both pass for me under both compilers with this change. Both of them pass `cc-options` and `cxx-options` and assert that they only pick up CPP flags from the correct one. If we are running them with 8.8 and 8.10 in the CI matrix, then I think we are in good shape.

ejconlon:
(Also, if you're planning on doing another RC for 3.4 I would love to get this change picked in!)

phadej:
Perfectly we would backport that to `Cabal-3.2` (shipped with GHC-8.10), as far as I'm aware GHC-8.10.3 will definitely happen.

@bgamari should I prepare `Cabal-3.2.1.0`?

cabal-testsuite/PackageTests/ForeignOptsC/foreign-opts-c.cabal
```diff
@@ -0,0 +1,14 @@
+cabal-version:       2.2
```
phadej, line 1: make a top directory for these tests, maybe `PackageTests/FFI/ForeignOptsC` and `..Cxx`. As you may notice, the top level is already quite crowded, lets not have individual tests there.

phadej (changes requested):
Squash commits, that will help backporting.

Write in the commit message what you have written in the description. Also write

```
Fixes https://github.com/haskell/cabal/issues/6421
```

(I prefer explicit URL, not an abbreviated number, as it makes easy to click from `git log`).

ejconlon:
(By the way, this is the second commit for which that Windows job flaked out with some handle error!)

bgamari:
> Perfectly we would backport that to `Cabal-3.2` (shipped with GHC-8.10), as far as I'm aware GHC-8.10.3 will definitely happen.
> 
> @bgamari should I prepare `Cabal-3.2.1.0`?

Yes, that would be great. Thank you1

phadej:
@ejconlon yes, windows jobs are flaky. Don't worry.

phadej:
Backported to 3.4 in https://github.com/haskell/cabal/pull/7074

phadej:

PR #7076, opened 2020-09-17
Disable RPath usage on macOS

angerman:
This interferes with dead_strip_dylib, and pollutes the load commands,
which have a size limit in recent macOS versions.  We now rely on GHC
to construct the correct -rpath entries *after* the linking phase by
inspecting the produced dynamic product and adding -rpaths for all
non-dead-stripped referenced libraries.

See https://gitlab.haskell.org/ghc/ghc/-/commit/4ff93292243888545da452ea4d4c1987f2343591
or  ghc/ghc@4ff9329

---

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -1726,7 +1726,21 @@ getRPaths lbi clbi | supportRPaths hostOS = do
     -- 'False', while those operating systems themselves do support RPATH.
     supportRPaths Linux       = True
     supportRPaths Windows     = False
-    supportRPaths OSX         = True
+    -- While macOS fundamentally supports RPaths, this flag will produce
+    -- linker -rpath flags for each dependency and pass them to GHC which will
+    -- in turn pass them verbatim to ld64.  This is not desirable as we need to
+    -- use dead_strip_dylibs on macOS to prevent the Load Commands to exceed the
+    -- Load Command Size Limit (32K) on macOS.  In GHC this was reworked in
+    -- ghc/ghc:4ff93292 (https://gitlab.haskell.org/ghc/ghc/-/commit/4ff93292243888545da452ea4d4c1987f2343591)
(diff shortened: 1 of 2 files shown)


phadej (changes requested):
I see no tests.

What happens with GHC-8.6 and older?

Cabal/src/Distribution/Simple/GHC.hs
```diff
@@ -1726,7 +1726,21 @@ getRPaths lbi clbi | supportRPaths hostOS = do
...
+    -- Load Command Size Limit (32K) on macOS.  In GHC this was reworked in
+    -- ghc/ghc:4ff93292 (https://gitlab.haskell.org/ghc/ghc/-/commit/4ff93292243888545da452ea4d4c1987f2343591)
+    -- to patch the -rpath's into the library *after* linking them
+    -- with dead_strip_dylib. GHC relies on proper passing of -L values to find
+    -- the referened lirbaries and subsequently produce the correct -rpath entries
+    -- in the library.  This will be backported to GHC 8.8, 8.10, and 9.0, as
+    -- well as being part of GHC 9.2 onwards.
+    --
+    -- We know the GHC in question supports this if it knows about
+    -- `install_name_tool` or `otool`.
```
phadej, line 1742: This suggests that there should some check of `install_name_tool`?

angerman:
> I see no tests.

Do you have suggestions how to test this properly? Effectively we strip cabal from passing `-rpath` values to `ghc`, and `ghc` will then do some post-link processing to inject the necessary rpaths.
 
> What happens with GHC-8.6 and older?

Won't get this feature. Hence cabal should pass `-rpath` values for older ghcs, but in general we should try to detect this based on the ghc at hand, in case someone decides to backport this to their favorit ghc.

Cabal/src/Distribution/Simple/GHC.hs
angerman: Yes. Need to figure out how.

PR #7080, opened 2020-09-18
Select buildable targets

tomjaguarpaw:
I found the `selectBuildableTargets...` functions a little bit harder to read than I think they should have been. The comprehensions they contain are all basically the same, so I factored out the commonality.  I think it is a modest readability improvement. What do you think?


tomjaguarpaw:
I know this PR seems trivial but small quality of life improvements really make it much easier for new developers to start contributing. The cabal codebase is enormous and complex and I'd like to make small incremental changes to make it less complex where I find them.

Mikolaj:
Is there a test that would batch a disastrous bug in this refactoring? E.g., if you return always the empty list, would any existing test break?

Mikolaj:
I suspect @tomjaguarpaw had tests fail when developing this, so a confirmation would satisfy me. But if @tomjaguarpaw got everything right the fist time, then indeed, there is a cost to pay for such perfection. ;)

tomjaguarpaw:
I'm not sure what you mean. I didn't have any tests fail because I didn't run the tests. I just let GitHub CI run the tests. They take hours so I don't want to run them on my machine!

As you can see, the CI tests pass. If there aren't already tests that would catch a bug in this refactoring then cabal has even bigger problems than I realised.

Mikolaj:
Excuse me being obnoxious, but I asked why you think we have enough tests to catch big problems in this code, not why you think your code fine (I think it's fine, too, but we needs that keep it being fine against changes by others or change of context).

An answer that would satisfy me is, e.g.: this code appears in each code path for `cabal build` so can't be side-stepped there and we have tests for `cabal build`. But I don't want to make up answer, while it's you who knows, or can know, this code and its context better, having just worked on it.

Mikolaj:
Another answer that would satisfy me: "I've changed the code to always return empty list and some tests failed".

tomjaguarpaw:
I'm not sure exactly what you're asking from me.  I offer this PR because I think it improves the codebase.  If you think it is insufficiently tested to merge then I respect your decision. If you have specific actions you would like me to take to validate this PR then please let me know and I will endeavour to take them. (I am not familiar with the codebase -- indeed I find it overwhelming -- so I would appreciate actions specified as clearly as possible).

tomjaguarpaw:
To be clear, I don't find that you are obnoxious but I do not understand what you are asking from me.

Mikolaj:
And one more clarification: you do know the concept of test coverage, right? Here's the bomb: we don't know what test coverage of cabal codebase is (PRs to plug cabal codebase into a coverage checker service or test are welcome). So, we at least strive to establish for each changed bit of code if it's covered or not and, if not, cover it.

tomjaguarpaw:
I can confirm that, running `cabal test` to get the `cabal-install` tests:

* My commit has a single test failure (git fetch failure). I guess this is something to do with the configuration of my machine because all the tests pass in CI.

* My commit with the following patch applied additionally fails the following tests
  * problems (build)
  * problems (repl)
  * problems (run)
  * problems (haddock)

so I believe there is sufficient test coverage for this change.

(The tests take just over three minutes, which is quite reasonable)

```diff
--- a/cabal-install/src/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/src/Distribution/Client/ProjectOrchestration.hs
@@ -767,8 +767,8 @@ selectBuildableTargets = selectBuildableTargetsWith (const True)
 
 zipBuildableTargetsWith :: (TargetRequested -> Bool)
                         -> [AvailableTarget k] -> [(k, AvailableTarget k)]
-zipBuildableTargetsWith p ts =
-    [ (k, t) | t@(AvailableTarget _ _ (TargetBuildable k req) _) <- ts, p req ]
+zipBuildableTargetsWith _ _ =
+    []
 
```

tomjaguarpaw:
> And to clarify some more ...

Thank you for clarifying. Your rationale seems very reasonable.

Mikolaj:
OOI, how do tests fail for you? Perhaps something should be fixed in these tests --- we want testing to be maximally comfortable, to encourage testing and adding tests.

Mikolaj:
Yes, thank you. Let me try if I get the error as well.

Mikolaj:
CI and mergifyIO does not cooperate today, so let me merge by hand.

Mikolaj:
Actually, `git fetch failure:  fatal: repository '/tmp/repos-2383662/src' does not exist` is expected, but another test is failing. Opened as #7785.

tomjaguarpaw:
Ah yes, that's right! The test suite output is as follows, and because I saw "1 out of 492 test failed" I assumed the one with the error message was the failure. I didn't see the `OK` hiding down there :)

```
      git fetch failure:    fatal: repository '/tmp/repos-299613/src' does not exist
OK
...
1 out of 492 tests failed (29.99s)
```

PR #7094, opened 2020-09-28
[macOS] better rpath logic

angerman:
This is the continuation of #7076

---

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -1740,7 +1740,7 @@ getRPaths lbi clbi | supportRPaths hostOS = do
     -- ghc/ghc:4ff93292 (https://gitlab.haskell.org/ghc/ghc/-/commit/4ff93292243888545da452ea4d4c1987f2343591)
     -- to patch the -rpath's into the library *after* linking them
     -- with dead_strip_dylib. GHC relies on proper passing of -L values to find
-    -- the referened lirbaries and subsequently produce the correct -rpath entries
+    -- the referenced libraries and subsequently produce the correct -rpath entries
     -- in the library.  This will be backported to GHC 8.8, 8.10, and 9.0, as
     -- well as being part of GHC 9.2 onwards.
     --
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -1717,6 +1717,12 @@ getRPaths lbi clbi | supportRPaths hostOS = do
     (Platform _ hostOS) = hostPlatform lbi
     compid              = compilerId . compiler $ lbi
 
+    ghcInfoMap          = compilerProperties . compiler $ lbi
+    hasInstallNameTool  =
+      case Map.lookup "install_name_tool command" ghcInfoMap of
+        Just path -> not (null path)
+        Nothing   -> False
+
     -- The list of RPath-supported operating systems below reflects the
     -- platforms on which Cabal's RPATH handling is tested. It does _NOT_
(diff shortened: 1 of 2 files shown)


chessai:
Restarted the failed windows jobs

chessai:
@phadej @bgamari could one of you review? This actually fixes a pretty annoying bug on MacOS. I can confirm that it works as intended (tested on 15+ dev machines).

phadej:
I'm uneasy on this. I'd like to see the tests, which would break if that code is changed again.

So if we need to wait until GHC 9.2 or 9.0 is released with `install_name_tool command` in the `--info` to write a test, then we will wait. (The patch won't work anyway if you aren't using it with such GHC).

Cabal/src/Distribution/Simple/GHC.hs
```diff
@@ -1726,7 +1732,23 @@ getRPaths lbi clbi | supportRPaths hostOS = do
...
-    supportRPaths OSX         = True
+    -- While macOS fundamentally supports RPaths, this flag will produce
+    -- linker -rpath flags for each dependency and pass them to GHC which will
+    -- in turn pass them verbatim to ld64.  This is not desirable as we need to
+    -- use dead_strip_dylibs on macOS to prevent the Load Commands to exceed the
+    -- Load Command Size Limit (32K) on macOS.  In GHC this was reworked in
+    -- ghc/ghc:4ff93292 (https://gitlab.haskell.org/ghc/ghc/-/commit/4ff93292243888545da452ea4d4c1987f2343591)
+    -- to patch the -rpath's into the library *after* linking them
+    -- with dead_strip_dylib. GHC relies on proper passing of -L values to find
+    -- the referened lirbaries and subsequently produce the correct -rpath entries
```
emilypi, line 1743:
```suggestion
    -- the referenced libraries and subsequently produce the correct -rpath entries
```

emilypi (commented):
Hi @angerman - let's talk about how to write tests for this, and I'd be happy to help see this through.

ulysses4ever:
This is marked as "In progress" for 3.10. Is it?

Kleidukos:
@angerman Yo! Do you need any help with this PR?

andreabedini:
@angerman should we pick this up again? AFAIK `ghc --info` includes `install_name_tool command` since 8.10.7.

alt-romes:
Note: if we decide to do this, we must carefully take into account that GHC will only keep as RPATHs the paths where linked libraries are found at link time. Specifically, some runtime paths should be included in the binary regardless of any library being found there at link time, if the user specifies so (namely `extra-lib-dirs` #7339).

So it is kind of fine if `getRPaths` were to return an empty list for GHCs which workaround the darwin linker issue by post-link-injecting the paths of the libraries found at link time to the RPATH entries, but I'm also not sure to what extent it is necessary, since the list is often not that long.

In fact, `getRPaths` is typically 2 or 3 directories only: the `.cabal/ghc` store where all external dependency libraries are stored. In my case, building a shared library these rpaths are computed:
```
-optl-Wl,-rpath,/Users/romes/.ghcup/ghc/9.8.1/lib/ghc-9.8.1/lib/aarch64-osx-ghc-9.8.1
-optl-Wl,-rpath,/Users/romes/.cabal/store/ghc-9.8.1/lib
-optl-Wl,-rpath,/Library/Developer/CommandLineTools/SDKs/MacOSX14.2.sdk/usr/lib
```
However, with nix stores I think one might get one path per dependency.

The paths determined by `getRPaths/depLibraryPaths` are likely where libraries are found at both linktime *and* runtime, so it should be fine to not pass them on on macOS, I guess.

ffaf1:

Issue #6772, opened 2020-05-10
Choosing `library` in init dialogue still creates `exe` dir


emilypi:
**Describe the bug**
Choosing the "library" option in the interactive init script still produces a `exe` directory

**To Reproduce**
Follow the init steps  for a new library. Example: 

```
λ П(a: A) haskell → mkdir foo
λ П(a: A) haskell → cd foo
λ П(a: A) foo → cabal init
Should I generate a simple project with sensible defaults? [default: y] n
What does the package build:
   1) Executable
   2) Library
   3) Library and Executable
Your choice? 2
Package name? foo
This package name is already used by another package on hackage. Do you want to choose a different name? [default: y] n
Package version? [default: 0.1.0.0] 
Author name? [default: Emily Pillmore] 
Maintainer email? [default: emilypi@cohomolo.gy] 
Project homepage URL? 
Project synopsis? 
Project category:
 * 1) (none)
   2) Codec
   3) Concurrency
   4) Control
   5) Data
   6) Database
   7) Development
   8) Distribution
   9) Game
  10) Graphics
  11) Language
  12) Math
  13) Network
  14) Sound
  15) System
  16) Testing
  17) Text
  18) Web
  19) Other (specify)
Your choice? [default: (none)] 
Add informative comments to each field in the cabal file (y/n)? [default: n] n

Guessing dependencies...

Generating LICENSE...
Generating Setup.hs...
Generating CHANGELOG.md...
Generating src/MyLib.hs...
Generating test/MyLibTest.hs...
Generating foo.cabal...

Warning: no synopsis given. You should edit the .cabal file and add one.
You may want to edit the .cabal file and add a Description field.
λ П(a: A) foo → ls -l
total 32
-rw-r--r--  1 emilypi  staff   106 May 10 01:53 CHANGELOG.md
-rw-r--r--  1 emilypi  staff  1526 May 10 01:53 LICENSE
-rw-r--r--  1 emilypi  staff    46 May 10 01:53 Setup.hs
drwxr-xr-x  2 emilypi  staff    64 May 10 01:53 exe
-rw-r--r--  1 emilypi  staff   854 May 10 01:53 foo.cabal
drwxr-xr-x  3 emilypi  staff    96 May 10 01:53 src
drwxr-xr-x  3 emilypi  staff    96 May 10 01:53 test
λ П(a: A) foo → 
```

**Expected behavior**
Shouldn't do that. 

**System information**
 -  Mac
 - Cabal 3.2
[...]

PR #7103, opened 2020-10-08
Issue #6772 library creates exe

JeremyScott-IO:
This fixes init adding an exe folder if the package type is Library. This is for issue #6772 .

I added an if statement for checking if the PackageType is Library. We can just return Just [] to appDir, and skip other evaluation. I also removed noAppDirIfLibrary as its functionality is now covered by the new if statement.

It's also possible to create a lib directory when source-dir specified after selecting exe package type.

`$ mkdir foo`
`$ cd foo`
`$ cabal init --non-interactive --exe --source-dir=lib`
`$ ls`
`CHANGELOG.md  Main.hs   Setup.hs  lib  test.cabal`

Should I work on that or raise a seperate issue?

Let me know what you think!

### DIFF
--- a/cabal-install/src/Distribution/Client/Init/Command.hs
+++ b/cabal-install/src/Distribution/Client/Init/Command.hs
@@ -468,20 +468,16 @@ getGenComments flags = do
 getAppDir :: InitFlags -> IO InitFlags
 getAppDir flags = do
   appDirs <-
-    return (applicationDirs flags)
-    ?>> noAppDirIfLibraryOnly
-    ?>> guessAppDir flags
-    ?>> promptUserForApplicationDir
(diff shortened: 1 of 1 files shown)


m-renaud:
Thanks for the fix! Could you add to the PR description a couple of example cabal sessions with this change? 

> It's also possible to create a lib directory when source-dir specified after selecting exe package type.... Should I work on that or raise a seperate issue?

A separate PR for that fix would be preferred to keep the individual changes small.

emilypi:
#7262 subsumes this PR. Do you mind if we close @JeremyScott-IO?

gbaz:
I'm closing based on above discussion.

Issue #7106, opened 2020-10-09
Build differences between `v2-run` and `v2-exec` not documented


pnwamk:
**Describe the bug**
Differences between `v2-run` and `v2-exec` w.r.t. build-related behavior is not documented.

**To Reproduce**
Observe that when running:

```
$ mkdir exec-vs-run && cd exec-vs-run
$ cabal init -n
$ cabal v2-build
...
$ cabal v2-exec exec-vs-run
Hello, Haskell!
$ sed -i 's/Hello/Bonjour/g'  Main.hs
$ cabal v2-exec exec-vs-run
Hello, Haskell!
$ cabal v2-run exec-vs-run
...
Bonjour, Haskell!
```

that `v2-exec` will not build or rebuild a project to incorporate any recent changes, whereas `v2-run` will.

**Expected behavior**
Cabal documentation alludes to this difference.

**System information**
 - Operating system: Any? (MacOS in my case)
 - `cabal` version 3.0.0.0, `ghc` version 8.8.3

PR #7107, opened 2020-10-09
Resolve https://github.com/haskell/cabal/issues/7106: document differences between v2-run and v2-exec

pnwamk:
Adds a note in the docs about build-related differences between `v2-build` and `v2-run`; see https://github.com/haskell/cabal/issues/7106.

### DIFF
--- a/doc/cabal-commands.rst
+++ b/doc/cabal-commands.rst
@@ -173,10 +173,11 @@ cannot be excluded for technical reasons).
 cabal v2-run
 -------------
 
-``cabal v2-run [TARGET [ARGS]]`` runs the executable specified by the
-target, which can be a component, a package or can be left blank, as
-long as it can uniquely identify an executable within the project.
-Tests and benchmarks are also treated as executables.
+``cabal v2-run [TARGET [ARGS]]`` builds (if necessary) and runs the
+executable specified by the target, which can be a component, a
+package or can be left blank, as long as it can uniquely identify
+an executable within the project. Tests and benchmarks are also 
+treated as executables.
 
 See `the v2-build section <#cabal-v2-build>`__ for the target syntax.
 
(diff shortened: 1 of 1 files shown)


doc/cabal-commands.rst
```diff
@@ -273,7 +274,10 @@ cabal v2-exec
 
 ``cabal v2-exec [FLAGS] [--] COMMAND [--] [ARGS]`` runs the specified command
 using the project's environment. That is, passing the right flags to compiler
-invocations and bringing the project's executables into scope.
+invocations and bringing the project's executables into scope. 
+
+Note that unlike ``v2-run``, ``v2-exec`` will *not* (re)build anything; in particular,
+any changes since the last build will not be incorporated.
```
phadej, line 280: There is conceptual difference that `v2-exec` executes *a command* in project environment, that command can (and often indeed is) completely unrelated to to the projects' packages/executables/... If that can be made clearer, then it should be more obvious *why* `v2-exec` doesn't rebuild anything.

In other words, I feel that comparing `v2-run` and `v2-exec` is on the level what they concretely do (e.g. rebuiding) is wrong, we should rather explain how they differ conceptually.

Also, I never needed `cabal v2-exec`, it's there for more or less legacy reasons. I'd be :+1: saying explicitly, "you are probably looking for `v2-run`" in the docs.

(Bad) examples of using `v2-exec` are
- `v2-exec which program` to find a built executable, there is now `cabal list-bin`
- `v2-exec doctest src/` which may or may not work, there isn't a right alternative at the moment though.

Sorry for not having concrete improvement, just a long comment.

fgaz: What about just adding

> If you want to run an executable from your project, you are probably looking for `v2-run`. `v2-exec` is more suited for running external programs that act on or inspect the project's environment.

at the end?

Still not perfect, but a net improvement over the status quo, no?

jneira: and both?

> If you want to run an executable from your project, you are probably looking for v2-run. v2-exec is more suited for running external programs that act on or inspect the project's environment. Using it for running project executables can have unexpected behaviours like doesn't rebuild the executable.

Kleidukos:

PR #7123, opened 2020-10-14
Add sync support for hg

sumo:
* [Y ] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ NA] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
* [ NA] The documentation has been updated, if necessary.

Change was tested using a cabal.project.local with the following:

source-repository-package
   type: hg
   location: https://hg.sr.ht/~sumo/pairing-serialize
   tag: b8c5d29837eb

First checkout was verified with tag and without. On rebuild followup fetch + checkouts were verified against hg revs verified by using the hg id in the local checkout.

### DIFF
--- a/cabal-install/src/Distribution/Client/VCS.hs
+++ b/cabal-install/src/Distribution/Client/VCS.hs
@@ -445,7 +445,35 @@ vcsHg =
                  -> ConfiguredProgram
                  -> [(SourceRepositoryPackage f, FilePath)]
                  -> IO [MonitorFilePath]
-    vcsSyncRepos _v _p _rs = fail "sync repo not yet supported for hg"
+    vcsSyncRepos _ _ [] = return []
+    vcsSyncRepos verbosity hgProg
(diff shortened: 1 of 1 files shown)


phadej:
Hmm. Don't we have tests running `hg`? Or don't the test sync repos ever. (both are bad, tell me which and I'll take a closer look).

sumo:
> 
> 
> Hmm. Don't we have tests running `hg`? Or don't the test sync repos ever. (both are bad, tell me which and I'll take a closer look).

No tests in cabal-install. Happy to help adding them in for sync repos if it is priority.

phadej:
>  Happy to help adding them in for sync repos if it is priority.

Would be great to have them. Syncing (vs. clean clone) is quite common operation in practice.

sumo:
I've created a new PR with the correct tests in place.

PR #7129, opened 2020-10-22
Fix #7126: Stricter checks for relative paths

### LINKED ISSUE #7126: cabal check doesn't catch denormalised paths
For example: `hs-source-dirs: src/../../other-packag/src` is fine.

phadej:


phadej:

PR #7133, opened 2020-10-25
Vcs hg

sumo:
---

* [X ] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [NA] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
* [NA] The documentation has been updated, if necessary.

Updated the cabal-install VCS tests to include hg specific tests as well. cabal-install tests are not run by default so the cabal.project needs to have tests enabled before switching out to use the .dev cabal file.

### DIFF
--- a/cabal-install/src/Distribution/Client/VCS.hs
+++ b/cabal-install/src/Distribution/Client/VCS.hs
@@ -445,7 +445,35 @@ vcsHg =
                  -> ConfiguredProgram
                  -> [(SourceRepositoryPackage f, FilePath)]
                  -> IO [MonitorFilePath]
-    vcsSyncRepos _v _p _rs = fail "sync repo not yet supported for hg"
+    vcsSyncRepos _ _ [] = return []
+    vcsSyncRepos verbosity hgProg
+                 ((primaryRepo, primaryLocalDir) : secondaryRepos) = do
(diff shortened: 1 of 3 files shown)


cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -445,7 +445,35 @@ vcsHg =
...
+      where
+        hg :: FilePath -> [String] -> IO ()
+        hg cwd args = runProgramInvocation verbosity $
+                          (programInvocation hgProg args) {
+                            progInvokeCwd = Just cwd
+                          }
+        cloneArgs      = ["clone", "--noupdate", (srpLocation repo), localDir]
+                        ++ verboseArg
+        verboseArg = [ "--quiet" | verbosity < Verbosity.normal ]
+        checkoutArgs = [ "checkout", "--clean" ]
```
arrowd, line 472: It is a matter of taste, but I'd use command's original name, `update`, rather that alias.

Mikolaj:
@sumo: any updates on this? E.g., regarding the review comment? @arrowd: other than the comment does it look fine to you?

PR #7137, opened 2020-10-29
Sync repo darcs

strake:
It was falsely advertized to be supported. Now it should work.

I know not how to reset to a given repo state with Darcs, so if the local copy is not in the wanted state, cabal blows it away and re-clones it. It's not ideal but it does the job.

Tested for a while in some personal projects of mine.


strake:
I'm not sure how the changelog works — should i make a file "changelog.d/sync-repo" or what?

strake:
ping

fgaz:
> I'm not sure how the changelog works — should i make a file "changelog.d/sync-repo" or what?

Yes, you can look at the other files for an example

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
 
     vcsSyncRepos :: Verbosity -> ConfiguredProgram
                  -> [(SourceRepositoryPackage f, FilePath)] -> IO [MonitorFilePath]
-    vcsSyncRepos _v _p _rs = fail "sync repo not yet supported for darcs"
+    vcsSyncRepos verbosity prog = \ case
```
fgaz, line 317: LambdaCase is almost absent from the cabal codebase, I think we can avoid it here too and just match directly on rs instead, since it's just two cases.

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
...
+      [] -> pure []
```
fgaz, line 318:
```suggestion
      [] -> return []
```

Though I like `pure` too, in both this file and the rest of the repo `return` is used when in a monadic context. (same for the other `pure`s)

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
...
+          [ vcsSyncRepo verbosity prog repo localDir (Just primaryLocalDir)
+          | (repo, localDir) <- secondaryRepos ]
+
+    vcsSyncRepo verbosity prog SourceRepositoryPackage{..} localDir _peer =
+      try (lines <$> darcsWithOutput localDir ["log", "--last", "1"]) >>= \ case
+        Right (_:_:_:x:_)
+          | Just tag <- (List.stripPrefix "tagged " . List.dropWhile Char.isSpace) x
+          , Just tag' <- srpTag
+          , tag == tag' -> pure ()
+        Left (_ :: SomeException) -> pure ()
```
fgaz, line 333: Why should we ignore the exception? The one that needed catching was already taken care of below, no?

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
...
+        darcs :: FilePath -> [String] -> IO ()
+        darcs = darcs' runProgramInvocation
+
+        darcsWithOutput :: FilePath -> [String] -> IO String
+        darcsWithOutput = darcs' getProgramInvocationOutput
+
+        darcs' f cwd args = f verbosity (programInvocation prog args)
+          { progInvokeCwd = Just cwd }
+
+        cloneArgs = ["clone"] <|> tagArgs <|> [srpLocation, localDir] <|> verboseArg
```
fgaz, line 349: Similarly to `pure`/`return`, other similar code uses `++` instead of `<|>`. Same below.

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
...
+      try (lines <$> darcsWithOutput localDir ["log", "--last", "1"]) >>= \ case
+        Right (_:_:_:x:_)
+          | Just tag <- (List.stripPrefix "tagged " . List.dropWhile Char.isSpace) x
+          , Just tag' <- srpTag
+          , tag == tag' -> pure ()
+        Left (_ :: SomeException) -> pure ()
+        _ -> do
+          removeDirectoryRecursive localDir `catch` \ case
+              e | isDoesNotExistError e -> pure ()
+                | otherwise -> throw e
```
fgaz, line 337:
```suggestion
          removeDirectoryRecursive localDir `catch` \e ->
              unless (isDoesNotExistError e) throw e
```

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
 
     vcsSyncRepos :: Verbosity -> ConfiguredProgram
                  -> [(SourceRepositoryPackage f, FilePath)] -> IO [MonitorFilePath]
-    vcsSyncRepos _v _p _rs = fail "sync repo not yet supported for darcs"
+    vcsSyncRepos verbosity prog = \ case
+      [] -> pure []
+      (primaryRepo, primaryLocalDir) : secondaryRepos ->
+        [ monitorDirectoryExistence dir
+        | dir <- primaryLocalDir : (snd <$> secondaryRepos) ] <$ do
```
fgaz, line 321: I think giving this a name, or just putting it at the end of the do block could help readability a lot here.

In this case, I think desugaring the comprehension and cons would also help:
```haskell
primaryMonitor <- monitorDirectoryExistence primaryLocalDir
monitors <- traverse (monitorDirectoryExistence . snd) secondaryRepos
return $ primaryMonitor:monitors
```

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -305,7 +314,43 @@ vcsDarcs =
...
+        vcsSyncRepo verbosity prog primaryRepo primaryLocalDir Nothing
+        sequence_
+          [ vcsSyncRepo verbosity prog repo localDir (Just primaryLocalDir)
+          | (repo, localDir) <- secondaryRepos ]
```
fgaz, line 325: What about

```suggestion
        forM_ secondaryRepos $ \(repo, localDir) ->
          vcsSyncRepo verbosity prog repo localDir $ Just primaryLocalDir
```

fgaz (changes requested):
Mostly style/clarity stuff :)

cabal-install/src/Distribution/Client/VCS.hs
strake: Done

cabal-install/src/Distribution/Client/VCS.hs
strake: Done

cabal-install/src/Distribution/Client/VCS.hs
strake: Ahh... this part is subtle and my code was indeed wrong. PTAL to see whether it makes more sense now.

cabal-install/src/Distribution/Client/VCS.hs
strake: Done

cabal-install/src/Distribution/Client/VCS.hs
strake: Proposed code is type-incorrect — done with amendment

cabal-install/src/Distribution/Client/VCS.hs
strake: Done

cabal-install/src/Distribution/Client/VCS.hs
strake: Done

gbaz (approved):

emilypi (approved):

emilypi:
@mergifyio update

PR #7138, opened 2020-10-30
Remove duplicated stanza

tomjaguarpaw:
duplicated in b0333ec5b73ba8f7a18223b203d999b38c75281d likely due to
bad resolve of merge conflict

Issue #6918, opened 2020-06-18
Add PerOptionalStanza


phadej:
so we don't need to use `Map`, which is a bit silly :)

EDIT: also `[OptinalStanza]` could be `PerOptionalStanza Bool`.

PR #7139, opened 2020-10-30
Enable -fwarn-incomplete-record-updates

tomjaguarpaw:
to prepare for the time when GHC enables this flag under `-Wall`.

### DIFF
--- a/cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs
+++ b/cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs
@@ -9,7 +9,7 @@ module Distribution.Solver.Types.OptionalStanza
 import Distribution.Solver.Compat.Prelude
 import Prelude ()
 import Distribution.Types.ComponentRequestedSpec
-            (ComponentRequestedSpec(..), defaultComponentRequestedSpec)
+            (ComponentRequestedSpec(..))
 
 data OptionalStanza
     = TestStanzas
@@ -23,11 +23,14 @@ showStanza BenchStanzas = "bench"
 
 -- | Convert a list of 'OptionalStanza' into the corresponding
 -- 'ComponentRequestedSpec' which records what components are enabled.
+
+-- Note: [OptionalStanza] could become PerOptionalStanza Bool.
+-- See https://github.com/haskell/cabal/issues/6918
 enableStanzas :: [OptionalStanza] -> ComponentRequestedSpec
-enableStanzas = foldl' addStanza defaultComponentRequestedSpec
-  where
-    addStanza enabled TestStanzas  = enabled { testsRequested = True }
-    addStanza enabled BenchStanzas = enabled { benchmarksRequested = True }
+enableStanzas optionalStanzas = ComponentRequestedSpec {
+    testsRequested      = any (== TestStanzas)  optionalStanzas
+  , benchmarksRequested = any (== BenchStanzas) optionalStanzas
+  }
 
 instance Binary OptionalStanza
 instance Structured OptionalStanza
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -58,7 +58,7 @@ library
   else
     build-depends: unix  >= 2.6.0.0 && < 2.8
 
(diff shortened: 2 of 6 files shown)


tomjaguarpaw:
Given https://github.com/haskell/cabal/pull/7138#issuecomment-719486433 I'm not sure this is the right thing to do. I tried making the change in `cabal-install.cabal.zinza` and then running

```
make cabal-install-prod cabal-install-dev
````

but this made *huge* changes to `cabal-install.cabal` so I'm not sure that was the right thing to do. What should I do?

**Comment by phadej** [comment]
I'll write it here and would appreciate a PR cleaning up the language and putting it in contributing.md or/and other appropriate places:

- `make cabal-install-prod` and `make cabal-install-dev` only copy pregenerarated "production" (only single executable component) and "development" (internal library, tests, ...) variants
- `make cabal-install-cabal` is used to pregenerate the files
    - We commit pregenerated files, as it makes life easier: in CI, and for most simple dev needs.
- Why we have two modes:
    - Because we cannot commit on maintaining `cabal-install` library, and cannot (yet) bump `cabal-version: ` of `cabal-install` to support internal libraries. We try to be nice to distribution packagers, by not using latest bells and whistles while they aren't absolutely necessary.

For more details, look what `Makefile` does.

**Comment by tomjaguarpaw** [comment]
OK so it seems to me that the correct thing to do is

```
make cabal-install/cabal-install.cabal.dev cabal-install-prod
```

which makes `cabal-install/cabal-install.cabal.dev` and both makes `cabal-install/cabal-install.cabal.prod` and installs it on top of `cabal-install/cabal-install.cabal`. I have updated this PR with the result of that. Do you agree that's the correct thing to do? If so I can update the docs too.

File: cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs [hunk 1]
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# OPTIONS_GHC -fno-warn-incomplete-record-updates #-}
```
phadej, line 3: Why?

tomjaguarpaw: Because these are an incomplete record update

https://github.com/haskell/cabal/blob/e76b56dc7b2eba7696f85844b690daa99e3eca20/cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs#L30-L31

See

https://github.com/haskell/cabal/blob/e76b56dc7b2eba7696f85844b690daa99e3eca20/Cabal/src/Distribution/Types/ComponentRequestedSpec.hs#L66-L70

phadej: Rather, fix those. Make setters `setTestsRequested` and `setBenchmarkRequested`.

phadej (changes requested):

cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs
tomjaguarpaw: I'm not sure what the intended semantics of those should be, specifically, what should they do on a `OneComponentRequestedSpec`?

As a compromise, how about the latest?

cabal-install/cabal-install-solver/src/Distribution/Solver/Types/OptionalStanza.hs
```diff
@@ -24,10 +24,10 @@ showStanza BenchStanzas = "bench"
 -- | Convert a list of 'OptionalStanza' into the corresponding
 -- 'ComponentRequestedSpec' which records what components are enabled.
 enableStanzas :: [OptionalStanza] -> ComponentRequestedSpec
-enableStanzas = foldl' addStanza defaultComponentRequestedSpec
-  where
-    addStanza enabled TestStanzas  = enabled { testsRequested = True }
-    addStanza enabled BenchStanzas = enabled { benchmarksRequested = True }
+enableStanzas optionalStanzas = ComponentRequestedSpec {
```
phadej, line 27: Leave a note about #6918

phadej:
merged in #7144

PR #7141, opened 2020-10-30
Add forward declarations to ForeignLib tests

phadej:


phadej:
Backported to 3.4 in https://github.com/haskell/cabal/pull/7166

PR #7152, opened 2020-11-05
Clarify cabal.project documentation

toonn:
I was confused about the resolution of settings in the presence of a `cabal.project` so I added some minimal documentation that would've helped me.

I was confused because in a project with both a `<pkgname>.cabal` and a `cabal.project`, v2-build complained about neither of them existing:
```
cabal: There is no <pkgname>.cabal package file or cabal.project file. To
build packages locally you need at minimum a <pkgname>.cabal file. You can use
'cabal init' to create one.

For non-trivial projects you will also want a cabal.project file in the root
directory of your project. This file lists the packages in your project and
all other build configuration. See the Cabal user guide for full details.
```

The problem turned out to be not setting `packages:` in the `cabal.project`, I'd only set `index-state:` which erases the default value.

I'd like to refine that error message to be more helpful in case a `cabal.project` file is present, explaining `packages:` should probably be set, rather than denying the files exist : )

---

### DIFF
--- a/doc/cabal-project.rst
+++ b/doc/cabal-project.rst
@@ -21,7 +21,8 @@ line flags that ``cabal install`` and other commands take. For example,
 file with ``profiling: True``.
 
 The full configuration of a project is determined by combining the
-following sources (later entries override earlier ones):
+following sources (later entries override earlier ones, except for appendable
+options):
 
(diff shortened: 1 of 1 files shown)


emilypi (approved):

emilypi:
thanks @toonn

PR #7158, opened 2020-11-10
Extensible Interface Directories 

JoshMeredith:
This patch aims to support future work on GHC with the intent of `.hi` interface files instead being directories (https://gitlab.haskell.org/ghc/ghc/-/issues/18954). This feature requires small breaking changes in Cabal, so the goal of this is to get ahead of the release cycle.

Currently, GHC supports extensibility features in standard `.hi` interfaces files by way of a custom format (https://gitlab.haskell.org/ghc/ghc/-/wikis/Extensible-Interface-Files). However, there are a number of advantages to doing away with a custom format and instead relying on the filesystem for this. In particular, by supporting `.hi` directories instead of having an extra custom directory for extensibility, we only need to support this feature once within Cabal for the `install` command - whereas an additional directory would require all tools to have knowledge of it.

### DIFF
--- a/Cabal/src/Distribution/Simple/Utils.hs
+++ b/Cabal/src/Distribution/Simple/Utils.hs
@@ -79,6 +79,7 @@ module Distribution.Simple.Utils (
         installExecutableFiles,
         installMaybeExecutableFiles,
         installDirectoryContents,
+        installDirectoriesContents,
         copyDirectoryRecursive,
 
         -- * File permissions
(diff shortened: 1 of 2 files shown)


Cabal/src/Distribution/Simple/Utils.hs
```diff
@@ -1365,15 +1366,24 @@ installMaybeExecutableFiles :: Verbosity -> FilePath -> [(FilePath, FilePath)]
                                -> IO ()
 installMaybeExecutableFiles v fp fs = withFrozenCallStack (copyFilesWith installMaybeExecutableFile v fp fs)
 
+-- | This is like 'copyFiles' but uses 'installDirectoryContents'.
+--
+installDirectoriesContents :: Verbosity -> FilePath -> [(FilePath, FilePath)] -> IO ()
+installDirectoriesContents v fp fs = withFrozenCallStack (copyFilesWith installDirectoryContents v fp fs)
+
 -- | This installs all the files in a directory to a target location,
```
phadej, line 1374: The documentation is outdated with this change.

phadej (changes requested):
I'd rather not accept this yet. This is reasonable change on its own, but if GHC decides to not use this approach it will be unnecessary addition.

Kleidukos:
Let's revive this when GHC adopts it.

ffaf1:

PR #7163, opened 2020-11-19
Use environment paths in GitHub CI

phadej:


phadej:
Looks like our Windows CI is busted, because `Error: Unable to process command '::add-path::C:\ProgramData\chocolatey\lib\ghc\tools\ghc-8.8.4\bin' successfully.` which comes from installing GHC packages.

jneira:
we take ride of the error  in ghcide f.e upgrading haskell action version 
https://github.com/haskell/ghcide/pull/915

phadej:
I will not use sliding tags. The whole issue is because environment changes without any control on it, I would not want to introduce any more instability into CI.

Or in another words, I think that Cabal's CIs maintainer's "job" is open for applications.

phadej:
ah, so `v1.1` is fixed tag, only `v1` is retagged. I see. Confusing. Let me try `v1.1.4`

jneira:
> I will not use sliding tags. The whole issue is because environment changes without any control on it, I would not want to introduce any more instability into CI.

Fair enough, only wanted to note that upgrading the action to last version (1.1.4) should fix the issue.

For reference the issue in github actions is:
https://github.com/actions/setup-haskell/issues/44

phadej:
> Fair enough, only wanted to note that upgrading the action to last version (1.1.4) should fix the issue.

Yes. It was my confusion. See https://github.com/actions/setup-haskell/issues/51

phadej:
Backported to 3.4 in https://github.com/haskell/cabal/pull/7166

PR #7190, opened 2020-12-06
Cabal: Improve cmdline help readability by breaking long lines

DanielG:
Currently usageInfo is very ridgid about wrapping the cmdline help
output. Only the help text itself is wrapped, the list of short and long
option names is kept on one line always.

This patch allows more wrapping to happen for very long long-option
names. For example cabal-install has the

    --write-ghc-environment-files=always|never|ghc8.4.4+

option which pushes the help text column over 80 cols for no good
reason. By simply wrapping the possible values after the "=" onto the next
line we can already return sanity to the output.

The patch also gets rid of the dedicated short options column. Previously
the long options column would be aligned to where the longest short option
ends, but since the vast majority of options are long options this seems
misguided. Now long and short options are simply smushed together into one
column.

---
* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ ] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
* [ ] The documentation has been updated, if necessary.

--- 

Current `cabal build --help`, notice the crazy amount of dead space because `--write-ghc-environment-files` pushes everything to the right

```
    --write-ghc-environment-files=always|never|ghc8.4.4+ Whether to create a
                                                         .ghc.environment file after a
                                                         successful build (v2-build
                                                         only)
    --enable-documentation                               Enable building of
                                                         documentation
    --disable-documentation                              Disable building of
                                                         documentation
    --doc-index-file=TEMPLATE                            A central index of haddock
                                                         API documentation (template
                                                         cannot use $pkgid)
    --dry-run                                            Do not install anything, only
                                                         print what would be
                                                         installed.
```

<details>

```
 -h --help                                               Show this help text
 -v --verbose[=n]                                        Control verbosity (n is 0--3,
                                                         default verbosity level is 1)
    --builddir=DIR                                       The directory where Cabal
                                                         puts generated build files
                                                         (default dist)
 -g --ghc                                                compile with GHC
    --ghcjs                                              compile with GHCJS
    --uhc                                                compile with UHC
    --haskell-suite                                      compile with a haskell-suite
                                                         compiler
    --cabal-file=PATH                                    use this Cabal file
 -w --with-compiler=PATH                                 give the path to a particular
                                                         compiler
    --with-hc-pkg=PATH                                   give the path to the package
                                                         tool
    --prefix=DIR                                         bake this prefix in
                                                         preparation of installation
[...]

PR #7191, opened 2020-12-08
GetOpt: Make OptHelp less polymorphic

DanielG:
This was a leftover from the first iteration in #7190 where we used to map over optNames, but it's not needed anymore.


emilypi (approved):
@fgaz mind +1'ing this and we can get it in?

fgaz (approved):

fgaz:
@DanielG Rebase on master to fix the missing CI job

PR #7195, opened 2020-12-10
Allow base16-bytestring 1.0 and newer

mimi1vx:
---

--> https://build.opensuse.org/project/show/devel:languages:haskell:ghc-8.10.x

### DIFF
--- a/cabal-install/src/Distribution/Client/HashValue.hs
+++ b/cabal-install/src/Distribution/Client/HashValue.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP                #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric      #-}
 module Distribution.Client.HashValue (
@@ -72,14 +71,8 @@ hashFromTUF (Sec.Hash hashstr) =
     --TODO: [code cleanup] either we should get TUF to use raw bytestrings or
     -- perhaps we should also just use a base16 string as the internal rep.
     case Base16.decode (BS.pack hashstr) of
-#if MIN_VERSION_base16_bytestring(1,0,0)
       Right hash -> HashValue hash
       Left _ -> error "hashFromTUF: cannot decode base16"
-#else
-      (hash, trailing) | not (BS.null hash) && BS.null trailing
-        -> HashValue hash
-      _ -> error "hashFromTUF: cannot decode base16 hash"
-#endif
 
 -- | Truncate a 32 byte SHA256 hash to
 --
--- a/cabal-install/src/Distribution/Client/HashValue.hs
+++ b/cabal-install/src/Distribution/Client/HashValue.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP                #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric      #-}
 module Distribution.Client.HashValue (
@@ -71,10 +72,14 @@ hashFromTUF (Sec.Hash hashstr) =
     --TODO: [code cleanup] either we should get TUF to use raw bytestrings or
     -- perhaps we should also just use a base16 string as the internal rep.
     case Base16.decode (BS.pack hashstr) of
+#if MIN_VERSION_base16_bytestring(1,0,0)
+      Right hash -> HashValue hash
+      Left _ -> error "hashFromTUF: cannot decode base16"
+#else
       (hash, trailing) | not (BS.null hash) && BS.null trailing
         -> HashValue hash
       _ -> error "hashFromTUF: cannot decode base16 hash"
-
+#endif
 
 -- | Truncate a 32 byte SHA256 hash to
 --
--- a/cabal-install/src/Distribution/Client/HashValue.hs
+++ b/cabal-install/src/Distribution/Client/HashValue.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP                #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric      #-}
 module Distribution.Client.HashValue (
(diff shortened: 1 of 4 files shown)


phadej:
This is blocked on https://github.com/haskell-hvr/resolv/pull/10

Bodigrim:
@emilypi since `resolv` issue has been resolved, could you please take a look at this? We need to allow `base16-bytestring-1.0` for `bytestring-0.11` transition.

cabal-install/src/Distribution/Client/HashValue.hs
```diff
@@ -71,10 +72,14 @@ hashFromTUF (Sec.Hash hashstr) =
     --TODO: [code cleanup] either we should get TUF to use raw bytestrings or
     -- perhaps we should also just use a base16 string as the internal rep.
     case Base16.decode (BS.pack hashstr) of
+#if MIN_VERSION_base16_bytestring(1,0,0)
+      Right hash -> HashValue hash
+      Left _ -> error "hashFromTUF: cannot decode base16"
+#else
```
emilypi, line 78: Looks correct

cabal-install/src/Distribution/Client/HttpUtils.hs
```diff
@@ -182,7 +183,11 @@ downloadURI transport verbosity uri path = do
     fragmentParser = do
         _ <- P.string "#sha256="
         str <- some P.hexDigit
+#if MIN_VERSION_base16_bytestring(1,0,0)
+        return (Base16.decodeLenient (BS8.pack str))
```
emilypi, line 187: So this is not quite equivalent behavior, because the previous behavior constructs valid bytestring "up to invalidity" in the source string, while `decodeLenient` passes through and does its best to construct a bytestring from what's left. Do we care about this in particular? It probably will not cause problems, but we should confirm.

fgaz:
> since resolv issue has been resolved

Just wanted to say that this made me laugh :D

cabal-install/src/Distribution/Client/HttpUtils.hs
Bodigrim: I do not see a good reason to use `decodeLenient` instead of at least `decode`. AFAIU ideally we should read an even number of hex digits into `str` (so that it's a valid base16 string), then parse with 
```
Right decoded <- Base16.decode (BS8.pack str)
return decoded
```
leveraging `instance MonadFail Parsec`.
(I do not have access rights to push necessary commits to this PR)

emilypi: That sounds more better to me if we want this to only accept morally correct inputs.

emilypi (changes requested):
@mimi1vx Could you please change the `HttpUtils` code to address our comments? Otherwise it looks good and we'll merge when that is in.

emilypi:
ping @mimi1vx

cabal-install/src/Distribution/Client/HttpUtils.hs
mimi1vx: @Bodigrim I added You to my fork, feel free to change it

mimi1vx: @emilypi  I don't think anybody cared about the validity of this bytestring. Code for older version of Base16-bytestring `return (fst (Base16.decode (BS8.pack str)))`  simply returns results without any care about error part of results tuple

cabal-testsuite/cabal-testsuite.cabal
```diff
@@ -58,7 +58,7 @@ library
     , aeson                 ^>= 1.4.2.0 || ^>=1.5.0.0
     , async                 ^>= 2.2.1
     , attoparsec            ^>= 0.13.2.2
-    , base16-bytestring     ^>= 0.1.1.6
```
emilypi, line 61: If possible, I would recomment bumping this to just `^>=1.0.0.0` if we can get away with it. The difference in performance is massive, and the new versions are completely back-compatible for the range of GHCs cabal cares about: https://matrix.hackage.haskell.org/#/package/base16-bytestring

Bodigrim: I do not have a strong opinion, but if it is a question of performance and not a question of security, I'd rather avoid a cliff age (switching from `<1` to `>1`) and allow a wider range of options. There are still many packages out there, which do not support `base16-bytestring-1.0`, https://packdeps.haskellers.com/reverse/base16-bytestring

emilypi: fair.

emilypi (approved):

mimi1vx:
@emilypi this PR we can also apply for cabal-install-3.4* and partially ( not needede HttpUtils.hs) to 3.2.*

cabal-testsuite/cabal-testsuite.cabal
fgaz: > I'd rather avoid a cliff age (switching from <1 to >1) and allow a wider range of options. There are still many packages out there, which do not support base16-bytestring-1.0

How is this a problem? base16-bytestring is a direct dependency (except for resolv) and cabal-install shouldn't have any reverse dependencies, should it? And both base16-bytestring-1.0 and resolv build on older ghcs too: https://matrix.hackage.haskell.org/#/package/base16-bytestring

>=1 would also save us the CPP (that we'd have to remember to remove when eventually bumping the lower bound)

Bodigrim: @fgaz no problem, pushed.

Bodigrim: @fgaz nope, it does not pass CI as configured, because `base16-bytestring-0.1.1.7` is freezed in bootstrap jsons and docker containers. Honestly, I feel that fixing them goes beyond the scope of this PR (and my role in it). I'd prefer to revert back 981f6d9, if this is fine with you.

fgaz: That's fine. Thanks for giving it a go

Bodigrim: Reverted 981f6d9, hopefully it is ready now.

Bodigrim:
@fgaz shall we merge this? This is a prerequisite to bump an upper bound for bytestring.

PR #7205, opened 2020-12-19
Use alex-3.2.6

phadej:


phadej:
Backported in https://github.com/haskell/cabal/pull/7220

PR #7206, opened 2020-12-19
Remove Win GHC-8.8.4 job

phadej:


phadej:
8.10.2 job started to segfault. So I'd rather not remove 8.8.4 job yet. Let's see if 8.10.3 job behaves better.

PR #7208, opened 2020-12-21
doc: Annotate package examples

toonn:
The simple package examples share module names among library and
executable components. To avoid potential confusion these have been
annotated to make it crystal clear the modules do not refer to the same
entities.

[ci skip]

An alternative approach is to simply uses non-conflicting names in the example, which would side-step the initial confusion but not teach users that module names do not have to be unique within a project.
Suggestions for improvement are welcome!

---


phadej:
The latter example is indeed bad. I'd rather
- use `hs-source-dirs` for library as well,
- use separate modules (as it's confusing if it's not true)
- use `main.hs` - lowercase to highlight that it is not a "module name" file.
- add dependency from executables to library, as that is common use case.

phadej (approved):

emilypi:
@toonn if you want to rebase to master, that should fix the CI and we can get this in

fgaz:
Let's get this improvement to the (soon to be) default version of the docs:

@Mergifyio backport 3.6

PR #7219, opened 2020-12-30
Paths no missing safe haskell mode

phadej:


phadej:
Backported in https://github.com/haskell/cabal/pull/7220

PR #7228, opened 2021-01-08
Remove the deprecated extension MonadFailDesugaring

Kleidukos:
This PR is the sibling of https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4767 that removes the deprecated extension MonadFailDesugaring. 

cc @phadej

### DIFF
--- a/Cabal/src/Language/Haskell/Extension.hs
+++ b/Cabal/src/Language/Haskell/Extension.hs
@@ -783,10 +783,6 @@ data KnownExtension =
   -- | Allow recursive (and therefore undecidable) super-class relationships.
   | UndecidableSuperClasses
 
-  -- | A temporary extension to help library authors check if their
-  -- code will compile with the new planned desugaring of fail.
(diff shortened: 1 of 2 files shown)


phadej (changes requested):
It cannot be removed. `Cabal` *MUST* support older GHCs. See e.g. https://hackage-search.serokell.io/?q=MonadFailDesugaring

Which I interpret as "extension list is monotonically growing one".

Kleidukos:
Then we're going to have a little problem with this test https://gitlab.haskell.org/ghc/ghc/-/blob/master/testsuite/tests/driver/T4437.hs right?

phadej:
I don't see a problem. You can add `MonadFailDesugaring` to cabal-only list, when the extension is removed from GHC itself.

Kleidukos:
Okay, I wasn't so sure it would be an acceptable way to solve it. Thanks!

PR #7233, opened 2021-01-09
Moved info for cabal install --lib to sub-heading

danidiaz:
The text for the `--lib` flag to the `cabal install` subcommand has been moved to its own subheading below that of `cabal install`. Also the text has been slightly reworded.

This is convenient because sometimes you want to link directly to the `cabal install --lib` behavior.

(Sorry if this is too trivial a change.)


fgaz:
> Sorry if this is too trivial a change

Docs patches are always welcome :)

I think you need to rebase to fix that stuck ci job

fgaz:
Thanks!

PR #7243, opened 2021-01-11
Add GHC-8.10.3 Windows CI job

phadej:


phadej:
GHC-8.10.3 job was successful, which is great. Let's hope it is not flaky

PR #7253, opened 2021-01-19
[ fix #7251 ] sort and nub module list inferred by cabal init --lib

### LINKED ISSUE #7251: `cabal init --lib` puts build-tools(happy, alex) generated modules twice into exposed-modules
`cabal init` nicely auto-detects `build-tools` like `happy` and `alex`.  However, it puts their generated modules twice into `exposed-modules`, like:
```
library
  exposed-modules: Main, Foo.Print, Foo.Lex, Foo.Skel, Foo.Par, Foo.Lex, Foo.Test, Foo.Abs, Foo.ErrM, Foo.Par
```
Note that `Foo.Lex` and `Foo.Par` are listed twice each.

`cabal build` then fails with a confusing error message:
```
Error:
    Module name  Foo.Lex  is exported multiple times.
    In the stanza 'library'
    In the inplace package 'DuplicateModules-0.1.0.0'
```
That's hard to understand.  Took me a while, esp. since I didn't know where to look for the "inplace package".

I suppose a simple `nub` would solve this issue.  Anyway, the order in `exposed-modules` looks random, wouldn't it anyway be nicer in alphabetical order?  Then `Set.toList . Set.fromList` would do the job.

Tested `cabal` versions: 3.2.0.0 and 3.5.0.0.

Steps to reproduce (using BNFC to generate .x and .y files cheaply):
```
cabal install BNFC
echo "Foo. Bar ::= Integer" > Foo.cf
bnfc -dm Foo.cf
make
cabal init --lib
cabal build
```

Trivium: `hpack` had the very same problem: https://github.com/sol/hpack/issues/406

andreasabel:
[ fix #7251 ] sort and nub module list inferred by cabal init --lib

This should fix the problem that duplicate entries end up in exposed-modules, e.g. from `Foo.x` and `Foo.hs` (with `alex` as build-tool).

* [x] > Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [x] > Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
  This fixes a broken behavior, shouldn't need mentioning.
* [x] > The documentation has been updated, if necessary.
  Ditto

> Please also shortly describe how you tested your change. Bonus points for added tests!

Couldn't test yet because of https://github.com/haskell/cabal/issues/7251#issuecomment-763050836 :-(

TODO:
- [ ] test
- [ ] changelog

### DIFF
--- a/cabal-install/src/Distribution/Client/Init/Command.hs
+++ b/cabal-install/src/Distribution/Client/Init/Command.hs
@@ -33,6 +33,8 @@ import System.FilePath
 
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map as M
+import qualified Data.Set as Set
+
 import Control.Monad
   ( (>=>) )
 import Control.Arrow
@@ -595,7 +597,7 @@ getModulesBuildToolsAndDeps pkgIx flags = do
   let sourceFiles = filter (isSourceFile (sourceDirs flags)) sourceFiles0
 
   Just mods <-      return (exposedModules flags)
-           ?>> (return . Just . map moduleName $ sourceFiles)
+           ?>> (return . Just . sortNub . map moduleName $ sourceFiles)
 
   tools <-     return (buildTools flags)
            ?>> (return . Just . neededBuildPrograms $ sourceFiles)
@@ -644,6 +646,10 @@ getModulesBuildToolsAndDeps pkgIx flags = do
                  , dependencies   = deps
                  , otherExts      = exts
                  }
+
+  where
+  sortNub :: Ord a => [a] -> [a]
+  sortNub = Set.toList . Set.fromList
 
 -- | Given a list of imported modules, retrieve the list of dependencies that
 -- provide those modules.


andreasabel:
Here is a test script now (using BNFC):
```
bnfc --version || cabal install BNFC
mkdir -p myparser
cd myparser
echo "Foo. Bar ::= Integer" > Foo.cf
bnfc -o src -d -m Foo.cf
make -C src
mv src/Foo/Test.hs Main.hs
cabal init --libandexe --application-dir=.
cabal build all
cat myparser.cabal
```
This will create a meaningful `myparser.cabal` except that the `executable` section does not have the necessary dependency on the `library`.

fgaz:
Having a test for this in cabal-testsuite would be good

fgaz:
> This fixes a broken behavior, shouldn't need mentioning.

Bug fixes go in the changelog too, plus there's the sorting

Here's how the changelog.d works: https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog

andreasabel:
Thanks for the feedback @fgaz!  
- Could you point me to a location where the new test should reside?
- Is there maybe some documentation how to add a test?

fgaz:
The most straightforward way to test this is with a package test. You can add it in `cabal-testsuite/PackageTests`, where you can also find a readme and many examples. Check out the "Hermetic tests" section in particular, since you'll need it for this test

fgaz:
@emilypi does your pr supersede this?

emilypi:
@fgaz yeah. Was this not linked? It should be.

PR #7254, opened 2021-01-19
No crash on bad package file.

strake:
If we can't decompress a package file and read the cabal file in it, warn the user rather than crash.

---

Tested locally with a local repo sans index containing a few empty ".tar.gz" files.

### DIFF
--- a/cabal-install/src/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/src/Distribution/Client/IndexUtils.hs
@@ -111,9 +111,13 @@ import Distribution.Compat.Directory (listDirectory)
 import Distribution.Utils.Generic (fstOf3)
 
 import qualified Codec.Compression.GZip as GZip
+import qualified Codec.Compression.Zlib.Internal as Zlib
 
 import qualified Hackage.Security.Client    as Sec
 import qualified Hackage.Security.Util.Some as Sec
+
+import Control.Monad.Trans.Class (lift)
+import Control.Monad.Trans.Maybe (MaybeT (..))
 
 -- | Reduced-verbosity version of 'Configure.getInstalledPackages'
 getInstalledPackages :: Verbosity -> Compiler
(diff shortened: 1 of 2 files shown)


strake:
I'm not sure how changelog.d works...

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -716,15 +720,15 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
...
-
-                case Tar.foldEntries (readCabalEntry pkgId) Nothing (const Nothing) entries of
+                tarGz <- lift $ BS.readFile (localDir </> file)
+                tar <- MaybeT $ (Just <$> evaluate (GZip.decompress tarGz)) `catch` \ (e :: Zlib.DecompressError) ->
+                  Nothing <$ warn verbosity ("Cannot read " ++ file ++ ": " ++ show e)
+                let entries = Tar.read tar
+                MaybeT $ case Tar.foldEntries (readCabalEntry pkgId) Nothing (const Nothing) entries of
                     Just ce -> return (Just ce)
-                    Nothing -> die' verbosity $ "Cannot read .cabal file inside " ++ file
+                    Nothing -> Nothing <$ warn verbosity ("Cannot read .cabal file inside " ++ file)
```
emilypi, line 732: Just a suggestion for more legible code here: 

```haskell
Just pkgId -> do 
  tarGz <- BS.readFile $ localDir </> file 
  tarM <- try (gGZip.decompress tarGz) >>= \case 
    Left (e :: ZLib.DecompressError) -> do 
      warn verbosity $ "Cannot read" ++ file ++ ": " ++ show e
      return Nothing
    Right a -> return $ Just a 
  
  case tarM of  
    Nothing -> return Nothing 
    Just tar -> do 
      let entries = Tar.read tar 
      case Tar.foldEntries (readCabalEntry pkgId) Nothing (const Nothing) entries of 
        Nothing -> do 
          warn verbosity $ "Cannot read .cabal file inside " ++ file 
          return Nothing 
        Just ce -> return $ Just ce 
```

emilypi (changes requested):
Hi Matthew, 

I like the idea of the PR, but the use of transformers and `<$` here are a bit of a put-off. I've suggested something in a style that the maintainership would approve of

cabal-install/src/Distribution/Client/IndexUtils.hs
phadej: I would not bother here at all, but let the exception bail up and handle it with `catches`. There is already a top-level `handle` catching `IOException`, that can handle `Zlib.DecompressError` too.

I don't see much value in bare warning when there is broken tarball. People don't read warnings and open issues that things don't work. Better to fail early.

emilypi:
Any commentary, or is this closable @strake ?

strake:
Ah, i assumed from @phadej's comment this would be closed as EWONTFIX, as bailing up to the `handle` is what i sought to avoid here, in case some package files are invalid. But if the maintainers remain open to this, i will respond to the review comments. Please let me know.

In case it helps: the motivation for this is a Hackage mirror — it seems a few package files on Hackage (or, at least, a few files with a URL fitting the package file URL scheme) are invalid.

At least, it would be much more useful for me to see a list of **all** invalid package files, rather than merely the first, so it's easier to delete them and then try again.

fgaz:
> In case it helps

It does, thanks!

> it seems a few package files on Hackage (or, at least, a few files with a URL fitting the package file URL scheme) are invalid

This is known: there is [a quirks file](https://github.com/haskell/cabal/blob/aabe56252fcceb58f2f3f80451bba7ca13109266/Cabal/src/Distribution/PackageDescription/Quirks.hs) that rewrites the problematic packages. This is done to avoid rewriting and invalidating the entire hackage index, which is more painful than the current solution. I wonder why this is not triggering for you? Or are the invalid packages not included there?

So I also think the failure should remain. Printing all the failures could be a good idea though.

gbaz:
closing based on above discussion.

PR #7266, opened 2021-01-31
Address old comment in check code

fendor:
According to the comment in 'checkConfiguredPackage', the
'GenericPackageDescription' should always be available and the function
'checkConfiguredPackage' should be removed. That's how we did it and the
only consumer of that function has 'GenericPackageDescription'
available.

---

I did not test it at all, since it is only parameter re-order.

Unfortunately, I can't tell if this requires additional documentation for the user, etc...

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Check.hs
+++ b/Cabal/src/Distribution/PackageDescription/Check.hs
@@ -24,7 +24,6 @@ module Distribution.PackageDescription.Check (
         -- * Package Checking
         PackageCheck(..),
         checkPackage,
-        checkConfiguredPackage,
 
         -- ** Checking package contents
         checkPackageFiles,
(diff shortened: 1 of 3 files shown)


fendor:
ok, turns out that breaks a lot. 

Should we try to fix the tests or is that some behaviour people started to rely on and, thus, we should not change it?

fendor:
I think the rework of the check comment will surpass this, closing as I am not intending to continue working on it atm

Mikolaj:
"rework of the check comment"? Could you kindly link to it?

fendor:
TBH, I haven't checked but https://github.com/haskell/cabal/pull/8427 seems like it is doing more than just reworking one code-path

Mikolaj:
Oh, "check command"? Thank you, now it's clear and the link goes both ways so your work should not get lost.

fendor:
Didn't notice the typo, sorry :D

PR #7276, opened 2021-02-07
Replace List with NonEmpty in TargetsMap

fendor:
Innocent refactoring with the intention to learn the code-base a bit better

### DIFF
--- a/cabal-install/src/Distribution/Client/ProjectPlanning.hs
+++ b/cabal-install/src/Distribution/Client/ProjectPlanning.hs
@@ -163,7 +163,7 @@ import qualified Data.Map as Map
 import qualified Data.Set as Set
 import           Control.Monad.State as State
 import           Control.Exception (assert)
-import           Data.List (groupBy)
+import           Data.List (groupBy, deleteBy)
 import qualified Data.List.NonEmpty as NE
 import           System.FilePath
 
@@ -1758,7 +1758,7 @@ elaborateInstallPlan verbosity platform compiler compilerprogdb pkgConfigDB
             -- package needs to be rebuilt.  (It needs to be done here,
             -- because the ElaboratedConfiguredPackage is where we test
             -- whether or not there have been changes.)
-            TestStanzas  -> listToMaybe [ v | v <- maybeToList tests, _ <- PD.testSuites elabPkgDescription ] 
+            TestStanzas  -> listToMaybe [ v | v <- maybeToList tests, _ <- PD.testSuites elabPkgDescription ]
             BenchStanzas -> listToMaybe [ v | v <- maybeToList benchmarks, _ <- PD.benchmarks elabPkgDescription ]
           where
             tests, benchmarks :: Maybe Bool
@@ -2503,7 +2503,7 @@ availableSourceTargets elab =
 -- We also allow for information associated with each component target, and
 -- whenever we targets subsume each other we aggregate their associated info.
 --
-nubComponentTargets :: [(ComponentTarget, a)] -> [(ComponentTarget, [a])]
+nubComponentTargets :: [(ComponentTarget, a)] -> [(ComponentTarget, NonEmpty a)]
 nubComponentTargets =
     concatMap (wholeComponentOverrides . map snd)
   . groupBy ((==)    `on` fst)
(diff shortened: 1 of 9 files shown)


cabal-install/src/Distribution/Client/ProjectPlanning.hs
```diff
@@ -2514,11 +2514,11 @@ nubComponentTargets =
     -- If we're building the whole component then that the only target all we
     -- need, otherwise we can have several targets within the component.
     wholeComponentOverrides :: [(ComponentTarget,  a )]
-                            -> [(ComponentTarget, [a])]
+                            -> [(ComponentTarget, NonEmpty a)]
     wholeComponentOverrides ts =
       case [ t | (t@(ComponentTarget _ WholeComponent), _) <- ts ] of
-        (t:_) -> [ (t, map snd ts) ]
-        []    -> [ (t,[x]) | (t,x) <- ts ]
+        (t:_) -> [ (t, NE.fromList (map snd ts)) ]
```
fendor, line 2520: Only unsafe function usage, should the pre-condition be documented somehow?

phadej:
@fendor I'm sorry, but I wont have time to review any time soon.

fendor:
No problem, obviously!

fendor:
Since this is a benign change, not really observable by others, I would merge this in about three days unless someone objects.

phadej (dismissed):
I object with any introduction of partial functions. There are enough already leading to unexpectable errors when invariants are violated. Please at least use specific `fromMaybe (error pointer-to-where-and-what-invariant-is-violated) ...`

phadej:
I also think that `NonEmpty` variant can always be maintained (maybe with some slightly larger change). But as said, I have no time to review and think properly about this case.

cabal-install/src/Distribution/Client/ProjectPlanning.hs
emilypi: yeah, just an inline comment should be fine here

fgaz: If it isn't much trouble I'd follow phadej's advice here. Dumping "NonEmpty.fromList: empty list" on the user isn't too nice

fendor: I came up with the following solution:

```haskell
    wholeComponentOverrides :: [(ComponentTarget,  a )]
                            -> [(ComponentTarget, NonEmpty a)]
    wholeComponentOverrides ts =
      case [ ta | ta@(ComponentTarget _ WholeComponent, _) <- ts ] of
        ((t, x):_) -> 
                let
                    -- Delete tuple (t, x) from original list to avoid duplicates.
                    -- Use 'deleteBy', to avoid additional Class constraint on 'nubComponentTargets'.
                    ts' = deleteBy (\(t1, _) (t2, _) -> t1 == t2) (t, x) ts
                in
                    [ (t, x :| map snd ts') ]
        []    -> [ (t, x :| []) | (t,x) <- ts ]
```

Longer, but no unsafe function!

m-renaud:
Nice changes @fendor! Could you expand the PR description with some information on where it's possible for the NonEmpty invariant to not hold (if that's the case), as well as the way that it can gracefully fail with a useful message?

fendor:
There is no partial code left, so I would assume the invariant has to hold everywhere. The only slightly fishy code is https://github.com/haskell/cabal/pull/7276/files#diff-2adae381a2eb2d8c804d5396694a431e1e3babd816a5d2408a62213cf774a893R2524 which has a comment.

m-renaud:
> There is no partial code left, so I would assume the invariant has to hold everywhere. The only slightly fishy code is https://github.com/haskell/cabal/pull/7276/files#diff-2adae381a2eb2d8c804d5396694a431e1e3babd816a5d2408a62213cf774a893R2524 which has a comment.

@phadej since the code now doesn't have any violations of the NonEmpty invariant, do you have any oppositions to merging this?

phadej:
I don't know. Don't block on my opinion.

m-renaud:
@fendor could you add to the PR description how you've tested the codepaths being modified? Ideally both unit testing and integration (CLI invocation) testing, but I'm not sure what the current state of testing is here.

@emilypi I see that you have a big PR out to rewrite a bunch of cabal init (which also touched/added a bunch of tests), do you have any guidance for how @fendor could best test this?

fendor:
As `nubComponentTargets` is an exposed function, it should be possible (and sensible) to write unit-tests for it.

m-renaud (approved):
Looks great, thanks!

fendor:
@Mikolaj Would you mind removing this review here, too, or reviewing this PR?

Mikolaj:
Sure. Emily tells me our local culture is 2 reviews per PR, so let me make a quick review.

Mikolaj (approved):
Looks great.

PR #7283, opened 2021-02-18
Explicitly close file handle to avoid resource exhaustion

### LINKED ISSUE #5115: cabal init: "getDirectoryContents:openDirStream: resource exhausted (Too many open files)"
Ran into this when i ran init accidentally in the wrong directory (with a lot of subdirectories with a lot of contents). The crawl for module contents combined with lazy IO clearly ran into the usual sorts of resource limits.

This is a corner case, but in general it is better to defensively code around these things.

fendor:
Attempts to close #5115 
I think this is not the cleanest solution, but it fixes the problem, and I am happy for feedback!
---

No tests added, yet, but I am happy to add some, given a pointer where a test should be added.


cabal-install/src/Distribution/Client/Init/Heuristics.hs
```diff
@@ -156,9 +160,23 @@ scanForModulesIn projectRoot srcRoot = scan srcRoot []
...
     ignoreDir dir      = dir `elem` ["dist", "_darcs"]
 
+-- | Read the contents of the handle and parse for Language pragmas and
+-- and other module names that might be part of this project.
 findImportsAndExts :: FilePath -> SourceFileEntry -> IO SourceFileEntry
-findImportsAndExts projectRoot sf = do
-  s <- readFile (sfToFileName projectRoot sf)
+findImportsAndExts projectRoot sf =
+    -- Avoid resource exhaustion, see https://github.com/haskell/cabal/issues/5115
+    bracket
```
phadej, line 168: this is just `withFile`, isn't it?

phadej: FWIW, I'd simply read the file with `Data.ByteString.readFile`.

fendor: Thank you for the review!

fendor: And then decode the file to utf8 manually, I presume? Or re-write the parsing to use `ByteString` instead and decode only the final result?

phadej: `fromUTF8BS` is simplest way forward. i would bother rewriting the parser bits. Here performance is not critical (we only need to be careful with scarse resources = file descriptors)

emilypi (approved):

emilypi:
@fendor thanks for this! feel free to press the merge button

fgaz:
Changelog entry?

emilypi:
@fgaz for a fix this small... maybe? I guess if we're tracking it as a bug, it'd be nice to list

fendor:
A bit of a dumb question, but where is the Changelog? Is it just like it is described here: https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog

fgaz:
Yes, the changelog gets generated from those files. This way we avoid conflicts

fendor:
@fgaz Changelog entry in the right format?

cabal-install/src/Distribution/Client/Init/Heuristics.hs
```diff
@@ -156,9 +158,11 @@ scanForModulesIn projectRoot srcRoot = scan srcRoot []
     ignoreDir ('.':_)  = True
     ignoreDir dir      = dir `elem` ["dist", "_darcs"]
 
+-- | Read the contents of the handle and parse for Language pragmas and
+-- and other module names that might be part of this project.
```
fgaz, line 162: double and

fgaz (commented):
The changelog looks fine

cabal-install/src/Distribution/Client/Init/Heuristics.hs
fendor: Thanks!

fgaz (approved):

fgaz:
CI should be fixed once you rebase

PR #7288, opened 2021-02-20
Allow Literate Haskell extension in penultimate position

dpwiz:
Allows Literate Haskell files with extra format extensions like "Main.lhs.md".

Together with -pgmL option this will make such files compilable and renderable
without forcing users to make symlinks.

---

`cabal test all`


phadej:
Does GHC recognize them?

Please add a test.

phadej:
```
% ghci-9.0.1  -pgmL=markdown-unlit foo.lhs.md          
GHCi, version 9.0.1: https://www.haskell.org/ghc/  :? for help
Warning: ignoring unrecognised input `foo.lhs.md'
```

Doesn't seem to work? What I'm missing?

dpwiz:
> Doesn't seem to work? What I'm missing?

`-x lhs`

```bash
$ ghc -x lhs -pgmL=markdown-unlit Main.lhs.md -o test; ./test
hi
```

phadej:
Will users be required to add `ghc-options; -x lhs`?

I think this emphasizes the need for a test case.

phadej:
For the reference: https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/using.html?highlight=x#ghc-flag--x%20%E2%9F%A8suffix%E2%9F%A9

> Causes all files following this option on the command line to be processed as if they had the suffix ⟨suffix⟩. For example, to compile a Haskell module in the file M.my-hs, use ghc -c -x hs M.my-hs.

If I understand right, then it won't be possible to mix-and-match literate and ordinary source files.

merijn:
GHC doesn't understand different file formats, because I have a several year old (open) issue about making GHC understand compound extensions for `lhs` :p

dpwiz:
Indeed, doing this properly may require some cooperation from GHC side.

I intended to use this for single-file targets like examples/tutorials.
That could be possible right now if not for cabal checker veto.

dpwiz:
Hm. Adding test with `markdown-unlit` will add a build-time dependency. It's only `base` and `base-compat`, but... :thinking:

phadej:
`PackageDescription.Check` is used by `hackage-server` to check that broken stuff don't gets into Hackage's index. Thus if you want to allow single file targets like examples/tutorials, allow only that.

Write a test case too.

You don't need to use `markdown-unlit`, `foo.lhs.tex` is as "valid", and that GHC should handle without `-pgmL`

phadej:
To be clear. I'm not sure this is a good idea. So let us wait a bit, and think properly (and not code in fury).

dpwiz:
> GHC doesn't understand different file formats, because I have a several year old (open) issue about making GHC understand compound extensions for `lhs` :p

Found it: https://gitlab.haskell.org/ghc/ghc/-/issues/9789

merijn:
I never ended up  doing anything with it, but related: https://gitlab.haskell.org/ghc/ghc/-/issues/9789

cabal-testsuite/PackageTests/LiterateHaskell/Suffixed/test.cabal
```diff
@@ -0,0 +1,10 @@
+cabal-version: 2.2
+name: test
+version: 0
+
+executable foo
+  main-is: Main.lhs.tex
+  ghc-options: -x lhs
```
phadej, line 7: As said, I'm not fan of this.

`cabal check` should not warn about `Main.lhs.tex` only if `ghc-options: -x lhs` is present. And I don't know how to make this check robust in presence of conditionals etc.

Making symbolic links feels to be more straight forward.

I'm sorry.

dpwiz:
It actually breaks on `cabal-install`. Or is it the test suite?..

First, `preprocessFile` drops too many extensions (all of them).
I tried to coerce it by adding `lhs.tex` or skipping preprocessing helps, but then the build fails.

For some reason the `-x lhs` argument isn't passed to GHC:

    ghc -c -fPIC ....  -package-id base-4.14.1.0 ./Main.lhs.tex -hide-all-packages
    Warning: the following files would be used as linker inputs, but linking is not being done: Main.lhs.tex
    ghc: no input files

ffaf1:

PR #7295, opened 2021-02-22
Add 'preferred-versions' support for LocalIndexRepo 

### LINKED ISSUE #7294: RepoIndex with NoIndexCache should support parsing the preferred-versions file
Currently, the only repo-index-type that reads the preferred-versions file is `RepoRemote`. 

However, for testing purposes we need the `RepoLocalNoIndex` to also support parsing the `preferred-versions` file.

The plan is to allow **one** preferred-versions file in the top-level of the LocalRepo location. 

Example:
```
index/
├── base-3.0.3.1.tar.gz
│   └── base.cabal
├── base-3.0.3.2.tar.gz
│   └── base.cabal
├── base-4.0.0.0.tar.gz
│   └── base.cabal
├── template-haskell-2.3.0.0.tar.gz
│   └── template-haskell.cabal
├── template-haskell-2.3.0.1.tar.gz
│   └── template-haskell.cabal
├── template-haskell-2.4.0.0.tar.gz
|   └── template-haskell.cabal
├── noindex.cache
└── preferred-versions
```

where `preferred-versions` contains the accumulated preferred versions, similar to how it looks in https://hackage.haskell.org/packages/preferred.

The `preferred-versions` are serialised into `noindex.cache` and can be read from the cache again.

Implementation: 
Afaict, this requires changing the `Binary` Instance of `NoIndexCacheEntry` and extending it with a field such as `NoIndexCachePreference`, more closely resembling the Constructors of `IndexCacheEntry`.

For the cabal test-suite, the idea is similar: allow **one** preferred-versions file that basically is copied into the right location. example structure:
```
repo/
├── base-3.0.3.1
│   └── base.cabal
├── base-3.0.3.2
│   └── base.cabal
├── base-4.0.0.0
│   └── base.cabal
├── template-haskell-2.3.0.0
│   └── template-haskell.cabal
├── template-haskell-2.3.0.1
│   └── template-haskell.cabal
├── template-haskell-2.4.0.0
|   └── template-haskell.cabal
└── preferred-versions
```

fendor:
Closes #7294 

Generated index parses `preferred-versions` file contents.

The plan is to allow **one** preferred-versions file in the top-level of the LocalRepo location. 

Example:
```
index/
├── base-3.0.3.1.tar.gz
│   └── base.cabal
├── base-3.0.3.2.tar.gz
│   └── base.cabal
├── base-4.0.0.0.tar.gz
│   └── base.cabal
├── template-haskell-2.3.0.0.tar.gz
│   └── template-haskell.cabal
├── template-haskell-2.3.0.1.tar.gz
│   └── template-haskell.cabal
├── template-haskell-2.4.0.0.tar.gz
|   └── template-haskell.cabal
├── noindex.cache
└── preferred-versions
```

where `preferred-versions` contains the accumulated preferred versions, similar to how it looks in https://hackage.haskell.org/packages/preferred.

The `preferred-versions` are serialised into `noindex.cache` and can be read from the cache again.

Implementation: 
Afaict, this requires changing the `Binary` Instance of `NoIndexCacheEntry` and extending it with a field such as `NoIndexCachePreference`, more closely resembling the Constructors of `IndexCacheEntry`.

For the cabal test-suite, the idea is similar: allow **one** preferred-versions file that basically is copied into the right location. example structure:
```
repo/
├── base-3.0.3.1
│   └── base.cabal
├── base-3.0.3.2
│   └── base.cabal
├── base-4.0.0.0
│   └── base.cabal
├── template-haskell-2.3.0.0
│   └── template-haskell.cabal
├── template-haskell-2.3.0.1
│   └── template-haskell.cabal
├── template-haskell-2.4.0.0
|   └── template-haskell.cabal
└── preferred-versions
```

---

### DIFF
--- a/cabal-install/src/Distribution/Client/IndexUtils.hs
+++ b/cabal-install/src/Distribution/Client/IndexUtils.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -40,7 +41,10 @@ module Distribution.Client.IndexUtils (
   writeIndexTimestamp,
   currentIndexTimestamp,
 
-  BuildTreeRefType(..), refTypeFromTypeCode, typeCodeFromRefType
+  BuildTreeRefType(..), refTypeFromTypeCode, typeCodeFromRefType,
+  -- * preferred-versions utilities
+  preferredVersions, isPreferredVersions, parsePreferredVersionsWarnings,
+  PreferredVersionsParseError(..)
   ) where
 
 import Prelude ()
@@ -75,7 +79,7 @@ import Distribution.Types.PackageName (PackageName)
 import Distribution.Version
          ( Version, VersionRange, mkVersion, intersectVersionRanges )
 import Distribution.Simple.Utils
-         ( die', warn, info, createDirectoryIfMissingVerbose )
+         ( die', warn, info, createDirectoryIfMissingVerbose, fromUTF8LBS )
 import Distribution.Client.Setup
          ( RepoContext(..) )
 
@@ -87,12 +91,13 @@ import           Distribution.Solver.Types.PackageIndex (PackageIndex)
 import qualified Distribution.Solver.Types.PackageIndex as PackageIndex
 import           Distribution.Solver.Types.SourcePackage
 
+import Data.Either
+         ( rights )
 import qualified Data.Map as Map
 import qualified Data.Set as Set
 import Control.Exception
 import Data.List (stripPrefix)
 import qualified Data.ByteString.Lazy as BS
-import qualified Data.ByteString.Lazy.Char8 as BS.Char8
 import qualified Data.ByteString.Char8 as BSS
 import Data.ByteString.Lazy (ByteString)
 import Distribution.Client.GZipUtils (maybeDecompress)
@@ -556,18 +561,67 @@ extractPkg verbosity entry blockNo = case Tar.entryContent entry of
 extractPrefs :: Tar.Entry -> Maybe [Dependency]
 extractPrefs entry = case Tar.entryContent entry of
   Tar.NormalFile content _
-     | FilePath.Posix.takeFileName entrypath == "preferred-versions"
+     | isPreferredVersions entrypath
     -> Just prefs
     where
       entrypath = Tar.entryPath entry
       prefs     = parsePreferredVersions content
   _ -> Nothing
 
+------------------------------------------------------------------------
+-- Filename and parsers for 'preferred-versions' file.
+--
+
+-- | Expected name of the 'preferred-versions' file.
+--
+-- Contains special constraints, such as a preferred version of a package
+-- or deprecations of certain package versions.
+--
+-- Expected format:
+--
+-- @
+-- binary > 0.9.0.0 || < 0.9.0.0
+-- text == 1.2.1.0
+-- @
+preferredVersions :: FilePath
+preferredVersions = "preferred-versions"
+
+-- | Does the given filename match with the expected name of 'preferred-versions'?
+isPreferredVersions :: FilePath -> Bool
+isPreferredVersions = (== preferredVersions) . takeFileName
+
+-- | Parse `preferred-versions` file, ignoring any parse failures.
+--
+-- To obtain parse errors, use 'parsePreferredVersionsWarnings'.
 parsePreferredVersions :: ByteString -> [Dependency]
-parsePreferredVersions = mapMaybe simpleParsec
-                       . filter (not . isPrefixOf "--")
-                       . lines
-                       . BS.Char8.unpack -- TODO: Are we sure no unicode?
+parsePreferredVersions = rights . parsePreferredVersionsWarnings
+
+-- | Parser error of the `preferred-versions` file.
+data PreferredVersionsParseError = PreferredVersionsParseError
+    { preferredVersionsParsecError :: String
+    -- ^ Parser error to show to a user.
+    , preferredVersionsOriginalDependency :: String
+    -- ^ Original input that produced the parser error.
+    }
+  deriving (Generic, Read, Show, Eq, Ord, Typeable)
+
+-- | Parse `preferred-versions` file, collecting parse errors that can be shown
+-- in error messages.
+parsePreferredVersionsWarnings :: ByteString
+                               -> [Either PreferredVersionsParseError Dependency]
+parsePreferredVersionsWarnings =
+  map parsePreference
+  . filter (not . isPrefixOf "--")
+  . lines
+  . fromUTF8LBS
+    where
+      parsePreference :: String -> Either PreferredVersionsParseError Dependency
+      parsePreference s = case eitherParsec s of
+          Left err -> Left $ PreferredVersionsParseError
+              { preferredVersionsParsecError = err
+              , preferredVersionsOriginalDependency = s
+              }
+          Right dep -> Right dep
 
 ------------------------------------------------------------------------
 -- Reading and updating the index cache
@@ -705,10 +759,23 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
 
     entries <- handle handler $ fmap catMaybes $ for dirContents $ \file -> do
         case isTarGz file of
-            Nothing -> do
-                unless (takeFileName file == "noindex.cache" || ".cabal" `isSuffixOf` file) $
-                    info verbosity $ "Skipping " ++ file
-                return Nothing
+            Nothing
+              | isPreferredVersions file -> do
+                  contents <- BS.readFile (localDir </> file)
+                  let versionPreferencesParsed = parsePreferredVersionsWarnings contents
+                  let (warnings, versionPreferences) = partitionEithers versionPreferencesParsed
+                  unless (null warnings) $ do
+                      warn verbosity $
+                          "withIndexEntries: failed to parse some entries of \"preferred-versions\" found at: "
+                              ++ (localDir </> file)
+                      for_ warnings $ \err -> do
+                          warn verbosity $ "* \"" ++ preferredVersionsOriginalDependency err
+                          warn verbosity $ "Parser Error: " ++ preferredVersionsParsecError err
+                  return $ Just $ NoIndexCachePreference versionPreferences
+              | otherwise -> do
+                  unless (takeFileName file == "noindex.cache" || ".cabal" `isSuffixOf` file) $
+                      info verbosity $ "Skipping " ++ file
+                  return Nothing
             Just pkgid | cabalPath `Set.member` contentSet -> do
                 contents <- BSS.readFile (localDir </> cabalPath)
                 for (parseGenericPackageDescriptionMaybe contents) $ \gpd ->
@@ -725,9 +792,20 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
                     Just ce -> return (Just ce)
                     Nothing -> die' verbosity $ "Cannot read .cabal file inside " ++ file
 
+    let (prefs, gpds) = partitionEithers $ map
+            (\case
+                NoIndexCachePreference deps -> Left deps
+                CacheGPD gpd _ -> Right gpd
+            )
+            entries
+
     info verbosity $ "Entries in file+noindex repository " ++ unRepoName name
-    for_ entries $ \(CacheGPD gpd _) ->
+    for_ gpds $ \gpd ->
         info verbosity $ "- " ++ prettyShow (package $ Distribution.PackageDescription.packageDescription gpd)
+    unless (null prefs) $ do
+        info verbosity $ "Preferred versions in file+noindex repository " ++ unRepoName name
+        for_ (concat prefs) $ \pref ->
+            info verbosity ("* " ++ prettyShow pref)
 
     callback entries
   where
@@ -772,8 +850,8 @@ readPackageIndexCacheFile :: Package pkg
 readPackageIndexCacheFile verbosity mkPkg index idxState
     | localNoIndex index = do
         cache0 <- readNoIndexCache verbosity index
-        pkgs   <- packageNoIndexFromCache verbosity mkPkg cache0
-        pure (pkgs, [], emptyStateInfo)
+        (pkgs, prefs) <- packageNoIndexFromCache verbosity mkPkg cache0
+        pure (pkgs, prefs, emptyStateInfo)
 
     | otherwise = do
         cache0   <- readIndexCache verbosity index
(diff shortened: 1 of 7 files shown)


cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -1080,13 +1099,26 @@ instance Structured NoIndexCache
...
+    put (NoIndexCachePreference dep) = do
+      put (1 :: Word8)
+      put dep
 
     get = do
-        bs <- get
-        case parseGenericPackageDescriptionMaybe bs of
-            Just gpd -> return (CacheGPD gpd bs)
-            Nothing  -> fail "Failed to parse GPD"
+        -- TODO: fendor, read it out of the cache, too
```
phadej, line 1110: TODO

phadej (dismissed):
The docs change would be welcome.

I don't understand how this works.

Is it so that we can have a **single** `preferred-versions` in the local-repo directory?

cabal-testsuite/src/Test/Cabal/Prelude.hs
```diff
@@ -507,11 +507,20 @@ withRepo repo_dir m = do
     -- 2. Create tarballs
     pkgs <- liftIO $ getDirectoryContents (testCurrentDir env </> repo_dir)
     forM_ pkgs $ \pkg -> do
+        let srcPath = testCurrentDir env </> repo_dir </> pkg
+        let destPath = package_dir </> pkg
+        isPreferredVersionsFile <- liftIO $
+            if pkg == "preferred-versions"
+                then doesFileExist srcPath
```
phadej, line 514: does this check that it's not a directory? Unclear to me.

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -705,14 +706,19 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
...
     entries <- handle handler $ fmap catMaybes $ for dirContents $ \file -> do
         case isTarGz file of
-            Nothing -> do
-                unless (takeFileName file == "noindex.cache" || ".cabal" `isSuffixOf` file) $
-                    info verbosity $ "Skipping " ++ file
-                return Nothing
+            Nothing
+              | takeFileName file == "preferred-versions" -> do
+                  contents <- BS.readFile (localDir </> file)
+                  let versionPreferences = parsePreferredVersions contents
```
phadej, line 712: I'd leave atl least a note that `parsePreferredVersions` never fails. That **will** confuse users, as there is no error reporting. (It's somewhat fine that we trust that Hackage generated index is ok, users: we couldn't).

I'd rather write something which at least warns.

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -722,14 +728,18 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
...
                 case Tar.foldEntries (readCabalEntry pkgId) Nothing (const Nothing) entries of
-                    Just ce -> return (Just ce)
+                    Just ce -> return (Just [ce])
                     Nothing -> die' verbosity $ "Cannot read .cabal file inside " ++ file
 
     info verbosity $ "Entries in file+noindex repository " ++ unRepoName name
-    for_ entries $ \(CacheGPD gpd _) ->
+    let flatEntries = concat entries
+    for_ flatEntries $ \case
+      (NoIndexCachePreference dep) ->
```
phadej, line 737: unnecessary parens. (Also most of the file is 4 space indentation).

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -1052,6 +1069,7 @@ data IndexCacheEntry
 
 data NoIndexCacheEntry
     = CacheGPD GenericPackageDescription !BSS.ByteString
+    | NoIndexCachePreference Dependency
```
phadej, line 1072: I think this should be `NoIndexCachePreference [Dependency]` I.e. just single entry? I'd imagine it would make loading a lot faster.

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: thanks, actually addressed that todo, even. Just forgot to remove it

cabal-testsuite/src/Test/Cabal/Prelude.hs
fendor: indeed yes

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: That is a very good point, will update this logic

fendor:
Yes, the intention is that you have a single `preferred-versions` file, in this tree structure:
```
repo/
├── base-3.0.3.1
│   └── base.cabal
├── base-3.0.3.2
│   └── base.cabal
├── base-4.0.0.0
│   └── base.cabal
├── template-haskell-2.3.0.0
│   └── template-haskell.cabal
├── template-haskell-2.3.0.1
│   └── template-haskell.cabal
├── template-haskell-2.4.0.0
|   └── template-haskell.cabal
└── preferred-versions
```

This is parsed and put into the `noindex.cache`.

cabal-testsuite/src/Test/Cabal/Prelude.hs
fendor: My thought process was bascially to have an additional check whether this is the correct file.

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: Included warnings

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: Changed to the suggested structure

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -556,18 +561,54 @@ extractPkg verbosity entry blockNo = case Tar.entryContent entry of
...
+parsePreferredVersions = rights . parsePreferredVersionsWarnings
+
+-- | Parse `preferred-versions` file, collecting parse errors that can be shown
+-- in error messages. Error message is the line that failed to parse.
+parsePreferredVersionsWarnings :: ByteString -> [Either String Dependency]
+parsePreferredVersionsWarnings =
+  map parsePreference
+  . filter (not . isPrefixOf "--")
+  . lines
+  . BS.Char8.unpack -- TODO: Are we sure no unicode?
```
fendor, line 606: This is actually wrong in my opinion. Can we fix that in this PR, or rather a follow-up?

phadej: you may use `fromUTF8BS` here, I don't think we need to be more elaborate here.

fendor:
Where should this be documented? To me, it looks like the `preferred-versions` file is basically not documented.

phadej:
Correct, `preferred-versions` is an implementation detail of indices which normal users never interact with. But now you are exposing this functionality to them.

I think mentioning that `preferred-versions` file is recognized, looks like a `build-depends` field (though that is incorrect, it should accept `Dependency` but a `PackageConstraint` - that is different story though), and give an example in https://cabal.readthedocs.io/en/3.4/installing-packages.html?highlight=local-repo#local-no-index-repositories would be enough.

fendor:
Thank you, it is documented now.

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -556,18 +560,54 @@ extractPkg verbosity entry blockNo = case Tar.entryContent entry of
...
+-- in error messages. Error message is the line that failed to parse.
+parsePreferredVersionsWarnings :: ByteString -> [Either String Dependency]
+parsePreferredVersionsWarnings =
+  map parsePreference
+  . filter (not . isPrefixOf "--")
+  . lines
+  . fromUTF8LBS
+    where
+      parsePreference :: String -> Either String Dependency
+      parsePreference s = case simpleParsec s of
```
phadej, line 608: FYI, there  is `eitherParsec` which returns an error why something failed parsed. Maybe use it? Something like

```haskell
Left err -> Left (s, e)
```

and prettify that when printing

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -726,8 +778,12 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
                     Nothing -> die' verbosity $ "Cannot read .cabal file inside " ++ file
 
     info verbosity $ "Entries in file+noindex repository " ++ unRepoName name
-    for_ entries $ \(CacheGPD gpd _) ->
-        info verbosity $ "- " ++ prettyShow (package $ Distribution.PackageDescription.packageDescription gpd)
+    for_ entries $ \case
+        NoIndexCachePreference deps ->
+            for_ deps $ \dep ->
+              info verbosity ("* " ++ prettyShow dep)
```
phadej, line 784: I think we need clearer separation.

I'd rather go over entries twice (or `partitionEithers` or ...) and print preferred versions independently under their own header (i.e. `info verbosity $ "Prefferred versions in ...`).

doc/installing-packages.rst
```diff
@@ -185,6 +185,27 @@ The part of the path will be used to determine the cache key part.
...
+then package deprecations and preferences will be taken into account by the solver.
+
+The contents of ``preferred-versions`` is a list of package version constraints, e.g.
+::
+
+    binary < 0.8.0.0 || > 0.8.0.0
+    text == 1.2.0.0
+
+thus, looks similar to a ``package-name.cabal``'s ``build-depends`` section.
+
```
phadej, line 208: I'm not sure where to add this, and whether it should be mentioned at all, but this adds a new (useful?) side-effect; We can have a `file+noindex` repository with *just* `preferred-versions` and that way affect other repositories (e.g. Hackage) indices.

There is a comment *-- Note: preferences combined without using CombineStrategy*, so there is no control over that *yet*.

phadej (commented):
This is getting in shape, great!

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: I extended it to a proper type, since I disliked the tuple.

cabal-install/src/Distribution/Client/IndexUtils.hs
fendor: Output is now better grouped

cabal-install/tests/UnitTests/Distribution/Client/IndexUtils.hs
```diff
@@ -0,0 +1,79 @@
+module UnitTests.Distribution.Client.IndexUtils where
```
fendor, line 1: Added these simple unit-tests for `preferred-versions` parser, since I thought, they should have some tests.

Therefore, we have to expose a bit more from `IndexUtils` to be able to test it.

doc/installing-packages.rst
fendor: Imo, it shouldn't be mentioned yet. At least some tests would be necessary before we mention it.

phadej: But they are affecting Hackage, I'm pretty sure. If we can be sure what will be in the future, there should be a warning.

fendor: I can't find anything obvious about merging repositories (maybe hidden or deeper in the guts?). Can you point me to some Module or something else I should take a closer look at?

phadej: https://github.com/haskell/cabal/blob/89df281c3f74c0976f6b95524fd85c468f77ce1a/cabal-install/src/Distribution/Client/IndexUtils.hs#L301-L332

fendor: Thanks, I will investigate!

fendor: Took me long enough, but I can finally confirm that indeed, the `preferred-versions` file can be used to restrict the package set from Hackage. Will update the documentation.

fendor: No idea whether it is really possible to add a test for it.

fendor:
@emilypi ping

fendor:
@emilypi ping again as requested!

cabal-install/src/Distribution/Client/IndexUtils.hs
```diff
@@ -705,10 +759,23 @@ withIndexEntries verbosity (RepoIndex _repoCtxt (RepoLocalNoIndex (LocalRepo nam
...
+              | isPreferredVersions file -> do
+                  contents <- BS.readFile (localDir </> file)
+                  let versionPreferencesParsed = parsePreferredVersionsWarnings contents
+                  let (warnings, versionPreferences) = partitionEithers versionPreferencesParsed
+                  unless (null warnings) $ do
+                      info verbosity $
+                          "withIndexEntries: failed to parse some entries of \"preferred-versions\" found at: "
+                              ++ (localDir </> file)
+                      for_ warnings $ \err -> do
+                          info verbosity $ "* \"" ++ preferredVersionsOriginalDependency err
```
emilypi, line 772: should these warnings be logged at `warn`?

emilypi (approved):
A few comments that aren't blockers

cabal-testsuite/src/Test/Cabal/Prelude.hs
Mikolaj: Actually, I think the blocker may be @phadej's review status set to 'request changes'. The review would need to be finished by Oleg or dismissed by an admin (e.g., me) to unblock this. Nevertheless, let me run CI again before asking Oleg for help.

Mikolaj: Resolving didn't work, so on @phadej's request I 'dismissed' the review and restarted the CI. Let's see...

fendor:
No idea, it used to work...

Mikolaj:
@emilypi, @fgaz: it seems CI is acting up. Github status is green. Any ideas? I have no experience with github CI, but I will troubleshoot if nobody has.

phadej:
@Mikolaj on https://github.com/haskell/cabal/actions/runs/854204943 the UI shows 
Screenshot from 2021-05-20 12-31-30

So something is broken with GHA, but whether its our or their fault - no idea

Mikolaj:
@phadej: how do you display that info? I see no such thing at the address you gave, but I'm clueless about github CI.

Mikolaj:
@fendor: could you `push -f` with a trivial change (e.g., extra space in a commit message) to try and unblock the hairy stuff (I know I'm desparate)?

Mikolaj:
@fgaz has shown me that github status caught up with reality and reported Github Action outages. So we probably just need to wait until it gets fixed. Still, a pity there are not text log files that show at which point it broke and with what error code.

Mikolaj:
@fendor: CI is apparently still stuck (I restarted it a couple of times) despite github having been repaired and CI working fine on my other PR. Could you please `push -f` a trivial change to prod the CI?

fendor:
will do!

PR #7296, opened 2021-02-22
Cabal outdated should not list deprecated versions

### LINKED ISSUE #7249: `cabal outdated` lists deprecated package versions
For e.g. [Agda-2.6.1.2](https://hackage.haskell.org/package/Agda-2.6.1.2) `cabal outdated` includes `binary`
```
$ cabal outdated
Outdated dependencies:
binary >=0.8.3.0 && <0.9 (latest: 0.10.0.0)
```
but both latest versions 0.9.0.0 and 0.10.0.0 of `binary` are deprecated, see https://hackage.haskell.org/package/binary/preferred .
> Deprecating a version adds a range which excludes just that version. 

So the effective latest version of `binary` is still 0.8.8.0 and `outdated` shouldn't print anything for `binary`.

Reproduce:
```
cabal unpack Agda-2.6.1.2
cd Agda-2.6.1.2
cabal outdated
```

Cabal versions 3.2.0.0 and 3.5.0.0 (5 January 2021).

fendor:
Closes #7249 

Essentially also takes a look at the `preferred-versions` file and prunes allowed versions accordingly.

~~PR depends on #7295 to write tests for it.~~

---


cabal-install/src/Distribution/Client/Outdated.hs
```diff
@@ -168,25 +172,33 @@ depsFromPkgDesc verbosity comp platform = do
 data ListOutdatedSettings = ListOutdatedSettings {
   -- | Should this package be ignored?
   listOutdatedIgnorePred :: PackageName -> Bool,
-  -- | Should major version bumps should be ignored for this package?
+  -- | Should major version bumps be ignored for this package?
```
emilypi, line 175: Question about these docs: do we mean minor version bumps here? Just judging by the name, i'd expect that to be minor rather than major.

emilypi (approved):
LGTM - one question that isn't a blocker for the PR

emilypi:
@fendor do you mind rebasing? should fix the CI failure

fendor:
This is not done anyways yet, as it depends on #7295 which needs some form of documentation for its newly added feature

cabal-install/src/Distribution/Client/Outdated.hs
fendor: Just checked, weirdly enough, this comment is correct. What it does is to check whether there is any *minor* version update and ignores any major version bumps. 

E.g. in the case of `binary == 0.8.5.0`, `cabal outdated --minor` reports:
```
Outdated dependencies:
binary ==0.8.5.0 (latest: 0.8.8.0)
```

fendor:
Ready for review

fendor:
@Mikolaj Would you mind removing this review here, too, or review this PR?

cabal-install/src/Distribution/Client/Outdated.hs
Mikolaj: I guess this may make sense: if minor updates are non-breaking bug-fixes, we don't want to miss any, while updating to breaking major versions is not something we want to be nagged about.

Mikolaj (approved):
Great TDD. LGTM.

cabal-install/src/Distribution/Client/Types/SourcePackageDb.hs
```diff
@@ -21,3 +26,38 @@ data SourcePackageDb = SourcePackageDb
...
+
+-- | Does a case-sensitive search by package name and a range of versions.
+--
+-- We get back any number of versions of the specified package name, all
+-- satisfying the version range constraint.
+--
+-- Additionally, `preferred-versions` (such as version deprecation) are
+-- honoured in this lookup, which is the only difference to
+-- 'PackageIndex.lookupDependency'
+lookupDependency :: SourcePackageDb -> PackageName -> VersionRange -> [UnresolvedSourcePackage]
```
fendor, line 38: Is this the right location for this kind of function?

cabal-install/src/Distribution/Client/Types/SourcePackageDb.hs
```diff
@@ -21,3 +26,38 @@ data SourcePackageDb = SourcePackageDb
...
+  where
+    pref = Map.lookup pname (packagePreferences sourceDb)
+
+
+-- | Does a case-sensitive search by package name.
+--
+-- Additionally, `preferred-versions` (such as version deprecation) are
+-- honoured in this lookup, which is the only difference to
+-- 'PackageIndex.lookupPackageName'
+lookupPackageName :: SourcePackageDb -> PackageName -> [UnresolvedSourcePackage]
```
fendor, line 50: Also, afaict there should be other codepaths that would benefit from these functions (e.g. list). Should I review those in a follow-up PR to unify logic?

cabal-install/src/Distribution/Client/Types/SourcePackageDb.hs
Mikolaj: I think so. The first used type is defined in this module and there are precedents of functions in such modules. Also, you didn't need to import any far away modules to make it work.

cabal-install/src/Distribution/Client/Types/SourcePackageDb.hs
Mikolaj: IMHO, yes, please.

cabal-install/src/Distribution/Client/Types/SourcePackageDb.hs
```diff
@@ -1,16 +1,21 @@
...
 
 import Distribution.Types.PackageName  (PackageName)
-import Distribution.Types.VersionRange (VersionRange)
+import Distribution.Types.VersionRange (VersionRange, withinRange)
+import Distribution.Package            (packageVersion)
 
 import Distribution.Client.Types.PackageLocation (UnresolvedSourcePackage)
+import qualified Distribution.Solver.Types.PackageIndex as PackageIndex
 import Distribution.Solver.Types.PackageIndex    (PackageIndex)
+import qualified Data.Map as Map
```
Mikolaj, line 18: I'm not sure if we have a consistent convention about grouping imports, but I'm quite sure an import from outside our packages should be separated from imports from our package(s).

emilypi (approved):

emilypi:
@fendor do you wanna update this branch, and then merge on green?

fendor:
@tapas rebased!

Issue #7290, opened 2021-02-21
Globbing support for mixins and rexported-modules


ekmett:
**Describe the bug**
Using backpack often requires using large large numbers of reexported-modules stanzas and mixins stanzas.

If you use a backpack package and that package produces 17 outputs in some namespace, I have to repeat myself 17 times in the output to rename it in the mixins statement, and another 17 times in the rexported-modules stanza.

If I then go to use that backpack package 17x on another than provides 17 definitions, I'm now stuck with 289, and the exponent keeps growing from there. I'm currently using a program to produce my cabal files. This is not "the Way". 

I don't want to use a custom setup to produce the boilerplate, because @phadej will be mad at me, but also because it gets in the way of incremental recompilation in this mode, as I understand the world.

This is currently the biggest blocker I have to using backpack 'in bulk' in my code.

I'd like to be able to rename in the mixins statement on the provides/requires sides with a wildcard notation of some sort:

in a sort of Makefile-like vocabulary:

`consDefs (% as Foo.%) requires (% as Baz.%)`

or since lowercase identifiers aren't a thing:

`mixins: consDefs (x as Foo.x) requires (x as Baz.x)`

or

`mixins: consDefs (* as Foo.*) requires (* as Baz.*)`

which gives an option of allowing recursive globbing or not

`mixins: consDefs (** as Foo.**) requires (** as Baz.**)`

Similarly,

```
reexported-modules:
  Foo.x as Bar.Internal.x
```

is practically required, as I have to re-export the things I get from all these bulk-work backpack modules.

options include using * and ** for recursive globbing or not. Other options are using something like a scheme like parameter pack syntax. I don't care what the syntax is, I just want _something_.

This would require cabal to be able to query a backpacked package for what signatures it has outstanding.

**To Reproduce**
Try to write large backpacked packages.

**Alternatives explored**
I am currently generating the entire cabal file with software, like I do with the `gl` package.

danidiaz:
> This would require cabal to be able to query a backpacked package for what signatures it has outstanding.

This would be useful on its own.

ezyang:
cc'ing myself

ezyang:
It'll be helpful to know exactly how the use case in question is working, so I can more easily tell if the specific request here is the right way to go about doing it.

[...]

PR #7303, opened 2021-02-24
Support qualified module renamings

ezyang:
Qualified module renamings let you bring all modules from a library
into scope, but qualified under some module prefix.  If you write
'pkg qualified Prefix', then if pkg exposes A and B, you will be
able to access them as Prefix.A and Prefix.B.  This functionality
doesn't require any GHC changes; Cabal takes care of desugaring
the qualified syntax into an explicit list of renamings.

Partially address #7290

---

### DIFF
--- a/Cabal/src/Distribution/Types/ModuleRenaming.hs
+++ b/Cabal/src/Distribution/Types/ModuleRenaming.hs
@@ -41,6 +41,11 @@ data ModuleRenaming
         -- | Hiding renaming, e.g., @hiding (A, B)@, bringing all
         -- exported modules into scope except the hidden ones.
         | HidingRenaming [ModuleName]
+        -- | Qualified renaming, @(qualified P)@, brining all
+        -- exported modules into scope with P prefix.  So if M
+        -- was provided by the package, it is now in scope as
+        -- P.M
+        | QualifiedRenaming ModuleName
     deriving (Show, Read, Eq, Ord, Typeable, Data, Generic)
 
 -- | Interpret a 'ModuleRenaming' as a partial map from 'ModuleName'
@@ -54,6 +59,8 @@ interpModuleRenaming (ModuleRenaming rns) =
 interpModuleRenaming (HidingRenaming hs) =
     let s = Set.fromList hs
     in \k -> if k `Set.member` s then Nothing else Just k
+interpModuleRenaming (QualifiedRenaming prefix) =
+    \k -> Just (prefix `joinModuleName` k)
 
 -- | The default renaming, if something is specified in @build-depends@
 -- only.
@@ -79,6 +86,8 @@ instance Pretty ModuleRenaming where
   pretty DefaultRenaming = mempty
   pretty (HidingRenaming hides)
         = text "hiding" <+> parens (hsep (punctuate comma (map pretty hides)))
+  pretty (QualifiedRenaming prefix)
+        = text "qualified" <+> pretty prefix
   pretty (ModuleRenaming rns)
         = parens . hsep $ punctuate comma (map dispEntry rns)
     where dispEntry (orig, new)
@@ -109,7 +118,7 @@ moduleRenamingParsec
     -> m ModuleRenaming
 moduleRenamingParsec bp mn =
     -- NB: try not necessary as the first token is obvious
-    P.choice [ parseRename, parseHiding, return DefaultRenaming ]
+    P.choice [ parseRename, parseHiding, parseQualified, return DefaultRenaming ]
   where
     cma = P.char ',' >> P.spaces
     parseRename = do
@@ -121,6 +130,11 @@ moduleRenamingParsec bp mn =
         P.spaces -- space isn't strictly required as next is an open paren
         hides <- bp (P.sepBy mn cma)
         return (HidingRenaming hides)
+    parseQualified = do
+        _ <- P.string "qualified"
+        P.spaces
+        prefix <- mn
+        return (QualifiedRenaming prefix)
     parseList =
         P.sepBy parseEntry cma
     parseEntry = do
--- a/Cabal/src/Distribution/Backpack/UnifyM.hs
+++ b/Cabal/src/Distribution/Backpack/UnifyM.hs
@@ -487,6 +487,13 @@ convertInclude ci@(ComponentInclude {
     req_rename_list <-
       case req_rns of
         DefaultRenaming -> return []
+        QualifiedRenaming _ -> do
+            -- TODO: support this for requires (req_rename_list
+            -- is a bad representation for this, alas, need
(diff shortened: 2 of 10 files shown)


ezyang:
This probably needs a version guard somewhere

ezyang:
And it also needs docs

ekmett:
LGTM. I'd like a more general globbing solution, but won't let the perfect be the enemy of 'hey we could ship this and improve the situation'

ezyang:
@phadej I'm not exactly sure how I'm supposed to operate the Described regex machinery.

Currently, IncludeRenaming is specified as:

```
instance Described IncludeRenaming where
    describe _ = mr <> REOpt (RESpaces <> "requires" <> RESpaces1 <> mr)
      where
        mr = describe (Proxy :: Proxy ModuleRenaming)
```

In particular, the space before requires is optional. This is currently unambiguous because prior to this PR `mr` never produces a token that would be ambiguous if cat'ed with requires. However, with this PR, an unparenthesized module name is permitted, in which case the space is mandatory, and this regex will produce garbage test input without it (failing propParsec).

However, I *can't* just modify the RESpaces to be RESpaces1, because the regular expressions are also being used to recognize valid strings, and it is permitted to omit the space between a trailing parentheses and requires.

So I could fix this by sort of just reimplementing the IncludeRenaming regex from scratch, inlining ModuleRenaming into it as necessary to resolve the ambiguity, but this seems very unsatisfactory. Is there an intended recipe for resolving this situation?

phadej (changes requested):
This should be guarded by `cabal-version` checks. We cannot extend syntax of old specs after they are allowed on Hackage.

phadej:
Re regexes. I'd write regex from scratch. Maybe having two braches, one with parens and once without.

Cabal/src/Distribution/Types/ModuleRenaming.hs
```diff
@@ -121,6 +130,11 @@ moduleRenamingParsec bp mn =
         P.spaces -- space isn't strictly required as next is an open paren
         hides <- bp (P.sepBy mn cma)
         return (HidingRenaming hides)
+    parseQualified = do
+        _ <- P.string "qualified"
```
phadej, line 134: This is wrong. It would allow `qualifiedBlabla`

phadej:
I'd ask you to discuss proposed syntax publicly, at least on `cabal-devel` list, or better some wider place. I *hope* there are other backpack users than @ekmett.

I'm feel uncomfortable you jumped straight into implementation. (Though it's good to try if an idea is even viable).

Cabal/src/Distribution/Types/ModuleRenaming.hs
ezyang: some day I will remember that Cabal's parser doesn't actually have a lexer stage before hand  🤣

ezyang:
@phadej Sorry, the intention wasn't to ram this PR through as is without discussion, I wanted to show ekmett that the change he wanted was easy to do (I renamed the subject to add POC to make this clearer). The cabal-devel mailing list seems pretty moribund these days but I'm happy to send an email there.

Cabal/src/Distribution/Types/ModuleRenaming.hs
phadej: Nor `parsec` has auto-`try`. (But `Described` generative tests should catch that)

Cabal-described/src/Distribution/Described.hs
```diff
@@ -414,10 +414,28 @@ instance Described ForeignLibType where
     describe _ = REUnion ["native-shared","native-static"]
 
 instance Described IncludeRenaming where
-    describe _ = mr <> REOpt (RESpaces <> "requires" <> RESpaces1 <> mr)
+    -- Unfortunately, we can't directly use ModuleRenaming, as
+    -- there is some ambiguity in the grammar that we can't express
```
phadej, line 418: Is this an expressivity problem of a framework, or is there an actual ambiguity in the grammar? I might sound like a SPJ, but can you write down the BNF-grammar then (assuming there is munching lexer, so without need to worry about spaces?)

phadej: Note, these regexes are virtually only documentation of backpack features (re: https://github.com/haskell/cabal/issues/4761).

And while `mixin` syntax definition in https://cabal.readthedocs.io/en/3.4/buildinfo-fields-reference.html is technically correct, it goes beyond what I'd consider "human understandable".

danidiaz:
How would this proposed syntax work with reexported-modules and signatures?

Cabal-described/src/Distribution/Described.hs
ezyang: > Is this an expressivity problem of a framework, or is there an actual ambiguity in the grammar?

Ooh, I used a bad turn of phrase here. Ambiguity isn't the right word.

First, here is the EBNF you requested:

```
IncludeRenaming ::= ModuleRenaming { "requires" ModuleRenaming }
ModuleRenaming ::=
    (* empty *)
  | "(" Renaming ")"
  | "hiding" "(" ModuleList ")"
  | "qualified" ModuleName

RenamingList ::=
    Renaming
  | Renaming "," RenamingList

Renaming ::=
    ModuleName
  | ModuleName "as" ModuleName

ModuleList ::=
    ModuleName
  | ModuleName "," ModuleList
```

OK, now to explain the problem here. The problem I am having with Distribution.Described lies *solely* in whitespace handling, as I mentioned in https://github.com/haskell/cabal/pull/7303#issuecomment-785584905 The problem is that it is only valid to omit the space before requires (RESpaces) *if* you didn't select the "qualified" production in ModuleRenaming. If you assume that everything is tokenized beforehand, the problem disappears because `M.Nrequires` will properly tokenize as a single module name, while `)requires` will tokenize as two tokens.

> Note, these regexes are virtually only documentation of backpack field syntax (re: #4761).

Yes, no excuse for this. I guess now is a good time to nail that issue :)

ezyang:
@danidiaz

> How would this proposed syntax work with reexported-modules and signatures?

So, the second part (not included in this PR) is an extension to reexported-modules to permit globbing of reexports (this is not relevant to signatures as we don't require you to write all the requirements in a package, we'll infer them). But this is substantially more fraught because this is tantamount to doing #5343 which means that we have to shave the "update cabal file when you sdist" yak first.

Cabal-described/src/Distribution/Described.hs
ezyang: @phadej Backpack docs exist and landed!

Kleidukos:

ffaf1:

ekmett:
I'm sad that we just let this one rot on the vine.

PR #7308, opened 2021-02-28
Let CI jobs run on all PRs, not just those rooted off master.

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* **#7308 Let CI jobs run on all PRs, not just those rooted off master.**

I want to use stack diffs (https://github.com/ezyang/ghstack/) to
submit PRs to Cabal.  Stack diffs, simply by virtue of what they
are, are PRs that are targeted to branches that are not master (they
target an appropriate base that lets GitHub render only the change
in question, even if there are other changes below it.)  However,
I don't get CI jobs on those stack diffs because the current
pull_request branch selector is too restrictive.  This PR makes
it less restrictive.

This PR is submitted using ghstack to validate that this indeed
works.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -4,8 +4,6 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
   release:
     types:
       - created
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -10,8 +10,6 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
   release:
     types:
       - created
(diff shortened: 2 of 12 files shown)


emilypi:
@fgaz thoughts on this one?

fgaz:
release branches do get ci though.. how does that happen? how does this interact wit it? also bootstrap is failing

ezyang:
@fgaz IDK how release branches work; my reading of the Actions code is that they shouldn't have stuff running. But whatever this PR does, it shouldn't interfere with release branches (I'm only running more jobs, not less)

Bootstrap has been failing for a while, it's unrelated to this PR.

emilypi (approved):
Can confirm that bootstrap CI just kind of does its own thing. Shouldn't block this PR though - we'll work it out in a separate PR.

fgaz:
@ezyang could you delete the gh/ezyang/2 branches?

ezyang:
oops i botched this merge

PR #7310, opened 2021-03-01
Canonicalize path in TargetSelector to fix No Targets bug on Windows

frabrooks:
On line 1752 in TargetSelector a call to `getCurrentDirectory` is not
canonicalized which prevents cabal-install from finding the primary
package on Windows when no targets are specified.

This happens because on line 1772 the `cwd` is compared as a string with
pinfoDirectory path which is canonicalized somewhere so we get:

       "c:\myProject\"  /=  "C:\myProject\"

This only happens under certain shells but it does completely break
cabal functionality in emacs' haskell-mode (on Windows) as it relies
on automatic package detection.

Only changed one line (adding the call to canonicalize path):

```
-  1752      cwd   <-  getCurrentDirectory
+  1752      cwd   <- canonicalizePath =<< getCurrentDirectory
```
---

---

Please also shortly describe how you tested your change:

 - Still compiles and fixes the bad behaviour and doesn't break anything far as I can see.


FraserBrooks:
See my original stack overflow posts for the full breakdown of how this breaks emacs' haskell-mode:

https://stackoverflow.com/questions/65606600/emacs-haskell-mode-with-cabal-project-unrecognised-target-syntax-for-ghc-o
https://stackoverflow.com/questions/65973200/how-does-cabal-work-out-what-directory-it-is-in-and-look-for-targets

Eason0210:
I have the same issue with GHC 8.10.5 + cabal 3.4.0.0  on Windows 10 Emacs 28.0.50, dante doesn't work with cabal project.
get the error as below:
```
cabal-3.4.0.0.exe: No targets given and there is no package in the current
directory. Use the target 'all' for all packages in the project or specify
packages or components by name or location. See 'cabal build --help' for more
details on target options.
```

jneira:
> Still compiles and fixes the bad behaviour and doesn't break anything far as I can see.

Hi, thanks for fixing this for windows. It needs more love than other os's :slightly_smiling_face: 
Would be any chance to add a regression test on this? Would you need any help to start writing it?

jneira:
> This only happens under certain shells but it does completely break
cabal functionality in emacs' haskell-mode (on Windows) as it relies
on automatic package detection.

What shells would be affected? powershell, dos,..?

jneira:
@Mikolaj i cant rerun the bootstrap on linux job, it seems a transient error though

Mikolaj:
> @Mikolaj i cant rerun the bootstrap on linux job, it seems a transient error though

OK. Apparently I can't rerun this time, either. However, once the branch is rebased, the tests should run again.

FraserBrooks:
> > This only happens under certain shells but it does completely break
> > cabal functionality in emacs' haskell-mode (on Windows) as it relies
> > on automatic package detection.
> 
> What shells would be affected? powershell, dos,..?

Actually, IIRC this behaviour presents *in and only in* emacs but no matter what shell one used or at least in all the different shells I tried (Powershell, CMD/DOS, Git Bash, Cygwin). 

I got as far as suspecting something odd is happening with the elisp `call-process` function:

The weird thing was that `(call-process shell ... "cabal build" )` returns the 'no-targets' error, but `(call-process shell ... "cd  .  ; cabal build")` worked just fine.

Somehow that `cd .` changes the format of the file path that cabal ends up using to look for targets.

Either way, the fix above solves things so I left it there.

Eason0210:
> > What shells would be affected? powershell, dos,..?
> 
> Actually, IIRC this behaviour presents in _and only in_ emacs but no matter what shell one used or at least in all the different shells I tried (Powershell, CMD/DOS, Git Bash, Cygwin).

I tested Cabal 3.4.0.0 with GHC 8.10.4  on Powershell  and  CMD , got the same issue.  
But when using with GHC 8.10.5 , it work well on  Powershell  and  CMD,  only got issue when using with Emacs ( such as eshell).

frabrooks:
> I tested Cabal 3.4.0.0 with GHC 8.10.4 and 8.10.5 on Powershell and CMD , got the same issue.

@Eason0210  Sorry, just for clarification: Do you mean, *in emacs* with your preferred shell set to Powershell/CMD, or just the usual terminals?  As I wasn't able to reproduce the problem outside of emacs

frabrooks:
> Would be any chance to add a regression test on this? Would you need any help to start writing it? 

@jneira I had a look at adding a regression test, but I'm a little out of my depth. I'm not sure what one would even test.

The bug is ultimately a result of not canonicalizing a call to `getCurrentDirectory` the result of which is then immediately used in a comparison. Would we want to assert that a search for a primary package doesn't fail due to file path capitalization on windows?

Afraid I don't know enough about the testing apparatus to go about implementing this currently.

Not to mention, the fact that it seems the bug only manifests *in emacs* and *on windows*, not sure how that would factor into writing an integration test

Eason0210:
> Do you mean, _in emacs_ with your preferred shell set to Powershell/CMD, or just the usual terminals?

Hi @frabrooks  I tested just the usual terminals without Emacs, Powershell/CMD.  
But it is very weird, I want to reproduce it again. now it worked outsite of Emacs.
```
➜ ~  mkdir haskell

    Directory: C:\Users\Aqua

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----            8/1/2021  4:28 PM                haskell

➜ ~
➜ ~  cd .\haskell\
➜ haskell  ls

➜ haskell
➜ haskell  mkdir testcabal

    Directory: C:\Users\Aqua\haskell

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----            8/1/2021  4:28 PM                testcabal

➜ haskell  cabal --version
cabal-install version 3.4.0.0
➜ haskell  ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.0.1
➜ haskell  ghcup --version
The GHCup Haskell installer, version v0.1.16.1
➜ haskell  cd .\testcabal\
➜ testcabal  ls

➜ testcabal  cabal init

Guessing dependencies...

Generating LICENSE...
Generating CHANGELOG.md...
Generating app\Main.hs...
Generating testcabal.cabal...

➜ testcabal  cabal run
Resolving dependencies...
Build profile: -w ghc-9.0.1 -O1
 - testcabal-0.1.0.0 (exe:testcabal) (first run)
Configuring executable 'testcabal' for testcabal-0.1.0.0..
Preprocessing executable 'testcabal' for testcabal-0.1.0.0..
Building executable 'testcabal' for testcabal-0.1.0.0..
[1 of 1] Compiling Main             ( app\Main.hs, C:\Users\Aqua\haskell\testcabal\dist-newstyle\build\x86_64-windows\ghc-9.0.1\testcabal-0.1.0.0\x\testcabal\build\testcabal\testcabal-tmp\Main.o )
Linking C:\Users\Aqua\haskell\testcabal\dist-newstyle\build\x86_64-windows\ghc-9.0.1\testcabal-0.1.0.0\x\testcabal\build\testcabal\testcabal.exe ...
Hello, Haskell!
```

jneira:
> @jneira I had a look at adding a regression test, but I'm a little out of my depth. I'm not sure what one would even test.

dont worry

But i am little bit worried by introducing `canonicalizePath` and their possible undesired effects. The last comment by @Eason0210  is decisive though (thanks!), as the issue can be at least reproduced outside emacs. A first step for a possible test.

jneira (approved):
LGTM, altough i would like to read the opinion of a more experienced maintainer (@phadej?)

jneira:
> 
> 
> LGTM, altough i would like to read the opinion of a more experienced maintainer (@phadej?)

@fgaz?

Mikolaj:
@jneira: next time try "young, but experienced" ;)

hasufell:
Alright, so I checked the code and I believe we don't want to run `canonicalizePath` there.

The actual culprit is:

https://github.com/haskell/cabal/blob/5390fa0bc9a5477981dd63ca5950882f0abc6ac6/cabal-install/src/Distribution/Client/TargetSelector.hs#L1771-L1772

This is *always* wrong. However, [equalFilePath](https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:equalFilePath) will already fix it for the use case you described.

The only thing where it would still break is if the current directory is a symbolic link. It's questionable whether that's a practical issue on windows. `canonicalizePath` will btw. not resolve junction points and `..` will not be returned by `getCurrentDirectory` anyway, so the only thing it canonicalizes is symbolic links.

I'd suggest to provide a helper function:

```hs
compareFilePath :: FilePath -> FilePath -> IO Bool
compareFilePath fp1 fp2
  | equalFilePath fp1 fp2 = pure True
  | otherwise = do
    c1 <- canonicalizePath fp1
    c2 <- canonicalizePath fp2
    pure $ equalFilePath c1 c2
```

And call it exactly where we do the comparison, not some place earlier.

Mikolaj:
@frabrooks: can I assist you in any way in moving this forward?

frabrooks:
@hasufell  Thanks for the suggestion. I've was going to provide the helper function you suggested inside `Client/Utils.hs` but it seems that IO, at least in `TargetSelector` needs to go through `DirActions` (the default of which is exactly `System.Directory.canonicalizePath`). So I changed the function to reflect this and included it in `TargetSelector.hs`:

https://github.com/haskell/cabal/blob/ad187a92a7bad124725e7e3f0f12ff363fb138ba/cabal-install/src/Distribution/Client/TargetSelector.hs#L2169-L2178

 `selectPrimaryPackage` now uses that util function rather than `dir == cwd`:

https://github.com/haskell/cabal/blob/ad187a92a7bad124725e7e3f0f12ff363fb138ba/cabal-install/src/Distribution/Client/TargetSelector.hs#L1766-L1773

Unless I missed a trick, there didn't seem to be a way to rewrite `selectPrimaryPackage` and still use `partition` with the comparison wrapped in the IO. So it's defined recursively instead, in a way that will always produce the same result as `partition`.

Also, as you mentioned it, I did some tests and it looks like `canonicalizePath` actually resolves junctions too. At least on Windows 10:

```
PS F:\drive\test\junctionFoler\test> ghci
GHCi, version 8.10.4: https://www.haskell.org/ghc/  :? for help
Prelude> import System.Directory
Prelude System.Directory> getCurrentDirectory
"F:\\drive\\test\\junction\\test"
Preluse System.Directory> canonicalizePath =<< getCurrentDirectory
"F:\\drive\\test\\foo\\bar\\test"
```

Mikolaj:
> > @frabrooks: can I assist you in any way in moving this forward?
> 
> You can let me know if I've committed any sins against Haskell now that I'm inserting more than a few characters smile

WIll do. Just remember that we are all sinners.

cabal-install/src/Distribution/Client/TargetSelector.hs
```diff
@@ -1763,14 +1763,15 @@ getKnownTargets dirActions@DirActions{..} pkgs = do
...
-          | dir == cwd = True
-        isPkgDirCwd _  = False
+    selectPrimaryPackage _ [] = return ([] , []) 
+    selectPrimaryPackage cwd (pkg : packages) = do
+      (ppinfo, opinfo) <- selectPrimaryPackage cwd packages
+      let mDir = fst <$> (pinfoDirectory pkg)
+      isPkgDirCwd <- maybe (pure False) (compareFilePath dirActions cwd) mDir
+      return (case isPkgDirCwd of
+                True  -> ( pkg : ppinfo , opinfo )
+                False -> ( ppinfo , pkg : opinfo ))
```
Mikolaj, line 1773: This can be just an `if` expression.

cabal-install/src/Distribution/Client/TargetSelector.hs
```diff
@@ -1763,14 +1763,15 @@ getKnownTargets dirActions@DirActions{..} pkgs = do
...
   where
-    selectPrimaryPackage :: FilePath
-                         -> [KnownPackage]
-                         -> ([KnownPackage], [KnownPackage])
-    selectPrimaryPackage cwd = partition isPkgDirCwd
-      where
-        isPkgDirCwd KnownPackage { pinfoDirectory = Just (dir,_) }
-          | dir == cwd = True
-        isPkgDirCwd _  = False
+    selectPrimaryPackage _ [] = return ([] , [])
```
Mikolaj, line 1766: Type signature here wouldn't hurt. Add `forall a.` to the type signature above it make it possible. And tnen `ScopedTypeVariables` to file extensions at the top, if it yells at you.

cabal-install/src/Distribution/Client/TargetSelector.hs
```diff
@@ -92,7 +92,7 @@ import qualified System.Directory as IO
          ( doesFileExist, doesDirectoryExist, canonicalizePath
          , getCurrentDirectory )
 import System.FilePath
-         ( (</>), (<.>), normalise, dropTrailingPathSeparator )
+         ( (</>), (<.>), normalise, dropTrailingPathSeparator , equalFilePath )
```
Mikolaj, line 95: A minor infraction here and elsewhere: space before comma.

frabrooks:
@Mikolaj  Made all the suggested changes. Also fixed a bug I introduced that was caught by two of the CI tests 😅

Mikolaj:
@frabrooks: looks good. Would it be possible to add a test that on Windows fails with the current code, but passes with yours?

frabrooks:
> @frabrooks: looks good. Would it be possible to add a test that on Windows fails with the current code, but passes with yours?

Ask, and ye shall receive. I've had a go. Reverted back to old code and test failed, with fix it passes.

The test is for the project root folder being a symlink which is how to reproduce the error outside of emacs. (I was also able to reproduce the error similarly to @Eason0210 )

Still have no idea, re my original issue, why, under some condition the call to `getCurrentDirectory` *within* cabal returns `F:\drive` rather than `f:\drive` when cabal has spun off from emacs. I don't know enough about the inner works of an OS to know for sure but my best guess is that maybe, being the parent process, emacs gives the file path with a capital letter and the OS is returning that to the IO call verbatim? Who knows...

Mikolaj:
@hasufell: given how crucial your input was, any last comments before we merge it?

hasufell:
The commits have a lot of reverse merge commits. I don't believe we want that, but a clean squash and rebase against master?

Additionally, I believe we want 2 commits in the following order:

1. Add the new test (it should fail)
2. Add the fix (the test should now succeed)

frabrooks:
> The commits have a lot of reverse merge commits. I don't believe we want that, but a clean squash and rebase against master?
> 
> Additionally, I believe we want 2 commits in the following order:
> 
> 1. Add the new test (it should fail)
> 2. Add the fix (the test should now succeed)

Done as requested. 

One last snag was that while moving stuff around I learned that under certain conditions/configurations of git on Windows the symlink for the test won't survive the pull; I guess because one needs admin to create a symlink on Windows. 

This wasn't caught by the CI for whatever reason as the tests here passed, but if I did a fresh clone of this pull request and then ran tests, the new test failed as the symlink wasn't there, not because the fix didn't work or anything.

So I opted to add a check to see if the symlink is there before running the test as I imagine so long as the test is running on *some* platforms that's good enough for all intents and purposes 👍

Mikolaj (approved):
Well done.

Mikolaj:
I cargo-culted the `rebase+merge me` label, but it doesn't seem to work. ;)

Mikolaj:
@mergifyio rebase

Mikolaj:
@Mergifyio rebase

Mikolaj:
@Mergifyio refresh

Mikolaj:
@Mergifyio refresh

Mikolaj:
@frabrooks: congratulations and thank you!

frabrooks:
> @frabrooks: congratulations and thank you!

Thank you for your help. First-ever contribution, but hopefully not the last 😄

swapneils:
To verify, is this fix included in Cabal 3.6.2.0? 
I'm running into the same issue in Emacs on Windows; Powershell/CMD can use `cabal build` and `cabal repl` in the package's home directory, but eshell/haskell-run-process can't detect the package.

Eason0210:
@swapneils   I tested cabal 3.6.2.0 on Windows 10， the issue still existed.

PR #7317, opened 2021-03-06
Suggest command on spelling mistakes

fgaz:
Patch from Francesco Ariis

---

### DIFF
--- a/Cabal/src/Distribution/Simple/Command.hs
+++ b/Cabal/src/Distribution/Simple/Command.hs
@@ -71,6 +71,8 @@ module Distribution.Simple.Command (
 import Prelude ()
 import Distribution.Compat.Prelude hiding (get)
 
+import qualified Data.Array as Array
+import qualified Data.List as List
 import qualified Distribution.GetOpt as GetOpt
 import Distribution.ReadE
 import Distribution.Simple.Utils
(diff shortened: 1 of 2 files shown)


Cabal/src/Distribution/Simple/Command.hs
```diff
@@ -513,6 +529,27 @@ commandsRun globalCommand commands args =
 
      where globalHelp = commandHelp globalCommand
 
+-- Levenshtein distance, from https://wiki.haskell.org/Edit_distance
+-- (Author: JeanPhilippeBernardy, Simple Permissive Licence)
+editDistance :: Eq a => [a] -> [a] -> Int
```
phadej, line 534: This is a bit sad as `cabal-install` already depends on `edit-distance` package. (But adding dependency to `Cabal` is hard).

fgaz: Yeah :-/

I can't think of a better solution though. cabal-install's option parsing is heavily tied to Cabal's

phadej: Yes. It's not a great design, but that what we have now.

emilypi (commented):
I agree with @phadej here. Not ideal, but definitely a quality of life improvemnt 👍

emilypi (approved):

Issue #7315, opened 2021-03-05
GHC 9.0 CI jobs


fgaz:
I'm a bit confused about how this is done. I guess by modifying ~~the files in `templates`~~ `cabal-dev-scripts/src/GenValidate.hs` and running `make github-actions`... ~~but those files say that they are generated too. Maybe it's like in `cabal-install.cabal.zinza`: the warning is there, but only because it has to go in the generated files somehow.~~

We probably want to document all this

/cc @emilypi

Mikolaj:
While we are at it, we can also remove the hack at

https://github.com/haskell/cabal/blob/3c4023f52cd582199c220eae7a0fd55d50ea7a73/cabal.project#L23-L24

to avoid publishing something depending on allow-newer.

l will bump base bounds of hackage-security as needed.

andreabedini:
@fgaz @Mikolaj is this still relevant?

Mikolaj:
Good find. Not having any `allow-newer` in CI is relevant, but that's it. Closing.

PR #7318, opened 2021-03-06
Add ghc 9.0 to ci

fgaz:
Let's see if this works...

#7315


fgaz:
I guess it has to be in master...

.github/workflows/linux.yml
```diff
@@ -17,6 +17,54 @@ on:
       - created
 
 jobs:
+  validate-9_0_1:
+    name: validate.sh ghc-9.0.1
+    runs-on: ubuntu-18.04
+    needs: validate-8_8_3
```
phadej, line 23: 9.0.1 job will start after 8.8.3 passes.

phadej:
At least in `master`:

```
% make cabal-install-dev
% cabal build -w ghc-9.0.1 --dry all
Resolving dependencies...
[__0] trying: Cabal-3.5.0.0 (user goal)
[__1] trying: base-4.15.0.0/installed-4.15.0.0 (dependency of Cabal)
[__2] trying: Cabal-described-3.5.0.0 (user goal)
[__3] next goal: rere (dependency of Cabal-described)
[__3] rejecting: rere-0.1 (conflict: base==4.15.0.0/installed-4.15.0.0, rere
=> base>=4.3.0.0 && <4.15)
[__3] fail (backjumping, conflict set: Cabal-described, base, rere)
goals I've had most trouble fulfilling: base, Cabal, Cabal-described, rere
Try running with --minimize-conflict-set to improve the error message.
```

So I doubt the job will succeed.

.github/workflows/linux.yml
fgaz: Aha, thanks. I assumed it would show as "waiting for dependency" or something

phadej: The *Waiting for status to be reported* jobs are there because they are `required` to pass before non-admins can merge. I.e. they come from repository settings, not in-repo CI config.

fgaz:
it did fail, but even before than the solving phase: there's no 9.0 container

phadej:
Hmm. I build the containers, but looks like I forgot to push them. They'll be up in an hour or so (many bytes to upload).

fgaz:
Blocked on https://github.com/haskell-hvr/cryptohash-sha256/issues/12 (or I could leave there the allow-newer)

phadej:
I pushed the containers for 9.0

fgaz:
Blocked on https://hub.darcs.net/shelarcy/regex-compat-tdfa/issue/2 too

phadej:
@fgaz, I'd drop `regex-compat-tdfa` from `cabal-testsuite` in favour of `regex-posix` or `regex-tdfa` (both of which work on windows nowadays).

fgaz:
...I just copied the only function we use and fixed it.

@phadej what would you replace regexes with?

phadej:
Good question. I'm not sure what `regex-compat-tdfa` does, looks like that not much, so I'd just use `regex-tdfa` https://hackage.haskell.org/package/regex-compat-tdfa-0.95.1.4/docs/src/Text-Regex.html

emilypi (approved):

PR #7324, opened 2021-03-10
Add -XUnliftedDatatypes

sgraf812:
This part of implementing the unlifted data types proposal,
https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2218.

### DIFF
--- a/Cabal/src/Language/Haskell/Extension.hs
+++ b/Cabal/src/Language/Haskell/Extension.hs
@@ -849,6 +849,9 @@ data KnownExtension =
   -- | Enable linear types.
   | LinearTypes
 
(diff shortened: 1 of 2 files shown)


sgraf812:
Am I supposed to act on the testsuite failures? If so, how?

fgaz:
Thanks!

We usually wait for a ghc pre-release / cabal branch-off, then add the new extensions all at once though ( https://github.com/haskell/cabal/blob/master/release-checklist.md ) so we'll do this later together with other extensions (and we'll update those hashes that are making ci fail)

sgraf812:
Ah, sorry. I should have read the Note in GHC.Driver.Session more carefully.

PR #7327, opened 2021-03-12
Bump Win32 bound

bgamari:
To accomodate 2.12.0.0. No changes necessary


fgaz (approved):
all items in the changelog are reflected in the types and I assume you built it with the new version, so +1

fgaz:
Wait, did you actually? I see for example some `c_GetShortPathName` references in `Cabal`

Issue #7312, opened 2021-03-03
Release planning for GHC 9.2 cycle (cabal 3.6)


bgamari:
It is typical for Cabal major releases to coincide with GHC major releases. This ticket is intended to facilitate coordination on the scoping and eventually cutting of a `Cabal-3.6` branch.

What major patches should be merged before this happens?

emilypi:
Just looking at the changelog, it looks like we'll need to support new extensions and flags/warnings:

- Support DuplicateRecordFields with PatternSynonyms (!4467, @adamgundry)

- -Wambiguous-fields (!4770, @adamgundry)

- NoFieldSelectors extension (!4743, @adamgundry)

- RecordDotSyntax (!4532, @shayne-fletcher-da)

- GHC2021 (!4853, @nomeata)

- -finfo-table-map and -fdistinct-constructor-tables (!3469, @mpickering)

- UnliftedDataTypes (!2218, @sgraf812)

PR #7349, opened 2021-04-03
Add new language extensions for 9.2

emilypi:
---

See: #7312

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC/ImplInfo.hs
+++ b/Cabal/src/Distribution/Simple/GHC/ImplInfo.hs
@@ -35,6 +35,7 @@ import Distribution.Version
 
 data GhcImplInfo = GhcImplInfo
   { supportsHaskell2010  :: Bool -- ^ -XHaskell2010 and -XHaskell98 flags
+  , supportsGHC2021      :: Bool -- ^ -XGHC2021 flag
   , reportsNoExt         :: Bool -- ^ --supported-languages gives Ext and NoExt
   , alwaysNondecIndent   :: Bool -- ^ NondecreasingIndentation is always on
   , flagGhciScript       :: Bool -- ^ -ghci-script flag supported
@@ -61,6 +62,7 @@ getImplInfo comp =
 ghcVersionImplInfo :: Version -> GhcImplInfo
 ghcVersionImplInfo ver = GhcImplInfo
   { supportsHaskell2010  = v >= [7]
+  , supportsGHC2021      = v >= [9,2]
   , reportsNoExt         = v >= [7]
   , alwaysNondecIndent   = v <  [7,1]
   , flagGhciScript       = v >= [7,2]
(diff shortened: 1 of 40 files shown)


emilypi:
@phadej does anything else need to happen to add support for new extensions?

fgaz (approved):

Cabal/src/Language/Haskell/Extension.hs
```diff
@@ -849,6 +849,18 @@ data KnownExtension =
...
   | LinearTypes
 
+  -- | Disable the generation of selector functions corresponding to record fields.
+  | NoFieldSelectors
+
+  -- | Enable the use of record dot-accessor and updater syntax
+  | RecordDotSyntax
+
+  -- | Enable the use of the GHC2021 collection of language extensions.
+  | GHC2021
```
phadej, line 859: This is not `Extension`. This is `Language`.

phadej (commented):
Yes, looks fine. 

Last time `cabal.vim` didn't exist, IIRC https://github.com/haskell/cabal/commit/753e006266a0ba5e83d6ccf7be1da14f6be598b1

I'd experiment with what `cabal-install-3.4` does when you give it an

```
library
  default-language: GHC2021
```

does it just pass it through (maybe with a warning that this is unknown language?)

That is how extension list works, IIRC, but maybe cabal is stricter with languages. I don't actually know.

Cabal/src/Language/Haskell/Extension.hs
emilypi: Shoot my bad - I was thinking extension for some reasno.

editors/vim/syntax/cabal.vim
```diff
@@ -186,6 +186,7 @@ syn keyword cabalExtension contained
   \ FunctionalDependencies
   \ GADTSyntax
   \ GADTs
+  \ GHC2021
```
phadej, line 189: Still not an extension :)

emilypi:
If this all looks good now, i'll merge when i get the 👍

Cabal/src/Distribution/Simple/GHC/ImplInfo.hs
```diff
@@ -61,6 +62,7 @@ getImplInfo comp =
 ghcVersionImplInfo :: Version -> GhcImplInfo
 ghcVersionImplInfo ver = GhcImplInfo
   { supportsHaskell2010  = v >= [7]
+  , supportsGHC2021      = v >= [9,2]
```
phadej, line 65: `9.1` if you want this to work with pre-releases.

bgamari:
This passes my local GHC validation. I have kicked off a full CI run [here](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5506).

emilypi:
@hvr any thoughts if you pop your head out?

emilypi:
Confirmed via @bgamari

fgaz:
It's too late to say this, but I think this branch needed a clean-up before merging. For example there are two commits that do the same thing (moving GHC2021 from the extensions to the languages) and the whole pr could have been one single commit since it can't be really split into multiple meaningful changes.

emilypi:
I squashed these on merge @fgaz - there is only a single commit in master.  https://github.com/haskell/cabal/commit/bd20f1b2f09f3dfe735afd66c4779491479670ce

fgaz:
Duh! Sorry for the noise then 😅

PR #7344, opened 2021-03-30
Cabal Init Omnibus

emilypi:
---
Please include the following checklist in your PR:

* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [x] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog).
* [x] The documentation has been updated, if necessary.

## New additions

This PR does the following: 

- Completely rewrites the `cabal init` command to have a saner structure that can be tested easily. 
- Optimizing for individual component target generation allows us to golden test individual stanzas under various scenarios instead of just whole cabal files
- Coverage reports may now be generated for every component of the command. Coverage works for the whole project now, full stop. 
- Control flow for interactive vs. non-interactive vs. simple target generation is completely decoupled.
- Cleans up the file creation spaghetti, and prunes unused or redundant behavior wherever it arises. 
- Rewrites the prompt code to be less unnecessarily typed (strings are fine!)

This comes with bugfixes: 
 
- [x] Fixes #7273 
- [x] Fixes #7253 
- [x] Fixes #7256 
- [x] Fixes #7255  
- [x] Fixes #7251
- [x] Fixes #6758 
- [x] Fixes #6864 
- [x] Fixes #1074

Issue #7224, opened 2021-01-05
Expose cabal.project.freeze parsing in the Cabal library


Fuuzetsu:
Currently it seems impossible to get at contents of `cabal.project.freeze` file without re-implementing a bunch of the `Distribution.Client.Config` that's only in the `cabal-install` binary and not available to be used.

Our specific use-case is that we generate `stack.yaml` file as a shim from our cabal files and the freeze file contains the set of flags that we should be using.

In general, it seems that parsing this file should be inside Cabal.

phadej:
Asking to add `cabal.project.freeze` parsing to `Cabal` library is the same as adding parsing support for `stack.yaml` there.

In the current state of configuration code, splitting out the parser of cabal-install files (`cabal.config`, `cabal.project`), and having even some degree of compatibility is unfortunately impossible.

Fuuzetsu:
>Asking to add cabal.project.freeze parsing to Cabal library is the same as adding parsing support for stack.yaml there.

I'm not sure how this would be. We already do the parsing in `cabal-install` using the `Cabal` package, it seems like some of the code should just migrate back into `Cabal`... 

>In the current state of configuration code, splitting out the parser of cabal-install files (cabal.config, cabal.project), and having even some degree of compatibility is unfortunately impossible.

We already parse `.cabal` files, isn't it exactly the same issue? Is `Cabal` only meant to parse the `.cabal` file and offer tools to parse other things in the cabal format, but nothing specific to cabal-install? Hence the `stack` comment. If yes, feel free to close. It is very unfortunate though. Maybe `cabal-install` should expose modules so we could just invoke an already written piece of code from there...

----

In the meantime, if anyone is interested, I am parsing out the bits I'm interested in (package flags in constraints) with following code:

```haskell
parseCabalFlags :: Action CabalFlags
parseCabalFlags = do
    need [cabalProjectFreezeFile]
    fields <- do
        freezeFileBs <- liftIO $ BS.readFile cabalProjectFreezeFile
        case Cabal.readFields freezeFileBs of
            Left e -> fail $ show e
            Right fields -> pure fields
    constraints <- case listToMaybe $ mapMaybe getConstraints fields of
        Nothing -> fail "No constraints section in cabal freeze file."
        Just constraints -> pure constraints

    let cabalFlags =
            let f m t = case Cabal.simpleParse (T.unpack t) of
                    Nothing -> m
                    Just (CabalPkgFlags pkg flags) ->
[...]

Issue #6090, opened 2019-06-19
Expose `cabal.project` in library component?


mitchellwrosen:
Would it be possible to expose `cabal.project` data structures and parsing code in a public `cabal-install` library component? 

This would be helpful for `cabal-install` wrapper tools. For example, I may want to write a tool that automagically runs one `ghcid` session for each component in a package (since `ghci` can only load one component at a time). Doable for single-package projects using the `Cabal` library, but more difficult for entire projects with multiple `.cabal` files in different subdirectories.

phadej:
Note: since `cabal-install-2.4` there are no such thing as `cabal-install` lib component.

I'm not saying that there couldn't be some small specific independent library, but that library ought to have well-designed API (i.e. not everything `cabal-install` uses internally).

23Skidoo:
This has been discussed previously, see #5476. I think that perfect is the enemy of good enough here, so personally am +1 on following the GHC API approach and exposing the existing internals with no guarantees of stability between major releases. Though I'd like to do this in gradual fashion and expose the solver lib first.

phadej:
The #5476 already mentioned the fact which is worth repeating:

If we don't care about minor releases breaking stuff, than having a public library is ok. But in that case `cabal-install` would be a PITA library to depend upon.

An alternative is to be very strict of which patches to get in.

E.g. http://hdiff.luite.com/cgit/cabal-install/diff?id=2.4.1.0&id2=2.4.0.0 has the breaking API change right at the top of the diff.

23Skidoo:
Right. I think that we can live with the latter if major `cabal-install` releases are frequent (every six months or so).

hvr:
> If we don't care about minor releases breaking stuff

If for one strongly care about this; and this would effectively mean that we won't be able to fix some things within a patch-level or minor-release, except with high costs, due to the hard requirement to not break the API contract; and fixes might be delayed by ~6 months because of this. Are we sure we want to incur that cost?

phadej:
@hvr, I don't think having a major release of `cabal-install` without corresponding `Cabal` major is a biggie.

E.g. 

- `cabal-install-3.0` uses `Cabal-3.0`
- `cabal-install-3.2` has a "breaking" fix, still uses `Cabal-3.0`
- `cabal-install-3.4 uses `Cabal-3.4`

So `Cabal-3.2` won't ever exist.

I actually don't care whether `cabal-install` and `Cabal` versions are in sync.

[...]

PR #7358, opened 2021-04-14
Expose cabal-install lib, remove dogfooding, expose cabal-install-solver as own lib, enable HPC coverage reporting. 

emilypi:
As discussed in https://github.com/haskell/cabal/pull/7344#issuecomment-819670133, this PR does the following: 

- Exposes `cabal-install` as a library, bumping the `cabal-install` cabal version to 2.2, and removing the dogfooding framework surrounding `cabal-install` to facilitate easier testing. 
- Moves `cabal-install-solver` out to its own directory, and treats it as its own separate project. Addresses #6090, #7224, and #3781
- Enable HPC coverage: absorb `cabal-install-solver-dsl` module into the test modules, and incorporate it into its own test target. Splits out unit tests from long-running tests. Also adds proper subcomponent coverage masking so we can finally generate baseline coverage reports for `cabal-install`. 

TODO:
 
- [x] Doc changes for testing.md, contributing.md
- [x] Changelog entry 
- [x] Changes to CI 
 
Fixes: 

- #7224 
- #6090
- #4798 
- #3781
- #1597

Baseline report: 

[cabal-install-baseline-coverage.zip](https://github.com/haskell/cabal/files/6313866/cabal-install-baseline-coverage.zip)
 

CC @fgaz @phadej @ptkato


cabal-install/cabal-install.cabal
```diff
@@ -12,68 +9,46 @@ Description:
...
-      if impl(ghc < 8.8)
-        ghc-options: -Wnoncanonical-monadfail-instances
 
+library
+    default-language: Haskell2010
+    ghc-options: -Wall -Wcompat -Wnoncanonical-monad-instances
+    if impl(ghc < 8.8)
+      ghc-options: -Wnoncanonical-monadfail-instances
       if impl(ghc >=8.10)
         ghc-options: -Wunused-packages
```
fgaz, line 48: What about using common stanzas now that we can? So that we only need to update this stuff in one place

cabal-install-solver/cabal-install-solver.cabal
```diff
@@ -39,18 +39,18 @@ library
   default-language: Haskell2010
   hs-source-dirs:   src
   hs-source-dirs:   src-assertion
-  ghc-options:      -Wall -fwarn-tabs -fwarn-incomplete-uni-patterns
-
-  if impl(ghc >=8.0)
```
fgaz, line 44: I think this and the other ghc compat stuff in this commit should be moved to the previous one, where the minimum ghc actually changed (or to a new commit)

fgaz: Oh, I see that file was `cabal-version: 2.2` even before the change in cabal-install. So I guess a new commit?

cabal-install/cabal-install.cabal
emilypi: Oo! good point. In the previous iteration i had `cabal-version: 2.0`, but `2.2` seemed like a better idea for internal libs. We definitely can do common stanzas.

cabal-install-solver/cabal-install-solver.cabal
emilypi: Yes, it was changed upon exposing the new library to match the `cabal-install.dev` cabal file.

cabal-install/cabal-install.cabal
emilypi: Done

fgaz (commented):
The workflow files are generated (see the `github-actions` make target)

edit: this was supposed to be a comment on commit 59b2fe57188b6de6eb62ee993be083d40db4a7ec

fgaz:
Could you squash the commits that do the same thing (eg. Setup.hs for cabal-install-solver → separate cabal-install-solver)?

emilypi:
Sure i'll restructure these and rebase tomorrow 👍

cabal-install-solver/cabal-install-solver.cabal
```diff
@@ -39,18 +39,18 @@ library
...
-    if impl(ghc <8.8)
-      ghc-options: -Wnoncanonical-monadfail-instances
+  ghc-options:
+    -Wall -Wcompat -Wnoncanonical-monad-instances
+    -fwarn-tabs -fwarn-incomplete-uni-patterns
 
+  if impl(ghc <8.8)
+    ghc-options: -Wnoncanonical-monadfail-instances
     if impl(ghc >=8.10)
       ghc-options: -Wunused-packages
```
fgaz, line 49: The inner condition is always false. I think it's one level too deep

cabal-install/cabal-install.cabal
```diff
@@ -12,68 +9,51 @@ Description:
...
-                     -Wnoncanonical-monad-instances
-      if impl(ghc < 8.8)
-        ghc-options: -Wnoncanonical-monadfail-instances
-
+common warnings
+    ghc-options: -Wall -Wcompat -Wnoncanonical-monad-instances
+    if impl(ghc < 8.8)
+      ghc-options: -Wnoncanonical-monadfail-instances
       if impl(ghc >=8.10)
         ghc-options: -Wunused-packages
```
fgaz, line 46: Same here

cabal-install/cabal-install.cabal
```diff
@@ -205,78 +185,21 @@ executable cabal
...
-
     build-depends:
         async      >= 2.0      && < 2.3,
         array      >= 0.4      && < 0.6,
-        base       >= 4.8      && < 4.15,
         base16-bytestring >= 0.1.1 && < 1.1.0.0,
         binary     >= 0.7.3    && < 0.9,
         bytestring >= 0.10.6.0 && < 0.12,
         Cabal      == 3.5.*,
+        cabal-install-solver,
```
fgaz, line 195: I think this needs a constraint like the Cabal one above (assuming we release cabal-install-solver in lockstep with Cabal and cabal-install)

cabal-install/src/Distribution/Client/Utils.hs
```diff
@@ -362,8 +360,4 @@ progressMessage verbosity phase subject = do
         ProgressCompleted   -> "Completed    "
 
 cabalInstallVersion :: Version
-#ifdef __DOCTEST__
-cabalInstallVersion = mkVersion [3,3]
-#else
-cabalInstallVersion = mkVersion' Paths_cabal_install.version
-#endif
+cabalInstallVersion = mkVersion [3,5]
```
fgaz, line 363: Hmm we probably need to add this to the release checklist (can be in another pr)

cabal-install/cabal-install.cabal
```diff
@@ -205,78 +185,21 @@ executable cabal
...
         containers >= 0.5.6.2  && < 0.7,
         cryptohash-sha256 >= 0.11 && < 0.12,
         deepseq    >= 1.4.1.1  && < 1.5,
         directory  >= 1.2.2.0  && < 1.4,
         echo       >= 0.1.3    && < 0.2,
         edit-distance >= 0.2.2 && < 0.3,
+        exceptions,
```
fgaz, line 202: where is this needed? I don't see any change that uses it. It also needs bounds

cabal-install/cabal-install.cabal
```diff
@@ -317,13 +236,171 @@ executable cabal
...
+          network-uri >= 2.6.2.0 && <2.6.4.0,
+          random,
+          tar,
+          time,
+          zlib,
+          tasty >= 1.2.3 && <1.4,
+          tasty-golden >=2.3.1.1 && <2.4,
+          tasty-quickcheck,
+          tasty-hunit >= 0.10,
+          tree-diff,
```
fgaz, line 315: tasty-quickcheck and tree-diff may also need a bound

cabal.project
```diff
@@ -1,5 +1,5 @@
 packages: Cabal/ cabal-testsuite/
-packages: cabal-install/cabal-install-solver/
+packages: cabal-install-solver/
```
fgaz, line 2: This should be in the previous commit

cabal.project
```diff
@@ -29,18 +29,56 @@ allow-newer: windns-0.1.0.1:base
...
+--
+-- NOTE: for library coverage in multi-project builds,
+-- see:
+--
+-- * https://github.com/haskell/cabal/issues/6440
+-- * https://github.com/haskell/cabal/issues/5213#issuecomment-586517129
+--
+-- We must mask coverage for dependencies of `cabal-install` in
+-- multiproject settings in order to generate coverage for
+-- the `cabal-install` library
```
fgaz, line 49: Question: how do we get the coverage for Cabal?

templates/ci-artifacts.template.yml
```diff
@@ -27,8 +27,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Package project
         run: |
-          cabal v2-sdist --builddir=sdist Cabal cabal-install
-          python3.8 release.py --tarlib=sdist/sdist/Cabal-*.tar.gz --tarexe=sdist/sdist/cabal-install-*.tar.gz
+          cabal v2-sdist --builddir=sdist Cabal cabal-install-solver cabal-install
+          python3.8 release.py --tarlib=sdist/sdist/Cabal-*.tar.gz --tarsolver=sdist/sdist/cabal-install-solver-3.*.tar.gz --tarexe=sdist/sdist/cabal-install-3.*.tar.gz
```
fgaz, line 31: why the "3."s?

CONTRIBUTING.md
```diff
@@ -26,54 +26,23 @@ cabal v2-build Cabal:unit-tests # build Cabal's unit test suite
 cabal v2-build cabal-tests # etc...
 ~~~~
 
-**Dogfooding HEAD.**
```
fgaz, line 29: This is still valid though, isn't it?

CONTRIBUTING.md
```diff
@@ -137,22 +105,16 @@ There are also other test suites:
...
+  cabal-install's dependency solver, VCS, and file monitoring code.
+  If you are working on the solver you should run this test suite.
 
 * `cabal-install:integration-tests2` are integration tests on some
   top-level API functions inside the `cabal-install` source code.
 
 For these test executables, `-p` which applies a regex filter to the test
 names.
 
-**Testing `cabal-install` Locally**
```
fgaz, line 150: I think this section should stay and just say "run cabal test". After all running tests locally is still encouraged!

changelog.d/pr-7358
```diff
@@ -0,0 +1,21 @@
+synopsis: Enable HPC coverage for `cabal-install`
```
fgaz, line 1: I think the main point _for users_ is the cabal-install-solver/cabal-install-library stuff more than hpc

fgaz (changes requested):
Small stuff and questions

cabal-install/cabal-install.cabal
fgaz: Nvm, this isn't new

cabal.project
emilypi: Through pain and suffering. We'd have to rework the `cabal-testsuite`, and associated tests to be coverable. I'm punting on that one until i understand it better.

templates/ci-artifacts.template.yml
emilypi: because now that `cabal-install-solver` exists, `cabal-install-*` matches two possible conditions. This allows us to match  on file names more precisely.

cabal-install/cabal-install.cabal
emilypi: Good idea.

cabal-install/cabal-install.cabal
emilypi: Artifact from the Cabal Init work. Nice catch - i'll remove.

CONTRIBUTING.md
emilypi: I think it's confusing to keep, since it implies having to think about something special. It is now the _only_ way we test `cabal-install`, so i don't think it makes sense to point out specifically.

CONTRIBUTING.md
emilypi: Same as above - calling `cabal test` is normal behavior, so i don't think it needs to be highlighted. We can debate this - I think it's unnecessary to point out details that people should _expect_ to already be a thing.

typedrat:
This would be massively useful to me.

cabal-install-solver/cabal-install-solver.cabal
emilypi: Fixed

cabal-install/src/Distribution/Client/Utils.hs
emilypi: Added

cabal.project
emilypi: :+1:

emilypi:
@fgaz okay - comments should be addressed, and I'll merge on green, following up with a PR with updates to the release checklist. The release checklist in particular needs to detail how to update CI, which i'll add as well. I'll do one more pass to see if anything stands out, and then I think we're good to merge.

fgaz (approved):

emilypi:
Was this closed by linking? I didn't physically close this issue.

PR #7252, opened 2021-01-19
Include `cmm-sources` when linking shared objects

mpickering:
Fixes #7182 

Patch is by @bgamari, I have just rebased it on master as requested by @phadej in #7183

---
Please include the following checklist in your PR:

* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [x] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
* [x] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

PR #7359, opened 2021-04-15
Follow-up to #7252

bgamari:
- Add a changelog entry
 - Carry out a small renaming suggested by @phadej to make the implementation a bit clearer

---

*  [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -557,7 +557,7 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
                       , toNubListR (cmmSources libBi)
                       , toNubListR (asmSources libBi)
                       ]
-      cObjs       = map (`replaceExtension` objExtension) cLikeSources
+      cLikeObjs   = map (`replaceExtension` objExtension) cLikeSources
       baseOpts    = componentGhcOptions verbosity lbi libBi clbi libTargetDir
       vanillaOpts = baseOpts `mappend` mempty {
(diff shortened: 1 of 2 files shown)


Cabal/src/Distribution/Simple/GHC.hs
```diff
@@ -551,13 +551,13 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
...
   -- modules?
-  let cLikeFiles  = fromNubListR $ mconcat
+  let cLikeSources  = fromNubListR $ mconcat
                       [ toNubListR (cSources   libBi)
                       , toNubListR (cxxSources libBi)
                       , toNubListR (cmmSources libBi)
                       , toNubListR (asmSources libBi)
                       ]
-      cObjs       = map (`replaceExtension` objExtension) cLikeFiles
+      cObjs       = map (`replaceExtension` objExtension) cLikeSources
```
phadej, line 560: this is still `cObjs`, which is confusing as it's `cLikeObjs`. Ditto `cProfObjc` etc (which include objects **not only** from C-code, but from **C-like** code, where likeness is vague notion as well).

TL;DR this is not what I suggested.

emilypi: I tend to agree here. Because we don't document, we have to improve the state of variable names. Though, preferably, I'd like both a docstring and better names.

emilypi:
@bgamari CI is fixed if you rebase to master

emilypi (approved):

Issue #7370, opened 2021-04-25
cabal.project compiler field seems to be ineffective


yobson:
**Describe the bug**
I am trying to build a project with a few packages compiled by GHCJS (v8.6) and a set with GHC (8.10)
and some libraries used by both.

I set the `compiler` field to `ghc` and `ghcjs` respectively in my `cabal.project` file in individual package sections and yet it uses `ghc` for all of them!

**To Reproduce**
[cabal-bug-report.zip](https://github.com/haskell/cabal/files/6372273/cabal-bug-report.zip)

I have attached a zip of a minimal example. `project1` is set to compile with `ghc` and `project2` with `ghcjs`
Simply run

```
$ cabal v2-build all
```
or
```
$ cabal v2-build project1
$ cabal v2-build project2
```

**Expected behavior**
To build `project1` with `ghc` and to build `project2` with `ghcjs`

**System information**
 - Operating system: Mac os Big Sur
 - `cabal 3.4`, `ghc 8.10.4`, `ghcjs 8.6.1`

fgaz:
It looks like the field does work when set at the top level. It probably was never intended to be used in individual packages, like with the similar `with-compiler` option:

> At the moment, it’s not possible to set with-compiler on a per-package basis, but eventually we plan on relaxing this restriction. If this is something you need, give us a shout.

So I'd say this is not a bug. Pull requests to fix the documentation and/or make compiler/with-compiler work for individual packages are welcome though!

yobson:
Thanks for the response!
I may take you up on that pull request! Docs probably. Cabal is massive and I feel my haskell abilities are still young!

My workaround was just adding
```
if impl(ghcjs)
    buildable: False/True
```
tastefully to various cabal files and then running
```
$ cabal build all
$ cabal build all --ghcjs
```
Is your two separate project files solution more elegant?

fgaz:
> Is your two separate project files solution more elegant?

Nah, both are fine :)

PR #7371, opened 2021-04-25
Documented limitation with compiler field in project files 

yobson:
Added a declaimer (copied and changed) from the `with-compiler` section to the `compiler` section of the cabal project description part of the docs to assist with #7370 

### DIFF
--- a/changelog.d/issue-7370
+++ b/changelog.d/issue-7370
@@ -1,0 +1,3 @@
+synopsis: Documented limitation to compiler section of cabal project files
+issues:
+  #7370
--- a/changelog.d/issue-7370
+++ b/changelog.d/issue-7370
@@ -1,3 +1,0 @@
-synopsis: Documented limitation to compiler section of cabal project files
-issues:
-  #7370
--- a/doc/cabal-project.rst
+++ b/doc/cabal-project.rst
@@ -703,9 +703,8 @@ feature was added.
 
     The command line variant of this flag is ``--compiler=ghc``.
 
-    At the moment, it's not possible to set :cfg-field:`compiler` on a
-    per-package basis, but eventually we plan on relaxing this
-    restriction. If this is something you need, give us a shout.
+    It's not possible to set :cfg-field:`compiler` on a
+    per-package basis. 
 
 .. cfg-field:: tests: boolean
                --enable-tests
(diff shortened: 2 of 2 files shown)


doc/cabal-project.rst
```diff
@@ -703,6 +703,10 @@ feature was added.
 
     The command line variant of this flag is ``--compiler=ghc``.
 
+    At the moment, it's not possible to set :cfg-field:`compiler` on a
+    per-package basis, but eventually we plan on relaxing this
+    restriction. If this is something you need, give us a shout.
```
phadej, line 708: Is there a plan? Can we promise that? I'd rather remove this promise (also from `with-compiler` field), as I simply don't see how it would be possible (without larger `cabal.project` overhaul).

changelog.d/issue-7370
```diff
@@ -0,0 +1,3 @@
+synopsis: Documented limitation to compiler section of cabal project files
+issues:
+  #7370
```
fgaz, line 3: No need for a changelog entry in this case, it isn't an user facing change

yobson: Do I just re-push without the file?

doc/cabal-project.rst
yobson: I don't know if there is a plan. But it would be a nice thing to have given the multitude of web haskell compilers. I can remove the promise if you would like?

changelog.d/issue-7370
fgaz: Yes

doc/cabal-project.rst
phadej: That line was introduced 5 years ago in https://github.com/haskell/cabal/commit/99d48fa0fe930e019ff940d59c152069d1a2aaf8

In that time per-package compiler setting didn't become any closer to be true. I would say that the best which could happen is that we get `if (..)` pragmas to `cabal.project` so same `cabal.project` file could be reused for different compilers (versions or kinds).

But having the same `cabal build all` building with two different compilers doesn't seem likely.
Thus, jremove `, but we plan...` - we don't. (In general, there should be links to open issues if there are plans)

yobson:
I've made the above changes!

fgaz:
Thanks! Could you also remove the promise from with-compiler field while we're at it?

yobson:
Also done!

fgaz (approved):

PR #7381, opened 2021-05-04
Mergify: configuration update

fgaz:
This change has been made by @fgaz from https://mergify.io config editor.


fgaz:
Misclicked while checking out their interface. Too bad I can't ninja-delete this.

PR #7188, opened 2020-12-05
directory: ensure xdg compliance

GovanifY:
---
Please include the following checklist in your PR:

* [X] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ ] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
^ Not sure how I should add that to the changelog, any input appreciated! 
* [X] The documentation has been updated, if necessary.

This ensures compliance with the [XDG base specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
This will need to be merged alongside the upstream ghc PR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4567 to avoid any unintended breakage. 
This fixes the upstream bug https://gitlab.haskell.org/ghc/ghc/-/issues/6077

Issue #1857, opened 2014-05-12
Remove the need to hardcode absolute paths for some entries in `.cabal/config`: support xdg home variables


KholdStare:
Hi!

I keep all my configuration files on github to have an easily reproducible environment on all my machines. I keep my `.cabal/config` file in there as well, and don't like the fact that some absolute paths have to be hardcoded (e.g. `remote-repo-cache`, `world-file` and `build-summary`). This is inconvenient if `$HOME` happens to be different on various machines (e.g. work vs home).

Some possible solutions (which could be wrong):
- Expand `~` in paths.
- Expand environment variables from the shell (perhaps even very specific ones)
- Allow usage of `$prefix` in all paths

If one of the above suggestions is worthwhile, I'd be up for implementing it. Just let me know :smile:

Thank you!

chris-martin:
Would someone with a commit bit mind weighing in on whether a patch would be accepted for this?

hvr:
I generally like the idea, but under the condition we do *not* open up support for arbitrary environment variables.

I'd concretely suggest to add support for a couple additional `$`-variables

- `$home` via `System.Directory.getHomeDirectory`  (if we want, we could support `~/` as sugar
- `$xdg_data` / `$xdg_config` / `$xdg_cache` (via `System.Directory.getXdgDirectory`)

which I believe should address most use-cases without bringing into scope an unknown set of new variable names.

The downside of allowing env-variables to affect the paths in the cabal config is that file-change-monitoring gets slighlty more complicated, but by using an explicitly enumerated set of input variables (i.e. `$home` /`$xdg_{data,config,cache}`) this remains simple enough.

sboosali:
yeah, I agree w.r.t. a fixed of environment variables. and the `~` sugar.

one issue is which environment variables are chosen for Windows (which is far as I know, isn't part of the XDG BaseDirs spex). For example, the standard implementations between Haskell and Racket differ (see below). I'd stick with whatever `directory` does (or if `Cabal` has its own XDG utilities, those), but I wanted to mention it.

Haskell (`directory` package):

* `$XDG_CONFIG_HOME` — `%APPDATA%`
* `$XDG_DATA_HOME`   — `%APPDATA%`
* `$XDG_CACHE_HOME`  — `%LOCALAPPDATA%`

Racket (`basedir` package):

* `$XDG_CONFIG_HOME` — `%LOCALAPPDATA%`
[...]

PR #7386, opened 2021-05-09
Move towards using XDG directories.

athas:
Yet another swing at #1857.  In contrast to #7188, this draft PR doesn't just move the `.cabal` directory somewhere, but distinguishes between the cabal configuration file (which goes in `$XDG_CONFIG_HOME/cabal`) and all the other files produced by Cabal, which go in `$XDG_CACHE_HOME/cabal`.  Here's the basic design (not all of which is implemented):

*   The Cabal configuration file goes in `$XDG_CONFIG_HOME/cabal/config`.
*   `cabal install` binaries go in` $HOME/.local/bin` (from the systemd file-hierarchy spec).
*   Everything else goes in `$XDG_CACHE_HOME/cabal`).

This might break people, of course. So if CABAL_DIR exists, then the current behaviour will be maintained (everything in CABAL_DIR). Further, as a (hopefully temporary) migration path, if $HOME/.cabal exists, then that will interpreted as CABAL_DIR being set to $HOME/.cabal.

I also suggest maintaining the current behaviour on Windows, since I have no idea what TRT would be there (but I'm pretty sure it's not XDG).

I have also found a few places where certain path computations seem to be duplicated instead of importing functions from `Distribution.Simple.InstallDirs`.  Should I make small preliminary PRs to fix these in advance?

---


athas:
> I have also found a few places where certain path computations seem to be duplicated instead of importing functions from Distribution.Simple.InstallDirs. Should I make small preliminary PRs to fix these in advance?

An example (where I have fixed it, I think): https://github.com/haskell/cabal/pull/7386/files#diff-ae0158c0ab0fa116a8e5deeea1f7a9809c9c061a1646444580df9b4c131263b0L66-L82

Cabal/src/Distribution/Simple/GHCJS.hs
```diff
@@ -1801,7 +1801,7 @@ pkgRoot verbosity lbi = pkgRoot'
       let ghcjsProg = fromMaybe (error "GHCJS.pkgRoot: no ghcjs program") $ lookupProgram ghcjsProgram (withPrograms lbi)
       in  fmap takeDirectory (getGlobalPackageDB verbosity ghcjsProg)
     pkgRoot' UserPackageDB = do
-      appDir <- getAppUserDataDirectory "ghcjs"
+      appDir <- getXdgDirectory XdgData "ghcjs"
```
phadej, line 1804: This is wrong. Here we look what `ghcjs` (or `ghc`) does. This is not up to Cabal to decide.

athas: Yeah, this was something I imported from #7188 and forgot to clean up (I did fix the similar problem in GHC.hs)...

Cabal-tests/tests/HackageTests.hs
```diff
@@ -63,23 +64,15 @@ import Data.TreeDiff.Pretty          (ansiWlEditExprCompact)
 parseIndex :: (Monoid a, NFData a) => (FilePath -> Bool)
            -> (FilePath -> B.ByteString -> IO a) -> IO a
 parseIndex predicate action = do
-    cabalDir   <- getAppUserDataDirectory "cabal"
-    configPath <- getCabalConfigPath cabalDir
+    configPath <- getConfigFilePath
     cfg        <- B.readFile configPath
     cfgFields  <- either (fail . show) pure $ Parsec.readFields cfg
+    repoCache  <- case lookupInConfig "remote-repo-cache" cfgFields of
+                    []        -> defaultCacheDir   -- Default
```
phadej, line 71: this is change is good, and could be merged without xdg stuff.

athas: Alright, I'll prepare a separate PR.

phadej:
The tricky part is where to put `/caba/store`. It's kind of a cache, so `XDG_CACHE_HOME` makes sense, OTOH installed executables (and libraries for environment files) are there as well, so if cache is purged things are broken (the cache is not autopopulated).

I guess `XDG_CACHE_HOME` is not bad, but this limitation is good to mention.

---

I think that `CABAL_DIR` should go away with this change. making its default equal to `XDG_CACHE_HOME` isn't future proof. What if something belongs to `XDG_DATA_HOME`, or [`XDG_STATE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) (*There is a single base directory relative to which user-specific state data should be written.* - is this a place to put nix-style store?)

---

I stress, if this change goes in, there should be a very clear "migration" documentation. (Which at simplest is: continue using explicit config, set by `CABAL_CONFIG` pointing to old directories you use, or remove everything and start from scratch).

phadej:
Please not comment here. This PR is draft, and discussion what it should do is in https://github.com/haskell/cabal/issues/680#issuecomment-835750887

Mikolaj:
I took the liberty of merging this branch with master, to kick off the CI.

athas:
CI wrinkle: the Debian `cabal` is used to generate the config file (with v2-update), which is then put in `$HOME/.cabal/config`, but the tests will expect it in `$HOME/.config/caba/config`.

gbaz:
Along with where cabal stores its own files, we should start using xdg for writing env files and other ghc interaction, cf for example: https://gitlab.haskell.org/ghc/ghc/-/issues/20684

Mikolaj:
Also, see the cabal issue about that: #7819

Mikolaj:
Hello! Can we help to push this forward?

Mikolaj:
I see a decision is needed. Could somebody remind what the decision is about? We could discuss that on this Thursday's meeting (all devs are invited).

athas:
I think the only real decision to be made is a set of instructions about where to put what, and whether `CABAL_DIR` should be used for anything except backwards compatibility.  I didn't stop working on it because it's particularly difficult; only because I got distracted by other things.  I recall only two meaningful technical challenges:

1. Backwards compatibility, which I think should be triggered if `CABAL_DIR` is set or `~/.cabal` exists.
2. Some parts of the test suite use the `CABAL_DIR` environment variable and will need to use the XDG config stuff instead, but they can't depend directly on the `cabal-install` package for reasons I don't remember clearly anymore.  This can probably be solved in ad-hoc ways, since it's just a test suite.

gbaz:
It sounds like you can just continue to push forward, including keeping back-compat for `CABAL_DIR` for now. If there's any specific question about what goes where, please highlight it -- otherwise, it looks like everything is on track, and we're excited to see this proceed.

Mikolaj:
> (in the documentation, we can note that we're discouraging future use of `CABAL-DIR`)

which implies that the answer to your question "whether CABAL_DIR should be used for anything except backwards compatibility" is that it should not. We plan to gradually deprecate CABAL_DIR and remove it, helping its users to find good alternate workflows along the way.

athas:
I am working on this again.  Currently without any `CABAL_DIR` compatibility; I think it is better to add that back in later (but before any PR is merged obviously).

I will completely remove the `getCabalDir` function, since there is no "cabal directory".

This PR is marked as blocked due to "decision-needed", but it's a bit hazy what decision that is.  I put a spec (sort of) in https://github.com/haskell/cabal/issues/680#issuecomment-868826313, and I don't recall any objections.

Mikolaj:
Regardless of what decision we had in mind, probably the most important one is how fast do we intend to migrate and how do we support users during this transition. It would be ideal if we kept full backward compatibility for a long time, while nudging users in the right direction and, if the user starts from scratch, creating the new cabal directory hierarchy(ies) according to XDG.

athas:
There is still lots to do (look at how many tests fail!), but one I'm not sure about how to handle is this:

https://github.com/athas/cabal/blob/d5a10e93e86067b32ca9ea19579f5f5a1632fbee/cabal-install/src/Distribution/Client/Config.hs#L541-L559

What is this "savedUserInstallDirs" business about?  I assume I should just change the default to be something else, but I have no idea what it is, and I'm having trouble connecting this bit of configuration to anything in the manual.  I don't think it's a feature I have ever used.

athas:
I don't think backwards compatibility will be difficult, although I'm not completely sure what cabal used to support.  The directories are currently accessed through functions like `defaultStoreDir`, `defaultCacheDir`, etc. and it seems pretty trivial to make these first look for a `CABAL_DIR` environment variable and prefer that if set.

I'm less sure about how to also support the various configuration knobs in the cabal config file.  For example, looking at my own config file I have found a "logs-dir" key, but during my work on this PR I have certainly seen code that seems to always use  `$CABAL_DIR/logs`.   Generally, there are several places in the code where something wants to write to a directory, but doesn't have the configuration object handy.  This is not going to be made worse by an XDG migration, though.

athas:
Tests are failing for (maybe among other reasons) bumping the `directory` lower bound to 1.3.7.0 in order to use XdgState.  I suppose we can work around this if `directory` cannot be bumped, but perhaps someone can explain why it fails?  Here is the GA error: https://github.com/haskell/cabal/runs/7906534227?check_suite_focus=true#step:3:1449

athas:
Is it because `directory` is a GHC boot library?

athas:
Oh, I understand, it's due to the bootstrapping stuff hardcoding package versions.  I think I can fix that.

Mikolaj:
> Oh, I understand, it's due to the bootstrapping stuff hardcoding package versions. I think I can fix that.

It's actually the easiest to regenerate the bootstrapping files (via Makefile) after bumping the bound. Otherwise, you need to copy hashes, etc.

Do we keep compatibilty with old GHCs, starting with GHC 8.2, after bumping the bound? E.g., isn't the lower bound of `base` in `directory` too restrictive?

athas:
The newest `directory` requires base 4.5, which is from GHC 7.4, so I think we're good on supporting old GHCs.

Even after updating the bootstrap JSONs it still seems unable to find the right directory version.  I must be missing something.

athas:
Progress!  Now it builds and some tests fail.  I think it is because some of the test harness depends on `CABAL_DIR` being respected, so I added support for that.  The policy is:

* If `$CABAL_DIR` is set, treat that as a classic monolithic directory (like current ~/.cabal) containing all subdirectories.
* If `~/.cabal` exists, treat that as if `CABAL_DIR=~/.cabal` is set (this is for backwards compatibility, and in time one might want to print a warning if ~/.cabal exists but `$CABAL_DIR` is unset).
* Otherwise it uses a variety of different XDG directories for different purposes.

I don't know anything about Windows, but I wonder if it might be better to keep the monolithic directory behaviour on Windows.  `directory` maps XDG paths to Windows paths, but I have no idea whether what it does is actually what Windows users like.

athas:
Another wrinkle (or design question): where should `script-builds` be stored?  Previously they were at `~/.cabal/script-builds`.  In the current branch they are at `~/.cabal/packages/script-builds`, but that's only because the `defaultCacheDir` function returns the "packages" directory, which doesn't make a great deal of sense to me (but maybe there's some part of the vision I'm not privy to).

Mikolaj:
Does XDG say anything about cache location?

athas:
Yes, `XDG_CACHE_HOME`.  It would be no problem to create a `script-builds` subdir there.

athas:
It looks like most things are turning green, so I am marking this as ready for review.  There are a few nits in the code left I want to fix ([like this one](https://github.com/haskell/cabal/pull/7386#issuecomment-1219911343)), but otherwise the big missing piece is documentation.  However, I don't want to write the documentation until I'm sure what behaviour is actually desired.  How would one go about creating consensus on that?

fgaz (commented):
Just from skimming the diff it looks good!

This will need a detailed changelog entry and something about the migration in the docs, as suggested here: https://github.com/haskell/cabal/pull/7386#issuecomment-835750489

fgaz:
Looks like more docs need updating:

```
$ git grep -E 'cabal/(config|logs|packages|store|bin)' | wc -l
98
```

athas:
Yes, all documentation remains to be written.  I suppose I will just start doing that.  I suggest the following narrative:

* `CABAL_DIR` (and `CABAL_CONFIG`) is actually still useful in many cases (e.g. Cabal's own test suite) and will be documented as a fully supported and intentional alternative to XDG paths.
* Treating the existence of `~/.cabal` as the same as a `CABAL_DIR` being set is for backwards compatibility (although I see no reason not to support it for a long time - the only risk is that if something creates `~/.cabal` for some reason, then already existing XDG directory contents will be ignored).

I still need someone to explain [what this prefix business is about](https://github.com/athas/cabal/blob/d5a10e93e86067b32ca9ea19579f5f5a1632fbee/cabal-install/src/Distribution/Client/Config.hs#L553) and what it should default to.

athas:
I have added documentation for how Cabal will use XDG directories.  Still no changelog entry or migration guide.  There are various places where program output help text hardcodes `~/.cabal/config`, which I will change to say `~/.config/cabal/config` instead (the XDG default).  I think this is defensible because it's not like the current text respects `CABAL_DIR` or `CABAL_CONFIG` environment variables anyway.  If someone wants to make this help text responsive, I think that belongs in a followup PR.

athas:
Where should I write the release notes/changelog entry for an unreleased version of Cabal?

With the behaviour as currently implemented (assuming it has no bugs), upgrading is pretty straightforward: as long as you have a ~/.cabal directory, nothing changes.  To switch to the new behaviour, delete ~/.cabal (maybe copy ~/.cabal/config to $XDG_CONFIG_HOME/cabal/config first).  We can perhaps document how to manually split an existing ~/.cabal among the three XDG directories we use, but I do not think anyone will go through that hassle.

ulysses4ever:
@athas 1) add an entry in the `changelog.d` directory; use examples that are already there; 2) update documentation in the `doc` directory (edit the `.rst` files there).

fgaz:
> Where should I write the release notes/changelog entry for an unreleased version of Cabal?

See https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog

> We can perhaps document how to manually split an existing ~/.cabal among the three XDG directories we use, but I do not think anyone will go through that hassle.

I'm not even sure splitting is possible without breaking the store

I think InstallDirs is used in v1-* commands. I personally never touched that section of my ~/.cabal/config

athas:
I suspect I will have to move some of the logic to the Cabal package itself, because it is needed in `Distribution.Simple.InstallDirs` to compute the default prefix.  It is currently `~/.cabal`, but I think it should follow the same rules as other Cabal directories and default to (I guess?) `~/.local` if `CABAL_DIR` is unset.

ulysses4ever:
I also think that InstallDirs is a v1- legacy. As such, we don’t need to update it necessarily; after all, v1- is not maintained **officially**. It will be dropped at some point.

I’m not trying to shut down your effort: if you want to, you can try to implement XDG support in InstallDirs too. I’m just saying this shouldn’t be a blocker or even high priority.

athas:
Alright, I'll keep it simple.  My main worry is that due to the backwards compatibility policy, things go haywire if something creates ~/.cabal by accident (that actually happens right now and I'm trying to figure out what).

athas:
From what I can tell, InstallDirs is not just legacy.  It is also used to populate the config file.

fgaz:
It's also used in Setup.hs:

```
$ runhaskell Setup.hs configure --user
Configuring tmp-i7xOiIOJbE-0.1.0.0...
$ runhaskell Setup.hs install --user
Installing executable tmp-i7xOiIOJbE in /home/fgaz/.cabal/bin
```

Not sure if that reads the config file at all or if it just uses a hardcoded ~/cabal/.bin unless --prefix or --bindir is passed

athas:
Yes, so I'll have to migrate a lot of this logic to the Cabal package itself.  No big deal, and it makes sense in the way that these default directories aren't really "config" anymore.

athas:
I have added a changelog entry and fixed the few minor code nits I mentioned.

I have been dogfooding this version of cabal-install myself for the past few days and it seems to work well.

athas:
I have also updated the original PR text to summarise the current design.

Cabal-tests/Cabal-tests.cabal
```diff
@@ -149,6 +149,7 @@ test-suite hackage-tests
     , Cabal
     , Cabal-syntax
     , Cabal-tree-diff
+    , cabal-install
```
phadej, line 152: Don't pollute `Cabal` tests, with `cabal-install` stuff. 

The finding of config file logic was duplicated on purpose.

Cabal/src/Distribution/Simple/InstallDirs.hs
```diff
@@ -47,6 +47,9 @@ module Distribution.Simple.InstallDirs (
         packageTemplateEnv,
         abiTemplateEnv,
         installDirsTemplateEnv,
+
+        maybeGetCabalDir,
```
phadej, line 51: it's kind of sad that `Cabal` has this functionality.

I'd try to structure this so, that `cabal-install` handles all of backward-compat/migration, and `Cabal` the library can have defaults independent of the legacy.

cabal-install/src/Distribution/Client/ProjectConfig.hs
```diff
@@ -622,7 +622,7 @@ writeProjectConfigFile file =
     writeFile file . showProjectConfig
 
 
--- | Read the user's @~/.cabal/config@ file.
+-- | Read the user's Cabal config file.
```
phadej, line 625: it is `cabal-install`s config file.

phadej (changes requested):
As this is change to `cabal-install` logic, i'm negatively surprised to see any changes in `Cabal`. Either they can be done in preparatory PR (refactorings), or there shouldn't be any.

Cabal-tests/Cabal-tests.cabal
phadej: I emphasize, this is **very** important. `Cabal` is tested with e.g. older GHCs for which in-repo `cabal-install` is not buildable.

Cabal/src/Distribution/Simple/InstallDirs.hs
athas: This is difficult because some of the stuff in InstallDirs influences the default configuration produced in cabal-install - essentially a kind of dynamic cyclic dependency.  I agree it would be better design, but InstallDirs is used many places in Cabal, and I doubt I have the knowledge (or frankly, inclination) to exfiltrate it.

Cabal-tests/Cabal-tests.cabal
athas: I guess I can bring back the duplication.  Basic CABAL_DIR support is perhaps enough for the test suite, if we can assume it is always set.

phadej: It's not. That's why there is logic for `CABAL_CONFIG`.

phadej:
Let's revisit how `cabal-install` works now:

- It finds `cabal.config` first. Its location is set by either `CABAL_CONFIG` env var, or is under `CABAL_DIR`.
- If `cabal.config` is found, done. Work from there.
  - IMO, here we can and should be strict, if some config path is not set, parsing should fail hard. II don't want `cabal-install` to start writing stuff somewhere in my personalized configration.
- If `cabal.config` is not found, there is logic to generate default locations etc.

Note this is about changing two things:
1. where initial `cabal.config` is written (`xdg_config/cabal/config`)
2. what are its contents

And the backward compatibility shim is to simply keep looking for `cabal.config` in new place(s). *Specifically* we shouldn't have a backward compatibility logic for default contents of freshly generated `cabal.config`.

This PR has changes it shouldn't, if you agree with above.

---

Currently that configation discovery flow is obfuscated. I'm also surprised that `Cabal` has some default locations in it, I'm confused why. (It would be very good if there was as little implicit stuff as possible).

If current structure of `Cabal` and `cabal-install` is not good enough to make the XDG change cleanly (which is quite small, based on above), then they need to be refactored first.

athas:
I agree with the above, but could you clarify which parts of the PR are unnecessary?

cabal-install/src/Distribution/Client/Config.hs
```diff
@@ -587,40 +590,62 @@ initialSavedConfig = do
...
-  return $ dir </> "config"
+defaultConfigFile =
+  getDefaultDir XdgConfig "config"
 
 defaultCacheDir :: IO FilePath
-defaultCacheDir = do
-  dir <- getCabalDir
-  return $ dir </> "packages"
+defaultCacheDir =
+  getDefaultDir XdgCache "packages"
```
phadej, line 612: E.g. this one shouldn't look at whether `~/.cabal` folder exists. The default cache dir should be just `getXdgDirectory </> "cabal" </> "packages"`.

phadej: ... but as I tried to say previously, I'm not sure if these functions are **also** used to fill missed fields in cabal `config`. I think that is bad functionality, and `cabal` should rather just fail (and separately provide a way to update the config, there is `cabal user-config update` after all).

athas: These functions are used to fill the missing fields in the config file.  It may be bad functionality, but I tried to preserve complete backwards compatibility in this PR.  In particular, if ~/.cabal already exists or `CABAL_DIR` is set, there will be no change in behaviour.  Changing the interpretation of missing config fields is a different feature, and can be done either before or after this PR, but it would not affect the ease of implementing XDG compliance.

Cabal-tests/Cabal-tests.cabal
athas: I duplicated enough logic to make this work now.

phadej:
I argue that it should be done before.

And I think it's even worth waiting one `cabal-install` release. (So people have time to fill missing entries in their configs).

athas:
That's fair, but it's not work that I am personally motivated to do, mostly because I don't believe it fixes any of the architectural ugliness (the InstallDirs business, which I think is completely orthogonal).

phadej:
`InstallDirs` is indeed orthogonal, but there is no hurry to change that.

And as far as I can see the code it's probably **better** if we don't touch `InstallDirs`. If some old code uses `./Setup configure --user` they most likely want stuff to be installed in `~/.cabal` even if they don't have that directory yet.

In use-cases when `cabal-install` calls `./Setup` the directories are set accordintly.

---

I'm :+1: for killing `./Setup configure --user` configuration. It's a thing of a past. `cabal-instal v2-install` doesn't use it. `stack` doesn't use it. nixpkgs doesn't use it. **BUT** `ghc-pkg` knows about it, on my machine it's

```
% ghc-pkg list --user
/home/phadej/.ghc/x86_64-linux-8.6.5/package.conf.d
```

(if someone needs to install something there, they can still call `./Setup` by explicitly pointing to that directory).

athas:
Not touching `InstallDirs` is perhaps possible, but not completely straightforward, since it influences the default configuration.  I'll have to check whether it only does so in a way that's easy to change.  My suspicion is that it is only used to set the default installation path based on the "prefix", which is a near-vestigial code path now.  I will investigate.

Personally, I am certainly in favour of savagely removing old features and misdesigns, *but* I would rather not have XDG support depend on such changes.  This PR shows that XDG support can be introduced in a pretty simple and backwards-compatible way that doesn't even add any technical debt (although it may piggyback on what's already there).

phadej:
> Not touching InstallDirs is perhaps possible, but not completely straightforward, since it influences the default configuration.

Let's keep changes to `Cabal` and `cabal-install` separate. If `cabal-install` uses `InstallDirs` for its own uses, that's not right, they should be allowed to develop independently.

Especially, if this change to `Cabal` somehow messes with `stack` use of `Cabal` = `./Setup`, that won't be fun.

athas:
It looks like things are working out even with zero changes to `Cabal`.  The only downsides are:

* HackageTests.hs duplicates some of the backwards-compatible lookup junk.
* ~/.cabal is still present as the "user install dir" in the default config (although commented out), but it's my impression that this is an old semi-deprecated feature anyway.

athas:
I have fixed a few doc typos.

I also realised that the `CabalDirLayout` type is now somewhat misleading.  It's not an urgent problem, because it was never complete anyway, and only contains two things: the layout of the store and the path to the log directory.  A followup PR could refactor this type away entirely (I can do that), but it is orthogonal to XDG support, so I don't think it should be part of this PR.

Is there anything else that should be done?  There is probably lots of related cleanup that could be done, but I don't think the changes currently listed in this PR add any technical debt.  The backwards compatibility code is perhaps a bit elaborate, but it is very localised and I think that is better than breaking compatibility (and it could be gradually boiled away once compatibility breaks are accepted, e.g. by making certain config fields mandatory).

Mikolaj:
@ulysses4ever: uhoh, do you see that label?

Mikolaj:
@ulysses4ever: you will have a hell of a story to tell in Mergify TV. :D 

Unless the label is fine, because it does not suffice for merging, and the merge timing was a coincidence (actually, mergify would do a rebase, right)?

ulysses4ever:
@Mikolaj do you mean the "merge delay passed" label was applied prematurely because the PR is not ready and it's just a coincidence that it was not touched for 2 days? If yes, then I agree, and it's easy to fix: just add the same conditions to the add label action (2 approvers, presence of "merge me" label).

I don't understand the question about rebase. But more generally, I don't have a good idea when Mergify rebases. What I am sure of is that: currently, not all conditions for the merge apply (no approvers, no "merge me"), so I'd be surprised if Mergify touches the branch (do a rebase or anything else with it).

Mikolaj:
Please forgive my cryptic remarks. I was worried mergify not only applied the label but also merged master into the PR and would proceed to merge PR to master if not for the conflict. However, it must have been a coincidence and a human merrily did the merge, oblivious to mergify's machinations. So the only surprise is that mergify applies the label in such cases, and it's pretty harmless (but IMHO it's better to avoid surprise).

ulysses4ever:
No worries! I see now. Yes, I hope that merging master was incidental. What generally worries me (unrelated to the delay business) is these force-pushes that seem to come from real people (randomly chosen it seems!) but in fact are results of Mergify trying to adjust the branch (I think). [I recently complained how some rebase came "from me"](https://github.com/haskell/cabal/pull/8383#issuecomment-1217223532) when it was a PR I had nothing to do with. I'd prefer wet fix it somehow...

Sorry for polluting the thread with offtopic. We probably should move somewhere... I'll open a PR fixing the label in a foreseeable future. And will also remove the label from old PRs that got it accidentally.

Mikolaj:
@mergify rebase

Mikolaj:
@athas: Your PR looks great, has tests and a changelog and we have CI under control, though there are some shallow conflicts now.

What we need is some thorough reviews, because this is an important and possibly disruptive change. Let me ask people on #hackage.

athas:
In it's current form this patch is actually not disruptive.  It is (perhaps excessively) backwards compatible, at the cost of more convoluted code.  I don't think a simpler or more backwards-compatible implementation is possible, so it's more of an overall design question whether Cabal wants XDG support at all.

Mikolaj:
> In it's current form this patch is actually not disruptive. It is (perhaps excessively) backwards compatible, at the cost of more convoluted code. I don't think a simpler or more backwards-compatible implementation is possible, so it's more of an overall design question whether Cabal wants XDG support at all.

@athas, excuse my paranoia. Nothing personal. :) I immensely appreciate you backward compat care. An, yes, we want XDG. Just the two routinely required reviews and we are go.

cabal-install/src/Distribution/Client/Config.hs
```diff
@@ -587,40 +590,87 @@ initialSavedConfig = do
...
-    Just dir -> return dir
+    Just dir -> return $ Just dir
+    Nothing -> do
+      defaultDir <- getAppUserDataDirectory "cabal"
+      dotCabalExists <- doesDirectoryExist defaultDir
+      return $ if dotCabalExists
+               then Just defaultDir
+               else Nothing
+
+-- The default behaviour of cabal-install is to use the XDG directory
```
Kleidukos, line 607: Any reason why this is not a proper Haddock declaration? :)

athas: I have the habit of not using Haddock declarations for non-exported functions.  I vaguely recall that there was a good reason for this, but I don't remember it anymore, so I will change this.

Cabal/src/Distribution/Simple/InstallDirs.hs
athas: This is gone now.

athas:
I don't understand why CI fails all of a sudden.

Mikolaj:
Let me re-run the failed bootstrap job. It sounded like a temporary directory could not be found.

Mikolaj:
Still failing, and CI works fine on other PRs (may be non-deterministic, though). This is the likely culprit, merged 4 days ago: https://github.com/haskell/cabal/pull/8368

athas:
I don't understand the implications of that PR, or why it would affect only the 8.8.4 bootstrapping.

Mikolaj:
> why it would affect only the 8.8.4 bootstrapping

CI cancels all the other bootstrapping jobs after one fails, so we don't know if it affects the others.

Mikolaj:
No idea what's going on, but the error mentions "_build/tarballs/directory.cabal" and your PR touches "directory" in bootstrap files and the other PR is about boostrap. May be a coincidence though.

Mikolaj:
Wow, how did the `null` get there? Did the tool for regenerating boostrap files enter it? And how come `null` is fine for the other entries? I'm worried the tool may have deteriorated.

athas:
I suspect I edited those files by hand because the tool didn't work for me at some point (don't remember the details, maybe it was just that I didn't have the right GHCs available).

Mikolaj:
I see. So perhaps `null` is fine if there are no revisions for that version, or the other way around, and perhaps the tool is fine. Thank you.

Mikolaj (approved):
I'm not able to devote as much time to the review as this well thought out PR deserves, but it does look very delicate in introducing the new defaults while respecting the old ones. Also, thank you so much for updating user docs and even code comments in places you don't change directly.

doc/config.rst
```diff
@@ -75,13 +81,39 @@ The configuration file location is determined as follows:
...
 ``cabal-install`` will generate the default one, with directories
-based on ``$CABAL_DIR`` (if set) or ``getAppUserDirectory "cabal"``
-prefix.
+based on ``$CABAL_DIR`` (if set) or according to the XDG Base
+Directory Specification.
+
+Directories
+-----------
+
+Unless the ``CABAL_DIR`` environment variable is set, Cabal will store
```
ulysses4ever, line 96:
```suggestion
Unless the ``CABAL_DIR`` environment variable is set or `~/.cabal` exists, Cabal will store
```

I think this section should be as complete as possible.

doc/config.rst
```diff
@@ -97,7 +129,7 @@ the repository to be the central Hackage server:
 
 The name of the repository is given on the first line, and can be
 anything; packages downloaded from this repository will be cached under
-``~/.cabal/packages/hackage.haskell.org`` (or whatever name you specify;
+``$XDG_CACHE_HOME/cabal/packages/hackage.haskell.org`` (or whatever name you specify;
```
ulysses4ever, line 132: Suddenly switch from explicit ~/… to $XDG? For the sake of consistency, I’m leaning to using $XDG stuff everywhere in the docs, so perhaps another euphemisms like $CABAL_DIR instead of explicit paths.

cabal-install/src/Distribution/Client/Config.hs
```diff
@@ -587,40 +590,87 @@ initialSavedConfig = do
...
   }
 
-defaultCabalDir :: IO FilePath
-defaultCabalDir = getAppUserDataDirectory "cabal"
-
-getCabalDir :: IO FilePath
-getCabalDir = do
+-- | If @CABAL\_DIR@ is set or @~/.cabal@ exists, return that
+-- directory.
+maybeGetCabalDir :: IO (Maybe FilePath)
```
ulysses4ever, line 595: A better name for this function would be `maybeGetDotCabalDir`, because it’s specifically ~/.cabal that it tries to find, not any $CABAL_DIR or anything.

cabal-install/src/Distribution/Client/Config.hs
```diff
@@ -587,40 +590,87 @@ initialSavedConfig = do
...
+  mDir <- maybeGetCabalDir
+  case mDir of
+    Just dir ->
+      return [dir </> "bin"]
+    Nothing -> do
+      dir <- getHomeDirectory
+      return [dir </> ".local" </> "bin"]
 
 defaultInstallPath :: IO FilePath
 defaultInstallPath = do
```
ulysses4ever, line 666: This function now largely duplicates `defaultInstallPrefix`. Couldn’t you just call it and append ‘bin’ to the result?

cabal-install/src/Distribution/Client/HttpUtils.hs
```diff
@@ -500,7 +500,7 @@ wgetTransport prog =
             warningMsg     =  "the 'wget' transport currently doesn't support"
                            ++ " range requests, which wastes network bandwidth."
                            ++ " To fix this, set 'http-transport' to 'curl' or"
-                           ++ " 'plain-http' in '~/.cabal/config'."
+                           ++ " 'plain-http' in '~/.config/cabal/config'."
```
ulysses4ever, line 503: Again, maybe better to use $CABAL_DIR here, because for some people it will be ~/.cabal still…

cabal-install/src/Distribution/Client/Install.hs
```diff
@@ -197,7 +197,7 @@ install verbosity packageDBs repos comp platform progdb
         warn verbosity $ "--root-cmd is no longer supported, "
         ++ "see https://github.com/haskell/cabal/issues/3353"
         ++ " (if you didn't type --root-cmd, comment out root-cmd"
-        ++ " in your ~/.cabal/config file)"
+        ++ " in your ~/.config/cabal/config file)"
```
ulysses4ever, line 200: Same concern about explicit paths.

cabal-install/src/Distribution/Client/Manpage.hs
```diff
@@ -46,7 +46,7 @@ data FileInfo = FileInfo String String -- ^ path, description
 -- | A list of files that should be documented in the manual page.
 files :: [FileInfo]
 files =
-  [ (FileInfo "~/.cabal/config" "The defaults that can be overridden with command-line options.")
+  [ (FileInfo "~/.config/cabal/config" "The defaults that can be overridden with command-line options.")
```
ulysses4ever, line 49: Same concern about explicit paths.

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -356,7 +356,7 @@ commandLineFlagsToProjectConfig globalFlags NixStyleFlags {..} clientInstallFlag
                                   , packageConfigHaddockLinkedSource = packageConfigHaddockLinkedSource pc
                                   })
 
--- | Convert from the types currently used for the user-wide @~/.cabal/config@
+-- | Convert from the types currently used for the user-wide Cabal config
```
ulysses4ever, line 359: Good way to avoid the explicit paths issue. Use it more widely?

cabal-install/src/Distribution/Client/ProjectOrchestration.hs
```diff
@@ -138,7 +138,7 @@ import qualified Distribution.Client.BuildReports.Anonymous as BuildReports
 import qualified Distribution.Client.BuildReports.Storage as BuildReports
          ( storeLocal )
 
-import           Distribution.Client.Config (getCabalDir)
```
ulysses4ever, line 141: I’m silly: how does it work that the import list is empty and yet you’re able to use e.g. mkCabalDirLayout? Empty import is suspicious anyway.

cabal-install/src/Distribution/Client/Sandbox/PackageEnvironment.hs
```diff
@@ -81,7 +81,7 @@ userPackageEnvironmentFile = "cabal.config"
 -- | Type of the current package environment.
 data PackageEnvironmentType
   = UserPackageEnvironment    -- ^ './cabal.config'
-  | AmbientPackageEnvironment -- ^ '~/.cabal/config'
+  | AmbientPackageEnvironment -- ^ '~/.config/cabal/config'
```
ulysses4ever, line 84: Same concern about explicit paths.

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -1270,7 +1270,7 @@ reportCommand = CommandUI {
     commandSynopsis     = "Upload build reports to a remote server.",
     commandDescription  = Nothing,
     commandNotes        = Just $ \_ ->
-         "You can store your Hackage login in the ~/.cabal/config file\n",
+         "You can store your Hackage login in the ~/.config/cabal/config file\n",
```
ulysses4ever, line 1273: Same concern about explicit paths.

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -2354,18 +2354,18 @@ userConfigCommand = CommandUI {
   commandDescription  = Just $ \_ -> wrapText $
        "When upgrading cabal, the set of configuration keys and their default"
     ++ " values may change. This command provides means to merge the existing"
-    ++ " config in ~/.cabal/config"
+    ++ " config in ~/.config/cabal/config"
```
ulysses4ever, line 2357: Same concern about explicit paths here and in the rest of this message (three more times).

ulysses4ever (approved):
Exceptional work, thank you! I’m not really qualified to judge the implementation approach and details here, that’s why I hold off my review for some time, but if it needs be, I think this should be merge even if we don’t get any more reviews. My comments are mostly cosmetic, so judge for yourself if you can address them (of course, I’d be happier if you could address them somehow). Thanks again!

doc/cabal-package.rst
```diff
@@ -1574,7 +1574,7 @@ in ghc's package DB and so we can figure out what the location of the library
 is. Foreign libraries however don't get registered, which means that we'd have
 to have a way of finding out where a platform library got installed (other than by
 searching the ``lib/`` directory). Instead, we install foreign libraries in
-``~/.cabal/lib``, much like we install executables in ``~/.cabal/bin``.
+``~/.cabal/lib``.
```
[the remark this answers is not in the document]
athas: No, `~/.config/cabal/lib` would be wrong.  In an ideal world, it should be somewhere in `$XDG_STATE_HOME` I think.  But the impression I got is that this specific behaviour for installing foreign libraries is legacy, and only used for exotic instances of v1-builds.  I did not change the code for this, so they will indeed install into `~/.cabal/lib`.  If I am wrong, then this is something that should definitely be addressed.

doc/config.rst
```diff
@@ -51,16 +51,22 @@ Various environment variables affect ``cabal-install``.
...
 ``CABAL_DIR``
-   Default content directory for ``cabal-install`` files.
-   Default value is ``getAppUserDataDirectory "cabal"``, which is
-   ``$HOME/.cabal`` on unix systems and ``%APPDATA%\cabal`` in Windows.
+
+   If set, *all* ``cabal-install`` content files will be stored as
+   subdirectories of this directory, including the configuration file
+   if ``CABAL_CONFIG`` is unset.  If ``CABAL_DIR`` is unset, Cabal
+   will store data files according to the XDG Base Directory
+   Specification.
```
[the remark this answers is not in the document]
athas: Done.

doc/cabal-commands.rst
```diff
@@ -121,7 +121,7 @@ Arguments and flags common to some or all commands are:
         $ cabal install --allow-newer=foo:base,lens --allow-newer=bar:time
 
     Finally, one can enable :option:`--allow-newer` permanently by setting
-    ``allow-newer: True`` in the ``~/.cabal/config`` file. Enabling
+    ``allow-newer: True`` in the ``~/.config/cabal/config`` file. Enabling
```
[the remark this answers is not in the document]
athas: Yes, I made the subjective decision to use the default path because it looks cleaner.  This is line with the existing documentation, which also says `~/.cabal/config` instead of `$CABAL_DIR/config`.  I think the vast majority of people will use the default config file location, and the minority who customises their XDG variables will know where to look when the documentation says `~/.config`.

doc/config.rst
athas: I don't have a strong preference.  I use `$XDG_CACHE_HOME` here because the default XDG cache dir is not as well known as `~/.config`.

cabal-install/src/Distribution/Client/Config.hs
athas: cabal-install seems to use the term "cabal dir" for "the directory that contains all cabal-related stuff", i.e. the directory that serves the same purpose as the old `~/.cabal`.  Whether it actually has a leading dot in its name is immaterial.  I tried to preserve that nomenclature.

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
athas: This is fine for an internal Haddock comment that will only be read by Cabal developers.  It is inappropriate for many of the other messages, which show up in CLI output, and may be read by novices.

doc/cabal-package.rst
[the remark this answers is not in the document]
athas: I confused this with something else (the InstallDirs business).  Now I'm actually curious where these foreign libraries get installed, since they haven't shown up in any of my tests.  I wonder if they end up in `~/.local/lib` now, since I don't remember seeing any code dealing with them.  Trying to find a buildable example that uses them...

doc/cabal-commands.rst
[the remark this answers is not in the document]
ulysses4ever: I'm not worried for those who customized XDG vars. But for those who will have preexisting ~/.cabal (which will be used instead of ~/.config) — probably, a vast majority of users. At least in the near future after we release this feature. Does it make sense?

doc/cabal-package.rst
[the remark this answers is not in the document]
athas: Oh, I guess it *is* currently v1-install only:  #6046

But with this patch, v1-install will put the library in `~/.local/lib`, which seems reasonable.

doc/cabal-commands.rst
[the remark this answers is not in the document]
athas: That is a problem, but it wouldn't be solved by using XDG variables in the documentation.  The only completely unambiguous solution is to repeat (or link to) the backwards compatibility rules whenever we mention the config file, but that would perhaps get quite tedious very quickly.

No matter what, I think we will have to suffer through a period where the location of the cabal-install config file will be a bit fuzzy to most users.  Hopefully most will likely not need to touch the config file at all.  The only truly dreadful situation will occur when novices use multiple versions of cabal-install simultaneously...

cabal-install/src/Distribution/Client/Config.hs
ulysses4ever: Ok, now I saw I was wrong it checks all sorts of things, not just ~/.cabal. In that case, I'm confused as to why it's a Maybe. Could  you document when it can give you a nothing? And maybe some more prose as to what our tried to achieve. The current comment seems completely wrong?

athas: I added a clarification.

ulysses4ever: Big improvement, thank you! I'm still not completely sure about the name of the function. How about: `maybeGetAllInOneCabalDir` or `maybeGetPreXdgCabalDir`?

doc/config.rst
ulysses4ever: Fair enough. I think it's probably good enough, thank you.

doc/cabal-package.rst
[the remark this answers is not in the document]
ulysses4ever: Just FYI: we still have plenty of v1 users (I'd add: unfortunately). Please, make sure this will not break for them (at least, not easily :-)).

doc/cabal-commands.rst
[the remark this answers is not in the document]
ulysses4ever: Right. My view is that it'd be more robust to say "Cabal config file" instead of "./config/cabal/config" here and in couple similar places. And the phrase "Cabal config file" could be a hyperlink to a section explaining the file and where you typically find it.

doc/cabal-package.rst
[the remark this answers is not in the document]
athas: It doesn't break; I just tested it.  If `~/.cabal` exists then the library is put in `~/.cabal/lib`.

doc/cabal-commands.rst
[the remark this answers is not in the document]
athas: That is possible in the manual, but not in the CLI help text.  And I expect most novices to read the CLI help text, but not the manual.

doc/cabal-package.rst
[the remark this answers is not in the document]
ulysses4ever: Have you tried to use a v1-installed library that war put in .local/lib?

Also, if.local/lib may be used, is it documented anywhere? I don't see it.

doc/cabal-commands.rst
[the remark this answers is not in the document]
ulysses4ever: That fair. But the current thread, for example, is under a page of the manual.

doc/cabal-package.rst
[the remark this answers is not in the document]
athas: "Use" in what sense?  Aren't foreign libraries for use by C programs that are not themselves built by Cabal?

doc/cabal-commands.rst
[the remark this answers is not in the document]
athas: Alright, I fixed this everywhere in the manual.

doc/cabal-package.rst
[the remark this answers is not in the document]
Mikolaj: > I confused this with something else (the InstallDirs business). Now I'm actually curious where these foreign libraries get installed, since they haven't shown up in any of my tests. I wonder if they end up in `~/.local/lib` now, since I don't remember seeing any code dealing with them. Trying to find a buildable example that uses them...

In case you are still searching, here are some examples: cabal-testsuite/PackageTests/ForeignLibs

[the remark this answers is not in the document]
athas: I managed to cobble a test together from the documentation itself, which is how I confirmed that it now installs to `~/.local/lib` unless `~/.cabal` exists.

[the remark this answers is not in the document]
ulysses4ever: Okay, I read up a bit more on this, and I think it is fine.

Mikolaj:
@athas: when you are ready, please add one of the merge_me labels to this PR and mergify is going to merge after two days of inactivity. Thank you again!

athas:
Great, thanks!  While I have been dogfooding this for a few weeks, it's unlikely that my use of cabal-install touches all of its dark corners.  It'll be interesting to see what breaks. I expect to stick around to fix anything that pops up.

ulysses4ever:
Just in case you missed it, there are two comments above that you didn't reply to.

cabal-install/src/Distribution/Client/Config.hs
athas: Yeah, that's the current mechanism behind it, but I'm not convinced it's the *policy*.  I'd rather duplicate the code to make it clear that conceptually the two things are not connected.

cabal-install/src/Distribution/Client/Config.hs
athas: This is ultimately subjective, and doesn't really matter.  "cabal dir" is a fairly common term of art in the code base, so I'm sticking with that for now.

ulysses4ever:
It seems that this PR is stuck because @phadej requested some changes. @phadej could you kindly look into whether it’s fine now?

Mikolaj:
I've re-requested review from @phadej. I hope this is non-blocking now?

ulysses4ever:
@Mergify refresh

Mikolaj:
Mergify does not cooperate this time, so let me prod it and just get it merged it already. :D

Mikolaj:
@mergify rebase

Mikolaj:
Huh, apparently there are rebase-conflicts. That does not mean there are squash-merge conflicts, but there might be. Perhaps that's what held up Mergify, though a pity we got no indication (I haven't check Mergify queue, though).

@athas: could you rebase one last time? Or would you like me to try a manual squash-merge without rebasing?

athas:
Rebasing the entire commit sequence is not viable - there are over forty commits here.  I have merged master into the branch, which should imply that a squash-merge is conflict-free.  I can squash+rebase manually if you think it'll help.

Mikolaj:
Thank you. No, I think this is enough. Let's just sort out the CI results and I will squash-merge manually.

athas:
I assume these CI failures are unrelated to my changes, and I see that master also fails.  I certainly don't remember touching changelog generation!

Mikolaj:
You are right. It seems we are hit by #7798. I will merge as soon as the other jobs succeed.

ulysses4ever:
Well, squashing manually could certainly increase chances of a smoother merge down the road.

Mikolaj:
I mean I will squash-merge via Mergify or the button. Certainly we don't want commits with conflicts in our repo.

Mikolaj:
Let me merge with master that now has CI fixed and merge ASAP after CI passes this time.

Mikolaj:
Or perhaps mergify does it this time. Let's see.

ulysses4ever:
I'd try to use mergify: I don't see any rush in merging this, and to the contrary: this is a big patch, so it's better not to rush.

Meanwhile the changelog job is failing...

Mikolaj:
OK, let's use mergify, unless @athas objects. Unfortunately, one more merge with CI workaround from master will be needed once the said workaround is merged.

Mikolaj:
Hah, apprently, the changelog CI check was not mandatory, so mergify merged anyway (I'm trying to fix that in #8505).

@athas: thank you again; this is one small step for a man, etc.

athas:
Great, thanks! Since this is a fairly major change in behaviour, I will try to see if I can convince people to try using a prerelease of cabal, so the most significant bugs can be found before the next release.

Mikolaj:
Good call. CI builds artifacts (binaries), so that may be helpful, I guess.

ulysses4ever:
@athas thanks for your effort!

One thing that people did in the past to raise awareness is to write a sorry blog post and announce it on Reddit, Discourse, Haskell-cafe, etc.

Mikolaj:
@athas: BTW, I've just had read you email about the XDG overhaul to haskell-cafe. Great job! I can't find it mentioned on https://discourse.haskell.org/ (and I haven't looked r/haskell nor on IRC/matrix). I think posting a link to that email and perhaps also pasting the contents would be a service to the community. What do you think?

ulysses4ever:
it's also easy to submit a link to the announcement to HaskellWeekly: just PR https://github.com/haskellweekly/haskellweekly. @athas let us know if you need help with posting any of it. We really want to have people try it as much as possible before the release.

athas:
I have posted to the Discourse.  I'll see about submitting a link to Haskell Weekly News.

mouse07410:
Here's my current `~/.cabal` directory contents:
```sh
$ ll ~/.cabal
total 40
drwxr-xr-x   11 ur20980  staff    352 Sep  9 09:51 ./
drwxr-xr-x+ 385 ur20980  staff  12320 Oct 20 13:24 ../
drwxr-xr-x   51 ur20980  staff   1632 Oct 12 20:52 bin/
-rw-r--r--    1 ur20980  staff   5523 Sep  9 09:51 config
-rw-r--r--    1 ur20980  staff   5533 Mar 20  2022 config.backup
drwxr-xr-x    4 ur20980  staff    128 Jul 17 13:22 lib/
drwxr-xr-x   13 ur20980  staff    416 Sep  5 17:10 logs/
drwxr-xr-x    3 ur20980  staff     96 Sep 28  2021 packages/
drwxr-xr-x    4 ur20980  staff    128 Jul 17 14:39 share/
drwxr-xr-x    8 ur20980  staff    256 Sep  5 17:10 store/
-rw-r--r--    1 ur20980  staff     15 Sep 29  2021 world
```

After migration - where exactly would each of those entries go?
Note: special attention to `~/.cabal/lib/` and `~/.cabal/store/` directories, because that's what dynamically built libraries and executables use.

ulysses4ever:
@mouse07410 thanks for trying it! I think, to get the new structure, you have to nuke ~/.cabal first.

athas:
> After migration - where exactly would each of those entries go?

Does [this section in the documentation](https://cabal.readthedocs.io/en/latest/config.html#directories) answer the question?  If not, we should improve the documentation.

I think the documentation does not explain where the `lib` directory ends up (it's `~/.local/lib`), but that's mostly because I'm a bit shaky on when exactly things are put in that directory, so I don't know how to describe it.

mouse07410:
> > After migration - where exactly would each of those entries go?
> 
> 
> 
> Does [this section in the documentation](https://cabal.readthedocs.io/en/latest/config.html#directories) answer the question?  If not, we should improve the documentation.

No, it doesn't, IMHO. 

What I'd find satisfactory, would be an **exhaustive** list of the possible legal entries in the current `~/.cabal` directory with the corresponding mapping of reach to its new location. 

> I think the documentation does not explain where the `lib` directory ends up (it's `~/.local/lib`), but that's mostly because I'm a bit shaky on when exactly things are put in that directory, so I don't know how to describe it.

And IMHO that's **bad**: not only will this collide with the Stack products (currently going to `~/.local/`), but I'm aware of other apps utilizing that directory too. Are we heading towards "Cabal hell" all over again, but on a larger scale? Because cabal-v1, at the very least, did not impact **other** ecosystems, while this change will.

athas:
> What I'd find satisfactory, would be an exhaustive list of the possible legal entries in the current ~/.cabal directory with the mapping of each to its new location.

I don't know what "legal" means in this context, and I'm not sure cabal ever had an exhaustive list of such directories.  While working on this PR, it seemed various parts of the code just asked for "the cabal dir" and then created whatever they needed as subdirectories.  It would be possible to describe movement of the most common directories in the changelog entry, but I'm not sure it's useful in the user's guide itself.

> And IMHO that's bad: not only will this collide with the Stack products (currently going to ~/.local/), but I'm aware of other apps utilizing that directory too. Are we heading towards "Cabal hell" all over again, but on a larger scale?

I don't think so, precisely because many other ecosystems have put stuff in that directory for years, and it doesn't seem to a widely reported problem.  You only have conflicts when you have identically named programs installed by different package managers, which seems to be a rare case.  And cabal at least makes it easy to change the installation directory in the configuration file.  This concern was also brought up in the original issue: https://github.com/haskell/cabal/issues/680

mouse07410:
Ki> I don't know what "legal" means in this context, and I'm not sure cabal ever had an exhaustive list of such directories

To me "legal" in this context means "what cabal itself is supposed to put there", as opposed to whatever something else might. 

> It would be possible to describe movement of the most common directories in the changelog entry, but I'm not sure it's useful in the user's guide itself.

It is darn useful for somebody who's trying to migrate, e.g., dynamic build (100% of all Cabal build on my systems, if you care to know). I did not see where `~/.cabal/store` is supposed to go. 

> You only have conflicts when you have identically named programs installed by different package managers, which seems to be a rare case.

It is pretty rare between, e.g., Cabal and Macports or Apt. It is practically 100% between your proposed Cabal and Stack.

> cabal at least makes it easy to change the installation directory in the configuration file

Does it work for dynamically-linked executables (which have always been second-class citizens)? Where would the dependencies (dynamic/shared libraries) go, and will the loader (that load the binary in question to run) be able to locate them?

> This concern was also brought up in the original issue: #680

I'm glad.

athas:
>  I did not see where ~/.cabal/store is supposed to go.

That might be worth mentioning explicitly.  I'll make a PR that adds it.

> Does it work for dynamically-linked executables (which have always been second-class citizens)? Where would the dependencies (dynamic/shared libraries) go, and will the loader (that load the binary in question to run) be able to locate them?

Where did they go previously?  Since store paths are cryptographic hashes, I'm assuming dynamically linked executables embedded the paths, since there is no way the dynamic linker could find them otherwise.  This should still work.  My understanding is that the lib `~/.cabal/lib` (now `~/.local/lib`) was used only for "foreign libraries", i.e. Haskell libraries compiled to be callable from C.  And while this is not a feature I ever used myself, I got the impression that users of this feature inevitably end up copying the .so files elsewhere anyway.

I can try building a dynamic binary and see if it still works.

athas:
As expected, dynamically linked Haskell executables contain full store paths.  They continue to work as well as they ever did.

mouse07410:
>> Where would the dependencies (dynamic/shared libraries) go, and will the loader (that load the binary in question to run) be able to locate them?
>
> Where did they go previously?

AFAIK, to the GHC version-related subdirectory of `~/.cabal/store`.

> My understanding is that the dir ~/.cabal/lib (now ~/.local/lib) was used only for "foreign libraries", i.e. Haskell libraries compiled to be callable from C. And while this is not a feature I ever used myself, I got the impression that users of this feature inevitably end up copying the .so files elsewhere anyway.

Well, I use it, and often do not copy the .so (or, in my case, .dylib) elsewhere. Sometimes I do. It really depends.

mouse07410:
OK, this got merged - now, what's the answer to the question "where does `~/.cabal/store` get migrated to"?

athas:
`$XDG_STATE_HOME/cabal/store`

jneira:
I know it is maybe too late (o is not?) but maybe it could be useful to make migrations smoother:
* in a minor o major release make it explicitly opt-in with a dedicated global flag: use-xdg-spec or whatever
  * setting it to false in new config files
  * document which are the changes, i.e. where will go each folder and possible caveats
* in the next major release set the global flag to true when writing new config files but honour existing ones flag value
  * and maybe warn about the fact? to encourage it's adoption?
  * maybe warn if you are using xdg explicitly but you still has some .cabal directory
* in the next major release (two after adding the feature) remove the flag handling from code, warn if it is set, and use xdg spec unconditionally

jneira:
I know it is maybe too late (o is not?) but maybe it could be useful to make migrations smoother:
* in a minor o major release make it explicitly opt-in with a dedicated global flag: use-xdg-spec or whatever
  * setting it to false in new config files
  * document which are the changes, i.e. where will go each folder and possible caveats
* in the next major release set the global flag to true when writing new config files but honour existing ones flag value
  * and maybe warn about the fact? to encourage it's adoption?
  * maybe warn if you are using xdg explicitly but you still has some .cabal directory
* in the next major release (two after adding the feature) remove the flag handling from code, warn if it is set, and use xdg spec unconditionally

athas:
> in a minor o major release make it explicitly opt-in with a dedicated global flag: use-xdg-spec or whatever 

This is a bit circular, since XDG behaviour *changes the location of the config file*.  Also, people can already use XDG right now if they want, since all paths can be set in the config file (except for the location of the config file itself).  In truth, all the XDG change really does is:

* Change the default location of the config file.
* Change the *default* directories when generating a config file (or equivalently, the default when the fields are missing from the config file).

A general cabal principle seems to be that all directories are supposed to be specified in the config file, and "using XDG" is mostly about selecting XDG directories by default when generating it.  Only the location of the config file itself is (relatively) baked-in.

If you want, you can get XDG-equivalent behaviour with released `cabal` by setting an environment variable `CABAL_CONFIG=$XDG_CONFIG_HOME/cabal/config` and then setting the various directories in the config file.

jneira:
And agree an exhaustive list of where will go actual directories is essential to help in the migration.
Imo this migration mini guide should be included in the release notes but in a specific section at the very beginning to make it visible

mouse07410:
Am I correct assuming that the old `~/.cabal` location will always be supported?

athas:
I can't predict the future, but I personally see no reason to remove it.  It's useful in some cases, e.g. CI.  I would say it is likely that at some point it will require you to set the `CABAL_DIR` environment variable to point to it (i.e. automatic backwards compatibility may not stick around forever).

PR #7387, opened 2021-05-09
Deduplicate config file path calculation in test.

athas:
Tested with `cabal test Cabal-tests:hackage-tests` (*man* does Cabal have many test suites).

---

### DIFF
--- a/Cabal-tests/Cabal-tests.cabal
+++ b/Cabal-tests/Cabal-tests.cabal
@@ -160,6 +160,7 @@ test-suite hackage-tests
     , bytestring
     , Cabal
     , Cabal-tree-diff
+    , cabal-install
     , containers
     , deepseq
     , directory
--- a/Cabal-tests/tests/HackageTests.hs
+++ b/Cabal-tests/tests/HackageTests.hs
@@ -20,13 +20,12 @@ import Data.Foldable                               (traverse_)
 import Data.List                                   (isPrefixOf, isSuffixOf)
 import Data.Maybe                                  (mapMaybe)
 import Data.Monoid                                 (Sum (..))
+import Distribution.Client.Config                  (getConfigFilePath, defaultCacheDir)
 import Distribution.PackageDescription.Check       (PackageCheck (..), checkPackage)
 import Distribution.PackageDescription.PrettyPrint (showGenericPackageDescription)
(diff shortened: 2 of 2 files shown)


phadej (changes requested):
Now looking at this in own context i see that this is not good. `defaultConfigFile` doesn't consider `$CABAL_CONFIG` environment variable, so that is wrong.

Perfectly there should be a function in `cabal-install` to find the config file (c.f. https://hackage.haskell.org/package/cabal-install-parsers-0.4.1/docs/src/Cabal.Config.html#findConfig) but AFAICS it's backed into `loadRawConfig` (which is not bad, you probably want to read the config - as in here).

Yet, adding `cabal-install` lib as dependency to `Cabal-tests` doesn't feel right. There is "configuration parsing library" screaming to get out, whole `cabal-install` is big.

So, either use `loadRawConfig` to not duplicate any parsing logic, or let's drop this change.

athas:
> Now looking at this in own context i see that this is not good. `defaultConfigFile` doesn't consider `$CABAL_CONFIG` environment variable, so that is wrong.

`defaultConfigFile` uses `getCabalDir`, which looks up the `CABAL_DIR` environment variable AFAICS.

I guess one could use `loadRawConfig`, but the types involved are completely different, so I have no idea what that would actually entail.

athas:
Oh, right, `CABAL_CONFIG`... well, is that needed for this test?  I'm not even completely sure what it is testing.

athas:
Now it uses `getConfigFilePath`, which respects `CABAL_CONFIG`.  I don't see a realistic way to get rid of the `cabal-install` dependency.

phadej:
> I'm not even completely sure what it is testing.

`hackage-tests` tests that whole Hackage's 01-index.tar can be parsed (when we do `.cabal` file parser changes). So it has to find 01-index.tar.

Mikolaj:
@athas: do you need help with this PR or a future incarnation of it? Please shout to me, if so.

athas:
I kind of lost motivation to do anything with it, since I don't fully understand the problem *or* the solution.  Maybe it will become necessary to address as I do some more work on #7386 (particularly when I implement the `CABAL_HOME` fallback behaviour).

Issue #7390, opened 2021-05-11
Make ghc-pkg discoverable from ghc executable


bgamari:
**Describe the bug**
Currently there is no robust way to locate the `ghc-pkg` executable corresponding to a particular compiler. For this reason, Cabal currently implements a variety of heuristics to do the same. However, these can end up failing, in particular when the executable name is version-suffixed.

@hasufell and I discussed this some months ago after a `ghc-up` user encountered this problem, which resulted in [GHC #18807](https://gitlab.haskell.org/ghc/ghc/-/issues/18807). To address this I proposed that `ghc --info` should report the path to the corresponding `ghc-pkg` executable. This solves the problem nicely since Cabal already must query `--info` when configuring a package. 

I have put together a very straightforward GHC implementation as [!4214](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4214). I have also written a [quick patch](https://gitlab.haskell.org/ghc/ghc/-/snippets/2710) implementing the Cabal support.

My sense is that this is the right solution to the problem. However, I'm also happy to hear alternatives. If there are no objections, I will backport to 8.10.5, 9.0.2, and 9.2.1 such that future users won't encounter this issue in the future.

hasufell:
I'll try the patch out tomorrow in conjunction with the GHC patch.

I also noticed that we could investigate another workaround in the following function

https://github.com/haskell/cabal/blob/4f8aeb2c8a0a3638e1af887dc869a17e291c8329/Cabal/src/Distribution/Simple/GHC.hs#L267-L278

What if we canonicalize first and then effectively follow the symlink if it is one... that would make us useally end up in the internal ghc directory, which should always work. No?

The advantage of this would be that it's a backwards compatible fix that doesn't need a GHC patch.

bgamari:
> Does this supersede/obsoletes #6648 or are both valuable?

I think they are essentially duplicates. I just didn't find #6648 until I had already opened this one.

Mikolaj:
How can we help to move this forward? @bgamari, are your patches alive?

andreabedini:
Just go out of curiosity, in what cases ghc-pkg path is not the the same as ghc plus -pkg?

phadej:
> Just go out of curiosity, in what cases ghc-pkg path is not the the same as ghc plus -pkg?

- `ghc-9.6.2` and `ghc-pkg-9.6.2`
- `ghc-9.8-alpha1` and `ghc-pkg-9.8-alpha1`

[...]

PR #7392, opened 2021-05-12
Prefer canonicalized path when guessing tools from GHC path

hasufell:
Motivation
----------
Often times, the user facing `ghc` binary is
symlinked by other forces, such as the package manager,
tooling like ghcup etc. As such, the naming convention
(version suffix in particular) may not align with the
assumptions made in Cabal and it may find an incorrect ghc-pkg.

See:
  - https://github.com/haskell/cabal/issues/7390
  - https://gitlab.haskell.org/ghc/ghc/-/issues/18807
  - https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/73

Solution
--------
Guessing the ghc-pkg path is already a hack and will be solved
more appropriately in the future, see
  - https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4214
  - https://gitlab.haskell.org/ghc/ghc/-/snippets/2710
These patches will solve the issue for future GHC versions.

As such, this patch provides a workaround for
older, already existing GHC versions by first always
following the symbolic link of the ghc binary (if it is one)
and prefering its target directory as the guess lookup
location.

Rationale
---------
The canonicalized path of the ghc binary usually points to the
bin/ directory unpacked from a bindist, which is less likely to be
tampered with by distributions and tools. As such, prefering the
canoncialized path should get us more robust results.

---

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -240,7 +240,7 @@ guessToolFromGhcPath tool ghcProg verbosity searchpath
                              | otherwise = [guessGhcVersioned dir suf,
                                             guessVersioned dir suf,
                                             guessNormal dir]
-           -- oder matters here, see https://github.com/haskell/cabal/issues/7390
+           -- order matters here, see https://github.com/haskell/cabal/issues/7390
            guesses = (if real_path == given_path
                         then []
                         else mkGuesses real_dir real_suf)
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -240,10 +240,11 @@ guessToolFromGhcPath tool ghcProg verbosity searchpath
                              | otherwise = [guessGhcVersioned dir suf,
                                             guessVersioned dir suf,
                                             guessNormal dir]
-           guesses = mkGuesses given_dir given_suf ++
-                            if real_path == given_path
-                                then []
-                                else mkGuesses real_dir real_suf
+           -- oder matters here, see https://github.com/haskell/cabal/issues/7390
+           guesses = (if real_path == given_path
+                        then []
+                        else mkGuesses real_dir real_suf)
+                     ++ mkGuesses given_dir given_suf
        info verbosity $ "looking for tool " ++ toolname
          ++ " near compiler in " ++ given_dir
        debug verbosity $ "candidate locations: " ++ show guesses
--- a/cabal-testsuite/PackageTests/GhcPkgGuess/SameDirectory/setup.cabal.out
+++ b/cabal-testsuite/PackageTests/GhcPkgGuess/SameDirectory/setup.cabal.out
@@ -1,3 +1,3 @@
 # Setup configure
 Configuring SameDirectory-0.1.0.0...
-cabal: Version mismatch between ghc and ghc-pkg: <ROOT>/./ghc is version <GHCVER> <ROOT>/./ghc-pkg is version 9999999
+cabal: Version mismatch between ghc and ghc-pkg: <ROOT>/./ghc is version <GHCVER> <ROOT>/ghc-pkg is version 9999999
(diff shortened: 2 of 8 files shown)


hasufell:
summoning some nix guys to check whether this would break something: @rvl @dmjio

Cabal/src/Distribution/Simple/GHC.hs
```diff
@@ -240,10 +240,11 @@ guessToolFromGhcPath tool ghcProg verbosity searchpath
                              | otherwise = [guessGhcVersioned dir suf,
                                             guessVersioned dir suf,
                                             guessNormal dir]
-           guesses = mkGuesses given_dir given_suf ++
-                            if real_path == given_path
-                                then []
-                                else mkGuesses real_dir real_suf
+           -- oder matters here, see https://github.com/haskell/cabal/issues/7390
```
maralorn, line 243:
```suggestion
           -- order matters here, see https://github.com/haskell/cabal/issues/7390
```

maralorn:
I am new to this logic, so I am not very confident about my judgement. I will document my understanding here so you can correct any misconceptions.

* This logic discovers the matching `ghc-pkg` binary belonging to the `ghc` binary cabal knows from PATH.
* If `~/.nix-profile/bin/ghc` is a symlink to `/nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc` this change will lead to `/nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc-pkg` being prefered over `~/.nix-profile/bin/ghc-pkg`.
* This code does not know that `/nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc` is actually a shell-wrapper around another ghc binary and so does not canonicalize the path further.

If this understand is correct I think this change is actually a slight improvement because it sounds at least possible that the user could find a way to have mismatching `ghc` and `ghc-pkg` in their PATH.

Yeah, I think this is fine for nixpkgs.

hasufell:
> This logic discovers the matching ghc-pkg binary belonging to the ghc binary cabal knows from PATH.

Yes

> If ~/.nix-profile/bin/ghc is a symlink to /nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc this change will lead to /nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc-pkg being prefered over ~/.nix-profile/bin/ghc-pkg.

Correct

> This code does not know that /nix/store/qi93hiqajr9v8lw23wjawyfnl4iwlmrs-ghc-8.10.4-with-packages/bin/ghc is actually a shell-wrapper around another ghc binary and so does not canonicalize the path further.

Right, canoncalization only means "follow symbolic links". For hardlinks or anything else, this is the same behavior.

emilypi (approved):

emilypi:
@hasufell I think this would benefit from a changelog entry. Do you mind writing one up?
@bgamari does this look okay to you? if yes, we'll merge.

hasufell:
> @hasufell I think this would benefit from a changelog entry. Do you mind writing one up?

Which one? They look [unmaintained](https://github.com/haskell/cabal/blob/master/cabal-install/changelog) [to me](https://github.com/haskell/cabal/blob/master/Cabal/ChangeLog.md).

[Release notes](https://github.com/haskell/cabal/tree/master/release-notes) don't seem to contain a section for the next version yet.

emilypi:
Just adding something in the `changelog.d` directory is fine. I'll aggregate and update in follow up PR's. I have one in the Cabal Init Omnibus one that updates the changelog completely.

hasufell:
Done

emilypi:
When CI is green, I'll merge @hasufell

hasufell:
:green_circle:

PR #7397, opened 2021-05-17
Add regression test for #5782, which requires extending test harness to v2-install

Mikolaj:
This adds a regression test for #5782.

I've extended cabal-testsuite harness to exes taken from v2-install (it assumed in-place install, as in v2-build, so AFAICT no amount of extra flags would help). And there is a separate issue that for cabal <= 3.2 my test requires a dummy `--installdir` option, but I hope we don't run regtression tests with old cabals? Otherwise it would fail anyway, because the test fails in cabal 3.2, being a regression test for an issue fixed (for good) in 3.4

---


Mikolaj:
@fgaz: I bet I missed some thing one needs to do with a new test, like disable it for too old cabals, etc. Any hints appreciated, especially as checks fail...

cabal-testsuite/PackageTests/Regression/T5782Diamond/cabal.test.hs
```diff
@@ -0,0 +1,34 @@
...
+--
+-- The dummy "--installdir=." is needed for cabal <= 3.2
+-- and also to match cabal output on different OSes
+-- (default installdir is different on various OSes).
+
+import Test.Cabal.Prelude
+main = cabalTest $
+    withSourceCopy . withDelay $ do
+        writeSourceFile "issue5782/src/Module.hs" "module Module where\nf = \"AAA\""
+        cabal "v2-install" ["--installdir=.", "--overwrite-policy=always", "issue5782"]
```
phadej, line 19: I don't remember, does '`cabal-testsuite` specify config with non-user `store-dir`.

I think it should (and could specify `install-dir` which it normalizes as well when comparing outputs), so users' store is not polluted, neither affect runs.

Mikolaj: Do you mean something like https://github.com/haskell/cabal/blob/master/cabal-testsuite/PackageTests/Regression/T5409/use-different-versions-of-dependency-for-library-and-build-tool.test.hs#L18

phadej: @Mikolaj that looks better though is still ugly.

`cabal-testsuite` is unfortunately not very good at isolating test environment (I mostly run it in docker container, see `Makefile`), which is not very nice.

Mikolaj: Thank you for the tips.  `withShorterPathForNewBuildStore` and Co let me avoid some path mismatches, etc., in the output, but MacOS still insists on processing internal libraries in a different order and Windows additionally still can't recognize that paths match. So I ended up adding `recordMode DoNotRecord`, which is fine in this case, because the problem manifests either as failed compilation or wrong exe output, which I do check.

Mikolaj:
This is now ready for review.

The only failing CI test is continuous-integration/appveyor/pr where I'm consistently getting 'heap exhauted' failures, e.g., https://ci.appveyor.com/project/23Skidoo/cabal/builds/39255946 but the PR can't be causing that, so it's unrelated. 

BTW, I don't have admin access to the AppVeyor CI and so can't, e.g., restart it; I wonder if it should be somehow granted based on github roles. 

BTW2, the 'Checks' menu of github doesn't know about AppVeyor --- are any other tests not present there? I guess this can't be easily fixed (no clue about github CI and UI).

Mikolaj:
@jneira: would you add a quick review, given that this affair is based in part on your findings?

jneira:
Hi, thanks for the mention (and the invitation! 😄 )
The test looke totally reasonable, to have a case of a diamond dependency seems to be good even if no there was not real bug before.

jneira (approved):
lgtm, thanks for take care of the possible regression

Mikolaj:
@jneira: thank you for the review and your reports and troubleshooting.

emilypi (approved):
LGTM

Mikolaj:
I've rebased and the CI failure about heap exhausted vanished, which proves it was caused by the state of master branch at that time (rummaging in CI history seems to confirm that). Right now I'm getting MacOS failures, in turn, but surely not my fault either --- probably an outage of wherever we are taking the MacOS archives from.

@fgaz, @phadej: I will merge as soon as you say 'go' or I get bored waiting. :)

emilypi:
@Mikolaj go

Mikolaj:
I go.

PR #7402, opened 2021-05-19
cabal v2-configure

### LINKED ISSUE #5591: Implement a 'v2-reconfigure' command
For old-build we have a useful `reconfigure` command (originally implemented by @ttuegel), which is like `configure`, but instead of fully overwriting the old settings file, it `mappend`s the new flags on top of the old ones and records that. I think it'd be useful to have this for `v2-build` as well.

/cc @bugthunk

### LINKED ISSUE #7180: Repeated v2-configure floods working directory with cabal.project.local files.
**Describe the bug**

Every time `cabal v2-configure --verbose=0` is ran, a new `cabal.project.local` file is created with the old one being moved to a new filename. This can very quickly flood the current directory which is especially annoying in IDE file explorers and such.

**To Reproduce**
Go to any cabal-capable Haskell project. Run

```console
$ ls -lh cabal.project*
ls: cannot access 'cabal.project*': No such file or directory

$ cabal v2-configure --verbose=0

$ cabal v2-configure --verbose=0

$ ls -lh cabal.project*
-rw-r--r-- 1 shana nogroup 33 Nov 25 09:26 cabal.project.local
-rw-r--r-- 1 shana nogroup 33 Nov 25 09:26 cabal.project.local~
```

The problem is made worse by the fact that the contents of each of these files is the same
```console
$ for f in cabal.project.local*; do echo $f; cat $f; done
cabal.project.local
verbose: 0
ignore-project: False
cabal.project.local~
verbose: 0
ignore-project: False
```

I have tried adding `verbose: 0` (not what I actually want though) and `ignore-project` into `cabal.project` file but the file is made regardless.

**Expected behavior**
My wishlist:
* allow a flag which stops the cabal.project.local file from being created at all
* if the file we're overwriting with is the same, don't make a new copy

**System information**
```
Linux aya 5.4.70 #1-NixOS SMP Wed Oct 7 06:01:31 UTC 2020 x86_64 GNU/Linux
cabal-install version 3.4.0.0
```
I'm using the 3.4-rc4 commit of cabal-install: 81fd8d935a25212d9b38c9b8b41fac2c22d42d58

**Additional context**
My use-case for running configure so many times is that we use a Shake-based build system to build our code but use cabal to provide all the dependencies. This means that we need the `plan.json` file to be present and up to date and so run `v2-configure` for it. Likewise, this file is needed for `cabal-cache` to work. I want to pass `--verbose=0` in configure stage only as I don't want to see the full configure output in middle of other messages from the build system. I can try to limit how often the configure is ran but the problem remains in that very quickly working directory ends up full of `cabal.project.local~N` files. I could remove these after running configure but that seems potentially wrong lest we do actually want to put some extra flags in there.

### LINKED ISSUE #7405: RFC: What to do about v2-reconfigure
Per a discussion in #7402, in which @ptkato implemented a solution for #5591, @phadej linked a discussion from #7180  in which there are some differences pointed out between `v1-(re)configure` and `v2-(re)configure`. Namely, the following: `v2-configure` is not a useful command, and therefore `v2-reconfigure` is similarly not useful. `v2-configure` does two things: generate a build plan, and create a `cabal.project.local` with project-local settings. The reasons this is not useful are two-fold: 

1. `v2-build` already  generates a build plan autonomously, making the `v2-configure` build plan step redundant 
2. modification of the `cabal.project.local` via command-line args passed to `v2-configure` is under-specified. One can only pass a subset of valid `cabal.project.local` flags via command line (Oleg points out that you cannot specify remote source repositories, for example). So in that sense, the implementation is under-specified and hides some implementation details. 

Here are a few ideas for what we can do here: 

1. Deprecate `v2-configure` and tell the users why it's not a bad idea, tossing away the `v2-reconfigure` work from #7402. Many people would probably balk at this, since it removes a useful workflow from people's tooling (script-local modifications are faster than hand-editing the local project files) there is a proliferation of content out there that alludes to `v1-configure`, and we'd have to be very loud and explicit about why that command in particular was deprecated. 

2. Modify `v2-configure` and re-spec it's purpose so that it only does modifications of the `cabal.project.local`, and see if we can come up with a decent middle ground for what flags we support on command line. 

Your thoughts welcome.

### LINKED ISSUE #7411: Remove the pre-build phase from v2-configure
This issue is only for documentation purposes and will be closed by #7402.

As of #7405, the [pre-build phase](https://github.com/haskell/cabal/blob/a39d590e01636a14986d34cd204079cba45b2d02/cabal-install/src/Distribution/Client/CmdConfigure.hs#L122-L142) shall be removed from the `v2-configure` command altogether.

ptkato:
~~This PR fixes #5591 and also adds a few tests to `v2-configure` and `v2-reconfigure` both.~~

After #7405, this PR now fixes #5591, fixes #7180, fixes #7411, and closes #7405, while also adding tests for this new implementation.

---


cabal-install/src/Distribution/Client/CmdConfigure.hs
```diff
@@ -140,9 +149,101 @@ configureAction flags@NixStyleFlags {..} _extraArgs globalFlags = do
...
+        readProjectLocalExtraConfig v (distDirLayout baseCtx)
+
+      buildCtx <- runProjectPreBuildPhase v baseCtx $ \plan -> 
+        return (plan, Map.empty)
+
+      printPlan v 
+        (baseCtx {buildSettings = (buildSettings baseCtx) {buildSettingDryRun = True}})
+        buildCtx
+
+      return $ Right (baseCtx, configurationMerge conf cliConfig)
```
emilypi, line 237: It doesn't seem useful here to lift these into an `Either`. You just end up calling `either go go` on them later, which implies the branching is unnecessary

emilypi, line 237:
```suggestion
      return $ Right (baseCtx, conf <> cliConfig)
```

cabal-install/src/Distribution/Client/CmdConfigure.hs
```diff
@@ -140,9 +149,101 @@ configureAction flags@NixStyleFlags {..} _extraArgs globalFlags = do
...
+    v = fromFlagOrDefault normal (configVerbosity configFlags)
+    cliConfig = commandLineFlagsToProjectConfig globalFlags flags mempty
+    quote s = "'" <> s <> "'"
+
+configurationMerge
+    :: ProjectConfig -- The current configuration
+    -> ProjectConfig -- The configuration to merge
+    -> ProjectConfig
+configurationMerge config cliConfig =
+    config <> cliConfig
```
emilypi, line 249: This doesn't meet my Fairbarn threshold. It makes more sense just to use `<>` and not introduce a new name.

cabal-install/src/Distribution/Client/CmdConfigure.hs
```diff
@@ -140,9 +149,101 @@ configureAction flags@NixStyleFlags {..} _extraArgs globalFlags = do
...
+        quote (takeFileName localFile)
+        <> " doesn't exist, calling configure instead."
+
+      -- @configureAction'@ is called, as opposed to @configureAction@, to avoid
+      -- calling on the writing function for configure, thusly maintaining the 
+      -- separation on the behaviour that modify files
+      Left <$> configureAction' flags extraArgs globalFlags
+  
+    else do
+      conf <- runRebuild (distProjectRootDirectory . distDirLayout $ baseCtx) $
```
emilypi, line 227: What does running a rebuild do here?

emilypi (commented):
On first pass, this is looking good.

cabal-install/src/Distribution/Client/CmdConfigure.hs
ptkato: Nothing really, it's just so `readProjectLocalExtraConfig` is wrapped by the `Rebuild` monad.

cabal-install/src/Distribution/Client/CmdConfigure.hs
ptkato: I thought about that, but decided to go this route to make it easier to modify it afterwards, if a more complex kind of configuration merging takes place. I have no strong feelings on this, this function can be dissolved into the other one, if desired.

cabal-install/src/Distribution/Client/CmdConfigure.hs
ptkato: Ya, I actually don't like that either, no pun intended. The thing is, I used it that way because in a case within the `reconfigureAction'`, I'm calling `configureAction'`, that way I can differ one thing from the other when actually doing the tests.

emilypi: Coverage gives you that data. It tells you what branch you're going down and if it's covered by a particular test. I don't think you should embed branch coverage tests into the user-facing source code

cabal-install/src/Distribution/Client/CmdConfigure.hs
emilypi: I don't think we'll have a more complicated merge than this one :smile:

cabal-install/src/Distribution/Client/CmdConfigure.hs
ptkato: It's not a matter of what is being covered, it's about being able to tell when it was reconfigure or configure within reconfigure that was called. If those two are perceived as equals, then I do agree with you, otherwise I find it harmless to `Either`ise that, since it goes beyond the testing.

ptkato: However, I can see the wisdom in dismissing that as an `Either`, after all it does add another layer into the fold. If that does pose to be more complicated than it's worth down the road, I won't oppose its refactor.

phadej:
I still don't understand what you need `v2-(re)configure` for. See discussion in https://github.com/haskell/cabal/issues/7180

I'm :-1: on adding this command.

Mikolaj:
Still, this #5591 was milestoned and re-milestoned a few times. That's not the same as approval, but quite unfortunate anyway, if the issue doesn't make sense. The very fact that #5591 and #7180 are not cross-linked is very unfortunate, though understandable in a project of this size with no project managers and devs doing at once coding/management/desing/troubleshooting/etc.

ptkato:
> I still don't understand what you need `v2-(re)configure` for. See discussion in #7180 (I assume people want to use `configure` though none of our docs suggest it, do they - let's fix docs!)

You're suggesting to, instead of calling on a separate `v2-reconfigure` command, call instead on `v2-configure` together with that extra flag? That wouldn't change the workflow of the actual code in this PR much, I suppose that's more of a semantics thingy, and would be easy enough to accomplish, yes.

phadej:
@ptkato I'm suggesting to not use `v2-configure` at all

emilypi:
@phadej I'm afraid I disagree. If i'm interfacing with other people's projects, and they have a fixed `cabal.project`, i tend to lean on `v2-configure` to set my project-local overrides and other settings `-O0`, coverage, testing, profiling etc. This is an important workflow. Reconfigure, I could see being a flag that overwrites and doesn't generate a new `cabal.project.local` backup, and I think that's useful as well. 

I'm curious to know why you think what you think, since it's not really clear from the issue you're citing, and it's not detailed anywhere in any issue. Right now, this MR is a cabal contributors worst nightmare: obscure comments from an issue from last year, not linked to the actual ticket, no comments _on_ the actual ticket, and an immediate thumbs down from you. So, I'd like to hear why that is.

phadej:
@emilypi yes, your use case is valid. I simply write `cabal.project.local` by hand, and edit it if I need something. Why I do so? because I see the state when I open an editor, and don't have to remember what I already configured and how and what I didn't. 

Your use case is **scriptable modification of `cabal.project.local`**, like `cabal user-config` and I'd be :+1: on making command for that (say `cabal package-config`). `v2-configure` and `v2-reconfigure` do more and **give wrong impression that you __have__ to run them to do build**, that is not true. Just `cabal v2-build`.

/to repeat/ `v2-configure` is not the spiritual successor of `v1-configure` but too many people think it is, that's why I think `v2-configure` is bad. (Note how `v2-install` is not the same as `v1-install` but in many issues people think they are interchangeable - e.g. for installing libraries).

From https://github.com/haskell/cabal/issues/5591, 

> For old-build we have a useful reconfigure command (originally implemented by @ttuegel), which is like configure, but instead of fully overwriting the old settings file, it mappends the new flags on top of the old ones and records that.

But `v2-configure` **is not like `v1-configure`**. The premise is false.

phadej:
And addition, more often then not I add `head.hackage` or some other (local no+index) repository or `source-repository-package`s, which `v2-configure` cannot. For `head.hackage` there a snippet you copy&paste, for `source-repository-package` you need repository url and commit, supplying either via command line interface is *in my opinion* non ergonomic.

For optimizations likely there is short `-O0` flag. Maybe if there were short flags for `--enable-tests` `--enable-profiling` the `v2-build` interface would be slightly more convenient too. (And you wouldn't write `-O0 --enable-profiling` by accident, with `v2-reconfigure` you may).

emilypi:
@phadej Alright, so correct me if I'm wrong, but the subtlety is this: 

1. `v2-configure` creates a build plan via dry-run *and* `cabal.project.local` with build flags passed in, and this is doing more than just modifying the `cabal.project.local`. Therefore, this command is redundant with `cabal build --dry` and <insert .local modification script command here>, and at the moment, we do not have the latter command. 

2. Being the ability to do `cabal.project.local` modifications can't possibly be fully encapsulated in the command, because there are some flags you can pass to a `cabal.project.local` file which are not describable as build flags (e.g. your `source-repository-package` example), and so a command that modifies the local project file would need immediate qualification in the docs, saying what subset of options it does and does not support. 

I think i see better now. I was wondering up above why we need to call `build` and `rebuild` if all we're doing is `mappend`'ing the options the file. This makes more sense. I'll open up an issue today and link against this PR and the relevant issues above and we can take discussion there.

emilypi:
@ptkato we have a decision at #7405

ptkato:
I see, however the plans there are to change multiple behaviours, does the removal of the `--dry-run` thingy even belong within this PR? I personally don't mind to sneak it in here, by the by.

ptkato:
Hello there, requesting review

@emilypi @fgaz @Mikolaj

Mikolaj:
Hey, congrats on completing the PR. I was the least engaged (except for trolling users), but others turn out too busy, I will gladly review.

Re "Merging master", I don't want to be a rebase nazi, but would it be feasible to rebase instead of merging? I know with long running PR and complex histories it's no longer easy to rebase, but perhaps here we can afford that luxury?

ptkato:
> Re "Merging master", I don't want to be a rebase nazi, but would it be feasible to rebase instead of merging? I know with long running PRs and complex histories it's no longer easy to rebase, but perhaps here we can afford that luxury?

Done.

Mikolaj:
> Done.

Yay, thank you. :)

ptkato:
I could even squash those earlier commits, since they are outdated, now that the implementation changed.

Mikolaj:
Yep, I like dirt history, too, just not tangled history. :)

emilypi (approved):
Looks great @ptkato. Do you mind opening up an issue in conjunction with this PR for stripping the builds/rebuilds from configure? I think that's the only step I'd follow up with. Other than that, we're free to bikeshed whether `--append` is the correct choice of name. personally, it makes sense, but i'd like to hear what others think

ptkato:
> Other than that, we're free to bikeshed whether `--append` is the correct choice of name. personally, it makes sense, but i'd like to hear what others think.

About that, the way it is set up, it will add two commands for each one: `--enable-append`, `--disable-append`, `--enable-overwrite`, and `--disable-overwrite`. Regarding the actual names, I think `--append` is fine, however `--overwrite` is terrible and quite ambiguous, maybe something like `--no-backup` (or something of the sort) might be better.

Cabal/src/Distribution/Simple/Setup.hs
```diff
@@ -201,6 +201,11 @@ data ConfigFlags = ConfigFlags {
     configPrograms_     :: Option' (Last' ProgramDb), -- ^All programs that
                                                       -- @cabal@ may run
 
+    configAppend        :: Flag Bool, -- ^specifies whether or not the @configure@
+                                      -- command should overwrite or append flags
+    configOverwrite     :: Flag Bool, -- ^specifies whether or not the @configure@
+                                      -- command should backup the configuration
+
```
fgaz, line 208: This is `Cabal`, but those are `cabal-install` flags, right? In that case, they should go in `cabal-install` (like eg. `v2-install` has its own additional `ClientInstallFlags`)

fgaz (changes requested):

Cabal/src/Distribution/Simple/Setup.hs
ptkato: What does that actually entails? Because it seems that a "second" `ClientInstallFlags` would just bloat everything, since it is everywhere within `cabal-install`. Is there any other option?

fgaz: > would just bloat everything, since it is everywhere within cabal-install

I'm not sure I understand what you mean. Wouldn't it be somewhat confined to the configure code, in the command-specific flags?

fgaz: I see what you mean. It's far more widespread than I thought.

Ideally we'd find a way to isolate the flags... but that's out of scope.

I guess you can put it in ConfigExFlags.

ptkato: Done.

fgaz: Thanks!

fgaz:
I know you already discussed it, but since we changed the merge strategy to "rebase", wouldn't it be better to squash this, for the sake of clean bisects (that cannot ignore intermediate commit without a merge commit) and to not make blames more difficult due to temporary changes?

Mikolaj:
Oh, I didn't know rebase has this interaction with `git bisect`. I take full blame for advocating rebase over merge. If for any reason merge seems better in a particular case, please feel free to merge manually --- the github UI setting are meant to help the devs by indicating an aspirational preference, but not to hinder them by outlawing the other choices. :)

ptkato:
Just for the record, that sack of commits there is because I accidentally had one of the merge commits included in the rebase without noticing it.

ptkato:
> Oh, I didn't know rebase has this interaction with `git bisect`. I take full blame for advocating rebase over merge. If for any reason merge seems better in a particular case, please feel free to merge manually --- the github UI setting are meant to help the devs by indicating an aspirational preference, but not to hinder them by outlawing the other choices. :)

Wait, is that so? I thought we were rebasing only the local changes, and then the PR itself would be merged into master.

fgaz:
Yeah, the merge strategy in the github repo settings was recently changed, this is what I see:

2021-05-31-190606_393x376_scrot

Mikolaj:
I'm fine with merging a series of commits, as long as they are fully rebased themselves (this is equivalent to `git rebase; git merge --no-fast-forward` --- normally merge commit is not created if things can be fast-forwarded). Is there a way to suggest this workflow is welcome and, at the same time, enforce that merged commits are fast-forward? I know there is in bitbucket, but I don't know about git.

Alternatively, would we get the same result if you, @ptkato, did `git merge --no-fast-forward` and then clicked the github button?

phadej:
FWIW, rebase and merge also doesn't leave a thread of original PRs in the git messages, so you are more or less forced to use GitHub UI to find which PR commit originated from.

I'm not a fan,

Mikolaj:
You are right, when the author does a manual `git merge --no-ff`, there is no guarantee the PR number would end up in the commit message (I guess git only suggests to put the branch name there or not even that?). I forgot github inserts the nice `Merge pull request #NN from <branch>`.

OK, I think I was wrong suggesting to remove the github-merge option. If so, squash is probably fine, too. However, can we automatically warn the PR author when the branch is not fully rebased (or even block merging until then)?

Cabal/src/Distribution/Simple/Setup.hs
ptkato: Is this request stuck? It shows as resolved, but at the bottom it still points to it.

Mikolaj: I have no clue if "resolve conversation" counts, or if @fgaz needs to resolve or if this doesn't matter nad @fgaz needs to mark his interview as "approved".

ptkato: It was marked resolved by @fgaz, I unresolved and resolved again just to double check.

fgaz: I think I have to mark it as approved or dismiss the review. Let me take a quick look (the last time I stopped at the flags) and I'll do it

cabal-install/src/Distribution/Client/CmdConfigure.hs
```diff
@@ -88,61 +89,39 @@ configureCommand = CommandUI {
...
-
-    -- TODO: Hmm, but we don't have any targets. Currently this prints
-    -- what we would build if we were to build everything. Could pick
-    -- implicit target like "."
-    --
-    -- TODO: should we say what's in the project (+deps) as a whole?
-    printPlan verbosity baseCtx' buildCtx
+    let overwrites = fromFlagOrDefault False $ configOverwrite configExFlags
+        appends    = fromFlagOrDefault False $ configAppend    configExFlags
+        backup     = localFile ++ "~"
```
fgaz, line 107: The rest of the file uses `<>`

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -637,6 +639,14 @@ configureExOptions _showOrParseArgs src =
       (reqArg "VERSION" (parsecToReadE ("Cannot parse cabal lib version: "++)
                                     (fmap toFlag parsec))
                         (map prettyShow. flagToList))
+  , option "" ["append"]
+      "appending the new config to the old config file"
+      configAppend (\v flags -> flags { configAppend = v })
+      (boolOpt [] [])
+  , option "" ["overwrite"]
+      "the backup of the config file before any alterations"
```
fgaz, line 647: This is not what the flag does. In fact, it's the opposite.

Rather than fixing the help text, I suggest changing `overwrite` to `backup` and negating its current effect, since _backups_ are what the flag is actually controlling, and `overwrite` looks like the opposite of `append`.

Let's discuss it in the rfc: https://github.com/haskell/cabal/issues/7405#issuecomment-851697157

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -637,6 +639,14 @@ configureExOptions _showOrParseArgs src =
       (reqArg "VERSION" (parsecToReadE ("Cannot parse cabal lib version: "++)
                                     (fmap toFlag parsec))
                         (map prettyShow. flagToList))
+  , option "" ["append"]
+      "appending the new config to the old config file"
```
fgaz, line 643: What about:
```suggestion
      "appending the given configuration to the one in cabal.project.local"
```

fgaz (changes requested):
Looks good, but the `overwrite` help needs fixing, or we could change the flag. Also some minor suggestions

cabal-install/src/Distribution/Client/Setup.hs
ptkato: I avoided calling it `cabal.project.local` explicitly because it could be called something else, through the use of the `--project-file` flag, but that's probably overzealousness on my part.

fgaz: no I agree, that's a good point. Maybe "appending the given configuration to the one in the configuration file". Or feel free to leave it as is

cabal-install/src/Distribution/Client/CmdConfigure.hs
ptkato: Done.

Mikolaj:
Regarding merge buttons, I've got a helpful suggestion on IRC and came up with a plan. Does it make sense? @fgaz, @emilypi, @ptkato, others?

> 10:42 < mikolaj> hsyl20: thank you! what I'd like to do is bring back "allow merge commits", but avoid non-trivial merges performed by github --- it seems "Require branches to be up to date before merging" does the job
> 10:43 < mikolaj> "require linear history" forbids even fast-forward merges via the github button at least and perhaps at all
> 10:44 < mikolaj> I'm inclied to bring back all the three merge buttons, check "Require branches to be up to date before merging" and, while we are at, increase the number of required reviews from 1 to 2, matching our culture (informal 
>                  policy)
> 10:44 < mikolaj> unless we make exceptions for trivial PRs and require only one review?
> 10:45 < mikolaj> any objections?
> 10:48 < mikolaj> this setup has the drawback that PR authors are not warned against manually merging master into their branch if they think rebase would be sufficient (verifying and expressing/exposing the assumption this way), but I 
>                  think at least the github buttons don't automatically do non-trivial merges any more, which is even more risky
>

cabal-install/src/Distribution/Client/Setup.hs
ptkato: I agree with your stance and made the alterations.

fgaz (approved):

fgaz:
Oh oops that didn't prompt for a comment. Well:

Looks good, but I wonder if we should hide the new flags for other commands. eg. they don't make sense for `build` or `test`. On the other hand we do have a lot of other flags that don't have an effect, so is one more bad?

emilypi (approved):
Great job @ptkato! Do you mind following this PR up with another issue addressing @phadej's comments? Thanks

emilypi:
@fgaz i think @ptkato should open another issue for pruning unused commands. I'd rather see this done in one sweep rather than piecemeal, so we can go down the checklist once.

Issue #7379, opened 2021-05-03
Some commands ignore --dry-run and --only-download


fgaz:
Follow-up to #7351, in which v1-install failed outright.

Apparently some commands completely ignore --dry-run (and now --only-download), and just perform the task. Ideally they should tell what would happen (eg. `the executable "xyz" would be run`), but just not doing the thing would be ok too.

Commands that behave incorrectly:

* v2-configure (creates the project file)
* v2-freeze (creates the freeze file)
* v2-run (runs the executable)
* v2-exec (runs the command)

Commands that behave correctly:

* v2-repl
* v2-build
* v2-install
* v2-haddock
* v2-test
* v2-bench (probably. I didn't check)

aaronallen8455:
I can work on this issue.

fgaz:
That's great!

Those commands are all implemented in `cabal-install/src/Distribution/Client/Cmd*.hs` and they have similar structure. Feel free to ask here or on IRC if you get stuck or need pointers.

aaronallen8455:
Thanks, will do! I have opened https://github.com/haskell/cabal/pull/7407 for this issue.

PR #7407, opened 2021-05-25
Increase effect of --dry-run flag (#7379)

aaronallen8455:
Changes the behavior of the following commands under the --dry-run flag
- `v2-configure` and `v2-freeze` no longer write files.
- `v2-exec` and `v2-run` print out the command that would've been
executed rather than doing so.

Adds package tests for the configure and freeze commands.

---

### DIFF
--- a/cabal-install/src/Distribution/Client/CmdConfigure.hs
+++ b/cabal-install/src/Distribution/Client/CmdConfigure.hs
@@ -89,39 +89,54 @@ configureCommand = CommandUI {
 -- "Distribution.Client.ProjectOrchestration"
 --
 configureAction :: NixStyleFlags () -> [String] -> GlobalFlags -> IO ()
-configureAction flags extraArgs globalFlags = do
-  (baseCtx, projConfig) <- configureAction' flags extraArgs globalFlags
-  writeProjectLocalExtraConfig (distDirLayout baseCtx) projConfig
+configureAction flags@NixStyleFlags {..} extraArgs globalFlags = do
+    (baseCtx, projConfig) <- configureAction' flags extraArgs globalFlags
+
+    if shouldNotWriteFile baseCtx
+      then notice v "Config file not written due to flag(s)."
+      else writeProjectLocalExtraConfig (distDirLayout baseCtx) projConfig
+  where
+    v = fromFlagOrDefault normal (configVerbosity configFlags)
 
 configureAction' :: NixStyleFlags () -> [String] -> GlobalFlags -> IO (ProjectBaseContext, ProjectConfig)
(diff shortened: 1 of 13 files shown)


aaronallen8455:
Tests fail because the output I added for freeze command is not agnostic to the version of ghc used i.e. the test runner produces base ==4.13 instead of base ==4.14.
Any thoughts on how to remedy this? Is this additional output even useful?

fgaz:
> Any thoughts on how to remedy this?

Look at https://github.com/haskell/cabal/blob/master/cabal-testsuite/README.md, sections "How do I turn of accept tests? My test output wobbles to much" and "How can I manually test for a string in output?". But as you wrote

> Is this additional output even useful?

in this case it's probably better to just output "cabal.project.freeze would be written" or similar. The freeze file could be _very_ long after all, and its contents would just clutter the terminal

fgaz (approved):
Looks good!

I see that this only takes --dry-run into account, but that's fine, this is a net improvement and the --download-only check can be added later

aaronallen8455:
@fgaz Thanks for reviewing. Should the `--only-download` flag have the same effect as `--dry-run` for all of these commands?

fgaz:
Yes, since those are all steps that would be performed after the download. Do you plan to add it to this pr (just to know whether to wait for that before merging)?

aaronallen8455:
I'll add it to this PR since it's a small change.

fgaz:
Oh, I almost forgot, this needs a [changelog entry](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog) (looks good otherwise).

Finally, please rebase on current master. We're trying a github setting that requires prs to be updated on the base branch

fgaz (approved):
Thanks!

Pinging @Mikolaj @emilypi for the second review

emilypi (approved):

fgaz:
@Mergifyio backport 3.6

emilypi:
@aaronallen8455 this looks good, and I'd like to merge, but the recent v2-configure work looks like it caused conflicts. If you want, @ptkato is around to help sort those out.

aaronallen8455:
I've resolved the conflict. Perhaps someone could take a look one more time.

ptkato:
I believe there's some confusion regarding `v2-configure`, that command does not run the pre-build phase anymore, see #7411. However it might seem otherwise due to the fact that there's a `runRebuild` function amidst its code, but that's only to satisfy the typechecker, since `readProjectLocalExtraConfig` is wrapped around the `Rebuild` monad.

fgaz:
Thanks for the contribution!

Issue #7384, opened 2021-05-07
Running cabal-install unit-tests fails due to Hpc failure


fendor:
**Describe the bug**
Executing the test-suite via `cabal run unit-tests` in `cabal-install` sub-directory works fine on the first run.
After changing the test-suite, followed by executing the tests again the following error is thrown:
```
Hpc failure: module mismatch with .tix/.mix file hash number
(perhaps remove unit-tests.tix file?)
```

**To Reproduce**

```bash
$ cd cabal-install
$ cabal v2-run unit-tests -- -p "forkPackages"
$ # Modify test-suite slightly
$ cabal v2-run unit-tests -- -p "forkPackages" # Fails
```

**Expected behavior**
Should not crash

**System information**
 - `Linux archlinux 5.11.16-arch1-1 #1 SMP PREEMPT Wed, 21 Apr 2021 17:22:13 +0000 x86_64 GNU/Linux`
 - `cabal 3.4.0.0` installed via ghcup
 - `ghc 8.10.2` installed via ghcup

emilypi:
Pinging @hvr. Historically, when I brought this fact up, it was brushed off as "Not Cabal's Problem", because it is primarily due to the way HPC produces garbage and doesn't clean up after itself. I tend to agree, but I think until HPC gets its ducks in a row, Cabal should handle the cleanup and teardown when it creates coverage reports.

ezyang:
I feel it is even worse than reported here; in particular, I can induce this problem without any reference to profiling, e.g., `cabal run cabal-testsuite:cabal-tests -- --with-cabal $(cabal list-bin cabal-install:cabal) cabal-testsuite/PackageTests/Backpack/T6385/cabal.test.hs`, rebuild cabal-install, and then run it again.

Mikolaj:
Now I understand why README.md say to use `cabal install --project-file=cabal.project.release cabal-install`. But requiring all cabal devs to specify the project file on each command is definitely too much.

fendor:
Is fixed (or work-arounded by #7421 )

BinderDavid:
This is due to an (in my opinion) anti-feature of hpc: At program startup it tries to read in the .tix file and accumulates coverage information across multiple runs. This behaviour is documented in the hpc user guide:


[...]

PR #7421, opened 2021-06-01
remove coverage from main cabal.project, add to dedicated file

emilypi:
This PR addresses #7384 by deliberately keeping the project coverage details encapsulated in a dedicated `.project` file so the main build will not result in .tix file litter.

### DIFF
--- a/cabal.project.coverage
+++ b/cabal.project.coverage
@@ -28,34 +28,49 @@ allow-newer: windns-0.1.0.1:base
 
 -- avoiding extra dependencies
 constraints: rere -rere-cfg
-constraints: these -assoc
+constraints: these
 
--- So us hackers get all the assertion failures early:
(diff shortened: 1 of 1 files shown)


Mikolaj (approved):
Do we currently do coverage in CI? If so, should we now refer to the new project file via commandline options?

ezyang:
Thanks, this was bugging me!

emilypi:
@Mikolaj we don't, but if we did, this would be the correct preparatory step for that work.

cabal.project
```diff
@@ -51,35 +51,21 @@ constraints: these -assoc
 --
```
fgaz, line 51: I think this last NOTE can be removed

emilypi: Done - i separated the notes out to their constituent files.

fgaz (approved):
Thanks!

PR #7424, opened 2021-06-03
standalone tests for `cabal init`

### LINKED ISSUE #7410: `cabal init --tests`  does not create any tests if not using --lib{andexe}
**Describe the bug**

`cabal init --tests`  does not create any tests.

**To Reproduce**

```
% cabal init --tests
Warning: The package list for 'hackage.haskell.org' is 23 days old.

Guessing dependencies...

Generating LICENSE...
Generating CHANGELOG.md...
Generating app/Main.hs...
Generating example.cabal...

% tree
.
├── app
│   └── Main.hs
├── CHANGELOG.md
└── example.cabal

1 directory, 3 files
```

**Expected behavior**

A directory is created with some tests. The tests are wired into `example.cabal` as appropriate.

**System information**

```
% cabal --version
cabal-install version 3.4.0.0
% uname
Linux
```

ptkato:
This PR adds a new functionality to `cabal init`, standalone tests, as discussed in #7410.

Closes #7410.

---


ptkato:
@emilypi here we go, I do believe these changes cover everything in that issue, yes?

Another point to thing about, the `--tests` flag might have become a little too cryptic.

cabal-install/src/Distribution/Client/Init/FlagExtractors.hs
```diff
@@ -135,7 +135,11 @@ getExtraDocFiles = pure
 
 -- | Ask whether the project builds a library or executable.
 getPackageType :: Interactive m => InitFlags -> m PackageType -> m PackageType
-getPackageType flags = fromFlagOrPrompt (packageType flags)
+getPackageType InitFlags
+  { initializeTestSuite = Flag True
+  , packageType         = NoFlag
+  } _ = return TestSuite
+getPackageType flags act = fromFlagOrPrompt (packageType flags) act
```
emilypi, line 142: Nice. that's easy i suppose.

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
```diff
@@ -137,6 +137,16 @@ createProject v pkgIx srcDb initFlags = do
       return $ ProjectSettings
         (mkOpts comments cabalSpec) pkgDesc (Just libTarget)
         (Just exeTarget) testTarget
+    
+    TestSuite -> do
+      testTarget <- addLibDepToTest pkgName <$>
```
emilypi, line 142: if we have just a test suite, there'll be no lib dependency.

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
```diff
@@ -229,7 +239,7 @@ genTestTarget
     => InitFlags
     -> InstalledPackageIndex
     -> m (Maybe TestTarget)
-genTestTarget flags pkgs = initializeTestSuitePrompt flags >>= go
+genTestTarget flags pkgs = initializeTestSuitePrompt (flags {initializeTestSuite = NoFlag}) >>= go
```
emilypi, line 242: If i have `--lib --tests`, then this stomps all over the test flag.

cabal-install/src/Distribution/Client/Init/Simple.hs
```diff
@@ -62,6 +62,12 @@ createProject v _pkgIx _srcDb initFlags = do
         return $ ProjectSettings
           (mkOpts False cabalSpec) pkgDesc
           (Just libTarget) (Just exeTarget) testTarget
+      
+      TestSuite -> do
+        testTarget <- addLibDepToTest pkgName <$> genSimpleTestTarget initFlags
```
emilypi, line 67: same here - no need for lib dep.

emilypi (changes requested):
Looking good! Some comments, and I think we should golden test this.

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
ptkato: How so? Can you please elaborate? As I see it, `initializeTestSuite` must be set to `NoFlag` at that point, so it always prompt for testing, like you mentioned in https://github.com/haskell/cabal/issues/7410#issuecomment-849913926, `--tests` corresponds for a test suite package type, and `lib`/`--libandexe` will always ask for testing, hardcoding it to `NoFlag` basically does that. The only problem I see is when one passes only `--tests`, then it would ask when it should go straight through.

ptkato: I see what you mean now.

ptkato: Okay, fixed, turns out I over-complicated things and it was more straightforward than I thought.

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
```diff
@@ -139,9 +139,10 @@ createProject v pkgIx srcDb initFlags = do
         (Just exeTarget) testTarget
     
     TestSuite -> do
-      testTarget <- genTestTarget initFlags pkgIx
+      let initFlags' = initFlags { initializeTestSuite = Flag True }
```
ptkato, line 142: Just an addendum: this is necessary in `--interactive` because if both package type and test flags are **not** passed, the user will be prompted for a package type, and it now includes `TestSuite` in the list. In short, that means you could have a `TestSuite` target, with `initializeTestSuite` set to `NoFlag`, which would prompt the user if they want a test suite or not, considering that the target is `TestSuite`, that prompt behaviour is undesired.

ptkato:
Hello there @emilypi @fgaz,

I believe this one is ready for some reviewing action.

emilypi (approved):

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
Mikolaj: That would make a great comment IMHO.

cabal-install/tests/UnitTests/Distribution/Client/Init/NonInteractive.hs
```diff
@@ -323,7 +320,7 @@ driverFunctionTest pkgIx srcDb comp = testGroup "createProject"
               , "[\"test/Main.hs\", \"test/Foo.hs\", \"test/bar.y\"]"
               ]
 
-        case (_runPrompt $ createProject comp silent pkgIx srcDb (emptyFlags {initializeTestSuite = Flag True})) inputs of
+        case (_runPrompt $ createProject comp silent pkgIx srcDb (emptyFlags {initializeTestSuite = Flag True, packageType = Flag LibraryAndExecutable })) inputs of
```
Mikolaj, line 323: This is a woefully long line, it doesn't even fit in github display.

cabal-install/tests/UnitTests/Distribution/Client/Init/NonInteractive.hs
```diff
@@ -462,7 +456,7 @@ driverFunctionTest pkgIx srcDb comp = testGroup "createProject"
               , "[\"test/Main.hs\", \"test/Foo.hs\", \"test/bar.y\"]"
               ]
 
-        case (_runPrompt $ createProject comp silent pkgIx srcDb (emptyFlags {initializeTestSuite = Flag True})) inputs of
+        case (_runPrompt $ createProject comp silent pkgIx srcDb (emptyFlags {initializeTestSuite = Flag True, packageType = Flag Library})) inputs of
```
Mikolaj, line 459: Long line.

cabal-install/tests/UnitTests/Distribution/Client/Init/NonInteractive.hs
Mikolaj: If indentation/minimal diff is the problem, I'd got with 1 character indentation to keep the diff small, but break the line.

Mikolaj: But that's probably not the problem. Perhaps a `let` or `where` with the big argument would help?

Mikolaj (approved):
Looks very solid from the surface.

cabal-install/src/Distribution/Client/Init/Interactive/Command.hs
ptkato: I actually just thought of that, I shall make it so.

PR #7429, opened 2021-06-05
Allow relative paths starting with single-character directories

### LINKED ISSUE #7426: isGoodRelativeFilePath rejects single-character top-level directories
I had previously been storing my C source files in a directory called `c` alongside my `.cabal` file.  When I tried to upload the package to Hackage, I got a weird error about a "same directory segment: .." in the path `c/cdio/compat/cdtext.c`; I believe I have narrowed it down to [this line](https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2197), which I think should be switching to `state5` instead of `state3` (note also that the message `state3` is giving on line 2216 has an extra dot); as it is, a directory consisting of a single letter fails if and only if it's at the very beginning of a relative path.  Was this deliberate, whether for OS/tooling compatibility or for reserving names (e.g. the dist-newstyle `x/` prefix for executable targets)?  If so, the error message should probably be edited, which might involve creating a new `state6`; if not, then it should be pointed to the right place.

ag-eitilt:
Should fix #7426.

---

Leaving testing to the CI, but the existing tests didn't seem to catch the bug, and going digging to figure out the test format to add one that would seemed like it would take longer than it did to find, diagnose, and fix in the first place.

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Check.hs
+++ b/Cabal/src/Distribution/PackageDescription/Check.hs
@@ -2146,11 +2146,17 @@ fileExtensionSupportedLanguage path =
     isHaskell = extension `elem` [".hs", ".lhs"]
     isC       = isJust (filenameCDialect extension)
 
--- | Whether a path is a good relative path.
+-- | Whether a path is a good relative path.  We aren't worried about perfect
+-- cross-platform compatibility here; this function just checks the paths in
+-- the (local) @.cabal@ file, while only Hackage needs the portability.
 --
 -- >>> let test fp = putStrLn $ show (isGoodRelativeDirectoryPath fp) ++ "; " ++ show (isGoodRelativeFilePath fp)
 --
--- >>> test "foo/bar/quu"
+-- Note that "foo./bar.hs" would be invalid on Windows.
+--
+-- >>> traverse_ test ["foo/bar/quu", "a/b.hs", "foo./bar.hs"]
+-- Nothing; Nothing
+-- Nothing; Nothing
 -- Nothing; Nothing
 --
 -- Trailing slash is not allowed for files, for directories it is ok.
@@ -2172,7 +2178,7 @@ fileExtensionSupportedLanguage path =
 -- Just "posix absolute path"; Just "posix absolute path"
 -- Just "empty path segment"; Just "empty path segment"
 -- Just "trailing same directory segment: ."; Just "trailing same directory segment: ."
--- Just "same directory segment: ."; Just "same directory segment: .."
+-- Just "same directory segment: ."; Just "same directory segment: ."
 -- Just "parent directory segment: .."; Just "parent directory segment: .."
 -- Just "reserved character '*'"; Just "reserved character '*'"
 --
@@ -2194,16 +2200,16 @@ isGoodRelativeFilePath = state0
     state0 (c:cs) | c == '.'     = state1 cs
                   | c == '/'     = Just "posix absolute path"
                   | isReserved c = Just ("reserved character " ++ show c)
-                  | otherwise    = state3 cs
+                  | otherwise    = state5 cs
 
-    -- after .
+    -- after initial .
     state1 []                    = Just "trailing dot segment"
     state1 (c:cs) | c == '.'     = state4 cs
                   | c == '/'     = state2 cs
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state5 cs
 
-    -- after ./
+    -- after ./ or after / between segments
     state2 []                    = Just "trailing slash"
     state2 (c:cs) | c == '.'     = state3 cs
                   | c == '/'     = Just "empty path segment"
@@ -2213,11 +2219,11 @@ isGoodRelativeFilePath = state0
     -- after non-first segment's .
     state3 []                    = Just "trailing same directory segment: ."
     state3 (c:cs) | c == '.'     = state4 cs
-                  | c == '/'     = Just "same directory segment: .."
+                  | c == '/'     = Just "same directory segment: ."
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state5 cs
 
-    -- after non-first segment's ..
+    -- after ..
     state4 []                    = Just "trailing parent directory segment: .."
     state4 (c:cs) | c == '.'     = state5 cs
                   | c == '/'     = Just "parent directory segment: .."
@@ -2226,7 +2232,7 @@ isGoodRelativeFilePath = state0
 
     -- in a segment which is ok.
     state5 []                    = Nothing
-    state5 (c:cs) | c == '.'     = state3 cs
+    state5 (c:cs) | c == '.'     = state5 cs
                   | c == '/'     = state2 cs
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state5 cs
@@ -2255,8 +2261,8 @@ isGoodRelativeDirectoryPath = state0
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state4 cs
 
-    -- after ./
-    state1 []                    = Nothing -- "./"
+    -- after initial ./ or after / between segments
+    state1 []                    = Nothing
     state1 (c:cs) | c == '.'     = state2 cs
                   | c == '/'     = Just "empty path segment"
                   | isReserved c = Just ("reserved character " ++ show c)
@@ -2269,8 +2275,8 @@ isGoodRelativeDirectoryPath = state0
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state4 cs
 
-    -- after non-first segment's ..
-    state3 []                    = Just "trailing parent directory segment: ."
+    -- after ..
+    state3 []                    = Just "trailing parent directory segment: .."
     state3 (c:cs) | c == '.'     = state4 cs
                   | c == '/'     = Just "parent directory segment: .."
                   | isReserved c = Just ("reserved character " ++ show c)
@@ -2283,7 +2289,7 @@ isGoodRelativeDirectoryPath = state0
                   | isReserved c = Just ("reserved character " ++ show c)
                   | otherwise    = state4 cs
 
-    -- after .
+    -- after initial .
     state5 []                    = Nothing -- "."
     state5 (c:cs) | c == '.'     = state3 cs
                   | c == '/'     = state1 cs
--- a/changelog.d/rel-path-check
+++ b/changelog.d/rel-path-check
@@ -1,0 +1,4 @@
+synopsis: Allow relative paths starting with single-charater directories
+packages: Cabal
+prs: #7429
+issues: #7426


Mikolaj:
@jneira: would you live to review? In particular, is there anything else wrong in this code, even comments?

Mikolaj:
@ag-eitilt: thank you for the PR. Are these doctests in the comment above this function? Are they executed? If so, could you add there a test that fails with old code, but works with your fix?

Mikolaj:
This

https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2229

and this

https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2353

don't match. Which one is right?

Mikolaj:
And indeed these are doctest and they are run, which is why CI failed. I guess the failure is due to the lesser of the two fixes, so a test for the main problem is not yet present.

jneira:
I think a regression test would be great: it provides evidence thah the patch works and help to not break it again.
I guess the place to add the test would be here: https://github.com/haskell/cabal/blob/master/Cabal-tests/tests/CheckTests.hs
Maybe [this commit](https://github.com/haskell/cabal/commit/ebe279979d1f0430b4005adcc7031ffb331d72c8) could help to write it  as example.

phadej:
> I think a regression test would be great: it provides evidence thah the patch works and help to not break it again.
> I guess the place to add the test would be here: https://github.com/haskell/cabal/blob/master/Cabal-tests/tests/CheckTests.hs
> Maybe [this commit](https://github.com/haskell/cabal/commit/ebe279979d1f0430b4005adcc7031ffb331d72c8) could help to write it as example.

Don't make that too difficult. `make doctest` works (or rather `cabal build Cabal && make doctest`). Alternatively the *current maintainers* could transfer the tests from doctests into the proper test suite if they find that desirable. (IMO it was simpler to do parsing related testing using `doctest`, there are plenty of doctests for `Parsec` and `Pretty` instances).

ag-eitilt:
I've added a regression test to the doctests (didn't see them for the first commit) and fixed the doctest which was expecting the typoed error; let's see if the CI likes it now.  I can certainly add something to `CheckTests.hs` as well, later, but I have limited time at the moment.

> This
> 
> https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2229
> 
> and this
> 
> https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2353
> 
> don't match. Which one is right?

Looking at the `"foo/./bar"` below, I'm guessing the first.  I've updated the second to match, but anyone let me know if that's wrong.

https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/PackageDescription/Check.hs#L2170-L2177

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -2350,7 +2351,7 @@ isGoodRelativeDirectoryPath = state0
 --     | x <= COtherReserved -> 6
 --     | otherwise           -> 5
 -- 5+ -> \x -> if
---     | x <= CDot           -> 5
+--     | x <= CDot           -> 3
```
phadej, line 2354: this is not correct. Or if you think it should be that way, then above regular expression construction is wrong (which I doubt). So most likely the Haskell `state5` implementation is also wrong.

```haskell
*Distribution.PackageDescription.Check> isGoodRelativeFilePath "foo./bar"
Just "same directory segment: .."
```

---

You can put the code above in the comment into Path.hs and

```
cabal-env kleene lattices
ghci Path.hs
*Main> Kleene.Internal.Pretty.putPretty $ fromTM filePathR
```

phadej (changes requested):

Cabal/src/Distribution/PackageDescription/Check.hs
ag-eitilt: Good catch, that's what I get for not diagramming things out.  State 3 is already reached via state 2, so this particular fix was exactly backwards.

ag-eitilt:
@phadej's mention of "foo./bar" (and [that link](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file) above 'isRelative') reminded me that Windows rejects any names ending in `'.'` or a space.  Do y'all want to reject those here as well?  It would require a `state6` for this and `isGoodRelativeDirectoryPath`, and would change the patch from strictly loosening the requirements to being a breaking change (in theory, though there's probably very few packages which would be affected).

Mikolaj:
CI is green, great job!

Re "I can certainly add something to CheckTests.hs as well" I think we agreed there is no need --- doctests work great in this case.

Re Windows silliness, I think it's a good idea to ban the offending cases. It's nasty if packages on hackages don't work for Windows users for reasons that a dev on Linux can't even imagine possible and tooling gives no hint.

I think this would require a separate PR with a user changelog (OTOH, I'm not sure this PR needs one, because this one is a bugfix and I hope no users depend on the bug or have elaborate workarounds they may now dismantle) and clearly explaining in the error message that it's a Windows limitation and providing a link (I can't quite find an explanation at the link you provided). If you'd be so kind as to create such a PR, please take your time, cabal has long release spans, so this won't be out tomorrow anyway and we strive not to burn out contributors (too soon) in this project. :)

Mikolaj (approved):
LGTM, though I depend on @phadej to give the final verdict.

Mikolaj:
Pasting an IRC chat, since no further discussion ensued:

> 08:38 < mikolaj_> sclv: apparently Windows bans spaces or dots at the end of filenames (or something equally silly) and we'd like to ban such paths in .cabal files, etc.
> 08:39 < mikolaj_> who needs to know and approve this API change and potential breakage?
> 08:40 < mikolaj_> and I guess it would be nice to crawl Hackage and see if any package is affected (which means it currently doesn't build on Windows, but many do for other reasons)
> 08:43 < sclv> uhhh… if we ban em in hackage but not in general we can keep it to a small group. like core maintainers can basically dictate policy with a little consultation and that’s that
> 08:43 < sclv> if we ban em in cabal files as such thats a change to the cabal grammar and its a much vaguer and broader group one needs signoff from?
> 08:44 < sclv> hackage already has more restrictive policies on cabal files than just the grammar
> 08:46 < sclv> i guess we need to ban it not just in cabal files but also tarballs
> 08:46 < sclv> or not, since sdist only packages whats in the cabal file anyway
> 08:47 < sclv> to scan hackage one need just grep with the right regex thru the index-01.tar thats in .cabal already, since that has all cabal files inside
> 08:57 < mikolaj_> oh, cool
> 09:00 < mikolaj_> if nothing is affected on Hackage right now, I think we can assume it's effectively a non-breaking change, despite being formally breaking
> 09:00 < mikolaj_> hmm, though some people use cabal for their closed-source packages
> 09:11 < mikolaj_> so we need a RFC ticket and some publicity

Mikolaj:
Continued:

> 16:35 < sclv> mikolaj_: note again that we can simply have an extended check in the cabal lib that is _not_ an error by default, but turn it into an error on package upload to the central cabal server
> 16:36 < sclv> er hackage server
> 16:38 < sclv> cf https://github.com/haskell/hackage-server/blob/master/Distribution/Server/Packages/Unpack.hs#L142 and surrounding
> 17:14 < mikolaj_> sounds good; after some years we may turn the warning into error or not
> 17:14 < mikolaj_> the only problem is it complicates the implementation (right now we only have path format errors, never warnings)
> 17:15 < mikolaj_> let's see how bad it gets...

Mikolaj:
The story unfolds:

> 18:26 < sclv> mikolaj_: actually we may already have logic basically in place! https://github.com/haskell/hackage-server/blob/master/Distribution/Server/Packages/Unpack.hs#L448
> 18:26 < sclv> basically if a tar file fails this check, hackage will reject it: https://hackage.haskell.org/package/tar-0.5.1.1/docs/Codec-Archive-Tar-Check.html#v:checkPortability
> 18:27 < sclv> so if that function doesn't actually catch all portability errors, it should be fixed anyway, for completeness
> 20:38 < mikolaj_> ok, if that rejects the trailing space and dot (should be checked), then we don't need to bother and make the internal cabal check complete wrt Windows quirks
> 20:39 < mikolaj_> people would be able to write Windows-incorrect paths in variuos cabal config files
> 20:39 < mikolaj_> but at least they would get an error already when uploading to Hackage instead of their users getting errors when trying to build
> 20:39 < mikolaj_> that's good enough
> 20:40 < mikolaj_> we don't have unlimited dev resources
> 20:41 < mikolaj_> perhaps a comment in the cabal-side check would be valuable, explaining this, so that we don't need this discussion next time somebody notices that cabal accepts Windows-unsafe paths
> 20:41 < sclv> anyway its worth somebody verifying what that function does or doesn't accept to be sure that things are working already
> 20:42 < sclv> the general rule of thumb is cabal files are _not_ required to be portable (since after all many packages are not) but uploads to hackage _are_
> 20:42 < mikolaj_> agreed
> 20:42 < mikolaj_> let's check; the upside being we don't need to crawl hackage, but only check the code in tar-0.5.1.1

phadej:
- Changes in `PackageDescription/Check.hs` **are not** changes to `.cabal` format, only to the `cabal check` (and transitively to hackage-server on upload).
    - which means that these changes can be done without worrying about backward-compatibility
- `cabal run hackage-tests -- --help` and look into the source what and how it does, specifically you can check how much new checks cause new `check` "errors" (i.e. warnings).

For the record, at the moment:

```
149498 files processed
41502 files have lexer/parser warnings
116492 files have check warnings
334 build impossible
12073 build warning
50671 build dist suspicious
35480 build dist suspicious warning
17934 build dist inexcusable
```

i.e. most of cabal files on Hackage are dubious :)

Mikolaj:
> i.e. most of cabal files on Hackage are dubious :)

Wow. Presumably a lot of those are ancient abandonware? I wonder how many build with newer GHCs on matrix.

Thank you for correcting my assumption these checks are for cabal config files. If they only affect `cabal check` then one more reason not to worry excluding the Windows-incorrect paths --- its probably (to be checked) already excluded in tar-0.5.1.1, and if not, we argued on IRC we should fix it there.

However, if even the data we check is the same, how about (exposing, if needed, and) calling the paths check from tar-0.5.1.1 that is used here https://github.com/haskell/hackage-server/blob/master/Distribution/Server/Packages/Unpack.hs#L448 instead of our maintaining own path correctness checks?

phadej:
> instead of our maintaining own path correctness checks?

These checks are something which we could (and IMO should) take in use in some future `cabal-version: ...`. This all started with `SymbolicPath` work, i.e. changing https://github.com/haskell/cabal/blob/fc4daed11271630aae1954e1520f9c16be2f4cd3/Cabal/src/Distribution/Utils/Path.hs#L62-
and https://github.com/haskell/cabal/blob/fc4daed11271630aae1954e1520f9c16be2f4cd3/Cabal/src/Distribution/PackageDescription/FieldGrammar.hs#L712-L766

Mikolaj:
OK, so I was at least right this *should* be used for paths in various config files. Then we are in the ideal world, because it's not yet used, so without endangering backward compatibility, we can make it stricter (I'm again flipping towards strengthening the check to include Windows-correctness). Assuming the checks in tar-0.5.1.1 already have the checks (otherwise we do have to worry about backward compatibility).

Still, why not re-use the code from tar-0.5.1.1, improving it, if not perfect?

phadej:
Because `.cabal` format should not depend on external sources, neither `Cabal` can depend on `tar`.

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -2150,7 +2150,8 @@ fileExtensionSupportedLanguage path =
 --
 -- >>> let test fp = putStrLn $ show (isGoodRelativeDirectoryPath fp) ++ "; " ++ show (isGoodRelativeFilePath fp)
 --
--- >>> test "foo/bar/quu"
+-- >>> traverse_ test ["foo/bar/quu", "a/b.hs"]
```
phadej, line 2153: please add the `foo./bar.hs` example as well

phadej (changes requested):

Cabal/src/Distribution/PackageDescription/Check.hs
ag-eitilt: Sounds good!  Was just waiting to see what the verdict was on whether it should throw an error or not; sounds like "not".

Mikolaj:
@phadej: would you have a cursory look (you can ignore the comments --- I hijacked this issue for admin discussion:)?

phadej (approved):

Mikolaj:
@ag-eitilt: Congratulations on your merged contribution. What would you like to do with the related considerations discussed in this issue? IIRC, e.g., verifying how consistent these checks are with the Hackage checks and verifying if the Windows paths restrictions apply to Hackage? Would you like to manage these things, e.g., open issues about them (not code tickets, but brainstorm/technical task issues, I suppose)?

ag-eitilt:
Thanks!  I'm too busy to own them for the next couple weeks, but I could likely do so by early july.  If anyone else is interested before I get to them, feel free to take over, but otherwise I'll open the relevant issues once I get time.

Mikolaj:
Deal. Thank you.

ag-eitilt:
I think those three issues/PRs handle the remaining questions.  Have I missed anything?

Mikolaj:
@ag-eitilt: AFAICT, this is it. Particularly interesting catch in the third issue. Thank you again and let's move the discussion to these issues.

PR #7439, opened 2021-06-12
Prefer ProjectFlags over additional field in OutdatedFlags

### LINKED ISSUE #6815: Use ProjectFlags in outdated command

fendor:
closes #6815 

---

I tested it manually that it uses the correct freeze file.

### DIFF
--- a/cabal-install/src/Distribution/Client/CmdOutdated.hs
+++ b/cabal-install/src/Distribution/Client/CmdOutdated.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE NamedFieldPuns #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Distribution.Client.CmdOutdated
@@ -16,19 +17,24 @@ module Distribution.Client.CmdOutdated
 where
 
 import Distribution.Client.Compat.Prelude
+import Distribution.Compat.Lens
+    ( _1, _2 )
 import Prelude ()
 
 import Distribution.Client.Config
     ( SavedConfig(savedGlobalFlags, savedConfigureFlags
                  , savedConfigureExFlags) )
 import Distribution.Client.IndexUtils as IndexUtils
+import Distribution.Client.DistDirLayout
+    ( defaultDistDirLayout
+    , DistDirLayout(distProjectRootDirectory, distProjectFile) )
 import Distribution.Client.ProjectConfig
     ( ProjectConfig(projectConfigShared),
       ProjectConfigShared(projectConfigConstraints), findProjectRoot,
       readProjectLocalFreezeConfig )
-import Distribution.Client.DistDirLayout
-    ( defaultDistDirLayout
-    , DistDirLayout(distProjectRootDirectory, distProjectFile) )
+import Distribution.Client.ProjectFlags
+    ( projectFlagsOptions, ProjectFlags(..), defaultProjectFlags
+    , removeIgnoreProjectOption )
 import Distribution.Client.RebuildMonad
     ( runRebuild )
 import Distribution.Client.Sandbox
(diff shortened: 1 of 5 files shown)


cabal-install/src/Distribution/Client/CmdOutdated.hs
```diff
@@ -212,8 +214,8 @@ outdatedOptions _showOrParseArgs =
 -------------------------------------------------------------------------------
 
 -- | Entry point for the 'outdated' command.
-outdatedAction :: OutdatedFlags -> [String] -> GlobalFlags -> IO ()
-outdatedAction OutdatedFlags{..} _targetStrings globalFlags = do
+outdatedAction :: (ProjectFlags, OutdatedFlags) -> [String] -> GlobalFlags -> IO ()
+outdatedAction (ProjectFlags{..}, OutdatedFlags{..}) _targetStrings globalFlags = do
```
Mikolaj, line 218: Is a lot the fields from `ProjectFlags` used there now or will be in the future (I guess that's the plan for the future and this is the crux of the PR?)? If it's only one field and no future plans, could you list it instead of the two dots? E.g., this would help me understand this during the review, so possibly it would help others reading the code.

Mikolaj (approved):
LGTM

cabal-install/src/Distribution/Client/CmdOutdated.hs
fendor: `ProjectFlags` has only two fields, I don't feel like this is significant. Afaict, this should not have a single addition to the help message, at most the role of `--project-file` is a bit harder to grasp.

I think you are rather thinking about `NixStyleFlags` which has a ton of fields.

Mikolaj: Got it. How about `ProjectFlags{theFieldUsedHere}`?

fendor: Done!

Mikolaj: Thank you. Now I see without asking.

emilypi (approved):

Mikolaj:
Windows CI seems to be borked. Let's hope only that one.

cabal-install/src/Distribution/Client/CmdOutdated.hs
```diff
@@ -161,10 +168,6 @@ outdatedOptions _showOrParseArgs =
       "Act on the new-style freeze file (default: cabal.project.freeze)"
       outdatedNewFreezeFile (\v flags -> flags {outdatedNewFreezeFile = v})
       trueArg
-  , option [] ["project-file"]
-      "Act on the new-style freeze file named PROJECTFILE.freeze rather than the default cabal.project.freeze"
```
fgaz, line 165: This will change the description to a slightly inaccurate "Set the name of the cabal.project file to search for in parent directories", no? This is particularly tricky for the user because the flag only sets the PROJECTFILE part of PROJECTFILE.freeze

fgaz: Hmm, though since this isn't the `freeze` command, one would expect --project-file to affect both PROJECTFILE and PROJECTFILE.freeze like the other commands (?). Is that what this flag does / will do?

fendor: Yes it does!

fendor: I agree, this way the flag is slightly less discoverable

fgaz: > Yes it does!

then it's accurate, though maybe we should improve the help text for --project-file, like in https://cabal.readthedocs.io/en/3.4/cabal-project.html#cmdoption-0

fgaz (approved):

Issue #701, opened 2012-05-24
Track license-file in ghc package database.


bos:
(Imported from [Trac #710](http://hackage.haskell.org/trac/hackage/ticket/710), reported by elliottt on 2010-07-12)

ghc-pkg has supported (since 6.4) a license-file field in the package database.  It would be useful to make this information available inside of Cabal, as guesses have to be made as to the location of the license file, currently.

PR #7443, opened 2021-06-15
[ re #7173 ] Add the license information more in Paths_*pkgname*

### LINKED ISSUE #7173: Feature request: provide text of LICENSE file as String/Text in generated Paths_XYZ
The cabal-generated `Paths_XYZ` gives me all kind of useful values from the installation, but not the text of the `LICENSE` file.

Feature: provide a def `license :: Text` in `Paths_XYZ` containing the text of `LICENSE`.

My current workaround is: Put a `Makefile` in the `src` directory that creates `License.hs`:
```make
License.hs : ../LICENSE Makefile
	echo "\
-- This file was autogenerated from LICENSE, do not edit!!\n\n\
{-# LANGUAGE QuasiQuotes #-}\n\n\
module License where\n\n\
import Data.String.QQ\n\n\
license :: String\n\
license = [s|" > $@
	cat $< >> $@
	echo "\n|]" >> $@
```
Then I add `License` to the module list in the cabal file and can `import License` in my executable.

However, this boilerplate would not be necessary if `license` was declared in `Paths_XYZ`.

Related feature request: #4887, but mine is more modest and has a straightforward semantics.

L-TChen:
I am trying to resolve #7173 by ~~embedding~~ adding the license information to the module `Paths_<pkgname>`. ~~There are a few design issues I am not sure about.~~

* ~~Should the license type be stored as `String` or as `License` type? In the current draft, `license :: String` is pretty printed using `Distribution.Pretty`, i.e. `Z.zLicense = show $ pretty $ license pkg_descr` in `Cabal/src/Distribution/Simple/Build/PathsModule.hs`. I don't know if there is any possible application requiring `License` instead of `String`.~~
  - ~~Making `license` of type `License` means that `Distribution.License` needs to be imported and thus `Paths_<pkgname>` depends on the package `cabal`. That is essentially the reason I am against using `License`.~~
  - ~~However, having a separate module like `License_<pkgname>` may alleviate the need for an extra dependency.~~

* ~~Should it be a separate module instead? I plan to embed license files using Template Haskell, but I have the impression that invoking TH slows down the compilation in general.~~
---
Update (Jun. 15): 

Anything fancier than `String` results in an additional dependency anyway, including `template-haskell`. I plan to add the following in `Paths_<pkgname>`:

- `license :: String` 
- `licenseFiles :: [FilePath]`

where `license` and `licenseFiles` correspond to the fields `license` and `license-file` in a `.cabal` file. Since license files are not copied into the destination folder (please correct me if it is not the case), a sensible way to access its content is via Template Haskell to embed its content.
 
---


Mikolaj:
Thank you for the PR. I've just read the issue and one question seems unanswered: what to do when there are multiple licenses. What's your preference there?

phadej:
- At the bare minimum this must be guarded by `cabal-version`
- Filepaths is a very bad idea. There are plenty good reasons to not hard code (installation) paths into libraries, don't make it harder to make it happen (at the every least the path getting function should be in `IO` c.f. `getDataDir`). Note: original issue is about embedding textual content of them.

L-TChen:
Alright, thanks for your reply. I was not so sure how to do it, but I think I can do more than just giving file paths.

> What to do when there are multiple licenses. What's your preference there?

I will have `[String]` instead.

> Filepaths is a very bad idea. 

You are right. I forgot that I can simply fetch the license file during generating the module `Paths_<pkgname>`.

> the dependency closure isn't so static nor easy to accumulate

The dependency graph will be available after `cabal` resolves the dependencies, right? Is there any way to fetch that information?

L-TChen:
By the way, one of the design choices I have in mind is to avoid any additional dependency. So I avoid using `Text` or `[Text]`. Otherwise, it will be a breaking change.

wenkokke:
Is there a reason to package the license file into the executable itself? I assumed it'd be sufficient to distribute the executable together with the licenses.

wenkokke:
@L-TChen What still needs to happen before this can be merged?

L-TChen:
Sorry for not getting back to you sooner. IIRC, this PR was for some discussion first to figure out the appropriate way to add license information to the module `Paths_<pkgname>`, and I only tried to locate what files to modify without much progress.

I plan to work on this PR this summer.

L-TChen:
I don’t find a way to get the text of the license file of every dependent package when generating the module `Paths_<pkgname>`. License files are usually placed in the cabal store, but their location does not seem available in the existing records. For installed packages, the type of license is already in the `InstalledPackageInfo` but the license file might be lost.

My current understanding is that it might be worthwhile to keep the license file as a field of type `[Text]` in `InstalledPackageInfo`, so it can be easily accessed from the field `installedPkgs :: InstalledPackageIndex` of `LocalBuildInfo`.

Any hint?

phadej:
> My current understanding is that it might be worthwhile to keep the license file as a field of type 

Not a great idea.

My cabal store has 12000 packages currently. Reading its package database (i.e. 12000 `InstalledPackageInfo`s) is already quite slow. Traversing through license text would slow it even further.

If you need a file location(s), add a field for field locations.

L-TChen:
I hit the issue #701 which says that `ghc-pkg` is supposed to address `license-file` but it does not. 
IIUC, `InstalledPackageInfo` is exactly from `Cabal-syntax`, so I only need to patch `Cabal-syntax` and re-compile `ghc-pkg`, then the toolchain should fix the above issue and I will have sufficient information to locate every license file of every dependency.

Does it make sense?

L-TChen:
I hit the issue #701. I am patching the record type `InstalledPackage` with a new field `licenseFiles :: [FilePath]`. 

IIUC, I will need to re-compile the `ghc-pkg` to get sync with the new field and I will also need to file a PR to update `ghc-pkg` with the patched `Cabal-syntax`. Is that right?

Issue #7270, opened 2021-02-03
Dependency on installed public sublibrary without --dependency flag


fgaz:
follow-up from https://github.com/haskell/cabal/issues/5660#issuecomment-772485875

Depending on a globally installed public sublibrary is problematic unless  `--dependency=pkgname:sublibname=installed-id` is specified.

The bug (?) does not get triggered when that flag is used, for example:

* when manually running Setup configure --dependency
* when running cabal-install _on source packages_

Here are two occasions in which it can manifest (`b` depends on `a:foo`):

In nix, which I think does not use --dependency and instead relies on Cabal's basic solver, I get this error:

<details><summary>build log</summary>
<p>

```
building '/nix/store/djsi6bxhgwgqlvymi3rf5078567jry3w-b-1.drv'...
setupCompilerEnvironmentPhase
Build with /nix/store/2lkbmrf9iw3v4dnf0yixhldvff5397xl-ghc-8.10.3.
unpacking sources
unpacking source archive /nix/store/0h1n08ri9fph049mx0rxxx8qnl4pz408-b
source root is b
patching sources
compileBuildDriverPhase
setupCompileFlags: -package-db=/build/setup-package.conf.d -j4 +RTS -A64M -RTS -threaded -rtsopts
[1 of 1] Compiling Main             ( /nix/store/4mdp8nhyfddh7bllbi7xszz7k9955n79-Setup.hs, /build/Main.o )
Linking Setup ...
configuring
[...]

PR #7470, opened 2021-06-30
Suppress warnings for verbosity

### LINKED ISSUE #7286: Please give some way to disable specific cabal warnings...
... or at least suppress the : in package name and phantom internal version bound ones. 

**Describe the bug**
cabal 3.4 makes it possible (nay, necessary) to use : in some package names.

However, as of 3.4.0.0: this results in a veritable flood of

```
Warning: unlifted.cabal:55:27: colon specifier is experimental feature (issue
#5660)
```

notifications.

Similarly, there appears to be phantom internally made up constraints on any internal library you build:

```
Warning: The package has an extraneous version range for a dependency on an
internal library: unlifted:{core, def, reps} ((>=0 && ==0) && ==0) && ==0,
unlifted:core ((>=0 && ==0) && ==0) && ==0. This version range includes the
current package but isn't needed as the current package's library will always
be used.
```

There's no way to suppress these either. In my current project I get something like 100k worth of build spam output... from just these two warnings alone.

**To Reproduce**

```
git clone https://github.com/ekmett/unlifted.git
cd unlifted
git checkout 09b4b1d037fff2
cabal build
```

with ghc 9.0.1 and cabal 3.4.0.0-rc4 (or rc7)

**Expected behavior**
I'd like to see the actual build messages and real errors.

**System information**
- OSX
- cabal 3.4.0.0-rc4, cabal 3.4.0.0-rc7, cabal 3.4.0.0
- ghc 9.0.1

**Additional information**
I expect to increase the amount of backpack usage in the project 10-fold, which will result in about a megabyte of logs spewed to my screen in the current configuration.

[updated with cabal 3.4.0.0 final release, which didn't change the situation.]

ptkato:
This PR will add the ability to suppress all warnings within a given command. For example, to suppress all warnings in `cabal build`, while keeping the output verbose: `cabal build --verbose="verbose+nowarn"`.

This PR serves as a palliative measure for #7286, while also doubling up on adding a new feature.

---

* [x] The documentation has been updated, if necessary

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Check.hs
+++ b/Cabal/src/Distribution/PackageDescription/Check.hs
@@ -539,13 +539,15 @@ checkFields pkg =
         ++ "for example 'tested-with: GHC==6.10.4, GHC==6.12.3' and not "
         ++ "'tested-with: GHC==6.10.4 && ==6.12.3'."
 
-  , check (not (null depInternalLibraryWithExtraVersion)) $
-      PackageBuildWarning $
-           "The package has an extraneous version range for a dependency on an "
-        ++ "internal library: "
-        ++ commaSep (map prettyShow depInternalLibraryWithExtraVersion)
-        ++ ". This version range includes the current package but isn't needed "
-        ++ "as the current package's library will always be used."
+  -- for more details on why the following was commented out,
+  -- check https://github.com/haskell/cabal/pull/7470#issuecomment-875878507
+  -- , check (not (null depInternalLibraryWithExtraVersion)) $
+  --     PackageBuildWarning $
+  --          "The package has an extraneous version range for a dependency on an "
+  --       ++ "internal library: "
+  --       ++ commaSep (map prettyShow depInternalLibraryWithExtraVersion)
+  --       ++ ". This version range includes the current package but isn't needed "
+  --       ++ "as the current package's library will always be used."
 
   , check (not (null depInternalLibraryWithImpossibleVersion)) $
       PackageBuildImpossible $
@@ -555,13 +557,13 @@ checkFields pkg =
         ++ ". This version range does not include the current package, and must "
         ++ "be removed as the current package's library will always be used."
 
-  , check (not (null depInternalExecutableWithExtraVersion)) $
-      PackageBuildWarning $
-           "The package has an extraneous version range for a dependency on an "
-        ++ "internal executable: "
-        ++ commaSep (map prettyShow depInternalExecutableWithExtraVersion)
-        ++ ". This version range includes the current package but isn't needed "
-        ++ "as the current package's executable will always be used."
+  -- , check (not (null depInternalExecutableWithExtraVersion)) $
+  --     PackageBuildWarning $
+  --          "The package has an extraneous version range for a dependency on an "
+  --       ++ "internal executable: "
+  --       ++ commaSep (map prettyShow depInternalExecutableWithExtraVersion)
+  --       ++ ". This version range includes the current package but isn't needed "
+  --       ++ "as the current package's executable will always be used."
 
   , check (not (null depInternalExecutableWithImpossibleVersion)) $
       PackageBuildImpossible $
@@ -617,12 +619,12 @@ checkFields pkg =
       , isInternal pkg dep
       ]
 
-    depInternalLibraryWithExtraVersion =
-      [ dep
-      | dep@(Dependency _ versionRange _) <- internalLibDeps
-      , not $ isAnyVersion versionRange
-      , packageVersion pkg `withinRange` versionRange
-      ]
+    -- depInternalLibraryWithExtraVersion =
+    --   [ dep
+    --   | dep@(Dependency _ versionRange _) <- internalLibDeps
+    --   , not $ isAnyVersion versionRange
+    --   , packageVersion pkg `withinRange` versionRange
+    --   ]
 
     depInternalLibraryWithImpossibleVersion =
       [ dep
@@ -630,12 +632,12 @@ checkFields pkg =
       , not $ packageVersion pkg `withinRange` versionRange
       ]
 
-    depInternalExecutableWithExtraVersion =
-      [ dep
-      | dep@(ExeDependency _ _ versionRange) <- internalExeDeps
-      , not $ isAnyVersion versionRange
-      , packageVersion pkg `withinRange` versionRange
-      ]
+    -- depInternalExecutableWithExtraVersion =
+    --   [ dep
+    --   | dep@(ExeDependency _ _ versionRange) <- internalExeDeps
+    --   , not $ isAnyVersion versionRange
+    --   , packageVersion pkg `withinRange` versionRange
+    --   ]
 
     depInternalExecutableWithImpossibleVersion =
       [ dep
--- a/Cabal/src/Distribution/Simple/Utils.hs
+++ b/Cabal/src/Distribution/Simple/Utils.hs
@@ -442,7 +442,7 @@ verbosityHandle verbosity
 --
 warn :: Verbosity -> String -> IO ()
 warn verbosity msg = withFrozenCallStack $ do
-  when (verbosity >= normal) $ do
+  when ((verbosity >= normal) && not (isVerboseNoWarn verbosity)) $ do
     ts <- getPOSIXTime
     hFlush stdout
     hPutStr stderr . withMetadata ts NormalMark FlagTrace verbosity
--- a/Cabal/src/Distribution/Verbosity.hs
+++ b/Cabal/src/Distribution/Verbosity.hs
@@ -52,6 +52,9 @@ module Distribution.Verbosity (
   -- * Stderr
   verboseStderr, isVerboseStderr,
   verboseNoStderr,
+
+  -- * No warnings
+  verboseNoWarn, isVerboseNoWarn
   ) where
 
 import Prelude ()
@@ -213,6 +216,7 @@ parsecVerbosity = parseIntVerbosity <|> parseStringVerbosity
             "timestamp"  -> return verboseTimestamp
             "stderr"     -> return verboseStderr
             "stdout"     -> return verboseNoStderr
+            "nowarn"     -> return verboseNoWarn
             _            -> P.unexpected $ "Bad verbosity flag: " ++ token
 
 flagToVerbosity :: ReadE Verbosity
@@ -239,6 +243,7 @@ showForCabal v
     showFlag VMarkOutput = ["+markoutput"]
     showFlag VTimestamp  = ["+timestamp"]
     showFlag VStderr     = ["+stderr"]
+    showFlag VNoWarn     = ["+nowarn"]
 
 showForGHC :: Verbosity -> String
 showForGHC   v = maybe (error "unknown verbosity") show $
@@ -290,6 +295,10 @@ verboseStderr = verboseFlag VStderr
 verboseNoStderr :: Verbosity -> Verbosity
 verboseNoStderr = verboseNoFlag VStderr
 
+-- | Turn off warnings for log messages
+verboseNoWarn :: Verbosity -> Verbosity
+verboseNoWarn = verboseFlag VNoWarn
+
 -- | Helper function for flag enabling functions
 verboseFlag :: VerbosityFlag -> (Verbosity -> Verbosity)
 verboseFlag flag v = v { vFlags = Set.insert flag (vFlags v) }
@@ -334,6 +343,10 @@ isVerboseTimestamp = isVerboseFlag VTimestamp
 -- @since 3.4.0.0
 isVerboseStderr :: Verbosity -> Bool
 isVerboseStderr = isVerboseFlag VStderr
+
+-- | Test if we should output warnings when we log.
+isVerboseNoWarn :: Verbosity -> Bool
+isVerboseNoWarn = isVerboseFlag VNoWarn
 
 -- | Helper function for flag testing functions.
 isVerboseFlag :: VerbosityFlag -> Verbosity -> Bool
--- a/cabal-testsuite/PackageTests/InternalVersions/BuildDependsExtra/setup.test.hs
+++ b/cabal-testsuite/PackageTests/InternalVersions/BuildDependsExtra/setup.test.hs
@@ -1,6 +1,6 @@
 import Test.Cabal.Prelude
 -- Test unneed version bound on internal build-tools deps
-main = setupAndCabalTest $ do
+main = setupAndCabalTest . expectBroken 7470 $ do
     setup' "configure" []
     assertOutputContains "extraneous version range"
         =<< setup' "sdist" []
--- a/changelog.d/pr-7470
+++ b/changelog.d/pr-7470
@@ -1,0 +1,4 @@
+synopsis: adds a verbosity flag "+nowarn", to suppress all warnings
+packages: cabal-install
+prs: #7470
+issues: #7286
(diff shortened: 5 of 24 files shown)


ptkato:
cc @ekmett 

Considering that your issue is the subject of this PR, would it suffice for your troubles or a more specific approach would be needed? As of now, it basically suppresses everything that begins with `"Warning:"`.

phadej:
Sweeping problems under a carpet. Ok fine.

Mikolaj:
@phadej: would you elaborate? what is the root problem here in your opinion?

fgaz:
I'm guessing #5119 (spurious warning) and #7270 (bug on `:` syntax that keeps it from getting past the "experimental" stage), since those are the ones mentioned in #7286.

Being able to hide warnings is nice, but I agree fixing the source should have priority

ptkato:
That is exactly the case, while fixing the root problems would be great, it still won't prevent other things coming up in the future and throwing warnings galore. An option to suppress warnings altogether is always nice, which is the objective of this PR.

phadej:
> would you elaborate? what is the root problem here in your opinion?

That flags hides **all** the warnings. #7286 asks to be able to hide **specific** warnings.

ptkato:
Oh, you meant that. Well yes, it would be possible to hide specific warnings in a similar manner, like I stated in the issue, however I took another approach due to the amount of different warnings out there.

emilypi:
@phadej do you have pointers for @ptkato to help with #7286? Embedding the warnings in the parsec instances was a naughty approach, and @ptkato's double naughtiness don't really solve the problem, but we're here because it's unclear. If you could help him, that would be very appreciated, and we can unblock @ekmett  at the same time.

phadej:
> Embedding the warnings in the parsec instances was already a naughty approach

What you mean by that?

All parser generated warnings come with an enum of their type: https://hackage.haskell.org/package/Cabal-3.4.0.0/docs/Distribution-Parsec-Warning.html#t:PWarnType so it's possible to granularly filter them. That's exactly why that enumeration is in place.

The

```
Warning: unlifted.cabal:55:27: colon specifier is experimental feature (issue
#5660)
```

is tagged with [`PWTExperimental`](https://hackage.haskell.org/package/Cabal-3.4.0.0/docs/Distribution-Parsec-Warning.html#t:PWarnType) so it's possible to filter just that. Yet, I still think that it's better to just make the feature non-experimental by implementing the outstanding issues. Because as long as feature is experimental, I'll do my best to prevent packages using such features to get into Hackage index, and that probably won't make Edward happy either.

---

Re

```
Warning: The package has an extraneous version range for a dependency on an
internal library: unlifted:{core, def, reps} ((>=0 && ==0) && ==0) && ==0,
unlifted:core ((>=0 && ==0) && ==0) && ==0. This version range includes the
current package but isn't needed as the current package's library will always
be used.
```

I admit it's a lost battle. To properly fix it some changes are needed to solver framework to distinguish which constraints are user written (i.e. are present in `.cabal` file) and which are auto-generated to make solver produce correct install plans. I don't remember (nor really ever understood) in which cases  they are needed, TL;DR you need a solver-hero to fix that issue properly, so one "pragmatic" (but sad) approach is just comment out that warning generation and make an issue if someone ever gets to it.

Adding constraints on the package itself by hand is wrong, (e.g. having `mylibrary ==0.0.1` in the test-suite), and I hope people don't do that anymore. Also it's something that `.Check` could possibly figure out, so solver won't need to care.

Cabal/src/Distribution/Fields/ParseResult.hs
```diff
@@ -188,7 +188,14 @@ parseString
     -> IO a
 parseString parser verbosity name bs = do
     let (warnings, result) = runParseResult (parser bs)
-    traverse_ (warn verbosity . showPWarning name) warnings
+
+    let noWarn = [wt | wt <- [PWTOther ..]
+            , (wt == PWTExperimental) && isVerboseNoWarnExperimental verbosity
+            ]
```
ptkato, line 194: If we ever want to add more warning suppressants for cabal parsing warnings, this fella here helps a tiny little bit with that, we just need to pop them in there.

Mikolaj: BTW, what is `PWTOther`? Is it `minBound` by any chance?

ptkato: I just picked the first on the list, but you have a good point, I should have used `minBound`, because the type does have the instances.

Mikolaj: The advantage of `minBound` is that it's clear it's the whole list. The disadvantage is that it's not clear what the type of the list is. But here it's obvious from the context.

Cabal/src/Distribution/Verbosity.hs
```diff
@@ -217,6 +218,7 @@ parsecVerbosity = parseIntVerbosity <|> parseStringVerbosity
             "stderr"     -> return verboseStderr
             "stdout"     -> return verboseNoStderr
             "nowarn"     -> return verboseNoWarn
+            "nowarnexperimental" -> return verboseNoWarnExperimental
```
ptkato, line 221: `"nowarnexperimental"` is a bit on the lengthy side, if anyone have suggestions, I'd be much obliged.

Mikolaj: I'm fine with length. However, the lack of separation of words make it hard for me to read. I think cabal uses hyphen in it's commandline? Could we somehow adopt the GHC convention perhaps (not sure how): https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html

ptkato: We could, but it would look really weird, and perhaps even out of place, in our context.

Mikolaj: Perhaps somebody on IRC would have an idea how to make `--nowarnexperimental` more readable?

Mikolaj: BTW, I can't find the other options from this code snippet in `cabal help`, so I lack context. How do I see them?

ptkato: They're not in the `cabal help`, see here:

https://github.com/haskell/cabal/blob/724e9f5cb458909766bb31e62262803a63a307fd/Cabal/src/Distribution/Verbosity.hs#L242-L249

and

https://github.com/haskell/cabal/blob/724e9f5cb458909766bb31e62262803a63a307fd/Cabal/src/Distribution/Verbosity/Internal.hs#L17-L25

Mikolaj: Thank you. All the others have the same style, so IMHO `+nowarnexperimental` is perfectly fine if we don't feel like changing everything (and worrying about backward compatibility).

Mikolaj (approved):

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -539,37 +539,39 @@ checkFields pkg =
...
+  --       ++ ". This version range includes the current package but isn't needed "
+  --       ++ "as the current package's executable will always be used."
+
+  -- , check (not (null depInternalExecutableWithImpossibleVersion)) $
+  --     PackageBuildImpossible $
+  --          "The package has an impossible version range for a dependency on an "
+  --       ++ "internal executable: "
+  --       ++ commaSep (map prettyShow depInternalExecutableWithImpossibleVersion)
+  --       ++ ". This version range does not include the current package, and must "
+  --       ++ "be removed as the current package's executable will always be used."
```
ptkato, line 574: If those are the only ones to be commented, I shall open the issue regarding them, as @phadej suggested.

Mikolaj: I think no other belong to this spam class, but are you sure all four appear as the spurious warnings?

ptkato: You're actually right, I got ahead of meself, only the first and third ones are to be commented out.

Cabal/src/Distribution/Simple/Utils.hs
```diff
@@ -442,7 +442,7 @@ verbosityHandle verbosity
 --
 warn :: Verbosity -> String -> IO ()
 warn verbosity msg = withFrozenCallStack $ do
-  when (verbosity >= normal) $ do
+  when ((verbosity >= normal) && not (isVerboseNoWarn verbosity)) $ do
```
Mikolaj, line 445: BTW, the parens around the inequality are spurious

ekmett:
This patch seems to paper over two actual issues with a fix that doesn't in the end, actually get me to a state where a user can just cabal install a thing without getting an endless scrolling session full of errors without knowing a specialized incantation. 

The two main issues I'm concerned with are a spurious complaint I get out of the bounds checker, where cabal complains about a >= 0 bound that it inserts itself and which nothing I can do can give a good user experience for, and one where you get yelled at for using a feature that was made mandatory. Both of these strike me as outright legitimate cabal design bugs.

There's no reason for either warning.

If I use multiple sub-libraries at all, which has been a big project we've been working on since the end of the 2.x era, I _have_ to incur the useless auto-inserted bound, as there is no way to ask for cabal to not hallucinate it into place. Without this being fixed outright, why did we ever build common blocks and all those things in the first place? Because to a first approximation, we simply can't use them.

If I ask for cabal 3.4 as a file format, I then _have_ to use the package:library syntax that I get warned about to use backpack at all. This was a bug fix for the fact that you couldn't have a sublibrary name that collided with a global package name. But as a result of that 'fix' I can't use sublibraries at all without spamming users.

If I have to make the user run some bizarre cabal command to explicitly opt out of these warnings, I'm not entirely clear how it is any better than asking the user to pipe their output to grep or something.

All packages that transitively depend on me then need to adopt this same bit of arcana, because when they invoke my build, they'll still get an endless stream of noise.

I confess, with the current state of this patch, I will probably (almost) never ship a package that incurs these warnings to hackage. This is a damn shame, because the alternative is for me to take like 30+ package names at a time that expose all the internal structure of my project, requiring 30x maintenance overhead, where I'd have to be constantly deprecate individually as that structure changes. Since _that_ alternative is so unpalatable, it just leads to the current status quo of me simply not really uploading new projects to hackage at all and living off github.

ekmett:
The current behavior for both of these seem actively hostile to active use.

It seems to me that the fix I'd actually want would be to find and suppress the warning about redundant version bounds when one of them is a `>= 0` bound. Its origin seems to be almost always this sort of auto-insert, and as there's no way for a user to legitimately write a cabal file that doesn't incur it, its not an actionable warning.

The warning on package:library similarly seems to be the kind of thing where warning about it on a version of cabal file format that makes it impossible to use without that notation is just blatantly saying this feature doesn't exist.

My end goal is to be able to in n + 3 years time, for n = whenever these two issues are resolved to be able to do things like break up the internals of `lens` into lots of little public sublibraries, so that users can incur just the build times for the portions they want, and yet I wouldn't take up a ton of hackage namespace.

This patch doesn't get me any closer to setting `n` to now.

Mikolaj:
@ekmett: don't worry. I think "This PR will close #7286." in description is a misleading left-over from an abandoned approach. We currently stand at "Being able to hide warnings is nice, but I agree fixing the source should have priority" from a later comment. So this PR is just a beginning. Do you have any constructive comments to this PR as such, without the assumption it's supposed to fix your #7286 problems? I already noted the remark that syntax of `cabal build --verbosity="verbose+nowarnexperimental"` is too cryptic. Any hints for an improvement? Anything else?

fgaz:
Strong agree with both @phadej and @ekmett's first comment. I think #7270 really has to be fixed before allowing hackage uploads though. Unfortunately I don't know when I'll have enough time to sit down and do that, so it'd be great if someone helped investigate it.

One solution could be to move that warning to cabal check (still keeping it fatal) or to hackage itself, so that it only gets printed when the user requests it.

fgaz (commented):
Honestly I'd remove `nowarnexperimental`: it's way too specific, and I think generalizing it would add a bit too much complexity for little gain since as per @ekmett's comment, there isn't a real need for it. If you do that it's a +1 from me.

Also remember to backport "Commenting out extraneous version warnings"! We don't want that warning to still be there in 3.6!

changelog.d/pr-7470
```diff
@@ -0,0 +1,12 @@
...
+prs: #7470
+issues: #7286
+
+description: {
+- adds a verbosity flag "+nowarn", to suppress all warnings
+- adds a verbosity flag "+nowarnexperimental", to suppress 
+  all warnings from experimental features
+- removes the warnings for extraneous versions
+  https://github.com/haskell/cabal/pull/7470#issuecomment-875878507
+}
```
fgaz, line 12: oh, and technically this should be two separate entries

ptkato: huh, how would that work? I thought it was only one changelog entry per PR.

fgaz: > I thought it was only one changelog entry per PR

CONTRIBUTING.md is indeed unclear

#7486 :)

ptkato: Done.

changelog.d/pr-7470
```diff
@@ -0,0 +1,8 @@
+synopsis: Suppress warnings for verbosity
+packages: cabal-install
+prs: #7470
+issues: #7286
+
+description: {
+- adds a verbosity flag "+nowarn", to suppress all warnings
+}
```
fgaz, line 8: I'd move the description text (description is optional) to the synopsis replacing the current synopsis. I think it's clearer for the user. Same for the other changelog entry

fgaz (approved):
Thanks!

A little suggestion for the changelog entries, since you'll have to push to fix the failing tests anyway

ptkato:
@Mikolaj & @fgaz, if you could re-review (there were changes since your reviews), I'd be much obliged.

Cabal/src/Distribution/Simple/Utils.hs
Mikolaj: Still spurious. :)

Mikolaj (approved):
Still fine. :)

cabal-testsuite/PackageTests/InternalVersions/BuildDependsExtra/setup.test.hs
```diff
@@ -2,5 +2,6 @@ import Test.Cabal.Prelude
 -- Test unneed version bound on internal build-tools deps
 main = setupAndCabalTest $ do
     setup' "configure" []
-    assertOutputContains "extraneous version range"
+    -- this specific warning was commented out, this check is inverted so it doesn't break the CI
+    assertOutputDoesNotContain "extraneous version range"
```
fgaz, line 6: Here I'd use `expectBroken` instead of flipping the condition, since the behavior we're testing for is not the one we want. Same for all the others

ptkato: Good catch, that way we needn't touch the expected outputs.

gbaz:
It looks like there's a legit test failure here?

UNEXPECTED FAIL: cabal-testsuite\PackageTests\Regression\T5213ExeCoverage\cabal.test.hs cabal-testsuite\PackageTests\Regression\T5213\cabal.test.hs

ptkato:
That's odd, they were passing before. I take it that that test was added afterwards? Well, no matter, I'll take a look into it.

Mikolaj:
@ptkato: indeed, I'm the felon in this case, the tests are from my recent PR.

Mikolaj:
Which probably means the tests are failing on 3.6 branch for a dual reason (I should have adjusted the expected results, not just backport them from master). Do we care if tests on 3.6 branch pass? I guess we do?

Mikolaj:
I've checked, #7493 and #7510 are not yet backported, so 3.6 is not broken. One of the backports will need to change expected test results to match these from the final version of this PR.

Mikolaj:
Now both above PRs are backported.

PR #7477, opened 2021-07-08
Json bytestring

fendor:
Rebase of #6948 

As a bonus, adds escaping fully conform with https://datatracker.ietf.org/doc/html/rfc8259#section-7 and tests


Mikolaj (approved):

emilypi (approved):

Cabal-tests/tests/UnitTests/Distribution/Utils/Json.hs
```diff
@@ -0,0 +1,42 @@
...
+      renderJson (JsonArray [JsonString "foo\"bar", JsonString "baz"]) @?= "[\"foo\\\"bar\",\"baz\"]"
+    , testCase "renders empty object" $
+      renderJson (JsonObject []) @?= "{}"
+    , testCase "renders singleton object" $
+      renderJson (JsonObject [("key", JsonString "foo\"bar")]) @?= "{\"key\":\"foo\\\"bar\"}"
+    , testCase "renders object" $
+      renderJson (JsonObject
+                    [ ("key", JsonString "foo\"bar")
+                    , ("key2", JsonString "baz")])
+          @?= "{\"key\":\"foo\\\"bar\",\"key2\":\"baz\"}"
```
fendor, line 30: @Mikolaj, @emilypi Do we have some form of formatting standards for new code?

Cabal/Cabal.cabal
```diff
@@ -75,7 +75,7 @@ library
 
   if !impl(ghc >= 7.8)
     -- semigroups depends on tagged.
-    build-depends:  tagged >=0.8.6 && <0.9
+    build-depends:  tagged >=0.8.6 && <0.9, bytestring-builder >= 0.10.8 && <0.11
```
fendor, line 78: From https://github.com/haskell/cabal/pull/6948#discussion_r452838749

> This is wrong as is, but I'll fix it before merge. One can have older bytestring on newer GHC, though that is contrived corner-case. But Cabal should be exemplary in own definition :)

Does that mean we don't need this dependency? Ill try to remove it and see whether CI is green.

Cabal-tests/tests/UnitTests/Distribution/Utils/Json.hs
emilypi: There's been no concrete discussion of standards. Don't worry about it for now, and we'll come up with one

Mikolaj: Right. We only have https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions, as mentioned in PR template. I guess, for now, use whatever conventions exist in the snippet you edit, unless they are utterly insane (in which case, create a reformatting commit first).

Cabal/Cabal.cabal
fendor: Ok, I guess CI tells us that it is not fine. I still don't know why it is wrong though. @phadej Would you mind elaborating?

fendor:
Theoretically ready to merge, but we need to resolve https://github.com/haskell/cabal/pull/7477#discussion_r667449590 before merging.

Cabal/Cabal.cabal
Mikolaj: @phadej: ping?

fendor:
@emilypi, @Mikolaj This PR is waiting for resolving https://github.com/haskell/cabal/pull/7477#discussion_r667449590, do you have any insights, or should be just merge?

emilypi:
@fendor the edge case seems like it'd go away as we drop GHC's in the next series of PRs. I say let's merge.

emilypi (approved):

jneira (approved):
thanks!

PR #7479, opened 2021-07-09
Properly link filename compatibility checks into `cabal check`

ag-eitilt:
This is issue 1 of 3 following up on potential cross-platform compatibility problems that were identified in #7429 (alongside #7480 and haskell/filepath#85).  That PR fixed a bug in the path validation state machine, but in doing so I realized that some of the Windows idiosyncrasies weren't being checked; the resulting discussion revealed that most of them were supposed to be checked elsewhere, but I volunteered to verify that anyway, once I got time.  I've since done so, and am now recording my results:

The good news is that characters which are invalid on Windows (`<>?`...) are properly caught and rejected.  The bad news is that there's a few file names/patterns which are invalid but *not* caught; `System.FilePath.Windows.isValid` is consulted on [line 1977](https://github.com/haskell/cabal/blob/cdd8d97268935e04ef0250194b98fbebc157787c/Cabal/src/Distribution/PackageDescription/Check.hs#L1977), but interestingly it's grandparent function is never actually called.  This patch just connects that logic as was likely intended.


ag-eitilt:
That's a bit of an embarrassing oversight.  Fixing now.

ag-eitilt:
I've taken a bit of a heavy hammer to the globbed paths, and just ignored all of them.  Would it be better to model this after `isGoodRelativeGlob` and replace the asterisks but test the rest of the path?

https://github.com/haskell/cabal/blob/05b280c36a195adc0184a1f81aac4c034d228f80/Cabal/src/Distribution/PackageDescription/Check.hs#L2244-L2249

At this point, there seems to only be a single test failing, and that test is failing purely because the wrong error message is being shown; it expects `reserved character ':'` and is getting a complete description of the entirety of bad Windows paths.  Should I disable that particular check as with the similar (but different) line above it?

https://github.com/haskell/cabal/blob/05b280c36a195adc0184a1f81aac4c034d228f80/Cabal-tests/tests/ParserTests/regressions/denormalised-paths.cabal#L63-L65

(Interestingly, the first "Important" pullout in [the Windows doc page](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#traditional-dos-paths) suggests that the path may actually be *correct*, if misleading, but that's probably something to take up with the `filepath` team instead.)

Mikolaj:
> Would it be better to model this after isGoodRelativeGlob and replace the asterisks but test the rest of the path?

Why not? Ideally there'd also be a minimal test that indicates this works.

> Should I disable that particular check as with the similar (but different) line above it?

I think what you propose makes perfect sense: delegate this to the filepath maintainers, perhaps parking this in a separate issue with a link to the filepath ticket.

ag-eitilt:
> > Would it be better to model this after isGoodRelativeGlob and replace the asterisks but test the rest of the path?
> 
> Why not? Ideally there'd also be a minimal test that indicates this works.

Good point, I've gone ahead with the replacement.  What would a minimal test would look like in this case, beyond the existing test cases passing?

> > Should I disable that particular check as with the similar (but different) line above it?
> 
> I think what you propose makes perfect sense: delegate this to the filepath maintainers, perhaps parking this in a separate issue with a link to the filepath ticket.

For the moment, I've just added the error line to get the test case to pass.  I'll eventually see what they think about the drive-relative paths (once the new maintainer settles in over there), though *if* they decide to allow it, we're going to have to revisit `isGoodRelativeFilePath` ourselves, as `C:foo/bar` is currently rejected separately by both packages.

Actually, given that, at this point we can probably remove the `isReserved` character checks from `isGoodRelativeFilePath` entirely.  The only thing we reject that `Windows.isValid` allows is a backslash, and we can give a better warning by specifically testing for that (`| c == '\\' = Just "use Unix path separators"` or something).  We'll also want to remove the `take 1`s in `checkPackageFileNames` if we do, but `isGoodRelativeFilePath` is only ever used within that same `checkPaths` function anyway, and I at least am not seeing any compelling reason to keep it.  Anyone want to go to bat for it?

Mikolaj:
> What would a minimal test would look like in this case, beyond the existing test cases passing?

Do I get it right that this is the only test for globs: `files/**/*.txt/` right now? Or did I miss a positive case, where a glob is accepted? I'd venture a single negative and single positive test would be enough until users report any tricky cases.

ag-eitilt:
Oh, is that all we have?  Yeah, I'll see about rounding that out, then.

Mikolaj:
> Oh, is that all we have?

I really don't know. You are the local sub-domain expert. :)

ag-eitilt:
Sorry it took me a while to get back around to this!  It's been a busy couple weeks.

I've added a few tests (including one which should be successfully validated) and removed the now-redundant reserved character checks from the `isGoodRelative*Path` state machines.  The tests could probably be more thorough still, but I think they cover all common uses, and am taking @Mikolaj's advice to wait until people come across edge cases in the wild.

I did discover that the question about drive-relative paths (`C:foo/bar`) has already come up; @dcoutts already prepended a `./` to the paths to explicitly treat everything as relative to the current directory.  I'll wait for anyone who's ever deliberately used those paths before reopening the question here, but we should be good for now -- it's not handled the most gracefully, but it's not going to come up with any frequency.

Finally, it's probably a given, but I'd recommend this PR be treated as a breaking change; the collection of Windows-invalid characters isn't anything that would be *too* commonly used, but I'd be surprised if there weren't two or three packages somewhere in the ecosystem with symbolic and/or informal file names, which now are unable to be uploaded to Hackage.

Mikolaj:
I've heard somewhere one can just grep the index file (or `less ~/.cabal/packages/hackage.haskell.org/01-index.tar` and search within `less`?) to verify whether Hackage contains any weird package names.

ag-eitilt:
Ah, that's where people get their Hackage stats.  (The specific index file you gave, @Mikolaj, doesn't include the file paths, but it got me looking in the right place.)  Unfortunately, what seems to be the only place which contains the file paths, `~/.cabal/packages/hackage.haskell.org/*/*/*`, only contains tarballs for the packages I've interacted with.  Is there any way to run `for pkg in */*/* ; do tar -tzf $pkg | grep '[<>:"\|?*z/]' ; done` on the database itself?  Or to download an offline mirror to do so locally?

Also, the build checks seem to have failed due to an issue with the CI infrastructure rather than anything I did.  I know that's happened to other projects, but it had been working just fine here before.

Mikolaj:
Yes, we have a CI bug, but it's fixed now. I've tried to rerun the checks, but it doesn't quite work. It will work once you rebase the branch (BTW, the CI is probably broken on your branch).

gbaz:
If the only failure mode is that previously uploadable packages now can't be uploaded, that's fine. This will just nudge people to fix their packages :-)

ag-eitilt:
Hello, doctest, my old friend.  I prefer running tests manually to relying on CI, at least before merging, but that does mean I don't always notice every issue.  That test failure should now be fixed, though I continue to have issues running doctest locally (it's now failing while saying that parsec is hidden...)

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -1966,25 +1972,41 @@ repoTypeDirname Pijul     = [".pijul"]
...
+    ]
+
+checkWindowsPaths :: [(Bool, FilePath)] -> Maybe PackageCheck
+checkWindowsPaths paths =
+    case filter (not . FilePath.Windows.isValid . escape) paths of
+      [] -> Nothing
+      ps -> Just $
+        PackageDistInexcusable $
+             "Unfortunately, the " ++ quotes (map snd ps) ++ " on Windows "
+          ++ "which would cause portability problems for this package. "
```
Mikolaj, line 1993: Not sure how to improve the wording. Would "paths" instead of "the files" be as accurate? Is 'C:/foo" a path or is it a drive and a path? Also, I guess there should be a comma before 'which'?

Mikolaj (approved):
Great job.

Cabal/src/Distribution/PackageDescription/Check.hs
ag-eitilt: How does "the path(s) ... is/are invalid on Windows, which" sound?  Getting into whether the drive is separate from the path is probably a bit more detail than necessary (most users would just consider it all a single path, correct or not), especially since the vast majority of paths in .cabal files are required to be relative.  We should also probably remove the "Unfortunately" to better fit the other warning messages; the *only* other ones which use it describe needing to specify a newer `cabal-version`.  The below is also probably worth comparing and updating for more consistency (if nothing else, not everything that fails is "reserved" per se):

https://github.com/haskell/cabal/blob/bb516ed31bf31758627e25a663071ac4d2acc8ba/Cabal/src/Distribution/PackageDescription/Check.hs#L441-L446

Mikolaj: Agreed on all accounts.

gbaz (approved):
lgtm. thanks for the detailed work on this!

Mikolaj:
@ag-eitilt: please feel free to rebase and merge

ag-eitilt:
@Mikolaj I just rebased it yesterday!  Y'all keep pulling the rug from under me. 😛  Really, though, thanks to all of you for helping walk me through this!

emilypi (approved):

ag-eitilt:
@emilypi Thanks!  I had left the checks to run overnight intending to merge in the morning, but that's certainly a bit less prone to being forgotten.

emilypi:
Mergify bot is a good boi :)

Mikolaj:
@ag-eitilt: congrats

Issue #7466, opened 2021-06-29
Cabal takes 3 minutes to resolve dependencies


dfithian:
**Describe the bug**
I'm converting a stack project to cabal. The project has over 200 packages. I'm noticing that the cabal solver takes 2-3 minutes while stack is pretty much instantaneous. I figure I must not be aware of some cabal feature that stack is leveraging, and that this in fact is not a bug, but I thus far have been unable to contact the maintainers.

**To Reproduce**
```
$ curl https://www.stackage.org/lts-17.10/cabal.config > cabal.project.freeze
$ cabal v2-update
# add `allow-older: all` and `allow-newer: all` to `cabal.project.freeze` and iterate on the solver output
$ cabal v2-build all
Resolving dependencies # takes 2-3 min
```

**Expected behavior**
There should be a way to resolve dependencies quickly. In lieu of that, perhaps there's a way that I'm missing to tell cabal to resolve dependencies once, and then cache the result (because the "Resolving dependencies" step happens _whenever_ `cabal build all` is run).

**System information**
- Mac OS Catalina
- Cabal is 3.4.0.0 (installed with ghcup)
- GHC is 8.10.4 (installed with ghcup)

**Additional context**
* When I run `cabal v2-freeze`, the dependency resolution time is 90 seconds. That's a lot better, but still very slow.
* Removing wildcards from `allow-newer`/`allow-older` has no effect.
* Adding `reject-unconstrained-dependencies: all` fails because `rts` is not pinned, but it's not available in Hackage as far as I can tell 😕 
* Adding `max-backjumps: 0` does not improve the performance

Mikolaj:
I hope somebody comes up with a hint to speed it up, particularly when repeating the builds. I'd also try with a nightly build of dev version of cabal (https://github.com/haskell/cabal/issues/7458#issuecomment-869494367).

If no easy solution, it's definitely worth investigating where the time is spent (unless it's widely known for projects of that size with freeze files). I'd guess, in the solver, because checking satisfiability of constraints (freeze file) is a special, but not too special, kind of constraint solving, which is costly even without backtracking.

(BTW, did you try `cabal build .` or `cabal build exe: foo`? I seem to remember there was a difference once I used it.)

fgaz:
> There should be a way to resolve dependencies quickly. In lieu of that, perhaps there's a way that I'm missing to tell cabal to resolve dependencies once, and then cache the result (because the "Resolving dependencies" step happens whenever cabal build all is run).

[...]

PR #7491, opened 2021-07-23
use config (i.e. freeze) constraints also as solver version prefs

gbaz:
This extends version prefs handed to the solver (to guide it) by any version constraints fixed directly in a project config (or freeze) file. It cuts down the number of local backjumps, because the prefs are applied before directly traversing the version tree. This was done to see if it could improve the situation in #7466 but my tests don't seem to show a significant speedup -- i.e. local backjumps appear cheap, similar to #7490 .

So I'm not sure if there's any significant benefit in ensuring preferences are applied before running the solver and just letting the constraint conflict machinery immediately apply the fixed version constraints from the freeze file. Again interested in feedback from the Usual solver experts (@dcoutts @grayjay @ezyang and anyone else I may not have thought of here).


grayjay:
As in #7490, I think that preferences shouldn't have a significant effect on performance if there are already constraints on the same choices, since the constraints should be enforced at the time that the choices are made. The solver also creates log messages about versions that were rejected by constraints by traversing those versions' nodes.

gbaz:
As discussed here and on the appropriate tickets it seems like this doesn't really do something useful, so closing for now.

fgaz:
Do we leave the branch there?

Mikolaj:
IMHO the users may want to test if this one brings any improvements, if they still yearn for more or their scenario changes. But maybe I'm cargo-culting: gbaz==speed.

PR #7497, opened 2021-07-27
Force cabal to use a build plan where exe:cabal-plan is enabled

fendor:
For some reason, `cabal v2-install cabal-plan` installs the library on macos instead of the executable, causing CI failures such as: https://github.com/haskell/cabal/pull/7477/checks?check_run_id=3170346447

Logs from CI:
``` 
  cd $(mktemp -d)
  cabal v2-install cabal-plan --constraint='cabal-plan ^>=0.6.2.0' --constraint='aeson +fast'
  shell: /bin/bash -e {0}
Resolving dependencies...
Build profile: -w ghc-8.8.3 -O1
In order, the following will be built (use -v for more details):
 - base-compat-0.11.2 (lib) (requires download & build)
 - base-orphans-0.8.4 (lib) (requires download & build)
 - base16-bytestring-0.1.1.7 (lib) (requires download & build)
 - dlist-1.0 (lib) (requires download & build)
 - hashable-1.3.2.0 (lib) (requires download & build)
 - indexed-traversable-0.1.1 (lib) (requires download & build)
 - integer-logarithms-1.0.3.1 (lib) (requires download & build)
 - primitive-0.7.1.0 (lib) (requires download & build)
 - splitmix-0.1.0.3 (lib) (requires download & build)
 - tagged-0.8.6.1 (lib) (requires download & build)
 - th-abstraction-0.4.2.0 (lib) (requires download & build)
 - transformers-compat-0.7 (lib) (requires download & build)
 - base-compat-batteries-0.11.2 (lib) (requires download & build)
 - time-compat-1.9.6 (lib) (requires download & build)
 - unordered-containers-0.2.14.0 (lib) (requires download & build)
 - data-fix-0.3.1 (lib) (requires download & build)
 - vector-0.12.3.0 (lib) (requires download & build)
 - scientific-0.3.7.0 (lib) (requires download & build)
 - random-1.2.0 (lib) (requires download & build)
 - distributive-0.6.2.1 (lib) (requires download & build)
 - attoparsec-0.14.1 (lib) (requires download & build)
 - uuid-types-1.0.5 (lib) (requires download & build)
 - comonad-5.0.8 (lib) (requires download & build)
 - bifunctors-5.5.11 (lib) (requires download & build)
 - assoc-1.0.2 (lib) (requires download & build)
 - these-1.1.1.1 (lib) (requires download & build)
 - strict-0.4.0.1 (lib) (requires download & build)
 - aeson-1.5.6.0 (lib) (requires download & build)
[...]

PR #7499, opened 2021-07-28
Backport 3.6: Force cabal to use a build plan where exe:cabal-plan is enabled

fendor:
Backport of #7497


fgaz (approved):
Please use the `-x` option in the `cherry-pick` command when backporting so that the original commit is tracked (maybe this should be documented in CONTRIBUTING.md...).

I'm approving in advance :)

Issue #7489, opened 2021-07-21
 Implementation and purpose of Show-Build-Info


fendor:
--- 

Nomenclature (uses recursive name bindings)
```
Cabal = `lib:Cabal`
cabal = `exe:cabal`
v2-build = nix-style build command for project building (happens in exe:cabal)
Setup.hs build = Cabal build system, e.g. what happens under the hood (happens in lib:Cabal)
v2-show-build-info = nix-style show-build-info command, cli interface for users (happens in exe:cabal)
Setup.hs show-build-info = Cabal show-build-info system, performs actual work (happens in lib:Cabal)
``` 
---

The command `show-build-info` is intended to help tooling based on Cabal to find detailed information on how to build each component individually. It is not intended to expose Cabal details, such as what preprocessors are executed, etc... It also does not subsume Cabal's build process, i.e. having the output of `show-build-info` does not suffice to build a cabal component in certain situations. For example it is still Cabal's task to generate `Paths_*.hs`. Therefore, `show-build-info` only works to compile a project, if Cabal has basically built the project, already.

But then tooling such as [hie-bios](https://github.com/mpickering/hie-bios/) and [hsinspect](https://gitlab.com/tseenshe/hsinspect) can use the output of `show-build-info` to feed them into a GHC session and provide IDE functionality.

## Implementation

Extracting compiler information is theoretically not a problem for cabal. Since we parse the `.cabal` file description and already solve for which dependencies to use, basically all information should be present, right? However, Cabal has been designed with extensibility in mind and achieved that by allowing developers to overwrite certain cycles of a build via custom Setup.hs hooks. These hooks are basically a way for a developer of a Cabal package to overwrite or redefine various build-steps in Cabal.
This means that while `exe:cabal` knows everything about the package to build, `lib:Cabal` can change compilation flags when building the project.
Therefore, `exe:cabal` alone can not provide us with the low-level details on how to compile a component, as only lib:Cabal has the full build information (when using `build-type: Custom`). Thus, `show-build-info` must be a part of lib:Cabal, as long as we have build hooks in order to generate the correct build-info reliably.

For the implementation itself, we present three ideas:

### Dump build info as side effect in `v2-build`

**Idea:** Add an option to `exe:cabal` v2-build to dump the build information during the build. We would explicitly invoke `Setup.hs show-build-info` from the `v2-build` command.

**Pros:**

* No extra command
* Keeps build-info up-to-date automatically
* Can be easily used by editors without extra invocation

**Con:**

[...]

PR #7500, opened 2021-08-01
Add `status` command to cabal

fendor:
Lightweight command that can query for very basic information
in a cabal project.
In particular, information about the compiler for the project and the
unit-id a given target belongs to.
Other flags are bound to follow.

A document that lays out the plan in more detail: https://gist.github.com/fendor/796b373aba26e12dbdecf7e40af78b32

This is the second part to #7489. It allows IDEs and hie-bios to quickly query for certain project meta information, in this case compiler information and what unit a target string belongs to.

Example:
```sh
> cabal status --output-format=json --compiler | jq
{
  "cabal-version": "3.7",
  "compiler": {
    "flavour": "ghc",
    "compiler-id": "ghc-8.10.7",
    "path": "/home/hugin/.ghcup/bin/ghc"
  }
}
```

```sh
> cabal status --output-format=json --compiler --target="lib:cabal" | jq
{
  "cabal-version": "3.7",
  "compiler": {
    "flavour": "ghc",
    "compiler-id": "ghc-8.10.7",
    "path": "/home/hugin/.ghcup/bin/ghc"
  },
  "targets": [
    {
      "target": "lib:cabal",
      "unit-id": "Cabal-3.7.0.0-inplace"
    }
  ]
}
```

Target strings that resolve to the same component are listed both times.

```sh
> cabal status --output-format=json --compiler --target="lib:cabal" --target="./Cabal/src/Distribution/Make.hs" | jq
{
  "cabal-version": "3.7",
  "compiler": {
    "flavour": "ghc",
    "compiler-id": "ghc-8.10.7",
    "path": "/home/hugin/.ghcup/bin/ghc"
  },
  "targets": [
    {
      "target": "lib:cabal",
      "unit-id": "Cabal-3.7.0.0-inplace"
    },
    {
      "target": "./Cabal/src/Distribution/Make.hs",
      "unit-id": "Cabal-3.7.0.0-inplace"
    }
  ]
}
```

Main advantage: 
* Very quick (only perform work you actually need)
* Extensible
* Composable

Disadvantage:
* Maintain cabal command
* Do we really need it?
* Are we sure we are not just blindly adding commands without considering a bigger picture?

Alternatives:
* Use `cabal build --dry-run` to generate `plan.json` and then parse that one.
    * How to find the location of `plan.json` given a target string?
* Merge this with another command
    * E.g. Add option `cabal build --dry-run --targets` or something like that, but it feels like terrible UX.

Missing:
- [ ] Write tests in cabal-testsuite
  - [x] No build-plan 
  - [x] Some third-party dependency doesn't build
  - [x] Some local dependency doesn't build
  - [x] Ignores invalid target selectors
  - [x] Disabled units
  - [x] Unit fails to build
  - [ ] GHCJS, what's the output?
  - [x] Check plan.json is generated
  - [x] Bench unit
  - [x] Exe unit
  - [x] Flib unit 
  - [x] Lib unit
  - [x] Test unit
  - [ ] Custom Setup
- [ ] Documentation
  - [ ] Write Command docs
  - [ ] Define json schema
  - [ ] Explicitly mark as experimental
  - [ ] Show some use-cases
  - [ ] Future features
    -  [ ] Cabal-plan why-depends

<details>
<summary>
Old PR description, here for completeness
</summary>

Lightweight command that can query for very basic information
in a cabal project.
In particular, you can list all targets in your project or print
information about the compiler. Other flags are bound to follow.

This is the second part to #7489. It allows IDEs and hie-bios to quickly query for certain project meta information, in this case compiler information and available targets.

Example:
```sh
> cabal ide --project-compiler
Resolving dependencies...
Compiler: ghc
Version: 8.10.2
Path: /home/hugin/.ghcup/bin/ghc-8.10.2
```
```sh
> cabal ide --targets
lib:Cabal
lib:cabal-testsuite
exe:setup
exe:cabal-tests
lib:cabal-install
exe:cabal
test:long-tests
test:integration-tests2
test:memory-usage-tests
test:unit-tests
lib:cabal-install-solver
test:unit-tests
lib:solver-benchmarks
exe:hackage-benchmark
test:unit-tests
lib:Cabal-QuickCheck
lib:Cabal-tree-diff
lib:Cabal-described
test:no-thunks-test
test:rpmvercmp
test:hackage-tests
test:custom-setup-tests
test:check-tests
test:parser-tests
test:unit-tests
test:cabal-benchmarks
lib:cabal-doctest
```

Main advantage: 
* Very quick (only perform work you actually need)
* Extensible

Disadvantage:
* Maintain cabal command
* Do we really need it?
* Are we sure we are not just blindly adding commands without considering a bigger picture?

Alternatives:
* Use `cabal build --dry-run` to generate `plan.json` and then parse that one.
    * How to find the location of `plan.json`?
* Merge this with another command
    * E.g. Add option `cabal build --dry-run --targets` or something like that, but it feels like terrible UX.

Missing:
* ~~[ ] Documentation~~
* ~~[ ] Test-cases~~
</details>

### DIFF
--- a/cabal-install/main/Main.hs
+++ b/cabal-install/main/Main.hs
@@ -83,6 +83,7 @@ import qualified Distribution.Client.CmdHaddock   as CmdHaddock
 import qualified Distribution.Client.CmdHaddockProject as CmdHaddockProject
 import qualified Distribution.Client.CmdInstall   as CmdInstall
 import qualified Distribution.Client.CmdRun       as CmdRun
+import qualified Distribution.Client.CmdStatus    as CmdStatus
 import qualified Distribution.Client.CmdTest      as CmdTest
 import qualified Distribution.Client.CmdBench     as CmdBench
 import qualified Distribution.Client.CmdExec      as CmdExec
@@ -274,6 +275,7 @@ mainWorker args = do
       , hiddenCmd  actAsSetupCommand actAsSetupAction
       , hiddenCmd  manpageCommand (manpageAction commandSpecs)
       , regularCmd CmdListBin.listbinCommand     CmdListBin.listbinAction
+      , regularCmd CmdStatus.statusCommand       CmdStatus.statusAction
 
       ] ++ concat
       [ newCmd  CmdConfigure.configureCommand CmdConfigure.configureAction
--- a/cabal-install/src/Distribution/Client/CmdStatus.hs
+++ b/cabal-install/src/Distribution/Client/CmdStatus.hs
@@ -1,0 +1,366 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE LambdaCase #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Client.CmdStatus
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- Implementation of the 'status' command. Query for project information
+-- such as targets in the project or which ghc version is going to be used
+-- to build the project.
+-----------------------------------------------------------------------------
+
+module Distribution.Client.CmdStatus (
+  statusCommand, statusAction,
+  ) where
+
+import qualified Data.Map as Map
+
+import Prelude ()
+import Distribution.Client.Compat.Prelude
+
+import Distribution.Client.DistDirLayout
+import Distribution.Client.TargetProblem
+import Distribution.Client.CmdErrorMessages
+import qualified Distribution.Client.InstallPlan as InstallPlan
+import Distribution.Client.NixStyleOptions
+         ( NixStyleFlags (..), nixStyleOptions, defaultNixStyleFlags )
+import Distribution.Client.ProjectOrchestration
+import Distribution.Client.ProjectPlanning
+import Distribution.Client.ProjectPlanning.Types
+import Distribution.Client.Setup
+         ( GlobalFlags, ConfigFlags(..), yesNoOpt )
+import Distribution.Client.Utils.Json
+         ( (.=) )
+import qualified Distribution.Client.Utils.Json as Json
+import Distribution.Client.Version
+         ( cabalInstallVersion )
+import Distribution.InstalledPackageInfo
+         ( InstalledPackageInfo )
+import Distribution.Parsec (parsecCommaList, parsecToken)
+import Distribution.ReadE
+         ( ReadE(ReadE), parsecToReadE )
+import Distribution.Simple.BuildPaths (buildInfoPref)
+import Distribution.Simple.Command
+         ( CommandUI(..), option, reqArg, ShowOrParseArgs, OptionField )
+import Distribution.Simple.Compiler
+import Distribution.Simple.Program
+import Distribution.Simple.Flag
+         ( Flag(..), fromFlagOrDefault )
+import Distribution.Simple.Utils
+         ( wrapText, die', withOutputMarker, ordNub )
+import Distribution.Verbosity
+         ( normal )
+import Distribution.Version
+
+-------------------------------------------------------------------------------
+-- Command
+-------------------------------------------------------------------------------
+
+statusCommand :: CommandUI (NixStyleFlags StatusFlags)
+statusCommand = CommandUI
+  { commandName = "status"
+  , commandSynopsis = "Query for simple project information"
+  , commandDescription  = Just $ \_ -> wrapText $
+      "Query for available targets and project information such as project GHC."
+  , commandNotes = Just $ \pname ->
+        "Examples:\n"
+     ++ "  " ++ pname ++ " status --output-format=json --compiler-info\n"
+     ++ "    Print the compiler that is used for this project in the json format.\n"
+     ++ "  " ++ pname ++ " status --output-format=json --build-info=./src/Foo.hs\n"
+     ++ "    Print the location of the component \"src/Foo.hs\" belongs to.\n"
+     ++ "  " ++ pname ++ " status --output-format=json --build-info=./src/Foo.hs\n"
+     ++ "    Print both, compiler information and build-info location for the given target.\n"
+     ++ "  " ++ pname ++ " status --output-format=json --build-info=./src/Foo.hs --build-info=./test/Bar.hs\n"
+     ++ "    Print build-info location for multiple targets.\n"
+  , commandUsage = \pname ->
+      "Usage: " ++ pname ++ " status [FLAGS]\n"
+  , commandDefaultFlags = defaultNixStyleFlags defaultStatusFlags
+  , commandOptions      = nixStyleOptions statusOptions
+
+  }
+
+-------------------------------------------------------------------------------
+-- Flags
+-------------------------------------------------------------------------------
+
+data StatusOutputFormat
+  = JSON
+  deriving (Eq, Ord, Show, Read)
+
+data StatusFlags = StatusFlags
+  { statusBuildInfo :: [String]
+  , statusCompiler :: Flag Bool
+  , statusOutputFormat :: Flag StatusOutputFormat
+  } deriving (Eq, Show, Read)
+
+defaultStatusFlags :: StatusFlags
+defaultStatusFlags = StatusFlags
+  { statusBuildInfo = mempty
+  , statusCompiler = mempty
+  , statusOutputFormat = mempty
+  }
+
+statusOutputFormatParser :: ReadE (Flag StatusOutputFormat)
+statusOutputFormatParser = ReadE $ \case
+  "json" -> Right $ Flag JSON
+  policy -> Left  $ "Cannot parse the status output format '"
+            <> policy <> "'"
+
+statusOutputFormatPrinter
+  :: Flag StatusOutputFormat -> [String]
+statusOutputFormatPrinter = \case
+  (Flag JSON) -> ["json"]
+  NoFlag      -> []
+
+statusOptions :: ShowOrParseArgs -> [OptionField StatusFlags]
+statusOptions showOrParseArgs =
+  [ option [] ["output-format"]
+    "Output Format for the information"
+    statusOutputFormat (\v flags -> flags { statusOutputFormat = v })
+    (reqArg "json"
+      statusOutputFormatParser
+      statusOutputFormatPrinter
+    )
+  , option [] ["build-info"]
+    "List all available targets in the project"
+    statusBuildInfo (\v flags -> flags { statusBuildInfo = v ++ statusBuildInfo flags})
+    (reqArg "TARGET" buildInfoTargetReadE (fmap show))
+  , option [] ["compiler-info"]
+    "Print information of the project compiler"
+    statusCompiler (\v flags -> flags { statusCompiler = v })
+    (yesNoOpt showOrParseArgs)
+  ]
+  where
+    buildInfoTargetReadE :: ReadE [String]
+    buildInfoTargetReadE =
+      parsecToReadE
+        -- This error should never be shown
+        ("couldn't parse targets: " ++)
+        -- TODO: wrong parser, kills filepaths with spaces
+        (parsecCommaList parsecToken)
+
+-------------------------------------------------------------------------------
+-- Action
+-------------------------------------------------------------------------------
+
+-- | Entry point for the 'status' command.
+statusAction :: NixStyleFlags StatusFlags -> [String] -> GlobalFlags -> IO ()
+statusAction flags@NixStyleFlags { extraFlags = statusFlags, ..} cliTargetStrings globalFlags = do
+  when (NoFlag == statusOutputFormat statusFlags) $ do
+    die' verbosity "The status command requires the flag '--output-format'."
+  when (not $ null cliTargetStrings) $
+    die' verbosity "The status command takes not target arguments directly. Use appropriate flags to pass in target information."
+
+  baseCtx <- establishProjectBaseContext verbosity cliConfig OtherCommand
+  (_, elaboratedPlan, elabSharedConfig, _, _) <-
+    rebuildInstallPlan verbosity
+                        (distDirLayout baseCtx)
+                        (cabalDirLayout baseCtx)
+                        (projectConfig baseCtx)
+                        (localPackages baseCtx)
+
+  let initialJson = Json.object
+        [ "cabal-version" .= jdisplay cabalInstallVersion
+        ]
+
+  compilerJson <- if not $ fromFlagOrDefault False (statusCompiler statusFlags)
+    then pure $ Json.object [] -- Neutral element
+    else do
+      let compiler = pkgConfigCompiler elabSharedConfig
+      compilerProg <- requireCompilerProg verbosity compiler
+      let progDb = pkgConfigCompilerProgs elabSharedConfig
+      (configuredCompilerProg, _) <- requireProgram verbosity compilerProg progDb
+      pure $ mkCompilerInfo configuredCompilerProg compiler
+
+  buildInfoJson <- if null (statusBuildInfo statusFlags)
+    then pure $ Json.object [] -- Neutral element
+    else do
+      let targetStrings = statusBuildInfo statusFlags
+      targetSelectors <- readTargetSelectors (localPackages baseCtx) Nothing targetStrings >>= \case
+        Left err   -> reportTargetSelectorProblems verbosity err
+        Right sels -> pure sels
+
+      -- Interpret the targets on the command line as build targets
+      -- (as opposed to say repl or haddock targets).
+      -- TODO: don't throw on targets that are invalid.
+      targets <- either (reportBuildTargetProblems verbosity) return
+                $ resolveTargets
+                    selectPackageTargets
+                    selectComponentTarget
+                    elaboratedPlan
+                    Nothing
+                    targetSelectors
+
+      pure $ mkBuildInfoJson (distDirLayout baseCtx) elabSharedConfig
+              elaboratedPlan targets targetSelectors targetStrings
+
+  let statusJson = mergeJsonObjects [initialJson, compilerJson, buildInfoJson]
+
+  -- Final output
+  putStrLn $ withOutputMarker verbosity $ Json.encodeToString statusJson
+  where
+    verbosity = fromFlagOrDefault normal (configVerbosity configFlags)
+    cliConfig = commandLineFlagsToProjectConfig globalFlags flags mempty
+
+-- ----------------------------------------------------------------------------
+-- Helpers for determining and serialising compiler information
+-- ----------------------------------------------------------------------------
+
+requireCompilerProg :: Verbosity -> Compiler -> IO Program
+requireCompilerProg verbosity compiler =
+  case compilerFlavor compiler of
+    GHC -> pure ghcProgram
+    GHCJS -> pure ghcjsProgram
+    flavour -> die' verbosity $
+      "status: Unsupported compiler flavour: "
+      <> prettyShow flavour
+
+mkCompilerInfo :: ConfiguredProgram -> Compiler -> Json.Value
+mkCompilerInfo compilerProgram compiler =
+  Json.object
+    [ "compiler" .= Json.object
+        [ "flavour"     .= Json.String (prettyShow $ compilerFlavor compiler)
+        , "compiler-id" .= Json.String (showCompilerId compiler)
+        , "path"        .= Json.String (programPath compilerProgram)
+        ]
+    ]
+
+-- ----------------------------------------------------------------------------
+-- Helpers for determining and serialising build info
+-- ----------------------------------------------------------------------------
+
+mkBuildInfoJson :: DistDirLayout -> ElaboratedSharedConfig -> ElaboratedInstallPlan -> TargetsMap -> [TargetSelector] -> [String] -> Json.Value
+mkBuildInfoJson distDirLayout elaboratedSharedConfig elaboratedPlan targetsMap targetSelectors targetStrings = Json.object
+  [ "build-info" .= Json.Array allTargetsJsons
+  ]
+  where
+    allTargetsJsons =
+      [ planPackageToJ elab ts
+      | (uid, elab) <- Map.assocs subsetInstallPlan
+      , (_, tss) <- targetsMap Map.! uid
+      , ts <- ordNub $ toList tss
+      ]
+
+    subsetInstallPlan = Map.restrictKeys (InstallPlan.toMap elaboratedPlan) (Map.keysSet targetsMap)
+
+    targetsTable = Map.fromList $ zip targetSelectors targetStrings
+
+    tsToOriginalTarget ts = targetsTable Map.! ts
+
+    planPackageToJ :: ElaboratedPlanPackage -> TargetSelector -> Json.Value
+    planPackageToJ pkg ts =
+      case pkg of
+        InstallPlan.PreExisting ipi -> installedPackageInfoToJ ipi
+        InstallPlan.Configured elab -> elaboratedPackageToJ elab ts
+        InstallPlan.Installed  elab -> elaboratedPackageToJ elab ts
+        -- Note that the --build-info currently only uses the elaborated plan,
+        -- not the improved plan. So we will not get the Installed state for
+        -- that case, but the code supports it in case we want to use this
+        -- later in some use case where we want the status of the build.
+
+    -- TODO: what should we do if we run in this case?
+    -- Happens on `--build-info=containers` while we are not in the containers project.
+    installedPackageInfoToJ :: InstalledPackageInfo -> Json.Value
+    installedPackageInfoToJ _ipi =
+      -- Pre-existing packages lack configuration information such as their flag
+      -- settings or non-lib components. We only get pre-existing packages for
+      -- the global/core packages however, so this isn't generally a problem.
+      -- So these packages are never local to the project.
+      --
+      Json.object []
+
+    elaboratedPackageToJ :: ElaboratedConfiguredPackage -> TargetSelector -> Json.Value
+    elaboratedPackageToJ elab ts = Json.object
+      [ "target" .= Json.String (tsToOriginalTarget ts)
+      , "path" .= maybe Json.Null Json.String buildInfoFileLocation
+      ]
+      where
+      dist_dir :: FilePath
+      dist_dir = distBuildDirectory distDirLayout
+                    (elabDistDirParams elaboratedSharedConfig elab)
+
+      -- | Only add build-info file location if the Setup.hs CLI
+      -- is recent enough to be able to generate build info files.
+      -- Otherwise, write 'null'.
+      --
+      -- Consumers of `status` can use the nullability of this file location
+      -- to indicate that the given component uses `build-type: Custom`
+      -- with an old lib:Cabal version.
+      buildInfoFileLocation :: Maybe FilePath
+      buildInfoFileLocation
+        | elabSetupScriptCliVersion elab < mkVersion [3, 7, 0, 0]
+        = Nothing
+        | otherwise
+        = Just (buildInfoPref dist_dir)
+
+-- ----------------------------------------------------------------------------
+-- Target selectors and helpers
+-- ----------------------------------------------------------------------------
+
+-- | This defines what a 'TargetSelector' means for the @status@ command.
+-- It selects the 'AvailableTarget's that the 'TargetSelector' refers to,
+-- or otherwise classifies the problem.
+--
+-- For the @status@ command select all components except non-buildable
+-- and disabled tests\/benchmarks, fail if there are no such
+-- components
+--
+selectPackageTargets :: TargetSelector
+                     -> [AvailableTarget k] -> Either TargetProblem' [k]
+selectPackageTargets targetSelector targets
+
+    -- If there are any buildable targets then we select those
+  | not (null targetsBuildable)
+  = Right targetsBuildable
+
+    -- If there are targets but none are buildable then we report those
+  | not (null targets)
+  = Left (TargetProblemNoneEnabled targetSelector targets')
+
+    -- If there are no targets at all then we report that
+  | otherwise
+  = Left (TargetProblemNoTargets targetSelector)
+  where
+    targets'         = forgetTargetsDetail targets
+    targetsBuildable = selectBuildableTargetsWith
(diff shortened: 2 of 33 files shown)


cabal-install/src/Distribution/Client/CmdIde.hs
```diff
@@ -0,0 +1,126 @@
...
+    putStrLn $ unlines
+      [ "Compiler: " ++ programId compilerProg
+      , "Version: " ++ maybe "UNKNOWN" prettyShow (programVersion compilerProg)
+      , "Path: " ++ programPath compilerProg
+      ]
+
+  when (fromFlagOrDefault False (ideTargets ideFlags)) $ do
+    putStrLn "Available Targets:"
+    knownTargets <- getKnownTargets defaultDirActions localPackages
+    for_ (knownComponentsAll knownTargets) $ \KnownComponent{..} -> do
```
fendor, line 116: This impl is wrong, but getting it right is actually non-trivial.

michaelpj:
If the intent is for these to be machine-readable, would it make sense to dump them out in JSON or similar by default?

phadej:
> If the intent is for these to be machine-readable, would it make sense to dump them out in JSON or similar by default?

What's wrong with reading `plan.json` then. `cabal-plan` library has [`SearchPlanJson`](https://hackage.haskell.org/package/cabal-plan-0.7.2.0/docs/Cabal-Plan.html#t:SearchPlanJson) and [`findPlanJson`](https://hackage.haskell.org/package/cabal-plan-0.7.2.0/docs/Cabal-Plan.html#v:findPlanJson) find the `plan.json` files.

fendor:
> What's wrong with reading `plan.json` then. 

Personally, I think it is a bit troublesome to find `plan.json` reliably without user input. If we can mitigate that somehow, the `ide` command could also be just a frontend to query `plan.json` efficiently and backwards compatible.

Additionally, what is the simplest/efficient way to generate `plan.json`? `cabal build all --dry-run`? That takes on the cabal code-base 2.5s, while the `ide` command needs a couple of milliseconds.

However, I can see how appealing it would be to not have to introduce a new command...

jneira:
> This is the second part to #7489. It allows IDEs and hie-bios to quickly query for certain project meta information, in this case compiler information and available targets.

So it the replacement od `cabal show-build-info`? Afaiu it is complementary to `cabal build --ddump-build-info=file`, no? I mean all the info this command can offer will be in the show-build-info json, right? (and then the json requested by @michaelpj would be already covered)
In any case i like the design, the discoverability and it could be used for other tooling than ide's, like shell scripts

fendor:
> I mean all the info this command can offer will be in the show-build-info json, right?
 
 Not only from `build-info.json`, e.g. available target information is from `plan.json`.
 
> So it the replacement od `cabal show-build-info`?

Basically yes, but intended to be a bit more general.

> Afaiu it is complementary to cabal build --ddump-build-info=file, no?

Yep, ideally a quick way to query for project metadata, in this case project compiler and targets.

emilypi:
@Mergifyio update

gbaz:
Some slight bikeshedding since we closed the old status pr in favor of this. I'd prefer the name be something like "status" (now available!) or "build-info" or the like, since I imagine end users and not just ides might want this.

Also I'd encourage taking a look at this `status` output from the old draft PR to see if all the information it provides is covered by this -- maybe the old work covers something missed/forgotten here? :-) 

https://github.com/haskell/cabal/pull/2882#issuecomment-150368527

emilypi:
@fendor what do you think about @gbaz's comment?

fendor:
I did not have time yet to revisit this PR.

jneira:
@fendor ping, would be great to see this in 3.8 if possible

andreasabel:
I agree with @gbaz that `ide` isn't the best name for a command that dumps some computed information.  `ide` promises to crank up an IDE for package development or something grand like that.
Alt: `values`, `show-info` etc.

fendor:
> Some slight bikeshedding since we closed the old status pr in favor of this. I'd prefer the name be something like "status" (now available!) or "build-info" or the like, since I imagine end users and not just ides might want this.
> 
> Also I'd encourage taking a look at this `status` output from the old draft PR to see if all the information it provides is covered by this -- maybe the old work covers something missed/forgotten here? :-)
> 
> [#2882 (comment)](https://github.com/haskell/cabal/pull/2882#issuecomment-150368527)

The suggested output seems to me like a harder-to-read `plan.json`. I can see some merits to it, but I think the bigger question currently is: What are desired features, and what will the interface look like?

I nominate some features:
* Quick access to project-metadata
  * Specifically, I want to access certain information (such as GHC version and location) without a valid build-plan.
* Easier access to project information we (a user of cabal) are commonly interested in. Probably just about anything `cabal-plan` provides fits here, but do we really want to copy `cabal-plan`?
* Allow us to query for more detailed information per component
  * E.g. given `cabal status <component-selector> --build-info` should give us build-info (details what this command entails can be ignored for *now*).
  
Of today, everything listed here (except the first point) can be easily achieved using `cabal-plan` and `cabal`. 

If everything can be already done with `cabal-plan`, do we really want to be a competitor to it? Is a command really warranted if the only thing that I want is just a quick way to get project metadata?

jneira:
Is this still blocked on a decision? :worried:

fendor:
I had a fairly long call with @Kleidukos about this, and we had the following idea:

`cabal status` should essentially be a query interface to information available in plan.json and build-info.json. 

The cli should form an algebra on this information, with this CLI roughly suggested:
```
# cabal status just operates in the local cabal project, not the global database.
$ cabal status
       --compiler # Gives the compiler path of the project
       --targets  # Gives all the targets to be built, present in the ".cabal" file.
       --target <TARGET> # Gives informations about a specific target present in the ".cabal" file
       Future "prisms" into plan.json or .cabal file
```

cli example:
```sh
$ cabal status --compiler --target Cabal:lib:Cabal

Compiler: /home/fendor/.ghcup/bin/ghc-8.10.7
Targets:
  Name: lib:Cabal
  Dependencies:
    Cabal == xxx
    array == xxx
    …
```

One more:
```sh
$ cabal status --compiler --target lib:Cabal --target lib:Foobar

Compiler: /home/fendor/.ghcup/bin/ghc-8.10.7
Targets:
  Name: lib:Cabal
  Dependencies:
    Cabal == xxx
    array == xxx
    …
  Name: lib:Foobar
  Dependencies:
    Cabal == xxx
    array == xxx
    ……
```
 
To be more machine-readable, a json format is also suggested:
```json
{ "targets": [
    "name": "Cabal:lib:Cabal",
    "dependencies": {
      "Cabal": "== xxx",
      "array": "== xxx"
    }
  ]
}
```

This command has a fairly large overlap with `cabal-plan`, if we want to integrate these commands into cabal as well, it might be worth merging `cabal-plan` and `cabal status` in some way.
Features that we would also want that are currently not covered:
* Access to "$PATH" (essentially to implement https://github.com/haskell/hie-bios/issues/125 and fix issues such as https://github.com/haskell/haskell-language-server/issues/176)

Mikolaj:
I can't speak for users (I don't use HLS myself), but +1 for merging with cabal-plan if you go that path.

Kleidukos:
Yes, we will benefit from a rationalisation of the `cabal status` workings, and an integration of cabal-plan's features.

jneira:
> 
> 
> Yes, we will benefit from a rationalisation of the `cabal status` workings, and an integration of cabal-plan's features.

including the actual `list-bins`, which could be deprecated and removed

jneira:
sound like a great plan!

gbaz:
I think we should first proceed on reviewing and merging this, and consider a broader based plan later. @fendor is this ready for review?

fendor:
No it is not, far from it. I didn't proceed with this due to a lack of vision/real plan forward.

gbaz:
What questions do you have outstanding that you think need resolved?

fendor:
I guess for most of this https://github.com/haskell/cabal/pull/7500#issuecomment-941994786 it is still kind of unclear to me, what we should do.

gbaz:
I think that the alternative approaches still require a build while the idea behind this is that it does not, correct? In that case, I think we should start with the original motivation here -- something fast and lightweight that satisfies the needs of hie-bios and ides. We can grow more features optionally and later. Let's start with something that satisfies what we know we need and want.

fendor:
Well, right, but if we only implement what *I* would need, we would have for example:
```
> cabal status --build-info <target> # Only if the target has been built before
<some  json>
> cabal status --project-compiler
<comp info>
```
And it either gives you the build-info or not.
However, my whole issue was, is that acceptable to just add ad-hoc commands just because they fit into my requirements?

jneira:
> However, my whole issue was, is that acceptable to just add ad-hoc commands just because they fit into my requirements?

well, I would say they looks pretty general to serve to other purposes

fendor:
Right, if that is fine, it is fine, and it is exactly what I wanted to discuss. Scope and CLI.

jneira:
I think merged this dont have to be excluyent with converting it in a plugin, it can be done thereafter honouring the cli interface

gbaz:
I would say that if its extensible to add other data as others want, and for the moment the data it gives is ad-hoc based on a particular set of actually desired purposes, that's fine. If other stuff comes up that people want which fits into this framework, it can be added -- and because the groundwork is laid, it will be easier to do so!

cabal-install/main/Main.hs
```diff
@@ -78,6 +78,7 @@ import qualified Distribution.Client.CmdBuild     as CmdBuild
 import qualified Distribution.Client.CmdRepl      as CmdRepl
 import qualified Distribution.Client.CmdFreeze    as CmdFreeze
 import qualified Distribution.Client.CmdHaddock   as CmdHaddock
+import qualified Distribution.Client.CmdIde   as CmdIde
```
Kleidukos, line 81:
```suggestion
import qualified Distribution.Client.CmdHaddock   as CmdHaddock
import qualified Distribution.Client.CmdIde       as CmdIde
```

cabal-install/src/Distribution/Client/CmdIde.hs
```diff
@@ -0,0 +1,126 @@
...
+  { ideTargets    :: Flag Bool
+  , ideProjectGhc :: Flag Bool
+  }
+
+defaultIdeFlags :: IdeFlags
+defaultIdeFlags = IdeFlags
+  { ideTargets    = mempty
+  , ideProjectGhc = mempty
+  }
+
```
Kleidukos, line 81: Extraneous empty line

cabal-install/src/Distribution/Client/CmdIde.hs
```diff
@@ -0,0 +1,126 @@
...
+-------------------------------------------------------------------------------
+-- Flags
+-------------------------------------------------------------------------------
+
+data IdeFlags = IdeFlags
+  { ideTargets    :: Flag Bool
+  , ideProjectGhc :: Flag Bool
+  }
```
Kleidukos, line 74: Could you `deriving stock (Eq, Show)` if possible?

cabal-install/src/Distribution/Client/CmdIde.hs
```diff
@@ -0,0 +1,126 @@
...
+  when (fromFlagOrDefault False (ideTargets ideFlags)) $ do
+    putStrLn "Available Targets:"
+    knownTargets <- getKnownTargets defaultDirActions localPackages
+    for_ (knownComponentsAll knownTargets) $ \KnownComponent{..} -> do
+      let compTarget = ComponentTarget cinfoName WholeComponent
+      putStrLn $ showComponentTarget cinfoPackageId compTarget
+
+    pure ()
+
+
```
Kleidukos, line 122: Extraneous empty lines

cabal-install/src/Distribution/Client/CmdIde.hs
Kleidukos: Any chance you can document the expected result so that it can be undertaken in another PR? :)

Kleidukos (changes requested):

hasufell:
@fendor what is missing here?

fendor:
I have created the following plan for the cli and output format: https://gist.github.com/fendor/796b373aba26e12dbdecf7e40af78b32. If no one fundamentally disagrees, that's what I would like to implement.

Kleidukos:
I agree with this design, and would love to have some tests to confirm that the rules of composition that you offer here can hold. :)

cabal-install/src/Distribution/Client/CmdIde.hs
fendor: This will be put on "maybe later" pile, so ignoring for now.

fendor:
I pushed a major update and changed the PR description. The ideas are ready to review, the actual implementation is too, but tests and documentation is missing.

Kleidukos:
@fendor Could you bring the design document as part of the PR please? :)

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE LambdaCase #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Distribution.Client.CmdIde
```
Kleidukos, line 7: Update the name of the module in the Haddocks

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
```
Kleidukos, line 9: Unsure if this means in terms of OS or in terms of Compiler

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+
+module Distribution.Client.CmdStatus (
+  statusCommand, statusAction,
+  ) where
+
+import qualified Data.Map as Map
+
+import Prelude ()
+import Distribution.Client.Compat.Prelude
+
```
Kleidukos, line 24: Extraneous empty line :)

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+        -- This error should never be shown
+        ("couldn't parse targets: " ++)
+        -- TODO: wrong parser, kills filepaths with spaces
+        (parsecCommaList parsecToken)
+
+-------------------------------------------------------------------------------
+-- Action
+-------------------------------------------------------------------------------
+
+-- | Entry point for the 'ide' command.
```
Kleidukos, line 151: I think you meant "status" instead of 'ide'? :)

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -273,6 +273,7 @@ globalCommand commands = CommandUI {
         , addCmd "hscolour"
         , addCmd "exec"
         , addCmd "list-bin"
+        , addCmd "ide"
```
Kleidukos, line 276: Same as above, I think you meant "status" here. :)

Kleidukos (changes requested):
Good PR, I probably missed some things but here are my main remarks. :)
Please request a review from me again once you have the documentation and tests pushed. :heart:

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+
+  baseCtx <- establishProjectBaseContext verbosity cliConfig OtherCommand
+  (_, elaboratedPlan, elabSharedConfig, _, _) <-
+    rebuildInstallPlan verbosity
+                        (distDirLayout baseCtx)
+                        (cabalDirLayout baseCtx)
+                        (projectConfig baseCtx)
+                        (localPackages baseCtx)
+
+  let initialJson = Json.object
```
hasufell, line 167: Shouldn't we already structure this function in a way that the information is decoupled from the formatting?

Return a big product type and then allow to feed that into the formatter function.

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: It is copy-pasted, I have no idea what it means :D

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+jdisplay :: Pretty a => a -> Json.Value
+jdisplay = Json.String . prettyShow
+
+mergeJsonObjects :: [Json.Value] -> Json.Value
+mergeJsonObjects = Json.object . foldl' go []
+  where
+    go acc (Json.Object objs) =
+      acc <> objs
+    go _   _                  =
+      error "mergeJsonObjects: Only objects can be merged"
```
hasufell, line 367: I suggest to use `MonadThrow` or `MonadFail` here, especially since the caller is in `IO`.

Alternatively, this function could be made total by filtering the input list for Objects, then simply running `mconcat` on them and manually wrapping it into `Value`:

```hs
mergeJsonObjects = Object . mconcat . catMaybes . fmap (\v -> case v of ; Object obj -> Just obj ; _ -> Nothing)
```

cabal-install/src/Distribution/Client/Setup.hs
fendor: Why did I even add it here...

Does this even do anything? Someone needs to clean up the cli. (Or document it)

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+                    Nothing
+                    targetSelectors
+
+      pure $ mkBuildInfoJson (distDirLayout baseCtx) elabSharedConfig
+              elaboratedPlan targets targetSelectors targetStrings
+
+  let statusJson = mergeJsonObjects [initialJson, compilerJson, buildInfoJson]
+
+  -- Final output
+  putStrLn $ withOutputMarker verbosity $ Json.encodeToString statusJson
```
hasufell, line 205: Since  cabal is rather annoying wrt piping output to files (e.g. you have to be careful to suppress other stuff printing to stdout)... having a `--file` option or so may be worthwhile.

cabal-install/src/Distribution/Client/Setup.hs
fendor: ah, I see, it does make sense and is required. Still :D

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: That's basically what the marker is helpful for, as you can filter for the markers and get only the json output.

I am open to adding a file output, but are other parts of the API using file output options as well?

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,367 @@
...
+
+-------------------------------------------------------------------------------
+-- Action
+-------------------------------------------------------------------------------
+
+-- | Entry point for the 'ide' command.
+statusAction :: NixStyleFlags StatusFlags -> [String] -> GlobalFlags -> IO ()
+statusAction flags@NixStyleFlags { extraFlags = statusFlags, ..} cliTargetStrings globalFlags = do
+  when (NoFlag == statusOutputFormat statusFlags) $ do
+    die' verbosity "The status command requires the flag '--output-format'."
```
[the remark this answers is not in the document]
Kleidukos: Good thinking, I agree

fendor:
One more update, now we no longer resolve to the build-info files (since there are situations slightly tricky to resolve) but we resolve each target to the unit-id it resolves to!

We can still obtain the build-info file by parsing `plan.json` externally, but we are more flexible in general and can likely implement more with this primitive than before.

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: Changed, better now?

Kleidukos:
Looks like some dependencies are not needed anymore: https://github.com/haskell/cabal/runs/5265397355?check_suite_focus=true#step:8:1438

fendor:
I doubt this is related to this PR... so I think I am just going to wait for upstream to patch this

cabal-install/src/Distribution/Client/CmdStatus.hs
hasufell: https://mail.haskell.org/pipermail/haskell/2004-March/013836.html

it appears this is an undocumented/useless haddock feature

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+  , commandOptions      = nixStyleOptions statusOptions
+
+  }
+
+-------------------------------------------------------------------------------
+-- Flags
+-------------------------------------------------------------------------------
+
+data StatusOutputFormat
+  = JSON
```
hasufell, line 95: can we document since when a particular format is supported?

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+-- Action
+-------------------------------------------------------------------------------
+
+-- | Entry point for the 'status' command.
+statusAction :: NixStyleFlags StatusFlags -> [String] -> GlobalFlags -> IO ()
+statusAction flags@NixStyleFlags { extraFlags = statusFlags, ..} cliTargetStrings globalFlags = do
+  when (NoFlag == statusOutputFormat statusFlags) $ do
+    die' verbosity "The status command requires the flag '--output-format'."
+  when (not $ null cliTargetStrings) $
+    die' verbosity "The status command takes not target arguments directly. Use appropriate flags to pass in target information."
```
hasufell, line 159:
```suggestion
    die' verbosity "The status command doesn't take target arguments directly. Use appropriate flags to pass in target information."
```

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+data ResolvedTarget = ResolvedTarget
+  { rtOriginalTarget :: String
+  -- | UnitId of the resolved target.
+  -- If 'Nothing', then the given target can not be resolved
+  -- to a target in this project.
+  , rtUnitId :: Maybe UnitId
+  }
+  deriving (Show, Read, Eq, Ord)
+
+serialise :: Verbosity -> Flag StatusOutputFormat -> StatusInformation -> IO String
```
hasufell, line 244: I'd suggest to return `Either String String` or use `MonadThrow` and turn it into `die` at the call site. Serialization shouldn't require IO.

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+          [ "target" .= Json.String (rtOriginalTarget rt)
+          , "unit-id" .= maybe Json.Null jdisplay (rtUnitId rt)
+          ]
+
+
+-- ----------------------------------------------------------------------------
+-- Helpers for determining and serialising compiler information
+-- ----------------------------------------------------------------------------
+
+requireCompilerProg :: Verbosity -> Compiler -> IO Program
```
hasufell, line 279: same here

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+-- Helpers for determining and serialising the unit-id
+-- ----------------------------------------------------------------------------
+
+mkBuildInfoJson :: ElaboratedInstallPlan -> TargetsMap -> Map TargetSelector String -> [String] -> [ResolvedTarget]
+mkBuildInfoJson elaboratedPlan targetsMap tsMap unresolvableTargetStrings =
+  [ ResolvedTarget str (Just uid)
+  | uid <- Map.keys subsetInstallPlan
+  , (_, tss) <- targetsMap Map.! uid
+  , str <- ordNub $ map tsToOriginalTarget $ toList tss
+  ]
```
hasufell, line 302: I've no idea what this does or if it's correct :D

cabal-install/src/Distribution/Client/CmdStatus.hs
```diff
@@ -0,0 +1,381 @@
...
+
+readTargetSelector :: [PackageSpecifier (SourcePackage (PackageLocation a))]
+                    -> Maybe ComponentKindFilter
+                    -> String
+                    -> IO (Either TargetSelectorProblem TargetSelector)
+readTargetSelector pkgs mfilter targetStr =
+  readTargetSelectors pkgs mfilter [targetStr] >>= \case
+    Left [problem] -> pure $ Left problem
+    Right [ts] -> pure $ Right ts
+    _ -> error $ "CmdStatus.readTargetSelector: invariant broken, more than "
```
hasufell, line 373: This should use `fail`

Kleidukos: Do you expect the exception to be caught?

hasufell: I don't know, but `error` doesn't have predictable semantics.

Kleidukos: ah, indeed that's a good reason. Is this explained more in-depth somewhere?

hasufell: > ah, indeed that's a good reason. Is this explained more in-depth somewhere?

Not directly, but https://hackage.haskell.org/package/base-4.16.0.0/docs/Control-Exception.html#v:evaluate explains some of it. Basically, `error` can leak through exception handlers (or never trigger at all), depending on where it's called.

E.g. try:

```hs
case Right (error "lol") of
    Right _ -> 1
     _ -> 2
```

And this:

```hs
-- we force the exceptions
(fmap join $ try @SomeException $ fmap (bimap (SomeException . userError) id)
  $ evaluate $ force $ Right @String (error "lol")) >>= \x -> case x of
    Right r -> putStrLn ("not caught " ++ r)
    Left e  -> putStrLn "caught"

-- not forced
(fmap join $ try @SomeException $ fmap (bimap (SomeException . userError) id)
  $ pure $ Right @String (error "lol")) >>= \x -> case x of
    Right r -> putStrLn ("not caught " ++ r)
    Left e  -> putStrLn "caught"

Mikolaj:
> Looks like some dependencies are not needed anymore: https://github.com/haskell/cabal/runs/5265397355?check_suite_focus=true#step:8:1438

This is indeed unrelated, non-deterministic and described in another ticket (I've forgotten if that's an issue or PR, but I remember it was hard to fix due to being non-deterministic).

Kleidukos:
@Mikolaj Might be worth removing the flag in CI then?

Mikolaj:
> @Mikolaj Might be worth removing the flag in CI then?

The -Werror flag for spurious deps? Sure, but in a way that reminds us to bring it back once this is fixed.

@andreabedini, @jneira: ping, since you are handling the fallout of the CI overhaul and so you may encounter this old failure, too.

andreabedini:
> @andreabedini, @jneira: ping, since you are handling the fallout of the CI overhaul and so you may encounter this old failure, too.

The only fault of the CI overhaul was to start testing with GHC 9.2 :D which is marked as experimental exactly because it's not working yet (work in #7907). The failure with GHC 9.0.2 is #7987.

I am realising the GitHub interface is sometimes unclear. We need to be aware that some tests are marked as "experimental" and that in that case the workflow is considered as successful even if one of the jobs fail.

One picture is worth a thousand words: the green tick at the top left is what matters

image

Mikolaj:
Haha, you are being hounded for all your hard work. But note that I'm not suggesting this is a fault of the CI overhaul, but that you may encounter it when working on CI and then it's good to know the problem is old and not introduced by you.

andreabedini:
> Haha, you are being hounded for all your hard work. But note that I'm not suggesting this is a fault of the CI overhaul, but that you may encounter it when working on CI and then it's good to know the problem is old and not introduced by you.

No offence taken! Also, I have noticed the GitHub UI might have a bug, the PR might get a red mark even if the workflows all pass.

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: Resorted to `MonadError` which should have a similar effect.

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: Good addition!

cabal-install/src/Distribution/Client/CmdStatus.hs
fendor: Tried to improve this piece of code, but it needs some refactoring anyway, due to its usage of `Map.!` which yields sub-par error messages.

Kleidukos (approved):
I think it's okay for me.

fendor:
It is still missing tests and proper documentation. Also, target resolving is buggy.

Kleidukos:
> It is still missing tests and proper documentation. Also, target resolving is buggy.

Need a hand with this? :)

fendor:
TODOS:

- [ ] Write tests in cabal-testsuite
  - [ ] No build-plan 
  - [ ] Some third-party dependency doesn't build
  - [ ] Some local dependency doesn't build
  - [ ] Ignores invalid target selectors
  - [ ] Disabled units
  - [ ] Unit fails to build
  - [ ] GHCJS, what's the output?
  - [ ] Check plan.json is generated
  - [ ] Bench unit
  - [ ] Exe unit
  - [ ] Flib unit 
  - [ ] Lib unit
  - [ ] Test unit
- [ ] Documentation
  - [ ] Write Command docs
  - [ ] Define json schema
  - [ ] Explicitly mark as experimental
  - [ ] Show some use-cases
  - [ ] Future features
    -  [ ] Cabal-plan why-depends

newhoggy:
Can this also print the location of the cabal store?

hasufell:
Printing configuration values should be part of `cabal user-config` command. It's true you can overwrite that store dir via cli (maybe even in project files), but that's a non-issue, because it means the user usually knows the value and doesn't need to query it.

fendor:
Closing as discussed in https://hackmd.io/mQjghm2zRgWgcyD4XKhBqA

A new PR will follow soonish, depending on the holidays.

PR #7507, opened 2021-08-04
Add compilerLocation to Compiler datatype

fendor:
Avoids compiler program lookup logic code in `ShowBuildInfo`.

Split off from #7498 based on the comment https://github.com/haskell/cabal/pull/7498#discussion_r682513126

### DIFF
--- a/Cabal/src/Distribution/Simple/Compiler.hs
+++ b/Cabal/src/Distribution/Simple/Compiler.hs
@@ -80,6 +80,7 @@ import Distribution.Compiler
 import Distribution.Version
 import Language.Haskell.Extension
 import Distribution.Simple.Utils
+import Distribution.Simple.Program.Types (ProgramLocation)
 
 import qualified Data.Map as Map (lookup)
 import System.Directory (canonicalizePath)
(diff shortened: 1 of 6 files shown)


DanielG:
My gut feeling is that this is a bad idea. If you insist on going this way you should replace all the use sites of `lookupProgram ghcProgram` and similar with this mechanism instead and see if it still makes sense.

fendor:
I do not insist, I thought it might be a smooth solution, but if it isn't, I feel free to abandon it again

DanielG:
I didn't mean to imply I don't think it's worth pursuing, I just think just adding this hardly tells us if this is a good idea or not. In my mind the ProgramDB is there for a reason and I'm not sure if moving some of the info it hast elsewhere wont lead to more problems than it solves. Showing how this would replace a lot of `error` prone (hehe) code would certainly make a stronger case for this approach.

fendor:
I don't think a single invocation of `requireProgram` or `lookupProgram` can be saved, as other data is usually required than just the program location. E.g. to invoke a program, you still need its `ConfiguredProgram`.

fendor:
This is probably not very good, so closed

PR #7519, opened 2021-08-06
solver: swap catas with recursion, fuse travs

gbaz:
This is the last refactor I had in mind to address the general performance issues discussed in https://github.com/haskell/cabal/issues/7466

It attains some modest but real improvement (mainly in allocations) from eliminating the intermediate `cata` calls in favor of explicit recursion in the solver tree.

For `trav` calls, rather than eliminate them entirely, it lifts out their intermediate functions to the top level and explicitly fuses them in the solver pipeline, which yields somewhat more significant improvements.

Overall not the huge win I hoped, but nonetheless given the straightforward nature of the changes to the code, still worth it I think.

### DIFF
--- a/cabal-install-solver/src/Distribution/Solver/Modular/Validate.hs
+++ b/cabal-install-solver/src/Distribution/Solver/Modular/Validate.hs
@@ -164,12 +164,12 @@ type VROrigin = (VR, ExposedComponent, DependencyReason QPN)
 type Conflict = (ConflictSet, FailReason)
 
 validate :: Tree d c -> Validate (Tree d c)
-validate = cata go
+validate = go
   where
-    go :: TreeF d c (Validate (Tree d c)) -> Validate (Tree d c)
+    go :: Tree d c -> Validate (Tree d c)
 
-    go (PChoiceF qpn rdm gr       ts) = PChoice qpn rdm gr <$> sequence (W.mapWithKey (goP qpn) ts)
-    go (FChoiceF qfn rdm gr b m d ts) =
+    go (PChoice qpn rdm gr       ts) = PChoice qpn rdm gr <$> sequence (W.mapWithKey (goP qpn) (fmap go ts))
+    go (FChoice qfn rdm gr b m d ts) =
       do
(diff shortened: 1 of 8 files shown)


cabal-install-solver/src/Distribution/Solver/Modular/Validate.hs
```diff
@@ -164,12 +164,12 @@ type VROrigin = (VR, ExposedComponent, DependencyReason QPN)
...
 validate :: Tree d c -> Validate (Tree d c)
-validate = cata go
+validate = go
   where
-    go :: TreeF d c (Validate (Tree d c)) -> Validate (Tree d c)
+    go :: Tree d c -> Validate (Tree d c)
 
-    go (PChoiceF qpn rdm gr       ts) = PChoice qpn rdm gr <$> sequence (W.mapWithKey (goP qpn) ts)
-    go (FChoiceF qfn rdm gr b m d ts) =
+    go (PChoice qpn rdm gr       ts) = PChoice qpn rdm gr <$> sequence (W.mapWithKey (goP qpn) (fmap go ts))
```
phadej, line 171: sequence . map seems bad idea. Should we add `traverseWithKey` to `WeightedPSQ` (separately)? Maybe it fuses, but  I'm somewhat sceptical about that.

gbaz: Sure. I take it you don't consider that a blocker for this PR, as its about changing a different thing than this PR currently does. Are you ok with approving this pr as is?

emilypi (approved):

cabal-install-solver/src/Distribution/Solver/Modular/Validate.hs
emilypi: I can fix it up with `traverseWithKey` in the follow up #7533

emilypi: Added it first, then pushed to this branch.

emilypi:
https://github.com/haskell/cabal/pull/7519

emilypi:
@mergifyio backport 3.6

PR #7532, opened 2021-08-11
defer build-tools-depends choices as well as setup choices

### LINKED ISSUE #7472: Slow solving with `build-tool-depends` on local packages with tests

gbaz:
This is effectively a one line change, modulo renaming, that seems to fully resolve #7472 giving a speedup of roughly 80% on the tricky case detailed there (cardano-node). Arguably this was the single key change to resolve the issue all along (although the other fixes resulted in improvements along the way anyway). All it does is extend the existing solver pass that defers solving setup depends until top-level goals are solved to _also_ defer build-tool-depends goals until top level goals are solved. This means that all the constraints at the top level (on versions, flags, stanzas, etc) are applied _first_ and only then are attempts made to build the less-constrained qualified goals, significantly streamlining the production of valid build plans.

This change is in line with how setup depends are already constrained, so its simply extending an already tested and correct strategy to another instance of the same problem. Also, since it just reorders goals, it doesn't change the solution space. As such, hopefully it can be reviewed and merged quickly.

### DIFF
--- a/cabal-install-solver/src/Distribution/Solver/Modular/Preference.hs
+++ b/cabal-install-solver/src/Distribution/Solver/Modular/Preference.hs
@@ -2,7 +2,7 @@
 -- | Reordering or pruning the tree in order to prefer or make certain choices.
 module Distribution.Solver.Modular.Preference
     ( avoidReinstalls
-    , deferSetupChoices
+    , deferSetupExeChoices
     , deferWeakFlagChoices
     , enforceManualFlags
     , enforcePackageConstraints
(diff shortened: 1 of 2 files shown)


gbaz:
whoops, accidentally pushed directly to master. i'm confident enough that i'm not going to revert, but just link the patch for posterity. sorry about bypassing the process.

https://github.com/haskell/cabal/commit/aae998e4a5603e162d85294eb6810c9fd29a460b

Mikolaj:
Wow, congratulations. This is 80% of the real world case, not 80% of the reproduction example for the issue?

michaelpj:
Amazing!

newhoggy:
Really excited for this.  Thanks so much!

emilypi:
@mergifyio backport 3.6

PR #6935, opened 2020-06-30
pkg-config: Use --libs --static when building with using --enable-executable-static

alexbiehl:
---
Please include the following checklist in your PR:

* [X] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [X] Any changes that could be relevant to users have been recorded in the changelog (add file to `changelog.d` directory).
* [X] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

PR #7536, opened 2021-08-12
Alex/library dirs static and extra libraries static

alexbiehl:
This PR is the continuation of https://github.com/haskell/cabal/pull/6935, in particular the next step in the plan laid out by https://github.com/nh2/static-haskell-nix/issues/62#issuecomment-552051386. 

The commits in this PR were kindly developed by @nh2, I just rebased and adjusted here and there for recent Cabal developments. 

We need to see how much adjustment we need for the testsuite to pass. Once that's done the I would prepare a patch for GHC honouring the new fields for static linking.

---


alexbiehl:
@Mikolaj I think this is ready for review!

Mikolaj:
Splendid. Am I right guessing this is supposed to help in situations like this:

> cabal new-build --enable-executable-static --ghc-options="-optl-lfreetype -optl-lpng -optl-lbz2 -optl-lbrotlidec-static -optl-lbrotlicommon-static"

which is what I needed to write in order to link SDL application statically? (and failed, see see https://github.com/utdemir/ghc-musl/issues/14#issuecomment-887856085)

emilypi (approved):
This is really complete. Thanks @alexbiehl

emilypi:
Also, do you mind adding a changelog entry?

alexbiehl:
> Also, do you mind adding a changelog entry?

@emilypi Done!

alexbiehl:
@Mikolaj That's exactly right. For now this PR doesn't do very much. Once merged, I will update Cabal for GHC, add a new flag to create static exes where I will make use of the newly added field in the InstalledPackageInfo. Only then we will have full end-to-end support for more convenient static linking. We are so close!!

Mikolaj:
@alexbiehl: I think @emilypi meant something like that: https://github.com/Mikolaj/cabal/blob/91195e844e44f585298c0cb78806103904c4f9bb/changelog.d/pr-7510

But perhaps your way of adding the changelog is fine, too, can't tell.

Mikolaj (approved):
Great job. Looking forward to GHC changes.

hasufell:
@alexbiehl what is the reasoning for this change?

Right now, we have:

* `library-dirs`
* `library-dirs-static`
* `dynamic-library-dirs`

This makes no sense to me. There is only two types: dynamic and static. But we have 3 directories/options.

mouse07410:
There are two types, but three states:

- Explicitly static (no dynamic);
- Explicitly dynamic - no static-related stuff here;
- Don't care and don't specify - let it be whatever system default is. 

Eliminating the 3rd one would force the user to always have to care and specify everything libraries-related explicitly. That would not be good. Plus, unpredictable effect on packages that should run on different platforms that have different defaults or properties.

hasufell:
@mouse07410 

What you're describing is not how GHC or cabal works.

GHC has no knowledge internally of whether you specified `-static` or you didn't (it is the default). There are no three states, there are only two:

https://github.com/stable-haskell/ghc/blob/a2d9d7c2073867ee0cabb8d49f93246d95ec0b09/compiler/GHC/Linker/Unit.hs#L58-L62

This is the same for cabal, there are no three states for `--enable-executable-static`. Either the flag is specified or it isn't.

https://github.com/haskell/cabal/blob/1e3c355eeab89fa8894ce9c6e4e649903e58d6b3/Cabal/src/Distribution/Simple/GHC/Build/Link.hs#L138-L142

What you are proposing is **bad practice** I would argue. CLI tools should not behave differently when someone invokes: `foo --bar` or `foo` while `--bar` is the default. If you have three different behaviors, you want 3 different cli switches and make one of them the default.

No one understands how this is supposed to behave and Cabal and GHC randomly mix those choices.

----

Edit: it's also not clear what *technical* purpose this solves. Why would there be 3 different directories? And what is inside of them? The directories themselves don't determine how linking is done. We could remove `library-dirs-static` and `dynamic-library-dirs` without losing any functionality. The convention has always been to place dynamic and static libraries next to each other. You can tell the linker whether to use a static library or a dynamic one. Playing tricks with directory search paths is not good practice.

mouse07410:
Are you talking about the input that user needs to provide? Or about what the GHC internals need to do for different kinds of libraries?

hasufell:
@mouse07410 this PR adds `library-dirs-static` to the package database format.

PR #7558, opened 2021-08-18
`cabal init` common stanzas and default flags

### LINKED ISSUE #7465: Cabal init: Init should add warning flags such as `-Wall`
This might be controversial, but I think people benefit from most warnings in `-Wall`, but a lot of beginners actually don't have this flag enabled, as it is not trivial to discover.

I propose that `cabal init` adds a set of default warnings to each stanza, and if cabal version > 3.0 maybe even with common stanzas, showing off another nice feature of cabal.

While I can see that some warnings are debatable, I think it is better for newcomers to have these warnings added automatically.

Such additions might be controversial as we impose our (maybe only mine?) opinions on other developers, but I argue that this benefits beginners while not impeding more advanced users.

ptkato:
This PR aims to add common stanzas to the `cabal init` command and possible default flags that could go with it. This PR will close #7465.

---

### DIFF
--- a/cabal-install/src/Distribution/Client/Init/Format.hs
+++ b/cabal-install/src/Distribution/Client/Init/Format.hs
@@ -20,6 +20,7 @@ module Distribution.Client.Init.Format
 , annNoComments
 , postProcessFieldLines
   -- * stanza generation
+, mkCommonStanza
 , mkLibStanza
 , mkExeStanza
 , mkTestStanza
@@ -111,10 +112,28 @@ postProcessFieldLines ann
 -- -------------------------------------------------------------------- --
 -- Stanzas
 
+-- The common stanzas are hardcoded for simplicity purposes,
+-- see https://github.com/haskell/cabal/pull/7558#discussion_r693173846
+mkCommonStanza :: WriteOpts -> PrettyField FieldAnnotation
+mkCommonStanza opts = case specHasCommonStanzas $ _optCabalSpec opts of
+  NoCommonStanzas -> PrettyEmpty
+  _ -> PrettySection 
+    annNoComments
+    "common"
+    [text "warnings"]
+    [field "ghc-options" text "-Wall" [] False opts]
+
 mkLibStanza :: WriteOpts -> LibTarget -> PrettyField FieldAnnotation
 mkLibStanza opts (LibTarget srcDirs lang expMods otherMods exts deps tools) =
   PrettySection annNoComments (toUTF8BS "library") []
-    [ field "exposed-modules" formatExposedModules (toList expMods)
+    [ case specHasCommonStanzas $ _optCabalSpec opts of
+        NoCommonStanzas -> PrettyEmpty
+        _ -> field "import" (hsep . map text) ["warnings"]
+          ["Import common warning flags."]
+          False 
+          opts
+
+    , field "exposed-modules" formatExposedModules (toList expMods)
       ["Modules exported by the library."]
       True
       opts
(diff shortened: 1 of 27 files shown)


cabal-install/src/Distribution/Client/Init/Types.hs
```diff
@@ -186,6 +191,20 @@ data TestTarget = TestTarget
...
+    { _commonName   :: String
+    , _commonFields :: [CommonField]
+    } deriving (Show, Eq)
+
+data CommonField = CommonField
+    { _commonFieldName    :: String
+    , _commonFieldValue   :: String
+    , _commonFieldComment :: [String]
+    } deriving (Show, Eq)
+
```
ptkato, line 207: @emilypi what are your thoughts? While this makes using multiple, separate, common stanzas trivial within `cabal init`, I'm not sure we really have the need for it.

emilypi: I don't think we need it even if it's trivial, but if you want to add in `-Wall` support and showcase them, be my guess.

ptkato: Yeah... I've been pondering, it adds complexity to the codebase; even though it's somewhat minimal right now, every little change/tuning/addition sums up and soon enough we'll be with a happy mess in our hands. The future proofing and flexibility that this change would provide (regarding common stanzas) is niftily nice, but I don't think it's worth the extra code that it incurs throughout the cabal init codebase for what it brings to the table. I'll likely do as you suggest and streamline the warnings' common stanzas to a hardcoded approach.

emilypi: Sounds good - :+1: 

Yeah, what i was thinking about was that in order to make this worthwhile, you'd probably also want to throw the base dep in there, but then the process of aggregating dependencies becomes a bit more tiresome, and you end up having to fetch the base dep early on, and then pass it through to all of the targets as a parameter, or you have to "fixup" the dependencies in the `createProject` function like we do when we add the library dep to the exe and test suite targets. It starts to become more tangled, and not for any tangible benefit aside from the showcase.

ptkato:
Hardcoding the common stanzas in there made things much much simpler, however each individual stanza test look a bit wonky, considering they now got a `import: warnings` field/value with no stanza in sight.

cabal-install/src/Distribution/Client/Init/Format.hs
```diff
@@ -111,10 +112,28 @@ postProcessFieldLines ann
...
+-- see https://github.com/haskell/cabal/pull/7558#discussion_r693173846
+mkCommonStanza :: WriteOpts -> PrettyField FieldAnnotation
+mkCommonStanza opts = case specHasCommonStanzas $ _optCabalSpec opts of
+  NoCommonStanzas -> PrettyEmpty
+  _ -> PrettySection 
+    annNoComments
+    "common"
+    [text "warnings"]
+    [field "ghc-options" text "-Wall" [] False opts]
+
```
emilypi, line 125: Hey look, `PrettyEmpty` paying dividends :)

cabal-install/src/Distribution/Client/Init/Format.hs
```diff
@@ -194,7 +220,14 @@ mkExeStanza opts (ExeTarget exeMain appDirs lang otherMods exts deps tools) =
 mkTestStanza :: WriteOpts -> TestTarget -> PrettyField FieldAnnotation
 mkTestStanza opts (TestTarget testMain dirs lang otherMods exts deps tools) =
     PrettySection annNoComments (toUTF8BS "test-suite") [suiteName]
-       [ field "default-language" id lang
+       [ case specHasCommonStanzas $ _optCabalSpec opts of
+           NoCommonStanzas -> PrettyEmpty
+           _ -> field "import" (hsep . map text) ["warnings"]
+             ["Common stanzas."]
```
emilypi, line 226: This comment isn't super useful. I'm suggesting the following so that it makes sense: 

```
Import common warning flags
```

emilypi (approved):

cabal-install/src/Distribution/Client/Init/Format.hs
ptkato: Done.

Mikolaj:
Actually, some of the CI errors are about `cabal init`.

gbaz (approved):

PR #7492, opened 2021-07-24
some type annotations

peterbecich:
I am investigating a possible issue in `FileMonitor`. ~This PR adds a few type annotations. It also replaces `$` with `.` in some places.~

I have only tested the change by building Cabal, and the Github Actions.

---
Please include the following checklist in your PR:

* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ ] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog).
* [ ] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

PR #7564, opened 2021-08-21
more type annotations

peterbecich:
---

Similar to https://github.com/haskell/cabal/pull/7492

Only tested by building `cabal-install` and CI.

### DIFF
--- a/cabal-install/src/Distribution/Client/Dependency.hs
+++ b/cabal-install/src/Distribution/Client/Dependency.hs
@@ -13,8 +13,8 @@
 -- Top level interface to dependency resolution.
 -----------------------------------------------------------------------------
 module Distribution.Client.Dependency (
-    DepResolverParams,
     -- * The main package dependency resolver
+    DepResolverParams,
     chooseSolver,
     resolveDependencies,
     Progress(..),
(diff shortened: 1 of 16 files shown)


cabal-install/src/Distribution/Client/Dependency.hs
```diff
@@ -13,6 +13,7 @@
 -- Top level interface to dependency resolution.
 -----------------------------------------------------------------------------
 module Distribution.Client.Dependency (
+    DepResolverParams,
```
peterbecich, line 16: Which section should this fall under?

Mikolaj: I bet @grayjay would know...

grayjay: The top section looks like a good match to me, since DepResolverParams is one of the top level types for running the dependency solver.

peterbecich: done

Mikolaj (approved):
Great job. Countless future generations of cabal hackers will benefit.

peterbecich:
Thank you!

grayjay (approved):

jneira (approved):
agree with @Mikolaj, thank you very much

PR #7567, opened 2021-08-23
handle build reports in v2-build

### LINKED ISSUE #6851: BuildReports don't seem to work with v2-build
A more focused follow-up to #6811

After some digging, it appears that the --build-summary (and --build-log?) flags when passed to new-build are no-ops. I looked through the code, and it seems to me the issue is that this info goes into the `installSummaryFile` config field, but the only place that (or the `BuildReports`) module is used is in `Distribution.Client.Install` such as here:

https://github.com/haskell/cabal/blob/06c3eff242154cdf6b3d1bbaffcfc1a5cbea9da8/cabal-install/Distribution/Client/Install.hs#L824

So v1-build didn't do this, only v1-install. Furthermore, v2-build, if I understand the system (and by implication v2-install) won't hit that path either.

It looks like we'd need to implement similar logic in `ProjectOrchestration.hs` if we want to make use of this? (https://github.com/haskell/cabal/blob/10f33badad67a471d8635e288d24fa495db7d504/cabal-install/Distribution/Client/ProjectOrchestration.hs)

Note that this is actually pretty important for GSoC work, because we can't move the docbuilder to v2-build if we don't have build reports as part of it.

Can someone confirm that this analysis makes sense?

gbaz:
Resolves #6851 -- or rather resolves it enough for the purposes of hackage builder, which is likely the only thing that ever cared about these flags.

In particular makes the `--report-planning-failure` and `--build-summary` flags do anything at all. Note that this preserves the v1 behavior that by default, build-summary information for all builds gets written to ~/.cabal/logs/build.log (which had been lost with v2-build before). The former flag writes build-summary info even if the solver fails (otherwise such info is written only if a build occurs). The latter controls where to write summary info to.

Some notes on what this doesn't do:

This does _not_ provide the still missing "remote build reports" (enabled by `--remote-build-reporting`) since hackage builder doesn't use them, and nobody else seems to use or miss them either. At this point I think we should probably remove the flag from v2 and just strip the functionality entirely (saving code) when v1 is removed (if not sooner). For background context, the idea of anonymous/remote build reports was that instead of a central docbuilder, cabals could be taught to tarball up all the information from their builds so they could automatically upload it and "report to a central server" any successes or failures, optionally with logs, etc. It was an interesting idea, but never fully implemented. The existing machinery is not very fleshed out on the tricky bits (i.e. actually uploading the reports, controlling policy well, etc.) and there was never a clear policy worked out nor any upstream support. As such, I think we can strip out the code for this from the codebase and if we embark on this in the future, start fresh with a clear architecture and design first, and then write new code to suit that.

Further, this does _not_ write detailed logs for `--report-planning-failure` but rather just the summary log. It seemed an extra bit of unnecessary functionality, given that hackage builder doesn't use these logs, and as far as I can tell nobody else did, who might miss them. This means that as far as I can see, the `--build-log` argument to v2-build never gets used (though I may be mistaken here). It would be future work to confirm these flags are unused at the moment, and potentially to clean them up. Certainly something is still writing logs to `~/.cabal/logs/ghc-x.x.x/pkg-name-hash.log` but I'm not sure of that codepath.

### DIFF
--- a/cabal-install/src/Distribution/Client/BuildReports/Anonymous.hs
+++ b/cabal-install/src/Distribution/Client/BuildReports/Anonymous.hs
@@ -25,6 +25,7 @@ module Distribution.Client.BuildReports.Anonymous (
     parseBuildReport,
     parseBuildReportList,
     showBuildReport,
+    cabalInstallID
 --    showList,
   ) where
 
--- a/cabal-install/src/Distribution/Client/ProjectOrchestration.hs
+++ b/cabal-install/src/Distribution/Client/ProjectOrchestration.hs
@@ -121,7 +121,10 @@ import           Distribution.Client.Types
                    ( GenericReadyPackage(..), UnresolvedSourcePackage
                    , PackageSpecifier(..)
                    , SourcePackageDb(..)
-                   , WriteGhcEnvironmentFilesPolicy(..) )
+                   , WriteGhcEnvironmentFilesPolicy(..)
+                   , PackageLocation(..)
+                   , DocsResult(..)
+                   , TestsResult(..) )
 import           Distribution.Solver.Types.PackageIndex
                    ( lookupPackageName )
 import qualified Distribution.Client.InstallPlan as InstallPlan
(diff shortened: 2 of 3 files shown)


Mikolaj:
I'm all for removing `--remote-build-reporting` ASAP (but not backporting the removal to 3.6) to given people the theoretical chance to yell at us if anybody uses it (and uses master branch, both unlikely but that's the best we can do with low effort).

Mikolaj (approved):
LGTM.

Is the new code copied from the v1- version? From where exactly?

gbaz:
The standard buildreport code is from here in v1, but I inlined the report generation since the types changed enough I couldn't make use of the existing functions:

https://github.com/haskell/cabal/blob/master/cabal-install/src/Distribution/Client/Install.hs#L821

The planning failure code is from here:

https://github.com/haskell/cabal/blob/master/cabal-install/src/Distribution/Client/Install.hs#L728

Let's get this merged, then I'll make another PR cleaning up remote build reporting (removing it from v1 as well as removing the useless flags from v2)

Mikolaj:
We need one more review, pretty please. :)

emilypi (approved):

PR #7572, opened 2021-08-23
Add file perm bits to current mode rather than overriding

### LINKED ISSUE #7560: Cabal does not respect the setgid bit when creating directories
When using `cabal install` to install packages on a Linux machine it creates quite a number of directories while building and installing components. It works pretty well until you want to use `CABAL_DIR` (or `dist-newstyle` directory) with a set `setgid` bit. In this case some of the directories created by cabal do not inherit the `setgid` bit of the parent directory (as in the following example):

```
$ export CABAL_DIR=/scratch/cabal
$ cabal update
Config file path source is default config file.
Config file /scratch/cabal/config not found.
Writing default configuration to
/scratch/cabal/config
$ cabal install zlib
# ... Output omitted ...
$
```
which gives the following under `/scratch/cabal`:
```
$ ll /scratch/cabal
total 28
drwxrwsr-x 5 user grp 4096 Aug 20 14:57 ./
drwxrwsr-x 7 user grp 4096 Aug 20 14:54 ../
-rw-rw-r-- 1 user grp 5508 Aug 20 14:55 config
drwxrwsr-x 3 user grp 4096 Aug 20 14:57 logs/
drwxrwsr-x 3 user grp 4096 Aug 20 14:55 packages/
drwxr-xr-x 3 user grp 4096 Aug 20 14:57 store/ # <== No setgid bit
$ ll /scratch/cabal/store
drwxr-xr-x 3 user grp  4096 Aug 20 14:57 ./
drwxrwsr-x 5 user grp  4096 Aug 20 14:57 ../
drwxr-xr-x 4 user user 4096 Aug 20 14:58 ghc-8.10.6/
$
```
where the `store` directory is created without the `setgid` bit. Thus, the files and directories under `store` have a different group, which could be rather bothersome when user's primary group has a strict disk quota enforced.

It seems that this behavior is related to the definition of the `createDirectoryVerbose` function which calls `setDirOrdinary` after a directory is created. In turn, the `setDirOrdinary` function calls `c_chmod` to set the permissions of the directory in question to `0o755` apparently clearing the `setgid` bit.

It would be really nice to have cabal to be able to recognize this kind of situation and set (or keep) the `setgid` bit accordingly. Or maybe it's already possible?

DanielG:
Fixes #7560

---

### DIFF
--- a/Cabal/src/Distribution/Compat/CopyFile.hs
+++ b/Cabal/src/Distribution/Compat/CopyFile.hs
@@ -20,6 +20,8 @@ import Distribution.Compat.Internal.TempFile
 import Control.Exception
          ( bracketOnError )
 import qualified Data.ByteString.Lazy as BSL
+import Data.Bits
+         ( (.|.) )
 import System.IO.Error
          ( ioeSetLocation )
 import System.Directory
@@ -35,7 +37,9 @@ import Foreign
 import System.Posix.Types
          ( FileMode )
 import System.Posix.Internals
-         ( c_chmod, withFilePath )
+         ( withFilePath
+         , c_chmod
+         , c_stat, sizeof_stat, st_mode )
 import Foreign.C
          ( throwErrnoPathIfMinus1_ )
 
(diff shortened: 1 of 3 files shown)


DanielG:
This should be ready for review now.

jneira (approved):
lgtm, although there are no tests, the pr has been tested manually by the issue reporter so it is ok for me (out of curiosity, would be hard to add those tests?)

DanielG:
> lout of curiosity, would be hard to add those tests?

Setting any of the upper bits including setgid needs root, so we'd have to setup something using `fakeroot` if we really wanted to test this :)

jneira:
@emilypi @fgaz do you think it could be merged?

emilypi (approved):
LGTM, though, can we get a changelog @DanielG?

fgaz:
CI got stuck again... let's see if this time it starts

Mikolaj:
I'm afraid it's stuck again and blocks the queue.

Mikolaj:
@Mergifyio refresh

Mikolaj:
@Mergifyio refresh

PR #7592, opened 2021-08-28
ci(Mergify): use rebase+merge queue

fgaz:
This change has been made by @fgaz from https://mergify.io config editor.

---

actions/merge/strict is deprecated, so I switched it to merge queues.

I also changed the merge method from squash to rebase+merge, for a clean history (provided prs are clean too), which _I think_ was the original intent behind the introduction of this bot. Opinions? Maybe there should be both this and a "squash+merge me" label for squashable atomic prs so that we don't have to ask the author to do it when it's necessary?
edit: added the squash strategy too

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,9 +1,29 @@
 pull_request_rules:
+  # rebase+merge strategy
   - actions:
-      merge:
-        strict: smart+fasttrack
-        method: squash
-    name: Automatically merge pull requests
+      queue:
+        name: default
+        # Merge into master with a merge commit
+        method: merge
+        # Update the pr branch with rebase, so the history is clean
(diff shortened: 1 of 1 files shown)


Mikolaj:
Huh, apparently CI failed, but no individual check is red so I could see the failure log...

fgaz:
Apparently one of them was canceled:

> 24 successful and 1 cancelled checks 

> MacOS / validate.sh macos ghc-8.6.5 (pull_request) Cancelled after 38m

Mikolaj (approved):
Judging by the description, makes sense.

emilypi (approved):

fgaz:
Great! I'm force-merging since the failure is spurious

PR #7594, opened 2021-08-29
Fix #7331: prefix fatal errors with "Error:"

### LINKED ISSUE #7331: Missing file notification from cabal-install doesn't specify if it's a warning or error.
**Describe the bug**

When running cabal install:
```
$ cabal install
cabal.exe: filepath wildcard 'ChangeLog.md' does not match any files.
```

The output: cabal.exe: filepath wildcard 'ChangeLog.md' does not match any files.
does not give a clear indication if this is an error or an warning. Nor if the executable was installed or not.

**Expected behavior**
Since it seems this prevents installation of the package the string should be prefixed with "error:"

andreasabel:
- [x] refactoring passed CI
  Re #7331: factor error message decoration in variants of `die`

  The existing code had quite some duplication, which makes it more brittle and harder to understand.

  I factored out atomic modifications of the error message that will eventually be raised by `die` and friends.  These factors can now be combined in a pipeline.

  The factors that require `IO` are:

  - prefix with program name (e.g. `cabal`)
  - annotate with time stamp and meta-data

  The lens `ioeErrorString` and its pure compagnion `ioeModifyErrorString` should really be in `base:System.IO.Error`, but they aren't, so I define them here.

- [x] Insert the word "Error:" automatically when producing a fatal error.
- [x] Fix golden values.
- [x] Get approval.
- [x] Reorganize commits.
- [ ] CHANGELOG
- [ ] Check whether documentation needs to be updated.

---


andreasabel:
@fgaz: From what we discussed, I removed "Error:" from many individual places as `die` etc. prefix the error message with it if it is not present already (see `addErrorPrefix`).  The backpack error reporting follows a slightly different style with `Doc`-formatted errors.  There, removing `Error:` was not possible without destroying the layout.  Thus, I kept the heuristics in `addErrorPrefix` to only add "Error:" if it is not already at the front.

emilypi (approved):

Mikolaj:
Don't things like

    cabal-install/src/Distribution/Client/Freeze.hs:        die' verbosity $ "internal error: 'resolveUserTargets' returned "
    cabal-install/src/Distribution/Client/Freeze.hs:        die' verbosity $ "internal error: 'resolveUserTargets' returned "

now result in "Error: internal error: seriously, error"? Got it from `git grep "error:"` on branch master.

Mikolaj (approved):
LGTM.

andreasabel:
Squashed commits into just two:
1. Refactoring.
2. Added feature and updated golded values.

Waiting for CI...

Mikolaj:
@andreasabel: and what about ""Error: internal error"? Cannot happen?

andreasabel:
> @andreasabel: and what about ""Error: internal error"? Cannot happen?

Ah, sorry, I thought you accepted anyway.

I searched the testsuite for output mentioning "error" in a message twice but could not find any.  Should a testcase in the future display such behavior, it should be fixed; the testcase will lead to the relevant code snippet.

I don't think internal errors will enter the testsuite because these usually point to bugs; at least this is how I understand the concept of _internal error_.  
(And if the user is exposed to an internal error, it is anyway too late for cosmetics; the image is ruined already...)

So, I'd argue there isn't anything urgent to be done right now.

Mikolaj:
Thank you. Given what you say, "internal error" probably amounts to "impossible error", so it's fine to ignore what is impossible. :)

Issue #7469, opened 2021-06-29
Dead link in "data-files:" documentation.


danidiaz:
The [Accessing data files from package code](https://cabal.readthedocs.io/en/latest/cabal-package.html#accessing-data-files-from-package-code) section of the user guide has a dead link to a non-existent section about "prefix independence":

>  (see prefix independence)

Googling for it, it seems that section [formerly belonged](https://downloads.haskell.org/~ghc/7.0.4/docs/html/Cabal/builders.html#prefix-independence) to the GHC User Guide—but it isn't present in the current GHC User Guide.

andreasabel:
No, the section "prefix independence" is still present in the Cabal docs: https://github.com/haskell/cabal/blob/7d4ce4787cb3d4fe30d08953ad64d19456a77690/doc/setup-commands.rst#prefix-independence

Only the link broke, maybe by 3752c3e5edc3c19b7262cdb0b70f4e0539b46362 which moved the text from `installing-packages.rst` to `setup-commands.rst`
bb16112746724ccef9686d407ec94c7ccf7f928a fixed some links, but must have missed this one.

andreasabel:
Fix in PR #7586.

andreasabel:
I took away the "wierd bug award", as broken references are easy to creep in.  
Broken refs can maybe be avoided by passing `-n` ("nitpick") to `sphinx-build`, see https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-n.  E.g., when I run with `-n`, this sleeping bug surfaces:
```
doc/cabal-project.rst:644: WARNING: 'envvar' reference target not found: PATH
```
I'll prepare a new PR that fixes this.

andreasabel:
> I'll prepare a new PR that fixes this.

PR #7603.

Mikolaj:
Good call, thank you. Broken links are indeed not that weird, as opposed to linking away to an imaginary section.

PR #7603, opened 2021-08-31
Docs: make Sphinx stricter

andreasabel:
Added flags to sphinx-build to make it stricter:
- `-n`: more warnings
- `-W`: warnings to errors
- `-E`: rebuild

This will hopefully prevent errors creeping in to the docs (e.g. broken references such as #7469).

Thanks to these flags, I fixed two issues (see additional commits).


Mikolaj (approved):
Not nearly enough experience on my part, but LGTM. Citing from IRC:

> somebody asked (here? on gibhub) if we build docs as part of CI; we don't, right? should we? I guess we should, unless it takes forever?

Mikolaj:
OK, got the question and the answer, so we now need a PR: https://github.com/haskell/cabal/pull/7586#issuecomment-907874571

andreasabel:
@fgaz wrote:
> I think we shouldn't have `-W` in the makefile, only in CI (like it's done for `-Werror` for haskell code).

Is the Makefile shipped?  I thought it is developers only.

> So in the makefile:
> 
> ```makefile
> SPHINX_FLAGS += -n --keep-going -E
> ```

Can I then still override the flags with `make SPHINX_FLAGS= users-guide`?

fgaz:
> Is the Makefile shipped? I thought it is developers only.

that's right, but I still think it'd be nicer to not fail on warnings when writing docs so that one can still preview them while fixing the warnings (like for haskell). 

> Can I then still override the flags with make SPHINX_FLAGS= users-guide?

yes. from what I understand `make users-guide SPHINX_FLAGS=...` overrides, while `SPHINX_FLAGS=... make users-guide` makes the `+=` append

fgaz (approved):
By the way that above is just a weak preference so I'll just give an approve. It can also be done later when someone adds CI for the guide

emilypi (approved):

andreasabel:
@fgaz
> but I still think it'd be nicer to not fail on warnings when writing docs so that one can still preview them while fixing the warnings

If it turns out that `-W` is a nuisance to developers, it can be changed back.  
Note also that with `--keep-going` you will still see _all_ the errors (at least so it is promised).  An advantage of `-W` by default (as opposed to just in the CI) is a quicker feedback cycle (if it passes locally, it also passed the CI).

Mikolaj:
Thank you!

PR #7608, opened 2021-08-31
glob expand extra-src-files when tracking for rebuild

gbaz:
Fixes partially the issue discussed in #4746 by expanding globs when tracking extra-src-files

### DIFF
--- a/cabal-install/src/Distribution/Client/SourceFiles.hs
+++ b/cabal-install/src/Distribution/Client/SourceFiles.hs
@@ -12,11 +12,14 @@
 -- we cannot "see" easily.
 module Distribution.Client.SourceFiles (needElaboratedConfiguredPackage) where
 
+import Control.Monad.IO.Class
+
 import Distribution.Client.ProjectPlanning.Types
 import Distribution.Client.RebuildMonad
 
 import Distribution.Solver.Types.OptionalStanza
 
+import Distribution.Simple.Glob (matchDirFileGlobWithDie)
 import Distribution.Simple.PreProcess
 
 import Distribution.Types.PackageDescription
@@ -35,6 +38,7 @@ import Distribution.ModuleName
 
 import Prelude ()
 import Distribution.Client.Compat.Prelude
+import Distribution.Verbosity (silent)
 
 import System.FilePath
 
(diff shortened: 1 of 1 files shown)


cabal-install/src/Distribution/Client/SourceFiles.hs
```diff
@@ -138,13 +142,14 @@ needBuildInfo pkg_descr bi modules = do
     -- A.hs-boot; need to track both.
     findNeededModules ["hs", "lhs", "hsig", "lhsig"]
     findNeededModules ["hs-boot", "lhs-boot"]
+    expandedExtraSrcFiles <- liftIO $ fmap concat . for (extraSrcFiles pkg_descr) $ \fpath -> matchDirFileGlobWithDie silent (\ _ _ -> return []) (specVersion pkg_descr) "." fpath
```
Mikolaj, line 145: Are we sure we don't want to die with an error message? Is there any risk of false positives that would justify that?

Mikolaj (changes requested):

cabal-install/src/Distribution/Client/SourceFiles.hs
gbaz: in this setting where we're just tracking setting stuff, dying seems weird.

The only other place this gets glob expanded is in the sdist step, where dying makes sense, because you're sdisting something with missing files. But here we're just checking if we need to trigger a reconfigure...

Mikolaj: Still, if the user thinks the compiler rebuilds his stuff, but in fact it's not, because the glob is incorrect, we should die ASAP, right? Unless false positives are possible. E.g., the user writes too general globs predicting some files/directories that are not all there yet. Is that or any other false positive possible and likely?

gbaz: The only error would be if the glob matches nothing. I'm extremely chary of having any sort of death or error reporting on a _purely administrative_ step. I don't think there's anywhere else inside the `Rebuild` monad that we die, so it seems like it would be a significant change from how things now occur. A user will notice if a rebuild happens or not. The behavior no rebuild happening is not substantially different than if they put nothing at all in their extra-src-files, or put in something that matches but doesn't match everything they intended, etc.

That it so say, we're not in a good position here to catch if this is correct. The only way to catch if it is correct is to see if the tarball generated by sdist actually builds.

Mikolaj: Good point about sdist tarball building. Let it be so.

Regarding possible errors, I thought it can fail in a variety of ways, both due to a syntax error in the glob and due to misssing directories, etc., as in https://github.com/haskell/cabal/blob/master/Cabal/src/Distribution/Simple/Glob.hs#L243-L250

Mikolaj (approved):

cabal-install/src/Distribution/Client/SourceFiles.hs
gbaz: thanks for the clarification on possible errors, I'm not particularly familiar with the glob code.

Mikolaj: On a second thought, I hope syntax errors in globs are caught already when .cabal file is parsed (it definitely is at `cabal check`, but that would not be enough). So, no point preparing to handle these errors again.

fgaz (commented):
This may need a changelog entry, especially since it fixes an issue that we're tracking

gbaz:
added

jneira (approved):
LGTM @fgaz: would be this ready to go after the changelog note?

Mikolaj:
@fgaz: ping :)

PR #7611, opened 2021-09-02
testsuite: remove file generated by autoheader

andreasabel:
Testsuite: remove file generated by autoheader.

After running the cabal-testsuite, I found changes in this file, so I suppose it is a generated file and can (and should) be removed from the source tree.

It actually says so:
```c
/* include/HsZlibConfig.h.in.  Generated from configure.ac by autoheader.  */
```


andreasabel:
Ok, now CI fails, i have to investigate...

ulysses4ever:
@mergify rebase

ulysses4ever:
I wonder if it fails because the `include` directory will cease to exist after this change (it may seem like the change only removes the file but it was the only one in that dir, and git doesn’t like empty dirs…).

ulysses4ever:
That was it.

ulysses4ever (approved):

Mikolaj (approved):

Mikolaj:
@andreasabel, are you ready to mergify-merge?

Mikolaj:
Let me step in and add the label...

andreasabel:
Thanks, @ulysses4ever, for finishing this!

andreasabel:
@ulysses4ever : Here, it claims that I forced-pushed.  ;-)

Issue #6114, opened 2019-06-26
--enable-tests causes recalculation


fommil:
**Describe the bug**

Repeatedly swapping between

```
cabal v2-build
```

and

```
cabal v2-build --enable-tests
```

causes a recalculation of all the dependencies (but no recompilation, after the first full build).

**To Reproduce**

See above

**Expected behavior**

I would expect (after the first full compilation) no further work by cabal.

**System informataion**

```
cabal --version
cabal-install version 2.4.1.0
compiled using version 2.4.1.0 of the Cabal library 
```

on any GNU/Linux with any ghc.

**Additional context**

My projects are quite large with multiple `.cabal` packages and many hackage dependencies. Hence recalculation is quite costly (e.g. 30 seconds every time the dependencies are recomputed). I'm also using a freeze file.

In addition,

```
cabal v2-run tasty
```

vs

```
cabal v2-run tasty --enable-tests
```

exhibits the same problem.

A workaround is to always use `--enable-tests` even when doing `v2-run tasty` and even though it is redundant.

hvr:
I'm afraid that what you seem to be asking is at odds with the goal for `cabal v2-build` to be stateless; if you want to be stateful, there's `cabal v2-configure` -- this is the only command that's supposed to persist the configuration settings which affects the solver (by modifying or creating your `cabal.project.local`).

And `--enable-tests` is such a solver input; if you toggle it, it *will* require the solver to be re-run. And since you're passing `--enable-tests` only to `v2-build` which by design is supposed to *not* persist the settings, and be forgetful about it on the next `v2-build` invocation (unless you re-pass the same flags), a solver recomputation is what you'll get.

However, one thing we could do is to create a solver-cache indexed by solver inputs, so that switching between `cabal v2-build --enable-tests` and `cabal v2-build --disable-tests` would be able to retrieve a previously cached plan w/ requiring a full resolving (if and only if all solver inputs remained the same).

gbaz:
I agree that a solver cache would be a very nice solution to this and related issues.

symbiont-sam-halliday:
I'm not sure if this is the same bug or not but using `--test-option(s)` with `v2-test` also causes recalculation (in cabal-install 3.0) and it shouldn't.

hvr:
it seems like `--test-option` isn't the only flag that has this effect from which one wouldn't expect this;

another one is `--test-show-detail` or `--test-machine-log` which seem to be considered input to the build-plan when modified:

```
[...]

PR #7614, opened 2021-09-02
don't consider test flags in reconfigure check

### LINKED ISSUE #6267: Options that change how tests are run should not cause a rebuild
**Describe the bug**
Building my project and the running the tests, causes the test command to rebuild everything.

I like to see the output of my test suites so I use `--test-show-details=direct` and I have multiple test suites and to avoid them all running at the same time (and having their output all come out at once) I use `-j1`.

**To Reproduce**
Using cabal version 3.0:
```
$ cabal build all
$ cabal test all -j1 --test-show-details=direct
```
If I run the same two commands again, the `build` command will rebuild everything and then the `test` command will again rebuild everything.

**Expected behavior**
The `test` command should only build the stuff that has not been built.

**System informataion**
 - Linuix
 - cabal 3.0
 - ghc 8.6.5

Related: https://github.com/haskell/cabal/issues/6129 https://github.com/haskell/cabal/issues/6069

gbaz:
Resolves #6267 and #6114

This is a conservative cut at which flags to ignore, on the grounds its better to be safe than sorry. It can always be expanded based on other requests.

### DIFF
--- a/cabal-install/src/Distribution/Client/ProjectBuilding.hs
+++ b/cabal-install/src/Distribution/Client/ProjectBuilding.hs
@@ -379,6 +379,12 @@ packageFileMonitorKeyValues elab =
     -- do not affect the configure step need to be nulled out. Those parts are
     -- the specific targets that we're going to build.
     --
+
+    -- Additionally we null out the parts that don't affect the configure step because they're simply
+    -- about how tests or benchmarks are run
+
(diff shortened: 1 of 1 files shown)


jneira (approved):
How many time saved with so few lines, really amazing

Mikolaj (approved):
LGTM

andreasabel:
Horray for this fix!

PR #7615, opened 2021-09-03
avoid bad duplicate module checks

### LINKED ISSUE #4629: cabal check doesn't account for conditionals
In [`binary-serialise-cbor`]() we are trying to provide `exposed-modules` for GHC 7.8, where `reexported-modules` is not supported. For instance,
```
library
  default-language:  Haskell2010
  hs-source-dirs:      src
  if impl(ghc >= 7.9)
    reexported-modules:
      Codec.Serialise            as Data.Binary.Serialise.CBOR,
      Codec.Serialise.Class      as Data.Binary.Serialise.CBOR.Class,
      Codec.Serialise.Decoding   as Data.Binary.Serialise.CBOR.Decoding,
      Codec.Serialise.Encoding   as Data.Binary.Serialise.CBOR.Encoding,
      Codec.CBOR.FlatTerm        as Data.Binary.Serialise.CBOR.FlatTerm,
      Codec.Serialise.IO         as Data.Binary.Serialise.CBOR.IO,
      Codec.CBOR.Magic           as Data.Binary.Serialise.CBOR.Magic,
      Codec.CBOR.Pretty          as Data.Binary.Serialise.CBOR.Pretty,
      Codec.Serialise.Properties as Data.Binary.Serialise.CBOR.Properties,
      Codec.CBOR.Write           as Data.Binary.Serialise.CBOR.Write,
      Codec.CBOR.Term            as Data.Binary.Serialise.CBOR.Term,
      Codec.Serialise.Tutorial   as Data.Binary.Serialise.CBOR.Tutorial
    exposed-modules: Data.Binary.Serialise.CBOR.Read
  else
    -- GHC 7.8 doesn't support module re-exports
    exposed-modules:
      Data.Binary.Serialise.CBOR,
      Data.Binary.Serialise.CBOR.Class,
      Data.Binary.Serialise.CBOR.Decoding,
      Data.Binary.Serialise.CBOR.Encoding,
      Data.Binary.Serialise.CBOR.FlatTerm,
      Data.Binary.Serialise.CBOR.IO,
      Data.Binary.Serialise.CBOR.Magic,
      Data.Binary.Serialise.CBOR.Pretty,
      Data.Binary.Serialise.CBOR.Properties,
      Data.Binary.Serialise.CBOR.Read,
      Data.Binary.Serialise.CBOR.Write,
      Data.Binary.Serialise.CBOR.Term

  build-depends:
    base < 5,
    bytestring < 1.0,
    cborg     == 0.1.*,
    serialise == 0.1.*

```
Unfortunately, `cabal check` fails on this with,
```
$ cabal check
The package will not build sanely due to these errors:
* Duplicate modules in library: Data.Binary.Serialise.CBOR,
Data.Binary.Serialise.CBOR.Class, Data.Binary.Serialise.CBOR.Decoding,
Data.Binary.Serialise.CBOR.Encoding, Data.Binary.Serialise.CBOR.FlatTerm,
Data.Binary.Serialise.CBOR.IO, Data.Binary.Serialise.CBOR.Magic,
Data.Binary.Serialise.CBOR.Pretty, Data.Binary.Serialise.CBOR.Properties,
Data.Binary.Serialise.CBOR.Read, Data.Binary.Serialise.CBOR.Term,
Data.Binary.Serialise.CBOR.Write

```

It seems that Cabal doesn't account for the conditional when computing this check.

gbaz:
Resolves #4629

This tries to be relatively minimal, by introducing a special flattening call which attempts to merge conditionals without introducing spurious duplication.

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Configuration.hs
+++ b/Cabal/src/Distribution/PackageDescription/Configuration.hs
@@ -18,6 +18,7 @@
 module Distribution.PackageDescription.Configuration (
     finalizePD,
     flattenPackageDescription,
+    flattenPackageDescriptionWithConditionals,
 
     -- Utils
     parseCondition,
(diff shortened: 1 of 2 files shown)


gbaz:
I still need to test this a bit before merge, and possibly we want to introduce an explicit regression test for it as well...

Cabal/src/Distribution/PackageDescription/Configuration.hs
```diff
@@ -598,3 +599,92 @@ transformAllBuildDependsN f =
...
+newtype WrappedLibrary = WrappedLibrary {unwrapWrappedLibrary :: Library}
+
+instance Semigroup WrappedLibrary where
+   WrappedLibrary a <> WrappedLibrary b = WrappedLibrary $ ab
+                                          {exposedModules = exposed', libBuildInfo = lbi', reexportedModules = reexported', signatures = sigs' }
+    where
+      ab = a <> b
+
+      -- if something exists in both branches of a conditional, only add it once
+      o x y = x ++ filter (not . (`elem` x)) y
```
phadej, line 646: There is `ordNub` in https://hackage.haskell.org/package/Cabal-3.6.0.0/docs/Distribution-Utils-Generic.html#v:ordNub

gbaz: The intent isn't exactly like `ordNub` -- it preserves duplication as long as that duplication exists already in either of the branches. So if `x` or `y` contains duplicates, then `x o y` should also contain them. (And as I write this, I see where it actually fails to have that semantics fully, but that's the intention, at least !).

gbaz:
Actually this approach isn't the cleanest -- the conditional logic is trickier to capture than I think this newtype approach allows. Will close and try a different angle.

PR #7617, opened 2021-09-03
remove unnecessary dotfiles, add to .gitignore

emilypi:
---


gbaz:
we may want to keep  .git-blame-ignore-revs ?

emilypi:
@gbaz it's been more than a year. Do we want to set a time when we want to stop caring?

phadej (changes requested):
`.git-blame-ignore-revs ` is definitely not unnecessary.

Mikolaj:
So, does HLS solve the import sorting problem for us? So that we can remove .stylish-haskell.yaml and not see the imports deteriorating over time? How many people use HLS? How many use stylish-haskell (I do)?

emilypi:
@Mikolaj 

> Does HLS solve the sorting problem? 

@jneira is this coming down the pipeline any time soon? Right now, no, but I also do not think Cabal should be the one to support tools. I don't use `stylish-haskell`, and I don't see users who regularly do. However, I haven't done a statistical analysis of our users. 

> not see the imports deteriorating over time?

I have some unfortunate news for you... *they deterioated despite having that config* :joy:. I'm not convinced that the config actually helps with this, since we do not enforce the _lack_ of its use. So, I would put a constraint on keeping it: 

1. If we keep it, CI must require its use and run a pass checking the new code additions against a formatting of the whole project
2. We let the users contribute and manage their own tools.

emilypi:
@phadej same question as the one i posed to @gbaz: when do we stop caring about those commits and blaming?

phadej:
@phadej, never. I have dug history of Cabal far back many times, few years is not a long time span. And I'm quite sure there will be future commits you want to ignore too (given you craft similar big moving commits separately, and IMO you should).

Say, if you reformat everything with ormolu...

emilypi:
@phadej Maybe, but i'm wondering about the utility of looking back in history after a significant number of changes has occurred such that the project is sufficiently different from the one many years back. That's why i'm asking. It seems like git blaming anything except recent commits on a project as large as this one yields diminishing returns the farther back you go.

phadej:
It doesn't really cost anything to keep that file around and updated. You never know. (I have looked past 2016 when I made that file, otherwise those files wouldn't be there).

emilypi:
@phadej fair enough!

gbaz (approved):

Mikolaj (approved):

fgaz:
Just commenting to emphasize @phadej's point. It happened to me too, multiple times, and it saved me hours. History (and that file) is important!

Issue #7559, opened 2021-08-19
Split `Cabal` into `Cabal-syntax` and `Cabal`?


gbaz:
As discussed in https://github.com/haskell/cabal/issues/7548#issuecomment-901001578 , some people find a dependency on the full `Cabal` library somewhat heavyweight for just doing cabal file parsing without actually executing any builds.

In general, I think splitting off the basic file parsing and syntax tree stuff of the lib into `Cabal-syntax` or the like would be a fine idea, and probably help with code organization to boot.

Maybe this could be a next step after the `exactprint` work, and it might also help encourage ides to integrate with the cabal library.

A bit more wild of an idea maybe the functionality of https://hackage.haskell.org/package/cabal-install-parsers could be deprecated by moving cabal-install's parsers into `Cabal-parsers` as well.

Ideally I think this (along with exact-print) would help tools be more comfortable using the already built utilities for cabal files instead of rolling their own (or punting).

gbaz:
Note that hvr also has an experimental cabal-parsers library (unreleased on hackage) that glues together the parsing code from various cabal versions so it can run them explicitly to check: https://github.com/haskell-hvr/cabal-parsers/ -- Pulling out Cabal's parser stuff into its own lib could considerably simplify such a package (but more generally pinning down a properly versioned cabal bnf etc could obviate the need for it).

phadej:
I'm :-1: on moving `cabal-install-parsers` stuff into `Cabal-syntax`. The `cabal-install` files (`cabal.config`, `cabal.project`) are `cabal-install` specific, e.g. `stack` will do completely nothing with them.

`cabal-parsers` idea is to be bug compatible. It's a pragmatic way to check that old `cabal-install`s can parse stuff uploaded on Hackage, removing the burden of being bug-compatible from the `Cabal(-syntax)` library itself.

phadej:
I'm pretty sure that `Cabal-syntax` dealing only with `.cabal` files could be even pulled out of this repository and developed and maintained by different (can be obviously overlapping) folks in their own pace. That's another argument to not couple it to `cabal-install`.

The only change needed for new GHC releases in format part of the lib, IIRC, is adding extensions & languages. Woiuld be nice to refactor the library to make those changes non-breaking changes PVP-wise. (I doubt anyone does exhaustive pattern match on extensions, yet there is only three languages, so someone might do that). TL;DR `Language` and `Extension` type could be made opaque.

gbaz:
[...]

Issue #7531, opened 2021-08-11
RFC: Dropping build support for GHC < 8


emilypi:
Hi Everyone,

This project currently supports GHC >=7. However, as of the latest [Haskell Survey](https://taylor.fausak.me/2020/11/22/haskell-survey-results/#s2q4), roughly 1% of the community is using a GHC < 8.0: 

![image](https://user-images.githubusercontent.com/10459342/129093338-c02c2eeb-0306-4e39-9cad-0f2903b5a84c.png)

However, looking at CI times, on average, we spend roughly 47% of CI time on GHC < 8.0. This means that we're spending close to half of all build time supporting 1% of the community for a niche case that disproportionately negatively affects the 99%. Dropping GHCs would *significantly* improve the contributor pipeline, and help ease the burden going forward for both users and maintainers.  Part of me suspects that we support old GHCs in order to not cast anyone into the lurch if they happen to not be able to upgrade. However, I think having a concrete stability policy is just as good, and will set expectations appropriately so that people know what's coming and can act accordingly. So, my proposal is this: 

1. We drop GHC < 8.0 for Cabal and `cabal-install`. Older GHC users can always use older versions of Cabal. The point to stress here is that we want to put cap *ongoing* support at roughly 5 years for a particular GHC.
2. We establish a concrete stability policy to this effect, and announce it far and wide. 
3. etc. (whatever we come up with here as refinements)

chris-martin:
I do maintain libraries that support versions of GHC older than the GHCs I "use" myself, so I'm wondering if 1% is perhaps an underestimation of who this might affect in some way. (That said, I don't think I actually still support GHC 7.x on anything, because 8.0 added Semigroup.)

One nicety that might make this entirely painless for some "niche case" users: Could the default Cabal version in the [setup](https://github.com/haskell/actions/tree/main/setup) GitHub action be "the latest Cabal version that supports the requested GHC version"? That might spare some people the extra brain space of mapping GHC versions to appropriate Cabal versions themselves.

chris-martin:
Ah. In that case I don't even understand how it would be possible to be affected by this.

emilypi:
> Ah. In that case I don't even understand how it would be possible to be affected by this.

I've heard-say that some people are running GHC 7.2 still in production and want progressive fixes for it :grin:. This would obviously put those people out, but the burden is too great for this team.

mightybyte:
[...]

Issue #6452, opened 2019-12-21
cabal-version: 3.4 Don't generate `Paths_pkgname.hs` it it's not in `autogen-modules`


phadej:
This is following explicit rule of "don't do anything you are not told to do".

- `cabal-version: 3.4.` only generate `Paths_pkgname.hs` if there is `autogen-modules: Paths_pkgname` (note, `other-modules` would not be enough!)
- `cabal-version: 3.0` or older: generate `Paths_pkgname.hs` always.

Along the same vein we could not generate `cabal_macros.h` (if there's no `*-extensions: CPP`), but that would be a silent breakage as recent GHCs provide some of definitions. So for `cabal_macros.h`, the best we can do is to not generate additional `TOOL_*` macros. #6442 is related

- `cabal-version: 3.4` and *recent GHC*: generate `cabal_macros.h` only if there **is** `*-extensions: CPP`
- `cabal-versions: 3.4` and *old GHC*:
    - generate full `cabal_macros.h` if there's `*-extensions: CPP`
    - generate only `TOOL_` and `CURRENT_` definitions if there **is not** `*-extensions: CPP`
- `cabal-version: 3.0` or older, and *recent GHC*: generate only `TOOL_` and `CURRENT_` definitions
- `cabal-version: 3.0` or older, and *old GHC*: generate full `cabal_macros.h`

Note: currently `Paths_pkgname.hs` uses CPP itself, but it can be changed.
We can generate Paths file suitable for specific compiler/dependencies version.

Discussion time: until 2020-02-01

PR #7620, opened 2021-09-04
Create Cabal-syntax for .cabal parsing and types

patrickdoc:
---

---

This PR splits `Cabal` into `Cabal-syntax` and `Cabal` (re: #7559).

The changes were generally conservative to avoid making design decisions. I mostly just pulled PackageDescription.hs and GenericPackageDescription.hs into the new package, and then moved code over as necessary to fix all the compilation errors.

The two code changes I did make were to move a few utility parsing functions so that Distribution.Verbosity stayed in `Cabal` rather than `Cabal-syntax`. Otherwise it was just import changes to avoid going through the Distribution.Simple.Utils module.

The result is split almost exactly in half (126 modules in `Cabal-syntax`, 121 modules in `Cabal`)


Cabal-syntax/Cabal-syntax.cabal
```diff
@@ -0,0 +1,243 @@
...
+    Distribution.Types.VersionRange.Internal
+    Distribution.Utils.Generic
+    Distribution.Utils.MD5
+    Distribution.Utils.Path
+    Distribution.Utils.ShortText
+    Distribution.Utils.Structured
+    Distribution.Version
+    Language.Haskell.Extension
+
+  other-modules:
```
phadej, line 141: Most of these modules should be `exposed-modules`. In fact, the different lists could be combiled.

I don't think any module have to be hidden (i.e. `other-module`), except maybe some `Compat` modules, though I'm not sure about these either.

Cabal-syntax/Cabal-syntax.cabal
```diff
@@ -0,0 +1,243 @@
...
+  exposed-modules:
+    Distribution.CabalSpecVersion
+    Distribution.Compat.Binary
+    Distribution.Compat.DList
+    Distribution.Compat.Exception
+    Distribution.Compat.Lens
+    Distribution.Compat.Newtype
+    Distribution.Compat.NonEmptySet
+    Distribution.Compat.Prelude
+    Distribution.Compat.Semigroup
```
fgaz, line 82: We may want to split the compat stuff between Cabal and Cabal-syntax, like it's done now with cabal-install and Cabal, but I'm not sure it's worth doing in this case.

Cabal-syntax/src/Distribution/Fields/ParseResult.hs
```diff
@@ -160,37 +154,3 @@ parseFatalFailure' = PR pr
...
 
     err = PError zeroPos "Unknown fatal error"
-
--- | Helper combinator to do parsing plumbing for files.
---
--- Given a parser and a filename, return the parse of the file,
--- after checking if the file exists.
---
--- Argument order is chosen to encourage partial application.
-readAndParseFile
```
phadej, line 170: Moving these seems wrong. (probably related that this module have to exposed?)

phadej (commented):
This indeed looks very simple.

Could you make changes I mentioned.

Making CI pass will need some changes there, but I can take a look at that.

Note: GHC devs will need to be aware of this change, their build system will need adjustments too.

Cabal-syntax/Cabal-syntax.cabal
phadej: I'd leave them for now, and just drop pre-GHC-8 support (#7531) before next major. That would cleanup plenty of stuff, leaving mostly "We rather not depend on `dlist` package and make our own copy".

phadej:
Most (maybe even all?) `test-suites` in `Cabal` can be moved to `Cabal-syntax`. These might highlight if things are in wrong place too (or imported from wrong place):

```
HackageTests.hs:import Distribution.Simple.Utils                   (fromUTF8BS, toUTF8BS)
UnitTests.hs:import Distribution.Simple.Utils
UnitTests.hs:import qualified UnitTests.Distribution.Simple.Glob
UnitTests.hs:import qualified UnitTests.Distribution.Simple.Program.GHC
UnitTests.hs:import qualified UnitTests.Distribution.Simple.Program.Internal
UnitTests.hs:import qualified UnitTests.Distribution.Simple.Utils
UnitTests.hs:    , testGroup "Distribution.Simple.Glob"
UnitTests.hs:        UnitTests.Distribution.Simple.Glob.tests
UnitTests.hs:    , UnitTests.Distribution.Simple.Program.GHC.tests
UnitTests.hs:    , testGroup "Distribution.Simple.Program.Internal"
UnitTests.hs:        UnitTests.Distribution.Simple.Program.Internal.tests
UnitTests.hs:    , testGroup "Distribution.Simple.Utils" $
UnitTests.hs:        UnitTests.Distribution.Simple.Utils.tests ghcPath
UnitTests/Distribution/Compat/Time.hs:import Distribution.Simple.Utils (withTempDirectory)
UnitTests/Distribution/Simple/Glob.hs:module UnitTests.Distribution.Simple.Glob
UnitTests/Distribution/Simple/Glob.hs:import Distribution.Simple.Glob
UnitTests/Distribution/Simple/Program/GHC.hs:module UnitTests.Distribution.Simple.Program.GHC (tests) where
UnitTests/Distribution/Simple/Program/GHC.hs:import Distribution.Simple.Program.GHC (normaliseGhcArgs)
UnitTests/Distribution/Simple/Program/GHC.hs:tests = testGroup "Distribution.Simple.Program.GHC"
UnitTests/Distribution/Simple/Program/Internal.hs:module UnitTests.Distribution.Simple.Program.Internal
UnitTests/Distribution/Simple/Program/Internal.hs:import Distribution.Simple.Program.Internal ( stripExtractVersion )
UnitTests/Distribution/Simple/Utils.hs:module UnitTests.Distribution.Simple.Utils
UnitTests/Distribution/Simple/Utils.hs:import Distribution.Simple.BuildPaths ( exeExtension )
UnitTests/Distribution/Simple/Utils.hs:import Distribution.Simple.Utils
```

Looks like very few unit tests are testing `Distribution.Simple`.

phadej:
`Distribution.Simple.Glob` could be renamed to `Distribution.Utils.Glob`.

Cabal-syntax/src/Distribution/Fields/ParseResult.hs
patrickdoc: So I moved these because "Verbosity" seems like a cabal-install specific thing as opposed to a Cabal-syntax thing, so I left it in `Cabal`.

These seem to be mostly for nice error reporting in a context where "Verbosity" makes sense, but I would guess most external use cases would rather handle the errors differently.

Cabal-syntax/Cabal-syntax.cabal
patrickdoc: Updated

Cabal-syntax/src/Distribution/Fields/ParseResult.hs
phadej: Good point about `Verbosity`.

patrickdoc:
Running into some small compatibility issues, and not sure what the standard practices are here.

1) I've locally moved the .cabal specific checks of this module: https://github.com/haskell/cabal/blob/master/Cabal/src/Distribution/PackageDescription/Check.hs to `Cabal-syntax`, but one of the checks requires `autogenPathsModuleName` (to verify `Paths_Cabal`, etc.). This is defined in Distribution.Simple.BuildPaths, and it doesn't seem reasonable to me to move all of those build paths into `Cabal-syntax`. Not sure if `Paths_Cabal` is an official part of the `PackageDescription` syntax, or if it's just an implementation detail of `Cabal`.

2) Moving `Distribution.Package` to `Cabal-syntax` breaks `hackage-security`. https://github.com/haskell/hackage-security/search?p=1&q=Distribution It would be pretty trivial to move `hackage-security` to use `Cabal-syntax`, but not sure how these types of changes are usually handled. Otherwise I could add a compatibility layer, but there might be some weirdness with name conflicts.

gbaz:
I'm not sure if moving the checks to cabal-syntax makes sense or not as a whole? I would thing the checks are for the _semantics_ of a well-formed cabal file, not for parsing and manipulating its syntax. I can see arguments either way, though.

On the latter, could you use reexported modules? https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-library-reexported-modules

I'm not sure if that fits in the support window for Cabal features, but I think it might?

phadej:
@gbaz, consider `hackage-server`. I think `Cabal-syntax` is enough for it, and `hackage-server` uses `PD.Check`. Also `hackage-cli` could use check. So i'd rather have that in `Cabal-syntax`, though I'm not against having `Cabal-check` lib too, but maybe that is too much at once!

gbaz:
As the comments say:

```
-- This has code for checking for various problems in packages. There is one
-- set of checks that just looks at a 'PackageDescription' in isolation and
-- another set of checks that also looks at files in the package. Some of the
```

Hackage-server needs both types of checks, so I imagine it would need stuff that went beyond cabal-syntax regardless...

phadej:
So? The package content checking interface is abstracted:

```haskell
-- | A record of operations needed to check the contents of packages.
-- Used by 'checkPackageContent'.
--
data CheckPackageContentOps m = CheckPackageContentOps {
    doesFileExist        :: FilePath -> m Bool,
    doesDirectoryExist   :: FilePath -> m Bool,
    getDirectoryContents :: FilePath -> m [FilePath],
    getFileContents      :: FilePath -> m BS.ByteString
  }

-- | Sanity check things that requires looking at files in the package.
-- This is a generalised version of 'checkPackageFiles' that can work in any
-- monad for which you can provide 'CheckPackageContentOps' operations.
--
-- The point of this extra generality is to allow doing checks in some virtual
-- file system, for example a tarball in memory.
--
checkPackageContent :: (Monad m, Applicative m)
                    => CheckPackageContentOps m
                    -> PackageDescription
                    -> m [PackageCheck]
```

---

Or put it other way around: `hackage-server` doesn't need to depend on `Distribution.Simple`, it doesn't build the packages. (builds do, but they probably use `cabal-install` or something higher level anyway!)

gbaz:
Right, we _could_ put all those checks into cabal syntax. My understanding is those checks do not belong in cabal-syntax, because they're beyond the scope of cabal-syntax, which is, as the name says, _cabal syntax_, and checks for well-formedness of a tarball don't make sense there. I'm proposing we put no checks in cabal syntax, since most clients will not need them, or those that will, will want all of them, including the ones that don't belong there.

As one of the primary people currently maintaining hackage-server, I really don't care if it needs to continue to depend on the whole `Cabal` library -- that seems fine to me.

patrickdoc:
Reexported modules were exactly what I needed, thanks.

Looking at the check code again, there wasn't really a compelling reason to move it in the first place. Just a little over-aggressive in trimming the test dependencies. I'll roll that back.

patrickdoc:
Rolled back the PackageDescription Check change, updated the rest of the modules, and I think CI should work now.

phadej:
The 

```haskell
-- | The name of the auto-generated Paths_* module associated with a package
autogenPathsModuleName :: PackageDescription -> ModuleName
autogenPathsModuleName pkg_descr =
  ModuleName.fromString $
    "Paths_" ++ map fixchar (prettyShow (packageName pkg_descr))
  where fixchar '-' = '_'
        fixchar c   = c
```

can very well be in `Cabal-syntax`. https://github.com/haskell/cabal/issues/6452 is an issue which would make the name very much relevant for `Cabal-syntax` part (though parsing doesn't change).

phadej:
I tried this patch and it `cabal-fmt` and [`cabal-extras`](https://github.com/phadej/cabal-extras) work with this patch by changing `Cabal` to `Cabal-syntax`.

Except of `cabal-docspec` which is expected, as it does cpp preprocessing itself, i.e. very much "builds" the packages.

Cabal-QuickCheck/Cabal-QuickCheck.cabal
```diff
@@ -13,8 +13,9 @@ library
   build-depends:
     , base
     , bytestring
-    , Cabal       ^>=3.7.0.0
-    , QuickCheck  ^>=2.13.2 || ^>=2.14
+    , Cabal         ^>=3.7.0.0
```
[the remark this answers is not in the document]
patrickdoc: `Cabal-QuickCheck` needs:

- `Distribution.Simple.Compiler  (DebugInfoLevel(..), OptimizationLevel(..), PackageDB(..),ProfDetailLevel(..),knownProfDetailLevels`
- `Distribution.Simple.Flag (Flag(..))`
- `Distribution.Simple.InstallDirs`
- `Distribution.Simple.Setup (HaddockTarget(..), TestShowDetails(..))`
- `Distribution.Utils.NubList`
- `Distribution.Verbosity`

`Cabal-described` needs `Distribution.Verbosity`
```
instance Described Verbosity where
    describe _ = REUnion
        [ REUnion ["0", "1", "2", "3"]
        , REUnion ["silent", "normal", "verbose", "debug", "deafening"]
          <> REMunch reEps (RESpaces <> "+" <>
            -- markoutput is left out on purpose
            REUnion ["callsite", "callstack", "nowrap", "timestamp", "stderr", "stdout" ])
        ]
```

patrickdoc:
Updated artifact/bootstrap/meta CI, but not sure how to teach `doctest` about the new package.

patrickdoc:
I think I've now fixed doctest and the Windows artifact issue, which just leaves the ghc 7.X failures due to reexported-modules not being supported.

jneira:
> 
> 
> Reexported modules were exactly what I needed, thanks.
> 
> Looking at the check code again, there wasn't really a compelling reason to move it in the first place. Just a little over-aggressive in trimming the test dependencies. I'll roll that back.

minor (?) caveat, reexported modules dont work well with `stack repl` see https://github.com/commercialhaskell/stack/issues/5077#issuecomment-713511471
if they dont work with `stack repl` the project could not be loaded in hls neither for now

phadej:
`reexported-modules` would also tightly tie `Cabal-syntax` and `Cabal` together: whenever there is a minor version of `Cabal-syntax`, there should be a minor version of `Cabal` too. That is probably not an issue for `Cabal`, but in general people have to be aware of that tradeoff. IMO `Cabal` don't need to re-export `Cabal-syntax`, though it would help in the migration, migration need is transient. (I expect very few projects having needs beyond `Cabal-syntax`).

phadej:
That said. It would be very nice to check what `stack` uses from `Cabal`. I.e. how much extra it needs beyond `Cabal-syntax`.

From my code-only scan, it seems  it needs only macro generation (i.e. the same as `cabal-docspec`).

And now I'm starting to think where those belong. :(

patrickdoc:
> `reexported-modules` would also tightly tie `Cabal-syntax` and `Cabal` together: whenever there is a minor version of `Cabal-syntax`, there should be a minor version of `Cabal` too.

To be clear, I was only reexporting the single module Distribution.Package so that hacakge-security didn't need to be patched for cabal-install to build successfully. I can also update hackage-security if that route is easier.

> That said. It would be very nice to check what `stack` uses from `Cabal`. I.e. how much extra it needs beyond `Cabal-syntax`.

From a quick look, the two most likely issues I saw were the `Setup.hs` files here: https://github.com/commercialhaskell/stack/search?q=Distribution.Simple

phadej:
[`Distribution.Package`](https://hackage.haskell.org/package/Cabal-3.6.0.0/docs/src/Distribution.Package.html) re-exports other stuff. Still, my argument holds, even single (small) re-exported module ties the packages together.

patrickdoc:
Any suggestions on how to deal with the breaking change for `hackage-security` without re-exporting `Distribution.Package`? It's needed for `cabal-install`, so we would need a patched version of `hackage-security` using the new `Cabal-syntax` for the builds to work.

gbaz:
Just make a branch of hackage-security with the patch. Since its under the haskell org umbrella, we can make sure the patch is accepted when this is merged.

That said, I don't think reexported modules is problematic here. The concern about stack is about backpack, afaik, not reexported modules, which is a much earlier feature, and the version number coupling should not be a concern in our _particular_ case. (i.e. the idea of an update to the syntax lib without an update to the main Cabal lib seems unlikely)

jneira:
>  The concern about stack is about backpack, afaik, not reexported modules, which is a much earlier feature

yeah, i overlooked it, i hit the issue using `mixins` and no `reexported-modules`

patrickdoc:
Rebased, removed the reexport, and created a patch for hackage-security: https://github.com/haskell/hackage-security/pull/254

fgaz:
@phadej do you think this is ready to be merged?

patrickdoc:
@fgaz There's some extended conversation on the related hackage-security pr: https://github.com/haskell/hackage-security/pull/254#discussion_r744166675

It looks like the recommendation is to release a dummy Cabal-syntax 3.6 that helps create mutually exclusive bounds, then update/release hackage security with the linked pr, then merge this. I'm not sure how to set that in motion.

Mikolaj:
I will help, I'm a maintainer of `hackage-security`. Tell me what to do.

Mikolaj (approved):
LGTM

Mikolaj:
The hackage-security PR is merged, the dummy Cabal-syntax is on Hackage, all seems ready to go. Last moment to have a critical look before merge. This is quite a fabulous PR IMHO, but also very disruptive, so it'd be good not to revert and re-revert it too many times...

phadej:
I think this is great overall (and we might or might not move some modules around in the future). However, to avoid reverts, you should make GHC devs aware that this is going to happen *soon*, cause if they realise this is too long, there will be too much pressure to revert indeed.

solver-benchmarks/solver-benchmarks.cabal
```diff
@@ -41,6 +46,11 @@ library
     vector
   default-language: Haskell2010
 
+  if flag(Cabal-syntax)
```
phadej, line 49: this package may depend on `Cabal & Cabal-syntax ==3.7.0.0` only, as its a test package

Mikolaj:
> I think this is great overall (and we might or might not move some modules around in the future). However, to avoid reverts, you should make GHC devs aware that this is going to happen _soon_, cause if they realise this is too late (in GHC release pipeline), there will be too much pressure to revert indeed.

Than you for the heads up. I've just told the GHC hackers and will liaison with them as needed.

solver-benchmarks/solver-benchmarks.cabal
patrickdoc: Updated to only `Cabal-syntax >= 3.7`

Cabal/src/Distribution/Simple/PackageDescription.hs
```diff
@@ -0,0 +1,74 @@
...
+-- Module      :  Distribution.PackageDescription.Parsec
+-- Copyright   :  Isaac Jones 2003-2005
+-- License     :  BSD3
+--
+-- Maintainer  :  cabal-devel@haskell.org
+-- Portability :  portable
+--
+-- This defines parsers for the @.cabal@ format
+
+module Distribution.Simple.PackageDescription (
```
fgaz, line 12: the `Module` haddock is inconsistent with the package name. Also why was this moved from `Cabal-syntax`'s `Distribution.PackageDescription.Parsec`?

fgaz (commented):
Could you add `Cabal-syntax` in `README.md`, under "This Cabal Git repository contains the following packages:"?

Cabal/src/Distribution/Simple/PackageDescription.hs
fgaz: > Also why was this moved from Cabal-syntax's Distribution.PackageDescription.Parsec?

Ah I see, `Verbosity` is in cabal-install

patrickdoc:
> Could you add `Cabal-syntax` in `README.md`, under "This Cabal Git repository contains the following packages:"?

Updated

Cabal/src/Distribution/Simple/PackageDescription.hs
patrickdoc: Yeah, these are mostly just error-handling wrappers around the actual parsers. And since they use `Verbosity` it felt better to keep them in `cabal-install` instead of letting that leak into `Cabal-syntax`.

bgamari:
I suspect that it shouldn't be too hard to adapt GHC to this change but I'll have to try to know for certain.

Mikolaj:
> I suspect that it shouldn't be too hard to adapt GHC to this change but I'll have to try to know for certain.

Should we wait?

Cabal-syntax/Cabal-syntax.cabal
```diff
@@ -0,0 +1,251 @@
...
+    base       >= 4.6      && < 5,
+    bytestring >= 0.10.0.0 && < 0.12,
+    containers >= 0.5.0.0  && < 0.7,
+    deepseq    >= 1.3.0.1  && < 1.5,
+    directory  >= 1.2      && < 1.4,
+    filepath   >= 1.3.0.1  && < 1.5,
+    mtl        >= 2.1      && < 2.3,
+    parsec     >= 3.1.13.0 && < 3.2,
+    pretty     >= 1.1.1    && < 1.2,
+    text       >= 1.2.3.0  && < 1.3,
```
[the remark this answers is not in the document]
Mikolaj: Indeed, we bumped this already on master, but mechanical rebasing did not reflect that in a PR a new package. I suppose the bounds need to be resynchronized with Cabal.cabal.

bgamari:
> > I suspect that it shouldn't be too hard to adapt GHC to this change but I'll have to try to know for certain.
> 
> Should we wait?

I am currently working on attempting the adaptation. So far so good but I will feel better once both build systems have passed the testsuite. I'll let you know shortly.

bgamari:
Is `Cabal-syntax/src/Distribution/Fields/Lexer.hs` guaranteed to be up-to-date?

phadej:
> Is `Cabal-syntax/src/Distribution/Fields/Lexer.hs` guaranteed to be up-to-date?

Yes. `Cabal`s own CI doesn't use `alex` either. And there is a check that it should be up to date.

bgamari:
I have confirmed that with the following patch I can bootstrap GHC:

```patch
diff --git a/Cabal-syntax/Cabal-syntax.cabal b/Cabal-syntax/Cabal-syntax.cabal
index 7649ac76b..91c04002d 100644
--- a/Cabal-syntax/Cabal-syntax.cabal
+++ b/Cabal-syntax/Cabal-syntax.cabal
@@ -41,7 +41,7 @@ library
     mtl        >= 2.1      && < 2.3,
     parsec     >= 3.1.13.0 && < 3.2,
     pretty     >= 1.1.1    && < 1.2,
-    text       >= 1.2.3.0  && < 1.3,
+    text       (>= 1.2.3.0  && < 1.3) || (>= 2.0 && < 2.1),
     time       >= 1.4.0.1  && < 1.12,
     -- transformers-0.4.0.0 doesn't have record syntax e.g. for Identity
     -- See also https://github.com/ekmett/transformers-compat/issues/35
diff --git a/Makefile b/Makefile
index 9af2ede22..2e218ef7f 100644
--- a/Makefile
+++ b/Makefile
@@ -23,15 +23,10 @@ lib-ghc-7.8 :
 
 # source generation: Lexer
 
-LEXER_HS:=Cabal-syntax/src/Distribution/Fields/Lexer.hs
+LEXER_HS:= Cabal-syntax/src/Distribution/Fields/Lexer.hs
 
 lexer : $(LEXER_HS)
 
-$(LEXER_HS) : templates/Lexer.x
-       alex --latin1 --ghc -o $@ $^
-       cat -s $@ > Lexer.tmp
-       mv Lexer.tmp $@
-
 # source generation: SPDX
 
 SPDX_LICENSE_HS:=Cabal-syntax/src/Distribution/SPDX/LicenseId.hs
diff --git a/ghc-packages b/ghc-packages
index 8d827f6f9..e9e369c42 100644
--- a/ghc-packages
+++ b/ghc-packages
@@ -1,2 +1,3 @@
 Cabal
+Cabal-syntax
```

Mikolaj:
Well spotted re `/ghc-packages`.

patrickdoc:
Fixed text and ghc-packages + bumped time to `< 1.13` to match `Cabal.cabal`

Mikolaj:
I think we are go. :)

Issue #7448, opened 2021-06-19
Make pkgconfig-depends usable


bgamari:
**Describe the bug**
Cabal has for a long time supported discovering native library dependencies through `pkgconfig-depends`. This is extremely useful as:

 * it can be used by some distributions to transparently provide native dependencies to `cabal-install` (e.g. see [nix-pkgconfig](https://github.com/bgamari/nix-pkgconfig/blob/master/cabal.project.local))
 * it makes Cabal packages more robust against non-traditional packaging decisions
 * it eliminates the need for Cabal to guess whether the target platform has a particular library

However, due to the fact that `pkg-config` is not available by default on some platforms, libraries tend to either not use `pkgconfig-depends` (e.g. see [`regex-posix`](https://hackage.haskell.org/package/regex-posix)) or guard it behind a default-off flag (e.g. see [`zlib`](https://hackage.haskell.org/package/zlib)). This severely cripples the usefulness of the mechanism.

Even worse, the tendency for packages to assume library non-availability on Windows can blow up catastrophically in some cases. This was the case in [GHC #19945](https://gitlab.haskell.org/ghc/ghc/-/issues/19945) where an ABI mismatch triggered by the assumption that Windows environments do not offer a POSIX regex implementation caused an innocuous-looking program to segfault, requiring many hours of debugging.

**Proposed design**

Users avoid using `pkgconfig-depends` by default because it can cause builds to fail when `pkg-config` is not available. I suggest one of the following (roughly in order of decreasing desirability):

 * `cabal-install` learn to backtrack if a `pkg-config` probe fails due to the executable not being available
 * `cabal-install` implement support for setting/clearing the `pkgconfig` flag on its own depending upon whether the `pkg-config` executable is available, sanctioning the current flag-based convention
 * `cabal-install` assume that the `pkgconfig-depends` are available in the event that `pkg-config` is not available

phadej:
There are related issues:
- https://github.com/haskell/cabal/issues/6771 (Allowing pkg-config to fail), mentions `has(pkg-config)` conditional|
- https://github.com/haskell/cabal/issues/5328 Extend the grammar/semantics of default: to support boolean expressions - with `has(pkg-config` would allow setting flags based on existence of pkg-config.

If these are implemented, then both first and second @bgamari suggestions will be possible.

gbaz:
[...]

PR #7621, opened 2021-09-05
Backtrack when no pkgconfigdb is present

gbaz:
Potentially resolves https://github.com/haskell/cabal/issues/7448

This is a one lexeme change that I would argue "fixes" `pkgconfig-depends` handling in the solver. Before, if there was no `pkg-config` that could be found, the solver would optimistically succeed. But since the builder then requires using `pkg-config` to get the linking flags, this would necessarily fail on build anyway. Now, we fail. This lets a user set up an auto flag that toggles between using a `pkgconfig-depends` and explicit `extra-libraries` etc stanza.

I _think_ this gives a nice solution that is strictly better than the prior situation.

### DIFF
--- a/cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs
+++ b/cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs
@@ -253,11 +253,12 @@ tests = [
         , runTest $ mkTest dbBuildable2 "choose version that sets buildable to false" ["A"] (solverSuccess [("A", 1), ("B", 2)])
          ]
     , testGroup "Pkg-config dependencies" [
-          runTest $ mkTestPCDepends Nothing dbPC1 "noPkgs" ["A"] anySolverFailure
+          runTest $ mkTestPCDepends (Just []) dbPC1 "noPkgs" ["A"] anySolverFailure
         , runTest $ mkTestPCDepends (Just [("pkgA", "0")]) dbPC1 "tooOld" ["A"] anySolverFailure
         , runTest $ mkTestPCDepends (Just [("pkgA", "1.0.0"), ("pkgB", "1.0.0")]) dbPC1 "pruneNotFound" ["C"] (solverSuccess [("A", 1), ("B", 1), ("C", 1)])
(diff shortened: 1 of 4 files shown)


Mikolaj (approved):
Another one-liner that solves world peace? Sounds legit. :D

But seriously, this could work. Any chance for a test and changelog?

gbaz:
Changelog, sure. I'm not quite sure how one might add a test for something like this, since it explicitly involves pkg-config not being on the path, etc. Advice in that direction is appreciated.

grayjay:
> I'm not quite sure how one might add a test for something like this, since it explicitly involves pkg-config not being on the path, etc.

You could add unit tests for solving without pkg-config, similar to the existing pkg-config tests: https://github.com/haskell/cabal/blob/55c426f4e5c63654b21660d2e17e41383d48f5bb/cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs#L255-L260

It looks like all of the tests create a pkg-config DB by calling pkgConfigDbFromList , so you would need to add an option for no DB: https://github.com/haskell/cabal/blob/55c426f4e5c63654b21660d2e17e41383d48f5bb/cabal-install/tests/UnitTests/Distribution/Solver/Modular/DSL/TestCaseUtils.hs#L222

gbaz:
Thanks! I implemented tests as suggested

cabal-install/tests/UnitTests/Distribution/Solver/Modular/Solver.hs
```diff
@@ -253,10 +253,11 @@ tests = [
         , runTest $ mkTest dbBuildable2 "choose version that sets buildable to false" ["A"] (solverSuccess [("A", 1), ("B", 2)])
          ]
     , testGroup "Pkg-config dependencies" [
-          runTest $ mkTestPCDepends [] dbPC1 "noPkgs" ["A"] anySolverFailure
```
grayjay, line 256: I think it would be useful to also keep the empty pkg-config test case, using `Just []`.

grayjay (approved):

fgaz:
> * [ ] The documentation has been updated, if necessary.

:eyes:

Issue #2612, opened 2015-05-27
expose more global options in cabal --help ?


gbaz:
Currently for global options, in Setup.hs we have the line "showOrParseArgs of ShowArgs -> take 6"

This means that the options for "remote-repo" "remote-repo-cache" "local-repo" "logs-dir" and "world-file" are all hidden.

It seems to me that as our infrastructure has evolved, these options are not as exotic as they once were, and perhaps should all be exposed by the --help command? This can be fixed by just swapping "take 6" for id there :-)

gbaz:
note that --store-dir also doesn't appear in global help. That should definitely be added too! (and of course world-file is very dead and probably should be ripped out if it hasn't been yet)

jneira:
`--store-dir` definitely must be listed

phadej:
Probably no, as this doesn't sound like a very big task.

Issue #5346, opened 2018-05-26
--offline flag doesn't seem to affect new-build


chris-martin:
This is an excerpt from the output of running `cabal new-build --offline`:

```
Downloading th-abstraction-0.2.6.0...
Configuring ansi-terminal-0.8.0.4 (lib)...
Configuring cookie-0.4.4 (lib)...
Configuring bsb-http-chunked-0.0.0.2 (lib)...
Configuring blaze-builder-0.4.1.0 (lib)...
Downloading wai-extra-3.0.22.1...
Downloading scotty-0.11.1...
Building bsb-http-chunked-0.0.0.2 (lib)...
Building cookie-0.4.4 (lib)...
Building ansi-terminal-0.8.0.4 (lib)...
```

Either the `--offline` flag isn't working, or that is some misleading build output?

```
$ cabal --version
cabal-install version 2.2.0.0
compiled using version 2.2.0.1 of the Cabal library 
```

Here's another trial, this time with my network link disabled and `~/.cabal` directory cleared:

```
$ cabal new-build --offline
Config file path source is default config file.
Config file /home/chris/.cabal/config not found.
Writing default configuration to /home/chris/.cabal/config
Warning: Caught exception during _mirrors lookup:res_query: does not exist
(Connection refused)
Warning: No mirrors found for http://hackage.haskell.org/
dieVerbatim: user error (cabal: Couldn't establish HTTP connection. Possible cause: HTTP proxy server
is down.
```

Again here it seems like it's trying to make HTTP requests despite the `--offline` flag which I thought was supposed to be an instruction to the contrary.

For context: My Nix shell environment should be providing all of the necessary dependencies, so I don't ever need or want cabal-install to download anything.

23Skidoo:
Looks like an oversight.

parsonsmatt:
This is affecting our workflow at the day job. Sometimes folks will update our `.cabal` file with a new dependency that isn't provided by `nix`. We'd like for the build to fail. Instead, it tries to download and build the missing dependency from Hackage. Passing `--offline` would fix this, but it doesn't appear to work.

fgaz:
@parsonsmatt if you have cabal >= 3.4, you can use [`active-repositories`](https://cabal.readthedocs.io/en/3.4/cabal-project.html#cfg-field-active-repositories):

```cabal
-- "Offline" mode
active-repositories: none
```

Maybe `--offline` could just be sugar for that

gbaz:
[...]

PR #7624, opened 2021-09-06
change documented global flags

gbaz:
Resolves https://github.com/haskell/cabal/issues/2612

Motivated by https://github.com/haskell/cabal/issues/5346 this exposes a few more global flags in `cabal --help` -- particularly active-repositories, and store-dir, which are both useful. It also hides default-user-config which is a v1-freeze thing that I think never acquired much traction. And while we're at it, documents a bit what the flags that aren't exposed are, to help with future cleanup efforts.


jneira (approved):
LGTM, thanks for clean up flags

Mikolaj (approved):
LGTM

PR #7625, opened 2021-09-07
Support Git submodules in source-package-respository

akshaymankar:
This PR tries to solve #5536. I tried to use `--recurse-submodules` flags in `clone` and `reset` but they don't work very well when the repo is cloned with `--no-checkout`. So, I just implemented the solution discussed in the issue, this includes changing `checkout` to `reset --hard`.

---

I manually tested this in two ways:
1. Without a dist-newstyle in a project which depends on a repository with a submodule. I used `cabal` from this fork and see if the repository gets cloned correctly and the project compiles.
2. With an already existing dist-newstyle (compiled by cabal 3.6.0.0) of the same project. This dist-newstyle had the previously cloned the dependency without the submodules checked out. After upgrading to this fork, it was able to use the old directory without any problems.

I am not sure where to add automated tests for this. I would be happy to add them if someone points me to the place to add them.


Mikolaj:
If you run `git grep -30 "git'\? "`, does any of tests that turn up seem a good candidate to copy-paste and modify? I'd avoid tests that invoke v1- commands and probabably also avoid Setup commands, but stick to v2- commands instead.

akshaymankar:
@Mikolaj Thanks for the pointer. I found some property tests and added submodule support in there. It uncovered an edge case in git's handling of submodules. Fortunately I was able to work around that :smile:

Mikolaj:
@akshaymankar: do the CI failures look relevant or is this only some transient breackage?

akshaymankar:
Looks like GHC 8.8 and below couldn't infer the kind of the type level variable I introduced. I have added an explicit annotation now. It at least compiles with 8.8.

akshaymankar:
The older git (2.17.1) in the bionic container is stricter about `git submodule foreach` argument ordering than the one on my computer (2.32.0). I have fixed that. 

I also noticed that I wasn't respecting the `verboseArg` in the new commands, I have also fixed that.

cabal-install/src/Distribution/Client/VCS.hs
```diff
@@ -386,17 +386,18 @@ vcsGit =
     vcsCloneRepo verbosity prog repo srcuri destdir =
         [ programInvocation prog cloneArgs ]
         -- And if there's a tag, we have to do that in a second step:
-     ++ [ (programInvocation prog (checkoutArgs tag)) {
-            progInvokeCwd = Just destdir
-          }
-        | tag <- maybeToList (srpTag repo) ]
+     ++ [ git (resetArgs tag) | tag <- maybeToList (srpTag repo) ]
+     ++ [ git (["submodule", "sync", "--recursive"] ++ verboseArg)
```
Mikolaj, line 390: How much does this, and other more complex git commands you introduced, slow down the cases where there are no submodules or where submodules are handled fine already (are there such cases?). What other risks there are from these more complex commands?

Mikolaj (commented):
Thank you so much for the tests.

cabal-install/src/Distribution/Client/VCS.hs
akshaymankar: > How much does this, and other more complex git commands you introduced, slow down the cases where there are no submodules?

I just tried timing the new commands on the cabal repo and got:

```
$ time git submodule deinit --force --all
git submodule deinit --force --all  0.03s user 0.04s system 113% cpu 0.056 total

$ time git submodule sync
git submodule sync  0.04s user 0.03s system 109% cpu 0.062 total

$ time git submodule update --init --recursive --force
git submodule update --init --recursive --force  0.03s user 0.04s system 109% cpu 0.066 total

$ time git submodule foreach --recursive "git clean -ffdxq"
git submodule foreach --recursive "git clean -ffdxq"  0.03s user 0.04s system 110% cpu 0.061 total

$ time git clean -ffdxq
git clean -ffdxq  0.00s user 0.01s system 96% cpu 0.011 total
```

So, with this imprecise test, it seems like this PR may have added extra latency of ~0.15 seconds per repository (which doesn't have submodules).

>  or where submodules are handled fine already (are there such cases?)

I don't think there were any cases where submodules worked.

> What other risks there are from these more complex commands?

I can see this becoming a breaking change for some setup where a package repository submodules a private repository which is not needed for compiling haskell. To support this, a flag might be required to disable submodule fetching. 

I can't think of any other risks.

Mikolaj: > I can see this becoming a breaking change for some setup where a package repository submodules a private repository which is not needed for compiling haskell. To support this, a flag might be required to disable submodule fetching.

How do you feel about waiting with that flag until we get actual reports that people need it? Is there a workaround for such people in the absence of the flag?

Mikolaj (approved):
LGTM and the overhead and risks seem acceptable, though my knowledge of the matter is limited (e.g., I can't imagine corner cases and I have no idea how many projects currently use the features and may either benefit from this PR or may need to adjust to it).

cabal-install/src/Distribution/Client/VCS.hs
akshaymankar: I am OK with waiting for the flag. 
As a work around maybe people could maintain forks of of depedencies which remove such submodules, but this would be very inconvenient for repositories which need to be updated a lot.

Mikolaj:
If nobody reviews this soon, I'm going to abuse my privilege and merge it alone. Please don't make me a worse person.

cabal-install/src/Distribution/Client/VCS.hs
jneira: So finally we will not have the flag to hide the new feature?

Out of curiosity, did you test it (performance oriented) with large git repos with many submodules?

jneira:
> 
> 
> If nobody reviews this soon, I'm going to abuse my privilege and merge it alone. Please don't make me a worse person.

i am gonna try it in windows with a hls version which had submodules: https://github.com/haskell/haskell-language-server/tree/e4f677e1780fe85a02b99a09404a0a3c3ab5ce7c

jneira:
The build has been succesful in my windows machine. :100: 
However i've observed the submodules are initialized even if you dont use a submodule as `subdir` or you dont have `subdir`defined at all, would be ask for too much only trigger the submodules stuff only when the subdir is one of them (afaik it would not be needed in other case)?
That would make the perf penalty not noticed for existing use cases, which cant use submodules right now.

jneira:
> 
> 
>  would be ask for too much only trigger the submodules stuff only when the subdir is one of them (afaik it would not be needed in other case)? That would make the perf penalty not noticed for existing use cases, which cant use submodules right now.

well a corner case would be use a s-r-p without subdir which has source dirs (or other dirs needed for the build) as submodules (or use a subdir which has source dirs as submodules) :thinking:

Mikolaj:
BTW, I've been told this project is a good test of a possible slowdown for projects with no submodules, because it has so many `source-repository-package` which are slow already, in particular when rebuilding things:

https://github.com/input-output-hk/plutus-pioneer-program/blob/main/code/week06/cabal.project

If this test shows there is no problem, I'm fine accepting the PR and we'd worry when users with unused submodules complain (but they may be willing to accept some slowdown in unused modules for convenience in used submodules; also the scheme with submodules tends to make caching rebuilds harder, so that's not the route to take if somebody prioritizes (re)compilation speed --- the link above is the current optimum, as verbose as it is).

jneira:
@akshaymankar i labeled with info-needed to relfect @Mikolaj request for manual testing perf over a repo with lot of s-r-p

Mikolaj:
@akshaymankar: I'm aware this review drags on for a month now. Please let us know if you have a deadline or if you are just fed up already. :) We are quite stretched, hence the delay, but we appreciate contributions all the more and we would hate to discourage yours.

akshaymankar:
I understand that OSS contributions can take some time to get responses. So, I am not fed up yet :smile: 
Unfortunately, I have had to travel on short notice and I am working and living in different time zones, so I have not had time to look at this. I will try to get to this by the end of the weekend :crossed_fingers:

Issue #7544, opened 2021-08-14
RFC: Exact-printer Mega-issue


emilypi:
# What is this issue? 

A central place to discuss the issue of a Cabal exact-printer so that we can push forward the work all in one place. Currently, The discussion stretches back across many issues over the past 6 years. 

## What is it? 

An *exact-printer*, as inspired by `[ghc-exactpring](https://hackage.haskell.org/package/ghc-exactprint)`, is a byte-for-byte bidirectional parser and pretty printer, which gaurantees the following constructs exist and are principled: 

- A `.cabal` file source tree representation, which is in bijection with the package description for a `.cabal` file for a given spec version. This representation should be parametrized with all offset and encoding information along with file data. 
- A method for introducing deltas to `.cabal` file entries. I should able to update and transform individual components of data deep in the representation without modifying the broader structure. 
- A pretty printer which adds offset annotations to the specification fragment so that the output should have a byte-for-byte round trip on no-op parse/print and print/parse tests.
- Tests which confirm this in `cabal-testsuite` and `cabal install`'s test suite for the individual commands. 

## Why do we need it? 

This work is tied very closely with the `format`, `init`, `gen-bounds` and other work as shown in historical discussions: 

 - https://github.com/haskell/cabal/issues/3614
 - https://github.com/haskell/cabal/pull/6621
 - https://github.com/haskell/cabal/issues/6187
 - https://github.com/haskell/cabal/issues/4965
 - [etc...](https://github.com/haskell/cabal/labels/exact-print)

This would free us up for several important quality of life improvements to the `cabal install` ecosystem, as well as general consumers of the `Cabal` library API. We would like to have this in by `Cabal-3.8.0.0`. 

## Who's in charge of this? 

I am currently overseeing @ptkato who has been tasked with taking this on. Please consider him and myself a point of contact for this work, and join us in `libera.chat#hackage` to discuss. 

@ptkato @gbaz @davean

gbaz:
We have an intermediate "lexed tree" structure in https://github.com/haskell/cabal/blob/1be581d5463adebf8dc8cc8a5375a9ca3051970a/Cabal/src/Distribution/Fields/Field.hs

[...]

PR #7626, opened 2021-09-07
cabal exact-printer

ptkato:

---

This PR aims to centralise the efforts done toward our goal of coming up with a exact-printer for cabal.

Refer to #7544 for more detailed information and discussion.


pranaysashank:
Should this go in the upcoming `Cabal-Syntax` package?

Cabal/src/Distribution/Fields/Field.hs
```diff
@@ -42,12 +44,14 @@ import           Prelude ()
 -- | A Cabal-like file consists of a series of fields (@foo: bar@) and sections (@library ...@).
 data Field ann
     = Field   !(Name ann) [FieldLine ann]
+    | Comment !ann !(CommentLine ann)
```
phadej, line 47: Why?

Comments can be extracted separately? See how `cabal-fmt` does  the job. (Also GHC AST is not polluted with comments, they are attached afterwards. That's why there is `ann` in the `Field` structure too).

pranaysashank: This was just something I wrote for quick prototyping. I'll checkout both `cabal-fmt` and `ghc-exactprint`

ptkato:
@pranaysashank, likely, however it's best we wait for #7620 to be merged first.

Cabal/src/Distribution/Fields/Field.hs
phadej: We should keep in mind that most time this code is hit when solver is traversing dozens if not hundreds of cabal files. The less stuff is done in hot path, the more responsive `cabal-install` is.

The "interactive" usage should not make compromises to the primary function. I'd argue that if it becomes a friction then having two separate parsers is not impossible idea either.

ptkato: Hmm... it's something to ponder; while having a solid base to build on top of is great, having a clean slate to shape the feature to our current needs, while not affecting the older code, is very appealing.

nomeata:
https://github.com/haskell/cabal/pull/7620 has been merged – what’s the status of this PR?

Mikolaj:
@ptkato: thanks for this version of the PR. Do you intend to work on it in the future?

ptkato:
@Mikolaj I'd like to, but at the moment I can't tackle something this size.

Mikolaj:
@ptkato: thanks. No pressure, just asking.

I wonder if perhaps this can be split into smaller parts and tackled gradually by multiple people.

ffaf1:

Issue #7619, opened 2021-09-04
`cabal-install` only works when executed in a project or package root


dschrempf:
When I execute `cabal` in a sub-folder of a project or package not containing a `*.cabal` or `cabal.project` file, the call fails with

```
cabal run

cabal: There is no <pkgname>.cabal package file or cabal.project file. To
build packages locally you need at minimum a <pkgname>.cabal file. You can use
'cabal init' to create one.

For non-trivial projects you will also want a cabal.project file in the root
directory of your project. This file lists the packages in your project and
all other build configuration. See the Cabal user guide for full details.
```

`cabal-install` should realize that this call happend within a cabal project or package, and use this configuration.

I use `cabal-install` version 3.4.0.0.

dschrempf:
This also affects Cabal scripts.

I have a script accessing a library called `mcmc-date` defined in the very same package. If this script is located in a sub-folder, let's say `scripts`, it does not work, and fails with:
```
 + ./trees-midpoint-root
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: fake-package-0 (user goal)
[__1] unknown package: mcmc-date (dependency of fake-package)
[__1] fail (backjumping, conflict set: fake-package, mcmc-date)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: fake-package, mcmc-date
```

If I move the script to the project root folder it works just fine, and the project-specific library `mcmc-date` is found. If I execute `./scripts/trees-midpoint-root` from the project root, the script also works!

fgaz:
> When I execute cabal in a sub-folder of a project or package not containing a *.cabal or cabal.project file

I'll interpret that as "where the subfolder does not containin a *.cabal or cabal.project file", please correct me if I interpreted it wrong.

The behaviour you describe is normal in the case of the `.cabal` file (there might be a `cabal.project` file further up in the tree after all), but it should work for projects that include a `cabal.project` file. Are you sure it happens in both cases? I can't reproduce it with 3.4

dschrempf:
Ah, no, you are right! It only happens with cabal packages. When I add a `cabal.project` file it works also in sub-folders. Thank you. Should we do something about this? I guess others will run into similar issues with cabal packages. Or is this intended behavior?

fgaz:
It's intended (hence the error "No cabal.project file or cabal file **matching the default glob './*.cabal'** was found.")

[...]

PR #7627, opened 2021-09-08
Improve the no cabal file found error

SamuelSchlesinger:
Just a small improvement based on: https://github.com/haskell/cabal/issues/7619

### DIFF
--- a/cabal-install/src/Distribution/Client/ProjectConfig.hs
+++ b/cabal-install/src/Distribution/Client/ProjectConfig.hs
@@ -754,9 +754,9 @@ renderImplicitBadPackageLocation bpl = case bpl of
     BadLocGlobEmptyMatch pkglocstr ->
         "No cabal.project file or cabal file matching the default glob '"
      ++ pkglocstr ++ "' was found.\n"
-     ++ "Please create a package description file <pkgname>.cabal "
-     ++ "or a cabal.project file referencing the packages you "
-     ++ "want to build."
+     ++ "No <pkgname>.cabal was found and the current working directory is not "
+     ++ "part of a cabal project. To build (or run) local packages you need a cabal file "
+     ++ "at least. You can use 'cabal init' to create one."
     _ -> renderBadPackageLocation bpl
 
 renderBadPackageLocation :: BadPackageLocation -> String


cabal-install/src/Distribution/Client/ProjectConfig.hs
```diff
@@ -754,9 +754,9 @@ renderImplicitBadPackageLocation bpl = case bpl of
     BadLocGlobEmptyMatch pkglocstr ->
         "No cabal.project file or cabal file matching the default glob '"
```
jneira, line 755: This one does not match well with `No <pkgname>.cabal was found and the current working directory is no...`

jneira:
Many thanks for trying to fix the error. Error messages can be tricky, it is difficult to be precise and helpful at same time.

cabal-install/src/Distribution/Client/ProjectConfig.hs
jneira: Maybe an improvement would be print the default glob when no explicit cabal.project exists which afaik is `./*.cabal`. In fact `matching the default glob` talks about the cabal file and no the cabal.project file (it is a little bit ambiguous).

https://github.com/haskell/cabal/blob/b26e736a04da5fca30dfa50ea8cf16abf1c8314c/cabal-install/src/Distribution/Client/ProjectConfig.hs#L537

```suggestion
        "No cabal.project file in the current dir nor a cabal file matching the default glob (./*.cabal) '"
```

but making `./*.cabal` a constant used here and in the previous code using it.

cabal-install/src/Distribution/Client/ProjectConfig.hs
```diff
@@ -754,9 +754,9 @@ renderImplicitBadPackageLocation bpl = case bpl of
...
      ++ pkglocstr ++ "' was found.\n"
-     ++ "Please create a package description file <pkgname>.cabal "
-     ++ "or a cabal.project file referencing the packages you "
-     ++ "want to build."
+     ++ "No <pkgname>.cabal was found and the current working directory is not "
+     ++ "part of a cabal project. To build (or run) local packages you need a cabal file "
```
jneira, line 758: Not sure if a directory can be part of a cabal.project without having a .cabal file.
In my tests i am getting other errors if i list a subdirectory in the `packages`field of a cabal.project and that dorectory has no .cabal file

cabal-install/src/Distribution/Client/ProjectConfig.hs
```diff
@@ -754,9 +754,9 @@ renderImplicitBadPackageLocation bpl = case bpl of
...
+     ++ "at least. You can use 'cabal init' to create one."
```
jneira, line 759: Lets take the case of the linked issue:
* you have cabal scripts in the ./scripts directory
* `cabal run ./scripts/file` works but `cd scripts && cabal run file` does not. The suggestion of make a cabal init would be wrong for that case where the right think is simply run the script from the parent folder, no?

cabal-install/src/Distribution/Client/ProjectConfig.hs
jneira: In fact this would make the error message more similar to the thrown by `cabal build`

jneira:
@SamuelSchlesinger do you need any help to make progress on that, any thought about my comments? thanks!

PR #7628, opened 2021-09-08
Allow text-2.0 in Cabal.cabal only

Bodigrim:
`text-2.0` is not yet released, but I would like to update `Cabal` (not `cabal-install`) HEAD so that we can bump `text` submodule in GHC source tree and start patching of `head.hackage` to iron out potential migration issues before a final release. Cf. https://github.com/haskell/parsec/pull/135

`Cabal` package does not really delve into `text` deep enough to be affected by `text-2.0` changes, and I checked all usages manually. 


emilypi (approved):

emilypi:
@fgaz do you think we should bump on `cabal-testsuite` as well? If it's not required, we can do this in a follow-up and unblock @Bodigrim.

gbaz (approved):

PR #7629, opened 2021-09-09
cabal-install.cabal: use -fhide-source-paths for ghc-8.2+

juhp:
This should only affect builds of cabal-install


phadej:
IMO these belongs to `cabal.project(.local)`.

emilypi:
I tend to agree with @phadej here. Line noise tends to not be an issue people bring up, and I'd hate to impose (as nice as the output becomes) by modifying the global `cabal-install.cabal` file. Does this solve a particular problem for you @juhp?

juhp:
I am going to try using the option for all Fedora builds.

emilypi:
Cool - I'd be willing to accept this on the basis that it bricks Fedora. Let us know what the outcome is :)

Issue #6520, opened 2020-01-31
Fix password issue for curl transport


newhoggy:
**Describe the bug**
Currently when using `cabal` with `curl` transport with private hackage, `curl` will submit `...` as the password and fail authentication.  To work around this issue, we've had to use `wget` which has the downside of slow `cabal` updates because `wget` doesn't support range requests.

**To Reproduce**
Upload a package to password protected private hackage then try to get `cabal` to download it.

When using `wget` transport:

```
 cabal v2-update
Downloading the latest package lists from:
- hackage.haskell.org
- phackage
Warning: the 'wget' transport currently doesn't support range requests, which
wastes network bandwidth. To fix this, set 'http-transport' to 'curl' or
'plain-http' in '~/.cabal/config'. Note that the 'plain-http' transport
doesn't support HTTPS.
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2020-01-31T06:17:33Z'
```

Steps to reproduce the behavior when using `curl` transport:

```
$ cabal get customer-data
Downloading  customer-data-1.0.7
Unexpected response 401 for https://mayhem:...@hackage.mayhem.arbor.net/package/customer-data-1.0.7.tar.gz
```

This works when using `wget` transport:

```
$ cabal get customer-data
Downloading  customer-data-1.0.7
Downloaded   customer-data-1.0.7
Unpacking to customer-data-1.0.7/
```

**Expected behavior**

**System information**
 - Operating system: MacOS (probably any system)
 - `cabal-3.0.0.0`, any `ghc` version

**Additional context**
None

newhoggy:
Possible fix here #6184

dsturnbull:
Related: https://github.com/haskell/cabal/issues/4743

gbaz:
resolved by above

PR #6184, opened 2019-08-09
pipe auth into curl

dsturnbull:
Addresses #4743

Same pattern as in the wget transport: pipe auth config into the curl process if auth info is found in the repo URI.

Added embedded auth in the repo uri, chose the `curl` transport and ran `v2-update`:

```
Downloading the latest package list from private
To revert to previous state run:
    cabal v2-update 'private,2019-08-07T00:56:21Z'
```

---
Please include the following checklist in your PR:

* [x] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [x] Any changes that could be relevant to users have been recorded in the changelog.
* [x] The documentation has been updated, if necessary.
* [ ] If the change is docs-only, `[ci skip]` is used to avoid triggering the build bots.

PR #7630, opened 2021-09-09
pipe auth into curl (v2)

gbaz:
Merge of https://github.com/haskell/cabal/pull/6184 with latest master


gbaz:
Resolves #6520 , resolves #4743

Mikolaj (approved):

emilypi (approved):

andreasabel:
Q1: Where is the test that demonstrates the problem to solve and that it is solved by the patch?
Q2: Which existing tests could be broken by changing the communication with the hackage server, like this patch does?

I also think that __comments in the code__ that explain what the problem was, and how it is solved by the changed would be very helpful for code comprehension, as well as links to issues __in the code__.

gbaz:
Ok, I think I see what's wrong, and why its my fault for not reviewing this better, ugh. The issue is that the `uriAuthority` of a URI is a `Just` instead of a `Nothing` even when there's no user login info. So I had thought this patch was safe because it would only add new behavior on the `user:password@server.com` case (which had no handling before -- so this would be strictly better) -- but it turns out the case matches even without the u/p pair. Thanks for catching the problem -- I'll make a PR with appropriate changes. Not sure if we have any good `curl` tests or how to add them, sadly.

PR #7638, opened 2021-09-11
CI for building doc with sphinx

andreasabel:
Simplified from agda/agda/.github/workflows/user-manual.yml

Simple minded.  Builds doc by
- installing python
- `make users-guide`
- upload-artifact

Maybe something more sophisticated could be done later where we push to `ghpages` or something.

Tested by downloading the artifact and having a rough look.

See new job __Users guide__ in the checks below.  Should maybe also switched to "Required".

PR #7647, opened 2021-09-14
Testing readthedocs' pr integration (do not merge)

fgaz:


fgaz:
@Mikolaj it works \o/

https://readthedocs.org/projects/cabal/builds/14708732/

Mikolaj:
Yay! How do we wire it into CI? Or do we keep the new CI docs check in CI and use this one only for people that want to see the actual results of doc changes? Does it email somebody when it fails or just show a broken page?

fgaz:
We could mark it as required and remove the github actions one. Before that we'd have to reflect the flags from the makefile to readthedocs, eg. https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning

> Does it email somebody when it fails or just show a broken page?

it shows a failed check in the CI section like github actions

Mikolaj:
Got it. That's great news. Let's keep them both until we are sure they work consistently and the check by @andreasabel is completely subsumed by the ultimate check using Read The Docs.

Mikolaj:
Or does #7638 introduce more/stricter checks and they can't be ported to readthedocs server proper? Or can/should it introduce stricter checks? Is it worth keeping another CI job?

fgaz:
I don't know yet, I didn't check in depth

andreasabel:
@fgaz You could try to run it on a PR that broke links in the docs, e.g., #7498.  
To prevent doc formatting issues (broken links etc.) I think there should be a CI check that fails in that case.

PR #7648, opened 2021-09-14
Change cabal-plan list-bin to cabal list-bin in Makefile

fgaz:
cabal list-bin was introduced in cabal-install 3.4, so it's safe to
assume every cabal _developer_ has it by now. By using that, we
remove the dependency on cabal-plan from the Makefile (though it's still
present in other places).

---

---

I'm not sure if CI will be fine with this though. some jobs are run with older images. on the other hand, CI doesn't use many make commands. let's see...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ $(TEMPLATE_PATHS) : templates/Paths_pkg.template.hs cabal-dev-scripts/src/GenPat
 
 buildinfo-fields-reference : phony
 	cabal build --builddir=dist-newstyle-bi --project-file=cabal.project.buildinfo buildinfo-reference-generator
-	$$(cabal-plan list-bin --builddir=dist-newstyle-bi buildinfo-reference-generator) buildinfo-reference-generator/template.zinza | tee $@
+	$$(cabal list-bin --builddir=dist-newstyle-bi buildinfo-reference-generator) buildinfo-reference-generator/template.zinza | tee $@
 
 # analyse-imports
 analyse-imports : phony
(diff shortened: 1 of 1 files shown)


jneira (approved):
LGTM

andreasabel:
@fgaz 
Broke for me (macOS):
```
$ make cabal-install-test-accept 
rm -rf .ghc.environment.*
cd cabal-testsuite && `cabal list-bin cabal-tests` --with-cabal=`cabal list-bin cabal` --hide-successes -j3 --accept 
Error: cabal: No or multiple targets given

/bin/sh: --with-cabal=/Users/abel/bin/src/cabal/dist-newstyle/build/x86_64-osx/ghc-8.10.7/cabal-install-3.7.0.0/x/cabal/build/cabal/cabal: No such file or directory
make: *** [Makefile:137: cabal-install-test-accept] Error 127

$ cabal list-bin cabal-tests
Error: cabal: The list-bin command is for finding a single binary at once. The
target 'Cabal-tests' refers to the package Cabal-tests-3 which includes the
test suite 'unit-tests', the test suite 'rpmvercmp', the test suite
'parser-tests', the test suite 'no-thunks-test', the test suite
'hackage-tests', the test suite 'custom-setup-tests' and the test suite
'check-tests'.

$ cabal-plan list-bin cabal-tests
/Users/abel/bin/src/cabal/dist-newstyle/build/x86_64-osx/ghc-8.10.7/cabal-testsuite-3/build/cabal-tests/cabal-tests
```
Could be a problem with case-insensitivity.

@fgaz After our conversation where you mentioned that `cabal-plan list-bin` was obsolete, I noticed already on my machine that it worked more reliable than `cabal list-bin` (which in this instance wanted to build all kinds of stuff rather than simply spitting out the name of the binary).

Suggestion: revert this PR.  (I have to do it anyway on my machine.)

phadej:
case insensitive file system strikes again. So dumb.

fgaz:
Even then, there's also a problem in... sensible file systems, _and_ when a qualified target is used:

```
> cabal list-bin cabal-testsuite:exe:cabal-tests
cabal: No or multiple targets given
```

This definitely looks like a bug in list-bin, _and_ a bad error message: is it zero, or multiple?

This is not fixed by #7408 by the way.

So yes, let's revert for now. I also opened #7679 to track the bug.

P.S.: yes, I originally only checked a couple of invocations and assumed it worked the same for all others, sorry..

Issue #7649, opened 2021-09-14
cabal-install-3.6 fails to install network-3.1.2.2 in windows


jneira:
* but cabal-install-3.4.0.0 is able to build it
* logs:

```
Configuring network-3.1.2.2...
configure: WARNING: unrecognized options: --with-compiler
checking build system type... x86_64-pc-msys
checking host system type... x86_64-pc-msys
checking for gcc... D:\ghcup\ghc\8.10.7\lib\../mingw/bin\gcc.exe
checking whether the C compiler works... yes
checking for C compiler default output file name... a.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether D:\ghcup\ghc\8.10.7\lib\../mingw/bin\gcc.exe accepts -g... yes
checking for D:\ghcup\ghc\8.10.7\lib\../mingw/bin\gcc.exe option to accept ISO C89... none needed
checking for an ANSI C-conforming const... yes
checking how to run the C preprocessor... D:\ghcup\ghc\8.10.7\lib\../mingw/bin\gcc.exe -E
[...]

PR #7650, opened 2021-09-14
Generate Paths_ module with qualified Data.List.last import

alexbiehl:
---

### DIFF
--- a/Cabal/src/Distribution/Simple/Build/PathsModule/Z.hs
+++ b/Cabal/src/Distribution/Simple/Build/PathsModule/Z.hs
@@ -60,6 +60,7 @@ render z_root = execWriter $ do
     return ()
   tell "\n"
   tell "import qualified Control.Exception as Exception\n"
+  tell "import qualified Data.List as List\n"
   tell "import Data.Version (Version(..))\n"
   tell "import System.Environment (getEnv)\n"
   tell "import Prelude\n"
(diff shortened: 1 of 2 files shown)


Mikolaj:
@alexbiehl: this is great for a quickfix, but what do you think about trying not to use Prelude at all and then adding a test that indeed turns off Prelude and compiles Paths_ so that don't stumble into this problem again?

Also, whatever happend to the previous PR? :)

phadej:
turning of prelude can be done with `NoImplicitPrelude` in the `Paths_` file (when such extension is available), and doing `import Prelude ()` otherwise.

jneira:
[After some problems to build cabal from this commit](https://github.com/haskell/cabal/issues/7649), i've tried with the policeman package and it worked 👍

jneira:
This is the minimum change (so minimum risk to break things) which fixes it so imo we could merge this and do the definitive fix in another one on top of it

alexbiehl:
Thanks @jneira for trying it on the Kowainik packages!!

Mikolaj:
> @alexbiehl: this is great for a quickfix, but what do you think about trying not to use Prelude at all and then adding a test that indeed turns off Prelude and compiles Paths_ so that don't stumble into this problem again?
> 
> Also, whatever happend to the previous PR? :)

@alexbiehl: ping? we should probably stick this into 3.6.2 RSN, so what's your recommendation about the wider context?

alexbiehl:
@Mikolaj My recommendation is to merge as is. Refactoring the code to avoid Prelude will take more time and I'd need to spend substantial time on it.

Mikolaj:
Got it. Let's do that.

Mikolaj (approved):

Mikolaj:
@mergifyIO backport 3.6

jneira (approved):
lgtm. many thanks for the fix

PR #7657, opened 2021-09-17
Regenerate bootstrap/ with GHC updated from 8.10.4 to 8.10.7 

Mikolaj:
This regenerates all generated bootstrap/ files, after changing GHC from 8.10.4 to 8.10.7 in Makefile. The goal is to fix CI that has broken (both on 3.6 and master) because GitHub Action no longer provide GHC 8.10.4 (we will see if the provide 8.10.7) and to provide 8.10.7 files on @bgamari 's request (Darwin/Arm are only supported on 8.10.7).

Transcript from IRC/Matrix:

> mikolaj
> Q: how do I actually build a file like bootstrap/linux-8.10.4.plan.json, but for 8.10.7? is there CI job for this or a script I need to run locally or a tiny Haskell program, e.g., bootstrap/cabal-bootstrap-gen.cabal?
> hmm, bootstrap/README.md suggests all of the above plus a Makefile
> plus manual tweaking of the resulting json :D
> yes, there is bootstrap-plans-linux in the Makefile
> and bootstrap-jsons-linux, but I don't know if it's run by CI somewhere or if I need to install 3 different GHCs and run it (I will probably only tweak and run the portion for 8.10.7, unless somebody suggests otherwise, while I'm taking a break)
> bgamariasked me to update the boostrap plans for 3.6.2, so even if this doesn't fix CI (because GHA dropped 8.10 completely), it will still be worthwhile
> bgamari
> indeed it would be good to have a bootstrap option for 8.10.7
> since at the moment that is the only release that supports Darwin/ARM
> mikolaj
> I guess I can remove the one for 8.10.4? or let it be? no idea if they'd conflict in any way
> mikolaj
> nobody objects, so I'm removing bootstrap for 8.10.4
> mikolaj
> oh, and I probably should have run the scripts on branch 3.6., not master, right? these are bootstraps for 3.6 (and older cabals for the older GHCs), not for unstable, right?
> which makes me wonder if I should perhaps refresh the other files, too, for older GHCs, so that all bootstraps are for cabal 3.6 --- after all, they accompany the 3.6 (and later) code, not the old code, so quite possibly they no longer work
> again, unless somebody objects before I return from another break, I'm going to re-create the boostraps on branch 3.6, for GHC 8.6.5, 8.8.4 and 8.10.7 (and gut out 8.10.4)
> thank you for your cooperation
> ugh, but for this I need to install two more GHCs
> fortunately both are on ghcup
> mikolaj
> the linux-8.10.7.json file generated on master is surprising different from the one on 3.6., so I think this is a good call to make it on 3.6 and also to regenerate the old ones (which do differ as well)
> mikolaj
> oh, and actually the bootstrap files on 3.6 seem to have been regenerated by tapas at some point, while those on master come from some other users (probably much earlier)
> tapas: I hope you do not mind me regenerating your files and also that I will copy them to master as well


Mikolaj:
This is ready to merge, please review. The CI failures are normal. This PR is the first aiming to eliminate them and gets rid of two job failures.

emilypi (approved):

.github/workflows/bootstrap.yml
```diff
@@ -40,17 +40,17 @@ jobs:
       - name: Install GHC
         run: |
           cd $(mktemp -d)
-          curl -sLO "https://downloads.haskell.org/~ghc/8.10.4/ghc-8.10.4-x86_64-apple-darwin.tar.xz"
+          curl -sLO "https://downloads.haskell.org/~ghc/8.10.7/ghc-8.10.7-x86_64-apple-darwin.tar.xz"
```
hasufell, line 43: what's this? Why not use ghcup as well? It's also pre-installed on mac.

templates/ci-bootstrap.template.yml
```diff
@@ -40,17 +40,17 @@ jobs:
       - name: Install GHC
         run: |
           cd $(mktemp -d)
-          curl -sLO "https://downloads.haskell.org/~ghc/8.10.4/ghc-8.10.4-x86_64-apple-darwin.tar.xz"
+          curl -sLO "https://downloads.haskell.org/~ghc/8.10.7/ghc-8.10.7-x86_64-apple-darwin.tar.xz"
```
hasufell, line 43: Same as above

.github/workflows/bootstrap.yml
Mikolaj: Thanks, I will try that on the master branch version of this PR. This one I will leave in peace.

PR #7665, opened 2021-09-18
Bring back cabal-install-dev call removed by backport of #7653

Mikolaj:
I guess I mistakenly removed it when resolving a conflict in backport
of #7653 in c3354c96cbfdad3fac8afcce9c5806f77187bcee,
because on master these omitted lines are absent.
BTW, this backport was not reviewed, by my mistake.
A good example why careful reviews are essential.
Sadly, this PR will get in without reviews, too, because we are 
greatly lagging behind with reviews this week.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -37,6 +37,8 @@ jobs:
       - name: Update Hackage index
         run: cabal v2-update
       - uses: actions/checkout@v2
+      - name: make cabal-install-dev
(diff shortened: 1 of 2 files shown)


Mikolaj:
Prediction: CI will only fail in the 2 dogfood jobs. If so, I will merge and then #7664 will finish CI all green (now it fails at 2 normal Windows jobs). At which point I will merge it too, with reviews or without.

Mikolaj:
Only "MacOS / validate.sh macos ghc-8.6.5 (pull_request)" is still running, so I assume the prediction was correct and I'm merging without review.

Issue #7666, opened 2021-09-19
Regression: upload fails with "No authorization provided" (works with 3.4)


andreasabel:
Failure with `cabal` from `master`:
```
$ cabal upload -v dist-newstyle/sdist/BNFC-2.9.3.tar.gz
Uploading dist-newstyle/sdist/BNFC-2.9.3.tar.gz...
/usr/bin/curl --config - 'https://hackage.haskell.org/packages/candidates' --form 'package=@dist-newstyle/sdist/BNFC-2.9.3.tar.gz' --write-out '
%{http_code}' --user-agent 'cabal-install/3.6 (osx; x86_64)' --silent --show-error --header 'Accept: text/plain' --location
Error uploading dist-newstyle/sdist/BNFC-2.9.3.tar.gz: http code 401
Error: No authorization provided
```
Success with 3.4:
```
$ cabal-3.4.0.0 upload -v dist-newstyle/sdist/BNFC-2.9.3.tar.gz
Uploading dist-newstyle/sdist/BNFC-2.9.3.tar.gz...
/usr/bin/curl --config - 'https://hackage.haskell.org/packages/candidates' --form 'package=@dist-newstyle/sdist/BNFC-2.9.3.tar.gz' --write-out '
%{http_code}' --user-agent 'cabal-install/3.4.0.0 (osx; x86_64)' --silent --show-error --header 'Accept: text/plain' --location
Package successfully uploaded as candidate. You can now preview the result at
'https://hackage.haskell.org/package/BNFC-2.9.3/candidate'. To publish the
candidate, use 'cabal upload --publish'.
```

Mikolaj:
Uhoh. 1000$ question: does it work for you with cabal 3.6?

Mikolaj:
Assuming others can reproduce this, it would be good to check this recent merged PR (try with it and without): #7630.

andreasabel:
> Uhoh. 1000$ question: does it work for you with cabal 3.6?

I didn't have the released cabal-3.6, but installed it, and yes, it works there, but not with cabal from master.

Mikolaj:
Thank you so much for checking. So we at least don't have to scramble. And the mentioned PR is all the more a likely culprit.

andreasabel:
I tried running `upload` with `-v3` but this did not provide more insight, in particular, did not show any difference between 3.6 and master.

gbaz:
Thanks for the report, working on it now.

PR #7673, opened 2021-09-20
fix curl auth on upload

gbaz:
Resolves https://github.com/haskell/cabal/issues/7666

This fixes up the logic from #7630 which in turn was an update of #6184

I think this is a collective failure that we didn't review that PR well -- the logic was pretty busted all around, and we wrongly thought that worst case it wouldn't cause regressions, just not handle new functionality well.

I tried to rework the logic so I'm pretty sure its sensible, and I will test this by hand a fair amount. We're sorely lacking in transport tests (not least because we have four transports, etc), so any help in adding a good regression/sanity check around this test would be very welcome. We should not actually have our upload tests hit the running hackage server (even with candidates), so to test the upload rather than fetch functionality properly we'd arguably need a `curl` mock? ugh.

To give the general context -- some users have repos that are pass-authenticated -- so they want urls like `http://user@pass:repo.com` -- and they need that authentication to occur even for download, not upload. This I guess always worked with the plain http transport, when we moved to other transports it was added to wget, but not curl. The patches merged tried to add it to curl. However, they _overrode_ other auth info, even when no u/p pair was provided. This refactor tries to make the logical branches very explicit and clear, preserving standard behavior when there's no user info in the uri, or when a u/p pair are otherwise provided in any manner, and only falling back to the uri info when that's not around (such as in the case of a plain get request).

### DIFF
--- a/cabal-install/src/Distribution/Client/HttpUtils.hs
+++ b/cabal-install/src/Distribution/Client/HttpUtils.hs
@@ -406,25 +406,26 @@ curlTransport prog =
 
     posthttp = noPostYet
 
-    addAuthConfig auth uri progInvocation = do
-      case uriAuthority uri of
-        Just (URIAuth u _ _) -> progInvocation
-          -- all `uriUserInfo` values have '@' as a suffix. drop it.
-          { progInvokeInput = Just $ IODataText $ unlines $
+    addAuthConfig explicitAuth uri progInvocation = do
+      -- attempt to derive a u/p pair from the uri authority if one exists
+      -- all `uriUserInfo` values have '@' as a suffix. drop it.
+      let uriDerivedAuth = case uriAuthority uri of
+                               (Just (URIAuth u _ _)) | not (null u) -> Just $ filter (/= '@') u
+                               _ -> Nothing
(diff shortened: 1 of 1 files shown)


gbaz:
So again, I'd like some tests, but would appreciate help on that front. Also, before merge, a few of us should give it a workout and sanity check it by hand -- I think we really dropped the ball in reviewing this properly to begin with, sigh.

gbaz:
I can verify by the way that I successfully both ran `cabal update` (for get) and `cabal upload` (for putting a file) with this patch and they worked

andreasabel:
This PR fixes my OP of #7666; `cabal upload` works again for me.

dtaskoff:
I'm sorry, none of the issue mentioned above seem familiar to me. I think you might've mistaken me for someone else, possibly during GitHub's username auto-completion?

mouse07410:
Overall, looks reasonable. But my experience programming in Haskell is weak. 

It would be nice if the function on question was documented. I wonder if it's feasible to pull the function on question fun the file is defined in, and place into a standalone "reproducer", and then see if it works as expected.

PR #7675, opened 2021-09-21
Revert "Change cabal-plan list-bin to cabal list-bin in Makefile"

andreasabel:
Reverts haskell/cabal#7648

The replacement of `cabal-plan list-bin` by `cabal list-bin` did not work for macOS.
Details see https://github.com/haskell/cabal/pull/7648#issuecomment-924111500.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ $(TEMPLATE_PATHS) : templates/Paths_pkg.template.hs cabal-dev-scripts/src/GenPat
 
 buildinfo-fields-reference : phony
 	cabal build --builddir=dist-newstyle-bi --project-file=cabal.project.buildinfo buildinfo-reference-generator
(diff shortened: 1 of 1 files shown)


jneira:
@andreasabel hi, it would be great to have a dedicated bug report about as the idea would be to fix it imo. Reverting the change could be considered if the fix cant be included in the next 3.6 release. And not sure if it should be reverted in master if we want to work on it. Sorry if a i am missing some discussion or additional infor about

andreasabel:
@jneira : The merged PR only affects developers (only macOS?) so I propose to revert it now and get us back into a good state where we can start investigating the underlying problem.

A change to a developer-only goal in the `Makefile` isn't something that is relevant for the release, I'd say.

jneira:
> A change to a developer-only goal in the Makefile isn't something that is relevant for the release, I'd say.

Agree, did not consider the scope of the change well

jneira (approved):

fgaz (approved):
See https://github.com/haskell/cabal/pull/7648#issuecomment-925300582

I also opened #7679

Issue #5773, opened 2018-12-04
cabal new-build --package-db=foo doesn't do anything


phadej:
if `foo` doesn't exist: builds
if `foo` exists: doesn't consider it.

I think that for `3.0` we should either remove package database flags from `new-build` commands; or teach these commands to use other package databases. I.e. to consider other, not only global database. I actually have a use case for this. ("obviously", if I use some weird `--package-db`, I'd use my own `--store-dir` etc. so this is intended for people who know what they are doing).

PR #7676, opened 2021-09-22
Add package-db flag for v2 commands

patrickdoc:
---

---

This PR teaches v2-* commands how to use the `--package-db` flag (#5773). The records were already present in the config types, just commented out. I've un-commented them and injected the values into the two places they were needed (initial plan and elaboration).

TODO:
* [x] Modify package hashes based on PackageDBStack
* [x] Maintain backwards-compat for the default stack hash
* [x] Handle flag passed in cabal.project file
* [x] Document how the flag works

The added tests cover:
- adding an additional package db to your build (`--package-db=xyz`)
- clearing all package dbs (`--package-db=clear`)
- clearing the default package dbs and using all manual (`--package-db=clear --package-db=global --package-db=xyz`)


jneira:
@patrickdoc  thanks for the pr, i will ping @phadej as he opened the original issue

fgaz:
Oooh this will be useful for debugging #7270

jneira:
Out of curiosity how this interacts with the V2 store package-db? is it cleared with `clear`? sorry if it is something obvius

gbaz:
My understanding/expectation is that it interacts as described above -- i.e., cabal will provide the packages in the store following the package-db stack, just as it does with the default stack? 

Let's think through the hash case closely. Suppose I have `a` in some stack, and `b` depends on `a` and is put in the store. Now I compile against a stack _without_ `a`. I think that what will happen is cabal's solver will realize both `b` and `a` are required, and in turn it will not be able to find `a` and so will rebuild it and place it in the store. The question is then if it will attempt to use the `b` linked against the "old" `a` or if it will instantiate a new one. Further, if it does the former, will that even be an issue?

I think its worth testing first to be sure that this path doesn't somehow work, but I also understand very much the concern here.

I'd also like to figure out where the documentation goes? The flag is already listed in the "Setup.hs flags" section of the docs, which is where most similar flags that cabal-install accepts but are passed to the underlying Setup are listed.

gbaz:
Right -- I was just trying to both fully express what the precise bad outcome could be and also expressing a question as to whether that outcome would occur or not -- i.e. would the hash _actually_ need to be changed, or is it possible that things would work without it.

Note that changing the hash can be problematic (without careful attention to backwards compat) because a simple cabal-install version bump could wind up invalidating an entire store.

gbaz:
@patrickdoc I think the "cabal.project" section of the documentation could have the flag (as that also has v2 opts) as well, rather than just the Setup.hs section.

Another thought could be that we could enable this flag _just_ with build and _not_ with install, which would bypass the issue of the store hashes in a very straightforward way, and might still suffice for all the intended use cases?

jneira:
thanks for the info, it is very useful indeed, I still miss the place of the store package db in that picture

patrickdoc:
Requested changes are complete:
- package db stack added to the hash
- flag documentation added to the guide

Kleidukos (approved):
LGTM.

doc/cabal-project.rst
```diff
@@ -269,6 +269,36 @@ package, and thus apply globally:
...
+    clear
+        Remove all current entries from the stack
+    global
+        Compiler-specific packageDB containing rts, base, etc.
+    user
+        User-specific packageDB. Generally not applicable with v2-* commands.
+    PATH
+        Absolute path to a packageDB.
+    
+    With Nix-style local builds, the default package DB stack is ``[global,
```
jneira, line 292: So if you uses `--package-db`, could not you specify the `<compiler store>` as one of the pieces of the stack? is it included always by default? or you have to set it explicitly with the path?

jneira: i would change `<compiler store>` with `<cabal store>` as that package db is created and used by cabal. to distinguish it from the global, which is tied to the compiler

jneira: > Before the change, the PackageDB stack was (relevant code in ProjectPlanning elaboration):
> [CompilerDB, InplaceDB]

> After the change, the PackageDB stack is:
> [CompilerDB, <user flag dbs>, InplaceDB]

If this is still up to date could be used as a model to add the note i miss

Mikolaj:
@phadej: I'm guessing your comments have been addressed sufficiently and so I will take the liberty of dismissing the review blockage, as per usual, unless you object. Thank you once again for your insight!

doc/cabal-project.rst
patrickdoc: So I tried to keep this relatively concise, but using my computer (debian) as an example:

`global`: `~/.ghcup/ghc/8.10.7/lib/ghc-8.10.7/package.conf.d`
`user`: not defined
store: `~/.cabal/store/ghc-8.10.7/package.db`
in-place: `./dist-newstyle/packagedb/ghc-8.10.7`

By default, the package stack you will have with v2 commands is:
`[global, store]`
So all remote packages required by your project will be registered in the store package db (because it is last).

When cabal starts building your local projects, it appends the in-place db to the end:
`[global, store, in-place]`
So your local packages get put in dist-newstyle instead of the store.

This flag manipulates the default prefix:
`[global, store]`

To more directly answer your questions:
- yes, you can specify the store package db with: `--package-db /home/patrick/.cabal/store/ghc-8.10.7/package.db`
- yes, it is included by default

The main tricky bit is the `clear` option, which removes all previous entries. Ex.
`--p-db foo` -> `[global, store, foo]`
`--p-db clear --p-db foo` -> `[foo]`
`--p-db clear --p-db foo --p-db clear --p-db bar --p-db baz` -> `[bar, baz]`

jneira: wow, that is a great explanation, imo it deserves be included in docs including the super useful final examples, would you mind do it in this pr as you habe it almost written?

patrickdoc: Sure. Just under the flag description or a more general place?

jneira: in the flag description sounds good to me, it is a somewhat advanced stuff and useful for someone who already wants to tweak seminternal things

Cabal/src/Distribution/Simple/Setup.hs
```diff
@@ -38,7 +38,7 @@ module Distribution.Simple.Setup (
   GlobalFlags(..),   emptyGlobalFlags,   defaultGlobalFlags,   globalCommand,
   ConfigFlags(..),   emptyConfigFlags,   defaultConfigFlags,   configureCommand,
   configPrograms,
-  configAbsolutePaths, readPackageDbList, showPackageDbList,
+  configAbsolutePaths, readPackageDb, readPackageDbList, showPackageDb, showPackageDbList,
```
fgaz, line 41: Please add @since annotations

doc/cabal-project.rst
patrickdoc: updated

Cabal/src/Distribution/Simple/Setup.hs
patrickdoc: updated

jneira (approved):
Many thanks for the pr and for addressing my suggestions, looks really great

Mikolaj:
@fgaz: have your comments been addressed?

Dear all, should we merge?

fgaz:
> @fgaz: have your comments been addressed?

Yes, let's merge!

PR #7681, opened 2021-09-23
Option to place comments after the cabal file field declaration

### LINKED ISSUE #7680: cabal init does not put cabal-version on the first line
**Describe the bug**

cabal init does not put cabal-version on the first line. it puts a comment instead. according to the spec, though, starting from 2.2, the version should be on the first line

**To Reproduce**

```
$ cabal init --non-interactive --lib

Guessing dependencies...

Using cabal specification: 3.0

Creating CHANGELOG.md...
Creating file CHANGELOG.md...
Creating directory ./src...
Creating file src/MyLib.hs...
Creating file tmp-Ztm69mNz3W.cabal...

You may want to edit the .cabal file and add a Description field

$ cabal build
Errors encountered when parsing cabal file ./tmp-Ztm69mNz3W.cabal:

tmp-Ztm69mNz3W.cabal:5:1: error:
cabal-version should be at the beginning of the file starting with spec version 2.2. See https://github.com/haskell/cabal/issues/4899

    5 | cabal-version:      3.0
      | ^


```

The generated cabal file starts with:

```
-- The cabal-version field refers to the version of the .cabal specification,
-- and can be different from the cabal-install (the tool) version and the
-- Cabal (the library) version you are using. As such, the Cabal (the library)
-- version used must be equal or greater than the version stated in this field.
cabal-version:      3.0

```

**Expected behavior**

cabal-version and the comment should be swapped

**System information**

 - cabal: current master

/cc @emilypi @ptkato

ptkato:
This PR will add the ability to choose whether a comment that accompanies a cabal file field goes before of the field declaration or after it. Fixes #7680, which was introduced by #7553.

---


ptkato:
cc @emilypi @fgaz

cabal-install/src/Distribution/Client/Init/Format.hs
```diff
@@ -87,21 +91,23 @@ fieldD fieldName fieldContents fieldComments includeField opts
...
 
 
 -- | A field annotation instructing the pretty printer to comment out the field
 --   and any contents, with no comments.
-commentedOutWithComments :: [String] -> FieldAnnotation
-commentedOutWithComments = FieldAnnotation True . map ("-- " ++)
+commentedOutWithComments :: CommentPosition -> FieldAnnotation
+commentedOutWithComments (CommentBefore cs) = FieldAnnotation True . CommentBefore $ map ("-- " ++) cs
+commentedOutWithComments (CommentAfter  cs) = FieldAnnotation True . CommentAfter  $ map ("-- " ++) cs
+commentedOutWithComments NoComment = FieldAnnotation True NoComment
```
ptkato, line 102: I'm not very confident on this solution, it looks too ad-hoc-y. Maybe I could get rid of the `NoComment` value altogether?

emilypi: i'd actually argue this is a fairly clean solution

cabal-install/src/Distribution/Client/Init/Format.hs
```diff
@@ -244,6 +250,8 @@ mkPkgDescription opts pkgDesc =
       , "and can be different from the cabal-install (the tool) version and the"
       , "Cabal (the library) version you are using. As such, the Cabal (the library)"
       , "version used must be equal or greater than the version stated in this field."
+      , "Starting from the specification version 2.2, the cabal-version field must be"
+      , "the first thing in the cabal file."
       ]
```
ptkato, line 255: I took the liberty to add this extra comment here, stating that the `cabal-version` field should be the first line in the file.

ptkato:
Hmm... It seems that those changes broke a considerable amount of parser tests from Cabal-tests.

emilypi (approved):

Cabal/src/Distribution/Fields/Pretty.hs
```diff
@@ -115,10 +118,18 @@ flattenBlocks = go0 where
 
 renderField :: Opts ann -> Int -> PrettyField ann -> Block
 renderField (Opts rann indent post) fw (PrettyField ann name doc) =
-    Block before after $ comments ++ post ann lines'
+    Block before after content
   where
+    content = case comments of
+      CommentBefore cs -> cs ++ post ann lines'
+      CommentAfter  cs -> post ann lines' ++ cs
+      _ -> post ann lines'
```
fgaz, line 126: Since it's the only remaining case, could you match on `NoComment` explicitly?

Cabal/src/Distribution/Fields/Pretty.hs
```diff
@@ -133,11 +144,16 @@ renderField (Opts rann indent post) fw (PrettyField ann name doc) =
...
+    
+    attachComments
+      (post ann [ PP.render $ PP.hsep $ PP.text (fromUTF8BS name) : args ])
     ++
     map indent (renderFields opts fields)
+  where
+    attachComments content = case rann ann of
+      CommentBefore cs -> cs ++ content
+      CommentAfter  cs -> content ++ cs
+      _ -> content
```
fgaz, line 156: Same here

cabal-install/src/Distribution/Client/Init/Format.hs
```diff
@@ -76,6 +76,10 @@ fieldD fieldName fieldContents fieldComments includeField opts
         -- If the "--minimal" flag is set, strip comments.
         in fieldSWithContents comments
   where
+    isCabalVersion fn
```
fgaz, line 79: I stared at the call site for a while thinking it was a `_ -> Bool` function. What about `commentPositionFor`?

Cabal/src/Distribution/Fields/Pretty.hs
```diff
@@ -33,6 +34,8 @@ import qualified Distribution.Fields.Parser as P
 import qualified Data.ByteString  as BS
 import qualified Text.PrettyPrint as PP
 
+data CommentPosition = CommentBefore [String] | CommentAfter [String] | NoComment
```
fgaz, line 37: Could you briefly explain why this type exists in a comment?

fgaz (commented):
Some really minor stuff.

Also it'd be nice to have a test that checks that init produces a package that can actually be built. I think it's possible to do it in cabal-testsuite (see the hermetic tests section)

ptkato:
> Also it'd be nice to have a test that checks that init produces a package that can actually be built. I think it's possible to do it in cabal-testsuite (see the hermetic tests section)

I'll look into it.

emilypi:
@fgaz I'd call that a separate PR, just so we don't derail this one.

fgaz:
Thanks @ptkato 

@emilypi ok, as long as we don't forget!

emilypi:
@fgaz agreed! @ptkato do you mind taking that with an issue?

fgaz:
#7689

fgaz (approved):

PR #7683, opened 2021-09-23
Fix #7569 for cabal-install: bump to 3.7

### LINKED ISSUE #7569: Bump Cabal version to 3.7
Development version should not be equal to release version.

This is blocking development on #7489, as the flag `--enable-build-info` is guarded s.t. that it can not be accidentally used with older Cabal versions.

andreasabel:
Fix #7569 for cabal-install: bump to 3.7

Also: move `cabalInstallVersion` to its own module `Distribution.Client.Version` as suggested by @emilypi at https://github.com/haskell/cabal/issues/7569#issuecomment-925942914.


cabal-install/src/Distribution/Client/Version.hs
```diff
@@ -0,0 +1,12 @@
+-- | Provides the version number of @cabal-install@.
+
+module Distribution.Client.Version
+  ( cabalInstallVersion
+  ) where
+
+import Distribution.Version
+
+-- TODO: write a test around this. Don't abuse Paths_cabal_install.
+--
```
emilypi, line 10: Comment is stale. Suggesting this as an alternative text: 

```suggestion
-- This value determines the `cabal-install --version` output.
-- 
-- It is used in several places throughout the project, including anonymous build reports, client configuration, 
-- and project planning output. Historically, this was a @Paths_*@ module, however, this conflicted with 
-- program coverage information generated by HPC, and hence was moved to be a standalone value.  
--
```

emilypi (approved):
Approved pending CI

Mikolaj (approved):
LGTM. Thank you so much.

PR #7686, opened 2021-09-26
addFileMode: Replace use of Posix.Internals module with public API

DanielG:
Another follow-up to #7572 

Turns out there's a less gnarly API for file modes than the .Internals stuff we were already using.

---
* [ ]

### DIFF
--- a/Cabal/src/Distribution/Compat/CopyFile.hs
+++ b/Cabal/src/Distribution/Compat/CopyFile.hs
@@ -36,12 +36,8 @@ import Foreign
 
 import System.Posix.Types
          ( FileMode )
-import System.Posix.Internals
-         ( withFilePath
-         , c_chmod
-         , c_stat, sizeof_stat, st_mode )
-import Foreign.C
-         ( throwErrnoPathIfMinus1_ )
+import System.Posix.Files
+         ( getFileStatus, fileMode, setFileMode )
 
 #else /* else mingw32_HOST_OS */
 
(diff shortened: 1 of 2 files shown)


Mikolaj:
@DanielG: what's the status of this one? May I help move this forward?

Mikolaj:
@DanielG: this looks like a great PR, simplifying things and probably improving or future-proofing as well. Is this finished?

Mikolaj:
@Mergifyio rebase

DanielG:
Oh yeah, absolutely. This should be good to go.

Mikolaj:
Amazing! Any volunteer to add a minimal changelog file? Another volunteer for a minimal test perhaps?

DanielG:
Do we really need a changelog for a purely internal change like this? Seems excessive to me. We're just replacing one wrapper around chmod with another.

Mikolaj:
So there is no change of semantics? If so, you are completely right.

DanielG:
Yeah it's completely equivalent. The only concern would have been whether it breaks the build with the Data.Bits change but since CI is green that's fine too :)

Mikolaj:
So you also suggest it's significantly tested by tests we already have?

Mikolaj (approved):
Amazing what ancient relics we still have in the codebase.

bacchanalia:
It's unclear to me why `cabal-install/src/Distribution/Client/Compat/FilePerms.hs` exists. All it does is duplicate code for some functions already exported by `Cabal/src/Distribution/Compat/CopyFile.hs`, and it is imported in exactly zero places.

Mikolaj:
Well spotted! And `Cabal/src/Distribution/Compat/CopyFile.hs` has the ancient version of the code.

Mikolaj:
Oh, of course, it has the ancient version --- it's the second of the files update in this PR.

From what you say it's safe to remove `Cabal/src/Distribution/Compat/CopyFile.hs`, though I'd first compare the code to make sure we are not removing a more up to date implementation of the same functionality.

Mikolaj:
It might have been included due some backward compatibility (given the `Distribution.Client.Compat` path), but the file history suggests nobody cares any more and also using `cabal-install` as a library (for such basic things) is discouraged and also `Cabal` can be used instead.

DanielG:
Looks to me like 6248c74280a54382f85079f58a1ea40ba0db9476 removed the last use-site for the Compat.FilePerms module. Weird I distinctly remember being annoyed by the code dupe too. I would have for sure tried to remove it but I guess maybe I was just too lazy :)

DanielG:
> It's unclear to me why `cabal-install/src/Distribution/Client/Compat/FilePerms.hs` exists. All it does is duplicate code for some functions already exported by `Cabal/src/Distribution/Compat/CopyFile.hs`, and it is imported in exactly zero places.

Keep in mind that cabal-install cannot depend on "new" stuff from lib:Cabal because it should build with older versions of Cabal too. That's why the Compat module was likely introduced  back in the day FYI.

Mikolaj:
@bacchanalia: would you like to review and, if you so desire, follow up by a PR removing the spurious file?

bacchanalia (approved):
LGTM

Issue #7394, opened 2021-05-12
shell hooks for augmenting how cabal does various actions?


cartazio:
maybe theres a ticket about this, but i was talking with @hasufell  about cabal and he mentioned that in old discussions with HVR, there was the idea of making it easier to glue cabal into a overall toolchain if there were the sort of shell hooks a la git-hooks or https://paludis.exherbo.org/configuration/hooks.html 

with example application for example being:
``
and then you could hook into phases of like pre_configure and invoke ghcup automatically, installing the requested project ghc version
``

at the moment, i think cabal's pre/post-`$FOO` (for some value of FOO) hooks are only supported as haskell function call backs, which hits a different point in the ease of scripting tradeoffs

cartazio:
there may already be a tracking ticket related to this, but i'm not super familiar with older tickets ;)

cartazio:
loosely related is https://github.com/haskell/cabal/issues/3600

danidiaz:
One possible way of supporting hooks is by extending the experimental branch described [here](https://github.com/haskell/cabal/issues/7420#issuecomment-860034936).

In that branch, functions like [runProjectPreBuildPhase](https://github.com/danidiaz/cabal/blob/6cb294c33a7624167a0e6dde9bcc65b8cbc1137e/cabal-install/src/Distribution/Client/ProjectOrchestration.hs#L328) and [freezeAction](https://github.com/danidiaz/cabal/blob/6cb294c33a7624167a0e6dde9bcc65b8cbc1137e/cabal-install/src/Distribution/Client/CmdFreeze.hs#L101) have been wrapped in newtypes.

Potential "hooks" that added behavior to those functions (shell invocations, logging, and so on) could be defined as functions `RunProjectPreBuildPhase -> RunProjectPreBuildPhase` or `FreezeAction -> FreezeAction`, respectively. And they could be applied to the original functions [here](https://github.com/danidiaz/cabal/blob/6cb294c33a7624167a0e6dde9bcc65b8cbc1137e/cabal-install/src/Distribution/Client/CompositionContext.hs#L197), in the so-called "composition context". If the transformations needed to allocate some resources of their own, it could also be done there.

fgaz:
This is for _`cabal-install`_ hooks, right?

cartazio:
I believe yes :)

danidiaz:
[...]

Issue #6726, opened 2020-04-21
Supporting BNFC


ivanperez-keera:
I'd like to propose that Cabal support BNFC as a build tool. BNFC is a well-known, multi-language parser generator that supports Haskell via Happy & Alex.

**Additional context**
From what I understand, the required changes would be:

- Adding bnfc to this list here: https://github.com/haskell/cabal/blob/0c9792076b30257ff638bf6dd5a13b77ec7b22be/Cabal/Distribution/Simple/BuildToolDepends.hs#L31-L45
- Adding "fc" to "bnfc" here: https://github.com/haskell/cabal/blob/72d67bfe965d0ed700d6879a97cddf7a15ab8ff7/cabal-install/Distribution/Client/Init/Heuristics.hs#L215-L225
- Adding bnfcProgram to this export list: https://github.com/haskell/cabal/blob/9537238f5bd070aac1c998217c3c5e0136e71b4d/Cabal/Distribution/Simple/Program/Builtin.hs#L13 and this list: https://github.com/haskell/cabal/blob/9537238f5bd070aac1c998217c3c5e0136e71b4d/Cabal/Distribution/Simple/Program/Builtin.hs#L69-L70, and introducing a bnfcProgram definition in that module. For bnfc, it appears that just --version is enough to detect the version number.

Questions:
- Would such a patch be accepted?
- If so, would those versions be enough?
- How to handle the need to arguments to bnfc (e.g., `--haskell -p <PATH> -o <directory> <file.cf>`? Would doing this immediately make cabal understand `--with-bnfc`, `--with-bnfc-options` and `bnfc-options:`?

phadej:
- I'm ok about adding `bnfc` to known suffix preprocessors, but
- I think we have strong assumption that `packagename` and `exename` is the same, which isn't: `bnfc` and `BNFC` which might require some bigger changes.
- `desugarBuildTool` is for legacy `build-tools`, and I think shouldn't be changed
- I suspect things need changes if there is no one-to-one mapping between source `MyLanguage.fc` and generated `MyLanguage.hs`. I guess `bnfc` generates some `.x` and `.y` (and `.hs`) files. I suspect the multiple output file would case architectural problems.

Adding `bnfc-options:` to `.cabal` file syntax is *a big deal*, so everything else have to be figured out until we can commit to that change, we'll need to live with if **forever**. Note, we don't have `alex-options` nor `happy-options` as these tools self-configure from the input file. I'd appreciate that kind of working mode.

[...]

PR #7688, opened 2021-09-27
code-generators field in test stanza.

gbaz:
* [x] Add tests.

This PR extends the `test` stanza of a cabal file with a `code-generators:` field containing executables. These executables can be generated as dependencies via the `build-tools-depends:` field. They are run during the build step after all other preprocessors, and before the main build step, and can be used to autogenerate modules for test suites, e.g. for doctests, or other "discover" style test running.

These executables are expected to follow a very simple api. They are invoked as `exe-name TARGETDIR [SOURCEDIRS] -- [GHCOPTIONS]` -- where targetdir is a temp directory in the `dist-newstyle` hierarchy where they are directed to place their generated code. This is followed by a list of source directories containing the source files of all the local lib components which the test stanza depends on. Finally, following a double-dash, are all the options that cabal would pass to ghc for a build.

The expected output of these executables is a newline-separated list of names of all the generated modules. This PR also prevents checking for the existence of the main-is module in the preprocessor step, because that main module itself may be generated by the code generator.

A proof-of-concept of how this works is at https://github.com/gbaz/cabal-doctest-demo -- this repo gives a package that has _two different_ doctest runners built on top of this functionality. One wraps up the `doctest` program with a wrapper generated from the ghc options (and so which does not rely on a ghc environment file to bring deps into scope, but rather extracts them directly form the build-depends). The other uses the lexer from https://github.com/phadej/cabal-extras/tree/master/cabal-docspec and generates a _fully static_ tasty testsuite from the doctest comments in the source lib.

I think this code is basically correct and clean. If people like it, to finish off this PR it needs a changelog, doc updates, and some proper extension of the cabal test suite.


Mikolaj:
This is amazing.

fgaz:
I wonder, could this be generalized to make the existing preprocessor feature extensible?

jneira:
> 
> 
> I wonder, could this be generalized to make the existing preprocessor feature extensible?

https://github.com/haskell/cabal/issues/7394??

fgaz:
> > I wonder, could this be generalized to make the existing preprocessor feature extensible?
> 
> #7394??

that's in cabal-install, while I'm talking about the Cabal/.cabal feature: https://cabal.readthedocs.io/en/latest/cabal-package.html#modules-and-preprocessors

#6726 mentions extensible preprocessors

Kleidukos:
@gbaz Hi! I've taken a look at the PR and I was wondering: I see you've used the syntax `docspec-static-demo:docspec-static`, but it's because you've looked-up the program database right? If I were to put my build-tool on Hackage, cabal would fetch it automatically right?

gbaz:
That syntax is existing syntax for how `build-tool-depends` works -- the first component is the package (pulled from hackage, or whatever other repos are specified, etc) and the second is the component that's built and depended on -- i.e. the particular tool from the package to be brought into the program-database.

The new thing is the code-generators field, which does not use that component syntax, but just looks up the tool by name in the program database.

Kleidukos:
@gbaz Thank you very much! Do you still need a hand to finalise this PR? I would also enable the syntax in other places like library components

gbaz:
As I discussed on irc I don't think it makes sense in library components because this enables things to generate new modules fed into the build pipeline. For libraries, we want all exposed modules declaratively specified in the file itself.

I'm finishing up the build issues on older ghcs, then will add some tests and docs, but otherwise yeah I'd like some review on it.

Kleidukos:
@gbaz Concretely what happens if a module isn't declared in `autogen-modules`? There is an immediate sanction for the developer right? Is there a difference of feedback / behaviour compared to what people can achieve today with `Setup.hs`? 

From what @swamp-agr showed me in [encoding](https://github.com/dmwit/encoding/blob/master/encoding.cabal), the process seems to be enforced just like you want for custom setups?

gbaz:
I should clarify. The point of this field is that it can generate modules that are not listed in the cabal file, and make use of them.

The use case of autogen-modules is totally different. It allows modules that _are_ listed in the cabal file to not cause failures with sdist when their source does not appear: https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-autogen-modules

gbaz:
What is the point of https://github.com/haskell/cabal/blob/master/Cabal-tests/tests/UnitTests/Distribution/Utils/Structured.hs btw?

Is it just to make us notice when we change the structure hash? I.e. this pr adds stuff to the TestSuite type, which necessarily affects the hash of GenericPackageDescription and LocalBuildInfo. But its ok to do that -- we've updated this file in the past to reflect changes to the hash. So what are we gaining by this?

Kleidukos:
> I should clarify. The point of this field is that it can generate modules that are not listed in the cabal file, and make use of them.

Ah I see, those two usecases are different. Thank you very much for the explanation. I imagine this justifies the separate `build-drivers` that I had imagined at the beginning. Then if the user generates modules that are depended on but not listed, Cabal will properly complain. :)

> What is the point of https://github.com/haskell/cabal/blob/master/Cabal-tests/tests/UnitTests/Distribution/Utils/Structured.hs btw?

I think @phadej originally introduced it in ae33fcf938949f88948e0611623a4b989a43a223.

phadej:
`Structured` helps with cache expiration. `binary` can do weird stuff when decoding something. (E.g. if it expects list length and gets some big `Int` indeed...). The test for GPD hash is there to notice that structure hash changes when it should and don't when it shouldn't.

gbaz:
Right. I understand and appreciate structured. Not a big deal but just remain unsure if there's ever a case where these tests will fail and it will mean some action other than "update the tests with the new hash".

gbaz:
added changelog, docs, tests. ready for review and merge

jneira (approved):
The pr loos great, it has docs, tests and no one has expressed any opposition.

After a first quick look I also thought it asks for a generalization but @gbaz arguments for the actual scope are convincing. If later users needs it in other components (benchmarks?) we could generalize it later.

Another caveat is the usual one cabal has already lot of fields some of them with very specific use cases, undoubtly important but not a big deal.

Kleidukos:
From my perspective it's all good. :)

Kleidukos (approved):

Cabal-syntax/src/Distribution/PackageDescription/FieldGrammar.hs
```diff
@@ -268,6 +268,7 @@ data TestSuiteStanza = TestSuiteStanza
     , _testStanzaMainIs     :: Maybe FilePath
     , _testStanzaTestModule :: Maybe ModuleName
     , _testStanzaBuildInfo  :: BuildInfo
+    , _testStanzaCodeGenerators :: [String]
```
andreasabel, line 271: `String` isn't a good name for a type as it does not reflect its intended semantics.  It is also badly searchable across the code base.
```haskell
type CodeGenerator = String
type CodeGenerators = [CodeGenerator]
```
and then use these type synonyms.  
Extra points for a `newtype`!

Cabal-syntax/src/Distribution/Types/TestSuite.hs
```diff
@@ -26,7 +26,8 @@ import qualified Distribution.Types.BuildInfo.Lens as L
 data TestSuite = TestSuite {
         testName      :: UnqualComponentName,
         testInterface :: TestSuiteInterface,
-        testBuildInfo :: BuildInfo
+        testBuildInfo :: BuildInfo,
+        testCodeGenerators :: [String]
```
andreasabel, line 30: A comment would be in order explaining this new field, maybe linking to this PR.

In particular I wonder, should this field be non-revisable on Hackage?
(By default, I suppose, "yes, cannot change in a revision".  But on the other hand, what is the harm of revising this?  It's just for the test suite.  Also, version of dependencies can be arbitrary revised, to the point of completely changing functionality... Why not test generaters...?)

Cabal-syntax/src/Distribution/PackageDescription/FieldGrammar.hs
```diff
@@ -315,23 +321,20 @@ testSuiteFieldGrammar = TestSuiteStanza
     <*> optionalFieldAla "main-is"     FilePathNT testStanzaMainIs
     <*> optionalField    "test-module"            testStanzaTestModule
     <*> blurFieldGrammar testStanzaBuildInfo buildInfoFieldGrammar
+    <*> monoidalFieldAla "code-generators"        (alaList'  CommaFSep Token)     testStanzaCodeGenerators
+          ^^^ availableSince CabalSpecV3_6 [] -- TODO 3_8
```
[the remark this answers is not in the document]
andreasabel: Fixed in 
- #8134

Mikolaj:
@gbaz: what do you plan to do with the extra comments donated to you after the merge?

gbaz:
They're useful comments, but I think too small to be actionable after the fact.

andreasabel:
> They're useful comments, but I think too small to be actionable after the fact.

@gbaz: Not sure I follow this logic.

But I'd anyhow appreciate some discussion/justification about revisability decisions...

Cabal-syntax/src/Distribution/Types/TestSuite.hs
gbaz: our principle for allowing revisions has always been "by default, negative, and allow revision only if there's demonstrated need"

gbaz:
I left a comment regarding revisability. Otherwise, I just meant, there's plenty else on my plate and I don't imagine I'll find time for stylistic/comment cleanup at this point. Maybe the next time somebody touches the relevant code they could do further work, but for now I'm going to focus elsewhere.

Cabal-syntax/src/Distribution/Types/TestSuite.hs
andreasabel: Thanks for the justification.

andreasabel:
> I left a comment regarding revisability.

Thanks!

sol:
I'm not up to date with all the details of this proposal and, as I'm swamped with other obligations, I'm not planning to get too involved.

## `hspec-discover` does not require this (or any other changes to Cabal)

The current approach is adequate, robust and relies on GHC features alone (read: you can easily use it with GHC, without relying on Cabal).

## `doctest` doesn't need this

By design, `doctest` should  behave as similar to `ghci` as possible. It also supports the same cli. Hence it makes sense to invoke it the same way as you would invoke `ghci`. The closest I've come up with is to invoke it with:

    cabal repl --with-ghc=doctest

This woks well in most situations and is my current recommendation. The drawback of this approach is that `doctest` will also be used as a compiler, not only as a `ghci` substitute. The `doctest` executable detects when it's asked to compile something and delegates these requests to the `ghc` on the path. Again, this works well, except for when compiling the `ghc-paths` package. That's why you want to make sure that all your dependencies are up to date first. Just do a regular build, before invoking `doctest`:

    cabal build && cabal repl --with-ghc=doctest

The full-fledged version, with all the bells and whistles, of this command is:

    cabal install doctest --overwrite-policy=always && cabal build && cabal repl --build-depends=QuickCheck --build-depends=template-haskell --with-ghc=doctest

## What would indeed help from a Doctest's perspective

If we had a flag `--with-ghci` or `--with-repl` which tells Cabal to use a different executable as the REPL command only, that would be neat! This will simplify my currently recommended approach and will make it more robust.

In addition, this caters for all kinds of interesting use cases, like say you could use a drop-in "improved" version of `ghci`.

## What I hope will not happen

1. The `.cabal` file format spec is not compact at all. It's big.  Adding more "useful features" will only make it bigger. That's why I personally think it makes sense to be conservative here.

2. I'm also using `--with-ghc` a lot for similar things, like speeding up test runs with `hspec/sensei`. Please don't break it**, or if you do so then add a `--with-ghci` option first.

** read: Don't key the "path to ghc", into the hashes for the nix-style package store!

Mikolaj:
Thank you for your remarks. They make a lot of sense.

> I'm not up to date with all the details of this proposal

Could you elaborate which proposal you have in mind, e.g., which ticket? This PR is an implementation, not a proposal, and it's finalized and merged and used in (demos of) external tools.

gbaz:
@sol what this feature (now merged) allows is writing small drivers so that one need not install `doctest` or any other executable ahead of time, nor does that executable or cabal need to be invoked with any special incantation.

Instead, with a single addition of a field and a supporting driver, one can simply run `cabal test`.

An example of such a cabal field and stanza (that works) is at: https://github.com/gbaz/cabal-doctest-demo/blob/main/cabal-doctest-demo.cabal#L36

The supporting driver is at: https://github.com/gbaz/cabal-doctest-demo/blob/main/doctest-codegen-driver/app/Main.hs

PR #7670, opened 2021-09-20
Remove debug info from end of error message (#7248)

ivanperez-keera:
The function that informs users that the package list is missing prints
an internal data structure at the end of the message, without formatting
it.

This extra info is likely just left over from debugging and does not
need to be printed.

This commit just removes that unnecessary show.

---
Please include the following checklist in your PR:

* [X] Patches conform to the [coding conventions](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#conventions).
* [ ] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/haskell/cabal/blob/master/CONTRIBUTING.md#changelog).
* [X] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

Issue #7248, opened 2021-01-15
Error message about missing package list shows internal data structure RemoteRepo


andreasabel:
On a fresh install of ghc+cabal, a call to e.g. `cabal unpack` prints an error message that contains a `show`ed data structure.
```
$ cabal unpack BNFC
Config file path source is default config file.
Config file C:\Users\abel\AppData\Roaming\cabal\config not found.
Writing default configuration to C:\Users\abel\AppData\Roaming\cabal\config
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.RemoteRepo {remoteRepoName = "hackage.haskell.org",
remoteRepoURI = http://hackage.haskell.org/, remoteRepoSecure = Just True,
remoteRepoRootKeys =
["fe331502606802feac15e514d9b9ea83fee8b6ffef71335479a2e68d84adc6b0","1ea9ba32c52
6d1cc91ab5e5bd364ec5e9e8cb67179a471872f6e26f0ae773d42","2c6c3627bd6c982990239487
f1abd02e08a02e6cf16edb105a8012d444d870c3","0a5c7ea47cd1b15f01f5f51a33adda7e655bc
0f0b0615baa8e271f4c3351e21d","51f0161b906011b52c6613376b1ae937670da69322113a246a
09f807c62f6921"],
remoteRepoKeyThreshold = 3, remoteRepoShouldTryHttps = True}
cabal.exe: There is no package named 'BNFC'.
You may need to run 'cabal update' to get the latest list of available
packages.

$ cabal --version
cabal-install version 3.2.0.0
compiled using version 3.2.0.0 of the Cabal library
```

ivanperez-keera:
@phadej I just hit this one.

In the interest of being constructive :) maybe I can send a PR?

I think this is the line producing that:

https://github.com/haskell/cabal/blob/ac5cb4c308079ce2a8d3eb6563e192aec41c58bf/cabal-install/src/Distribution/Client/IndexUtils.hs#L408

What would be a good solution? Adding a new line and printing the repo info in a nicer format? Or just not showing it at all?

Mikolaj:
I'd just show nothing at all. This looks like a debug print that made it into a release. If we had a DEBUG_PRINT flag, we could CPP it, but AFAIK, we don't.

ivanperez-keera:
Ok I'll send that small change then. Thanks for the quick reply @Mikolaj .

Mikolaj:
[...]

PR #7702, opened 2021-10-06
Re #7670: WIP: test case for #7248

andreasabel:
Two problems remaining:
- Warning output is not recorded in the golden value.
- The correct warning is not produced yet.

Current warning:

    Warning: No remote package servers have been specified. Usually you would have one specified in the config file.

---


Mikolaj:
I'm guessing from a quick code-dive, but how about `cabalTest . recordMode RecordAll $`? Also, perhaps parameters `["--config-file", conf]` and `repository` specified in the `conf` file would suffice to warn about the right thing?

andreasabel:
@Mikolaj wrote:
> how about `cabalTest . recordMode RecordAll $`?

This does record the warnings, but also the call stack, which is unwanted because it is too brittle as a golden value:
```
Warning: No remote package servers have been specified. Usually you would have one specified in the config file.
  withMetadata, called at src/Distribution/Simple/Utils.hs:379:14 in Cabal-3.7.0.0-inplace:Distribution.Simple.Utils
-----BEGIN CABAL OUTPUT-----
Error: cabal: There is no package named 'BNFC'. 
You may need to run 'cabal update' to get the latest list of available packages.
-----END CABAL OUTPUT-----
```
The call-stack includes line numbers and package versions which are volatile.

The default seems to be `recordMode RecordMarked`; and the third alternative `DoNotRecord`, does not sound promising.

I notice that the warning I was triggering is hidden explicitly (since cb5a778b3c61861c0c2b51dcf11bc1413f67c5b2):
https://github.com/haskell/cabal/blob/bb08616b0bd96aac880b2ceb37402bc8bd4a067a/cabal-install/src/Distribution/Client/IndexUtils.hs#L225-L227
So I was chasing a wild goose.

>  perhaps parameters `["--config-file", conf]` and `repository` specified in the `conf` file

For some reason, `unpack` does not support the `--config-file` option.  Is this a bug?
```
$ cabal unpack --config-file=FOO BAR
Error: cabal: unrecognized 'unpack' option `--config-file=FOO'
```
`cabal --help` lists `--config-file` as a global option:
```
Global flags:
 -h, --help                     Show this help text
 -V, --version                  Print version information
 --numeric-version              Print just the version number
 --config-file=FILE             Set an alternate location for the config file
...
```
But also, according to `cabal --help`, `unpack` isn't even a valid command:
```
$ cabal --help | grep unpack
(NOTHING)
```

Mikolaj:
> For some reason, unpack does not support the --config-file option. Is this a bug?

Yes, I think so.

I wonder how other tests cope with no config files (and so no `repository: foo.org` defined). @gbaz, @fgaz, any idea? Should we just `echo "repository: foo.org" > ~/.cabal/config` and hope it doesn't break subsequent tests?

andreasabel:
@Mikolaj: It seems like I got the test working now, thanks to your feedback.

The warning contains also an `unrecognized stanza` which is not so desirable, but was good enough to trigger the warning we care about.

Commits should be merged _squashed_.

fgaz:
> unpack isn't even a valid command

`unpack` is a legacy alias for `get`

Mikolaj:
CI fails due to #7707, so let's merge anyway.

Issue #7455, opened 2021-06-22
Should CABAL_DIR really be dropped in the future?


danidiaz:
There was a [recent question](https://stackoverflow.com/questions/68079550/clean-installation-of-haskell-onto-a-windows-10-machine) in Stack Overflow, dealing with how to put the files related to Cabal in a non-standard folder. That question has a pretty straightforward solution which consists in setting the [`CABAL_DIR`](https://cabal.readthedocs.io/en/3.4/installing-packages.html?highlight=CABAL_DIR#environment-variables) environment variable. Once set, the configuration, the store and other cabal content will be contained in the folder `CABAL_DIR` points to.

However the user guide [says the following](https://cabal.readthedocs.io/en/3.4/installing-packages.html?highlight=CABAL_DIR#environment-variables):

> The CABAL_DIR might be dropped in the future, when cabal-install starts to use XDG Directory specification.

Following the XDG specification is a good idea, but should it require abandoning `CABAL_DIR`? That would forgo some versatility. 

A good compromise could be following `CABAL_DIR` if defined, and follow the XDG specification otherwise. Alternatively, perhaps the burden of maintaining both options is not worth it?

jneira:
In windows systems we suffer the long standing issue with max path limits. A way to workaround them for cabal has been using `CABAL_DIR` (ghcup added recently support for windows and uses it). So for that case (and i guess many others) i would not want change my entire xdg tree of dirs but only cabal.

Mikolaj:
Is it possible that XDG Directory specification specifies a way to define non-standard directories instead of the usual ones?

phadej:
Don't forget about `CABAL_CONFIG`. IMHO that's  the way you should specify the configuration, `CABAL_DIR` is confusing, as in examples like

```
CABAL_DIR=/home/oleg/.cabal
```

and the `$CABAL_DIR/cabal.config` (`/home/oleg/.cabal/cabal.config) having e.g.

```
store-dir: /cabalstore
```

means that store dir **is** `/cabalstore` and not `$CABAL_DIR/store` (`/home/oleg/.cabal/store`).

---

[...]

PR #7715, opened 2021-10-07
Fix #7705: only write default configuration if no --config-file

### LINKED ISSUE #7705: `cabal --config-file=NONEXISTENT ...` will write a default configuration to NONEXISTENT
If the FILE given with `--config-file` option does not exist, `cabal` will be so _helpful_ to write a default configuration to the FILE.

I do not think this is a good UI; we also do not write a template package description into a non-existent `.cabal`-file passed on the command line.  

I think writing a default configuration should be limited to an explicit command: 
```
cabal --config-file=FILE user-config init
```
For all other commands, giving a non-existent FILE should be an error.

andreasabel:
Fix #7705: only write default configuration if no config file is default, i.e. neither given via option `--config-file` nor via environment variable `CABAL_CONFIG`.

New behavior, e.g.:
```
$ cabal --config-file=ABSENT info happy

Config file path source is commandline option.
Error: cabal: Config file not found: ABSENT
(Config files can be created via the cabal-command 'user-config init'.)

$ CABAL_CONFIG=missing.file cabal info frown

Config file path source is environment variable CABAL_CONFIG.
Error: cabal: Config file not found: missing.file
(Config files can be created via the cabal-command 'user-config init'.)
```
Whether the hint in parenthesis should be given is up to taste, but information is easier ignored than conjured.

Golden value tests of these errors are included.

---

### DIFF
--- a/doc/installing-packages.rst
+++ b/doc/installing-packages.rst
@@ -6,10 +6,11 @@ Configuration
 Overview
 --------
 
-The global configuration file for ``cabal-install`` is
-``~/.cabal/config``. If you do not have this file, ``cabal`` will create
-it for you on the first call to ``cabal update``. Alternatively, you can
-explicitly ask ``cabal`` to create it for you using
+The global configuration file for ``cabal-install`` is by default
+``$HOME/.cabal/config``. If you do not have this file, ``cabal`` will create
+it for you on the first call to ``cabal update``
+(details see `configuration file discovery`_).
+Alternatively, you can explicitly ask ``cabal`` to create it for you using
 
 .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


andreasabel:
Fixed the unit tests and added a changelog.

jneira:
> 
> 
> > i wonder how many scripts in the wild are relying in this "feature" thinking
> 
> We'll find out soon. ;)
> 
> I'd recommend merging this without much ceremony, because there is an easy workaround and there's still a lot of time till 3.8 for people to object.

fair enough

hasufell:
I don't have a strong opinion here, but if this is merged, I'd argue it's one more reason to not deprecate `CABAL_DIR`, which has much better ergonomics

https://github.com/haskell/cabal/issues/7455

ghcup (and windows and many CIs) use `CABAL_DIR`.

PR #7726, opened 2021-10-08
Fix #7714: use `nroff -man | less` as backend for `cabal man`

### LINKED ISSUE #7714: `cabal man` does not work on macOS
Re #6548:

`cabal man` does not work on macOS
```
$ cabal man 
man: invalid option -- l
man, version 1.6c

usage: man [-adfhktwW] [section] [-M path] [-P pager] [-S list]
	[-m system] [-p string] name ...

  a : find all matching entries
...
fd:13: commitBuffer: resource vanished (Broken pipe)
```

Also, there is a typo in an error message:
```
$ cabal man happy
Error: cabal: 'manpage' doesn't take any extra arguments: happy
```

andreasabel:
Fix #7714: use `nroff -man /dev/stdin | less` as backend for `cabal man`

WIP: This does not work yet, I need some help with constructing the pipe correctly in Haskell.

I seem to follow the example at
https://stackoverflow.com/questions/41030168/haskell-using-the-handle-created-from-createprocess-and-createpipe-to-pipe-int
which works for me, but my patch does not seem to invoke `less` correctly.

---

### DIFF
--- a/cabal-install/src/Distribution/Client/Manpage.hs
+++ b/cabal-install/src/Distribution/Client/Manpage.hs
@@ -25,7 +25,6 @@ import Prelude ()
 import qualified Data.List.NonEmpty as List1
 
 import Distribution.Client.Init.Utils   (trim)
-  -- TODO #7744: move 'trim' to a more canonical place
 import Distribution.Client.ManpageFlags
 import Distribution.Client.Setup        (globalCommand)
 import Distribution.Simple.Command
--- a/cabal-install/src/Distribution/Client/Manpage.hs
+++ b/cabal-install/src/Distribution/Client/Manpage.hs
@@ -34,6 +34,8 @@ import Distribution.Simple.Utils
   ( IOData(..), IODataMode(..), createProcessWithEnv, ignoreSigPipe, rawSystemStdInOut )
 import qualified Distribution.Verbosity as Verbosity
 import System.IO                        (hClose, hPutStr)
+import System.Environment               (lookupEnv)
+import System.FilePath                  (takeFileName)
 
 import qualified System.Process as Process
 
@@ -78,11 +80,14 @@ manpageCmd pname commands flags
 
         unless (ec1 == ExitSuccess) $ exitWith ec1
 
+        pager <- fromMaybe "less" <$> lookupEnv "PAGER"
+        -- 'less' is borked with color sequences otherwise
+        let pagerArgs = if takeFileName pager == "less" then ["-R"] else []
         -- Pipe output of @nroff@ into @less@
         (Just inLess, _, _, procLess) <- createProcessWithEnv
           Verbosity.normal
-          "less"
-          []
+          pager
+          pagerArgs
           Nothing  -- Inherit working directory
           Nothing  -- Inherit environment
           Process.CreatePipe  -- in
--- a/cabal-install/src/Distribution/Client/Manpage.hs
+++ b/cabal-install/src/Distribution/Client/Manpage.hs
@@ -22,12 +22,17 @@ module Distribution.Client.Manpage
 
 import Distribution.Client.Compat.Prelude
 import Prelude ()
+import qualified Data.List.NonEmpty as List1
 
+import Distribution.Client.Init.Utils   (trim)
+  -- TODO #7744: move 'trim' to a more canonical place
 import Distribution.Client.ManpageFlags
 import Distribution.Client.Setup        (globalCommand)
-import Distribution.Compat.Process      (createProcess)
 import Distribution.Simple.Command
 import Distribution.Simple.Flag         (fromFlagOrDefault)
+import Distribution.Simple.Utils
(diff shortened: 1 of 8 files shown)


andreasabel:
> WIP: This does not work yet, I need some help with constructing the pipe correctly in Haskell.

Weirdly enough, it could be because the generated text is too long (800kB).  If I truncate it to 192kB (this is a precise hard bound, 1 char more is too much(!)), it works...

phadej:
> The documentation suggests that a pipe is created when the `CreatePipe` constructor is supplied

Yes a pipe is created, for example `stdout`. Library gives the writing end to the spawned process and gives the reading end to you. A pipe (= two FDs) is created, but you get only one end of it.

phadej:
For the record, e.g. `git commit --help` seems to simply call `man git-commit` as git installs the man pages. That works on BSD-derivatives too, obviously. Also doesn't need `*roff` to be available.

phadej:
> and do stdin writing as elsewhere.

refers to the wish that we rather add a new helper to `Distribution.Simple.Utils` to spawn processes with inherited outputs, rather then going low-level in a particular command implementation.

andreasabel:
Can this PR be merged now? I may not be perfect, but it improves on the status quo, I think.

Mikolaj:
Any further comments, anybody? 

@hasufell, would you like to review?

cabal-install/src/Distribution/Client/Manpage.hs
```diff
@@ -22,12 +22,17 @@ module Distribution.Client.Manpage
 
 import Distribution.Client.Compat.Prelude
 import Prelude ()
+import qualified Data.List.NonEmpty as List1
 
+import Distribution.Client.Init.Utils   (trim)
+  -- TODO #7744: move 'trim' to a more canonical place
```
fgaz, line 28: Looks like #7744 is fixed

cabal-install/src/Distribution/Client/Manpage.hs
```diff
@@ -50,24 +55,44 @@ manpageCmd pname commands flags
...
+          Nothing  -- Inherit environment
+          (Just $ IODataText contents)
+          IODataModeText
+
+        unless (ec1 == ExitSuccess) $ exitWith ec1
+
+        -- Pipe output of @nroff@ into @less@
+        (Just inLess, _, _, procLess) <- createProcessWithEnv
+          Verbosity.normal
+          "less"
```
hasufell, line 85: `most` is much better and also seems to have no problem with the newlines I think.

One way would be to check the `$PAGER` env var and fall back to `less` if that isn't set.

hasufell: https://github.com/andreasabel/cabal/pull/1

hasufell: using `less` is borked here:

## Less

Screenshot_2022-01-20_11-51-49

## Most

Screenshot_2022-01-20_11-52-06

fgaz: less needs `-R` (or `-r`) to display colours

hasufell: > less needs `-R` (or `-r`) to display colours

Is that safe to add unconditionally?

```
       -r or --raw-control-chars
              Causes "raw" control characters to be displayed.  The default is to display control characters using the caret notation; for  ex‐
              ample,  a  control-A (octal 001) is displayed as "^A".  Warning: when the -r option is used, less cannot keep track of the actual
              appearance of the screen (since this depends on how the screen responds to each type of control character).  Thus,  various  dis‐
              play problems may result, such as long lines being split in the wrong place.

              USE OF THE -r OPTION IS NOT RECOMMENDED.
```

hasufell: Seems `-R` should be ok and also works under FreeBSD

hasufell: https://github.com/andreasabel/cabal/pull/1/commits/c4140b19e827a97fc9152dc60fbe763ce5d02529

hasufell (changes requested):
@andreasabel I added some fixes here: https://github.com/andreasabel/cabal/pull/1

Mikolaj (approved):

andreasabel:
Mergiofy was waiting on the "Artifacts" workflows, but these seem gone.  So I deleted them from the required statuses list for master in Settings/Branches.  Hope that was intended.

Mikolaj:
> Mergiofy was waiting on the "Artifacts" workflows, but these seem gone. So I deleted them from the required statuses list for master in Settings/Branches. Hope that was intended.

Yes, it was, thank you!

fgaz:
This will need a changelog entry

andreasabel:
> This will need a changelog entry

@fgaz: See https://github.com/haskell/cabal/pull/7935

PR #7746, opened 2021-10-13
Remove any remnant of the 'world' file

### LINKED ISSUE #6767: Remove World file functionality
It's used in `v1-install`.

I wonder if it works properly currently.

fendor:
In v1-install, 'world' was used to trace what packages have been
installed, and re-install on-demand everything listed in it. However, in
v2-install, this is no longer needed and this file leads to code bloat.
Additionally, 'world' code-path is probably not up-to-date, remove it instead of
having partial features.

closes #6767


fendor:
Oh, I forgot to update documentation

jneira:
lgtm thanks for the clean up

Mikolaj:
BTW, none of the v1- users is admitting to use `world`. I suspect they are biding their time until 3.8 is out to get that perfect storm.

gbaz:
@Mikolaj @fendor so I realized a small problem with this pr. in particular, people like me still have old cabal configs that were created when world files existed.

So using HEAD of cabal I get warnings like `Warning: /Users/gershom/.cabal/config: Unrecognized field world-file on line
29`

I know how to fix this (update my config) but I think this will be a confusing PITA for cabal users. Can we have a PR to somehow suppress this warning?

andreasabel:
As a general principle, if one removes a component of a language, like the `world-file` syntax from the CABAL specification language, one should not remove it from the parser.  Rather, keep it in the parser so that you can give a meaningful error message that this component is no longer supported.
A meaningful error message would be "World files were removed in version 3.foo.bar.0".

See e.g. the removal of the `delimiters` pragma in BNFC: https://github.com/BNFC/bnfc/blob/b9d96e87a6f9051792555b7060cdd5ad2d31ff80/source/src/BNFC/GetCF.hs#L373-L376

Mikolaj:
That's a good idea. In particular, we could tell that it's usually just the matter to remove the "world-whatever" line from the config, because almost no users depend on it any more and so no further action is necessary.

gbaz:
In this case I don't think we should give an error message or warning -- just silently discard. It was added by default in previous cabals and forcing users to all remove it by hand when it was never cared about before will cause friction.

andreasabel:
We could have a warning that only shows with `-v2` or `-v3`.

PR #7754, opened 2021-10-16
Select buildable targets on 3.6

tomjaguarpaw:
I found the `selectBuildableTargets...` functions a little bit harder to read than I think they should have been. The comprehensions they contain are all basically the same, so I factored out the commonality.  I think it is a modest readability improvement. Small quality of life improvements really make it much easier for new developers to start contributing.

What do you think?

----

Alternative to https://github.com/haskell/cabal/pull/7080 because I'm not sure which branch PR's should be based on. If PRs should be based on master please close this one.


jneira:
yeah PRS are against master and backported if needed

PR #7756, opened 2021-10-16
Remove unused ConstraintSourceModifiedAddSourceDep

tomjaguarpaw:
This seems to be unused. Can we get rid of it?


tomjaguarpaw:
Looks like the last use was removed in https://github.com/haskell/cabal/commit/6248c74280a54382f85079f58a1ea40ba0db9476 by @phadej  when sandboxes were removed.

tomjaguarpaw:
> I would ping the author of the commit

Which commit do you mean?

jneira:
the one which added the definition this is removing, but the commit you have mentioned is informative enough

PR #7757, opened 2021-10-16
Terminate child processes on sigINT/sigTERM

hasufell:
For more information, check out https://github.com/ndmitchell/shake/issues/169

Questions that may need to be discussed:

1. Is the additional dependency allowed? It's a small single-module package that could be incorporated. It seems to work even on windows, utilizing some [foreign calls](https://hackage.haskell.org/package/signal-0.1.0.4/docs/src/System-Signal.html#installHandler) (also see [MS documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-160)), maybe @Mistuke can comment
2. I don't see an obvious way to clean up the handlers themselves cross-platform. Will consecutive calls to running external binaries cause problems here?
3. How safe is this in general?
4. Another approach is to use [`exec`](https://hackage.haskell.org/package/rio-0.1.21.0/docs/RIO-Process.html#v:exec) (at least for `cabal run/exec`), but that doesn't seem to be available on windows.

### DIFF
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -46,7 +46,8 @@ library
     filepath   >= 1.3.0.1  && < 1.5,
     pretty     >= 1.1.1    && < 1.2,
     process    >= 1.1.0.2  && < 1.7,
-    time       >= 1.4.0.1  && < 1.12
+    time       >= 1.4.0.1  && < 1.12,
+    signal
 
   if flag(bundled-binary-generic)
     build-depends: binary >= 0.5.1.1 && < 0.7
--- a/Cabal/src/Distribution/Compat/Process.hs
+++ b/Cabal/src/Distribution/Compat/Process.hs
@@ -6,10 +6,13 @@ module Distribution.Compat.Process (
     rawSystem,
     -- * Additions
     enableProcessJobs,
+    cleanUpProcessOnInterrupt,
     ) where
 
+import Control.Monad (forM_)
 import System.Exit (ExitCode (..))
 import System.IO   (Handle)
+import System.Signal
 
 import           System.Process (CreateProcess, ProcessHandle)
 import qualified System.Process as Process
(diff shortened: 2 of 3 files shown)


hasufell:
Note that we don't really re-throw the signal to the child, but rather run `cleanupProcess`, which runs [terminateProcess](https://hackage.haskell.org/package/process-1.6.13.2/docs/src/System.Process.html#terminateProcess), which runs:

* on unix: https://github.com/haskell/process/blob/e448450049ac51a6fcd5d1f0be86448d0931d71a/cbits/posix/runProcess.c#L199-L203 (equivalent to `SIGTERM`)
* on windows: https://github.com/haskell/process/blob/master/cbits/win32/runProcess.c#L467-L494 (this runs [`TerminateProcess`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess))

hasufell:
So I tracked down the callstack for the downloaders that invoke curl etc... and it seems they also invoke `rawSystemStdInOut` asynchronously via `withAsync`. I assume we won't really kill all curl processes that way?

hasufell:
Another possibility is to do it like `rio`:

https://hackage.haskell.org/package/rio-0.1.21.0/docs/src/RIO.Process.html#exec

```hs
-- | Execute a process within the configured environment.
--
-- Execution will not return, because either:
--
-- 1) On non-windows, execution is taken over by execv of the
-- sub-process. This allows signals to be propagated (#527)
--
-- 2) On windows, an 'ExitCode' exception will be thrown.
--
-- @since 0.0.3.0
exec :: (HasProcessContext env, HasLogFunc env) => String -> [String] -> RIO env b
#ifdef WINDOWS
exec = execSpawn
#else
exec cmd0 args = do
    wd <- view workingDirL
    envStringsL <- view envVarsStringsL
    cmd <- preProcess cmd0
    withProcessTimeLog wd cmd args $ liftIO $ do
      for_ wd setCurrentDirectory
      executeFile cmd True args $ Just envStringsL
#endif
```

But this seems to me it would only work nicely on unix.

robx:
Very much in favour of doing something here! I think `cabal` *should* ensure all its children exit on `SIGTERM`.

Some vague thoughts:

- I agree that for `run` and `exec`, `exec(3)`-like behavior would be ideal. It also seems like it would be worthwhile fixing `run/exec` first if that's easier.
- Regarding the lack of clean-up, I'm also a bit unsure what it means to keep all of these handlers around. I imagine we might run into issues such as running out of file descriptors keeping all those fd handles around?
- I'm wondering whether we could do something like:
  - install SIGTERM handler that swallows the signal and throws an async exception at the main thread (not exactly sure whether that's sensible and can be achieved)
  - bracket the `createProcess` calls with `cleanupProcess`

PR #7764, opened 2021-10-20
Cabal: Use -flink-rts when available

### LINKED ISSUE #7763: Cabal incorrectly constructs libHSrts library path
**Describe the bug**

A recent GHC bug ([#20520](https://gitlab.haskell.org/ghc/ghc/-/issues/20520)) brought to light the fact that Cabal includes knowledge of internal implementation details of GHC (specifically, the particular file name of `libHSrts`). This appears to have been introduced by 382143aa73 and 1c07f8354b8e3c9900a5a913334bc9831ed7f7a8.

This naming is unstable and has indeed changed in recent releases, hence the ghc#20520.

bgamari:
Previously Cabal did quite some headstands to link against libHSrts.
Note only this is complex but it couples very tightly to GHC's
implementation. Thankfully, as of GHC 9.0 GHC provides a `-flink-rts`
flag for precisely this purpose. Use it when available.

Fixes #7763.

---

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -1294,9 +1294,7 @@ gbuild verbosity numJobs pkg_descr lbi bm clbi = do
       implInfo   = getImplInfo comp
       runGhcProg = runGHC verbosity ghcProg comp platform
 
-  let (bnfo, threaded) = case bm of
-        GBuildFLib _ -> popThreadedFlag (gbuildInfo bm)
-        _            -> (gbuildInfo bm, False)
+  let bnfo = gbuildInfo bm
 
   -- the name that GHC really uses (e.g., with .exe on Windows for executables)
   let targetName = gbuildTargetName lbi bm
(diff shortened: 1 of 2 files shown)


bgamari:
This appears to fix the original ticket on Windows. However, my testing on Linux suggests that this feature is actually slightly broken on that platform. Specifically, `Cabal` must ensure that objects linked into a shared object are built with `-fPIC` (which may mean using the `dyn` way, since all transitive objects in the final link must be position independent). It's quite unclear how the original design intended this to be ensured.

Mikolaj:
@bgamari: you mentioned Windows and Linux. Could this affect M1 as well, as in #7837? If so, we can ask @vmchale to test once the PR is complete.

bgamari:
> Is it broken with GHC 8.10 as well? I assume it was broken for a long time?

Indeed, as far as I can tell this is just a hole in the design. Specifically, one can only expect to build a `foreign library` on most platforms in the `dynamic` way. Perhaps Cabal should throw an error to catch this case?

Mikolaj:
> My suspicion is that Cabal should either implicitly enable `--enable-dynamic-libraries' when building `foreign library`s on some platforms or refuse to build them if it's disabled.

Sounds reasonable. I think, as long as the error messages are clear, it's less surprising when stuff refuses to build and asks to enable `--enable-dynamic-libraries` than when it's enabled implicitly, the user doesn't notice and is utterly confused why the build products are not static as requested.

bgamari:
Unfortunately the issue is even more confused than I initially thought. It appears that, in an attempt to mitigate the issue described above, `GHC. `Distribution.Simple.GHC.gbuild` explicitly enables `-fPIC` when building a foreign library. Meanwhile, it passes `-static`, a combination that makes very little sense and causes GHC with this patch to link against the wrong RTS.

bgamari:
It appears that the `-static` flag arises from `Distribution.Simple.GHC.gbuildNeedDynamic`. Specifically, that function returns `False` if the library type is standalone. This seems to be quite intentional but it's not clear to me *why* this it would do this and, as far as I can tell, it's wrong.

bgamari:
> It appears that the `-static` flag arises from `Distribution.Simple.GHC.gbuildNeedDynamic`. Specifically, that function returns `False` if the library type is standalone. This seems to be quite intentional but it's not clear to me _why_ it would do this and, as far as I can tell, it's wrong.

Thinking about this a bit more, I suspect that this is due to the fact that the author wanted a "standalone" library to have no dynamic dependencies. That is, they expected GHC to link against static Haskell dependencies. However, it's unclear to me how this could have ever worked on platforms that require dynamic libraries to be PIC.

hsyl20:
> Ultimately GHC doesn't make things easier here as it defines -dynamic to serve a dual-purpose: it both tells GHC to link against Haskell code dynamically and it tells GHC to produce PIC code. 

Couldn't we use `-fexternal-dynamic-refs` instead?

Mikolaj:
@bgamari: any news?

Mikolaj:
@Bodigrim points our attention to https://gitlab.haskell.org/ghc/ghc/-/issues/20520#note_411402, citing:

> Unfortunately, this is at its core a Cabal bug. I have tried to fix this bug in [Cabal #7764](https://github.com/haskell/cabal/pull/7764) but have run into inconsistencies in the current implementation of foreign-library that I can't currently explain. Specifically, Cabal passes -static to GHC when building a type: standalone library. However, this is inconsistent with the users intent to build a dynamic library. I'm afraid the fix for this will need to be bumped to [%9.2.3](https://gitlab.haskell.org/ghc/ghc/-/milestones/376) at the very least.

@bgamari: do you have any vague suggestions how to fix that in Cabal?

gbaz:
I think based on discussion that this is a strict improvement, but there remain cases that make no sense in general with the static/dynamic issue. So we should merge this, but also open a new ticket trying to specify what the semantics actually are.

Bodigrim:
This does not seem to resolve the reproducer in https://gitlab.haskell.org/ghc/ghc/-/issues/20520, and I'm not qualified to jusge about other benefits.

Mikolaj:
> This does not seem to resolve the reproducer in https://gitlab.haskell.org/ghc/ghc/-/issues/20520, and I'm not qualified to jusge about other benefits.

Oh dear, I thought it does, based on "This appears to fix the original ticket on Windows."

@bgamari: do we need to do anything more to fix the original issue? Remove the `--static` that cabal inserts? Add `-fexternal-dynamic-refs` @hsyl20 is suggesting?

Bodigrim (approved):
Sorry for confusion, I tested a wrong branch. This indeed fixes the reproducer.

Mikolaj (approved):

Mikolaj:
@MergifyIO rebase

jneira:
This should had included a changelog entry, added in #8111

Issue #6528, opened 2020-02-05
[Feature] remote freeze files


hasufell:
If you want to use the same freeze file for a plethora of internal projects, then a "global" freeze file might make sense. Currently this would require copying and synchronising such a file across repositories, or having a mono-repo. Being able to set a remote location as a freeze file in `cabal.project` would solve this use case.

Cli options could allow to say what to do in case fetching fails:

* abort resolution
* continue without a freeze file

Kleidukos:
This is a great idea and will certainly help bringing cabal to feature parity with stack in enterprise environments.

merijn:
@Hvr spent considerable time to keep the critical path to produce a `plan.json` fast, so that a no-op `cabal build` and `cabal run` latency would be <100ms in order to keep unconditional `cabal build` and `cabal run` usable in interactive settings or when integrating with other build tools.

Such a remote fetch for a freeze file would be and almost certainly ruin this criticial path.

cdsmith:
IMO, it's very reasonable to be concerned about re-fetching a remote freeze file every time you run `cabal build`, and this is a convincing reason not to do it that way.

There are other ways to do it:

1. A cache that would only re-fetch a freeze file occasionally.  Taking an extra couple seconds once a day isn't anything like the same barrier.  But the unpredictable timing of picking up changes would be unfortunate.
2. My favorite: Give names to packaged constraint sets, and fetch them as part of `cabal update`.  This opens the door to future work mirroring them in case one source is down, migrating hosting or even hosting on Hackage, integrating with signing and other security measures, etc.

It's not like the latency concern is a dead end for the concept.  It's just an argument against one of many possible implementations.

hasufell:
Wrt 2.: Sounds like a controversial choice to me, since remote freeze files very likely would be per project (although there might be a global one too). `cabal update` so far has no project specific semantics afaik, so this would be a pretty drastic change in workflow.

As for the latency concerns: I do not share them.

1. This is an optional feature
2. Users who rely on it will care more about the feature than latency. It has zero impact on hackage and no impact on repos you use through `source-repository-package`. If you don't want it, it doesn't impact you.
3. Fetch strategies could be added to the configuration, to allow for immutable remote freeze files (uniquely identified by the name). Then you would only fetch once.

I believe that `configure` would be the correct place for this, since it also fetches remote repos at that stage.

gbaz:
[...]

PR #7783, opened 2021-10-28
Extended project files (conditionals and imports)

gbaz:
Adds conditionals and imports to cabal.project files. The basic idea is we introduce a ProjectSkeleton that reuses the conditional architecture from cabal files (CondTree) to hold a tree of project data, resolving imports as we go. Then the project skeletons need to be instantiated with concrete information for OS, compiler, platform before use.

Additional checks (not yet added) should prevent setting of compiler info within conditionals (since we can have conditionals depend on this info). Additionally, we should enforce that cabal.project.local files cannot have conditionals or imports, as we don't have to want to write machine generation to preserve that structure for now (necessary for cabal reconfigure).

---

### DIFF
--- a/Cabal-syntax/src/Distribution/System.hs
+++ b/Cabal-syntax/src/Distribution/System.hs
@@ -151,8 +151,8 @@ buildOS = classifyOS Permissive System.Info.os
 -- ------------------------------------------------------------
 
 -- | These are the known Arches: I386, X86_64, PPC, PPC64, Sparc,
--- Arm, AArch64, Mips, SH, IA64, S39, Alpha, Hppa, Rs6000, M68k,
--- Vax, and JavaScript.
+-- Arm, AArch64, Mips, SH, IA64, S390, S390X, Alpha, Hppa, Rs6000,
+-- M68k, Vax, and JavaScript.
 --
 -- The following aliases can also be used:
 --    * PPC alias: powerpc
@@ -164,7 +164,7 @@ buildOS = classifyOS Permissive System.Info.os
 --
 data Arch = I386  | X86_64  | PPC  | PPC64 | Sparc
           | Arm   | AArch64 | Mips | SH
-          | IA64  | S390
+          | IA64  | S390    | S390X
           | Alpha | Hppa    | Rs6000
           | M68k  | Vax
           | JavaScript
@@ -178,7 +178,7 @@ instance NFData Arch where rnf = genericRnf
 knownArches :: [Arch]
 knownArches = [I386, X86_64, PPC, PPC64, Sparc
               ,Arm, AArch64, Mips, SH
-              ,IA64, S390
+              ,IA64, S390, S390X
               ,Alpha, Hppa, Rs6000
               ,M68k, Vax
               ,JavaScript]
--- a/Cabal-tests/Cabal-tests.cabal
+++ b/Cabal-tests/Cabal-tests.cabal
@@ -155,8 +155,6 @@ test-suite hackage-tests
   type:               exitcode-stdio-1.0
   main-is:            HackageTests.hs
 
-  if os(windows)
-    buildable: False
 
   hs-source-dirs:     tests
   build-depends:
--- a/Cabal-tests/tests/HackageTests.hs
+++ b/Cabal-tests/tests/HackageTests.hs
@@ -63,7 +63,7 @@ import Data.TreeDiff.Pretty          (ansiWlEditExprCompact)
 parseIndex :: (Monoid a, NFData a) => (FilePath -> Bool)
            -> (FilePath -> B.ByteString -> IO a) -> IO a
 parseIndex predicate action = do
-    cabalDir   <- getAppUserDataDirectory "cabal"
+    cabalDir   <- getCabalDir
     configPath <- getCabalConfigPath cabalDir
     cfg        <- B.readFile configPath
     cfgFields  <- either (fail . show) pure $ Parsec.readFields cfg
@@ -74,6 +74,11 @@ parseIndex predicate action = do
         tarName repo = repoCache </> repo </> "01-index.tar"
     mconcat <$> traverse (parseIndex' predicate action . tarName) repos
   where
+    getCabalDir = do
+        mx <- lookupEnv "CABAL_DIR"
+        case mx of
+            Just x  -> return x
+            Nothing -> getAppUserDataDirectory "cabal"
     getCabalConfigPath cabalDir = do
         mx <- lookupEnv "CABAL_CONFIG"
         case mx of
--- a/Cabal-tests/tests/UnitTests/Distribution/Utils/Structured.hs
+++ b/Cabal-tests/tests/UnitTests/Distribution/Utils/Structured.hs
@@ -27,9 +27,9 @@ tests = testGroup "Distribution.Utils.Structured"
     -- The difference is in encoding of newtypes
 #if MIN_VERSION_base(4,7,0)
     , testCase "GenericPackageDescription" $
-      md5Check (Proxy :: Proxy GenericPackageDescription) 0xafbf1cfb39ece402a2008d07f5e5ffd8
+      md5Check (Proxy :: Proxy GenericPackageDescription) 0x3713da01e295903b046123a7cba0fb14
     , testCase "LocalBuildInfo" $
-      md5Check (Proxy :: Proxy LocalBuildInfo) 0xd8b4c7f04e75345f0303fe2c3093bc29
+      md5Check (Proxy :: Proxy LocalBuildInfo) 0xd4d0449f6db57c79b68b6e412b5f4f55
 #endif
     ]
 
--- a/Cabal/src/Distribution/Backpack/Id.hs
+++ b/Cabal/src/Distribution/Backpack/Id.hs
@@ -41,7 +41,7 @@ computeComponentId
 computeComponentId deterministic mb_ipid mb_cid pid cname mb_details =
     -- show is found to be faster than intercalate and then replacement of
     -- special character used in intercalating. We cannot simply hash by
-    -- doubly concating list, as it just flatten out the nested list, so
+    -- doubly concatenating list, as it just flatten out the nested list, so
     -- different sources can produce same hash
     let hash_suffix
             | Just (dep_ipids, flags) <- mb_details
--- a/Cabal/src/Distribution/GetOpt.hs
+++ b/Cabal/src/Distribution/GetOpt.hs
@@ -83,7 +83,7 @@ data OptHelp = OptHelp {
 -- second argument.
 usageInfo :: String                    -- header
           -> [OptDescr a]              -- option descriptors
-          -> String                    -- nicely formatted decription of options
+          -> String                    -- nicely formatted description of options
 usageInfo header optDescr = unlines (header : table)
   where
     options = flip map optDescr $ \(Option sos los ad d) ->
--- a/Cabal/src/Distribution/Simple/PreProcess.hs
+++ b/Cabal/src/Distribution/Simple/PreProcess.hs
@@ -722,6 +722,7 @@ platformDefines lbi =
       SH          -> []
       IA64        -> ["ia64"]
       S390        -> ["s390"]
+      S390X       -> ["s390x"]
       Alpha       -> ["alpha"]
       Hppa        -> ["hppa"]
       Rs6000      -> ["rs6000"]
--- a/cabal-install-solver/src/Distribution/Solver/Modular/Dependency.hs
+++ b/cabal-install-solver/src/Distribution/Solver/Modular/Dependency.hs
@@ -212,7 +212,7 @@ qualifyDeps QO{..} (Q pp@(PackagePath ns q) pn) = go
     -- dependency on R. We do not do this for the base qualifier however.
     --
     -- The inherited qualifier is only used for regular dependencies; for setup
-    -- and base deppendencies we override the existing qualifier. See #3160 for
+    -- and base dependencies we override the existing qualifier. See #3160 for
     -- a detailed discussion.
     inheritedQ :: Qualifier
     inheritedQ = case q of
--- a/cabal-install/src/Distribution/Client/FileMonitor.hs
+++ b/cabal-install/src/Distribution/Client/FileMonitor.hs
@@ -453,7 +453,7 @@ checkFileMonitorChanged
       where
         -- In fileMonitorCheckIfOnlyValueChanged mode we want to guarantee that
         -- if we return MonitoredValueChanged that only the value changed.
-        -- We do that by checkin for file changes first. Otherwise it makes
+        -- We do that by checking for file changes first. Otherwise it makes
         -- more sense to do the cheaper test first.
         checkForChanges :: IO (Maybe (MonitorChangedReason a))
         checkForChanges
@@ -768,7 +768,7 @@ probeMonitorStateGlobRel _ _ root dirName
 
 
     return (MonitorStateGlobFiles glob mtime' children)
-    -- Again, we don't force a cache rewite with 'cacheChanged', but we do use
+    -- Again, we don't force a cache rewrite with 'cacheChanged', but we do use
     -- the new mtime' if any.
   where
     probeMergeResult :: MergeResult (FilePath, MonitorStateFileStatus) FilePath
--- a/cabal-install/src/Distribution/Client/GlobalFlags.hs
+++ b/cabal-install/src/Distribution/Client/GlobalFlags.hs
@@ -112,7 +112,7 @@ data RepoContext = RepoContext {
     --
     -- NOTE: It is important that we don't eagerly initialize the transport.
     -- Initializing the transport is not free, and especially in contexts where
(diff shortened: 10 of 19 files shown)


phadej:
>  we should enforce that cabal.project.local files cannot have conditionals or imports

I'd rather make `reconfigure` fail in that case. (I never use `cabal configure` or `cabal reconfigure`, but I do write my `.local` files manually for various in development needs).

gbaz:
This last pr finishes up the core logic -- now it needs some testing and debugging.

Mikolaj:
@mergifyIO rebase

cabal-testsuite/PackageTests/ConditionalAndImport/cabal.project
```diff
@@ -0,0 +1,3 @@
+packages: .
+
+import extra.project
```
phadej: `.cabal` files use fields, not section headers.

My reasoning was that why to allow having

```
import foo
  something-in-section: ...
```

to begin with.

gbaz: you're commenting on an older version. with the parser cleanup this uses `import: foo` as you seem to desire.

phadej: :+1:

cabal-install/src/Distribution/Client/CmdOutdated.hs
```diff
@@ -86,11 +83,16 @@ import Distribution.Simple.Command
...
+import Distribution.Client.HttpUtils
+import Distribution.Utils.NubList
+         ( fromNubList )
+
 
 import qualified Data.Set as S
 import System.Directory
     ( getCurrentDirectory, doesFileExist )
 
+
```
Kleidukos, line 95: Extraneous empty line

Kleidukos (approved):
Only an extraneous empty line, but overall I'm satisfied with how this PR looks.

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -93,11 +104,127 @@ import Distribution.Simple.Command
...
+       where addWarnings (ParseOk ws' x') = ParseOk (ws' ++ ws) x'
+             addWarnings x' = x'
+    liftPR _ (ParseFailed e) = pure $ ParseFailed e
+
+    fetchImportConfig :: ProjectConfigImport -> IO BS.ByteString
+    fetchImportConfig pci = case parseURI pci of
+         Just uri -> do
+            let fp = cacheDir </> map (\x -> if isPathSeparator x then 'X' else x) (show uri) -- TODO can we do better?
+            createDirectoryIfMissing True cacheDir
+            _ <- downloadURI httpTransport verbosity uri fp
```
hasufell, line 212: I'm not sure from this piece of code if and how these remote files are cached. I think we discussed that wrt remote freeze files somewhere and "cache strategy" here: https://github.com/haskell/cabal/issues/6528#issuecomment-851846895

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -93,11 +104,127 @@ import Distribution.Simple.Command
...
+            BS.readFile fp
```
hasufell, line 213: Would it be overkill to add a max size limit here for remote files (configurable?), given that we read these strictly into memory without further ado. Or maybe I'm just paranoid.

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -93,11 +104,127 @@ import Distribution.Simple.Command
...
+    liftPR :: (a -> IO (ParseResult b)) -> ParseResult a -> IO (ParseResult b)
+    liftPR f (ParseOk ws x) = addWarnings <$> f x
+       where addWarnings (ParseOk ws' x') = ParseOk (ws' ++ ws) x'
+             addWarnings x' = x'
+    liftPR _ (ParseFailed e) = pure $ ParseFailed e
+
+    fetchImportConfig :: ProjectConfigImport -> IO BS.ByteString
+    fetchImportConfig pci = case parseURI pci of
+         Just uri -> do
+            let fp = cacheDir </> map (\x -> if isPathSeparator x then 'X' else x) (show uri) -- TODO can we do better?
```
hasufell, line 210: I think on windows, we also want to get rid of `:`, because that's not a valid filename character. So we could run [`makeValid`](https://hackage.haskell.org/package/filepath-1.4.2.2/docs/System-FilePath.html#v:makeValid) first and then git rid of the separators. Maybe `_` is better than `X` (makeValid also uses that).

So on windows:

```
> fmap (\x -> if isPathSeparator x then '_' else x) $ W.makeValid "https://some.remote.source/subdir/cabal.config"
"https___some.remote.source_subdir_cabal.config"```

File: cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs (hunk 3 above)
Comment (gbaz) [reply cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs:212]: This uses the same caching strategy used by cabal's caching of metadata from hackage, etc. which attempts to use built in http features with the `if-modified-since` header and etags.

File: cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs (hunk 5 above)
Comment (gbaz) [reply cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs:210]: Nice suggestion -- and good point about makeValid. I'll do that.

File: cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs (hunk 3 above)
Comment (hasufell) [reply cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs:212]: I think that might still be excessive in some cases, e.g. when using `cabal list-bins`.

In ghcup, relying only on etags etc causes a considerable slow down as well of 200-300ms (much more when using wget). There I use an additional  5minute period after last file access to determine whether to invoke any http methods.

Not sure that's the right thing here. Maybe we do want this configurable.

**Comment by gbaz** [comment]
changelog added, ready for review and merge

**Comment by hasufell** [comment]
Do we protect against accidental cyclic imports?

**Comment by gbaz** [comment]
good point, added.

**Comment by hasufell** [comment]
I tested this locally and it worked fine.

File: cabal-install/src/Distribution/Client/CmdOutdated.hs [hunk 6]
```diff
@@ -220,18 +220,23 @@ outdatedAction (ProjectFlags{flagProjectFileName}, OutdatedFlags{..}) _targetStr
...
 
     sourcePkgDb <- IndexUtils.getSourcePackages verbosity repoContext
+    (comp, platform, _progdb) <- configCompilerAux' configFlags
     deps <- if freezeFile
             then depsFromFreezeFile verbosity
             else if newFreezeFile
-                then depsFromNewFreezeFile verbosity mprojectFile
-                else depsFromPkgDesc verbosity comp platform
+                then do
+                       httpTransport <- configureTransport verbosity
```
Mikolaj, line 234: Why is this needed with `newFreezeFile`, but not with `freezeFile`? And how is this related to the main PR's topic?

cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs
```diff
@@ -94,11 +105,130 @@ import Distribution.Simple.Command
...
+         (ParseUtils.F l "import" importLoc) ->
+            if importLoc `elem` seenImports
+              then pure . parseFail $ ParseUtils.FromString ("cyclical import of " ++ importLoc) (Just l)
+              else do
+                let fs = fmap (\z -> CondNode z [importLoc] mempty) $ fieldsToConfig (reverse acc)
+                res <- parseProjectSkeleton cacheDir httpTransport verbosity (importLoc : seenImports) importLoc =<< fetchImportConfig importLoc
+                rest <- go [] xs
+                pure . fmap mconcat . sequence $ [fs, res, rest]
+         (ParseUtils.Section l "if" p xs') -> do
+                           subpcs <- go [] xs'
```
Mikolaj, line 171: Why is this block indented by 18 characters, while the previous, analogous block is indented by 3 characters?

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
```diff
@@ -0,0 +1,3 @@
+packages: .
+
+import: cabal-cyclical.project
```
jneira, line 3: Maybe i am a little bit paranoic and probably the code already take it in account but maybe testing an indirect cycle would be good: cabal-root.project -> cabal-child.project -> cabal-root.project

jneira (commented):
This pr is amazing and it will make cabal.project dramatically more flexible.
One caveat i can think of it is you will not be able to parse cabal.project's with a simple grep.
Doing that is fragile indeed but you have a static set of possible sources of project config and this change will make it so fragile than it cant be used anymore. And there are script out there doing such greps. Adding "ifs" to a config "language" is not for free, neither from an user pov.

But i agree is totally necessary and break such fragile scripts is justified.

But it think it would be good to merge `cabal status` asap after this, before the release if possible, to allow users other ways to query the final configuration of the project and gives them another, better alternative.

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
jneira: Also i miss a simple positive test over imports and conditionals, to exercise and demosntrate the pr main goal.

hasufell:
> One caveat i can think of it is you will not be able to parse cabal.project's with a simple grep.

I don't know of any such script. That's what https://hackage.haskell.org/package/cabal-install-parsers-0.4.5/docs/Cabal-Project.html is for and it shall be extended to resolve the project configuration.

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
hasufell: https://github.com/haskell/cabal/pull/7783#issuecomment-1027334061

jneira: > Also i miss a simple positive test over imports and conditionals, to exercise and demosntrate the pr main goal.

oops, the same test is testing both things, in addition to cycles, sorry, i was assuming they were to going placed in two separate tests (not needed)

jneira: > [#7783 (comment)](https://github.com/haskell/cabal/pull/7783#issuecomment-1027334061)

afaics the test only checks a cabal.project cant import itself but no a cycle with more then one node (maybe it is not possible by design??)

jneira: just checked it locally and the cycle check works perfectly, but i can imagine a bug breaking that invariant while not allowing auto imports

jneira:
Just tested a possible corner case: conditions over the compiler, being the compiler one of the values you can change in the cabal.project itself

* Setting the compiler in a imported cabal.project inside a if throws an error: :+1:

```
Cannot set compiler in a conditional clause of a cabal project file
```

* conditions over compiler version honours the compiler set in the same cabal.project indpendently of its position :+1: 

Such corner cases could be new tests but they could be added after merging it

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
jneira: and a test checking remote imports would not be bad as well, it could be in other pr as well

jneira:
In windows a remote import like:

```
packages: .

import: https://raw.githubusercontent.com/haskell/cabal/master/cabal.project
```

is failing for me with

```
D:\ws\haskell\issues\importifs\dist-newstyle\src\htt2033.project: renameFile:renamePath:MoveFileEx "\\\\?\\D:\\ws\\haskell\\issues\\importifs\\dist-newstyle\\src\\htt2033.project" Just "\\\\?\\D:\\ws\\haskell\\issues\\importifs\\dist-newstyle\\src\\https:XXraw.githubusercontent.comXhaskellXcabalXmasterXcabal.project":invalid argument (El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.)
```

cabal-install compiled with ghc-8.6.5 and executed also with ghc-8.6.5
it was predicted by @hasufell [here ](https://github.com/haskell/cabal/pull/7783#discussion_r792163807) so i guess it is matter to apply the fix?

jneira:
> > One caveat i can think of it is you will not be able to parse cabal.project's with a simple grep.
> 
> I don't know of any such script. That's what https://hackage.haskell.org/package/cabal-install-parsers-0.4.5/docs/Cabal-Project.html is for and it shall be extended to resolve the project configuration.

"I've seen things you people wouldn't believe. Attack ships on fire off the shoulder of Orion", etc etc :-P
But it was an example, see how ifs in .cabal file made the duplicate module checks harder and error prone lately

jneira:
> Additional checks (not yet added) should prevent setting of compiler info within conditionals (since we can have conditionals depend on this info). Additionally, we should enforce that cabal.project.local files cannot have conditionals or imports, as we don't have to want to write machine generation to preserve that structure for now (necessary for cabal reconfigure).

so the first check is done but just tested the second one is not, conditions and imports are allowed in cabal.project.local

jneira:
And the last one, i promise, there is a minor caveat related with syntax error reporting. If you have a cabal.project:

```
packages: .
import: cabal.1.project
```

and a cabal.1.project

```



package: foo
```

the warning does not include the offending file but it does for the line number:

```
Warning: D:\ws\haskell\issues\importifs\cabal.project: Unrecognized field
'package' on line 4
```

cabal-install/src/Distribution/Client/CmdOutdated.hs
gbaz: a freeze file is an old style freeze file which is a "~/.cabal.config" file. a new freeze rile is really a project file. so to parse it we handle fully imports and conditionals as well now, which requires a bit more setup.

gbaz:
"so the first check is done but just tested the second one is not, conditions and imports are allowed in cabal.project.local"

I changed my mind on that. Instead we just error if attempting to update-reconfigure when there's conditionals. So they're allowed, but we just block the failing action: https://github.com/haskell/cabal/pull/7783/files#diff-50fda1f9b998a33220632ddccf47717d9c21b9992268697b40fdaf149bd2c700R139

cabal-install/src/Distribution/Client/CmdOutdated.hs
Mikolaj: Got it, thanks.

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
gbaz: i'm fine with a remote import test -- currently we have no tests relying on network connectivity so i didn't want to add one. but i think the cycle check test is fine as is -- indeed by design we check in a way that's robust to any cycle, so checking for 3-cycles vs 2-cycles isn't telling us anything special.

gbaz:
improved the error message a bit to list all files, and fixed the filename on windows thing (i think) and added one more test. I think this is ready for merge now and more tests can be added as desired?

jneira (approved):
> improved the error message a bit to list all files, and fixed the filename on windows thing (i think) and added one more test. I think this is ready for merge now and more tests can be added as desired?

thanks for take in account review comments
agree about add more tests later, will open an issue about if you dont mind

cabal-testsuite/PackageTests/ConditionalAndImport/cabal-cyclical.project
jneira: > i'm fine with a remote import test -- currently we have no tests relying on network connectivity so i didn't want to add one. 

i've just checked 

```
packages: .
import: file://D:/ws/haskell/cabal/cabal.project
```

works. which is really nice. It would test part of the code path and will not go out the file system, no?

changelog.d/pr-7783
```diff
@@ -0,0 +1,10 @@
+synopsis: Conditionals and imports in cabal.project files
+packages: cabal-install
+prs: #7783
+issues: #7556
+
```
fgaz, line 5: I'd say this is `significance: significant` :)

doc/cabal-project.rst
```diff
@@ -35,6 +35,32 @@ options):
...
+::
+
+    if(os(darwin))
+      optimization: False
+    elif(os(freebsd))
+      packages: freebsd/*.cabal
+    else
+      optimization: True
+
+    import: https://some.remote.source/subdir/cabal.config
```
fgaz, line 58: Maybe we should have a stackage example too?

bapcyk:
So what is the right way to set `with-compiler:` conditionally?

Kleidukos:
I personally would generate a local project file with this information.

PR #7802, opened 2021-11-07
Make solver able to reject build-type: Custom packages

phadej:
---

### DIFF
--- a/cabal-install-solver/src/Distribution/Solver/Modular/IndexConversion.hs
+++ b/cabal-install-solver/src/Distribution/Solver/Modular/IndexConversion.hs
@@ -344,6 +344,7 @@ convCondTree flags dr pkg os arch cinfo pn fds comp getInfo solveExes@(SolveExec
               ++ L.map (\e -> D.Simple (LDep dr (Ext  e)) comp) (allExtensions bi) -- unconditional extension dependencies
               ++ L.map (\l -> D.Simple (LDep dr (Lang l)) comp) (allLanguages  bi) -- unconditional language dependencies
               ++ L.map (\(PkgconfigDependency pkn vr) -> D.Simple (LDep dr (Pkg pkn vr)) comp) (pkgconfigDepends bi) -- unconditional pkg-config dependencies
+              ++ [D.Simple (LDep dr (BT (buildType pkg))) comp]
               ++ concatMap (convBranch flags dr pkg os arch cinfo pn fds comp getInfo solveExes) branches
               -- build-tools dependencies
               -- NB: Only include these dependencies if SolveExecutables
--- a/cabal-install-solver/src/Distribution/Solver/Types/PackageConstraint.hs
+++ b/cabal-install-solver/src/Distribution/Solver/Types/PackageConstraint.hs
@@ -141,8 +141,8 @@ showPackageConstraint pc@(PackageConstraint scope prop) =
 packageConstraintToDependency :: PackageConstraint -> Maybe PackageVersionConstraint
 packageConstraintToDependency (PackageConstraint scope prop) = toDep prop
   where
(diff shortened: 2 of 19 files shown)


grayjay:
Could you please add more background on this flag to the commit message?

phadej:
@grayjay will do once the idea is in general accepted (along as writing docs and the changelog).

gbaz:
@phadej to accept the idea in general would probably require a little motivation -- what's the plan for why we'd want this and how it would be used?

phadej:
`build-type: Custom` packages can run arbitrary code (technically `build-type: Configure` packages too). IMO it's fair to be able to ask solver to avoid them. Probably we'll need a more granular constraints, to allow *some* `build-type: Custom` packages, but this is a start.

gbaz:
Is there a concrete use case for this? Are there consumers of cabal-install we can imagine wanting to turn this on by default in certain of their build systems or the like, or is the motivation more "this should exist" and somewhat speculative? Like the code looks clean to me, and the approach reasonable. I'm just not clear on if there's a general direction of work this is considered a part of or etc?

phadej:
>  Are there consumers

Myself. E.g. for GHCJS (or crosscompiling if it worked).

gbaz:
Thanks, ghcjs was the sort of motivating use case that I was asking about! Being able to rule out plans ghcjs can't use makes perfect sense.

phadej:
GHCJS can use `build-type: Custom` packages (fortunately!) but things are smoother when these are avoided

gbaz (approved):

cabal-install-solver/src/Distribution/Solver/Modular/IndexConversion.hs
```diff
@@ -344,6 +344,7 @@ convCondTree flags dr pkg os arch cinfo pn fds comp getInfo solveExes@(SolveExec
               ++ L.map (\e -> D.Simple (LDep dr (Ext  e)) comp) (allExtensions bi) -- unconditional extension dependencies
               ++ L.map (\l -> D.Simple (LDep dr (Lang l)) comp) (allLanguages  bi) -- unconditional language dependencies
               ++ L.map (\(PkgconfigDependency pkn vr) -> D.Simple (LDep dr (Pkg pkn vr)) comp) (pkgconfigDepends bi) -- unconditional pkg-config dependencies
+              ++ [D.Simple (LDep dr (BT (buildType pkg))) comp]
```
grayjay, line 347: This line seems to add the build type constraint to every component and the branches of every conditional. Could the constraint be associated with the whole package instead, similar to UnsupportedSpecVer?

cabal-install-solver/src/Distribution/Solver/Types/PackageConstraint.hs
```diff
@@ -141,8 +141,8 @@ showPackageConstraint pc@(PackageConstraint scope prop) =
...
-    toDep (PackagePropertyVersion vr) = Just $ PackageVersionConstraint (scopeToPackageName scope) vr
-    toDep (PackagePropertyInstalled)  = Nothing
-    toDep (PackagePropertySource)     = Nothing
-    toDep (PackagePropertyFlags _)    = Nothing
-    toDep (PackagePropertyStanzas _)  = Nothing
+    toDep (PackagePropertyVersion vr)  = Just $ PackageVersionConstraint (scopeToPackageName scope) vr
+    toDep (PackagePropertyInstalled)   = Nothing
+    toDep (PackagePropertySource)      = Nothing
+    toDep (PackagePropertyFlags _)     = Nothing
+    toDep (PackagePropertyStanzas _)   = Nothing
```
grayjay, line 148: It looks like this file has only whitespace changes.

phadej:
Cleaning up old PRs.

PR #7805, opened 2021-11-08
Fix #7803: doc: replace broken link to --constraint doc by s.th. reasonable

### LINKED ISSUE #7803: Doc: link to "command line version of constraint field" leads nowhere reasonable
See
https://github.com/haskell/cabal/blob/5b4258cbfa536b6362c4e20578ced51676206ea8/doc/cabal-project.rst?plain=1#L325-L327
```rst
.. cfg-field:: constraints: constraints list (comma separated)
               --constraint="pkg >= 2.0", -c "pkg >= 2.0"
    :synopsis: Extra dependencies constraints.

    Add extra constraints to the version bounds, flag settings,
    and other properties a solver can pick for a
    package. 

    ...

    Valid constraints take the same form as for the `constraint
    command line option
    <installing-packages.html#cmdoption-setup-configure-constraint>`__.
```

The command option `--constraint` seems to be documented above, I suppose the link dates from more ancient version of the documentation!?

andreasabel:
Fix #7803: doc: 
- remove broken link to `--constraint` doc 
- instead, explain difference between `--constraint` and `constraint:`

### DIFF
--- a/doc/cabal-project.rst
+++ b/doc/cabal-project.rst
@@ -322,9 +322,9 @@ The following settings control the behavior of the dependency solver:
         constraints: bar == 2.1,
                      bar +foo -baz
 
-    Valid constraints take the same form as for the `constraint
-    command line option
-    <installing-packages.html#cmdoption-setup-configure-constraint>`__.
+    The option ``--constraint`` only takes a single constraint, not a
+    comma-separated list of constraints.  To make up for this, it can
+    appear several times on the command line.
 
 .. cfg-field:: preferences: preference (comma separated)
                --preference="pkg >= 2.0"


andreasabel:
Relevant CI passes successfully.

Mikolaj (approved):
LGTM

fgaz:
The linked section moved to https://cabal.readthedocs.io/en/3.6/setup-commands.html#cmdoption-runhaskell-Setup.hs-configure-constraint, we can link to that, and maybe also to https://cabal.readthedocs.io/en/3.6/cabal-package.html#pkg-field-build-depends

Mikolaj:
> The linked section moved to https://cabal.readthedocs.io/en/3.6/setup-commands.html#cmdoption-runhaskell-Setup.hs-configure-constraint. We can link to that, and maybe also to https://cabal.readthedocs.io/en/3.6/cabal-package.html#pkg-field-build-depends

@andreasabel: what do you think?

andreasabel:
> @andreasabel: what do you think?

@Mikolaj : I didn't find the time to continue on this PR yet, since I am teaching atm.  If someone else wants to finish it, I would not mind.  Otherwise, I probably find time over the holidays.

Mikolaj:
Oh, sure, there is no hurry. Have fun! :)

andreasabel:
@fgaz wrote:
> The linked section moved to [cabal.readthedocs.io/en/3.6/setup-commands.html#cmdoption-runhaskell-Setup.hs-configure-constraint](https://cabal.readthedocs.io/en/3.6/setup-commands.html#cmdoption-runhaskell-Setup.hs-configure-constraint). We can link to that, and maybe also to [cabal.readthedocs.io/en/3.6/cabal-package.html#pkg-field-build-depends](https://cabal.readthedocs.io/en/3.6/cabal-package.html#pkg-field-build-depends)

It seems that the field description in `cabal-project.rst` _also_ gives the command line option syntax.  So, apparently, someone decided that the documentation for a feature should reside with the field description, and then other occurrences of the same feature (as in command line options) can link there.  Or am I mistaken?

I think it is not good to duplicate documentation for the same reason as it is not good to clone code.
We should decide on one of these and document our design decision in the developer manual:
1. The documentation for a feature resides with the field description of the cabal.project file.
2. The documentation for a feature resides with the command line option.
3. The features are described in a separate place and both of the above locations link there.

Which of these, I do not care (also, there might have been a decision in the past that was only partially implemented).  Option 3. might need some restructuring of the docs.

Attn: @Mikolaj

Mikolaj:
Fully agreed and I think we should choose 1 or 2. I'm not aware of a previous choice, but we should ask, e.g., on the meeting this Thursday. 

How often there is 1 but not 2, or the reverse? If there is a significant imbalance, I guess we might just choose the one that covers more of the options.

andreasabel:
The new PR #7902 fixes the OP without changing the text.

Issue #7798, opened 2021-11-04
Overhaul CI to use ghcup and not a mixture of obsolete sources of GHC, etc.


Mikolaj:
We really need a proper overhaul moving from a mixture of containers, Ubuntu packages, curl, chocolatey and downloads of random old snapshots of packages to ghcup. E.g., we are stuck at 8.10.4 (or even much earlier, e.g., 8.6.5) due to that.

Related: #7747, #7742, #7708, #7655, #7315 and many others.

phadej:
FWIW, I'd really appreciate if Cabal kept testing that it can control older GHCs: Currently we have

```
                    , GhcJob "8.8.4"  True  "--lib-only"                True  ["8.8.4"] $
                        libSteps ++
                        [ "lib-suite-extras --extra-hc /opt/ghc/7.0.4/bin/ghc-7.0.4"
                        , "lib-suite-extras --extra-hc /opt/ghc/7.2.2/bin/ghc-7.2.2"
                        , "lib-suite-extras --extra-hc /opt/ghc/7.4.2/bin/ghc-7.4.2"
                        ]
```

job, which tests that `cabal-install` & `Cabal` can use these old GHCs as compilers.

This is different then being buildable with them, that's not that important and there the 5/3 year policy is very conservative already.

jneira:
My point of view about ci:

1. We should put as much logic as possible in the test suite itself: it is haskell code and it would help to migrate ci between providers (and cabal was moved several times) easier and use several providers at same time (as we are doing with github/gitlab)
2. I would try to minimize shell and other langs like python scripts, they are not always os independent and add complexity and cognitive load as you have to learn and maintain them
3. I would not use templates even written in haskell unless the complexity they add reduce data duplication and maintance cost dramatically (it may be the case with cabal, dont know). ci providers has their own ways to reduce duplication for things which cant be put in the test suite itself. Templates makes ci providers features unusable (if some day dependabot could make automatic prs to update haskell deps we will not able to use it for example, see #7809)

jneira:
I would say we should use haskell/action/setup, abstracting away the way we get ghc's, even abstracting over ghcup, and try to make possible download older ghcs there (it it is not already possible). It would help all the haskell community

Mikolaj:
I'm a fan of baby steps, but I'm not averse to dropping existing templates one by one, once we have a replacement and we've thought through the cost of each such switch.

[...]

PR #7809, opened 2021-11-10
add dependabot: add Actions CI updates merge requests automation

Anton-Latukha:
Dependabot would weekly check updates for project's [.github/workflows/workflow.yml](https://github.com/haskell/actions/blob/main/.github/workflows/workflow.yml) & send information & patch updates ([like these](https://github.com/haskell-nix/hnix-store/pull/169)).

PR to CI configuration run CI with changes applied, which is a self-test.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+
+  # NOTE: Dependabot official configuration documentation:
+  # https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem
+
+  # Maintain dependencies for internal GitHub Actions CI for pull requests
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


Mikolaj:
@Anton-Latukha: hi! Do you think using https://github.com/haskell/actions would aid us in solving #7798? Is that preferable to ghcup or does it complement it?

jneira:
> 
> 
> @Anton-Latukha: hi! Do you think using https://github.com/haskell/actions would aid us in solving #7798? Is that preferable to ghcup or does it complement it?

They are complementary, for example it uses chocolatey for win and ghcup for the rest automatically so you dont have to. When the action will switch to use ghcup for al os's it will be transparent for us.

Mikolaj:
Great, let's use it. Whom does the dependabot send info to? Is it via email?

jneira:
> 
> 
> Great, let's use it. Whom does the dependabot send info to? Is it via email?

I think security info about deprecated actions will be here: https://github.com/haskell/cabal/security/dependabot
Like the recent ones about python deps when we changed the doc generation 

And it will open auto pr's to update dependencies

fgaz:
> And it will open auto pr's to update dependencies

our workflow files are generated, so prs would be useless (but issues wouldn't)

Anton-Latukha:
> > And it will open auto pr's to update dependencies
> 
> our workflow files are generated, so prs would be useless (but issues wouldn't)

Ok. Haskell-CI is both blessing & a curse.

I am "on the way" & think contributing some GitHub automation into it (I want to merge there HLint automation). But first I need to arrange stuff for that in `actions`.

---

> @Anton-Latukha: hi! Do you think using https://github.com/haskell/actions would aid us in solving #7798? Is that preferable to ghcup or does it complement it?

I do not have access to `actions` repository & it seems without a maintainer, lives on drive-by maintenance. I can adopt it if needed.

`actions/setup` - yes - mainly uses `ghcup`:
https://github.com/haskell/actions/blob/f0ecab9af1c178879d2243035bba5832e5529b43/setup/src/installer.ts#L148-L166

The solution to the question of #7798 - is to make `ghcup` being able to provide (even if hidden from CLI & only for internal users) - support for old versions. That requires to make code duplication there to keep legacy compatibility & more hands than one person.

I'm thinking about helping in `ghcup` more. Besides Haskell - I love POSIX shell programming. After a while & would look at `ghcup` internals & remember what was discussed here. Maybe, as I see Ospald knows `shell` better - we should be more compatible now. Now style in `ghcup` strangely reassembles style I adopted.

Because Cabal CI as said - is autogenerated - yes - do not see a reason to keep PR around, if only to use dependabot PRs as notifications.

Anton-Latukha:
Looked into CI.

Thought through the situation from different angles.

The situation is obviously complex.

I can not recommend anything. Especially because almost nothing in workflows can be done, deshotgun-surgered or shelled-out, because they are autogenerated from [GenValidate.hs](https://github.com/haskell/cabal/blob/master/cabal-dev-scripts/src/GenValidate.hs). & that is made by `phadej` pretty recently & who knew what & how to do - better than him. If there was a place to leverage `haskell-ci` - he'd probably used it.

& I do not know what `Zinza` is [& Hoogle actually returns `null`](https://hoogle.haskell.org/?hoogle=zinza) (for me) but it is on hackage [Hackage](https://hackage.haskell.org/package/zinza) - I notified `phadej` on that occasion with his package.

In time I currently can only work on the fringes here and there taking responsibilities to ease the situation (like maybe helping-out in `ghcup` & `haskell-ci`. I already have a number of promises maid which I should deliver.

andreabedini:
@Anton-Latukha the CI has been rehauled in #7952. Do you think we could reconsider this analysis?

Issue #371, opened 2012-05-24
Error in Data.IntSet documentation


bos:
(Imported from [Trac #378](http://hackage.haskell.org/trac/hackage/ticket/378), reported by RayNbow on 2008-10-21)

[http://hackage.haskell.org/packages/archive/containers/0.1.0.1/doc/html/Data-IntSet.html#v%3Asplit](http://hackage.haskell.org/packages/archive/containers/0.1.0.1/doc/html/Data-IntSet.html#v%3Asplit)

<tt> split 3 (fromList [1..5]) == (fromList [1,2], fromList [3,4]) </tt>

should be:

<tt> split 3 (fromList [1..5]) == (fromList [1,2], fromList [4,5]) </tt>

PR #7822, opened 2021-11-17
Enable TypeOperators to use (~)

int-index:
[GHC Proposal #371](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0371-non-magical-eq.md) requires `TypeOperators` to use type equality `a~b`. The following lines are affected:

```
Cabal/src/Distribution/Compat/Graph.hs
181:instance (IsNode a, IsNode b, Key a ~ Key b) => IsNode (Either a b) where

Cabal/src/Distribution/Utils/IOData.hs
61:instance a ~ Char => KnownIODataMode [a] where
```

The fix is to simply enable the extension.


Mikolaj (approved):
Thank you. Accepted as soon as CI passes.

fgaz (approved):

PR #7823, opened 2021-11-18
Warn about needing to opt in to overwrite-policy for install.

### LINKED ISSUE #7275: overwrite-policy is not documented
While I did find a mention of `install-method` in the docs, I did not find a mention of `overwrite-policy`, unless the search at 
https://cabal.readthedocs.io/en/3.4/search.html?q=%22overwrite-policy%22&check_keywords=yes&area=default
is failing me.

Given that every user who wants to re-install an executable with `cabal install` will have to think about this policy, this is probably worth documenting.

philderbeast:
Doc change only, added to the existing warning block for `cabal install`, fixes #7275.

### DIFF
--- a/doc/cabal-commands.rst
+++ b/doc/cabal-commands.rst
@@ -298,7 +298,8 @@ symlinks/copies their executables in ``installdir`` (usually ``~/.cabal/bin``).
 .. warning::
 
   If not every package has an executable to install, use ``all:exes`` rather
-  than ``all`` as the target.
+  than ``all`` as the target. To overwrite an installation, use
+  ``--overwrite-policy=always`` as the default policy is ``never``.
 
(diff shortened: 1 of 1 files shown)


Mikolaj (approved):
LGTM

Mikolaj:
BTW, if you said in commit message "fixes #7275", it would close the issue automatically: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

jneira (approved):
lgtm, thanks for updating docs

fgaz:
> BTW, if you said in commit message "fixes #7275", it would close the issue automatically: 

it also works in the pr description, but it has to be on its own line

Issue #5079, opened 2018-01-30
new-test should imply --enable-tests


dmwit:
I observe the following behavior. Summary: `cabal new-test` fails to resolve dependencies, while `cabal new-test --enable-tests` succeeds and runs the tests. I think they should both succeed and run the tests.

    % cabal new-test
    cabal: Cannot test the package bad-tests-0.1.0.0 because none of the
    components are available to build: the test suite 'bad-tests' is not available
    because the solver did not find a plan that included the test suites
    
    zsh: exit 1     cabal new-test
    % cabal new-test --enable-tests
    Resolving dependencies...
    Build profile: -w ghc-8.2.1 -O1
    In order, the following will be built (use -v for more details):
     - bad-tests-0.1.0.0 (first run)
    Preprocessing test suite 'bad-tests' for bad-tests-0.1.0.0..
    Building test suite 'bad-tests' for bad-tests-0.1.0.0..
    Running 1 test suites...
    Test suite bad-tests: RUNNING...
    Test suite bad-tests: PASS
    Test suite logged to:
    <redacted>
    1 of 1 test suites (1 of 1 test cases) passed.

The project itself is quite minimal. Here's `bad-tests.cabal`:

    name:                bad-tests
    version:             0.1.0.0
    build-type:          Simple
    cabal-version:       >=1.2
    
    test-suite bad-tests
      type: exitcode-stdio-1.0
      main-is: Main.hs
      build-depends: base, tasty >= 0.10 && < 0.12

`Main.hs` has just `main = return ()`, and `Setup.hs` is the standard `import Distribution.Simple; main = defaultMain`.

hvr:
Iirc this was a deliberate design choice by @dcoutts to avoid invocations of `cabal new-test` w/o any flags to suddenly force recomputing a new install-plan, only to have to recompute yet another one when we invoke `cabal new-build` or `cabal new-bench` (which btw suffers from the same symptom). The current behaviour causes the install-plans to remain stable when alternativing between invocations of `build/bench/test`.

dmwit:
Given that constraint (which I'm not sure I agree with), I suspect the correct behavior is to default to creating a build plan that includes the tests and benchmarks (even in new-build), then prune those away along with any dependencies that are in the plan exclusively for them if the user hasn't asked for them. And of course one could fall back to the current way of creating an install plan when --disable-tests/--disable-benchmarks is supplied.

A stable build plan is nice. Having a build plan at all in the first place is even nicer.

[...]

PR #7829, opened 2021-11-21
[WIP] Only buildable tests

gelisam:
[As requested](https://github.com/haskell/cabal/issues/5079#issuecomment-974270197), I am trying to implement an `--only-buildable-tests` flag whose behaviour matches the default behaviour when neither `--enable-tests` nor `--disable-tests` is specified.

This PR is a WIP, as I am still at the beginning of my investigation. So far, I have only added a test which exercises the current behaviour, and I have slightly tweaked how `--enable-tests` and `--disable-tests` are parsed (which slightly affects the `--help` output). Still, my impression is that cabal is a complex and fragile beast, so I would like to make sure that this small change hasn't already broken something. I am thus following [the recommendation](https://github.com/haskell/cabal/blob/cff9b1a7c6188796873ffcb7ba9332dca718548f/CONTRIBUTING.md#running-tests) and opening a WIP PR to get the tests to run.

---

### DIFF
--- a/Cabal-QuickCheck/src/Test/QuickCheck/Instances/Cabal.hs
+++ b/Cabal-QuickCheck/src/Test/QuickCheck/Instances/Cabal.hs
@@ -29,6 +29,7 @@ import Distribution.Simple.Setup                   (HaddockTarget (..), TestShow
 import Distribution.SPDX
 import Distribution.System
 import Distribution.Types.Dependency
+import Distribution.Types.EnableComponentType
 import Distribution.Types.Flag                     (FlagAssignment, FlagName, mkFlagAssignment, mkFlagName, unFlagAssignment)
 import Distribution.Types.IncludeRenaming
 import Distribution.Types.LibraryName
@@ -491,6 +492,13 @@ instance Arbitrary PackageDB where
 -------------------------------------------------------------------------------
 
 instance Arbitrary DumpBuildInfo where
+    arbitrary = arbitraryBoundedEnum
+
+-------------------------------------------------------------------------------
+-- EnableComponentType
+-------------------------------------------------------------------------------
+
+instance Arbitrary EnableComponentType where
     arbitrary = arbitraryBoundedEnum
 
 -------------------------------------------------------------------------------
--- a/Cabal/Cabal.cabal
+++ b/Cabal/Cabal.cabal
@@ -255,6 +255,7 @@ library
     Distribution.Types.VersionInterval.Legacy
     Distribution.Types.GivenComponent
     Distribution.Types.PackageVersionConstraint
+    Distribution.Types.EnableComponentType
     Distribution.Utils.Generic
     Distribution.Utils.Json
     Distribution.Utils.NubList
--- a/Cabal/src/Distribution/Simple/Configure.hs
+++ b/Cabal/src/Distribution/Simple/Configure.hs
@@ -80,6 +80,7 @@ import Distribution.Types.PackageVersionConstraint
 import Distribution.Types.LocalBuildInfo
 import Distribution.Types.ComponentRequestedSpec
 import Distribution.Types.GivenComponent
+import Distribution.Types.EnableComponentType
 import Distribution.Simple.Utils
 import Distribution.System
 import Distribution.Version
@@ -401,12 +402,15 @@ configure (pkg_descr0, pbi) cfg = do
                                 -- nomenclature; it's just a request; a
                                 -- @buildable: False@ might make it
                                 -- not possible to enable.
-                                { testsRequested = fromFlag (configTests cfg)
+                                { testsRequested =
+                                  fromFlag (configTests cfg) == EnableAll
                                 , benchmarksRequested =
-                                  fromFlag (configBenchmarks cfg) }
+                                  fromFlag (configBenchmarks cfg) == EnableAll
+                                }
     -- Some sanity checks related to enabling components.
     when (isJust mb_cname
-          && (fromFlag (configTests cfg) || fromFlag (configBenchmarks cfg))) $
+          && ( fromFlag (configTests cfg) == EnableAll
+            || fromFlag (configBenchmarks cfg) == EnableAll)) $
         die' verbosity $
               "--enable-tests/--enable-benchmarks are incompatible with" ++
               " explicitly specifying a component to configure."
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -701,10 +701,10 @@ configureOptions showOrParseArgs =
       ,multiOption "tests"
          configTests (\v flags -> flags { configTests = v })
          [noArg (Flag True) []
-                ["enable-test","enable-tests"]
+                ["enable-tests", "enable-test"]
                 "Build all the test suites listed in the package description file."
          ,noArg (Flag False) []
-                ["disable-test","disable-tests"]
+                ["disable-tests", "disable-test"]
                 "Do not build any test suites."
          ]
 
@@ -727,10 +727,10 @@ configureOptions showOrParseArgs =
       ,multiOption "benchmarks"
          configBenchmarks (\v flags -> flags { configBenchmarks = v })
          [noArg (Flag True) []
-                ["enable-benchmark","enable-benchmarks"]
+                ["enable-benchmarks", "enable-benchmark"]
                 "Build all the benchmarks listed in the package description file."
          ,noArg (Flag False) []
-                ["disable-benchmark","disable-benchmarks"]
+                ["disable-benchmarks", "disable-benchmark"]
                 "Do not build any benchmarks."
          ]
 
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -701,7 +701,7 @@ configureOptions showOrParseArgs =
       ,multiOption "tests"
          configTests (\v flags -> flags { configTests = v })
          [noArg NoFlag []
-                ["only-buildable-tests", "only-buildable-test"]
+                ["enable-tests-when-possible", "enable-test-if-possible"]
                 "Build the tests if a build plan can be found, don't build them otherwise. The decision is made independently for each package, not for each test suite."
          ,noArg (Flag True) []
                 ["enable-tests", "enable-test"]
@@ -730,7 +730,7 @@ configureOptions showOrParseArgs =
       ,multiOption "benchmarks"
          configBenchmarks (\v flags -> flags { configBenchmarks = v })
          [noArg NoFlag []
-                ["only-buildable-benchmarks", "only-buildable-benchmark"]
+                ["enable-benchmarks-when-possible", "enable-benchmark-if-possible"]
                 "Build the benchmarks if a build plan can be found, don't build them otherwise. The decision is made independently for each package, not for each benchmark."
          ,noArg (Flag True) []
                 ["enable-benchmarks", "enable-benchmark"]
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -698,10 +698,15 @@ configureOptions showOrParseArgs =
             (parsecToReadE ("Cannot parse module substitution: " ++) (fmap (:[]) parsecModSubstEntry))
             (map (Disp.renderStyle defaultStyle . dispModSubstEntry)))
 
-      ,option "" ["tests"]
-         "dependency checking and compilation for test suites listed in the package description file."
+      ,multiOption "tests"
          configTests (\v flags -> flags { configTests = v })
-         (boolOpt [] [])
+         [noArg (Flag True) []
+                ["enable-test","enable-tests"]
+                "Build all the test suites listed in the package description file."
+         ,noArg (Flag False) []
+                ["disable-test","disable-tests"]
+                "Do not build any test suites."
+         ]
 
       ,option "" ["coverage"]
          "build package with Haskell Program Coverage. (GHC only)"
@@ -719,10 +724,15 @@ configureOptions showOrParseArgs =
          (\v flags -> flags { configExactConfiguration = v })
          trueArg
 
-      ,option "" ["benchmarks"]
-         "dependency checking and compilation for benchmarks listed in the package description file."
+      ,multiOption "benchmarks"
          configBenchmarks (\v flags -> flags { configBenchmarks = v })
-         (boolOpt [] [])
+         [noArg (Flag True) []
+                ["enable-benchmark","enable-benchmarks"]
+                "Build all the benchmarks listed in the package description file."
+         ,noArg (Flag False) []
+                ["disable-benchmark","disable-benchmarks"]
+                "Do not build any benchmarks."
+         ]
 
       ,option "" ["relocatable"]
          "building a package that is relocatable. (GHC only)"
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -100,6 +100,7 @@ import Distribution.Verbosity
 import Distribution.Utils.NubList
 import Distribution.Types.ComponentId
 import Distribution.Types.DumpBuildInfo
+import Distribution.Types.EnableComponentType
 import Distribution.Types.GivenComponent
 import Distribution.Types.Module
 import Distribution.Types.PackageVersionConstraint
@@ -264,8 +265,8 @@ data ConfigFlags = ConfigFlags {
       -- package does not use Backpack, or we just want to typecheck
       -- the indefinite package.
     configConfigurationsFlags :: FlagAssignment,
-    configTests               :: Flag Bool, -- ^Enable test suite compilation
-    configBenchmarks          :: Flag Bool, -- ^Enable benchmark compilation
+    configTests               :: Flag EnableComponentType, -- ^Enable test suite compilation
+    configBenchmarks          :: Flag EnableComponentType, -- ^Enable benchmark compilation
     configCoverage :: Flag Bool, -- ^Enable program coverage
     configLibCoverage :: Flag Bool, -- ^Enable program coverage (deprecated)
     configExactConfiguration  :: Flag Bool,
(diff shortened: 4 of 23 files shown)


gelisam:
As I had feared, this seemingly no-op change did manage to break several tests.

First, `UnitTests.Distribution.Client.Configure` has a few broken tests which say that `True` cannot be parsed, presumably because a `boolOpt` has an obvious way to parse boolean values but a multiOption needs more effort. I'm thinking of using `EnableAll`, `DisableAll`, and `OnlyBuildable`, with `True` and `False` being synonyms for the first two, for backwards compatibility.

gelisam:
Second, `UnitTests.Distribution.Client.ProjectConfig`'s `ProjectConfig printing/parsing round trip` test fails because `Flag False` round trips to `NoFlag`. SEND Seems related, but why is the parsing succeeding with `NoFlag` this time instead of failing with a parse error? Maybe unparseable fields are ignored, and so we are left with the default value `NoFlag`?

Mikolaj:
Let me clarify for people viewing this PR, the idea is to make `--enable-tests` the default, but then we need a flag that corresponds to the current behaviour (`--only-buildable-tests`), so that the rare cases failing with `--enable-tests` can be handled by `--only-buildable-tests` and `tests: True` in cabal.project*. Did I get that right?

No idea about the failing tests. Feel free to ignore and/or redo them until we get reviewers that know what's going on.

gelisam:
> the idea is to make `--enable-tests` the default, but then we need a flag that corresponds to the current behaviour (`--only-buildable-tests`), so that the rare cases failing with `--enable-tests` can be handled by `--only-buildable-tests`

I was not planning to change the default, but sure, that sounds like a good default!

> by `--only-buildable-tests` and `tests: True` in `cabal.project*`.

Currently, `--enable-tests` is the same as `tests: True`, `--disable-tests` is the same as `tests: False`, and not specifying either flag is the same as not specifying a `tests` setting. I was thus planning to continue this 1-to-1 equivalence, by adding an `--only-buildable-tests` which would be equivalent to specifying `tests: OnlyBuildable`.

I don't quite understand the semantics of what you are proposing instead. Would there be an `only-buildable-tests: True` setting as well, so that the current behavior can be obtained either by specifying both `--enable-tests` and `--only-buildable-tests`, or both `tests: True` and `only-buildable-tests: True`, or an equivalent combination? What would the `tests: False; only-buildable-tests: True` combination mean?

What I had in mind was that `tests: True` with

No idea about the failing tests. Feel free to ignore and/or redo them until we get reviewers that know what's going on.

gelisam:
> No idea about the failing tests. Feel free to ignore and/or redo them until we get reviewers that know what's going on.

I'm not blocked, just thinking out loud. I'm working on fixing the tests.

Mikolaj:
> Currently, `--enable-tests` is the same as `tests: True`, `--disable-tests` is the same as `tests: False`, and not specifying either flag is the same as not specifying a `tests` setting. I was thus planning to continue this 1-to-1 equivalence, by adding an `--only-buildable-tests` which would be equivalent to specifying `tests: OnlyBuildable`.

Oh, sure, 1-1 is actually better. I didn't know this can be that simple. Please ignore my overcomplicated version.

BTW, it would be great to add tests of cabal invocations that succeed with `--only-buildable-tests` and then run the same tests with each of the old flags and see them fail, just to demonstrate that the new flag is not useless.

gelisam:
> BTW, it would be great to add tests of cabal invocations that succeed with `--only-buildable-tests` and then run the same tests with each of the old flags and see them fail, just to demonstrate that the new flag is not useless.

Isn't that what the test I've already added already does?

Mikolaj:
I think my confusion stems from this comment: https://github.com/haskell/cabal/issues/5079#issuecomment-362062727

I think the third option that @phadej asks for is not just simply the current default behaviour (which is to solve for no tests at all) but it's a behaviour that would be best to have as the default for `cabal test` in the future, namely to solve for as much tests as possible, but don't insist (and sometimes fail) to build all (and `--enable-tests` does). Do I get this right this time?

gelisam:
> the third option that @phadej asks for is not just simply the current default behaviour (which is to solve for no tests at all)

Okay, that really is the crux of the confusion then. I do think that the third option is the same as the current default behavior, and I do think that the current default behavior is to try to find a solution which includes tests, and if that can't be done, to look for a solution which does not include the tests. In fact, that's what my test demonstrates: with no arguments, `cabal test` successfully solves for, builds, and runs `buildable-test`, and it does not run nor build `unbuildable-test`, because it cannot find a build plan which includes it.

Meanwhile, `cabal --enable-tests` fails because we are asking to run all the tests but a plan which includes both tests cannot be found. And `cabal --disable-tests` also fails, because the build plan includes zero tests, and because `cabal test` chooses to error out when it is suspiciously asked to run zero tests.

Mikolaj:
I'm afraid @andreasabel and other people in that ticket report otherwise: https://github.com/haskell/cabal/issues/5079#issuecomment-974173311

Perhaps that's because you report what Cabal's Setup interface does, while we mean `cabal v2-test`? They do a different thing, IIRC.

Mikolaj:
Hmm, but your test indeed behaves as you claim, which is completely contrary to my experience.

Mikolaj:
I now think, probably all people surprised by the `cabal test` behaviour had `tests: False` specified somewhere. Indeed, if we can change the default in `cabal.project` and whenever else to `tests: Buildable` or whatever it's called, that can be the best of both worlds. Would that solve for tests, but not build not run them?

gelisam:
> Would that new flag, when used in cabal build solve for tests, but not build them?

My understanding is that "the new flag" (that is, the current default behavior) causes the solver to pick a plan which may include some tests and exclude others. Which of the included targets are built and/or ran depends on the command:

* `cabal build` defaults to `cabal build all`, which means "all libs and all executables", so `cabal build` does not build the tests.
* `cabal build *:test:*` builds all the tests, but it does not run them. Libraries and executables are not built either unless they are dependencies of those tests.
* `cabal test` runs all the included tests, and thus also builds them first.

Mikolaj:
This is perfection. I didn't even know this is possible.

gelisam:
> Oh, and the hint that is displayed when `cabal test` fails should suggest `tests: OnlyBuildable` instead of the current `tests: True`.

Are you sure? If people set `tests: OnlyBuildable` and their test is not buildable, the build plan will not include the non-buildable test. When they run `cabal test`, there will be zero tests to run, and so they will still get that hint. The hint will be misleading, because they already have `tests: OnlyBuildable`. In that case though, `cabal test --enable-tests` will not succeed either, because the test is not buildable.

Thinking about this further, I suspect that even in the case in which people do not explicitly set `tests: False`, it might still be possible to get the situation in which `cabal test` fails but `cabal test --enable-tests` succeeds. I think Cabal's algorithm is more complex than depth-first search, in which case maybe Cabal can end up picking a build plan which does not include a test even though there exists one. Perhaps it's more like breadth-first search, and a build plan A which does not include the tests happen to be encountered earlier than a plan B which does include them, and so `cabal test` fails because plan A does not include the tests, while `cabal test --enable-tests` succeeds because plan A is not available and so Cabal is forced to continue searching until it finds plan B.

> Generally, we'll need to do `git grep -i "tests:"`.

I don't follow this bit. Surely cabal should not assume that it is ran from inside a git repo? And why are we grepping for `tests:`? cabal knows whether `tests` is set or not.

Mikolaj:
> > Oh, and the hint that is displayed when `cabal test` fails should suggest `tests: OnlyBuildable` instead of the current `tests: True`.
> 
> Are you sure? If people set `tests: OnlyBuildable` and their test is not buildable, the build plan will not include the non-buildable test. When they run `cabal test`, there will be zero tests to run, and so they will still get that hint. The hint will be misleading, because they already have `tests: OnlyBuildable`. In that case though, `cabal test --enable-tests` will not succeed either, because the test is not buildable.

And `cabal test --enable-tests` is equivalent to `tests: True`? If that's the case, then the currently present hint, advising to add `tests: True` already has this problem, so we are not making things worse, only better (people using the new hint won't have their tests built by `cabal build` and so won't be tempted to set `tests: False`). 

> Thinking about this further, I suspect that even in the case in which people do not explicitly set `tests: False`, it might still be possible to get the situation in which `cabal test` fails but `cabal test --enable-tests` succeeds. I think Cabal's algorithm is more complex than depth-first search, in which case maybe Cabal can end up picking a build plan which does not include a test even though there exists one. Perhaps it's more like breadth-first search, and a build plan A which does not include the tests happen to be encountered earlier than a plan B which does include them, and so `cabal test` fails because plan A does not include the tests, while `cabal test --enable-tests` succeeds because plan A is not available and so Cabal is forced to continue searching until it finds plan B.

Oh, that's a shame. If that's true, indeed we can't modify the hint.

> > Generally, we'll need to do `git grep -i "tests:"`.
> 
> I don't follow this bit. Surely cabal should not assume that it is ran from inside a git repo? And why are we grepping for `tests:`? cabal knows whether `tests` is set or not.

I mean, we as developers changing that code, not 'we' as 'cabal when it's running'. I was speaking in the context of `tests: True` either advised or automatically added, so this grep would let us (the developers) see this in our repo and correct.

gelisam:
Should we maybe schedule a call? I feel like we're really struggling to communicate in this thread.

> And `cabal test --enable-tests` is equivalent to `tests: True`?

Yes.

> then the currently present hint, advising to add `tests: True` already has this problem

Depends what you mean by "this problem"! In the case in which a test is not buildable, the current hint does not have the problem that it is misleading, because setting `tests: True` does allow the user to make forward progress. Without a `tests` entry, their tests are mysteriously excluded from the build plan and they cannot even try to build it. With `tests: True`, the user now sees the source of the error: `cabal test` says it cannot find a build plan.

The current hint does have the problem that its suggestion is insufficient to completely solve the problem, as running `cabal test` still fails after setting `tests: True`, it merely fails with a different error message. Perhaps the hint should instead say "set `tests: True` _and fix the version bounds so that a build plan which includes the tests can be found_"?

Mikolaj:
> Should we maybe schedule a call? I feel like we're really struggling to communicate in this thread.

With pleasure. However, this is more a sequential strings of my misunderstandings than a block that can be dealt with in one go. And each past case is clarified as far as I'm concerned.

> The current hint does have the problem that its suggestion is insufficient to completely solve the problem, as running `cabal test` still fails after setting `tests: True`, it merely fails with a different error message. Perhaps the hint should instead say "set `tests: True` _and fix the version bounds so that a build plan which includes the tests can be found_"?

Yes, agreed, this also makes it more likely the user will looks closer at the error message with `tests: True` instead of assuming it's as unhelpful as without. It wouldn't also hurt to prefix with "To debug this problem, set ..." to make sure `tests: True` is not considered a permanent solution (and, after a period of suffering from longer builds, switched to equally wrong `tests: False`).

gelisam:
All right, back to thinking out loud about the failing tests.

There seems to be two related problems. The first problem is that the setting `packageConfigTests = Flag True` is now pretty-printed as `"tests: enable-tests"` instead of `"tests: True"`. The parser expects `enable-tests`, and thus barfs on `True`.

The second problem is that the setting `packageConfigTests = Flag False` is now pretty-printed as `""`, presumably because something thinks `False` is the default, but `""` gets parsed to `NoFlag`, not `Flag False`, thus causing the round-trip to fail.

gelisam:
I explored an alley that turned out to be a dead end.

It is possible to give multiple synonyms for a command-line flag. I am already using this feature to allow `--enable-test` as a synonym for `--enable-tests`. I thus tried to add `--True` as a silly synonym for `--enable-tests`, so that `tests: enable-tests`, `tests: enable-test`, and `tests: True` can all be accepted in the configuration file. Unfortunately, that doesn't work, as the first entry in the synonym list is treated specially. First, it is the entry which is used in the `--help` output, so I should definitely put `enable-tests` first. Second, it is the _only_ entry which is allowed in configuration files. So in order to allow `tests: True`, I'd have to list `True` first, which would make `--True` appear in the `--help` output instead of `--enable-tests`, which would be silly.

gelisam:
I wrote a custom parser for the `tests` and `benchmarks` settings, so I can accept both `tests: True` and `tests: EnableAll`. Unfortunately, the other problem remains: `False` is still being pretty-printed as `""`, which in turns gets parsed to `NoFlag` instead of `Flag False`.

gelisam:
Very strange, it's the parse error on `True` which remains, while the roundtrip error on `False` has disappeared! I expected the opposite, since I only fixed the parser.

gelisam:
Ah, I know why the roundtrip error on `False` disappeared! I didn't just write a customized parser, I also wrote a customized pretty-printer. The bug was in the pretty-printer: it was pretty-printing `Flag False` to `""` instead of to `"tests: False"`. The old and new parsers both parse `""` to `NoFlag`, and that's the correct behavior.

gelisam:
Ah, I know why the roundtrip error on `False` disappeared! I didn't just write a customized parser, I also wrote a customized pretty-printer. The bug was in the pretty-printer: it was pretty-printing `Flag False` to `""` instead of to `"tests: False"`. The old and new parsers both parse `""` to `NoFlag`, and that's the correct behavior.

gelisam:
And the reason why the parse error on `True` still happens is because I've only tweaked the parser used by `parseLegacyProjectConfig`, but I haven't yet tweaked the parser used by `parseConfig`. That's because I didn't think there was also a non-legacy version! I was misled by the following comment:

https://github.com/haskell/cabal/blob/master/cabal-install/src/Distribution/Client/ProjectConfig/Legacy.hs#L106-L113

That comment is simply saying that the parser for `tests: True` is derived from the parser for `--enable-tests`. The comment is apologetic, but to avoid duplication, I think that's a good idea.

What misled me is that the comment is right above the `LegacyProjectConfig` datatype definition, so I thought the comment was explaining why `LegacyProjectConfig` was needed. I thought it said that `LegacyProjectConfig` has parsers but `ProjectConfig` doesn't, which is why we still use `LegacyProjectConfig`. This is not the case: `parseLegacyProjectConfig` and `parseConfig` both derive their parsers from the same list of command-line options, and then perform a different set of tweaks on top of that (which is why I now need to make my tweak a second time).

Given this, would it make sense to move that comment somewhere else, perhaps closer to `parseConfig`? Then `parseLegacyProjectConfig` could have a shorter comment saying it uses the same reuse strategy as `parseConfig`, but has its own distinct set of tweaks.

Mikolaj:
Definitely, please amend and extend the comments as needed to prevent the misunderstanding it caused you.

gelisam:
All right, I have tweaked the comments (please let me know if you agree it is an improvement!) and I have added the missing parser. I hope the tests are passing this time!

gelisam:
They do pass, hurray! Now I can move on to actual feature: adding an `--only-buildable-tests` flag.

Now that I am familiar with the way in which flags and options work, I know that the easiest way to implement this would be to parse `--only-buildable-tests` as `NoFlag`. However, it would probably be clearer and more useful if I instead defined a custom sum type, `data EnableStanza = OnlyBuildable | EnableAll | DisableAll`, right?

gelisam:
I keep calling them "stanza", but it should be "component type". If a package provides a library and two tests, that's three stanzas (or three components): the library, the first test, and the second test. But it's only two component types: libraries and tests. Each of the `tests` and `benchmarks` setting is about a particular component type, not about a particular stanza/component.

gelisam:
Darn, "OnlyBuildable" might also be an incorrect name! I found this comment which gives a different definition for "buildable" in Cabal:

https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/Types/ComponentRequestedSpec.hs#L33-L37

Whereas what we mean is whether a build plan which includes that component type can be found. "OnlyPlannable"? "IfPossible"?

gelisam:
This other comment explains that `--enable-tests` is also poorly-named, because it actually _requests_ the test, it does not enable them. Now that I know there is a distinction between the internal nomenclature and the user-facing nomenclature, maybe `OnlyBuildable` isn't too bad?

https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/Simple/Configure.hs#L397-L403

Mikolaj:
Actually, `OnlyBuildable` did manage to confuse me at some point, due to the `buildable` field you can have in .cabal components. I'm as bad as naming as any nerd, but I like "IfPossible", except that it looks a bit like a starting keyword of a conditionsl. Perhaps "WhenPossible", "WheneverPossible", "WhenPlanPossible", "WhenPlanExists"?

gelisam:
Thanks for the suggestions! How about `--enable-tests-when-possible` and `tests: EnableWhenPossible`?

Mikolaj:
Not bad. What would be the names for the other modes?

gelisam:
* `--enable-tests` / `tests: EnableAll` or `tests: True` for backwards compatibility
* `--disable-tests` / `tests: DisableAll` or `tests: False` for backwards compatibility

Mikolaj:
I see, so `EnableWhenPossible` for each test, enables it, when possible, while `EnableAll` force-enables all tests. We could spend a few more days on that, but I think this is good enough, hints broadly enough and prevents too simple assumptions. Documentation will tell the whole story. which is too large to be contained in the names alone.

gelisam:
@phadej , do you happen to remember what this uncommented TODO was about? You introduced it in 84d2a5e502443edcb28f910bd633471f42e54b43.

https://github.com/haskell/cabal/blob/cc598812da41e4ad6103d9f65eece558071c28be/cabal-install/src/Distribution/Client/ProjectPlanning.hs#L2602-L2604

Also, while we're looking at this code, that comment should say "it is an optional stanza, so a testsuite or benchmark", not "it is **not** an optional stanza, so a testsuite or benchmark", correct?

gelisam:
Since this mysterious TODO is on a case expression which picks an error message, perhaps it is related to this other TODO, which suggests replacing the Bool we're pattern-matching on with a 3-valued sum-type in order to affect the error message? Note that his 3-valued sum-type is _not_ the same as the `EnableComponentType` 3-valued sum-type I'm adding in this PR.

https://github.com/haskell/cabal/blob/cc598812da41e4ad6103d9f65eece558071c28be/cabal-install/src/Distribution/Client/ProjectPlanning/Types.hs#L222-L234

Mikolaj:
> Also, while we're looking at this code, that comment should say "it is an optional stanza, so a testsuite or benchmark", not "it is **not** an optional stanza, so a testsuite or benchmark", correct?

Since there is no response, feel free to amend the comment and TODOs.

gelisam:
> feel free to amend the comment and TODOs

done! They now look like this:

https://github.com/haskell/cabal/pull/7829/commits/8af774b188ed4decb65e791380a13cfb480bf5fc#diff-2adae381a2eb2d8c804d5396694a431e1e3babd816a5d2408a62213cf774a893R2603-R2609

andreasabel:
Ah, too bad this PR has stalled.  Any chance to pick it up again, @gelisam ?  Would be great to have this feature!

gelisam:
Sure, thanks for reminding me about this!

gelisam:
I unfortunately haven't been able to find the time for this, does anybody else want to take over?

ffaf1:

gelisam:
I am no longer working on this MR and I don't think I will ever find the time to come back to it, so I think it makes sense to close it, thank you.

PR #7831, opened 2021-11-23
Cabal init fileformat tests

ptkato:
This PR adapts `cabal init`'s `FileCreators` module, so it has the capability to be tested through the testing machinery proper. It is also a compromise for the stalemate in #7807, as it will be unblocked.

---

### DIFF
--- a/cabal-install/src/Distribution/Client/Init/Types.hs
+++ b/cabal-install/src/Distribution/Client/Init/Types.hs
@@ -37,6 +37,7 @@ module Distribution.Client.Init.Types
 , BreakException(..)
 , PurePrompt(..)
 , evalPrompt
+, Severity(..)
   -- * Aliases
 , IsLiterate
 , IsSimple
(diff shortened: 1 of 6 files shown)


Mikolaj (approved):
Great job.

cabal-install/src/Distribution/Client/Init/Types.hs
```diff
@@ -406,6 +417,12 @@ popList = pop >>= \a -> case P.safeRead a of
     Nothing -> throwPrompt $ BreakException ("popList: " ++ show a)
     Just as -> return as
 
+checkInvalidPath :: String -> a -> PurePrompt a
+checkInvalidPath path act = 
+      if path `elem` ["."] then
```
Kleidukos, line 422: Do we expect to add more "invalid paths" to the list? :) It would be nice to document the intent so that a subsequent refactoring doesn't remove this line in favour of a direct match in the function head.

Kleidukos (changes requested):

cabal-install/src/Distribution/Client/Init/Types.hs
ptkato: Yes, that is the case, I shall add a comment stating such intent explicitly.

Kleidukos: Wonderful

ptkato: Done.

Kleidukos (approved):
Wonderful :)

Kleidukos:
@ptkato Just take care of the conflict caused by the latest merge, and we should be good! :)

Issue #5595, opened 2018-09-27
new-run doesn't suppress its own stdout when being invoked as part of a pipe


hasufell:
Invoking `cabal new-run target -- something > data` is kinda broken, because I get cabal information messages at the front.

phadej:
@hasufell, have you tried

`cabal new-run -v0 target -- something > data`

hasufell:
Yes, didn't change anything

typedrat:
Honestly, I think that some of that information should be going to stderr instead.

ulysses4ever:
Should this be closed as superseded by #7790?

Issue #4652, opened 2017-08-01
cabal run: "Preprocessing executable" message shouldn't be in stdout


Rufflewind:
The `Preprocessing executable 'X'...` message is currently written to stdout, even for `cabal run`.  This contaminates `stdout`, so `cabal run` cannot be used with any program that outputs program-readable through `stdout` (e.g. Pandoc filters).

hvr:
Would adding `-v0` to `cabal run` help?

Rufflewind:
It does work around the problem, yes.

23Skidoo:
We discussed making `-v0` the default for `new-run` to make it more usable in scripts/pipelines and decided against it. Normally you want to see compilation output, otherwise it can be hard to tell whether your program is running or still compiling. So you should use `-v0` if you're piping the output of `cabal run` to other programs. We should change the help message for `run`/`new-run` to mention `-v0`.

Rufflewind:
In any case, I think those messages are more appropriate for stderr, especially for a "wrapper" command like cabal run.

23Skidoo:
I don't think so, GHC for example prints it compilation output to stdout and only actual error messages go to stderr.

Rufflewind:
GHC doesn't wrap any programs though, which makes this a rather atypical situation.  stderr doesn't necessarily mean "errors-only" -- it can be used for any kind of diagnostic message.

Other package managers do keep their stdout clean:

```
$ cargo run 2>/dev/null
[PROGRAM OUTPUT]
```
```
$ cargo run >/dev/null
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/X`
```
```
$ pulp run 2>/dev/null
[PROGRAM OUTPUT]
```
```
$ pulp run >/dev/null
* Building project in /home/rf/stuff/jucys/tools
* Build successful.
```
```
$ stack build --exec X 2>/dev/null
[PROGRAM OUTPUT]
```
```
$ stack build --exec X >/dev/null
X-0.1.0.0: configure (exe)
Configuring X-0.1.0.0...
X-0.1.0.0: build (exe)
Preprocessing executable 'X' for X-0.1.0.0...
[1 of 1] Compiling Main             ( ... )
Linking .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/X/X ...
X-0.1.0.0: copy/register
Installing executable(s) in
[...]

PR #7838, opened 2021-11-25
Write debug output to stderr

### LINKED ISSUE #7790: `cabal run` should send info logging to stderr not stdout
`cabal run` prints stuff like "up to date" to stdout. So running an executable and `cabal run` an executable give different output to stdout, which is no good when the executable is producing output for e.g. machine consumption. We should do our info logging to stderr.

fredefox:
This fixes #7790

---

I tested this by running `cabal run` against another projects, verifying that all cabal-related output went into stderr, and output from my application went to stdout.


fredefox:
I struggle a bit to parse the output of the test failure. If anyone could nudge me in the right direction. That would be appreciated.

Mikolaj:
From what I see, it expected to find the following on stderr, but didn't see it:

```
Test suite LibV09-Deadlock: FAIL
Test suite logged to:
setup-deadlock.dist/work/dist/test/LibV09-0.1-LibV09-Deadlock.log
```

and you can find in the log earlier that this appears in stdout now, nondeterministically interleaved with some other output to stdout. Or it may all be the other way around.

fredefox:
I'm realising now that the problem is not as simple as I first thought. For a starter `cabal-testsuite` does not keep track of whether output was written to stdout or stderr. That would be the first thing to fix. As an added kink, `cabal-testsuite` writes program output to a log file in which the expectation is (of course) that the program output occurs in the order it was written. Regardless of which pipe it was written to!

If we want to retain that we'd need some sort of multiplexing added in `Run.hs`. Although I am not sure that it can be guaranteed that data is read in the same order that it was written if `stderr != stdout`.

fredefox:
Had a stab at it just now. See my attempts in `cabal-testsuite/src/Test/Cabal/Run.hs`. It's a bit messy with all those pipes. It doesn't quite work. In testing I discovered that the value of `rAll` appears out of order. I tested it by executing `/bin/bash /tmp/test.sh` with `test.sh` being:
```
echo stdout
echo stderr >&2
```
The value of `rAll` in this case was:
```
stderr
stdout
```

Mikolaj:
Concurrency is hard. If that helps, feel free to open a new PR exclusively about a given kind of minimal cabal-testsuite improvement. That would also make reviewing and testing easier.

fredefox:
> Concurrency is hard. If that helps, feel free to open a new PR exclusively about a given kind of minimal cabal-testsuite improvement. That would also make reviewing and testing easier.

Yes. I think this is the way to go. I'll concern myself with splitting it up later though. For now I'm just hacking around. So to be clear there are currently two tasks:

* [ ] Fix `cabal-testsuite` so that it recognizes which file descriptor output goes to. This may also mean significant changes to each individual test, since they currently make no such distinction. We could also consider a testing mode where we don't care which file descriptor receives output.
* [ ] Change where Cabal writes output to. Fixup tests accordingly.

Mikolaj:
> We could also consider a testing mode where we don't care which file descriptor receives output.

Yes, I guess, that would make it possible to cheaply migrate most test and tweak only a couple to test the actual split. However, the mode would need to guarantee the combined output is deterministic. That's probably hard to do without flushing all the time.

fredefox:
I see there was previous effort to work on this https://github.com/haskell/cabal/pull/4789. I wonder if that PR is still receiving life support. Judging by the creation date I would guess it's covered in cob webs. It also looks like there was a bit of a discussion around the desired functionality. @hvr mentioned that there might be cases where you would actually want Cabal to write to stdout. I think it could be good to consolidate some of the different discussions that are happening at (at least) #7838 (here), #4789, #7790, #5595, #4652. An authoritative and slightly more in-depth answer to the question "what output should go where?" would be needed to tackle #7790 but shouldn't stand in the way of expanding the capabilities of `cabal-testsuite` mentioned in bullet no. 1 above.

Mikolaj:
Agreed, somebody would need to summarize and maintainers would need to decide. However, there's lots of stalled PRs and undecided discussions and that should not stop progress that only comes through merged PRs and user feedback. All the other people involved previously are invited to experiment with the new implementation, tweak and amend afterwards.

fredefox:
I think I managed to actually fix this issue with connecting a bunch of pipes together and waiting in the correct order. Laziness and asynchronicity is a potent coctail. Not for the faint of heart!

fredefox:
I've changed this PR to not include the change that actually alters _where_ cabal writes output to. I figure the test-change is sufficiently complicated to warrant a separate discussion. I can make another PR for the changes I excluded or if #4789 is not dead we can roll with that one. I imagine that, that change should come after merging this. I've left`changelog.d/issue-7790` in this PR as I suppose this PR still pertains to that issue.

It should be fairly easy to migrate the existing tests. Simply 1) copy all `.out` files to `.stdout` 2) run all test suites 3) updated the failing tests accordingly.

fredefox:
It should be fairly easy to migrate the tests over. Simply:

1) Copy all `.out` files to `.stdout`
2) Run all tests
3) Fix broken tests accordingly.

Mikolaj:
So is the plan of action to ask for a preliminary review approval and, if obtrained, complete the legwork for the tests? How many relevant tests are passing already? How many are not?

fredefox:
> So is the plan of action to ask for a preliminary review approval and, if obtrained, complete the legwork for the tests?

No I'll make all the tests green before requesting a review. I think that's the only proper way to go. I should've probably marked this as a draft until then. It's just a question of getting around to actually doing it :)

Mikolaj:
Hi! How is it going?

fredefox:
> Hi! How is it going?
Hi Mikolaj. I've reached the conclusion that I cannot implement it the way I had envisaged where both stdout, stderr and their collective output are caught and the order of messages are preserved. I don't think it can be done. We should consider how we would want to implement this then. My proposal would be the following:

1. The tests checks if there are `.stdout` or `.stderr` files associated with the test.
2. If there is, then capture the stdout and stderr output of the program and disregard any `.out`-files.
3. Otherwise. Run tests as they are today.

The implication of this is that you cannot test _both_ which output goes where, and the order in which the collective output is generated.

Mikolaj:
> The implication of this is that you cannot test _both_ which output goes where, and the order in which the collective output is generated.

I don't think we mind the global order often, if at all. Does it mean we could get the same output as today (or at least deterministic enough) in tests, so that we can handle the golden tests method that we use in the biggest test suite(s)?

fredefox:
> Does it mean we could get the same output as today (or at least deterministic enough) in tests, so that we can handle the golden tests method that we use in the biggest test suite(s)?

With my sketched design it should be backwards compatible. But you cannot test both modes. So if you want to switch to testing `stdout` and `stderr` output you won't be able to test how those are interleaved.

Mikolaj:
Fair enough, that's an absolute improvement as is, worth merging IMHO. What about UI? How does the test writer request either normal mode or the seperate stdout and seperate stderr mode?

fredefox:
> What about UI? How does the test writer request either normal mode or the seperate stdout and seperate stderr mode?

So I would imagine that it ought to just switch into the new mode if you there is either a `.stdout` or a `.stderr` file present in the test directory. Mind you I still haven't implemented this yet.

ulysses4ever:
Hey @fredefox! I'm curious do you think you could finish it any time soon? Great work, and I hope it gets merged!

ulysses4ever:
@merify rebase

ulysses4ever:
@Mergify rebase

ulysses4ever:
@mergify rebase

ffaf1:

Issue #6354, opened 2019-11-14
cabal triggers a full build each time a script is executed


jneira:
**Describe the bug**
cabal triggers a full build each time a script is executed instead reuse the previous one

**To Reproduce**
given a simple script:

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends:
    base
-}

main = putStrLn "hello"
```

* we can observe executing it twice

```
D:\ws\haskell\cabal-test>cabal v2-run script.hs
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
In order, the following will be built (use -v for more details):
 - fake-package-0 (exe:script) (first run)
Configuring executable 'script' for fake-package-0..
Preprocessing executable 'script' for fake-package-0..
Building executable 'script' for fake-package-0..
[1 of 1] Compiling Main             ( Main.hs, D:\ws\haskell\cabal-test\dist-new
style\build\x86_64-windows\ghc-8.6.5\fake-package-0\x\script\build\script\script
-tmp\Main.o )
Linking D:\ws\haskell\cabal-test\dist-newstyle\build\x86_64-windows\ghc-8.6.5\fa
ke-package-0\x\script\build\script\script.exe ...
hello

D:\ws\haskell\cabal-test>cabal v2-run script.hs
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
In order, the following will be built (use -v for more details):
 - fake-package-0 (exe:script) (configuration changed)
Configuring executable 'script' for fake-package-0..
Preprocessing executable 'script' for fake-package-0..
Building executable 'script' for fake-package-0..
[1 of 1] Compiling Main             ( Main.hs, D:\ws\haskell\cabal-test\dist-new
style\build\x86_64-windows\ghc-8.6.5\fake-package-0\x\script\build\script\script
-tmp\Main.o )
Linking D:\ws\haskell\cabal-test\dist-newstyle\build\x86_64-windows\ghc-8.6.5\fa
ke-package-0\x\script\build\script\script.exe ...
hello
```
that the executable is built and linked every time

**Expected behavior**
The executable installed the first time should be reused like executing `cabal v2-run target` in a project

**System information**
 - Operating system: windows
 - `cabal`: 3.0.0.0
 - `ghc` versions: tested with `8.6.5`

**Additional context**
[...]

Issue #7506, opened 2021-08-04
Per package/component `cabal clean`


michaelpj:
`cabal clean` is both simple and effective, by virtue of just deleting everything. However, in a multi-package project, I usually want to clean just one package or component that I'm working on. 

Of course, the current implementation works for this, but once you accrue a few local packages and a few `source-repository-package`s, redoing *all* the work instead of just the last bit becomes painful.

At the moment I work around this by deleting targeted fragments of `dist-newstyle/build`. This works, but:
- it's clumsy
- it's not discoverable
- it's probably not comprehensively cleaning the package (I have no idea if it gets all the files!)

I think it would be nice to provide a per package/component `cabal clean`, and hopefully it would just be a matter of listing all the relevant files and deleting them.

Mikolaj:
I guess we'd need to delete the package or component files *and* the common files. The trouble is identifying those, in particular if they are not generated with `cabal build` but only with some other commands or with some specific options.

What does `cabal clean` do exactly? Only remove `dist-newstyle` or anything else?

Mikolaj:
I see there is option

    -s, --save-config              Save configuration, only remove build artifacts

I wonder what's 'configuration' and 'artifacts' and if there is anything more. Code-diving would help.

fendor:
> I see there is option
> 
> ```
> -s, --save-config              Save configuration, only remove build artifacts
> ```
> 
> I wonder what's 'configuration' and 'artifacts' and if there is anything more. Code-diving would help.

As far as I can tell, that flag is just being ignored, but the code path *should* delete `dist-newstyle/build` only which is still not fine grained enough for the proposed feature.

fendor:
Looks rather intentional to me: https://github.com/haskell/cabal/blob/769e4ede066108cf722ac4828884788429c2d310/cabal-install/src/Distribution/Client/CmdClean.hs#L72 but the commit message itself gives no indication whether it really is intentional

Mikolaj:
May it be that 'configuration' is only `cabal.project.local`, which is not removed by `cabal clean` and that's why `cabal clean` never removes configuration in the new v2- world? If so, we should mark the `-s` flag "deprecated; configuration is always saved now" once v1- commands are removed.

However, if  'configuration' is `cabal.project.local` and artifacts are `dist-newstyle/build`, is there anything else? E.g., any kind of caches? E.g., when downloading packages form repos?

fendor:
[...]

Issue #6977, opened 2020-07-20
Meta-issue: Remove need for fake-package


phadej:
The internals of `cabal-install` should allow project planning without writing things to file system.

This is high-level goal, and I don't think that it could be solved with single PR. But internal changes moving us closer to it will improve state of the codebase, which is very welcome!

fendor:
Where is this fake-package used? 
I know of:
* cabal scrips

phadej:
@fendor, `install`, `run`, `repl` outside of project directories. About everywhere where we "solve", but don't have a local project.

phadej:
#6885 is another (upcoming) example, where planning without relying on "real world" / "filesystem" inputs is desired.

alexbiehl:
We could modify the `Rebuild` monad. Currently it's the one that creates all the caches and triggers rebuilding in case any input changed. By providing configuration or a different interpreter we could just make it not cache things and rebuild all the time. 

Pair that with allowing an empty, implicit project and we might be golden.

phadej:
@alexbiehl yes, that what I have tried. Yet `Rebuild` monad is used very little, here and there. One clear subgoal is to make project planning contained completely in `Rebuild` monad.

alexbiehl:
> One clear subgoal is to make project planning contained completely in Rebuild monad

Yes. Though a less invasive change might be to pass down the configuration/interpreter to the runRebuild calls. They are not too many. Maybe it even makes sense to tie the runRebuild to DistDirLayout; we are passing that around everywhere and usually the call looks like `runRebuild distDirProjectRoot $ do ...`.

phadej:
... except the very first `runRebuild` is `runRebuild ""`.

I also don't understand why we work with relative paths at all. That's another yak to shave. #6667 might help.

typedrat:
This is very interesting. I never liked the whole `fake-package` chicanery, and I think I still understand how that part of Cabal works pretty well, even if I haven't been up to my arms in it recently.

I had a few thinks about how this could be done, actually. Let me try to find my old notes!

fgaz:
#6999 is a specific bug caused by fake-package

Mikolaj:
[...]

PR #7851, opened 2021-12-02
Better support for running scripts.

### LINKED ISSUE #7842: Better support for running scripts.
`cabal run` can run scripts, but the utility of this is limited because under the current design it recompiles on every invocation.

I propose adding the following:
- fake packages for scripts to be written under `~/.local/share` (or os equivalent)
  - specifically `~/.local/share/cabal/<abs path of script without extension>/` to avoid collisions 
- the option `--tmp-script-artifacts` (current behaviour. keep as default?)
- the option `--persist-script-artifacts` (new behaviour)
- the option `--script-artifact-dir=DIR` (place fake package in user supplied directory)
- script support for `cabal build` and `cabal repl`
- updates to the relevant docs

I have done a quick test with hard coded paths, and it works as expected without further changes. The only changes to `run` necessary are to add the flags and pass the correct directory based on them. I haven't scoped out the work needed to add support to `build` and `repl`, but I assume it would be very similar to how it works now in `run`.

bacchanalia:
Add caching to `cabal run script` and script support with caching to `build`, `repl`, and `clean`.

This is intended to close #7842, #6354, and #6149

edit: Title and body changed to reflect end of WIP.


jneira:
hi! many thanks for open the wip pr, i hope it could help to get some general feedback from maintainers.

Maybe you have thought on but for completeness i think the pr would need:
- documentation for all new public functions (it seems they all already have it)
- some tests i can think of:
  - `cabal build/repl/run script.hs` generates the expected known directory with build products
  - all `cabal run script.hs` run effectively the script
  - `cabal repl script.hs` effectively opens a repl with the script module in scope
  - a second `cabal repl/build/run script.hs` does not rebuild anything
  - after a `cabal build script.hs`, `cabal run/repl script,hs` does not rebuild anything
 
Not sure if you have some knowledge about the [functional test suite](https://github.com/haskell/cabal/tree/master/cabal-testsuite), feel free to ask anything about. Surely it will contain similar tests over the script feature, and some about build/repl which can be used as a model. 

Although this is still a WIP i would ask maintainers to share their general thoughts and the possible drawbacks about the chosen implementation, specially if they could be a blocker for merging the pr

One of the concerns could be the interaction of this with the fake-package machinery. Afaiu the plan is to eventually remove it and replace it with a all-in-memory implementation (see https://github.com/haskell/cabal/issues/6977).
One posible adaptation could be replace caching of all build products with the caching only the executable. This could work for two consecutive runs or a build + run. Not sure how `cabal repl script.hs` could be cached without writing anything to disk though. Arguably `cabal repl` will be faster without cache but it could be noticeable for larger scripts.

//cc @gbaz @fgaz @Kleidukos

bacchanalia:
Also for completeness it needs changes to the non-haddock docs also, because it changes the interface for `build`, `clean` and `repl`

bacchanalia:
The following tests are still failing and it's not clear to me why:
- cabal-testsuite/PackageTests/Regression/T4202/cabal.test.hs
- cabal-testsuite/PackageTests/BuildTools/Internal/cabal.test.hs

In addition, the existing tests caught an issue with --builddir being set. Are there any other flags I should potentially worry about that existing tests might not catch?

Mikolaj:
The testsuite has gaps, so it can fail to catch anything. It's definitely worth debugging the return code 1. Processes are fragile and there are known issues, but we should at least try to determine what exactly is misbehaving and perhaps file a bug report about that.

bacchanalia:
@Mikolaj I found a problem in my code that may have been causing those tests to fail (I was passing the wrong ComponentKind for build), but I'm in the middle of of a refactor to solve another issue I found while writing tests (repl didn't handle the case when it was in a project correctly). Hopefully those tests will pass after I finish my current work.

bacchanalia:
I added the tests suggested by @jneira, with the except that I don't test that build artifacts are created in the cache dir because the test runner always passes a --builddir flag which overrides that location. I do test that the fake project artifacts are created in the cache dir. I have manually tested that the cache dir is used correctly when --builddir is not passed.

In addition, it would in theory be nice to tests for the code path in which the script resides outside of a project directory, but this is impossible afaik because all tests live inside the cabal-tests project directory.

Mikolaj:
I don't remember --- does cabal walk up the directory tree checking all ancestors for a project file? If so, can't a test script create a directory in /tmp, cd to it and run there? I don't count on such functionality already in the test suite and probably not worth adding it, but can such an ad-hoc solution work for this particular test?

bacchanalia:
@Mikolaj cabal does walk up the tree. If we wanted to add the functionality, then placing a test directory in /tmp would probably be how to do it, but I think it might require a significant changes to the test runner.

emilypi:
Champion! I'll get a review out asap today :3

cabal-install/src/Distribution/Client/CmdClean.hs
```diff
@@ -80,16 +87,19 @@ cleanAction CleanFlags{..} extraArgs _ = do
         mdistDirectory = flagToMaybe cleanDistDir
         mprojectFile   = flagToMaybe cleanProjectFile
 
-    unless (null extraArgs) $
-        die' verbosity $ "'clean' doesn't take any extra arguments: "
-                         ++ unwords extraArgs
+    -- assume all files passed are the names of scripts
+    notScripts <- filterM (fmap not . doesFileExist) extraArgs
```
jneira, line 91: Maybe we should be more precise about checking if target is a script file. Not sure if it is feasible but running a raw main.hs file gives:

```
> cabal run .\main.hs
cabal-3.6.2.0.exe: Failed extracting script block: `{- cabal:` start marker
not found
```

soi guess the check lives somewhere in the codebase

cabal-install/src/Distribution/Client/CmdBuild.hs
```diff
@@ -95,19 +97,16 @@ defaultBuildFlags = BuildFlags
...
         targetAction
             | onlyConfigure = TargetActionConfigure
             | otherwise = TargetActionBuild
 
-    baseCtx <- establishProjectBaseContext verbosity cliConfig OtherCommand
-
-    targetSelectors <-
-      either (reportTargetSelectorProblems verbosity) return
-      =<< readTargetSelectors (localPackages baseCtx) Nothing targetStrings
+    baseCtx <- updateContextAndWriteScriptProjectFiles ctx targetCtx
```
jneira, line 109: Maybe i am missing something, this seems to be executed for all targets but the code is in ScriptUtils. If the code can handle other targets than script files maybe should be more generic in the name and place (or make it more concrete)

bacchanalia: updateContextAndWriteScriptProjectFiles uses the targetCtx passed to it to only write in the script case. The name is a little misleading, but I thought I mentioned it in the doc string.

cabal-install/src/Distribution/Client/CmdClean.hs
bacchanalia: @jneira I'm slightly confused by if your taking about `clean` or `run` or both?
re: `clean` to handle that properly clean would have to resolve targets which is #7506 that people have been talking about.
re: `run` is a new bug I introduced or did it work like that that before?

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -195,19 +189,52 @@ replCommand = Client.installCommand {
...
           ++ "You may wish to use 'build --only-dependencies' and then "
           ++ "use 'repl'."
 
+    let projectRoot = distProjectRootDirectory $ distDirLayout ctx
+
+    (replType, baseCtx) <- case targetCtx of
+      ProjectContext     -> return (ProjectRepl, ctx)
+      GlobalContext      -> do
+        unless (null targetStrings) $
+          die' verbosity $ "'repl' takes no arguments or a script argument outside a project: " ++ unwords targetStrings
```
jneira, line 205: So we cant do a `cabal repl scriptFileInsideAProjectDir.hs`? 😞 
It will be surprising to being able to do `cabal run` and `cabal build` over such script and no `cabal repl`

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+--
+-- CABAL_DIR\/script-builds\/
+getScriptCacheDirectoryRoot :: IO FilePath
+getScriptCacheDirectoryRoot = do
+  cabalDir <- getCabalDir
+  return $ cabalDir </> "script-builds"
+
+-- | Get the hash of a script path.
+getScriptHash :: String -> FilePath -> IO String
+getScriptHash prefix script = showHashValue . hashValue . fromString . (prefix ++) <$> canonicalizePath script
```
fendor, line 104: Maybe relevant to note, at what point the hash changes (when the filename changes).

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+data AcceptNoTargets
+  = RejectNoTargets -- ^ die on 'TargetSelectorNoTargetsInProject'
+  | AcceptNoTargets -- ^ return a default 'TargetSelector'
+  deriving (Eq, Show)
+
+-- | Information about the context in which we found the 'TargetSelector's.
+data TargetContext
+  = ProjectContext -- ^ The target selectors are part of a project.
+  | GlobalContext  -- ^ The target selectors are from the global context.
+  | ScriptContext { scriptPath :: FilePath, scriptExecutable :: Executable, scriptContents :: BS.ByteString }
```
fendor, line 133: Not sure what the cabal coding guideline is here, do we *want* to have partial accessor functions such as `scriptPath`?

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+
+-- | Add the 'SourcePackage' to the context and use it to write a fake-package.cabal file.
+updateContextAndWriteProjectFile :: ProjectBaseContext -> SourcePackage (PackageLocation (Maybe FilePath)) -> IO ProjectBaseContext
+updateContextAndWriteProjectFile ctx srcPkg = do
+  let projectRoot = distProjectRootDirectory $ distDirLayout ctx
+      projectFile = projectRoot </> "fake-package.cabal"
+      writeProjectFile = writeGenericPackageDescription (projectRoot </> "fake-package.cabal") (srcpkgDescription srcPkg)
+  projectFileExists <- doesFileExist projectFile
+  -- TODO This if is here to prevent reconfiguration of cached repl packages.
+  -- It's worth investigating why it's need in the first place.
```
fendor, line 224:
```suggestion
  -- TODO This is here to prevent reconfiguration of cached repl packages.
  -- It's worth investigating why it's needed in the first place.
```

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+      tmpDir <- getTemporaryDirectory >>= flip createTempDirectory "cabal-repl."
+      putMVar m tmpDir
+      return tmpDir
+    rmTmp m _ = tryTakeMVar m >>= maybe (return ()) (handleDoesNotExist () . removeDirectoryRecursive)
+
+-- | Add the 'SourcePackage' to the context and use it to write a fake-package.cabal file.
+updateContextAndWriteProjectFile :: ProjectBaseContext -> SourcePackage (PackageLocation (Maybe FilePath)) -> IO ProjectBaseContext
+updateContextAndWriteProjectFile ctx srcPkg = do
+  let projectRoot = distProjectRootDirectory $ distDirLayout ctx
+      projectFile = projectRoot </> "fake-package.cabal"
```
fendor, line 220: Can we turn the name into some constant?

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+  -- It's worth investigating why it's need in the first place.
+  if projectFileExists then do
+    contents <- readUTF8File projectFile
+    when (contents /= showGenericPackageDescription (srcpkgDescription srcPkg))
+      writeProjectFile
+  else writeProjectFile
+  return (ctx & lLocalPackages %~ (++ [SpecificSourcePackage srcPkg]))
+
+-- | In a script context, add a 'SourcePackage' to the context and write a fake-package.cabal file
+-- and the script source file (Main.hs or Main.lhs).
```
fendor, line 233: Since this is going to be in `CmdBuild`, etc... it should also mention what happens if we are not in a script context, imo.

doc/cabal-commands.rst
```diff
@@ -184,6 +192,17 @@ cannot be excluded for technical reasons).
 
     $ cabal v2-repl --build-depends vector --no-transitive-deps
 
+Scripts can be passing the path to the script as the target.
+
+::
+
+    $ cabal v2-repl script.hs
+
+If the name of the script matches another target, the other target is preferred.
```
fendor, line 201: In what context may that arise? I suspect in a project (e.g. cabal.project exists) and `script.hs` is part of the cabal project?

fendor: Also, the same caveat will apply for `v2-build` as well, right?

doc/cabal-commands.rst
```diff
@@ -131,6 +136,9 @@ cause extra store packages to be built (for example,
 ``--enable-profiling`` will automatically make sure profiling libraries
 for all transitive dependencies are built and installed.)
 
+When building a script, the executable is cached under the cabal directory.
```
fendor, line 139: Is it also cached when executed only via `v2-run`?

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+data TargetContext
+  = ProjectContext -- ^ The target selectors are part of a project.
+  | GlobalContext  -- ^ The target selectors are from the global context.
+  | ScriptContext { scriptPath :: FilePath, scriptExecutable :: Executable, scriptContents :: BS.ByteString }
+  -- ^ The target selectors refer to a script.
+  deriving (Eq, Show)
+
+-- | Determine whether the targets represent regular targets or a script
+-- and return the proper context and target selectors.
+-- Report problems if selectors are valid as neither regular targets or as a script.
```
fendor, line 139: How are problems reported?
I suspect just like before, which was `print to stderr >> die`, but can we document that here as well?

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+import qualified Text.Parsec as P
+import System.Directory
+         ( getTemporaryDirectory, removeDirectoryRecursive, doesFileExist, canonicalizePath )
+import System.FilePath
+         ( (</>), takeExtension )
+
+
+-- | Get the directory where script builds are cached.
+--
+-- CABAL_DIR\/script-builds\/
```
fendor, line 96: In haddock, renders more nicely when you have:
```suggestion
-- @CABAL_DIR\/script-builds\/@
```

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,342 @@
...
+         ( NixStyleFlags (..) )
+import Distribution.Client.Setup
+         ( GlobalFlags(..), ConfigFlags(..) )
+import Distribution.Client.Config
+         ( getCabalDir )
+import Distribution.Simple.Flag
+         ( fromFlagOrDefault )
+import Distribution.Simple.Setup
+         ( Flag(..) )
+import Distribution.CabalSpecVersion (CabalSpecVersion (..), cabalSpecLatest)
```
fendor, line 34: Probably format this line as well, just like the others?

fendor (changes requested):
Looking great! (Assuming everything works as intended ;D)
I basically have only minor nitpicks, regarding docs, formatting, etc...

Really great work, can't wait to have `cabal repl` support to finally unbreak HLS on cabal scripts :scream:

cabal-install/src/Distribution/Client/CmdClean.hs
```diff
@@ -80,16 +87,19 @@ cleanAction CleanFlags{..} extraArgs _ = do
...
+    unless (null notScripts) $
+        die' verbosity $ "'clean' extra arguments should be script files: "
+                         ++ unwords notScripts
```
fendor, line 94: I think this should have a TODO to accept arbitrary cabal components, since `v2-clean` will hopefully generalise nicely in the future for all cabal components.

cabal-install/src/Distribution/Client/CmdClean.hs
fendor: Actually the error message looks like a regression, running `cabal v2-run cabal-install/main/Main.hs` yields:
```
Resolving dependencies...
cabal: The run command can only run an executable as a whole, not files or
modules within them, but the target 'main/Main.hs' refers to the file
main/Main.hsin the executable cabal.
```

which makes total sense, but the current error message does not make sense at all for newcomers.

cabal-install/src/Distribution/Client/ScriptUtils.hs
fendor: Additionally, is there a definition of what `CABAL_DIR` is anywhere? Just for my own curiosity

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: I'll also make a note about prefixs, which I meant to do but apparently forgot.

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: I thought it had value for documentation and to use with recordwildcards, and I didn't think about their partiality. Should I remove them?

Mikolaj: We don't have `ghc-options: -Wpartial-fields` in our cabal files, so this is legal. My personal preference is to add this warning and then remove all places where it's triggered (no idea how many and if that would break backward compatibility for Cabal lib users). But that requires an experiment, search through old issues and, if all positive, a separate RFC ticket. Volunteers welcome.

doc/cabal-commands.rst
bacchanalia: I should probably remove the .hs because script doesn't necessarily have to end with .hs. so you could have a target called `mangler` and a script called `mangler` or whatever. You're right that it can occur for the other commands as well.

doc/cabal-commands.rst
bacchanalia: yes

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: By CABAL_DIR, I meant the environment variable or default location. I'm open so suggestions about a better way to render this.

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: A lot of the code for ScriptUtils came from CmdRun, but since it's not being explicitly diffed with CmdRun, I guess there's no reason to keep it.

cabal-install/src/Distribution/Client/CmdClean.hs
bacchanalia: to elaborate on precise checking: I didn't want to explicitly check if all passed args are scripts because it's possible they might refer to things that where built as scripts in the past and where either deleted or turned into normal source files and you still want to clean the cache for them. That's a bit of an edge case so it might not be worth the worse error reporting.

re error messages: The exact error messages cabal prints depends on the context. In this case I believe it's whether there is a .cabal file in the current directory or not. It should probably give the same error message when you try to run the Main.hs of a package regardless of the directory your in, but this is the same behavior as before the pr, not a regression.

doc/cabal-commands.rst
fendor: Please document that, thank you :)

cabal-install/src/Distribution/Client/ScriptUtils.hs
fendor: I figure so, but it is probably hard to discover... Don't know how we can improve that right now.

bacchanalia:
@fendor could you elaborate on what v2-run tests you'd like me to add?

fendor:
> @fendor could you elaborate on what v2-run tests you'd like me to add?

What I had in mind is a test each for run:
* Trying to run a Main.hs from a cabal component (that fails right now, no?)
* Trying to run a script where the script part is missing (must fail)

At least the latter will probably document that we could improve the error message quite a bit.
However, if you feel like this is out-of-scope for this PR, don't feel pressured by me to add those tests, my PR approval does not depend on it.

bacchanalia:
@fendor It's a little strange that `cabal run app/Main.hs` does not work considering that `cabal build app/Main.hs` does, but I think that is outside the scope of this pr. Adding the tests is no problem.

fendor:
Exactly, that's clearly out-of-scope, but adding the test would be incredibly awesome!

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,354 @@
...
+        let projectRoot = distProjectRootDirectory $ distDirLayout ctx
+        writeFile (projectRoot </> "scriptlocation") =<< canonicalizePath script
+
+        (executable, contents) <- readScriptBlockFromScript verbosity =<< BS.readFile script
+
+        let executable' = executable & L.buildInfo . L.defaultLanguage %~ maybe (Just Haskell2010) Just
+        return (ScriptContext script executable' contents, ctx, defaultTarget)
+      else reportTargetSelectorProblems verbosity err
+
+withTemporaryTempDirectory :: (IO FilePath -> IO a) -> IO a
```
jneira, line 218: Just in case: have you considered use https://github.com/haskell/cabal/blob/6a8c3f80173b71436386d9724b64eb67ccb1cf40/Cabal/src/Distribution/Simple/Utils.hs#L1467-L1470?

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -324,77 +352,15 @@ data OriginalComponentInfo = OriginalComponentInfo
...
-  let
-    targetSelectors = [TargetPackage TargetExplicitNamed [pkgId] Nothing]
-    finalizer = handleDoesNotExist () (removeDirectoryRecursive tempDir)
-
-  return (baseCtx, targetSelectors, finalizer, GlobalRepl ghciScriptPath)
+-- Tracks what type of GHCi instance we're creating.
+data ReplType = ProjectRepl | GlobalRepl
+  deriving (Show, Eq)
+
+relativePathBackToCurrentDirectory :: FilePath -> IO FilePath
```
jneira, line 359: not sure what is the intent but have you take a look to

https://github.com/haskell/cabal/blob/4309d0f60089bfe308af63de8671e1723c284c92/cabal-install/src/Distribution/Client/Utils.hs#L222-L230
?

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: It's a temporary temp directory, that's twice as temporary! Being serious though, it does use `withTempDirectory`. With `withTempDirectory` creates a directory in a given dir using the template meant to me used as a temp directory. `withTemporaryTempDirectory` does that under /tmp or os equivilent, hence temporary temp.

cabal-install/src/Distribution/Client/CmdRepl.hs
bacchanalia: makeRelativeToDir seems like what I want. Good catch!

bacchanalia: When removing `relativePathBackToCurrentDirectory` in favor of `makeRelativeToDir`, I discovered a bug in my code that I'm having a lot of trouble fixing properly. When configuring the project, the script file is relative to the hs-source-dir, but when running the repl, the script file is actually loaded after the repl changes directory, so it needs a different path to the script file.

The final command line used to launch ghc would be easy to fix, but I think it gets constructed somewhere deep in the code that I don't think I have easy access to, and none of the components exposed in replAction seem to help, as far as I've been able to tell.

bacchanalia: The hacky solution: use a path with a number of "../"' which is max (# path segments of project root) (# path segments of working directory) followed by the absolute path to the script. That way we're guaranteed to go up to the root and can then find the script by it's absolute path. This is also why it seemed to work before, because in all my testing (# path segments of project root) was greater than (# path segments of working directory).

Other possible solution: relax the restriction that hs-source-dirs has to be relative. Maybe we could check that the package-id maches the fake package-id constant and only allow it then?

bacchanalia: As recently as 3.4.0.0 cabal did accept absolute paths in `hs-source-dirs`

Mikolaj: See https://github.com/haskell/cabal/issues/6787#issuecomment-991206955. (The plot thickens).

bacchanalia:
Re #6787, if we use the same path trick for `build`/ `run` as I'm using for `repl`, instead of copying the source file, that should get us multi-module support basically for free.

jneira:
Hi. i am trying this pr in windows and i've got a somewhat weird issue:

* given this standalone script file

```haskell
#! /usr/bin/env cabal
{- cabal:
build-depends: base >= 4.3 && <5, haskell-say
-}

import HaskellSay

main :: IO ()
main = haskellSay "Hello World"
```

* `cabal run say.hs` works fine although paths reported are a little bit ugly:

```
PS D:\dev\ws\haskell\issues\scripts> cabal run .\say.hs
Build profile: -w ghc-9.0.1 -O1
 - fake-package-0 (exe:script) (file ..\..\..\..\..\dev\ws\haskell\issues\scripts\say.hs changed)
[1 of 1] Compiling Main             ( ..\..\..\..\..\dev\ws\haskell\issues\scripts\say.hs, D:\cabal\script-builds\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\dist-newstyle\build\x86_64-windows\ghc-9.0.1\fake-package-0\x\script\build\script\script-tmp\Main.o )
Linking D:\cabal\script-builds\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\dist-newstyle\build\x86_64-windows\ghc-9.0.1\fake-package-0\x\script\build\script\script.exe ...
```

* but if i change the cabal metadata (adding or removing a dependency or another field) and try again it fails:

```
PS D:\dev\ws\haskell\issues\scripts> cabal run .\say.hs
D:\cabal\script-builds\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\fakCA05.tmp: renameFile:renamePath:MoveFileEx "\\\\?\\D:\\cabal\\script-builds\\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\\fakCA05.tmp" Just "\\\\?\\D:\\cabal\\script-builds\\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\\fake-package.cabal": permission denied (Acceso denegado.)
```

It does not happen if i only changed the hs file contents

bacchanalia:
@jneira github doesn't let me react to to posts with "😡", but "😡" none the less

bacchanalia:
Last patch has some serious issues. I need to at least partially revert it or find a better solution.

bacchanalia:
currently blocked on bug #7867

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,343 @@
...
+  return cacheDir
+
+-- | What your command should do when no targets are found.
+data AcceptNoTargets
+  = RejectNoTargets -- ^ die on 'TargetSelectorNoTargetsInProject'
+  | AcceptNoTargets -- ^ return a default 'TargetSelector'
+  deriving (Eq, Show)
+
+-- | Information about the context in which we found the 'TargetSelector's.
+data TargetContext
```
jneira, line 134: It seems to me that this, `AcceptNoTargets` and `withContextsAndSelectors` should live in a more generic module, maybe one with other target related definitions

cabal-install/src/Distribution/Client/ScriptUtils.hs
```diff
@@ -0,0 +1,343 @@
...
+-- and return the proper context and target selectors.
+-- Die with an error message if selectors are valid as neither regular targets or as a script.
+--
+-- In the case that the context refers to a temporary directory,
+-- delete it after the action finishes.
+withContextAndSelectors
+  :: AcceptNoTargets     -- ^ What your command should do when no targets are found.
+  -> Maybe ComponentKind -- ^ A target filter
+  -> NixStyleFlags a     -- ^ Command line flags
+  -> [String]            -- ^ Target strings or a script and args.
```
jneira, line 152: Have you consider convert this in a sum type like `TargetInfo` (or whatever), tracking in the type system this useful comment? too much work at call sites?

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: `withContextAndSelectors` I think makes since in a `ScirptUtils` as it's specifically a utility function for helping commands deal with script targets. I guess `withContextAndSelectors, AcceptNoTargets(..), TargetContext(..),` could be in their own module for `SciptTargetUtils` (open to name suggestions), but I'm nor sure what the benefit is.

I'm not entirely happy with the name `ScirptUtils`. Maybe a solution is to find a better name for the module as a whole instead of splitting it?

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: `NixStyleFlags a -> [Strings] -> GlobalFlags ->` are what get passed to all command options
`AcceptNoTargets -> Maybe ComponentKind ->` I added as I realized I needed them to account for the differences between commands.

Are you suggesting combing just the first two or all five? I don't think it would be too much work at call sites, but I'm not sure I see a clear benefit either.

bacchanalia:
@Kleidukos proposes that for this PR (and I assume as a new coding style guideline for cabal generally) moving from return to pure. I'm sympathetic to their arguments, basically that return is a confusing historical accident (also pure is shorter), but I wanted to get the feedback of other maintainers.

Mikolaj:
Mandating 'pure' in guidelines will require an RFC ticket or a discussion elsewhere, but the status quo is that `pure` is permitted and so IMHO go ahead and try this style.

cabal-install/src/Distribution/Client/ScriptUtils.hs
jneira: I was thinking in something like

```haskell

-- With a better name for Targets
data TargetInfo = Targets [String] | ScriptTarget String [String] 

withContextAndSelectors :: AcceptNoTargets -> Maybe ComponentKind -> NixStyleFlags a -> TargetInfo  -> GlobalFlags -> (TargetContext -> ProjectBaseContext -> [TargetSelector] -> IO b) -> IO b
```
Cause [String] is being used for two different things and mixing them should be a type error.
But if `[String]` is used everywhere it is out of the scope of the pr

cabal-install/src/Distribution/Client/ScriptUtils.hs
jneira: Hmm the signature of the function, its name and types are generic and is being used for run, build and repl, as a higher order function.
It is a little bit surprising see it living in script utils only looking to the signature.
But its implementation is mainly script related, as it prapares context and selectors for the script case.

Idk, maybe changing the name to `withScriptContextAndSelectors` could help? As you consider.

Then if you look at the implementation, it is handling three cases:

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: Thinking about this, `withContextAndSelectors` truncates `targetStrings` based on the the `ComponentKind`, which is a bit ad-hoc. Really it's `CmdRun,hs` that knows that only the first one is a target string and should only pass that one.

cabal-install/src/Distribution/Client/ScriptUtils.hs
bacchanalia: I don't love the name `TargetContext` because Context is so heavily overloaded, but I didn't know what else to call it.  I'm not sure I like `withScriptContextAndSelectors` because it implies it's specific to when there is a script rather than optionally accepting scripts.

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -195,19 +190,42 @@ replCommand = Client.installCommand {
...
+            & lSrcpkgDescription . L.condLibrary
+            .~ Just (CondNode library [baseDep] [])
+          library = emptyLibrary { libBuildInfo = lBuildInfo }
+          lBuildInfo = emptyBuildInfo
+            { targetBuildDepends = [baseDep]
+            , defaultLanguage = Just Haskell2010
+            }
+          baseDep = Dependency "base" anyVersion mainLibSet
+
+        (,,) GlobalRepl Nothing <$> updateContextAndWriteProjectFile' ctx sourcePackage
```
fendor, line 219: Gotta be honest, I hate this :D

TupleSections maybe? 
```suggestion
        (GlobalRepl,Nothing,) <$> updateContextAndWriteProjectFile' ctx sourcePackage
```

If you prefer it this way, your choice!

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -274,29 +292,35 @@ replAction flags@NixStyleFlags { extraFlags = (replFlags, envFlags), ..} targetS
...
-            _                                   -> replFlags'
+
+        -- The '-ghci-script' flag is path to a GHCi script responsible for changing to the
+        -- correct directory. Only works on GHC >= 7.6, though. 🙁
+        replFlags'' <- case replType of
+          GlobalRepl | compilerCompatVersion GHC compiler >= Just minGhciScriptVersion -> do
+            let ghciScriptPath = projectRoot </> "setcwd.ghci"
+            cwd <- getCurrentDirectory
+            writeFile ghciScriptPath (":cd " ++ cwd)
+            return $ ("-ghci-script" ++ ghciScriptPath) : replFlags'
```
fendor, line 303: Can we maybe abstract this a bit? E.g. A function with a real name.

Additionally, why do we need this ghci-script? Why isn't it in the correct directory? I know that we both know the answer right now, but do we still know it in a year? :D would you mind documenting this "hack"?

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -274,29 +292,35 @@ replAction flags@NixStyleFlags { extraFlags = (replFlags, envFlags), ..} targetS
...
+          _ -> return replFlags'
 
         return (buildCtx, replFlags'')
 
+    incDir <- maybeToList <$> mapM (canonicalizePath . dropFileName) mScript
+    script <- if replOptionsNoLoad replFlags == Flag True
```
fendor, line 309:
```suggestion
    script <- if fromFlagOrDefault False (replOptionsNoLoad replFlags)
```

Avoids the `Flag` constructor. However, noticeably longer and maybe more confusing

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -274,29 +292,35 @@ replAction flags@NixStyleFlags { extraFlags = (replFlags, envFlags), ..} targetS
...
+              then return []
+              else maybeToList <$> mapM (\s -> if isAbsolute s then return s else makeRelativeToCwd s) mScript
     let buildCtx' = buildCtx
           { elaboratedShared = (elaboratedShared buildCtx)
             { pkgConfigReplOptions = replFlags
-              { replOptionsFlags = (replOptionsFlags replFlags) ++ replFlags''
+              { replOptionsFlags  = replOptionsFlags replFlags ++ replFlags'' ++ fmap ("-i" ++) incDir ++ script
```
fendor, line 315: magic include seems weird and breaks some abstractions in my opinion. I.e. do we anywhere just randomly add include directories?

Why does the repl need it? Needs at least documentation why we add it here, I am fine with refactoring it later though.

cabal-install/src/Distribution/Client/CmdRepl.hs
```diff
@@ -324,77 +348,9 @@ data OriginalComponentInfo = OriginalComponentInfo
...
-      distDirLayout
-      [SpecificSourcePackage sourcePackage]
-      OtherCommand
-
-  let
-    targetSelectors = [TargetPackage TargetExplicitNamed [pkgId] Nothing]
-    finalizer = handleDoesNotExist () (removeDirectoryRecursive tempDir)
-
-  return (baseCtx, targetSelectors, finalizer, GlobalRepl ghciScriptPath)
+-- Tracks what type of GHCi instance we're creating.
```
fendor, line 351:
```suggestion
-- | Tracks what type of GHCi instance we're creating.
```

Haddock ftw! Because HLS can show you this documentation and believe me, you want this ;D 

Additionally, maybe mention why or how we handle this differently? (Just bonus though, I am happy with haddock comment)

cabal-testsuite/PackageTests/NewBuild/CmdRun/ScriptBad/cabal.project
```diff
@@ -0,0 +1 @@
+packages: .
```
fendor, line 1: For consistency, should have a trailing newline.

fendor (changes requested):
Looking good to me already, I think this is my final round of minor complaints! Thank you for pushing this through, I am besides my minor nitpicks happy with this PR!

fendor (approved):
Good to go from my point of view!

Big thank you for this huge amount of code documentation you added in the process!

jneira:
The relative path until root does not look very nice tbh. I've tried last version of the pr:

```
PS D:\dev\ws\haskell\issues\scripts> cabal run say.hs
Resolving dependencies...
Build profile: -w ghc-8.10.7 -O1
Warning: 'hs-source-dirs: ..\..\..\dev\ws\haskell\issues\scripts' is a
relative path outside of the source tree. This will not work when generating a
tarball with 'sdist'.
[1 of 1] Compiling Main             ( ..\..\..\dev\ws\haskell\issues\scripts\say.hs, D:\cabal\script-builds\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\dist-newstyle\build\x86_64-windows\ghc-8.10.7\fake-package-0\x\script\build\script\script-tmp\Main.o )
Linking D:\cabal\script-builds\e94785007928b1a03e1723fddefd5bddd0aac4147a5c6b64c8162b91509d8755\dist-newstyle\build\x86_64-windows\ghc-8.10.7\fake-package-0\x\script\build\script\script.exe ...
```

Note it is emitting a warning.

Would be possible to make some conversion to absolute paths at some point?

bacchanalia:
The latest version just uses relative path to the script from the fake-project directory. It no longer needs to go back to root for repl, because I pass the script location to the repl explicitly. I agree that absolute paths would be preferable, but due to other design choices they are not so easy to implement right now. I was assuming that #6977 would make it irrelevant eventually, but if it's a blocking issue for you I can look into it now.

jneira:
oh sorry, I thought I was checking the last version
i. fine if you are planning in work in it afterwards

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -1099,6 +1100,13 @@ checkPaths pkg =
   | pat <- extraDocFiles pkg
   , Left err <- [parseFileGlob (specVersion pkg) pat]
   ]
+  ++
+  [ PackageBuildWarning $
+        "Scripts do not support the field: " ++ field ++ " "
+     ++ "If you need this field, please use a full package."
```
jneira, line 1106: minor thaing about the output: the upper case without a previous period look a bit strange:

```
Warning: Scripts do not support the field: include-dirs If you need this
field, please use a full package.
```
```suggestion
        "Scripts do not support the field: '" ++ field ++ "'. "
     ++ "If you need it, please use a full package."
```

bacchanalia:
I put in the checks, because if a user adds a field to a script with a relative path it probably won't work like they think it will, but cabal-install also adds some of those fields itself when it's elaborating a package, and I don't have a good idea of which ones can be added by cabal-install, so I have removed the checks.

jneira (approved):
Looks great, many many thanks for work on this and address all our requests and suggestions.
The pr has a good test suite and make scripts works as a reasonable user would expect.

jneira:
will merge after wait one year if nobody disagree 😝

bacchanalia:
who's supposed to hit the button?

Kleidukos:
@Mergifyio rebase

PR #7861, opened 2021-12-09
docs: move flags not supported by Setup.hs, fixes #7859

### LINKED ISSUE #7859: --allow-newer with Setup.hs configure?
Docs list `--allow-newer` under flags that [`runhaskell Setup.hs configure` supports][1], but it doesn't seem to actually do anything. The only way I've found to break the bounds with `Setup.hs configure` is to use `--exact-configuration` and specify everything.

[1]: https://cabal.readthedocs.io/en/3.6/setup-commands.html#cmdoption-runhaskell-Setup.hs-configure-allow-newer

adamse:
---

### DIFF
--- a/doc/cabal-commands.rst
+++ b/doc/cabal-commands.rst
@@ -1,8 +1,146 @@
 cabal-install Commands
 ======================
 
-We now give an in-depth description of all the commands, describing the
-arguments and flags they accept.
+We now give an in-depth description of all the commands, first describing the
+arguments and flags that are common to some or all of them.
+
+
+.. option:: --default-user-config=file
+
+    Allows a "default" ``cabal.config`` freeze file to be passed in
+    manually. This file will only be used if one does not exist in the
+    project directory already. Typically, this can be set from the
+    global cabal ``config`` file so as to provide a default set of
+    partial constraints to be used by projects, providing a way for
+    users to peg themselves to stable package collections.
+
+
+.. option:: --allow-newer[=pkgs], --allow-older[=pkgs]
+
+    Selectively relax upper or lower bounds in dependencies without
+    editing the package description respectively.
+
+    The following description focuses on upper bounds and the
+    :option:`--allow-newer` flag, but applies analogously to
+    :option:`--allow-older` and lower bounds. :option:`--allow-newer`
+    and :option:`--allow-older` can be used at the same time.
+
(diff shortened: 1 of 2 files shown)


doc/setup-commands.rst
```diff
@@ -1030,66 +957,6 @@ Miscellaneous options
         $ cabal install --constraint="foo:bar:exe.baz == 1.0"
 
 .. option:: --preference=preference
```
Mikolaj, line 959: leftover?

doc/cabal-commands.rst
```diff
@@ -51,6 +51,143 @@ is in place, which moves the old configuration to a ``cabal.project.local~``
 file, this feature can also be disabled by using the ``--disable-backup``
 flag.
 
+.. option:: --default-user-config=file
```
Mikolaj, line 54: Here, we are in a subsection about `v2-configure`. Is there a section/subsection somewhere about commands common to all v2- commands?

adamse: i can put them at the top section (https://cabal.readthedocs.io/en/3.6/cabal-commands.html)?

Mikolaj: Sure. Or create a distinct subsection --- whichever you choose. Now that I think about it, I'm not sure these are common to all v2-commands. There may be exceptions. Are there? If so, a couple of sentences before the command listing starts should say that they are common to many v2-commands, but to determine which belongs to which, one should use `cabal help cmd`.

Mikolaj: BTW, this links is better: https://cabal.readthedocs.io/en/latest

Mikolaj: I mean, the reader of the docs needs to experiment. Yes, moving is enough, with a sentence that says something, but not too much --- this is what I'm agonizing about.

Mikolaj:
I'm re-running the failing CI job.

doc/cabal-commands.rst
Mikolaj: Perhaps just tweak "describing the arguments and flags they accept"? E.g., "first describing the arguments and flags that are common to some or all of them"? That implies that args and flags in subsections are not common.

Mikolaj (approved):
Looks great. Thank you!

Kleidukos (approved):
LGTM. :)

Issue #7867, opened 2021-12-12
super-exponential explosion of --repl-options


bacchanalia:
**Describe the bug**
When passing multiple --repl-options to cabal repl they get replicated super-exponentially

**To Reproduce**
```
$ cabal v2-repl --repl-options="A" --repl-options="B"
```
or if you want to see 2^15-1 errors
```
$ cabal v2-repl --repl-options="A" --repl-options="B" --repl-options="C" --repl-options="D" 2>&1 | grep error | wc -l
Prelude> :q
```

**Expected behavior**
Each option is passed once.

**System information**
 - nixos
 - `ghc-8.10.7`
 - `cabal` HEAD

**Additional context**
Introduced after 3.6.2.0

bacchanalia:
This seems to fix it, but still not sure exactly why it was causing the exact bad behavior, so it might be masking a second issue.
```diff
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -1700,7 +1700,7 @@ instance Monoid ReplOptions where
   mappend = (<>)
 
 instance Semigroup ReplOptions where
-  (<>) = gmappend
+  _ <> r = r
 
 data ReplFlags = ReplFlags {
     replProgramPaths :: [(String, FilePath)],
```
Even if we don't want to keep the left arg to (<>), why is (<>) getting called enough for us to see the 2^(2^n) behavior?

bacchanalia:
The offending commit is 785ea62d2

bacchanalia:
It was the option parser, which makes sense. Fix:
```diff
diff --git a/Cabal/src/Distribution/Simple/Setup.hs b/Cabal/src/Distribution/Simple/Setup.hs
index 6ed08e7ed..bfd020b62 100644
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -1808,7 +1808,7 @@ replOptions _ =
     trueArg
   , option [] ["repl-options"]
     "use this option for the repl"
-    replOptionsFlags (\p flags -> flags { replOptionsFlags = p ++ replOptionsFlags flags })
+    replOptionsFlags (\p flags -> flags { replOptionsFlags = p })
     (reqArg "FLAG" (succeedReadE (:[])) id)
   ]
 ```

jneira:
[...]

PR #7868, opened 2021-12-13
Fix --repl-options parsing

bacchanalia:
Bug #7867

### DIFF
--- a/Cabal/src/Distribution/Simple/Setup.hs
+++ b/Cabal/src/Distribution/Simple/Setup.hs
@@ -1808,7 +1808,7 @@ replOptions _ =
     trueArg
   , option [] ["repl-options"]
     "use this option for the repl"
-    replOptionsFlags (\p flags -> flags { replOptionsFlags = p ++ replOptionsFlags flags })
+    replOptionsFlags (\p flags -> flags { replOptionsFlags = p })
     (reqArg "FLAG" (succeedReadE (:[])) id)
   ]
 


jneira (commented):
Hi, thanks for the standalone fix, it would need a regression test, but if you have tested manually and it is included in your main pr in some way it would be fine for me.
A changelog entry would be great.

jneira:
As noted by @bacchanalia the bug is not in any release so the changelog would not be needed and the main pr is already testing it, could you add a brief note about what test is and how is testing the bug?

bacchanalia:
In the latest design of #7851, I use replOptionsFlags to pass the script location between cabal-install and Cabal, so either `cabal-testsuite/PackageTests/NewBuild/CmdRepl/Script` or `cabal-testsuite/PackageTests/NewBuild/CmdRepl/ScriptRerun` would catch it because the repl would fail to load from the script file being passed multiple times.

jneira:
The pr (#7578) by @sirlensalot had specific tests so i hope this change will not break the functionality added there

jneira (approved):
looks good to me given the notes about the changelog and tests, thanks for the fix!

Mikolaj:
I think the backport of this bug is in 3.6.2 (#7593), ain't it so?

jneira:
it seems #7593 the pr was closed wisely by @fgaz without merging it

Mikolaj:
Wow, I missed it. Well done, team. :)

Mikolaj (approved):
Super.

jneira:
no need for backport then (i was not sure if the bug was backported aftere the release)

sergv:
Although I'm not the original author, I can confirm that the commit is correct. I did encounter this problem myself and did a bit of investigation.

The original commit (cf https://github.com/haskell/cabal/commit/785ea62d2b6c7d2e91654eb7a23cafccac3ec08c#diff-793c63febac9c808425678cd5535233aebb1aca838a829a665f85d4452f62346L1766) did the following change
<pre><code>
-replOptions :: ShowOrParseArgs -> [OptionField [String]]
-replOptions _ = [ option [] ["repl-options"] "use this option for the repl" <b>id</b>
-              <b>const</b> (reqArg "FLAG" (succeedReadE (:[])) id) ]
+replOptions :: ShowOrParseArgs -> [OptionField ReplOptions]
+replOptions _ =
+  [ option [] ["repl-no-load"]
+    "Disable loading of project modules at REPL startup."
+    replOptionsNoLoad (\p flags -> flags { replOptionsNoLoad = p })
+    trueArg
+  , option [] ["repl-options"]
+    "use this option for the repl"
+    <b>replOptionsFlags</b> <b>(\p flags -> flags { replOptionsFlags = p ++ replOptionsFlags flags })</b>
+    (reqArg "FLAG" (succeedReadE (:[])) id)
+  ]
</code></pre>

Note that what previously was `id` and `const` now became `replOptionsFlags` and a lambda `(\p flags -> ...)`. This is a getter and setter pair however previously the setter was not updating the old flags value. That's because ultimately during option parsing the appending of options will take place at https://github.com/haskell/cabal/blob/master/Cabal/src/Distribution/Simple/Command.hs#L159 (notice the `mappend`).

So the short summary is that previously there were no `++` and it shouldn't have been added in the https://github.com/haskell/cabal/commit/785ea62d2b6c7d2e91654eb7a23cafccac3ec08c commit.

Mikolaj:
@sergv: thank you for the analysis and confirmation. I think the merge is overdue. Thank you all!

sirlensalot:
Sorry for the delay responding, but obviously I concur with it as the author of the bug! Clearly my tests didn't catch this.

Indeed my only comment would have been: coverage? Can we get the bugfix covered by a unit test. I wrote a test for my functionality, which clearly didn't cover so I'm assuming this broke other workflows. Can we get coverage for those in a follow-on PR please?

sirlensalot:
Happy to help if somebody can give the use-case clearly

bacchanalia:
I though this would be covered by the tests in the other PR (#7851) I was working on, but due to code changes in that pr the tests no longer cover this, so we do still need tests for it.

PR #7873, opened 2021-12-20
Extra argument for `cabal init` as starting dir

### LINKED ISSUE #7871: Allow `cabal init` to create a new project directory
Currently, `cabal init` initializes a new project in the current directory. It would be nice if it could also create a new directory where the project is located, so that one doesn't have to do that manually, similar to `stack new`. This could also be a new `cabal new` command.

ptkato:
Closes #7871. `cabal init` will now take the first value in `extraArgs` and treat it as the root directory for the new project; i.e.:
```
cabal init -n -- my-project
```

---

### DIFF
--- a/cabal-install/main/Main.hs
+++ b/cabal-install/main/Main.hs
@@ -146,7 +146,7 @@ import Distribution.Simple.Program.Db (reconfigurePrograms)
 import qualified Distribution.Simple.Setup as Cabal
 import Distribution.Simple.Utils
          ( cabalVersion, die', dieNoVerbosity, info, notice, topHandler
-         , findPackageDesc, tryFindPackageDesc )
+         , findPackageDesc, tryFindPackageDesc, createDirectoryIfMissingVerbose )
 import Distribution.Text
          ( display )
 import Distribution.Verbosity as Verbosity
@@ -160,7 +160,8 @@ import System.FilePath          ( dropExtension, splitExtension
                                 , takeExtension, (</>), (<.>) )
 import System.IO                ( BufferMode(LineBuffering), hSetBuffering
                                 , stderr, stdout )
-import System.Directory         (doesFileExist, getCurrentDirectory)
+import System.Directory         ( doesFileExist, getCurrentDirectory
+                                , withCurrentDirectory)
 import Data.Monoid              (Any(..))
 import Control.Exception        (try)
 
@@ -907,10 +908,19 @@ unpackAction getFlags extraArgs globalFlags = do
   getAction getFlags extraArgs globalFlags
 
 initAction :: InitFlags -> [String] -> Action
-initAction initFlags extraArgs globalFlags
-    | not (null extraArgs) =
-      die' verbosity $ "'init' doesn't take any extra arguments: " ++ unwords extraArgs
-    | otherwise = do
+initAction initFlags extraArgs globalFlags = do
+  -- it takes the first value within extraArgs (if there's one)
+  -- and uses it as the root directory for the new project
+  case extraArgs of
+    [] -> initAction'
+    [projectDir] -> do
+      createDirectoryIfMissingVerbose verbosity True projectDir
+      withCurrentDirectory projectDir initAction'
+    _ -> die' verbosity $
+      "'init' only takes a single, optional, extra " ++
+      "argument for the project root directory"
+  where
+    initAction' = do
       confFlags <- loadConfigOrSandboxConfig verbosity globalFlags
       -- override with `--with-compiler` from CLI if available
       let confFlags' = savedConfigureFlags confFlags `mappend` compFlags
@@ -922,7 +932,7 @@ initAction initFlags extraArgs globalFlags
       withRepoContext verbosity globalFlags' $ \repoContext ->
         initCmd verbosity (configPackageDB' confFlags')
           repoContext comp progdb initFlags'
-  where
+
     verbosity = fromFlag (initVerbosity initFlags)
     compFlags = mempty { configHcPath = initHcPath initFlags }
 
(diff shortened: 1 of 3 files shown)


Mikolaj (approved):
LGTM.

Mikolaj:
@konsumlamm: would that help you as a user? How does it compare to what `stack` has?

konsumlamm:
Is there any reason we can't just have `cabal init my-project` instead of `cabal init -- my-project`? For reference, the syntax for stack is `stack new my-project`.

ptkato:
Oh, you can type it like that, iirc `--` exists only to avoid ambiguity in certain cases.

konsumlamm:
> Oh, you can type it like that, iirc `--` exists only to avoid ambiguity in certain cases.

Ah, if that's the case, I'm happy with the implementation. Thanks for working on this so fast!

How does documenting new features usually work, will there be a separate PR for that?

ptkato:
It's usually together with the feature PR, but I personally wait for it to be greenlit, so I don't document something that won't go forward.

cabal-install/main/Main.hs
```diff
@@ -907,10 +908,16 @@ unpackAction getFlags extraArgs globalFlags = do
...
 
 initAction :: InitFlags -> [String] -> Action
-initAction initFlags extraArgs globalFlags
-    | not (null extraArgs) =
-      die' verbosity $ "'init' doesn't take any extra arguments: " ++ unwords extraArgs
-    | otherwise = do
+initAction initFlags extraArgs globalFlags = do
+  -- it takes the first value within extraArgs (if there's one)
+  -- and uses it as the root directory for the new project
+  case nonEmpty extraArgs of
```
bacchanalia, line 914: what's the point of using nonEmpty hear instead of pattern matching on extraArgs directly?

cabal-install/main/Main.hs
```diff
@@ -907,10 +908,16 @@ unpackAction getFlags extraArgs globalFlags = do
...
+    Just (projectDir :| []) -> do
+      createDirectoryIfMissing True projectDir
+      withCurrentDirectory projectDir initAction'
+    _ -> initAction'
```
bacchanalia, line 918: This silently ignores all arguments if there isn't exactly one. Instead I think it would be better behavior for it to only init in the current directory on zero arguments, and die with an error on two or more arguments.

bacchanalia (commented):
I'm not a maintainer, but I took a quick look at the code and had a couple comments.

ptkato:
It's ok, I don't think you'd need to be a maintainer to submit feedback. :P

Those are nice finds, though; I'll fix them.

Mikolaj:
@konsumlamm: could you have one last look at the documentation?

konsumlamm:
Just a small comment, isn't `-n` the default now?

Mikolaj:
> Just a small comment, isn't `-n` the default now?

By "now" do you mean 3.6.2 or master branch? The defaults have swapped on master branch.

konsumlamm:
> > Just a small comment, isn't `-n` the default now?
> 
> By "now" do you mean 3.6.2 or master branch? The defaults have swapped on master branch.

I mean 3.6.2, yeah. So apparently the behaviour was changed back again.

bacchanalia (commented):
There is also help text for cabal init that needs to be changed in Distribution.Client.Setup

ptkato:
Oki, I believe we're all covered, yes?

Mikolaj:
> Oki, I believe we're all covered, yes?

I think so. yes. More than usually, thanks to live feedback from a user, and moreover a user who is the original ticket creator.

@bacchanalia, is your review positive now? If so, could you tick it?

Mikolaj:
Doh, I've linked from all over the world to this issue, by mistake (off-by-1). I'm afraid we are famous now. ;)

cabal-install/main/Main.hs
```diff
@@ -907,10 +908,19 @@ unpackAction getFlags extraArgs globalFlags = do
...
+  -- it takes the first value within extraArgs (if there's one)
+  -- and uses it as the root directory for the new project
+  case extraArgs of
+    projectDir : [] -> do
+      createDirectoryIfMissing True projectDir
+      withCurrentDirectory projectDir initAction'
+    _ : _ -> die' verbosity $
+      "'cabal init' only receives a single, optional," ++
+      "extra argument for the project root directory"
+    [] -> initAction'
```
bacchanalia, line 921: A few small suggestions, reorder cases, use createDirectoryIfMissingVerbose, formatting on the message to be more consistent with the rest of the file. Sorry for being a bit nitpicky.
```suggestion
  case extraArgs of
    [] -> initAction'
    [projectDir] -> do
      createDirectoryIfMissingVerbose verbosity True projectDir
      withCurrentDirectory projectDir initAction'
    _ -> die' verbosity $ "'init' only takes a single, optional,"
                       ++ "extra argument for the project root directory"
```

cabal-install/src/Distribution/Client/Setup.hs
```diff
@@ -2034,7 +2034,7 @@ initCommand = CommandUI {
       ++ "which will let you specify the options via flags and will use the defaults for the rest.\n",
     commandNotes = Nothing,
     commandUsage = \pname ->
-         "Usage: " ++ pname ++ " init [FLAGS]\n",
+         "Usage: " ++ pname ++ " init [PROJECT ROOT] [FLAGS]\n",
```
bacchanalia, line 2037: I'd change the above text from "Calling init with no arguments" to "Calling init with no flags", since we can now pass a non-flag argument. I think we should also add a sentence describing what [PROJECT ROOT] does. something like "Passing a project root will init in that directory instead of the current directory."

bacchanalia (approved):
Latest changes look good to me.

PR #7874, opened 2021-12-21
Rebase #5440: Respect ghc-options and with-gcc when compiling C sources

jberryman:
This is https://github.com/haskell/cabal/pull/5440 rebased on master

I don't understand the patch well enough to really help further right now. Please close if this isn't mergeable.

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -215,6 +215,8 @@
     ([#5431](https://github.com/haskell/cabal/pull/5431)).
   * Pass command line arguments to `hsc2hs` using response files when possible
     ([#3122](https://github.com/haskell/cabal/issues/3122)).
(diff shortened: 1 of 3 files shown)


Mikolaj:
@jberryman: thank you so much!

Any volunteer to add a simple test for that?

Mikolaj (approved):
LGTM, but I think we really need to test this. Ideally by adding a simple test. Trying out reproducing problems that people had that prompted this would would also be great. Perhaps we can ask them to try out the branch?

Mikolaj:
I did ask them. Let's wait some longer, given that it's Christmas and New Year.

Mikolaj:
@jberryman: thank you for #7900, which is now merged, which ends the saga! Please, everybody, test and reopen if anything doesn't work as expected.

PR #7882, opened 2021-12-28
OpenUnitId (fix typo in comments)

askeblad:


Mikolaj (approved):
Thank you.

fgaz (changes requested):
I think "lossly" may be correct here, meaning "with loss".

Rather, "represent it as an 'OpenUnitId uid'" should be changed to "represent it as an 'DefiniteUnitId uid'"

Mikolaj:
Oh, I couldn't make much sense of this comment either. What kind of loss exactly is that? Can we perhaps make this bit more precise so that it doesn't look like a typo?

askeblad:
@ezyang

In coding lossless has the correlate lossy.

Mikolaj:
So that would be https://en.wiktionary.org/wiki/lossily ? Isn't that too obscure? Let's reword.

askeblad:
@Mikolaj

https://gitlab.haskell.org/ghc/ghc/-/issues?label_name=backpack

[Wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/backpack)
[backpack](https://gitlab.haskell.org/ghc/ghc/-/wikis/backpack/history) · History

askeblad:
@bgamari and @simonpj appear to have modified the Wiki page

Mikolaj:
I'm afraid all three people you mentioned are terribly busy, so it's probably on us to understand and reword the blurb. Or, alternatively, you could just fix what @fgaz pointed out and move on. There is plenty of other low hanging fruit in cabal.

askeblad:
[Backpack documentation is lacking](https://gitlab.haskell.org/ghc/ghc/-/issues/17525) (GitLab issue)

I'll close if no response from the Wiki [page](https://gitlab.haskell.org/ghc/ghc/-/wikis/backpack) contributors is forthcoming.

PR #7900, opened 2022-01-08
Update #7874: Add test, fix changelog

sergv:
I have added test for #7874. It works by cabal package having a c file which will only compile if all the necessary defines are defined. I'm passing defines via various ways, including `cabal configure`, fields of cabal file and test that `gcc` is overridden by passing a custom script which defines final macro without which the c file will not compile.

The original patch seems to be working fine as without it the test doesn't pass. However, I noticed that the original patch only results in `cabal` passing `-pgmc` flag to `ghc`, thus overriding the C compiler. However original issue #4439 also mentioned the intention to override linker, which should have resulted in `-pgml` being passed to the `ghc`, but this doesn't happen in the original patch.

### DIFF
--- a/Cabal/ChangeLog.md
+++ b/Cabal/ChangeLog.md
@@ -215,8 +215,6 @@
     ([#5431](https://github.com/haskell/cabal/pull/5431)).
   * Pass command line arguments to `hsc2hs` using response files when possible
     ([#3122](https://github.com/haskell/cabal/issues/3122)).
-  * `ghc-options` and `--with-gcc` are now passed to GHC when compiling
-    C and C++ sources ([#4439](https://github.com/haskell/cabal/issues/4439)).
 
 ----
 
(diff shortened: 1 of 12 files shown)


Mikolaj:
Let me clarity: this PR provides a test, #7874 passed the test, and, based on the test, this PR slightly amends #7874 as well? If so, how exactly is #7874 amended? Is this reflected in the recorded test results?

sergv:
This PR takes originall commit from #7874, rebases it on master and adds test and changelog. The original commit is not changed.

Mikolaj:
OK, thanks, I got confused.

Mikolaj (approved):
Indeed, this matches the old PR and, while I'm not experience in this area, the test looks great. Thank you!

Kleidukos (approved):
Fantastic, thank you very much. :)

PR #7907, opened 2022-01-13
Gb/build with 9.2

gbaz:
build all now works with 9.2. For now, removing vendoring of cabal-doctest (not sure why it was vendored to begin with). If that's wrong let me know. There were no real comments why it was vendored. Also removing a few stray allow-newer clauses no longer necessary, apparently.


fgaz:
Nice! Could you add 9.0 and 9.2 CI jobs too?

Bodigrim:
> Amazing. That probably means @Bodigrim bumped all remaining hackage-security bounds on Hackage in advance of the release. And that's fine, because hackage-security CI passes.

Not me this time :) I think `resolv` was the last stumbling block and @gbaz just revised it.

phadej:
IIRC `cabal-doctest` was vendored because some dependency used it in its custom setup, and there was a change in `Cabal` which broke it.

If you checkout the commit it was added, and (`--dry`) build with `--index-state` of that date, you'll see that:

```
[Cabal-tests-3:test:"hackage-tests"]
 ├─ Cabal-tree-diff-3.5.0.0
 │   └─ tree-diff-0.1
 │       ├─ QuickCheck-2.14.1
 │       │   ├─ containers-0.5.6.2 ┄┄
 │       │   ├─ random-1.2.0 ┄┄
 │       │   └─ template-haskell-2.10.0.0 ┄┄
 │       ├─ aeson-1.5.4.1
 │       │   ├─ attoparsec-0.13.2.4
 │       │   │   ├─ scientific-0.3.6.2
 │       │   │   │   ├─ hashable-1.3.0.0 ┄┄
 │       │   │   │   ├─ integer-logarithms-1.0.3
 │       │   │   │   │   └─ array-0.5.1.0 ┄┄
 │       │   │   │   └─ primitive-0.7.1.0
 │       │   │   │       ├─ deepseq-1.4.1.1 ┄┄
 │       │   │   │       ├─ fail-4.9.0.0 ┄┄
 │       │   │   │       └─ transformers-0.4.2.0 ┄┄
 │       │   │   └─ semigroups-0.19.1 ┄┄
 │       │   ├─ base-compat-batteries-0.11.2
 │       │   │   ├─ base-compat-0.11.2
 │       │   │   │   └─ unix-2.7.1.0 ┄┄
 │       │   │   ├─ bifunctors-5.5.8
 │       │   │   │   └─ [comonad-5.0.6:all]
 │       │   │   │       ├─ comonad-5.0.6
 │       │   │   │       │   └─ [distributive-0.6.2:all]
 │       │   │   │       │       ├─ distributive-0.6.2
 │       │   │   │       │       │   ├─ base-orphans-0.8.3
 │       │   │   │       │       │   │   └─ base-4.8.2.0 ┄┄
 │       │   │   │       │       │   └─ semigroups-0.19.1 ┄┄
 │       │   │   │       │       └─ [setup]
 │       │   │   │       │           └─ cabal-doctest-1.0.8
 │       │   │   │       │               └─ Cabal-3.5.0.0 ┄┄
 │       │   │   │       └─ [setup]
 │       │   │   │           └─ cabal-doctest-1.0.8 ┄┄
```

Luckily `comonad` doesn't use `build-type: Custom` anymore.

phadej:
> > Amazing. That probably means @Bodigrim bumped all remaining hackage-security bounds on Hackage in advance of the release. And that's fine, because hackage-security CI passes.
> 
> Not me this time :) I think `resolv` was the last stumbling block and @gbaz just revised it.

I'm a co-maintainer of `resolv`, but didn't yet updated the Hackage maintainer field as there weren't need to.

I see a PR from 5 days ago, and now there is a revision. **and no mention that revision was made.**

gbaz:
You're right I should have mentioned the revision on the github. Anyway, I made a revision!

phadej:
Also `windns`. That doesn't have an issue or PR at all :(

fgaz:
@phadej could you push 8.10.7, 9.0.2, and 9.2.1 containers?

Mikolaj:
One or two jobs are already switched to ghcup, but these jobs are very different beasts, so this wouldn't carry over easily, I'm sure.

fgaz:
```
P.hs:1:8: error:
    Could not find module `Prelude'
    Perhaps you haven't installed the profiling libraries for package `base-4.13.0.0'?

[...]

*** unexpected failure for PackageTests/NewBuild/T3827/cabal.test.hs
```

huh. I guess that image does not have profiling libraries?

AlistairB:
> huh. I guess that image does not have profiling libraries?

Yup, they [are stripped out to save some space](https://github.com/haskell/docker-haskell/blob/master/9.2/buster/Dockerfile#L96). This was mimicked from the old haskell debian packages. I don't think this is a huge saving though, I'd be happy to put them back if there is a need.

Although the other thing to consider is the official images only keep the last 2-3 major versions of GHC up to date. So for example 8.8 still has cabal-install 3.4 and won't be updated.

gbaz:
What is the plan here? If we continue to be blocked on getting CI working with newer ghcs beyond a certain point, I would rather merge the fixes that allow cabal to build with newer ghcs, and then have a separate pr to actually give CI on those configurations.

fgaz:
@gbaz, @andreabedini is working on it in #7952

andreabedini:
My work on #7952 is going a bit slowly because each workflow run has to be approved by a maintainer, which means it might that days to test again after fixing a typo.

Is there any chance we can relax that constraint, even temporarily?

fgaz:
I just found and enabled a setting that does just that, you should be able to iterate quickly now. If that does not work, I believe you can test the workflows on your fork

andreabedini:
I'll review this soon but I believe off this changes have been incorporated into #7952. You can see a failed build [here](https://github.com/haskell/cabal/runs/5228622591?check_suite_focus=true#step:8:1428). @Mikolaj has already suggested a fix in https://github.com/haskell/cabal/pull/7952#issuecomment-1042693990.

jneira:
It seems this can be closed, no? we have succesful build with failing tests

Mikolaj:
Could somebody make a diff a verify? I believe this can be done both on commandline (or with gitk) and also using github.

gbaz:
No need. This branch was just intended to get things working. It wasn't special. If another branch has picked up enough changes to get things working, I'm happy :-)

andreabedini:
We still fail on 9.2 because of the overlapping patterns (see https://github.com/haskell/cabal/runs/5285920627?check_suite_focus=true#step:8:1574).

I will have a go at it today.

andreabedini:
FYI cabal now builds with GHC 9.2 but some haddock related tests fail because of #7987 this particular issue is truly done and dusted :)

PR #7912, opened 2022-01-20
Remove the failing and not needed Artifacts CI job

Mikolaj:


.github/workflows/artifacts.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-18.04
     container:
       # Older Ubuntu for older glibc
```
phadej, line 16: this comment is meaningful. artifcts build on bionic don't work on xenial.

Yet, IIRC artifacts are built by gitlab CI, aren't they; then the whole workflow could be removed. (as well as release.py which isn't used by that scripts).

Mikolaj: Yes, I understand the comment, but Xenial is just too old, not even the ppa supports it, it seems, hence the errors. So Focal is now the new "older glibc" container and it seems to work.

However, it's even better to remove things, let so let me try just that.

Mikolaj:
OK, full success. Only the breakage from broken network package persists. The breakage from python is gone. Please review and I will force-merge after a single review.

Kleidukos (approved):
love to see all this stuff going away. Godspeed!

Mikolaj:
@mergifyIO backport 3.6

Mikolaj:
The backport has merge conflicts, so it needs to be done manually. I'm not sure we'd be backporting anything to 3.6 branch any more, so let's procrastinate instead and fixup in one fell swoop, if needed.

PR #7917, opened 2022-01-20
Fix scripts without an .hs or .lhs extension

### LINKED ISSUE #7910: Scripts without an .hs or .lhs extension no longer work
**Describe the bug**
In cabal version 3.6.2.0 and before, scripts could be named anything. But with a cabal-install built from HEAD attempting to run any script file without an ".hs" or ".lhs" suffix fails.

**To Reproduce**
Save the following as a file named "script":

```
{- cabal:
build-depends: base
-}

main = putStrLn "Hello"
```

Then,

```
$ cabal --version
cabal-install version 3.7
compiled using version 3.7.0.0 of the Cabal library 

$ cabal v2-run script
Resolving dependencies...
Build profile: -w ghc-9.2.1 -O1
Error: cabal: The 'main-is' field must specify a '.hs' or '.lhs' file (even if
it is generated by a preprocessor), or it may specify a C/C++/obj-C source
file.
```

**Expected behavior**
Scripts should continue to work regardless of what they are named.

Although, it seems like this new behavior has fixed #5702 so error messages no longer confusingly mention "Main.hs". It would be nice to keep the new behavior but without this regression.

**System information**
 - cabal-install version 3.7 (built from commit 7775be851a4d661ab7e7d4e5efbc9268a8688975)
 - GHC version 9.2.1
 - NixOS 21.05

bacchanalia:
* Pass such scripts with: -x hs script
* Disable extension check for scripts
* Disable running preprocessors for scripts because they're based on
  extension
* Add test case for script with no extension

Resolves #7910


jneira:
Many thanks for the quick fix

> Disable running preprocessors for scripts because they're based on extension

What would be the practical implications? doesnt sound good at first

bacchanalia:
@jneira It just means that the script has to be a plain Haskell file, not `GreenCard, c2hs, hsc2hs, happy, alex and cpphs` (to quote from `Distribution.Simple.PreProcess`). This was also effectively the case before #7851, because the source file was copied to a `Main.hs` within the package. It would be possible to loosen this and allow those file types to be preprocessed but I'm not sure that's something useful to support. It would also be a change from previous behavior.

Cabal/src/Distribution/Simple/GHC.hs
```diff
@@ -1207,7 +1209,8 @@ gbuildSources verbosity specVer tmpDir bm =
       let mainModName = fromMaybe ModuleName.main $ exeMainModuleName exe
           otherModNames = exeModules exe
 
-      if isHaskell main
+      -- Scripts have fakePackageId and are always Haskell but can have any extension.
+      if isHaskell main || pkgId == fakePackageId
```
jneira, line 1213: the code is using fakePackageId to check if it is handling a script, but the fake package has other uses (cabal install iirc), could it interfere in some way?

bacchanalia: The only use of `fakePackageId` other than scripts is for the global (outside of a project context) repl, in which case there is no exe to be built. The only case where there is both an exe that is being built and `fakePackageId` is set is if we are running a script.

evidence: `fakePackageId` is only set in StriptUtil.hs in the function `fakeProjectSourcePackage`. `fakeProjectSourcePackage` is only used internally and in CmdRepl.hs

bacchanalia:
Scripts with spaces in the name are not being accepted. I think it makes sense to include the fix for that as part of this PR. Investigating now.

Mikolaj (approved):
LGTM

bacchanalia:
@Mikolaj should have waited for the tests to run

Mikolaj:
> @Mikolaj should have waited for the tests to run

I knew they would pass. Eventually.

cabal-install/src/Distribution/Client/TargetSelector.hs
```diff
@@ -815,6 +816,14 @@ reportTargetSelectorProblems verbosity problems = do
          ++ "packages in your project and all other build configuration. "
          ++ "See the Cabal user guide for full details."
 
+    case [ t | TargetSelectorNoScript t <- problems ] of
+      []  -> return ()
+      target:_ ->
+        die' verbosity $
+            "The script '" ++ showTargetString target ++ "' does not exist, "
+         ++ "and only script targets may contain whitespace characters or end "
+         ++ "with ':'"
```
jneira, line 825: what about a test with this corner case (spaces are already being tested)? it is easy to lose that corner case

bacchanalia: Spaces are already being tested. I can't easily include a test with a script that ends with ':' because files that end in ':' are not valid on windows, so putting one in the git repo is a no-go.

bacchanalia: I can probably check the OS is the cabal.test.hs and move it to end with ':' if the os is linux.

jneira: there is no a existing way to run tests only in a specific os (or equivalently ignore it in a specific os)?

jneira (approved):
lgtm, many thanks for fixing it so fast
i checked this in windows 7 with weird names like `sayñ¿Ç#$ä  2èáê` and worked fine :+1:

Issue #6322, opened 2019-10-31
cabal-install doesn't respect ^C


phadej:
**Describe the bug**

One have to ^C a lot to make cabal cancel. This is super annoying

**To Reproduce**

`cabal v2-build` something with plenty (fresh to build) dependencies. Try to ^C build right away. It won't. (on Linux).

**Expected behavior**

`cabal` should cancel promptly.

**System information**

- Operating System: Linux
- cabal-install-3.0.0.0

**Additional context**

E.g. when built through `make`, make seems to forward ^C, and then dies itself, leaving "downloading" cabal on the background. That's bad.

Feels like some missing `bracket` somewhere.

phadej:
```
^Cghc-pkg: interrupted
gi^Cghc-pkg: interrupted
^Cghc-pkg: interrupted
^Cghc-pkg: interrupted
^Cghc-pkg: interrupted
^C^C^Cghc-pkg: interrupted
```

23Skidoo:
I remember that we already had some code/tickets related to this, but don't remember where. @dcoutts?

robx:
I might have a fix for this in #7929, but it's a bit unclear if there's a different bug. I see the described behaviour, but only when there are (multiple) packages to be downloaded, and I try to interrupt during download. I've managed to reproduce that bug, and fix it in the PR. From the description and terminal log here it's unclear whether downloading was involved.

ulysses4ever:
@robx I think we should close it, do you agree? I just tried to Ctrl-C at various stages while building `lens` on Linux. It worked like a charm.

robx:
Yes, that sounds good to me -- this report is Linux-specific, and to the best of my knowledge the issue is fixed there. (I think the fix didn't make it into 3.8, but should be there in current master and 3.10.)

I'll close, of course feel free to reopen this if it resurfaces, or if we should wait for the 3.10 release before closing the issue.

ulysses4ever:
I was trying on 3.8 by the way. And  #7921 definitely got in 3.8.

robx:
> I was trying on 3.8 by the way. And #7921 definitely got in 3.8.

I think it's #7929 that fixed this, which did make it into 3.8.

(The primary effect of #7921 is to ensure that `kill`ing `cabal run` kills whatever `cabal run` is running as well as `cabal` itself.)

PR #7921, opened 2022-01-22
Handle SIGTERM properly, on unix systems

### LINKED ISSUE #7914: Cabal run leaves children running when killed
**Describe the bug**

When `cabal run` receives `SIGTERM`, it exits, but leaves its children running.

I'm not entirely sure whether this is buggy behaviour, but I was surprised by this,
in the context of running a Haskell tool via `cabal run` and [`withProcessTerm`](https://hackage.haskell.org/package/typed-process-0.2.8.0/docs/System-Process-Typed.html#v:withProcessTerm).

If I run `cabal run target` from the shell, and run
`kill <cabal-run-pid>`, I would have expected both `cabal` and `target`
to stop. Instead, `cabal` exits but `target` keeps running.

Note that when run interactively, `Ctrl-C` sends `SIGINT` to the whole
process group, which is how `cabal run` interrupts nicely on `Ctrl-C`.

**To Reproduce**

```
$ cat x.hs
{- cabal:
build-depends: base
-}

module Main where

import Control.Concurrent (threadDelay)
import Control.Exception
import Control.Monad (when)

main = loop `catch` logInterrupt
  where
    loop = do
      putStrLn "yo"
      threadDelay 1000000
      loop
    logInterrupt e = do
      when (e == UserInterrupt) $ putStrLn "ouch"
      throwIO e
```

Then call `cabal run x.hs`, in a different terminal:

```
$ ps | grep cabal
77565 ttys002    0:04.09 cabal run x.hs
77726 ttys002    0:00.01 /s/cabal/dist-newstyle/build/x86_64-osx/ghc-8.10.7/fake-package-0/x/script/build/script/script
$ kill 77565
```

Then the child keeps running although `cabal` exits:

```
yo
yo
yo
yo
Terminated: 15
$ yo
yo
yo
```

**Expected behavior**

I'd like the child process to be killed, too.

**System information**
- Operating system: macos
- cabal-install version 3.4.0.0, compiled using version 3.4.0.0 of the Cabal library 
- ghc 8.10.7

robx:
This make cabal-install reliably terminate its children before exiting when cabal-install itself is killed (fixes #7914), by:
- replacing use of [createProcess](https://hackage.haskell.org/package/process-1.6.9.0/docs/System-Process.html#v:createProcess) by [withCreateProcess](https://hackage.haskell.org/package/process-1.6.9.0/docs/System-Process.html#v:withCreateProcess), which provides async-exception-safe subprocess cleanup
- installing a signal handler which converts SIGTERM into an asynchronous exception that's thrown at the main thread

Some notes:
- The use of `withCreateProcess` helps with Ctrl-C behaviour on Windows, fixing the second half of #6322, as a follow-up to #7929.
- The `SIGTERM` change itself doesn't affect Windows -- there appears to be no meaningful way currently to make a similar change on Windows. (A previous version of this PR used the `signal` package to implement this in a cross-platform way, but that turns out to wrap a non-functional mingw API only.)
- `kill -TERM` of `cabal run` as illustrated in #7914 works correctly with this change on unix (macos); this is verified by test.
- There's a bug in `process` versions before 1.6.14 (https://github.com/haskell/process/pull/231) which causes
  spurious prints of  `cabal: waitForProcess: does not exist (No child processes)` when interrupting with Ctrl-C.
- It might be reasonable to use `installTerminationHandler` more widely, e.g. everywhere that `topHandler` is used.

---

### DIFF
--- a/Cabal/src/Distribution/Simple/Program/Run.hs
+++ b/Cabal/src/Distribution/Simple/Program/Run.hs
@@ -124,12 +124,10 @@ runProgramInvocation verbosity
   } = do
     pathOverride <- getExtraPathEnv envOverrides extraPath
     menv <- getEffectiveEnvironment (envOverrides ++ pathOverride)
-    exitCode <- rawSystemIOWithEnv verbosity
+    maybeExit $ rawSystemIOWithEnv verbosity
                                    path args
                                    mcwd menv
                                    Nothing Nothing Nothing
-    when (exitCode /= ExitSuccess) $
-      exitWith exitCode
 
 runProgramInvocation verbosity
   ProgramInvocation {
--- a/Cabal/src/Distribution/Simple/Utils.hs
+++ b/Cabal/src/Distribution/Simple/Utils.hs
@@ -789,9 +789,10 @@ rawSystemProcAction :: Verbosity -> Process.CreateProcess
                     -> IO (ExitCode, a)
 rawSystemProcAction verbosity cp action = withFrozenCallStack $ do
   logCommand verbosity cp
-  (mStdin, mStdout, mStderr, p) <- Process.createProcess cp
-  a <- action mStdin mStdout mStderr
-  exitcode <- Process.waitForProcess p
+  (exitcode, a) <- Process.withCreateProcess cp $ \mStdin mStdout mStderr p -> do
+    a <- action mStdin mStdout mStderr
+    exitcode <- Process.waitForProcess p
+    return (exitcode, a)
   unless (exitcode == ExitSuccess) $ do
     let cmd = case Process.cmdspec cp of
           Process.ShellCommand sh -> sh
--- a/cabal-install/cabal-install.cabal
+++ b/cabal-install/cabal-install.cabal
@@ -165,6 +165,7 @@ library
         Distribution.Client.Security.HTTP
         Distribution.Client.Setup
         Distribution.Client.SetupWrapper
+        Distribution.Client.Signal
         Distribution.Client.SolverInstallPlan
         Distribution.Client.SourceFiles
         Distribution.Client.SrcDist
--- a/cabal-install/main/Main.hs
+++ b/cabal-install/main/Main.hs
@@ -113,6 +113,8 @@ import Distribution.Client.Manpage            (manpageCmd)
 import Distribution.Client.ManpageFlags       (ManpageFlags (..))
 import Distribution.Client.Utils
          ( determineNumJobs, relaxEncodingErrors )
+import Distribution.Client.Signal
+         ( installTerminationHandler )
 import Distribution.Client.Version
          ( cabalInstallVersion )
 
@@ -170,6 +172,7 @@ import Control.Exception        (try)
 --
 main :: IO ()
 main = do
+  installTerminationHandler
   -- Enable line buffering so that we can get fast feedback even when piped.
   -- This is especially important for CI and build systems.
   hSetBuffering stdout LineBuffering
--- a/cabal-install/src/Distribution/Client/ProjectConfig.hs
+++ b/cabal-install/src/Distribution/Client/ProjectConfig.hs
@@ -120,7 +120,7 @@ import Distribution.Simple.InstallDirs
          ( PathTemplate, fromPathTemplate
          , toPathTemplate, substPathTemplate, initialPathTemplateEnv )
 import Distribution.Simple.Utils
-         ( die', warn, notice, info, createDirectoryIfMissingVerbose, rawSystemIOWithEnv )
+         ( die', warn, notice, info, createDirectoryIfMissingVerbose, maybeExit, rawSystemIOWithEnv )
 import Distribution.Client.Utils
          ( determineNumJobs )
 import Distribution.Utils.NubList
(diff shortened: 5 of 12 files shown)


hasufell:
cabal run and cabal exec should use `execSpawn` like stack does, see https://github.com/haskell/cabal/pull/7757#issuecomment-955179928 ...there's no reason to keep the parent process alive

robx:
I agree that it would make sense for `cabal run` and `cabal exec` to essentially `exec(3)` without forking. I think the change here is necessary regardless, though, because `cabal` shouldn't be leaving behind ghc processes or similar either when it is killed.

jneira:
@robx I could give a try to this on windows, what build step could be more appropriate?

robx:
> @robx I could give a try to this on windows, what build step could be more appropriate?

Thanks, that'd be great!

The ~~two~~ three things I'd try if I had access to a Windows system would be:
1. `cabal run` for some long-running executable, then kill it (but I'm not sure exactly what killing means on windows -- from process manager?), and see if that kills the child, too
2. `cabal build` along the lines of what you tested for #7929 (i.e. does interrupt work better now? I think it might)
3. `cabal build` again, but this time "kill" the process, and see if it exits completely

Just 1 would be great already. More detail there:

```
$ cat Main.hs 
import Control.Concurrent (threadDelay)

main = do
  putStrLn "whoop"
  threadDelay 500000
  main
$ cat RunKill.cabal 
name: RunKill
version: 1.0
build-type: Simple
cabal-version: >= 1.10

executable exe
  default-language: Haskell2010
  build-depends: base
  main-is: Main.hs
$ cat cabal.project 
packages: .
```

Then on unixy systems, what I'd do is:

- in one terminal, call `cabal run RunKill`
- in another, I'd call `ps | grep cabal`, which gives me two processes: `cabal run RunKill` (pid 1000) and something like `.../dist-newstyle/.../exe` (pid 2000)
- I'd call `kill 1000` in the second terminal, and the first prints `Terminated: 15` because `cabal` exits, but `exe` continues to run, printing `whoop` to the terminal, until I kill it explicitly with `kill 2000`

That's with cabal prior to this PR. After this PR, `kill 1000` is enough to kill both processes, and output to the terminal stops. Ideally, we'd see the same behaviour on Windows.

jneira:
> (but I'm not sure exactly what killing means on windows -- from process manager?)

we can do some things in the shell in windows too :-P `taskkill /pid` in cmd or `stop-process -pid` in powershell

thanks for the tips, will follow them

jneira:
Hi i ve got to build this pr and test it:

1. Unfortunately killing the main cabal process did not kill the child one in my windows 7

```
D:\ws>tasklist | grep cabal
cabal.exe                     4244 Console                    1    61.516 KB
cabal-script-Delay.hs.exe     8112 Console                    1     4.088 KB

D:\ws>taskkill /pid 4244 /F
Correcto: se terminó el proceso con PID 4244.

D:\ws>tasklist | grep cabal
cabal-script-Delay.hs.exe     8112 Console                    1     4.184 KB
```

In the shell running the script there is

```
whoop
whoop
whoop

D:\ws\haskell\cabal>whoop  # <------------ here the cabal kill
whoop
whoop
```

jneira:
2. Good news, hitting ctrl+c has killed cabal immediately in all my tests in several steps of the build: downloading, building, etc

Sometimes it prints `AsyncCancelled` and sometims it does not though

3. killing the cabal process while building has worked fine in all my tests, trying to trigger it in different phases of the build (with ghc/cc/ld spawned)

robx:
Thanks a lot! I've done a bit of digging, and it seems that signals as such don't exist meaningfully on Windows / mingw. Best reference I found is https://sourceforge.net/p/mingw/mailman/message/16209215/.

The `signal` library just appears to wrap mingw's `signal.h` https://github.com/jonasstrandstedt/MinGW/blob/master/MinGW/include/signal.h (not sure if that's the real MinGW source):

> `SIGTERM` comes from what kind of termination request exactly?

I've also checked, and the "can only be terminated forcefully" thing applies to any Haskell processes, so seems like the PR doesn't break anything, either.

I think what I'll do is

1. drop the `signal` dependency, and `#ifdef notwindows` the new signal handling code, using posix instead
2. raise an issue with the `signal` package to document that it doesn't really do anything useful (or give the author a chance to clarify my misunderstanding)

Anyway, effect should be that we just get the better Ctrl-C exit on Windows, and the SIGTERM fix on Unix only.

robx:
(There's probably something the ghc windows runtime could do to make it respond to `taskkill` without `/f`...)

hasufell:
> cabal run and cabal exec should use `execSpawn` like stack does, see [#7757 (comment)](https://github.com/haskell/cabal/pull/7757#issuecomment-955179928) ...there's no reason to keep the parent process alive

Will this be addressed?

robx:
> > cabal run and cabal exec should use `execSpawn` like stack does, see [#7757 (comment)](https://github.com/haskell/cabal/pull/7757#issuecomment-955179928) ...there's no reason to keep the parent process alive
> 
> Will this be addressed?

I didn't intend to address this, at least not in this PR, since it seems orthogonal to the change as is.

Mikolaj (approved):
Thanks a lot.

jneira:
@Mergifyio rebase master

jneira (commented):
lgtm, thanks!

jneira (approved):

hasufell:
The cabal pre-release now includes this change and I'm getting a lot of this lately when cancelling builds:

```
cabal-3.8.0.20220526: waitForProcess: does not exist (No child process)
```

Pretty sure it's this patchset.

Mikolaj:
That's on LInux. Is the problem only the alarming message or is there a visible delay or are zombies created or whatever?

hasufell:
Oh no. So I was wrong.

I guess we can set a constraint in cabal.project.

PR #7922, opened 2022-01-22
Raise process lower bound to 1.2.1.0

robx:
I ran into these shims while working on #7921. Do I interpret the "support window" info right in that we can do this?

---


robx:
Oh I see this duplicates a tiny amount of the work done in https://github.com/haskell/cabal/pull/7534. @emilypi would you prefer I close this one as duplicate?

Mikolaj:
Great find re #7534, That confirms the "support window" permits this upgrade.

Cabal/Cabal.cabal
```diff
@@ -96,7 +96,6 @@ library
     Distribution.Utils.MD5
     Distribution.Utils.String
     Distribution.Utils.Structured
-    Distribution.Compat.CreatePipe
```
robx, line 99: We could keep `Distribution.Compat.CreatePipe` as a trivial reexport to avoid requiring library users to make changes here if they were using the module. And/or I guess this module going away should be documented explicitly in the Changelog?

Mikolaj: If it's so cheap, I'd say let's keep the compat. The PR breaks (legally) enough stuff.

Cabal/src/Distribution/Compat/CreatePipe.hs
```diff
@@ -1,81 +1,5 @@
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE RankNTypes #-}
-
 module Distribution.Compat.CreatePipe (createPipe) where
```
andreasabel, line 1: Should this module get a deprecation warning, maybe?

robx: Done.

Mikolaj (approved):

andreasabel (approved):

Mikolaj:
@robx: would you like to merge? If so, add the merge_me label (it rebases, etc.). Thank you for the PR!

Issue #7774, opened 2021-10-25
list-bin is missing from user's manual


Bodigrim:
`list-bin` command, added in 3.4, is still missing from user's manual as of 3.6: https://cabal.readthedocs.io/en/3.6/cabal-commands.html.

Mikolaj:
@ philderbeast: would you like to turn that summary into a PR? If so, don't worry about how it fits the existing docs too much --- create a new section at a likely place and dump the knowledge.

philderbeast:
> @ philderbeast: would you like to turn that summary into a PR? If so, don't worry about how it fits the existing docs too much --- create a new section at a likely place and dump the knowledge.

Yes, happy to.

Mikolaj:
Splendid. Thank you.

Issue #7073, opened 2020-09-15
Ability to build a cabal script into executable


ethercrow:
When I have a haskell program in a form of a single file script like

```
#!/usr/bin/env cabal
{- cabal:
build-depends: base, lens, shake, ...
-}
main = putStrLn "All done."
```

I'd like to be able to build it into a standalone executable.

Currently it seems like you can only build something with `cabal v2-build` if there is a cabal file and/or a project file, but building single file scripts is not supported:

```
> cabal v2-build Main.hs
No cabal.project file or cabal file matching the default glob './*.cabal' was found.
Please create a package description file ...
```

Creating a package description file is exactly what I'd like to avoid for small programs, since all the necessary information is already contained in the source file.

jneira:
In fact cabal generates an executable when you does a `cabal v2-run script.hs`, it is under dist-newstyle\build\x86_64-windows\ghc-8x.x.x\fake-package-0\x\script\build\script\script.exe in my windows system (the precise path can be different in other os). Maybe it would not be a final solution but maybe a working workaround.

phadej:
@jneira as if I remembered that `cabal script` should just use `runghc`. Whether it should first compiler and run executable, or just interpret is better left to the choice of user. (I think that interpreting is more reasonable default for scripts).

bacchanalia:
I think #7851 resolved this issue.

Mikolaj:
It seems so. @ethercrow (and other interested parties), could you kindly confirm if it solves your use case? Ideally by compiling the master branch and trying it out.

georgefst:
#7851 certainly improves the situation. But there's no easy way to find the resulting binary. It would be great if `list-bin` worked for scripts (and/or `install`, but I've always had issues with that one).

bacchanalia:
There are so many cabal sub commands I didn't even realize `list-bin` was a thing. Adding `list-bin` support for scripts seems reasonable. I think installing a script going against the spirit of a script, though.

georgefst:
> There are so many cabal sub commands I didn't even realize `list-bin` was a thing.

Not unreasonably: #7774.

> Adding `list-bin` support for scripts seems reasonable. I think installing a script going against the spirit of a script, though.

Well, if you have `list-bin` then I was thinking `cabal install Script.hs` is just a shortcut for (roughly) `cabal build Script.hs && cabal list-bin Script.hs | cp ~/.local/bin`. But I seem to remember that `install` isn't defined like that, and I've never understood why.

PR #7925, opened 2022-01-23
Add list-bin support for scripts.

bacchanalia:
Small PR to add script support for list-bin. It includes a test.
Because of #7774, I do not  believe there is any documentation that needs to be updated.

Related: #7073

### DIFF
--- a/cabal-testsuite/PackageTests/ListBin/Script/cabal.test.hs
+++ b/cabal-testsuite/PackageTests/ListBin/Script/cabal.test.hs
@@ -1,0 +1,8 @@
+import Test.Cabal.Prelude
+
+main = cabalTest . void $ do
+    res <- cabal' "list-bin" ["script.hs"]
+
+    env      <- getTestEnv
+    cacheDir <- getScriptCacheDirectory $ testCurrentDir env </> "script.hs"
+    assertOutputContains cacheDir res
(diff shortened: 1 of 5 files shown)


Mikolaj (approved):
Splendid.

jneira (approved):
perfect companion of your main pr improving cabal script, including a regression test, thanks!

andreasabel:
@bacchanalia wrote:
> Small PR to add script support for list-bin. It includes a test. 

I checked the [test output](https://github.com/haskell/cabal/pull/7925/files#diff-3d6637ed4ab2ee658f06eaa968642fc10f891f059a167e3a927f1908a623b1fd) but it does not contain the path to the binary.  I'd expect that `list-bin` produces this path.  Is everything working correctly?

> Because of #7774, I do not believe there is any documentation that needs to be updated.

There is [documentation of `list-bin`](https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-list-bin) now, but does not include your use case with scripts.  This is
- #7986

Would you be able to supply this documentation now to close #7986?

Mikolaj:
Indeed I missed the test output problem and the above suggestions make sense to me.

cabal-testsuite/PackageTests/ListBin/Script/cabal.test.hs
```diff
@@ -0,0 +1,8 @@
+import Test.Cabal.Prelude
+
+main = cabalTest . void $ do
+    res <- cabal' "list-bin" ["script.hs"]
+
+    env      <- getTestEnv
+    cacheDir <- getScriptCacheDirectory $ testCurrentDir env </> "script.hs"
+    assertOutputContains cacheDir res
```
andreasabel, line 8: @bacchanalia : I don't understand how this `assertOutputContains` succeeds with above output (that does not contain something resembling a `cacheDir`).  But apparently it does.  How?

PR #7928, opened 2022-01-25
docs: typo

ulysses4ever:
Unless my English is severely lacking, this seems like a tiny typo in the docs.

### DIFF
--- a/doc/nix-local-build.rst
+++ b/doc/nix-local-build.rst
@@ -211,7 +211,7 @@ build products like executables.
 Caching
 -------
 
-Nix-style local builds sport a robust caching system which helps to reduce
+Nix-style local builds support a robust caching system which helps to reduce
 the time it takes to execute a rebuild cycle. While the details of how
 ``cabal-install`` does caching are an implementation detail and may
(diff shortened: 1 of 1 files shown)


jneira (approved):
good catch, many thanks!

Mikolaj:
That's quite a frivolous language, but I think it's intended: https://en.wiktionary.org/wiki/sport#Verb

ulysses4ever:
Oh, I had a hunch and should have tried the dictionary. Sorry for the noise and feel free to close it then. Unless you want to stay closer to "simple English" :-) (that's up to you of course).

Mikolaj:
Simple English has it's merits and it was written 2 years ago, so we won't be disrespecting the contributor who graciously contributed the fixes, so that's an option. 

@jneira:  what do you think?

jneira:
it is a good opportunity to know about less used definitions but support is extensively used in tech writings so I would change it anyways

Mikolaj (approved):
We've had fun for 2 years. Now let's refresh and simplify.

PR #7929, opened 2022-01-25
Fix concurrency/exception bugs in asyncFetchPackages

### LINKED ISSUE #6322: cabal-install doesn't respect ^C

robx:
This fixes Ctrl-C not interrupting cabal properly while it is downloading. I believe
that might be #6322, although it would be good to verify that.

There are ~~two~~ three concrete bugs:
- `try` catches asynchronous exceptions, preventing `withAsync` from interrupting it when its `body` action finishes prematurely.
- the use of `withAsync` is not safe since we don't wait for the async and hence lose exceptions it might throw
- the use of regular `bracket` in `withTempFileName` can "convert" `AsyncCancelled` into a regular IO exception

The first bug manifests during `cabal build` when hitting Ctrl-C during
download, e.g.:

1. Ctrl-C interrupts a curl process
2. This interrupt is converted into a UserInterrupt exception
   via the process package's delegate_ctlc functionality.
3. UserInterrupt bubbles up through `fetchPackage`, is caught by
   the `try` (fine) and writen to the result mvar.
   Meanwhile, `fetchPackage` continues its loop, starting to
   fetch the next package.
4. Some `rebuildTarget` is waiting for the interrupted download;
  `waitAsyncFetchPackage` rethrows UserInterrupt in that thread.
5. UserInterrupt bubbles up through `collectJob`, `execute`, the
   `body` action.
6. `withAsync`'s finalizer cancels the `fetchPackages` async.
7. The `fetchPackages` async receives the AsyncCancelled exception
   while fetching the next package (see 3. above). This interrupts
   the download action (it shouldn't matter whether we happen to
   be running curl again or not), and AsyncCancelled bubbles up
   through `fetchPackage`, is caught by the `try` (not fine!) and
   written to the mvar. But no-one is reading that mvar anymore
   because the build job already aborted (step 5), and that
   AsyncCancelled exception is lost.
8. `fetchPackages` keeps doing its thing, exiting eventually.

To fix this, use `try` from safe-exceptions, which doesn't catch
asynchronous exceptions. Note that this change affects both
instances of `try` in this story: `UserInterrupt` is also an asynchronous
exception, so after the change the `UserInterrupt` takes a different path
from step 3.

This relates to the second bug, uncovered during testing of the fix for
the first bug: `UserInterrupt` can now cause the async part of
`withAsync` to crash. That one is fixed by using `concurrently` instead
of withAsync.

Finally, the third bug showed on Windows CI during testing, where temporary
file cleanup trigged by async cancellation fails with an IO exception, which is
then treated as a regular download failure, while the attempt to cancel the
async is ignored.

- To verify manually, call `cabal build` for a package with several undownloaded
  dependencies, and hit Ctrl-C when it gets started. Prior to these changes, `cabal`
  doesn't exit until it's done downloading, afterwards it does. I've checked this
  on macOS.
- The PR includes unit tests for `asyncFetchPackages` with a faked HTTP transport,
  simulating several failure scenarios where the `asyncFetchPackages` might not exit
  promptly or even deadlock. Some of these fail before the change, and now pass.

---

### DIFF
--- a/bootstrap/linux-8.10.7.json
+++ b/bootstrap/linux-8.10.7.json
@@ -255,6 +255,15 @@
             "version": "0.1.2.0"
         },
         {
+            "cabal_sha256": "8bc9cd9991863a238b3531dfc663f262016adbbd814f30b1c63a6ce914ff7906",
+            "revision": 0,
+            "src_sha256": "69637f794146a8e7bfbc2db2bd0501c274ec99504b597728e203187790064895",
+            "flags": [],
+            "package": "safe-exceptions",
+            "source": "hackage",
+            "version": "0.1.7.2"
+        },
+        {
             "cabal_sha256": "b83dec34a53520de84c6dd3dc7aae45d22409b46eb471c478b98108215a370f0",
             "revision": 1,
             "src_sha256": "484df85be0e76c4fed9376451e48e1d0c6e97952ce79735b72d54297e7e0a725",
--- a/bootstrap/linux-8.6.5.json
+++ b/bootstrap/linux-8.6.5.json
@@ -255,17 +255,6 @@
             "version": "0.1.2.0"
         },
         {
-            "cabal_sha256": "b83dec34a53520de84c6dd3dc7aae45d22409b46eb471c478b98108215a370f0",
-            "revision": 1,
-            "src_sha256": "484df85be0e76c4fed9376451e48e1d0c6e97952ce79735b72d54297e7e0a725",
-            "flags": [
-                "-bench"
-            ],
-            "package": "async",
-            "source": "hackage",
-            "version": "2.2.4"
-        },
-        {
             "cabal_sha256": "7ed09aed03683d5b4337088061106c2389d274b3472031a330ff1b220bad2b2d",
             "revision": 3,
             "src_sha256": "4d0bfb4355cffcd67d300811df9d5fe44ea3594ed63750795bfc1f797abd84cf",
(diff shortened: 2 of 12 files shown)


robx:
This is verified locally (on macos, with what appears to be the `curl` transport).

This is a bit hard to cover with tests: What we need is to hit `cabal build` with Ctrl-C during the download phase. Note that `kill -INT` is not enough, Ctrl-C in a terminal hits the whole process group, which is what cabal needs for its delegate_ctlc stuff to kick in. Feels a bit too hard for automated tests.

robx:
I'm putting some finishing touches on these commits, then I think this should be ready for review.

Need some help with the bootstrap update though, since I don't have a Linux environment ready.
(Incidentally, should those `.plan.json` files be checked in? They seem transient and contain info about the committer's local dev environment.)

Mikolaj:
Great job. Re bootstrap files, I cargo-culted them recently (IIRC you just run a script), but I have no idea how they work. Please ask about them on #hackage IRC/Matrix --- I'd also like to learn.

robx:
Any idea who would be a good person to ask to review this? (And is there some way I can make it easier to review?)

jneira:
I suppose it should be tested in windows, in addition to Linux, will try if I have time but don't wait for me

Mikolaj:
We have waited so long, we can wait a little longer.

jneira:
i've tried in windows and it has responded pretty well to ctrl+c, i had only to press it more than one time:

```
Downloading  fourmolu-0.3.0.0   <--------------------------------------------------- first ctrl+c
Warning: ghc-lib-parser.cabal:149:5: The field "autogen-modules" is available
only since the Cabal specification version 2.0. This field will be ignored.
Warning: ghc-lib-parser.cabal:1:22: Packages with 'cabal-version: 1.12' or
later should specify a specific version of the Cabal spec of the form <------------- again here i think
'cabal-version: x.y'. Use 'cabal-version: 1.22'.
Configuring library for ghc-lib-parser-8.10.7.20210828..
Downloaded   fourmolu-0.3.0.0
Downloading  ormolu-0.1.4.1
Downloaded   ormolu-0.1.4.1
Downloading  stylish-haskell-0.13.0.0
AsyncCancelled <----------------------------------------------------------------------- here it made effect
```

But it is infinitely better than now for sure, great job

jneira (approved):
Thank you

robx:
Thanks for testing! Unfortunately I don't have access to Windows, would love to dig in to what's happening there. I *hope* that using `withCreateProcess` might help there, which I'm planning to do anyway (compare https://github.com/haskell/cabal/pull/7921).

jneira:
@MergifyIO rebase master

hasufell:
amazing

Mikolaj:
CI jobs are stuck, so let me merge manually.

Mikolaj:
https://github.com/Mergifyio rebase

Mikolaj:
Mergifyio rebase

Mikolaj:
@Mergifyio rebase

Issue #7930, opened 2022-01-25
docs: cabal.project search path is not documented


ulysses4ever:
My understanding is that cabal.project affects a build even if the file lives in a parent directory, not only the current directory. Maybe this is well-known, but no for me, and I've been experiencing spooky actions at a distance for a little while. The closest the docs get to is (if I'm not mistaken):

> The full configuration of a project is determined by combining the following sources (later entries override earlier ones, except for appendable options):
> 
> -    ~/.cabal/config (the user-wide global configuration)
> 
> -    cabal.project (the project configuration)
> 
> -    cabal.project.freeze (the output of cabal v2-freeze)

Which, if anything, suggests that the global config file and the two from the current directory affect the build. At least, having an abslute path in the first bullet point indicates to me that the rest are probably relative paths.

If you agree that it's an omission, I could supply some kind of amendment to the above text.

Mikolaj:
Related: #7695

Yes, we need better and consistent docs for that: in code, readthedocs, commandline help, if appropriate.

andreasabel:
@ulysses4ever 
> My understanding is that cabal.project affects a build even if the file lives in a parent directory

This sounds like a misfeature to me.  Can we get rid of it?  What motivated it in the first place?

Mikolaj:
I guess, cabal needs context to build a package and that context is provided by a project. If that package is not part of any project, some default context is assumed. Project can be determined by files in parent directories as well as by commandline overrides. An example of the contest info is the location of some package dependencies (that, e.g., are not on Hackage at all).

I guess one may ask why in such cases packages are not built by staying in the directory where the project file is. I don't know --- perhaps cabal `cd`s to subdirectories where it builds things, e.g., because tools it calls assume they are run in the same directory as the source?

ulysses4ever:
It's probably clear from my original message that I also find this feature confusing. The argument for it I  got from someone is that many build tools work this way today (I think `bazel` does); also, `git`.

fgaz:
I think the feature can be useful, though I don't use it often. Like with git, it's nice to not have to cd back to the root of the project when invoking a command in a subdirectory.

What kind of confusing behaviour did you experience?

ulysses4ever:
I had a completely unrelated `cabal.project` file hanging up the directory, which changed the way the current package builds in a subtle way.

ulysses4ever:
I guess, ideally I'd expect `cabal-install` messaging me about loading a `cabal.project` file.

[...]

PR #7931, opened 2022-01-26
docs: cabal.project in parent directories

ulysses4ever:
This is a stop-gap measure against #7930. It's probably not ideal so I'm fine with just waiting for a better feature, if you think it's the right way forward.

### DIFF
--- a/doc/cabal-project.rst
+++ b/doc/cabal-project.rst
@@ -32,7 +32,7 @@ options):
 
 4. ``cabal.project.local`` (the output of ``cabal v2-configure``)
 
-A call to ``cabal build`` will consider ``cabal.project*`` from parent 
+Any call to ``cabal build`` will consider ``cabal.project*`` files from parent 
 directories when there is none in the current directory.
 
 Specifying the local packages
--- a/doc/cabal-project.rst
+++ b/doc/cabal-project.rst
@@ -32,6 +32,8 @@ options):
 
 4. ``cabal.project.local`` (the output of ``cabal v2-configure``)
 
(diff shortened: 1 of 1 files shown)


Mikolaj (approved):
That's definitely progress.

doc/cabal-project.rst
```diff
@@ -32,6 +32,8 @@ options):
 
 4. ``cabal.project.local`` (the output of ``cabal v2-configure``)
 
+A call to ``cabal build`` will consider ``cabal.project*`` from parent
```
andreasabel, line 35: Just some wording suggestion:
_Any_ call to ... will consider ... _files_ from parent ...

andreasabel (commented):
I'd like to point out that _parent directory_ is ambiguous when symlinks are involved.
If I have a directory `a/d` and a link `b/d -> a/d` and I am sitting in `b/d` then both `a` and `b` can be parent directories depending on whether the current directory is canonicalized (becoming `a/d` or not staying `b/d`).  Also _going up_ is ambiguous and OS-specific when symlinks are involved.  
(We discussed this in Agda in the context of project files, but atm I cannot find the discussion in the issues/PRs...)

I don't think this PR should be blocked because of this ambiguity, but an issue should be opened to point out the problem with a TODO that the documentation needs to be clarified.

doc/cabal-project.rst
ulysses4ever: Done, thanks!

Kleidukos (approved):

PR #7948, opened 2022-02-02
Remove dead module Distribution.Client.Compat.FilePerms

bacchanalia:
Follow up to #7686

`Distribution.Client.Compat.FilePerms` is imported nowhere and duplicates functionality from `Distribution.Compat.CopyFile`. It provides a single function function, `setFileHidden`, but again this function is used no where in the code base. If we want to preserve it, I suggest moving it into `Distribution.Compat.CopyFile`.

This PR would remove `Distribution.Client.Compat.FilePerms` as dead code.


Mikolaj (approved):
LGTM

robx (approved):
(Let's see if my review counts?)

PR #7949, opened 2022-02-02
Simplify bootstrap generation

robx:
There doesn't seem to be a particular point in checking in the `.plan.json` files,
and generation appears to work just fine on macOS.

If desired, we could check in per-GHC-version freeze files to make the bootstrap
generation more reproducible, but... seems fine without?

The regeneration commit could be left out; I included it to show that the diff is
largely trivial (there's a bit of reordering).

I've validated this change via PR #7929.

---


Mikolaj (changes requested):
I really don't know how bootstrapping works, but any simplification is welcome.

Could you add the mini-changelog file?

robx:
Sure, will do! (Thought the change maybe wasn't relevant to users.)

Mikolaj:
I'm thinking about "users" that actually use the bootstrap mechanisms to build cabal on new setups. I think it's fine to remove `plan.json` files if you think they are unneeded, but in case the users are surprised by the lack of the files, let's mark it in the changelog.

Mikolaj (approved):
Great. If the users need the .plan files for anything, they now know where to find us.

Kleidukos (approved):

fgaz:
> and generation appears to work just fine on macOS.

the plans could be different though, couldn't they?

if we extend bootstrapping to macOS maybe we should provide separate jsons

PR #7952, opened 2022-02-04
Avoid templating GitHub Actions workflow

andreabedini:
GitHub Actions workflow have sufficient power to express what we need.
We don't need to maintain and additional templating solution on top.

This converts: `linux.yml`, `macos.yml` and `windows.yml` to native GitHub Actions syntax. ~The result should closely (but perhaps not perfectly) match the original version.~

EDIT:

~It would be better to check GHC versions with `startsWith(matrix.ghc, "8.10")` otherwise bumping minor versions is going to be error prone~ I am in

---


Mikolaj:
@fgaz is just extending our CI to GHC 9 and 9.2 and converting it in the process in #7907, so it may be beneficial to coordinate...

andreabedini:
Thanks @Mikolaj, I think I was able to include most of the changes of #7907. Happy to coordinate to finish the job.

fgaz:
Hmm.. why isn't the linux job running?

.github/workflows/linux.yml
```diff
@@ -15,211 +10,36 @@ on:
...
+          - "8.6.5"
+          - "8.4.4"
+          - "8.2.2"
+          - "8.0.2"
+          - "7.10.3"
+          - "7.8.4"
+
+    name: validate ${{ matrix.ghc }}
     runs-on: ubuntu-18.04
-    needs: validate-8_8_4
```
fgaz, line 212: This got lost, right? Is there a way to restore it? We do have high parallelism, but still, running n jobs for n ghcs before checking that the thing actually builds seems wasteful

cabal-dev-scripts/src/GenValidate.hs
```diff
@@ -1,159 +0,0 @@
...
-                    ]
-                , zWinJobs =
-                    [ mkWinGhcJob "8.6.5"  Nothing           []
-                    , mkWinGhcJob "8.10.4" (Just "8.10.4")   []
-                    ]
-                , zMangleVersion = map mangleChar
-                , zOr            = (||)
-                , zNot           = not
-                , zLibOnly       = (== "--lib-only")
-                }
```
fgaz, line 55: Having every parameter (version, flags...) in a single place helped a lot when updating stuff... is there a way to do something similar natively?

.github/workflows/linux.yml
andreabedini: well, yes and no, I wasn't sure the intent of the force serialisation so I left it. We can implement either a fail-fast or a partial serialisation depending on our needs (see [docos](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs)).

cabal-dev-scripts/src/GenValidate.hs
andreabedini: I tried different approaches. Typically one uses a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs) but we seem to be cherry-picking desparate configurations without a clear meaning here.

A bit extreme but one could also build the matrix entierly manually by listing its entries (see [this](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs#using-environment-variables-in-a-matrix) example, nevermind the use of environment variables). This would be close to what we do currently.

Also, I haven't yet opened `validate.sh` do see what it does, done that I can try to unpack and simplify this configuration.

andreabedini:
> Hmm.. why isn't the linux job running?

/shurg I don't know. I am working to be able to test this out locally (with [act](https://github.com/nektos/act)).

I will review the failures on Monday.

fgaz:
> > Hmm.. why isn't the linux job running?
>
> /shurg I don't know. I am working to be able to test this out locally (with [act](https://github.com/nektos/act)).

Looks like the workflow is invalid https://github.com/haskell/cabal/actions/runs/1794333751

andreabedini:
Rebased

andreabedini:
The workflows for macos and linux at this point are very similar and could be merged. They only differ in how they install `cabal-plan` (one downloading a binary, one installing from hackage)

templates/ci-linux.template.yml
```diff
@@ -1,73 +0,0 @@
...
-  push:
-    branches:
-      - master
-  pull_request:
-  release:
-    types:
-      - created
-
-jobs:
-{% for job in jobs %}
```
jneira, line 18: i wonder why a matrix was not used in the first place, were they not available since the beginning of gha?

.github/workflows/macos.yml
```diff
@@ -15,108 +10,63 @@ on:
...
+      matrix:
+        ghc:
+          - "9.2.1"
+          - "9.0.2"
+          - "8.10.7"
+          - "8.8.4"
+          - "8.6.5"
+          - "8.2.2"
+          # - "8.0.2" disabled because no-cli suite
+          # TODO merge with linux.yml and use the same logic for both
```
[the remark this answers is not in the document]
fgaz: > maybe we could do in this pr or is it asking for too much?

let's merge this first, so we have 9.2 support

templates/ci-linux.template.yml
andreabedini: No idea, I believe they were available.

.github/workflows/macos.yml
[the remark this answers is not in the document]
andreabedini: at this stage merging linux and mac is a no brainer but for sake of delivery I need to get this out ASAP. This means all the tests we expect to pass need to pass. I am a bit confused as how to tell GH that some combinations can fail. Once I fix that, I will be able to see if tests pass for GHC <9
It won't take too long I promise!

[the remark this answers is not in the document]
fgaz: No rush, afaik this isn't really blocking anything else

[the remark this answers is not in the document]
andreabedini: my long term plan would be 1) remove the validate script 2) unify linux, macos *and windows*. This likely involves rethinking the meaning of all those flags, so I will defer it to another PR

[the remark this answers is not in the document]
jneira: > at this stage merging linux and mac is a no brainer but for sake of delivery I need to get this out ASAP. This means all the tests we expect to pass need to pass. I am a bit confused as how to tell GH that some combinations can fail. Once I fix that, I will be able to see if tests pass for GHC <9 It won't take too long I promise!

continue-on-error?

andreabedini:
Current status:

**GHC 9.2.1**

Both [Linux](https://github.com/haskell/cabal/runs/5178990898?check_suite_focus=true) and [MacOS](https://github.com/haskell/cabal/runs/5178990705?check_suite_focus=true) fail because of `StarIsType`. This is good because it's what we wanted to test.

**GHC 9.0.2**

some unexpected failures.

[Linux](https://github.com/haskell/cabal/runs/5178990935?check_suite_focus=true)
* PackageTests/Backpack/Includes2/setup-external.test.hs
* PackageTests/Backpack/Includes3/setup-external-ok.test.hs
* PackageTests/Backpack/Includes2/setup-per-component.test.hs

[MacOS](https://github.com/haskell/cabal/runs/5178990727?check_suite_focus=true)
* PackageTests/Backpack/Includes2/setup-external.test.hs
* PackageTests/Backpack/Includes3/setup-external-ok.test.hs
* PackageTests/Backpack/Includes2/setup-per-component.test.hs
* PackageTests/Regression/T4025/setup.test.hs

**GHC 8.10.7**

[MacOS](https://github.com/haskell/cabal/runs/5178990763?check_suite_focus=true) has one unexpected failure 
* PackageTests/Regression/T4025/setup.test.hs

Notice, previously we were not testing this combination (GHC 8.10 on MacOS)

**GHC 8.0.2**

[MacOS](https://github.com/haskell/cabal/runs/5178990918?check_suite_focus=true) has one unexpected failure
* PackageTests/ForeignLibs/setup.test.hs PackageTests/Regression/T4270/setup.test.hs

Notice, previously we were not testing this combination (GHC 8.10 on MacOS)

I will investigate the failures of 9.0.2 on Linux. You see a green tick overall because I have temporarly marked all combinations as experimental (so I can test them in parallel rather than in sequence).

andreabedini:
What the heck is [this](https://github.com/haskell/cabal/runs/5178990935?check_suite_focus=true#step:10:1382) :thinking: 

```
haddock: internal error: haddock:iface
  error, called at utils/haddock/haddock-api/src/Haddock/Interface.hs:178:16 in main:Haddock.Interface
```

jneira:
> i wonder why a matrix was not used in the first place, were they not available since the beginning of gha?

To answer myself i think the goal of that was being independent of the ci provider. `cabal` has used several ci providers even at same time (travis, appveyor, gha, gitlab) and maybe it could change again in the future. We are using gha and gitlab right now but the latter only uses one ghc version.
The more logic you put in the ci provider configuration the more are you lock in it. With this pr the list of ghc used for the matrix which is the axis of the build will live in the gh configuration instead in a ci provider agnostic program.

That said we should also consider:
* All ci providers offer a native way to set up a matrix of jobs
* In the gha case there is the alternative of externalize the list of ghcs and inject it in the matrix configuration using `toJson`.
  * So you could extract it from the `tested-with` field of the .cabal file for example 
  * contra: no all ci providers has a similar meta evaluation 
* Templating makes miss other advantages of each specific ci provider, like automatic prs from bots (f.e. dependabot)

To be clear *i think this should be merged* but taking in account the context and the drawbacks.

> my long term plan would be 1) remove the validate script 2) unify linux, macos and windows. This likely involves rethinking the meaning of all those flags, so I will defer it to another PR

Given the above considerations i am not sure if removing the validate script would be a good idea. It contains all the build logic, abstracted out from any ci provider and makes ci config steps deadly simple, only passing the needed parameters. It is used for docker and could be used in gitlab or any other ci system.

It is not total os agnostic, as you need a bash shell to run it and `make` to generate it but in practice all os's has a way to do it.

-----------------

Sorry, maybe the pr is not he best place to write down this but there is no issue about.

cabal-dev-scripts/src/GenValidate.hs
jneira: the matrix could be defined dynamically using external data (i would use the list of ghcs in the `tested-with` field of the .cabal file like `haskell-ci`), see https://www.cynkra.com/blog/2020-12-23-dynamic-gha/ for example

jneira:
bootstrap jobs failed with any log or workflow error :thinking: 

https://github.com/haskell/cabal/runs/5178991035?check_suite_focus=true

andreabedini:
> Sorry, maybe the pr is not he best place to write down this but there is no issue about.

The way I see it there is no "lock in", there is adoption. Adopting third-party tools always comes with trade-offs. In this particular case, I believe the effort in abstraing over the CI provider is not worth it.

I'll defer discussing the validate script to a future PR :)

jneira:
Re PackageTests/Regression/T4025/setup.test.hs for linux 8.10.7 and macos 9.0.2

https://github.com/haskell/cabal/blob/dd7a18dfbd78e8b95efb51301737843f5b6d6ea0/cabal-testsuite/PackageTests/Regression/T4025/setup.test.hs#L9-L13

so the test fails cause the dynamically linked produced exe is succesfull. The change from ubuntu-18.04 to ubuntu-latest (20.04 since https://github.com/actions/virtual-environments/issues/1816) migh be related for linux (and maybe for the other failing tests)

.github/workflows/macos.yml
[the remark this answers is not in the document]
andreabedini: > continue-on-error

yes, it should be fixed now. When I wrote this I had written  `continue-on-error: ${{ matrix.experimental }}` but not all matrix elements had `experimental` defined.

andreabedini:
> bootstrap jobs failed with any log or workflow error thinking

I notice they are still queued. Perhaps we need more runners! :)

jneira:
> > bootstrap jobs failed with any log or workflow error thinking
> 
> I notice they are still queued. Perhaps we need more runners! :)

github gave to haskell org 180 parallel runs so maybe someone is busy running things in other projects or some other thing is not working properly :thinking:

andreabedini:
> github gave to haskell org 180 parallel runs so maybe someone is busy running things in other projects or some other thing is not working properly thinking

We do use  GitHub Actions quite intesively, this branch is doing 36 runs. At least I added caching to linux and macos (windows to come later).

> so the test fails cause the dynamically linked produced exe is succesful. The change from ubuntu-18.04 to ubuntu-latest (20.04 since [actions/virtual-environments#1816](https://github.com/actions/virtual-environments/issues/1816)) migh be related for linux (and maybe for the other failing tests). For macos the new tested ghc version could be related.

Thanks for giving it a look. I will check again tomorrow. I didn't have much time for this today.

jneira:
> At least I added caching to linux and macos (windows to come later).

That is a great move, the cabal build does not take so much time as the hls one (where i helped to setup gha) but faster feedback and saving reources are always very welcomed

andreabedini:
I tried running `PackageTests/Regression/T4025/setup.test.hs` on pristine ubuntu 20.04 and 18.04. It passes in both cases unfortunately.

andreabedini:
> **GHC 9.2.1**
> 
> Both [Linux](https://github.com/haskell/cabal/runs/5178990898?check_suite_focus=true) and [MacOS](https://github.com/haskell/cabal/runs/5178990705?check_suite_focus=true) fail because of `StarIsType`. This is good because it's what we wanted to test.

My analysis is wrong. Seeing the warning about StartIsType I thought we were using `-Werror` but we are not. The real error is the following. Also StarIsType warning comes from hackage-security. It would be better to split building the dependencies from building the local packages but that touches the validate script so it has to be deferred.

``
<no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - transformers-0.5.6.2
      - deepseq-1.4.6.0
```

**Comment by Mikolaj** [comment]
Wow, we are not using transformers? In which component? Only on some OSes?

**Comment by andreabedini** [comment]
> Wow, we are not using transformers? In which component? Only on some OSes?

cabal-install apparently, https://github.com/haskell/cabal/runs/5178990898?check_suite_focus=true#step:8:1426

**Comment by andreabedini** [comment]
Anyway, this https://github.com/haskell/cabal/pull/7952#issuecomment-1038748275 is the cause of the failing tests (the ones we really care about 9.0.2 on Linux, which passed before). I can reproduce locally both on ubuntu 18.04 and 20.04 but I have no clue what causes it (and why it worked before).

File: .github/workflows/linux.yml [hunk 5]
```diff
@@ -15,520 +10,97 @@ on:
...
+          cabal-version: latest # default, keeping for visibility
+
+      #  See the following link for a breakdown of the following step
+      #  https://github.com/haskell/actions/issues/7#issuecomment-745697160
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ${{ steps.setup-haskell.outputs.cabal-store }}
+            dist-newstyle
+          key: ${{ runner.os }}-${{ matrix.ghc }}-${{ github.sha }}
```
jneira, line 72: this cache key is too restrictive, hashing all files which can control the cache (*.cabal, *.project) would be a still to broader one, doing for the specific config files we know are being used would be beven better

andreabedini:
Actually I don't see those tests been run at all in the current CI. Not skipped, not mentioned.This is confusing.

current ci: https://github.com/haskell/cabal/runs/5203722340?check_suite_focus=true#step:12:1
new ci: https://github.com/haskell/cabal/runs/5178990935?check_suite_focus=true#step:10:1

new ci fails on `PackageTests/Backpack/Includes2/setup-external.test.hs`
current ci doens't mention `PackageTests/Backpack/Includes2/setup-external.test.hs`

.github/workflows/linux.yml
jneira: ah and the hackage index being used could be useful too

andreabedini: This is a bit of a trick, the comment I left above is relevant.

Our cache is basically append only. We want GitHub to always save and restore the cache. To get it to do that we use a key that (as you notice) is unique but we also instruct it to use as a backup anything that matches up to `${{ runner.os }}-${{ matrix.ghc }}-`. GitHub will pick the most recent matching key.

If we used a same key (for example just `${{ runner.os }}-${{ matrix.ghc }}`) GitHub would not update the cache.

So the alterative is 1) either we are 100% precise (and I guess we could) 2) or we make it append only with the trick above.

See https://github.com/haskell/actions/issues/7#issuecomment-745697160 for the original discussion

jneira: Thanks for the pointer to the coment, very interesting
Maybe i am missing something obvious but usually all prs are adding new commits so the real used cache is the fallback of the previous (or whatever) commit. It will take lot of cache space as each commit in the repo will augment it and the limit is 20 Gb, after that gh will start to evict old caches. But most of them will be identical (not sure if gh is able to optimize it underneath). If you returns to a old pr you will get a first rebuild for sure.
Otoh one commit could have very different dependencies and they change in the cabal store not only by version but with thing like ghc-options. Interleaved commits between prs will invalidate the cache for all them causing unnecessary rebuilds.


The comment talks about using the cabal.project.freeze, which is even better than hashing cabal.project and .cabal files, as it captures cli flags. I nfact it is what we are using in hls (with the hackage index as base key):

https://github.com/haskell/haskell-language-server/blob/96ea854debd92f9a54e2270b9b9a080c0ce6f3d1/.github/actions/setup-build/action.yml#L131-L142
But well cabal dependency set is relatively small and usually there are no lot simultaneous prs so maybe it can work in practice.

jneira: forgot I said about interleaved commits, as all commits are appending library with duff hashes in the cabal store, all of them will have what they need

andreabedini:
@fgaz @Mikolaj @jneira ok, it's the difference between 9.0.1 and 9.0.2. I can reproduce locally that on 9.0.2 the tests mentioned above fail while on 9.0.1 they pass. I am using the validate script on Ubuntu 20.04. Both ghc versions have been installed with ghcup.

andreabedini:
I have marked the ghc versions causing failures as experimental, I believe these are real new failures we didn't detect before. I also added GHC 9.0.1 which should pass and GHC 9.0.2 (marked as experimental).

Mikolaj:
@Kleidukos: any idea what the error may mean?

@andreabedini: Now that you've overhauled CI and secret ancient tests are revealed, the fate dictates that you need to fix them all! 

Kidding. :D

Could you kindly open separate tickets for everything that was newly failing or looks suspect for any reason? Thank you.

.github/workflows/linux.yml
andreabedini: Rephrasing, GitHub cache is immutable, given a key, once it's built, it is assumed to be correct, complete and "done".

On the other hand, cabal store is append only, it is its own cache if you want. Each package gets built only once (per configuration).

The trick is to make the former behave like the latter.

You raise an interesting point and I believe what you are saying is correct. If I understand you, you are saying that given the cache is not linked to the tree at all, the most recent cache might be inappropriate for the branch you are building.
Nevertheless I don't believe this is going to be a problem, we only use order of ~50MB per cache entry (link](https://github.com/haskell/cabal/runs/5218127255?check_suite_focus=true#step:4:10)). I think with 20GB we can build half hackage and keep it cached :-)

Using `cabal.project.freeze` would mean that any change in the plan will throw away the entire cache. E.g. if we change an option in `cabal-install` we will end up rebuilding `criterion` (a dependency of `cabal-benchmark`).

If anything the real issue I see with this setup is that the cache doens't get saved unless the workflow succeed. This has caused unneeded rebuilding of the dependencies, for example. Ideally, I would like the cabal store be persisted regardless of the workflow success. It's a known issue with `actions/cache`, see [here](https://github.com/actions/cache/issues/92).

andreabedini:
Alright,

I created #7987, #7988, #7989 and mentioned next to `experimental: true` in the workflows. I also tested locally on my MacOS laptop and reproduced the failures where I expect them.

One last bit of confusion is the difference between two recent runs on MacOS with GHC 9.0.2

- https://github.com/haskell/cabal/runs/5178990727?check_suite_focus=true
- https://github.com/haskell/cabal/runs/5218126500?check_suite_focus=true

One fails to build with the message

```
<no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - transformers-0.5.6.2
      - deepseq-1.4.5.0
```

while the other builds ok but fails the tests with #7987

Mikolaj:
> One last bit of confusion is the difference between two recent runs on MacOS with GHC 9.0.2

I see both built `unit-tests`, but one reported `not needed for compilation` and the other didn't. The only solution to the puzzle I can think of is that the solver picked a different build plan (the solver is not determined solely by the .cabal, project file and Hackage) and in one of them the packages are not needed. The fix would be to tighten the bounds so that we never need the extra packages and then remove them from dependencies.

andreasabel:
I am also a bit worried about removing the Ci generation layer.  Concerns have already voiced:
- @fgaz: https://github.com/haskell/cabal/pull/7952#discussion_r799554837
- @jneira: https://github.com/haskell/cabal/pull/7952#issuecomment-1039943611

The Haskell community has been severely burned by the treason of Travis CI.  CI just vanished for hundreds (or thousands?) of Haskell projects when travis-ci.org was shut down.  The community is still recovering from this blow.  I donated a GHA CI to dozens of Haskell projects.  This was possible through @phadej 's heroic migration of [Haskell-CI](https://github.com/haskell-CI/haskell-ci/) from Travis to GHA.  
Maybe CI on GHA will not be free anymore at some point in the future.  If the CI is generated by a tool, only the tool needs to be updated to a new CI backend.  (It is like programming in C rather than in assembler: you can migrate to a new machine by updating the compiler, rather by rewriting every single program...)
Of course, we cannot use Haskell-CI here; complex projects need a custom setup.  But if there is already a templating infrastructure, why get rid of it?
I don't know much about it, but it wasn't too hard to modify it when I did #7726.

Ideally, one can run the testsuites both locally and via CI, and the way the mechanism is the same (only that on the CI setup code is needed).
[This is more an outsider than an insider comment, arguing from principles.]

Mikolaj:
My 2 cents are that I emergency-fixed the current CI sediment a few times and attempted to upgrade it a couple of times, mostly failing, and that was a miserable experience. This is just old and unmaintained. Some parts more than others, but nobody makes the effort to understand it and document which bits should stay and why. Give that, I welcome a refresh, even if the method is to kill everything and start from scratch.

andreasabel:
> attempted to upgrade it a couple of times, mostly failing, and that was a miserable experience. This is just old and unmaintained.

Fair enough.  I won't stand in the way.
As an aside, in the GHA CI for Agda I did some minimal templating, to enable some YAML features that GHA do not support, like sharing subtrees.  The "templating" is just a `json2yaml . yaml2json` which normalizes YAML-DAGs to YAML-trees by going through JSON.  This is used to [share triggers between `push` and `pull_request`](https://github.com/agda/agda/blob/d4333bbfbdb93fb9cc2ea321438bf552ec0adee6/src/github/workflows/cabal.yml#L9-L18).  I wish that YAML allowed more modularity, like importing and let-binding.  Without some templating, there is a lot of cut-and-paste which can easily become inconsistent across workflows.

.github/workflows/linux.yml
jneira: yeah with that small dep footprint you can let grow the store quite time
In hls each key matrix has 200mb and we can have ~20 combinations of os+ghc so we have to reuse cache more aggressively even having to rebuild part of deps
But even with 20 mb you will reach the limit at some point and the actual cache key prefix is totally fixed. A manual reset changing the prefix will be needed.

In hls we have pinned the hackage index in the cabal.project and using it in the cache key gives us a pretty stable dependency set. We update the index from time to time and that make generate a clean cache shared for all PRS. So changes in the freeze file doesnt suppose a full rebuild almost never.

But hls scenario is pretty different from cabal one.


I have to take a look to the cache on pr failure issue but in theory the failure can be caused precisely by introducing a bad dep in the cache or even a corruption in the store (improbable but still, it happened to me locally) making it unusable and forcing you to invalidate manually changing the prefix. It can be avoided building only dependencies in a separate job, save the cache if it is successful and then build the local packages in other job reusing the cache.

jneira:
>  Bootstrap tests still don't run (as in, they get stuck in the queue and get cancelled after two days). 

Weirs, other prs have the bootstrap workflow running after queuing this one: https://github.com/haskell/cabal/actions/workflows/bootstrap.yml

.github/workflows/bootstrap.yml
```diff
@@ -15,37 +10,21 @@ on:
...
-    name: Bootstrap on macOS
-    runs-on: macos-latest
+  bootstrap:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+    name: Bootstrap ${{ matrix.os }}
+    runs-on: $${ matrix.os }}
```
jneira, line 20:
```suggestion
    runs-on: ${matrix.os}
```

that might be the culprit

andreabedini: oh dear

andreabedini: Thanks @jneira <3

jneira: Maybe it might be related with self host runners? maybe they can have arbitrary names, idk

andreabedini: Yes I think so. and I think somehow that was expanded to something like and empty string, so I didn't notice any bad name anywhere. My fat fingers!

jneira:
> The Haskell community has been severely burned by the treason of Travis CI. CI just vanished for hundreds (or thousands?) of Haskell projects when travis-ci.org was shut down. The community is still recovering from this blow. I donated a GHA CI to dozens of Haskell projects. This was possible through @phadej 's heroic migration of [Haskell-CI](https://github.com/haskell-CI/haskell-ci/) from Travis to GHA.

That was a unsung epic travispocalypse, thank you for your work.
Lot of work was invested in travis and its sunset was really painful. The lessons from that pain should not be forgotten.
However here in Spain we have a proverb: "Gato escaldado, del agua fria huye" (in english it would be "once bitten twice shy" i think)

I think there are two directions in ci lock in (or adoption :-P):
1. Up-down: this is about the top level *inputs* of workflows. The matrix of ghc versions and oss and the different parameters for each combination of them.
  * That was abstracted from ci thanks to GenValidate and that is which we are giving out to github with this pr
  * They are *essential* for the ci logic but
    * The size of data is relatively small and dont vary too much
    * They are not actual code so no ifs, no loops
    * All ci providers have similar ways to encode them so it is relatively easy convert between them
2. Bottom: The actual individual steps of the workflow, full of logic and default values which needs to be programmed
  * Imho that logic should live, in that precise order: 
     1. haskell code (test suites, benchmarks, build config)
     2. other language scripts: sh, python, etc
     3. using ci specific features
   * cabal uses 1 and 2 (with validate.sh and make etc) and i think we should keep them (but we can discuss that in other issue)

And for the top inputs we always could have it encoded in a ci agnostic data format and inject it in the ci provider, if ci allows to do it (see the linked above  https://www.cynkra.com/blog/2020-12-23-dynamic-gha/)

.github/workflows/linux.yml
andreabedini: perhaps it could be worth exploring using an external cache, it shouldn't be hard do to with S3 `aws s3 sync s3://my-cache` and back.

jneira:
> They are not actual code so no ifs, no loops

For completeness, they are not code but it could be computed, deriving it from other sources, to avoid duplication of such sensistive data.
For example we could extract the ghc versions from `tested-with` like `haskell-ci` does. But being a relatively small dataset, encoding its generation to avoid oversights and have to do greps should be very cheap to worth, imho.

jneira:
I have a minor nitpick: the experimental builds are marked as failing, so the entire build is marked that way. Those workflows will not be added to branch protection ruels and prs will be merged anyways but aesthetically it does not look nice
(Btw we have to remember add the succesful workflows to branch protection rules once this is merged and rebase the rest of prs // @Mikolaj)

Otoh, have them red is a compelling reason to work on fixing this so i am not strongly against it, thoughts?

andreabedini:
> I have a minor nitpick: the experimental builds are marked as failing, so the entire build is marked that way.

Only the "branches" are marked as failing, the entire workflow is green if all the required branched pass, see https://github.com/haskell/cabal/actions/runs/1863088120

> For example we could extract the ghc versions from `tested-with` like `haskell-ci` does.

This is a good idea (especially in relation to the post about dynamic workflows you posted). But, again, my intention is to try to simplify all variations rather than capturing them perfectly.

.github/workflows/linux.yml
jneira: i think it will not be as fast to restore and save it as the github one, located very near but maybe i am wrong (or the perf gain is not significant, to investigate)

cabal-dev-scripts/src/GenValidate.hs
```diff
@@ -1,159 +0,0 @@
...
-            writeFile tgt w
-        _ -> putStrLn "Usage source.yml.zinza target.yml"
-
-mangleChar :: Char -> Char
-mangleChar '.' = '_'
-mangleChar c   = c
-
-defSteps :: [String]
-defSteps =
-    [ "print-config"
```
jneira, line 74: hmm steps also lives here, sad to see they go away

jneira (approved):
Many thanks again, really a great work

jneira:
This pr cant be merged automatically, as it changes required jobs names as commented above we can change branch protection rules to match new identifiers and merge it (to check both pr jobs and updated ruels) and then aaaall prs need to be rebased to pass ci

//cc @Mikolaj

fgaz (approved):
Looks good! I'll merge tomorrow if nobody objects. Specifically, let's squash-merge since there are a lot of fixup commits. @andreabedini if you don't want them squashed please tell!

fgaz:
Thanks a lot @andreabedini!

Now I just have to re-select the required jobs... please ping me or @Mikolaj if prs are stuck because of that

.github/workflows/linux.yml
```diff
@@ -15,520 +10,101 @@ on:
...
-        run: sh validate.sh -j 2 -w ghc-8.8.4 -v --lib-only -s build
-      - name: Validate lib-tests
-        run: sh validate.sh -j 2 -w ghc-8.8.4 -v --lib-only -s lib-tests
-      - name: Validate lib-suite
-        run: sh validate.sh -j 2 -w ghc-8.8.4 -v --lib-only -s lib-suite
-      - name: Validate lib-suite-extras --extra-hc /opt/ghc/7.0.4/bin/ghc-7.0.4
-        run: sh validate.sh -j 2 -w ghc-8.8.4 -v --lib-only -s lib-suite-extras --extra-hc /opt/ghc/7.0.4/bin/ghc-7.0.4
-      - name: Validate lib-suite-extras --extra-hc /opt/ghc/7.2.2/bin/ghc-7.2.2
-        run: sh validate.sh -j 2 -w ghc-8.8.4 -v --lib-only -s lib-suite-extras --extra-hc /opt/ghc/7.2.2/bin/ghc-7.2.2
-      - name: Validate lib-suite-extras --extra-hc /opt/ghc/7.4.2/bin/ghc-7.4.2
```
phadej, line 533: I don't see these steps in new CI setup. Is `cabal-install` dropping support for older GHC.

AFAICS the oldest GHC you try to run tests with is GHC-8.0

Is this correct and intentional, @fgaz, @Mikolaj ?

jneira: Afaics it is not intentional and should be restored.

jneira:
> Thanks a lot @andreabedini!
> 
> Now I just have to re-select the required jobs... please ping me or @Mikolaj if prs are stuck because of that

The linux and macos validate jobs has the same name and in my experience tha does not play well with branch protection rules, which needs unique jobs names

.github/workflows/linux.yml
jneira: As we are deprecating *build* cabal with ghc < 8.0, all older version should be checked using a newer build (including 7.8 and 7.10) like < 7.6 before this pr

jneira: the oldest ghc instalable with ghcup is 7.10.3, so we have to continue using apt if we want to keep older ones

andreabedini: @phadej It was intentional, following #7534 but now I think I have misunderstood.

We are deprecating building with GHC < 8.0 but we still want to test cabal can use GHC 7.8 and 7.10. Is this correct?

It's a bit of headache (which GHC version we should use to compile the cabal that compiles with what GHC version?) but it can definitely be readded.

Mind adding an issue to track? Please mention me there.

andreabedini:
Thanks a lot everyone <3 It's been a jouney for me!

.github/workflows/linux.yml
Mikolaj: @andreabedini: great job! I might have mislead you re #7534. Let's clean up the fallout later on.

Issue #7828, opened 2021-11-19
Cabal fails to set object format when joining objects on Windows


bgamari:
**Describe the bug**
Due to limitations of the PE object format, it is in general necessary to pass `--oformat=pe-bigobj-x86-64` to the linker when joining Haskell object files on Windows lest one encounter linker errors due to too many sections. This is particularly true when split-sections is enabled (as it produces significantly more sections than otherwise would be produced).

Currently Cabal fails to do this, resulting in [GHC#20682](https://gitlab.haskell.org/ghc/ghc/-/issues/20682).

**To Reproduce**
Steps to reproduce the behavior (on Windows)

```
$ git clone https://gitlab.haskell.org/bgamari/T20506-repro
$ cd T20506-repro
$ cabal build -v3
```
The build will fail with:
```
[...]

PR #7960, opened 2022-02-06
Use ghc --merge-objs when available to join object files

bgamari:
Previously Cabal would rely on its own object-merge logic to produce
GHCi libraries. This was not ideal as it meant that this logic was
replicated in both GHC and Cabal. Moreover, Cabal's implementation
historically hasn't handled the more subtle aspects of interacting with
the linker, resulting in issues like #7828.

In GHC 9.4 we have introduced a new compiler mode, `--merge-objs`, which
exposes GHC's object merging mechanism. Here we teach Cabal to take
advantage of this mechanism when it is available.

---

### DIFF
--- a/Cabal/src/Distribution/Simple/GHC.hs
+++ b/Cabal/src/Distribution/Simple/GHC.hs
@@ -934,15 +934,13 @@ buildOrReplLib mReplFlags verbosity numJobs pkg_descr lbi lib clbi = do
       whenVanillaLib False $ do
         Ar.createArLibArchive verbosity lbi vanillaLibFilePath staticObjectFiles
         whenGHCiLib $ do
-          (ldProg, _) <- requireProgram verbosity ldProgram (withPrograms lbi)
-          Ld.combineObjectFiles verbosity lbi ldProg
+          combineObjectFiles verbosity lbi ghcProg
             ghciLibFilePath staticObjectFiles
 
       whenProfLib $ do
         Ar.createArLibArchive verbosity lbi profileLibFilePath profObjectFiles
         whenGHCiLib $ do
-          (ldProg, _) <- requireProgram verbosity ldProgram (withPrograms lbi)
-          Ld.combineObjectFiles verbosity lbi ldProg
+          combineObjectFiles verbosity lbi ghcProg
             ghciProfLibFilePath profObjectFiles
 
       whenSharedLib False $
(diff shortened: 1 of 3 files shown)


Mikolaj:
Amazing! I suppose it doesn't need documenting? How about a tiny changelog file and a tiny test?

bgamari:
It's technically an internal, non-user-visible change but I suppose a changelog entry won't hurt. I'm a bit unclear on what a test would look like; I would have thought that this functionality is exercised by nearly all of the "package tests" currently?

Mikolaj:
Great, so I guess a minimal changelog file short suffice.

Mikolaj:
I will write the changelog file, but please somebody review --- this is a tricky change.

jneira:
the main concern could be we are not testing it for ghc > 9.2 but if it works for older ghcs in ci so the falback works i guess we will hit the possible bug in the other code branch sooner or later

or has been this tested in ghc itself?

jneira:
> I will write the changelog file, but please somebody review --- this is a tricky change.

please do, it would help to move this forward

Mikolaj:
I probably can't push to the branch, so here's the changelog (in the worst case I will merge it in a subsequent PR):
```
synopsis: Use 'ghc --merge-objs' whenever available, instead of internal cabal hacks, to join object files
packages: Cabal
issues: #7828
prs: #7960
```

jneira:
@Mikolaj by default when you open a pr in github it allows upstream maintainers to do commits in the pr branch, you have to protect it on purpose so most pr branches are mofifiable but maintainers

jneira (approved):
lgtm, thanks @bgamari hope you dont mind i added the suggested changelog

jneira:
@Mergifyio rebase master

jneira:
It seems it closes #7828, at least for ghc >= 9.4

bgamari:
Thank you, @jneira!

I have been holding off on pushing this MR forward since I have recently been engaged in a related set of work, moving GHC to use an LLVM-based toolchain on Windows, fixing a number of long-standing issues. Sadly, one of the compromises that we have had to make is to drop support for GHCi objects on Windows entirely due to deficiencies in the new (and old) linker. I have opened https://github.com/haskell/cabal/pull/8062 disabling support in Cabal.

As a result this MR is no longer needed as the problem that it solves will no longer exist with GHC 9.4 and later

jneira:
@bgamari don't worry, it seemed to me than leverage that option would be good on its own, as ghc will know more about how to do the linking, as described in the initial comment
could not be this merge anyways to make more robust the ghc part?

Mikolaj:
@bgamari: any harm if we merge this?

Mikolaj:
@jneira: we decided, while this PR makes some sense, it's not strictly needed any more, and the stuff works as is, so it's safer not to touch it. However, if needed, e.g., if somebody has to change this bit of code, this PR is there.

PR #7962, opened 2022-02-07
Fix missing space in error message

andreasabel:
Add the missing space in error messages like:
```
  (conflict: requires NumDecimalswhich is not supported)
```


fgaz (approved):

andreasabel:
Btw @Mikolaj how do you keep track which commits/PRs are going into the next release?

Mikolaj:
> Btw @Mikolaj how do you keep track which commits/PRs are going into the next release? (For Agda, we use milestones for this.)

I guess the system here is that we cut the branch at some point and then only backport to it. When cutting, all of master goes there.

Which implies, we should not merge semi-autonomous steps of big overhauls just before cutting a release. I don't think we do ATM.

Issue #7776, opened 2021-10-26
Regression on master: building hashable-1.3.0.0 fails with `Duplicate modules in benchmark`


andreasabel:
Current master (e74a53b3d8b1a52cb86a8232e2e745d1bd64a8d7) of cabal-install fails to build `hashable-1.3.0.0`:
```
$ cabal install hashable-1.3.0.0
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
In order, the following will be built (use -v for more details):
 - hashable-1.3.0.0 (lib) (requires build)
Configuring library for hashable-1.3.0.0..
Error: cabal: Duplicate modules in benchmark: Data.Hashable.RandomSource

Error: cabal: Failed to build hashable-1.3.0.0. The failure occurred during
the configure step.

$ cabal --version
cabal-install version 3.7
compiled using version 3.7.0.0 of the Cabal library 
```

Build succeeds with version 3.6:
```
$ cabal-3.6 install -w ghc-8.10.7 hashable-1.3.0.0
...
Installing library in /Users/abel/.cabal/store/ghc-8.10.7/incoming/new-49369/Users/abel/.cabal/store/ghc-8.10.7/hshbl-1.3.0.0-02ec191e/lib

$ cabal-3.6 --version
cabal-install version 3.6.0.0
compiled using version 3.6.1.0 of the Cabal library 
```

Build also succeeds with `Setup.hs`:
```
$ cabal get hashable-1.3.0.0
$ cd hashable-1.3.0.0
$ runhaskell Setup.hs configure
$ runhaskell Setup.hs build
```
(However, `runhaskell Setup.hs configure --enable-benchmarks` does not succeed to construct a build plan.)

This is the content of the `cabal.project.local` file:
```
$ cat cabal.project.local 
ignore-project: False
benchmarks: True
```

gbaz:
Hrm, on reflection I'm surprised this stuff was working before, as I thought the patch in question made the checks strictly less strong.

I guess it strengthened some things and rendered others weaker. As such, yeah, I agree there's a problem here. If someone can explain how cabal's prior checks ever accepted examples such as above I'd appreciate it -- it seems to me that `flattenPackageDescription` should necessarily have created duplicate modules in such cases: https://downloads.haskell.org/~ghc/latest/docs/html/libraries/Cabal-3.6.0.0/Distribution-PackageDescription-Configuration.html#v:flattenPackageDescription

Perhaps the semigroup combination of package lists is sort of funny?

[...]

PR #7966, opened 2022-02-10
Fix the duplicate modules check

gbaz:
Resolves https://github.com/haskell/cabal/issues/7776

Now if a package _may_ have duplicate modules (subject to conditionals), cabal warns on this.

But if a package _necessarily_ has duplicate modules (because they are not guarded by conditionals), cabal will error.

### DIFF
--- a/Cabal/src/Distribution/PackageDescription/Check.hs
+++ b/Cabal/src/Distribution/PackageDescription/Check.hs
@@ -2144,15 +2144,21 @@ checkDuplicateModules pkg =
     checkTest  = checkDups "test suite" testModules
     checkBench = checkDups "benchmark"  benchmarkModules
     checkDups s getModules t =
-               let libMap = foldCondTree Map.empty
-                                         (\(_,v) -> Map.fromListWith (+) . map (\x -> (x,(1::Int))) $ getModules v )
-                                         (Map.unionWith (+)) -- if a module may occur in nonexclusive branches count it twice
-                                         (Map.unionWith max) -- a module occurs the max of times it might appear in exclusive branches
+               let sumPair (x,x') (y,y') = (x + x' :: Int, y + y' :: Int)
+                   mergePair (x, x') (y, y') = (x + x', max y y')
+                   maxPair (x, x') (y, y') = (max x x', max y y')
+                   libMap = foldCondTree Map.empty
+                                         (\(_,v) -> Map.fromListWith sumPair . map (\x -> (x,(1, 1))) $ getModules v )
+                                         (Map.unionWith mergePair) -- if a module may occur in nonexclusive branches count it twice strictly and once loosely.
+                                         (Map.unionWith maxPair) -- a module occurs the max of times it might appear in exclusive branches
                                          t
-                   dupLibs = Map.keys $ Map.filter (>1) libMap
-               in if null dupLibs
-                    then []
-                    else [PackageBuildImpossible $ "Duplicate modules in " ++ s ++ ": " ++ commaSep (map prettyShow dupLibs)]
+                   dupLibsStrict = Map.keys $ Map.filter ((>1) . fst) libMap
+                   dupLibsLax = Map.keys $ Map.filter ((>1) . snd) libMap
+               in if not (null dupLibsLax)
+                      then [PackageBuildImpossible $ "Duplicate modules in " ++ s ++ ": " ++ commaSep (map prettyShow dupLibsLax)]
+                      else if not (null dupLibsStrict)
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
I don't understand the code, but I tested the product on `hashable-1.2.7` and `hashable-1.3` and found now that it works as expected: build and install succeed, check gives this warning:
```
Warning: Potential duplicate modules (subject to conditionals) in benchmark:
Data.Hashable.RandomSource
```

Cabal/src/Distribution/PackageDescription/Check.hs
```diff
@@ -2144,15 +2144,21 @@ checkDuplicateModules pkg =
...
-                   dupLibs = Map.keys $ Map.filter (>1) libMap
-               in if null dupLibs
-                    then []
-                    else [PackageBuildImpossible $ "Duplicate modules in " ++ s ++ ": " ++ commaSep (map prettyShow dupLibs)]
+                   dupLibsStrict = Map.keys $ Map.filter ((>1) . fst) libMap
+                   dupLibsLax = Map.keys $ Map.filter ((>1) . snd) libMap
+               in if not (null dupLibsLax)
+                      then [PackageBuildImpossible $ "Duplicate modules in " ++ s ++ ": " ++ commaSep (map prettyShow dupLibsLax)]
+                      else if not (null dupLibsStrict)
+                           then [PackageDistSuspicious $ "Potential duplicate modules (subject to conditionals) in " ++ s ++ ": " ++ commaSep (map prettyShow dupLibsStrict)]
```
fgaz, line 2160: Shouldn't this be `PackageDistInexcusable`, since in this case there's a branch where modules are duplicated?

gbaz: Inexecusable would prevent hackage upload. Given that people can write multiple if branches that are not obviously nonexclusive, but whose conditions (e.g. ghc versions) logically can't all be satisfied at once, this would make the check more restrictive than prior. So e.g. ghc-src-gen could be built but not uploaded. One could argue this would be cleaner, but my goal was to clarify existing checks and allow more clearly legit stuff to pass, not to add any new restrictions.

gbaz:
Master without this patch has 1116 build impossible With it, there are only 525 build impossible, with the remaining 591 are now marked as suspicious.

So that seems about right.

fgaz:
What is the number with #7616 reverted?

gbaz:
Damn it, i ran that test the other day and then forgot to post the results, which are now on a lost terminal. But the build impossible was no less than with this patch, and perhaps even the exact same -- i.e. the results tied out as one would hope.

jneira:
@MergifyIO rebase master

jneira (commented):
not fully understand the code neither but trust it 🙂
but I think this really need a changelog entry and a regression test, based in the hashable case f.e., bonus if it tests the warning

jneira (changes requested):
I think this really need a changelog entry and a regression test, based in the hashable case f.e., bonus if it tests the warning

jneira:
@Mergifyio rebase master

jneira:
Full hackage test check data for each case

* with master:

```
D:\ws\haskell\cabal>cabal run Cabal-tests:hackage-tests -- check
Up to date
Reading index from: D:\cabal\packages\hackage.haskell.org\01-index.tar
157946 files processed
42067 files have lexer/parser warnings
115687 files have check warnings
1118 build impossible
7980 build warning
51952 build dist suspicious
36949 build dist suspicious warning
17688 build dist inexcusable
```

* with this pr:

```
157946 files processed
42067 files have lexer/parser warnings
115687 files have check warnings
525 build impossible
7980 build warning
52545 build dist suspicious
36949 build dist suspicious warning
17688 build dist inexcusable
```

* with #7616 reverted (https://github.com/jneira/cabal/tree/check-7616-reverted)

```
157946 files processed
42067 files have lexer/parser warnings
114905 files have check warnings
336 build impossible
7980 build warning
51952 build dist suspicious
36949 build dist suspicious warning
17688 build dist inexcusable

**Comment by gbaz** [comment]
@Mergifyio rebase master

**Review by jneira — APPROVED** [review]
lgtm, thanks for the fix

**Comment by andreasabel** [comment]
Great to see this fixed!

BEFORE:
```
Configuring library for snap-server-1.1.2.0..
Error: cabal: Duplicate modules in library: System.SendFile

Duplicate modules in executable: System.SendFile

Duplicate modules in executable: System.SendFile

Duplicate modules in test suite: System.SendFile, System.SendFile.Tests
```

AFTER: works!

Issue #5952, opened 2019-03-20
cabal new-update fails for head.hackage


AndreasPK:
Either I get an invalid response from the server, or cabal fails with a file locking issue.

```
Andi@Horzube MINGW64 /e/bench_hc_libs/libs/perf-xml/build
$ cabal v2-update 'hackage.haskell.org,2019-02-19T16:58:55Z'
Downloading the latest package list from hackage.haskell.org
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2019-02-20T16:58:55Z'

Andi@Horzube MINGW64 /e/bench_hc_libs/libs/perf-xml/build
$ cabal v2-update 'head.hackage,2019-02-19T16:58:55Z'
Downloading the latest package list from head.hackage
C:\ghc\msys64\tmp\tra2CC0: DeleteFile "\\\\?\\C:\\ghc\\msys64\\tmp\\tra2CC0": permission denied (The process cannot access the file because it is being used by another process.)

Andi@Horzube MINGW64 /e/bench_hc_libs/libs/perf-xml/build
$ cabal v2-update
Downloading the latest package lists from:
- hackage.haskell.org
- head.hackage
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2019-02-19T16:58:55Z'
C:\ghc\msys64\tmp\tra3EEA: DeleteFile "\\\\?\\C:\\ghc\\msys64\\tmp\\tra3EEA": permission denied (The process cannot access the file because it is being used by another process.)

Andi@Horzube MINGW64 /e/bench_hc_libs/libs/perf-xml/build
$ cabal v2-update
Downloading the latest package lists from:
- hackage.haskell.org
- head.hackage
Unexpected response 416for http://head.hackage.haskell.org/01-index.tar.gz
```

The unexpected response might not be a Cabal bug. But at least the failing DeleteFile seems like one.

This was with cabal-install built from commit 8e903011
After updating cabal-install to HEAD I only got unexpected response.

After cleaning ~/.cabal I was able to update successfully.

23Skidoo:
This is probably a server issue. HTTP 416 means "range not satisfiable".

gbaz:
debugging the delete file issue would require knowing which transport was used, which can be detected with a verbose flag.

But if the delete file issue doesn't occur with HEAD, then I guess it doesn't need to be debugged?

mpickering:
This happened to me as well.

AndreasPK:
```
[...]

PR #7970, opened 2022-02-11
ensure range headers are only sent when from < to

gbaz:
Speculative fix for https://github.com/haskell/cabal/issues/5952

Should be a strict improvement even if it doesn't resolve that ticket, as it just puts in place a sanity check to ensure range requests are not malformed.

### DIFF
--- a/cabal-install/src/Distribution/Client/Security/HTTP.hs
+++ b/cabal-install/src/Distribution/Client/Security/HTTP.hs
@@ -119,11 +119,16 @@ mkRangeHeader from to = HTTP.Header HTTP.HdrRange rangeHeader
     rangeHeader = "bytes=" ++ show from ++ "-" ++ show (to - 1)
 
 mkReqHeaders :: [HC.HttpRequestHeader] -> Maybe (Int, Int) -> [HTTP.Header]
-mkReqHeaders reqHeaders mRange = concat [
+mkReqHeaders reqHeaders mRange' = concat [
       tr [] reqHeaders
     , [mkRangeHeader fr to | Just (fr, to) <- [mRange]]
     ]
   where
+    -- guard against malformed range headers.
+    mRange = case mRange' of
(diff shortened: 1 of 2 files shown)


cabal-install/src/Distribution/Client/Security/HTTP.hs
```diff
@@ -119,11 +119,16 @@ mkRangeHeader from to = HTTP.Header HTTP.HdrRange rangeHeader
...
-mkReqHeaders reqHeaders mRange = concat [
+mkReqHeaders reqHeaders mRange' = concat [
       tr [] reqHeaders
     , [mkRangeHeader fr to | Just (fr, to) <- [mRange]]
     ]
   where
+    -- guard against malformed range headers.
+    mRange = case mRange' of
+        Just (fr, to) | fr >= to -> Nothing
+        _ -> mRange'
```
robx, line 130: Isn't this wrong? If we're being asked to get an empty range, I'd imagine the caller's expectation (of `getRange`) would be to end up with a `BodyReader` that reads an empty `ByteString`. By leaving out the range header, they'd instead get the whole file.

(Also seems like something that might be better handled in hackage-security.)

gbaz: this is fine because we already handle the case where we request a range but the server doesn't understand it and gives back the whole thing (the response code is different between the two cases)

jneira:
@Mergifyio rebase master

Kleidukos (approved):

jneira:
@Mergifyio rebase master

Mikolaj (changes requested):
Looks fine. Is it tested, though?

gbaz:
manually tested. any other sort of test runs into the "we don't have a path for testing network requests against hackage servers" issue.

Mikolaj (approved):
FIne.

